PR #6289, opened 2015-02-12
Database: log how many rows DB result has

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-836

Extends sampled logging of SQL queries introduced in #3787

@pchojnacki / @owend / @drsnyder

### DIFF
--- a/includes/db/Database.php
+++ b/includes/db/Database.php
@@ -3506,9 +3506,11 @@ abstract class DatabaseBase implements DatabaseType {
 		$queryId = MWDebug::query( $sql, $fname, $isMaster );
 		$start = microtime(true);
 
+		// Wikia change: DatabaseMysql returns a resource instead of ResultWrapper instance
+		/* @var $ret resource */
 		$ret = $this->doQuery( $sql );
 
-		$this->logSql( $sql, $fname, microtime(true) - $start, $isMaster );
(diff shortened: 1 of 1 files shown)


includes/db/Database.php
```diff
@@ -3519,21 +3521,24 @@ protected function executeAndProfileQuery( $sql, $fname, $isMaster ) {
...
 		if ($this->getSampler()->shouldSample()) {
 			$this->getWikiaLogger()->info( "SQL $sql", [
 				'method'      => $fname,
 				'elapsed'     => $elapsedTime,
+				'num_rows'    => ( $ret instanceof ResultWrapper ? $ret->numRows() : mysql_num_rows( $ret ) ),
 				'cluster'     => $wgDBcluster,
 				'server'      => $this->mServer,
 				'server_role' => $isMaster ? 'master' : 'slave',
 				'db_name'     => $this->mDBname,
+				'exception'   => new Exception(), // log the backtrace
```
wladekb: we keep adding exception here and there, maybe we should add it by default in the logger?

macbre: I don't think backtrace is required in all cases (given the fact that it can be pretty big).

wladekb:
lgtm

drsnyder:
:+1:

PR #5664, opened 2014-12-08
CON-2257 media galleries float fix

macbre:
Pinging @lizlux (code ownership)

PR #6291, opened 2015-02-12
Con 2257 media galleries float fix

lizlux:
https://github.com/Wikia/app/pull/5664 was the old PR for this. 

https://wikia-inc.atlassian.net/browse/SOC-365 is the ticket. 

This PR makes it so the new media galleries can wrap around floated elements like infoboxes.


lizlux:
Code reviewed in #5664

PR #6292, opened 2015-02-12
PLATFORM-879: Add reasonable limit for the lookahead in geshi regexp

wladekb:
This fixes an issue in GeSHi that triggers a stack limit exceeded error and segfault when CSS contains an invalid and long input.

https://wikia-inc.atlassian.net/browse/PLATFORM-879

/cc @macbre @sqreek


macbre:
Is this regex written in Chinesse?

LGTM, because I trust you, Władek :)

wladekb:
No, it's the next-gen Perl flavor...

PR #6294, opened 2015-02-12
Soc-163 Adding form validation to facebook login and singup forms

lizlux:
Ticket: https://wikia-inc.atlassian.net/browse/SOC-163

This PR attempts to unify the way we do form validation in the UserLogin extension. It involves changing the `name` attribute of the `username` and `password` inputs for a number of forms, so that they can use the same validation mechanisms. 

While I was in this code I noticed some bugs with changing a password from the login modal and the facebook modal, each have been ticketed and there's some comments in the code.


extensions/wikia/UserLogin/UserLoginForm.class.php
```diff
@@ -225,8 +251,14 @@ public function addNewAccountInternal() {
 		return parent::addNewAccountInternal();
 	}
 
-	public function mainLoginForm( $msg, $msgtype = 'error', $errParam='' ) {
-		$this->msgType = $msgtype;
+	/**
+	 * Apply a validation message (defaults to error message) to this form instance
+	 * @param $msg
```
armonr: please add type

extensions/wikia/UserLogin/UserLoginSpecialController.class.php
```diff
@@ -364,8 +370,12 @@ public function providersTop() {
 	}
 
 	public function modal() {
-		$this->loginToken = UserLoginHelper::getLoginToken();
-		$this->signupUrl = Title::newFromText('UserSignup', NS_SPECIAL)->getFullUrl();
+		$this->response->setData([
```
armonr: white space

extensions/wikia/UserLogin/UserLoginSpecialController.class.php
```diff
@@ -602,18 +612,32 @@ public function changePassword() {
...
-		$this->loginToken = $this->request->getVal( 'loginToken', '' );
-		$this->returnto = $this->request->getVal( 'returnto', '' );
+		$username = $this->request->getVal( 'username', '' );
+		$password = $this->request->getVal( 'password', '' );
+		$newPassword = $this->request->getVal( 'newpassword', '' );
+		$retype = $this->request->getVal( 'retype', '' );
+		$loginToken = $this->request->getVal( 'loginToken', '' );
+		$returnto = $this->request->getVal( 'returnto', '' );
+
+		// TODO: figure out why we were setting editToken twice, once via request var and once
```
armonr: It seems the request `editToken` was always being overwritten.
:+1:  on adding a todo. This could potentially be a bug.

extensions/wikia/UserLogin/UserLoginSpecialController.class.php
```diff
@@ -630,14 +654,14 @@ public function changePassword() {
 			if( $this->wg->User->matchEditToken( $this->editToken ) ) {
 
 				if ( $this->wg->User->isAnon()
-					&& $this->loginToken !== UserLoginHelper::getLoginToken()
+					&& $loginToken !== UserLoginHelper::getLoginToken()
 				) {
 					$this->result = 'error';
```
armonr: Please also convert these to use the response setter :)

extensions/wikia/UserLogin/FacebookSignupController.class.php
```diff
@@ -162,6 +159,8 @@ public function modal() {
 			'returnToUrl' => $returnToUrl,
 			'queryString' => $returnToParams,
 			'loginToken' => UserLoginHelper::getSignupToken(),
+			'usernameKey' => UserLoginForm::SIGNUP_USERNAME_KEY,
+			'passwordKey' => UserLoginForm::SIGNUP_PASSWORD_KEY,
```
rafalkalinski: It is a good move to get rid of the consts from this module and unify them. 
And what do you think about not importing these from 'sibling' module which has different responsibility?. Ideally would be that both modules (FacebookSignupController and UserLoginForm) inherit these consts from one place in UserLogin extension.

extensions/wikia/UserLogin/UserLoginSpecialController.class.php
Grunny: This was/is a bug. The one from the request is what should be checked below with the `$this->wg->User->matchEditToken( $this->editToken )`, and then setting the value is used to make sure the form has the right token. Setting it twice as is being done means the check didn't do anything useful. So, it should match the behaviour of SpecialChangePassword.php from core ([1](https://github.com/Wikia/app/blob/58714681833/includes/specials/SpecialChangePassword.php#L69), [2](https://github.com/Wikia/app/blob/58714681833/includes/specials/SpecialChangePassword.php#L163).

extensions/wikia/UserLogin/UserLoginSpecialController.class.php
```diff
@@ -602,18 +612,32 @@ public function changePassword() {
...
+		$password = $this->request->getVal( 'password', '' );
+		$newPassword = $this->request->getVal( 'newpassword', '' );
+		$retype = $this->request->getVal( 'retype', '' );
+		$loginToken = $this->request->getVal( 'loginToken', '' );
+		$returnto = $this->request->getVal( 'returnto', '' );
+
+		// TODO: figure out why we were setting editToken twice, once via request var and once
+		// via wgUser->getEditToken. (SOC-351)
+		$editToken = $this->wg->User->getEditToken(); // $this->request->getVal( 'editToken', '' );
+		$this->editToken = $editToken;
```
rafalkalinski: Do we need $editToken variable if we still want to have this value in $this->editToken?

extensions/wikia/UserLogin/js/FacebookLogin.js
```diff
@@ -301,7 +305,10 @@
 
 			this.loginForm = new window.FacebookFormConnectUser($modal.find('.UserLoginFacebookRight'), {
 				ajaxLogin: true,
+				ajaxValidation: true,
 				skipFocus: true,
+				usernameInputName: 'userloginext01',
+				passwordInputName: 'userloginext02',
```
rafalkalinski: Is there a chance to put these strings ('userloginext01' and 'userloginext02') into some common variable? Just to have it defined in one place.

extensions/wikia/UserLogin/js/UserBaseAjaxForm.js
```diff
@@ -62,19 +75,34 @@
...
-			'login',
-			{
-				loginToken: this.loginToken,
-				username: this.inputs.username.val(),
-				password: this.inputs.password.val(),
-				keeploggedin: this.inputs.keeploggedin.is(':checked')
-			},
-			this.submitLoginHandler.bind(this)
-		);
+	UserBaseAjaxForm.prototype.ajaxLogin = function () {};
```
rafalkalinski: In UserLogin.js there are two another empty functions like this one.
Maybe we could define it in one place for this whole extension and use only reference to it. I mean something similar like jQuery.noop().
Three uses like in this extension is not much, but it is always a good practice to reuse a functions which does something (or nothing like in here). Ideally it would be to have one place for our whole code for utilities like this.

extensions/wikia/UserLogin/js/UserLoginAjaxForm.js
```diff
@@ -87,8 +96,15 @@
 	 * @param {string} html
 	 */
 	UserLoginAjaxForm.prototype.retrieveTemplateCallback = function (html) {
-		var content = $('<div>').hide().append(html),
-			form = this.form;
+		var content, form;
+
+		if (typeof this.options.retrieveTemplateCallback === 'function') {
```
rafalkalinski: My suggestion:
I wouldn't do this check. It may be not politically correct (or it might be that I don't have enough context :) ), but if someone are using this extension, and trying pass to _options.retrieveTemplateCallback_ something that is not a function, then he is asking for troubles, or more probably he is doing some mistake. It would be helpful for that person if this code will blow up - impossible to miss, and easy to understand why.

extensions/wikia/UserLogin/js/UserLoginModal.js
```diff
@@ -75,53 +75,35 @@
...
-										contentBlock.slideDown(400);
-									});
-								}
-							});
-						}
-					});
+						};
+
+					// UserLogin.js sends a callback function for reloading the classic editor after forced login
+					if (typeof options.callback === 'function') {
```
rafalkalinski: the same comment as with _this.options.retrieveTemplateCallback_ above

extensions/wikia/UserLogin/js/UserLoginAjaxForm.js
lizlux: Would you suggest just checking to see if `typeof this.options.retrieveTemplateCallback !== undefined`? Either way we have to check if it's been passed, since it's an optional setting, and in most scenarios it will be undefined. 

I agree that it's not super great code, but i've actually been refactoring and deleting these optional callback functions as much as possible. This one is still around tho for the change password flow inside a modal. Actually, it might be better to trigger an event instead...

extensions/wikia/UserLogin/js/UserLoginModal.js
lizlux: I'll see what i can do here.

extensions/wikia/UserLogin/templates/UserLoginSpecial_changePassword.php
```diff
@@ -69,8 +63,8 @@
 		)
 	);
 
-	$form['isInvalid'] = !empty($result) && empty($errParam) && !empty($msg);
-	$form['errorMsg'] = !empty($msg) ? $msg : '';
+	$form['isInvalid'] = !empty( $result ) && $result == 'error';
```
rafalkalinski: is there a need for `!empty( $result )`?

extensions/wikia/UserLogin/js/UserLoginAjaxForm.js
rafalkalinski: Yes, I was thinking about something like this - just to check if there is anything - cos it's optional parameter.
On the other hand you could create a default empty function, and always call it - but all this function should be changed then because of that return below.

And don't worry - it's a good code. I just have different view on it and can't refrain from sharing it :)

extensions/wikia/UserLogin/FacebookSignupController.class.php
lizlux: Totally agreed. Garth and I actually discussed creating a ticket to clean up the back end of this extension, however since we're going to redo a lot of the user login and signup flow soon, we're going to forgo code optimizations like this for now.

lizlux:
Weird, not sure how this happened, but my css changes from https://github.com/Wikia/app/pull/6351 got mixed up in this PR.

extensions/wikia/UserLogin/UserLoginSpecialController.class.php
lizlux: Setting it to match the core behavior didn't fix the issue in https://wikia-inc.atlassian.net/browse/SOC-351, but it didn't make it worse, so I'm going to wait till that ticket comes up in the backlog to troubleshoot this further.

lizlux:
Closing this and opening a PR into dev branch.

extensions/wikia/UserLogin/js/UserBaseAjaxForm.js
lizlux: The idea behind this is that it's an abstract class and method, even though JavaScript doesn't really have that concept, so it looks like it's just an empty function. Since all child classes override this method, there's no point in putting anything in here. However, it is a required method, so I don't want to delete it. 

I didn't see any empty functions in UserLogin.js. Are you sure that's the file you meant?

extensions/wikia/UserLogin/UserLoginSpecialController.class.php
lizlux: I think I'll leave it as is for now and whoever picks up the change password bug ticket can sort it out.

PR #6296, opened 2015-02-13
SOC-273 making sure facebook button shows up with force login modal

lizlux:
Fixing race condition where the facebook button inside the forced login modal wasn't being selected b/c it wasn't on the page at the time of the DOM query. 

I tried to find a good lifecycle event for the modal so I could know when the content had actually reached the DOM, but I couldn't find one. Maybe @kvas-damian knows more about it?

I also found a bug with the facebook sdk loading script where it didn't fire a callback on subsequent calls, so this PR fixes that.

### DIFF
--- a/extensions/wikia/UserLogin/js/UserLoginModal.js
+++ b/extensions/wikia/UserLogin/js/UserLoginModal.js
@@ -75,6 +75,11 @@
 
 					// Init facebook button inside login modal
 					if (window.FacebookLogin) {
+						// SOC-273 remove 'hidden' class even if element isn't in the DOM yet
+						$.loadFacebookAPI()
+							.done(function () {
+								$loginModal.find('.sso-login').removeClass('hidden');
+							});
 						window.FacebookLogin.init(window.FacebookLogin.origins.MODAL, $loginModal);
 					}
 
--- a/extensions/wikia/UserLogin/js/UserLoginModal.js
+++ b/extensions/wikia/UserLogin/js/UserLoginModal.js
@@ -80,7 +80,7 @@
 							.done(function () {
 								$loginModal.find('.sso-login').removeClass('hidden');
 							});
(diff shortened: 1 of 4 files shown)


extensions/wikia/UserLogin/js/UserLoginModal.js
```diff
@@ -77,6 +73,16 @@
...
 					var $loginModal = loginModal.$element;
 
+					// Init facebook button inside login modal
+					if (window.FacebookLogin) {
+						// SOC-273 remove 'hidden' class even if element isn't in the DOM yet
+						$.loadFacebookAPI()
+							.done(function () {
+								$loginModal.find('.sso-login').removeClass('hidden');
+							});
+						window.FacebookLogin.init(window.FacebookLogin.origins.MODAL, $loginModal);
```
kvas-damian: ```$loginModal``` is never used - probably it's leftover from initial fix.

kvas-damian:
@lizlux I'm not sure if real problem is with when we add content to the DOM.
Main problem here is fact that this modal is recreated (data and assets retrieved from backend) every time modal is shown. Connecting this fact with ```FacebookLogin.init()``` which executes code only once per pageview https://github.com/Wikia/app/blob/dev/extensions/wikia/UserLogin/js/FacebookLogin.js#L37

This way when we initialize ```FacebookLogin``` somewhere else (see dropdown, or open modal, then close it and open again) this code isn't fired again. I think your solution is valid in this case, because we should avoid initializing ```FacebookLogin``` again.

bkoval:
LGTM if it works :)

extensions/wikia/UserLogin/js/UserLoginModal.js
lizlux: Ah yes, thanks for catching.

kvas-damian:
LGTM

PR #6297, opened 2015-02-13

jsutterfield:


extensions/wikia/FacebookClient/FacebookClientXFBML.php
```diff
@@ -48,12 +48,13 @@ public static function parserHook( $tag, $text, $args, Parser $parser ) {
 			unset( $args['profileid'] );
 		}
 
+		// add custom attribute to look for in DOM. We use this to determine
+		// if the Facebook SDK should be loaded.
+		$args['data-type'] = 'xfbml-tag';
```
jsutterfield: I couldn't find a good way to search for all the custom fb tags (eg `<fb:like>, <fb:share>, <fb:follow>'), so I'm adding a custom attribute to check for in the DOM.

extensions/wikia/FacebookClient/scripts/FacebookClient.XFBML.js
```diff
@@ -2,19 +2,34 @@
...
 
-				// scan the new content for any fb tags
-				FB.XFBML.parse($content[0]);
-			});
+		if ((xfbmlTagsOnPage())) {
+			if (facebookSDKNotLoaded()) {
+				// XFBML tags rendered automatically when SDK loads
+				$.loadFacebookSDK();
+			} else {
+				renderXFBMLTags();
```
jsutterfield: The original implementation relied on done() to fire `FB.XFBML.parse($content[0]);`. There were a couple of problems with this:

1.) The config object we pass to FB.init [here](https://github.com/Wikia/app/blob/dev/resources/wikia/libraries/jquery/loadLibrary/jquery.wikia.loadExternalLibrary.js#L63) is already instructing FB to parse the XFBML tags. Calling it manually as well was causing some race condition where the first XFBML tag would never render.
2.) The way that that `$.loadFacebookSDK()` works if the SDK is already loaded then the resolve on the promise is never fired. I tried to a couple different methods for fixing that, but none worked particularly well. I figured it's a pretty edge case need, so I could just keep the logic out here.

lizlux: @jsutterfield, sorry about the trouble you ran into with `$.loadFacebookSDK`. I actually noticed that bug and fixed it in [this PR](https://github.com/Wikia/app/pull/6296/files#diff-d8d8cc9a4a23644a051db21dc3b2ce3aL49). Once that's merged you shouldn't need to do these work-arounds, you should just be able to call `$.loadFacebookSDK().done(...)` and it should work fine after that. Again, sorry for the bug and for not communicating that. It looks like it would have saved you some time.

resources/wikia/libraries/jquery/loadLibrary/jquery.wikia.loadExternalLibrary.js
```diff
@@ -64,7 +64,10 @@
 				cookie: true,
 				version: 'v2.1'
 			});
-			// resolve after FB has finished inititalizing
+
+			// show facebook login button
+			$('.sso-login').removeClass('hidden');
```
lizlux: This looks like a nice alternative to the previous method. We'll be changing it again with https://wikia-inc.atlassian.net/browse/SOC-240 but this is an improvement for now.

extensions/wikia/FacebookClient/scripts/FacebookClient.XFBML.js
lizlux: Also, hopefully my fix to the library loading function will fix point number one in your comment, but if not, I'd be curious to figure out why.

extensions/wikia/FacebookClient/scripts/FacebookClient.XFBML.js
```diff
@@ -2,19 +2,34 @@
...
+			}
+		}
+
+		function xfbmlTagsOnPage() {
+			var numOfXFBMLTags = $content.find('[data-type="xfbml-tag"]').length;
```
lizlux: I've always written this kind of thing like `return $content.find('[data-type="xfbml-tag"]').length`, although that will return a number instead of a boolean, so that might not be as clear. But type casting like `return !!$content.find('[data-type="xfbml-tag"]').length` is also acceptable, as is your currently solution, which is arguably the most clear, but also seems unnecessarily verbose to someone that's used to working with jQuery.

extensions/wikia/FacebookClient/scripts/FacebookClient.XFBML.js
```diff
@@ -2,19 +2,34 @@
 
 /**
  * Render FB XML tags emebedded into wikitext
- * This file is loaded via ResourceLoader only when FB XML tags are present on the page
  */
 $(function () {
```
lizlux: I wonder if this is superfluous now b/c `mw.hook` is called when the DOM is loaded.

extensions/wikia/FacebookClient/scripts/FacebookClient.XFBML.js
jsutterfield: No worries at all, I should have reached out to you when I noticed it rather than trying to work around it. I merged your fix it and it works great. And it does fix point number 1, so I'm taking that `fbxml: true` out of the config so we can just call `FB.XFBML.parse($content[0]);` manually whenever the wikipage.content event fires.

extensions/wikia/FacebookClient/scripts/FacebookClient.XFBML.js
jsutterfield: Yea, I've seen that `!!` syntax in javascript before and never really liked it. It saves a bit of typing, and if you're well acquainted with the language it makes sense, but I think it makes things slightly less clear for the sake of brevity. Idk, could just be that I haven't written enough js to get that operator drilled into my head. If it works for you, I think I'll stick with current solution. I like that it's pretty unambiguous and straightforward to read.

PR #6298, opened 2015-02-13
CONCF-41 remove never assigned variable

bognix:
I removed $useRDFData variable which was never assigned and $timer variable which was never used.
$useRDFData variable is causing PHP Erros.
@lizlux you removed the assignment in https://github.com/Wikia/app/commit/64643127112194a59f2e51c3e7a6609edec31543 - is it OK if I remove the rest?

### DIFF
--- a/extensions/wikia/Thumbnails/ThumbnailVideo.class.php
+++ b/extensions/wikia/Thumbnails/ThumbnailVideo.class.php
@@ -160,12 +160,6 @@ class ThumbnailVideo extends ThumbnailImage {
 			$linkAttribs['id'] = $options['id'];
 		}
 
-		if ( $useRDFData ) { // bugId: #46621
-			$linkAttribs['itemprop'] = 'video';
-			$linkAttribs['itemscope'] = '';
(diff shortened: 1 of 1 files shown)


vforge:
LGTM

macbre:
:+1:

lizlux:
Thanks for looking into this @bognix. My only concern is that we added the RDF data to thumbnails in order to increase SEO. I don't remember the full context of why I removed it, but it's possible that I did it by accident, and that the actual fix is to put it back in place. @shareif would have more info on this than me. Please check in with him to see if the we should reinstate RDF data in the thumbnail HTML. Thanks!

shareif:
Yeah, it looks like something fairly important for Video SEO.

bognix:
Ok, I'll bring back the variable assignment.

bognix:
This change is only affecting wikiamobile skin.

vforge:
LGTM

hakubo:
LGTM

lizlux:
Ah okay, I looked at the code more and it does in fact only affect WikiaMobile. It would be nice to adjust the code to reflect that better, but this is probably fine for now.

PR #6299, opened 2015-02-13
(CE-1271) Add property translations for Polish to SemanticDrilldown

Grunny:
Fix PHP Warning in SemanticDrilldown by adding property translations
for Polish to SemanticDrilldown. If an SD_Language*.php translation
file is to be used, it expects both the namespace and properties to
be included, but for Polish only the namespaces were added.

/cc @lgarczewski


adamkarminski:
LGTM :+1:

PR #6300, opened 2015-02-13
Exclude gallery params from caption string

hakubo:
https://wikia-inc.atlassian.net/browse/CONCF-8

This has to be tested across skins.

It was always broken, it's just that Mercury displays this data whereas other skins did not.

### DIFF
--- a/extensions/wikia/WikiaPhotoGallery/WikiaPhotoGallery.class.php
+++ b/extensions/wikia/WikiaPhotoGallery/WikiaPhotoGallery.class.php
@@ -118,6 +118,21 @@ class WikiaPhotoGallery extends ImageGallery {
 			'hideoverflow' => false,
 			'sliderbar' => array('bottom','left')
 		);
+
+		$this->mAvalaibleUniqueParams = array_values(
+			array_unique(
+				array_reduce(
+					array_filter(
+						array_values( $this->mAvalaibleGalleryParams ),
+						function ($var) {
+							return !empty($var);
+						}
+					),
+					array_merge,
+					[]
+				)
+			)
+		);
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


extensions/wikia/WikiaPhotoGallery/WikiaPhotoGallery.class.php
```diff
@@ -118,6 +118,21 @@ function __construct() {
 			'hideoverflow' => false,
 			'sliderbar' => array('bottom','left')
 		);
+
+		$this->mAvalaibleUniqueParams = array_values(
```
rogatty: It's misspelled, should be `mAvailableUniqueParams`. And yes, `mAvalaibleGalleryParams` is misspelled too.

rogatty: Also, should it be declared like https://github.com/Wikia/app/pull/6300/files#diff-6ddf00ef82821d11d4d1cffd5bd8a74cR56?

extensions/wikia/WikiaPhotoGallery/WikiaPhotoGallery.class.php
```diff
@@ -402,7 +418,12 @@ public function parse(&$parser = null) {
 				} else if (substr($part, 0, 10) == 'shorttext=') {
 					$shorttext = substr($part, 10);
 				} else {
-					$captionParts[] = trim($part);
+					$tempPart = trim($part);
+
+					//If it looks like Gallery param don't treat it as a caption part
+					if (!in_array($tempPart, $this->mAvalaibleUniqueParams)) {
```
rogatty: Does it mean that we won't display captions like _small_, _landscape_ etc.? Is it working the same way in MediaWiki?

hakubo:
@rogatty I am not entirely sure, that why it needs some good QA expertise...

rogatty:
LGTM

PR #6302, opened 2015-02-13
Dat 2507

RafLeszczynski:
Limit tile set collection in create map modal to 49 element

### DIFF
--- a/extensions/wikia/WikiaMaps/controllers/WikiaMapsMapController.class.php
+++ b/extensions/wikia/WikiaMaps/controllers/WikiaMapsMapController.class.php
@@ -13,7 +13,7 @@ class WikiaMapsMapController extends WikiaMapsBaseController {
 	public function getTileSets() {
 		$params = [
 			'sort' => 'desc',
-			'limit' => 50
+			'limit' => 49
 		];
 		$searchTerm = $this->request->getVal( 'searchTerm', null );
 
--- a/extensions/wikia/WikiaMaps/controllers/WikiaMapsMapController.class.php
+++ b/extensions/wikia/WikiaMaps/controllers/WikiaMapsMapController.class.php
@@ -4,6 +4,9 @@
  * AJAX entry points for actions connected to map creation
  */
 class WikiaMapsMapController extends WikiaMapsBaseController {
+	// number of tilesets to be fetched from maps service. We want to have 50 items in the list in UI and the first
+	// item is create new tileset button. That's why the number is 49
+	const TILE_SET_ITEM_COUNT = 49;
 
 	/**
 	 * Gets list of tile sets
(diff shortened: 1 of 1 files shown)


extensions/wikia/WikiaMaps/controllers/WikiaMapsMapController.class.php
```diff
@@ -11,7 +11,10 @@ class WikiaMapsMapController extends WikiaMapsBaseController {
 	 * @return Array
 	 */
 	public function getTileSets() {
-		$params = [];
+		$params = [
```
nandy-andy: Do we want to limit `tilesets` results always? 1) If the user creates a custom map and a list of tilesets is presented to her and 2) if the user filters presented tilesets list with a keyword?

RafLeszczynski: AFAIK Zach wanted to always display 50 latest tile set also when using search. I set it up to 49 cause the first item i the list is upload button and it makes 50 items together with it

RafLeszczynski: @nandy-andy do you think I should add a test is we stay with 49 always?

nandy-andy: It's hard to say no to tests :japanese_ogre: It's up to you to decide. You could add a test checking if there are `sort`, `limit` parameters in `$params` always and if there is the additional one `search` element when passed with the request.

But again: it's up to you. I might be guessing it's additional value for you to spend some time with PHPUnit and a test is always an additional value for our code quality.

RafLeszczynski: True :-) will do it

nandy-andy: Sweet! :sweet_potato: Good luck and have fun :smile_cat:

extensions/wikia/WikiaMaps/controllers/WikiaMapsMapController.class.php
```diff
@@ -11,7 +11,10 @@ class WikiaMapsMapController extends WikiaMapsBaseController {
...
+			'sort' => 'desc',
+			'limit' => 49
```
nandy-andy: Oh, one more thing. We might want to add a comment line here why it is `49` - the thing about the upload button :) Extracting it to a const may be an overkill if it's being used only in this one place... on the other hand it's a good practice to keep "magic numbers" in const. I'd ask one or few more experienced developers what's their opinion about it :neckbeard:

extensions/wikia/WikiaMaps/controllers/WikiaMapsMapController.class.php
jacekjursza: IMO the reason to write unit test here would be to 'freeze' values of parameters 'sort' and 'limit'. But after talking to Rafał I believe there is no reason to do that - as the service back-end will handle it anyway and there is no risk of braking something easily. ;) 

On the others hand - I would move '49' number to const of this class and describe it - why it's 49.

jacekjursza:
LGTM

nandy-andy:
LGTM :smile:

PR #6305, opened 2015-02-13
ArticleService: /e modifier is not really needed here

macbre:
Fixes `Deprecated: preg_replace(): The /e modifier is deprecated, use preg_replace_callback instead in /usr/wikia/source/app/includes/wikia/services/ArticleService.class.php on line 212` when run using PHP 5.6
- extracted `cleanArticleSnippet` method to make unit testing a bit easier

@owend / @SebastianMarzjan


michalroszka:
Quite a clean up. Looks good!

PR #6307, opened 2015-02-13
Transaction: set "hostname" parameter and sent it as a context

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-905

As a part of PHP upgrade we would like to set of the Apaches in the poll to use PHP 5.6 to do the final round of testing and compare the performance.

In order to do that I'll added the `hostname` parameter that will be sent with all performance messages.

@wladekb


wladekb:
lgtm

PR #6308, opened 2015-02-13
MAIN-3868: Fix problem by switchich from old reCaptcha to new reCaptcha

inez:
About month ago there was an effort to switch from old reCaptcha to new reCaptcha and perhaps that change wasn't tested with VisualEditor.

https://wikia-inc.atlassian.net/browse/MAIN-3868


nandy-andy:
LGTM :+1:

PR #6310, opened 2015-02-16
DAT-2398 Add article type to GA custom variables

sebastianmarzjan:


jacekjursza:
LGTM. I was wondering if it's worth to refactor ArticleService class in order to make 1 call to Solr instead of 2 (snippet + article type), but then I read the comment about memoization. Cool :+1:

PR #6314, opened 2015-02-16
Helios endpoint improvements

ArturKlajnerok:
Please take a look: @michalroszka

### DIFF
--- a/extensions/wikia/Helios/HelperController.class.php
+++ b/extensions/wikia/Helios/HelperController.class.php
@@ -15,8 +15,10 @@ class HelperController extends \WikiaController
 	{
 		$this->response->setFormat( 'json' );
 		$this->response->setCacheValidity( \WikiaResponse::CACHE_DISABLED );
+		$this->response->setVal( 'allow', false );
 
 		if ( $this->getVal( 'secret' ) != $this->wg->TheSchwartzSecretToken ) {
+			$this->response->setVal( 'message', 'invalid secret' );
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


extensions/wikia/Helios/HelperController.class.php
```diff
@@ -15,8 +15,10 @@ public function checkAntiSpoof()
 	{
 		$this->response->setFormat( 'json' );
 		$this->response->setCacheValidity( \WikiaResponse::CACHE_DISABLED );
+		$this->response->setVal( 'allow', false );
```
michalroszka: @ArturKlajnerok — this one is not needed here as it is set in `return` statements.

extensions/wikia/Helios/HelperController.class.php
```diff
@@ -52,7 +56,7 @@ public function updateAntiSpoof()
 
 		if ( !empty( $this->wg->EnableAntiSpoofExt ) ) {
 			$oSpoofUser = new \SpoofUser( $sName );
-			$oSpoofUser->record();
+			return $this->response->setVal( 'success', $oSpoofUser->record() );
```
michalroszka: `SpoofUser::record()` returns true even on failure, so the check for success might be inconclusive and misleading.

extensions/wikia/Helios/HelperController.class.php
```diff
@@ -95,7 +99,8 @@ public function sendConfirmationEmail()
 			return;
 		}
 
-		$sMemKey = $this->getMemKeyConfirmationEmailsSent( $oUser->getId() );
+		$userLoginHelper = ( new \UserLoginHelper );
```
michalroszka: I would use the Hungarian notation for consistency, so `$oUserLoginHelper`.

extensions/wikia/Helios/HelperController.class.php
ArturKlajnerok: My plan is to return the result here to know what was the operation outcome and to fix SpoofUser::record() implementation.

PR #6317, opened 2015-02-16
CE-1313 Fix fatal error and add documentation

adamkarminski:
The pull request adds an additional check that prevents anything but a **_ResultWrapper_** object to be passed to isNativeWikiaValid.

Ping @macbre


macbre:
You're fast :) LGTM

PR #6323, opened 2015-02-17
Remove unused Wikia::getEnvironmentName

macbre:
```
app$ ack-grep getEnvironmentName
includes/wikia/Wikia.php
2320:   static public function getEnvironmentName() {
```

@michalroszka / @wladekb


michalroszka:
The method seems unused, indeed.

PR #6334, opened 2015-02-17
GlobalNotification -> BannerNotification(s)

bkoval:
This is an attempt to refactor GlobalNotification. Summary:
- Renaming it to BannerNotification
- Moving all the related files to the single extension directory
- Creating an AMD module for BN, but keeping the window global
- Adding support for multiple instances of notifications
- Adding generic method for ajax connection errors.
- Notifications are now objects that can be instantiated with 'new' keyword
- Setters and methods can be chained

PHP Usage

``` php
BannerNotificationController::addNotification(...);
```

instead of 

``` php
NotificationController::addNotification(...);
```

JavaScript Usage

``` javascript
//where possible
require(['BannerNotification'], function(BannerNotification){

    new BannerNotification('MyMessage", 'error').show();

    new BannerNotification().setContent('MyMessage').setType('error').show();

    var notification = new BannerNotification('MyMessage').setType('notify');
    notification.show();
    notification.hide();
});

//Where not possible to use AMD, there's still support for:
new window.BannerNotification().setContent('Hello').setType('confirm').show();
```

Also applied AMD version where it was possible and made WIkiaInYourLang and CookiePolicy use $element exposed by notification instance for binding events.

@kvas-damian @lizlux Please, review and after i address the Code Review notes I will ask owners of all the extensions using notifications to review the code.

I also added a wikia.onScroll module as a global place for binding to scroll event on window:

``` javascript
require(['wikia.onScroll'], function(onScroll){
    function doSomething(){}

    onScroll.bind(doSomething);
    //doSomething is getting triggered on scroll event now

    onScroll.unbind(doSomething);
    //doSomething has been unboud, it's not getting triggered anymore
});
```


extensions/wikia/BannerNotifications/BannerNotifications.setup.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+/**
+ * Docs are on internal in UI Style Guide
+ */
+$wgExtensionCredits['bannernotifications'][] = array(
+	'name' => 'BannerNotifications',
+	'descriptionmsg' => 'bannernotifications-desc',
+	'author' => 'Hyun Lim',
```
lizlux: Let's update this value since this extension has changed a bunch since Hyun last touched it.

extensions/wikia/BannerNotifications/css/BannerNotifications.monobook.scss
```diff
@@ -0,0 +1,6 @@
+@import 'skins/shared/styles/buttons/buttons-wikia-chiclet';
```
lizlux: Is this import necessary?

extensions/wikia/BannerNotifications/js/BannerNotifications.js
```diff
@@ -0,0 +1,269 @@
...
+	 */
+	function hideAll(callback) {
+		removeFromDOM($('.banner-notification'), callback);
+	}
+
+	/**
+	 * Bind close event to close button
+	 */
+	function setUpClose() {
+		$(document.body).on('click', '.close', onCloseClicked);
```
lizlux: The selector `.close` is probably not specific enough, since there's going to be `.close` elements in places other than banner notifications, and there may be unexpected consequences if `onCloseClicked` is fired on elements like modals.

extensions/wikia/BannerNotifications/BannerNotifications.setup.php
bkoval: I'll check the git history for this file and will update it :)

extensions/wikia/BannerNotifications/css/BannerNotifications.monobook.scss
bkoval: It's here so the 'close' button looks like it does on Oasis. Deleting it would make the close button look unstyled. I could add some minimal styles so it's visible enough, I just added these styles to make BN on both skins look the similar. WHat do you think, @lizlux ?

extensions/wikia/BannerNotifications/js/BannerNotifications.js
bkoval: Good catch! I had in mind to change the filtering selector to '.banner-notification .close' but must've forgot to do so, thanks!

extensions/wikia/BannerNotifications/css/BannerNotifications.monobook.scss
lizlux: It's really not a big deal, but I wish there were a more semantic place to put that include. 

Effectively this will include the chiclet styles in all of monobook, since the BannerNotifications extension is included on every page view. This is, of course, what we want, but what happens when we remove this extension and there were other places in monobook that were relying on that chiclet styling to be there? I wish we had a `monobook.scss` like we have `oasis.scss`, that included all the core css modules needed in monobook. Maybe monobook is just too far gone to be worrying about semantic code optimizations like this though. 

Just food for thought, not a blocker as far as I'm concerned.

extensions/wikia/BannerNotifications/js/BannerNotifications.js
```diff
@@ -0,0 +1,269 @@
...
+		modal,
+		isModal;
+
+	/**
+	 * Constructs jQuery element with the notification
+	 * @param {String} content
+	 * @param {String} [type]
+	 * @returns {jQuery}
+	 */
+	function createMarkup(content, type) {
```
lizlux: It would be nice to create a mustache template and import it via `grunt-mustache`. https://one.wikia-inc.com/wiki/Engineering/Grunt-Mustache

extensions/wikia/BannerNotifications/js/BannerNotifications.js
```diff
@@ -0,0 +1,269 @@
...
+		bannerNotification.setType(type);
+
+		addToDOM(
+			bannerNotification.$element,
+			$parent
+		);
+
+		// Share scroll event with WikiaFooterApp's toolbar floating (BugId:33365)
+		if (window.WikiaFooterApp) {
+			window.WikiaFooterApp.addScrollEvent();
```
lizlux: Did you look into removing this dependency? I can see why they did it this way for performance reasons, but it might be nice to abstract the scroll event handling into a separate module that both `WikiaFooterApp` and `BannerNotifications` could plug into. The current dependency seems brittle and unexpected from the point of view of someone looking at this for the first time.

skins/oasis/js/WikiaFooter.js
```diff
@@ -12,7 +12,7 @@ var WikiaFooterApp = {
 			this.footer = $( '#WikiaFooter' );
 		}
 		this.toolbar = this.footer.children( '.toolbar' );
-		this.gn = $( '.global-notification' );
+		this.gn = $( '.banner-notification' );
```
lizlux: `gn` => `bn`

skins/oasis/js/WikiHeader.js
```diff
@@ -393,7 +393,7 @@
 						$( '#publish' ).remove();
 						// TODO: use mustache and promise pattern along with .getMessages
 						notifications =
-							'<div class="global-notification error">' +
+							'<div class="banner-notification error">' +
```
lizlux: Looks like this should be re-written to just use `BannerNotification.show(...)`

lizlux:
@bkoval, I checked out the code and tried out the new JS interfaces you mentioned above, there there's a few items that didn't quite work as expected: 

* When there's two notifications present, only one of them scrolls with the page. 
* Using the following syntax doesn't work, since `setType` doesn't actually return the instance 
`var notification = BannerNotifications.show('MyMessage').setType('notify');
 notification.setContent('OtherMessage').hide();`

My suggestion would be to create a JS class for BannerNotifications, so that when we need a new instance, we just call `new BannerNotification('message', 'type')` or `new BannerNotification().setMessage('message').setType('type')`. In order for that chaining to work, you have to return `this`, which is typically done by creating new instances, as mentioned above. The end result should be that any number of notifications can be added to the page, each with their own scroll handling and close events, as well as internal references to instance methods like `setType`, etc. 

I'm happy to discuss this more OO approach further, although I'm sure there's good resources in the PL office for that discussion as well.

extensions/wikia/BannerNotifications/js/BannerNotifications.js
bkoval: Done :dancer:

extensions/wikia/AssetsManager/config.php
```diff
@@ -922,7 +925,9 @@
 		'#group_imglzy_js',
 		'#group_spotlights_js',
 		'//resources/wikia/libraries/ghostwriter/gw.min.js',
-		'//extensions/wikia/GlobalNotification/GlobalNotification.js',
+		'//extensions/wikia/BannerNotifications/js/BannerNotifications.js',
+		'//extensions/wikia/BannerNotifications/js/templates.mustache.js',
```
kvas-damian: Can we extract those assets into group and avoid repeatition?

extensions/wikia/AutomaticWikiAdoption/AutomaticWikiAdoptionHelper.class.php
```diff
@@ -337,14 +337,14 @@ static function onSkinTemplateOutputPageBeforeExec(&$skin, &$tpl) {
...
+			BannerNotificationsController::addNotification(
+				wfMsg( 'wikiadoption-notification',
 						$wgSitename,
-						Wikia::SpecialPageLink('WikiAdoption','wikiadoption-adopt-inquiry')),
+						Wikia::SpecialPageLink( 'WikiAdoption','wikiadoption-adopt-inquiry' ) ),
 				array(
 					'name' => 'AutomaticWikiAdoption',
 					'dismissUrl' => $wgScript . '?action=ajax&rs=AutomaticWikiAdoptionAjax&method=dismiss',
-			), NotificationsController::NOTIFICATION_CUSTOM);
+			), BannerNotificationsController::NOTIFICATION_CUSTOM );
```
kvas-damian: Is it possible to expose the same API for JS and PHP?

extensions/wikia/BannerNotifications/BannerNotifications.setup.php
```diff
@@ -0,0 +1,47 @@
...
+ */
+$wgAutoloadClasses[ 'BannerNotificationsController' ] = $dir . 'BannerNotificationsController.class.php';
+
+/**
+ * Hooks
+ */
+// confirmations
+$wgHooks['ArticleDeleteComplete'][] = 'BannerNotificationsController::addPageDeletedConfirmation';
+$wgHooks['ArticleUndelete'][] = 'BannerNotificationsController::addPageUndeletedConfirmation';
+#$wgHooks['EditPageSuccessfulSave'][] = 'BannerNotificationsController::addSaveConfirmation'; // BugId:10129
```
kvas-damian: What do you think about removing this line? I don't expect it get used again when we introduced VE.

extensions/wikia/FacebookClient/scripts/preferences.js
```diff
@@ -54,7 +58,10 @@
 				callback: function (data) {
 					if (data.status === 'ok') {
 
-						window.GlobalNotification.show($.msg('fbconnect-preferences-connected'), 'confirm');
+						bannerNotification
+							.setContent($.msg('fbconnect-preferences-connected'))
```
kvas-damian: I know this code was already there, but we could introduce escaping messages: http://www.mediawiki.org/wiki/Manual:Messages_API#Use_of_the_messages

extensions/wikia/BannerNotifications/BannerNotificationsController.class.php
```diff
@@ -5,7 +5,7 @@
  * @author Maciej Brencz
  */
 
-class NotificationsController extends WikiaController {
+class BannerNotificationsController extends WikiaController {
```
kvas-damian: I'm not sure if this class should be renamed. It's also responsible for displaying Oasis notification: https://one.wikia-inc.com/wiki/Oasis/Notifications_and_confirmations

skins/oasis/js/WikiaFooter.js
```diff
@@ -30,7 +33,7 @@ var WikiaFooterApp = {
 	},
 	addScrollEvent: function() {
 		'use strict';
-		WikiaFooterApp.windowObj.off( 'scroll.FooterAp' ); // GlobalNotifications could be re-binding this event.
+		WikiaFooterApp.windowObj.off( 'scroll.FooterAp' );
 		WikiaFooterApp.windowObj.on( 'scroll.FooterAp', WikiaFooterApp.resolvePosition ).triggerHandler( 'scroll' );
```
lizlux: Let's create a follow-up ticket to have WikiaFooter use the nifty new onScroll module (if one hasn't been created already).

lizlux:
@bkoval, for some reason I keep getting `Uncaught Module BannerNotification is not defined.` when I run this code in a browser. The error is coming from the AMD code so I'm not quite sure why it's happening. To reproduce, you can try logging out and logging back in again.

extensions/wikia/BannerNotifications/js/BannerNotifications.js
```diff
@@ -0,0 +1,251 @@
...
+		addToDOM(this.$element, this.$parent);
+
+		this.hidden = false;
+
+		// Close notification after specified amount of time
+		if (typeof this.timeout === 'number') {
+			self = this;
+			setTimeout(function () {
+				self.hide();
+			}, this.timeout);
```
lizlux: You could write this as: 

```
setTimeout(function () {
	this.hide();
}.bind(this), this.timeout);
```
to remove the need for the local var `self`.

extensions/wikia/BannerNotifications/js/BannerNotifications.js
```diff
@@ -0,0 +1,251 @@
...
+			if (this.$element) {
+				this.$element
+					.find('.msg')
+					.html(content);
+			}
+		}
+		return this;
+	};
+
+	BannerNotification.prototype.showConnectionError = function () {
```
lizlux: Let's add more flushed out doc blocks for the file. Also, it would be nice to order the functions a bit better too. For example, you could put the private functions below all the prototype methods, and place each method in the order in which it might be called.

extensions/wikia/BannerNotifications/js/BannerNotifications.js
```diff
@@ -0,0 +1,251 @@
...
+		if (typeof this.timeout === 'number') {
+			self = this;
+			setTimeout(function () {
+				self.hide();
+			}, this.timeout);
+		}
+		return this;
+	};
+
+	BannerNotification.prototype.hide = function (callback) {
```
lizlux: What do you think about getting rid of this `callback` param? I think it was only added to be used internally, and it looks like it's no longer being used anywhere. It would be nice to simplify the code, and it can always be added back in later if we need it in the future.

extensions/wikia/WikiaStyleGuide/templates/WikiaStyleGuideSpecial_index.php
```diff
@@ -250,29 +250,29 @@
...
+<!-- Banner Notification Sample Code -->
 <script>
 $(function() {
-	$('#UISGGlobalNotification').toggle(function() {
-		GlobalNotification.show('Hello there, I am Global Notification', 'notify');
+	$('#UISGBannerNotification').toggle(function() {
+		BannerNotifications.show('Hello there, I am Banner Notification', 'notify');
 	}, function() {
-		GlobalNotification.hide();
+		BannerNotifications.hideAll();
```
lizlux: Does this method exist?

extensions/wikia/WikiaInYourLang/modules/ext.wikiaInYourLang.js
```diff
@@ -101,24 +102,23 @@ require(
...
 			// Bind tracking of clicks on a link and events on close
-			bindEvents();
+			bindEvents(notification.$element);
 		}
 
-		function bindEvents() {
-			$('.global-notification.notify')
+		function bindEvents($element) {
+			$element
 				.on('click', '.close', function() {
```
lizlux: This might be better handled as an `onClose` event that's emitted from the BannerNotification instance. That way we would be free to change stuff like the `.close` element without breaking stuff like this. 

Alternately, we could expose the `.close` element as `notification.$close` and bind to that.

skins/oasis/css/core/responsive-core.scss
```diff
@@ -230,7 +230,7 @@
 			width: auto;
 		}
 
-		.global-notification {
+		.banner-notification {
 			width: 100%;
```
lizlux: This is causing some funky behavior on small screens. 
screen shot 2015-02-19 at 6 10 02 pm

skins/oasis/js/WikiHeader.js
```diff
@@ -392,12 +391,12 @@
 					if ( errorMessages.length > 0 ) {
 						$( '#publish' ).remove();
 						// TODO: use mustache and promise pattern along with .getMessages
```
lizlux: This comment should be updated.

skins/oasis/js/WikiaFooter.js
```diff
@@ -12,12 +15,12 @@ var WikiaFooterApp = {
 			this.footer = $( '#WikiaFooter' );
 		}
 		this.toolbar = this.footer.children( '.toolbar' );
-		this.gn = $( '.global-notification' );
+		this.bn = $( '.banner-notification' );
```
lizlux: Actually it looks like we can remove references to banner notifications from this file all together, since we don't need the scroll event or the resize event for BN.

extensions/wikia/WikiaStyleGuide/templates/WikiaStyleGuideSpecial_index.php
```diff
@@ -250,29 +250,29 @@
...
+&lt;div id=&quot;UISGBannerNotification&quot; class=&quot;uisg-sample-pane&quot;&gt;<br/>	Click me to toggle Notification near top of the window.<br/>&lt;/div&gt;
 
 javascript:
-$('#UISGGlobalNotification').toggle(function() {
-	GlobalNotification.show('Hello there, I am Global Notification', 'notify');
+$('#UISGBannerNotification').toggle(function() {
+	BannerNotifications.show('Hello there, I am Banner Notification', 'notify');
 }, function() {
-	GlobalNotification.hide();
+	BannerNotifications.hideAll();
```
lizlux: Same

extensions/wikia/BannerNotifications/BannerNotificationsController.class.php
lizlux: We'll have to warn QA that the oasis notifications will need a pass since they are affected.

skins/oasis/js/WikiaFooter.js
bkoval: https://wikia-inc.atlassian.net/browse/SOC-401

extensions/wikia/BannerNotifications/js/BannerNotifications.js
bkoval: Let's do so, it's not used anywhere (at least by my investigation, which could have ommited something) :)

extensions/wikia/AssetsManager/config.php
```diff
@@ -922,7 +924,8 @@
 		'#group_imglzy_js',
 		'#group_spotlights_js',
 		'//resources/wikia/libraries/ghostwriter/gw.min.js',
-		'//extensions/wikia/GlobalNotification/GlobalNotification.js',
+		'#group_banner_notifications_js',
+		'//skins/shared/scripts/onScroll.js',
```
kvas-damian: ```#group_banner_notifications_js``` is dependent on ```//skins/shared/scripts/onScroll.js```. Don't you think we should solve this dependancy somehow?

extensions/wikia/BannerNotifications/templates/BannerNotifications.mustache
```diff
@@ -0,0 +1,6 @@
+<div class="banner-notification">
+	<button class="close wikia-chiclet-button">
+		<img src="{{imageSource}}">
```
kvas-damian: Is it possible to render this image using CSS?
This way we could limit 1 request for assets and make JS code cleaner.

extensions/wikia/BannerNotifications/js/BannerNotifications.js
```diff
@@ -0,0 +1,294 @@
...
+	}
+
+	/**
+	 * Creates an instance of a notification sets on the backend side
+	 * (if such one exists)
+	 */
+	function createBackendNotification() {
+		var $backendNotification = $('.banner-notification');
+		if ($backendNotification.length) {
+			backendNotification = new BannerNotification();
```
kvas-damian: Don't we need to pass at least ```type``` and ```content``` arguments?

extensions/wikia/BannerNotifications/js/BannerNotifications.js
```diff
@@ -0,0 +1,294 @@
...
+	}
+
+	/**
+	 * Bind close event to close button
+	 */
+	function setUpClose(notification) {
+		$(notification.$element).on('click', '.close', function (event) {
+			notification.hide();
+			if (notification.onCloseHandler) {
+				notification.onCloseHandler(event);
```
kvas-damian: what do you think about defining ```onClonseHandler``` as ```Function.prototype``` by default and just call it without checking if it's defined.

extensions/wikia/BannerNotifications/js/BannerNotifications.js
```diff
@@ -0,0 +1,294 @@
...
+		return modal.is(':visible');
+	}
+
+	/**
+	 * Removes notification element from DOM and executes callback
+	 * @param {jQuery} $elements - elements to be removed from DOM
+	 */
+	function removeFromDOM($elements) {
+		if ($elements.length) {
+			$elements.animate({
```
kvas-damian: Do you think this animation should be part of ```removeFromDOM``` method?

skins/shared/scripts/onScroll.js
```diff
@@ -0,0 +1,54 @@
...
+		handlers.forEach(function (handler) {
+			handler(event);
+		});
+	}
+
+	/**
+	 * Binds global scroll event listener
+	 */
+	function init() {
+		window.addEventListener('scroll', trigger);
```
kvas-damian: Don't you think we need at least small debounce for better performance?

extensions/wikia/BannerNotifications/js/BannerNotifications.js
kvas-damian: Do we support multiple instances of ```backendNofications```?

extensions/wikia/FacebookClient/scripts/preferences.js
bkoval: @kvas-damian What do you think about creating a ticket to do so and taking care of escaping messages problem on a separate basis?

extensions/wikia/WikiaInYourLang/modules/ext.wikiaInYourLang.js
bkoval: Went with the first approach.

extensions/wikia/FacebookClient/scripts/preferences.js
kvas-damian: Sounds good for me.

skins/shared/scripts/onScroll.js
bkoval: Let's add a minimal one.

extensions/wikia/BannerNotifications/templates/BannerNotifications.mustache
bkoval: Done.

extensions/wikia/AssetsManager/config.php
bkoval: After offline discussion with @kvas-damian , we decided to keep things the way there are and consider resolving dependency management for AssetManager as a seperate cause.

bkoval:
Few changes:
 - To avoid the undefined module problem, I wrapped the module into a 'require' block that exports an AMD module at the very end, to trigger the code right away
 - Divided controllers for BannerNotifications and Oasis Notifications
 - Added 'onClose' handler
 - Added little debounce (5ms) to onScroll module

extensions/wikia/BannerNotifications/css/BannerNotifications.scss
```diff
@@ -102,8 +83,36 @@
 		width: 14px;
 
 		img {
-			height: 11px;
-			width: 12px;
+			background-image: url('/skins/oasis/images/icon_close.png');
```
kvas-damian: Please use inline image to avoid additional request.

extensions/wikia/BannerNotifications/BannerNotificationsController.class.php
```diff
@@ -0,0 +1,292 @@
...
+
+	public function init() {
+		$this->confirmation = null;
+		$this->confirmationClass = null;
+	}
+
+	/**
+	 * Add confirmation message to the user session (so it persists between redirects)
+	 */
+	public static function addConfirmation($message, $type = 1) {
```
kvas-damian: please use constant for ```$type```

extensions/wikia/BannerNotifications/BannerNotificationsController.class.php
```diff
@@ -0,0 +1,292 @@
...
+	 * Show confirmation stored in user session
+	 */
+	public function executeConfirmation() {
+		wfProfileIn(__METHOD__);
+
+		// call hook to trigger user messages from extensions
+		#wfRunHooks('SkinTemplatePageBeforeUserMsg', array(&$ntl));
+
+		// add testing confirmation
+		// self::addConfirmation('test');
```
kvas-damian: I don't think we need debugging code here.

extensions/wikia/BannerNotifications/BannerNotificationsController.class.php
```diff
@@ -0,0 +1,292 @@
...
+
+		if (!empty($_SESSION[self::SESSION_KEY])) {
```
kvas-damian: whitespaces ;)

extensions/wikia/BannerNotifications/BannerNotificationsController.class.php
```diff
@@ -0,0 +1,292 @@
...
+			$entry = $_SESSION[self::SESSION_KEY];
+
+			$this->confirmation = $entry['message'];
+
+			switch($entry['type']) {
+				case self::CONFIRMATION_NOTIFY:
+					$this->confirmationClass = 'notify';
```
kvas-damian: what do you think about storing ```$type``` as a string value in session?
We could avoid this mapping then.

extensions/wikia/BannerNotifications/BannerNotificationsController.class.php
```diff
@@ -0,0 +1,292 @@
...
+	/**
+	 * Handle confirmations from special preferences
+	 */
+	public static function addPreferencesConfirmation(&$prefs) {
+		global $wgRequest;
+		wfProfileIn(__METHOD__);
+
+		if ( F::app()->checkSkin( 'oasis' ) ) {
+			if ($wgRequest->getCheck('success')) {
+				self::addConfirmation(wfMsg('savedprefs'));
```
kvas-damian: Please use new ```wfMessage``` instead of deprecated ```wfMsg``` and escape strings.

extensions/wikia/BannerNotifications/BannerNotificationsController.class.php
```diff
@@ -0,0 +1,292 @@
...
+
+	/**
+	 * Registering and adding JS package for front-end usage
+	 *
+	 * @param object $out
+	 * @param string $text
+	 *
+	 * @return true
+	 */
+	function onOutputPageBeforeHTML( &$out, &$text ) {
```
kvas-damian: please use full method signature - including visibility and static keyword

bkoval: static public it is! ;)

extensions/wikia/BannerNotifications/js/BannerNotifications.js
```diff
@@ -0,0 +1,306 @@
...
+	};
+
+	/**
+	 * Generic method for indicating problem with AJAX connection
+	 * @returns {BannerNotification}
+	 */
+	BannerNotification.prototype.showConnectionError = function () {
+		return this
+			.setType('error')
+			.setContent($.msg('bannernotifications-general-ajax-failure'))
```
kvas-damian: Please use escaped message ;)

lizlux:
@bkoval, I tried to check out the code and run it, but I'm seeing escaped HTML as the output in the browser. See screenshot for error message in page source: 

screen shot 2015-02-26 at 3 10 25 pm

extensions/wikia/BannerNotifications/js/BannerNotifications.js
```diff
@@ -0,0 +1,309 @@
...
+		} else {
+			$pageContainer = $('.WikiaPageContentWrapper');
+			headerHeight = $('#globalNavigation').height();
+			onScroll.bind(handleScrolling);
+		}
+		createBackendNotification();
+	}
+
+	/**
+	 * Creates an instance of a notification set on the backend side
```
lizlux: This message seems unclear to someone that doesn't know about notifications originating from the server. Maybe update the wording to something like: 

"Create instance of BannerNotification based on HTML passed from the server on page load."

bkoval:
@lizlux Could you try to checkout this branch again? I guess this is connected to the Mac's filesystem being case-insensitive, I forced the changes in git to use the capital letter now :)

lizlux:
@bkoval thanks for the quick fix!

lizlux:
I'm still seeing that JS error that the module BannerNotification isn't defined, right at page load. This is definitely a race condition tho b/c I'm consistently seeing it being defined after the page has fully loaded. 

screen shot 2015-02-26 at 3 35 44 pm

Saw the JS error here: http://lizlux.liz.wikia-dev.com/wiki/TabViewContent (just a random article on my wikia).

lizlux:
I must say the functionality is looking really good, and the API is really smooth. It's a pleasure to play around with it. 
:+1:

extensions/wikia/WikiaInYourLang/WikiaInYourLang.hooks.php
```diff
@@ -9,12 +9,12 @@
...
 
 	/**
-	 * Add JS module to the output
+	 * Add JS assets package to the output
 	 * @param  \OutputPage $out  An output object passed from a hook
 	 * @return bool
 	 */
 	public static function onBeforePageDisplay( \OutputPage $out ) {
-		$out->addModules( 'ext.wikiaInYourLang' );
+		\Wikia::addAssetsToOutput( 'wikia_in_your_lang_js' );
```
lizlux: Is this how we're supposed to be doing this now? I feel like there's 100 ways to do add assets. :grimacing:

skins/oasis/css/core/responsive-core.scss
```diff
@@ -13,19 +13,23 @@
...
-	.global-notification,
+	.banner-notifications-wrapper,
 	.WikiaHeader .page-width-container {
 		@include box-sizing(border-box);
 		width: $width-outside;
 	}
 
-	.global-notification {
+	.banner-notification {
+		@include box-sizing(border-box);
```
lizlux: We don't need a mixin for this anymore. `box-sizing: border-box` works on all supported browsers now. :)
http://caniuse.com/#search=box-sizing

Someone should really clean that up...

skins/oasis/modules/NotificationsController.class.php
```diff
@@ -7,14 +7,6 @@
 
 class NotificationsController extends WikiaController {
 
-	const SESSION_KEY = 'oasisConfirmation';
```
lizlux: (Not related to this line, but) should the doc block on line 3 be updated to remove the reference to "(green stripe below global nav)"? It looks like you split this file out into the two types of notifications (which is good).

extensions/wikia/BannerNotifications/BannerNotifications.setup.php
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Classes
+ */
+$wgAutoloadClasses[ 'BannerNotificationsController' ] = $dir . 'BannerNotificationsController.class.php';
+
+/**
+ * Hooks
+ */
+$wgHooks['ArticleDeleteComplete'][] = 'BannerNotificationsController::addPageDeletedConfirmation';
```
lizlux: It would be good to mention to QA that these are some back end instances to test, since it's less well know that these can come from the back end.

extensions/wikia/BannerNotifications/BannerNotificationsController.class.php
```diff
@@ -0,0 +1,291 @@
+<?php
+/**
+ * Handles adding and rendering of confirmations (green stripe below global nav) and notifications (blue bubble boxes just above the fold)
+ *
+ * @author Maciej Brencz
+ */
+
+class BannerNotificationsController extends WikiaController {
+
+	const SESSION_KEY = 'oasisConfirmation';
```
lizlux: Should we update this to `BannerNotification`?

extensions/wikia/BannerNotifications/BannerNotificationsController.class.php
```diff
@@ -0,0 +1,291 @@
...
+		wfProfileIn( __METHOD__ );
+
+		if ( !empty( $_SESSION[self::SESSION_KEY] ) ) {
+			$entry = $_SESSION[self::SESSION_KEY];
+
+			$this->confirmation = $entry['message'];
+			$this->confirmationClass = $entry['type'];
+
+			// clear confirmation stack
+			unset( $_SESSION[self::SESSION_KEY] );
```
lizlux: Looks like `clearConfirmation` is never being called. Let's either delete it or call it from here.

extensions/wikia/BannerNotifications/BannerNotificationsController.class.php
```diff
@@ -0,0 +1,291 @@
...
+		if ( !empty( $message ) ) {
+			$_SESSION[self::SESSION_KEY] = array(
+				'message' => $message,
+				'type' => $type,
+			);
+
+			wfDebug( __METHOD__ . " - {$message}\n" );
+		}
+
+		wfProfileOut( __METHOD__ );
```
lizlux: Let's rip these `wfProfileOut` `wfProfileIn` calls out since we don't need them anymore.

extensions/wikia/FacebookClient/FacebookClientHooks.class.php
```diff
@@ -127,7 +127,7 @@ public static function onSkinTemplatePageBeforeUserMsg( &$html ) {
 				// check if current user is connected to facebook
 				$map = FacebookClient::getInstance()->getMapping();
 				if ( !empty( $map ) ) {
-					NotificationsController::addConfirmation( wfMessage( 'fbconnect-connect-msg' )->plain() );
+					BannerNotificationsController::addConfirmation( wfMessage( 'fbconnect-connect-msg' )->plain() );
```
lizlux: `escaped()`

extensions/wikia/SpecialCss/SpecialCssController.class.php
```diff
@@ -41,29 +41,29 @@ public function index() {
 				);
 
 				if (!$status) {
-					NotificationsController::addConfirmation(
+					BannerNotificationsController::addConfirmation(
 						wfMessage('special-css-merge-error')->plain(),
```
lizlux: escape

extensions/wikia/SpecialCss/SpecialCssController.class.php
```diff
@@ -41,29 +41,29 @@ public function index() {
...
-						NotificationsController::CONFIRMATION_ERROR
+						BannerNotificationsController::CONFIRMATION_ERROR
 					);
 					$this->diff = $this->app->sendRequest(__CLASS__, 'getDiff', ['wikitext' => $content])->getVal('diff');
 				} else if ($status->isOk()) {
-					NotificationsController::addConfirmation( wfMessage('special-css-save-message')->plain() );
+					BannerNotificationsController::addConfirmation( wfMessage('special-css-save-message')->plain() );
```
lizlux: escape

extensions/wikia/SpecialCss/SpecialCssController.class.php
```diff
@@ -41,29 +41,29 @@ public function index() {
...
+						BannerNotificationsController::CONFIRMATION_ERROR
 					);
 					$this->cssContent = $content;
 				}
 		}
 
 		if ($this->request->getVal('oldid', null) !== null) {
-			NotificationsController::addConfirmation(
+			BannerNotificationsController::addConfirmation(
 				wfMessage('special-css-oldid-message')->plain(),
```
lizlux: escape (obviously not your code but would be good to update)

extensions/wikia/BannerNotifications/js/BannerNotifications.js
```diff
@@ -0,0 +1,309 @@
+/* global require, define*/
+/*
+ * Handles the color-coded notification messages that generally appear
+ * at the top of the screen or inside a modal.
+ *
+ * AMD module is defined at the bottom of the 'require' block
+ *
+ * Use like:
+ * new BannerNotification('Content', 'error').show().hide();
+ */
```
lizlux: Might be worth noting from a security perspective that this module doesn't escape content before inserting it into the DOM, so it's only to be used with content that's already considered safe (i.e. no UCG).

extensions/wikia/BannerNotifications/js/BannerNotifications.js
```diff
@@ -0,0 +1,309 @@
...
+require([
```
lizlux: Also, let's add some unit tests! (Can be done in a separate ticket)

skins/oasis/css/core/responsive-core.scss
kvas-damian: Lets do that: https://wikia-inc.atlassian.net/browse/SOC-445

bkoval:
Hey @lizlux ! Thanks for checking out the branch and for CR. This weird race condition worries me though. I was trying to reproduce it using SF devbox (dev-interactivemaps) and was wandering around different wikias for quite some time. I didn't see the race condition error though, all was loading fine:

https://www.dropbox.com/s/3sn4kvxe60yfuy4/Zrzut%20ekranu%202015-02-27%2011.41.25.png?dl=0

Not sure where does it come from, but I'll set up a Sandbox today and try to reproduce it as well.

extensions/wikia/WikiaInYourLang/WikiaInYourLang.hooks.php
bkoval: I like this way of adding assets, this makes it independent from the hook it is used from. But I'm open to suggestions if you feel there's a better way :)

extensions/wikia/BannerNotifications/BannerNotifications.setup.php
bkoval: Sure thing, I'll prepare a description for QA-ing this :)

bkoval:
https://wikia-inc.atlassian.net/browse/SOC-446 - ticket for adding unit tests.

kvas-damian:
Besides error mentioned by @lizlux it looks ok for me.
I tried to find it but with no luck. Liz are you still able to reproduce this issue? 
If yes, it'd be helpful if you provide us more info, like browser etc.

PR #6336, opened 2015-02-17
MAIN-2878 use replace instead of update (both for watched item and globa...

jsutterfield:
...l watchlist)

When working on UC-153, a query inside of WatchedItem.php was changed from an REPLACE to an UPDATE. This wasn't the same behavior and was causing problems when trying to rename a page to a title which already exists. Revert to the REPLACE query, and update the globalwatchlist logic to account for that case.

Ticket: https://wikia-inc.atlassian.net/browse/MAIN-3878


armonr:
LGTM

PR #6338, opened 2015-02-17
SOC-361 make $timestamp a default argument

jsutterfield:
Make the 3rd argument to `timestamp` default to null. This fixes PHP warnings which were arising from this method being called in places without the `timestamp` argument.

Ticket: https://wikia-inc.atlassian.net/browse/SOC-361


garthwebb:
LGTM!

PR #6340, opened 2015-02-18
DAT-2518 Performance fixes

sebastianmarzjan:
- Factor in limit earlier in the process
- Do not double ImageServing queries because of re-using a variable

@macbre

### DIFF
--- a/includes/wikia/api/ArticlesApiController.class.php
+++ b/includes/wikia/api/ArticlesApiController.class.php
@@ -358,18 +358,20 @@ class ArticlesApiController extends WikiaApiController {
 			$ns = array_unique( $ns );
 		}
 
-		$key = self::getCacheKey( self::NEW_ARTICLES_CACHE_ID, '', [ implode( '-', $ns ) , $minArticleQuality ] );
+		$key = self::getCacheKey( self::NEW_ARTICLES_CACHE_ID, '', [ implode( '-', $ns ) , $minArticleQuality, $limit ] );
 		$results = $this->wg->Memc->get( $key );
 		if ( $results === false ) {
-			$solrResults = $this->getNewArticlesFromSolr( $ns, self::MAX_NEW_ARTICLES_LIMIT, $minArticleQuality );
+			$solrResults = $this->getNewArticlesFromSolr( $ns, $limit, $minArticleQuality );
 			if ( empty( $solrResults ) ) {
 				$results = [];
 			} else {
 				$articles = array_keys( $solrResults );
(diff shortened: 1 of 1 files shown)


includes/wikia/api/ArticlesApiController.class.php
```diff
@@ -358,7 +358,7 @@ public function getNew(){
 			$ns = array_unique( $ns );
 		}
 
-		$key = self::getCacheKey( self::NEW_ARTICLES_CACHE_ID, '', [ implode( '-', $ns ) , $minArticleQuality ] );
+		$key = self::getCacheKey( self::NEW_ARTICLES_CACHE_ID, '', [ implode( '-', $ns ) , $minArticleQuality, $limit ] );
```
macbre: How many different values can `$limit` have? Maybe it would be better to set a maximum value of `$limit`, cache the whole results set and slice cached results when requested with smaller limit.

includes/wikia/api/ArticlesApiController.class.php
```diff
@@ -366,10 +366,12 @@ public function getNew(){
 				$results = [];
 			} else {
 				$articles = array_keys( $solrResults );
+				$articles = array_slice( $articles, 0, $limit );
+
```
macbre: :+1:

includes/wikia/api/ArticlesApiController.class.php
SebastianMarzjan: Cool idea, I'll take a look

macbre:
LGTM, thanks for making DB less busy ;)

includes/wikia/api/ArticlesApiController.class.php
SebastianMarzjan: It was actually the original implementation. This is why we needed to ask ImageServing for all of ids. I'll try to pull out the $limit distribution to see if change to this method actually makes any sense.

SebastianMarzjan: It looks like 80% of the queries are for limit=1 and limit=10 and ~0% on the maximum. Furthermore, we have caching on ImageServing end too (per article). I'd leave it as in this PR to avoid slicing on vast majority of requests.

macbre: Sure. As we use only two values in most of the cases, I'm fine with the current implementation. Thanks for looking into it :)

macbre:
Ok, I think we're fine to merge this one.

idradm:
LGTM

PR #6347, opened 2015-02-18

jsutterfield:

Note: the original pull request was #6297 which contains additional discussion. The commit history got mangled with a couple bad merges and it was just easier to cherry-pick out the important commits, put them in this new branch, and open a new pull request.

### DIFF
--- a/resources/wikia/libraries/jquery/loadLibrary/jquery.wikia.loadExternalLibrary.js
+++ b/resources/wikia/libraries/jquery/loadLibrary/jquery.wikia.loadExternalLibrary.js
@@ -65,8 +65,9 @@
 				version: 'v2.1'
 			});
 
+			// show facebook login button
 			$('.sso-login').removeClass('hidden');
-			// resolve after FB has finished inititalizing
+			// resolve after FB has finished initializing
 			$deferred.resolve();
 		};
 
--- a/resources/wikia/libraries/jquery/loadLibrary/jquery.wikia.loadExternalLibrary.js
+++ b/resources/wikia/libraries/jquery/loadLibrary/jquery.wikia.loadExternalLibrary.js
@@ -56,25 +56,28 @@
 			$deferred.done(callback);
 		}
 
-		// This is invoked by Facebook once the SDK is loaded.
-		window.fbAsyncInit = function () {
-			window.FB.init({
-				appId: window.fbAppId,
-				xfbml: true,
-				cookie: true,
-				version: 'v2.1'
-			});
-
-			// show facebook login button
-			$('.sso-login').removeClass('hidden');
-			// resolve after FB has finished initializing
+		if (typeof window.FB === 'object') {
+			// Since we have our own deferred object, we need to resolve it if FB is already loaded.
+			// We can't rely on the type check inside $.loadExternalLibrary.
 			$deferred.resolve();
-		};
+		} else {
+			// This is invoked by Facebook once the SDK is loaded.
+			window.fbAsyncInit = function () {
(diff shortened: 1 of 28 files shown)


resources/wikia/libraries/jquery/loadLibrary/jquery.wikia.loadExternalLibrary.js
```diff
@@ -46,7 +46,7 @@
 	 * @returns {jQuery} Returns a jQuery promise
 	 * @see https://developers.facebook.com/docs/javascript/quickstart/v2.2
 	 */
-	$.loadFacebookAPI = function (callback) {
+	$.loadFacebookSDK = function (callback) {
```
lizlux: This function needs the `$('.sso-login').removeClass('hidden')` code put back in.

armonr: Also, until we fix [SOC-240](https://wikia-inc.atlassian.net/browse/SOC-240), shouldn't `$('.sso-login').removeClass('hidden')` take place on `.done()`?

jsutterfield: Ah, I hadn't actually dove into that ticket. If that's the case, why don't we just remove the 'hidden' class from that FB login?

armonr: @jsutterfield Good question. This was done originally to only show FB controls if FB is reachable and not blocked as in countries like China. This is why we only display them on successful FB library load. But again, we'll change this soon as described in SOC-240, but not yet :)

jsutterfield: @armonr after looking into it, I think the placement of `$('.sso-login').removeClass('hidden')` should still be okay. It's inside of the `window.fbAsyncInit` function which is called by the FB SDK after it's successfully loaded. If there's an issue with the SDK (like it being blocked b/c you're in China), then the function will never get called and the login button will remain hidden.

lizlux: I would move it to the `done` method [here](https://github.com/Wikia/app/blob/SOC-261-2/resources/wikia/libraries/jquery/loadLibrary/jquery.wikia.loadExternalLibrary.js#L56) so that it gets called every time we load the library. This will have a better chance of accounting for `.sso-login` elements that weren't previously on the page if the library is being called for a second time. Again, it will change with SOC-240, but for now I'd say this would be safer.

lizlux:
@jsutterfield, this PR https://github.com/Wikia/app/pull/6296/files introduced a new instance of `$.loadFacebookAPI()` so that should be updated with this PR.

resources/wikia/libraries/jquery/loadLibrary/jquery.wikia.loadExternalLibrary.js
jsutterfield: @lizlux I think the problem with adding it there is that `$.loadFacebookSDK` is only called with a callback function in side of FacebookClient.XFBML.js. Everywhere else there's no callback. Maybe I could add an additional `$('.sso-login').removeClass('hidden')` [here](https://github.com/Wikia/app/blob/SOC-261-2/resources/wikia/libraries/jquery/loadLibrary/jquery.wikia.loadExternalLibrary.js#L62)?

PR #6350, opened 2015-02-19
Monetization (MON-347): MonetizationModule - updated styles to use sprite

saipetch:
Monetization (MON-347): use sprite in MonetizationModule


tedgill:
I assume the unit test failures are nonsense? If so, LGTM.

PR #6359, opened 2015-02-19
DAT-2540 edit mode styling for title and description

RafLeszczynski:


extensions/wikia/NjordPrototype/css/Njord.scss
```diff
@@ -261,14 +261,21 @@
 	}
 
 	.title-wrap {
-		margin: 30px 330px 20px 30px;
+		margin: 0;
```
SebastianMarzjan: We don't really need this line

extensions/wikia/NjordPrototype/css/Njord.scss
```diff
@@ -261,14 +261,21 @@
...
+			margin: 0px 330px 0px 20px;
+			padding: 30px 0px 25px 10px;
+		}
+
 		.edit-box {
-			background-color: rgba(0, 0, 0, .03);
 			bottom: -64px;
 			display: none;
 			width: 100%;
+			text-transform: uppercase;
```
SebastianMarzjan: Please move this before width (ordering)

SebastianMarzjan:
Two small things and good to go

extensions/wikia/NjordPrototype/css/Njord.scss
RafLeszczynski: yes we do, unless we do refactoring of this stylesheet which I'd like to do in the next iteration of MoM project

SebastianMarzjan: Can you elaborate? The following one seems to override all 4 values.

RafLeszczynski: sorry this line shouldn't be there it was a leftover that I forgot to remove. My mistake, and good catch :+1:

PR #6360, opened 2015-02-19
CE-1318 Add AutoFollow extension to DefaultSettings

adamkarminski:
See [CE-1318](https://wikia-inc.atlassian.net/browse/CE-1318)

Besides minor conventions and documentation fixes, the most important part here is to add AutoFollow to the wikia's DefaultSettings since it should be run site-wide and catch users of our officially supported languages.


lgarczewski:
LGTM.

PR #6361, opened 2015-02-19
CONCF-143 Manually remove html tags from caption

dianafa:
https://wikia-inc.atlassian.net/browse/CONCF-143

This ticket is connected with https://github.com/Wikia/app/pull/6318. 
Maybe it's not the best piece of code, but it seems to be a xss safe and good looking solution.

### DIFF
--- a/extensions/wikia/WikiaMobile/js/modal.js
+++ b/extensions/wikia/WikiaMobile/js/modal.js
@@ -241,7 +241,7 @@ define( 'modal', ['throbber', 'jquery'], function modal ( throbber, $ ){
 
 	function setCaption ( cap ) {
 		if ( typeof cap === 'string' && cap !== '' ) { 			
-			caption.innerText = cap.replace(/<(?:.|\n)*?>/gm, '');	
+			caption.innerText = $('<i>').html(cap).text();	
 			caption.style.display = 'block';
 		} else {
 			caption.style.display = 'none';
--- a/extensions/wikia/WikiaMobile/js/modal.js
+++ b/extensions/wikia/WikiaMobile/js/modal.js
@@ -240,11 +240,8 @@ define( 'modal', ['throbber', 'jquery'], function modal ( throbber, $ ){
 	}
 
 	function setCaption ( cap ) {
-		var escapedCaption;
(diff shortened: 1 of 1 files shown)


extensions/wikia/WikiaMobile/js/modal.js
```diff
@@ -240,11 +240,8 @@ define( 'modal', ['throbber', 'jquery'], function modal ( throbber, $ ){
 	}
 
 	function setCaption ( cap ) {
-		var escapedCaption;
-		if ( typeof cap === 'string' && cap !== '' ) {
-			escapedCaption = document.createTextNode(cap);
-			caption.innerHTML = '';
-			caption.appendChild(escapedCaption);
+		if ( typeof cap === 'string' && cap !== '' ) { 			
+			caption.innerText = cap.replace(/<(?:.|\n)*?>/gm, '');
```
rogatty: >m modifier: multi-line. Causes ^ and $ to match the begin/end of each line (not only begin/end of string)

It doesn't make sense here, right?

rogatty: How about `/<[^>]*>?/g`?
http://stackoverflow.com/questions/822452/strip-html-from-text-javascript#comment4970970_822464

extensions/wikia/WikiaMobile/js/modal.js
```diff
@@ -240,11 +240,8 @@ define( 'modal', ['throbber', 'jquery'], function modal ( throbber, $ ){
 	}
 
 	function setCaption ( cap ) {
-		var escapedCaption;
-		if ( typeof cap === 'string' && cap !== '' ) {
-			escapedCaption = document.createTextNode(cap);
-			caption.innerHTML = '';
-			caption.appendChild(escapedCaption);
+		if ( typeof cap === 'string' && cap !== '' ) { 			
+			caption.innerText = $('<i>').html(cap).text();
```
rogatty: `$('<i>').html("").text()`

How about `/<[^>]*>?/g`?
http://stackoverflow.com/questions/822452/strip-html-from-text-javascript#comment4970970_822464

rogatty:
This is obsolete after https://github.com/Wikia/app/pull/6369

PR #6364, opened 2015-02-19
MOB-3421 remove CuratedContentPreview

sqreek:
Remove Special:CuratedContentPreview. It's shouldn't be on production yet.


nandy-andy:
Cool, 404 when you're trying visit the failed PHP unit test page on Jenkins :smile_cat:

nandy-andy:
LGTM - red means removing ;) 

However, how about wrapping it with a condition checking a new WF variable. This way next time you'll be ready to release it you'll be able to just switch a WF variable value - no code changes needed. Just an idea - maybe you decided to do it this way for some reasons I'm not familiar with.

sqreek:
Curated content preview, is not ready for production. It's need more work there, and using WF variable is too early. Someone could change it, and You get not working page.

PR #6365, opened 2015-02-19
SOC-365 fixing gallery toggle button float issue

lizlux:
Since we're no longer floating the gallery items (they are `display:inline-block`), we don't have to set `clear:both` on the show more/less buttons. Removing the clear property of that button wrapper avoids issues where the buttons clear a tall floated item (like an infobox) and are separated vertically from the gallery items.


jsutterfield:
LGTM!

PR #6368, opened 2015-02-19

jsutterfield:

Note: the original pull requests were #6297 and #6347 which contain additional discussion. Somehow I mangled both of those branches and it was easier to create a new one. After this ticket I'm going to sit down and have a long discussion with git about how we can work together better in the future.


lizlux:
LGTM!

PR #6371, opened 2015-02-19
[WikiaApiQueryPageinfo::getPageViews] The database query returns 114k+ rows

macbre:
Fix the performance of `WikiaApiQueryPageinfo` by avoiding querying `page_visited` DB table without `WHERE` provided.

The fix is in df1eb35

@michalroszka


michalroszka:
:+1:

PR #6379, opened 2015-02-20
Ignore .mweval_history

onei:
@Grunny Using `maintenance/eval.php` creates a log file: `maintenance/.mweval_history`. This makes git ignore it :)


Grunny:
:+1:

PR #6387, opened 2015-02-20
Concf 154

bognix:


extensions/wikia/VideoHandlers/VideoHandlers.setup.php
```diff
@@ -278,7 +274,6 @@
 	27 => 'Twitchtv',
 	28 => 'Ooyala',
 	29 => 'Iva',
-	30 => 'Snappytv',
```
bognix: @lizlux @saipetch @garthwebb - I didn't know if it's ok to fix the ordering in this array. I searched the code and I didn't find any relation between Ustream and id:31 (the same for the next ones). Guys, do you know about possible side effects of changing ids in this array?

lizlux: @saipetch would know the most about this. I'm unfamiliar with this code.

garthwebb: @bognix, this array isn't being used anymore and you can remove this element without changing the index of any of the other elements.  I recommend this whole array should be removed and the code that relies on it, but that's a code cleanup decision for the content team at this point.

bognix: I created the ticker for removing not used code: https://wikia-inc.atlassian.net/browse/CONCF-159.
For now I'm not going to fix the indexing in array as it's not needed at all.

vforge:
LGTM

hakubo:
LGTM

hakubo:
Tests ar passing.

PR #6390, opened 2015-02-20
Mercury opt in for whole wikia

Warkot:
Cookie was set only for the specific wiki domain (e.g. `.elderscrolls.wikia.com`), now user will have the mercury skin forced/not forced on the broader domain: `.wikia.com` or `.wikia-dev.com` (but custom domains stay the same like `.wowwiki.com`).


hakubo:
LGTM, and messages :)

PR #6391, opened 2015-02-20
CE-1309: Unify font size and fix textarea visability after CE-1322 fix

lukaszkonieczny:
https://wikia-inc.atlassian.net/browse/CE-1309


Grunny:
LGTM.

PR #6392, opened 2015-02-20
Make a softer is_object() check - selectRow return value inconsistency

adamkarminski:
Since Database.php method selectRow() can't decide if it wants to return an stdClass object or a ResultWrapper one - the check should be soften to accept valid stdClass results.

Ping @macbre


macbre:
LGTM, please double check unit tests failure if it's related.

adamkarminski:
```
Time: 11.73 seconds, Memory: 102.25Mb

OK (58 tests, 212 assertions)
```

`./php-all` run on devbox

PR #6393, opened 2015-02-20
SOC-281 Update Captcha messaging

jsutterfield:
With the updated captcha, the help text that went along with the old captcha was no longer accurate. This ticket updates that messaging.

Ticket: https://wikia-inc.atlassian.net/browse/SOC-281


garthwebb:
LGTM

PR #6274, opened 2015-02-12
PLATFORM-868: allow MW_INSTALL_PATH to control $IP in wikia-land also

owend:
This change should be transparent, leaving us in the same place we were before, just using realpath() instead of **DIR**.  However, it allows us to override the path based on an env var. @macbre

PR #6398, opened 2015-02-23
Revert "PLATFORM-868: allow MW_INSTALL_PATH to control $IP in wikia-land...

macbre:
Reverts Wikia/app#6274

@owend


owend:
Ok, looks like we need to figure out a different approach here.

PR #6402, opened 2015-02-23
CONCF-125 add wikiaVariables caching for 1 minute

dianafa:
https://wikia-inc.atlassian.net/browse/CONCF-125

### DIFF
--- a/extensions/wikia/MercuryApi/MercuryApiController.class.php
+++ b/extensions/wikia/MercuryApi/MercuryApiController.class.php
@@ -9,7 +9,8 @@ class MercuryApiController extends WikiaController {
 	const NUMBER_CONTRIBUTORS = 5;
 	const DEFAULT_PAGE = 1;
 
-	const WIKIA_VARIABLES_CACHE_TTL = 60;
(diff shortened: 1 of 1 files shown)


extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -256,6 +257,9 @@ public function getWikiVariables() {
 
 		$this->response->setVal( 'data', $wikiVariables );
 		$this->response->setFormat( WikiaResponse::FORMAT_JSON );
+
+		//cache wikiaVariables for 1 minute		
+		$this->response->setCacheValidity( self:: WIKIA_VARIABLES_CACHE_TTL);
```
macbre: Can we cache it for a bit longer time, like an hour or a day? And invalidate the cache when required.

dianafa: invalidating cache will require quite a lot of effort in this case. Can I find you tomorrow and can we talk about this?

macbre: Sure :)

hakubo:
LGTM

PR #6409, opened 2015-02-23
Add X-Request-Id header to Helios calls

wladekb:
This adds X-Request-Id derived from MW's one so we'll be able to correctly correlate events on both ends.

/cc @michalroszka @ArturKlajnerok


ArturKlajnerok:
LGTM :+1:

PR #6411, opened 2015-02-23
Soc 163 Adding form validation to facebook singup form

lizlux:
Ticket: https://wikia-inc.atlassian.net/browse/SOC-163

This PR attempts to unify the way we do form validation in the UserLogin extension. It involves changing the name attribute of the username and password inputs for a number of forms, so that they can use the same validation mechanisms.

While I was in this code I noticed some bugs with changing a password from the login modal and the facebook modal, each have been ticketed and there's some comments in the code.

Old PR was https://github.com/Wikia/app/pull/6294


extensions/wikia/UserLogin/js/UserLoginModal.js
```diff
@@ -80,51 +78,18 @@
 						window.FacebookLogin.init(window.FacebookLogin.origins.MODAL);
 					}
 
-					UserLoginModal.loginAjaxForm = new window.UserLoginAjaxForm($loginModal, {
-						ajaxLogin: true,
-						// context is this instance of UserLoginAjaxForm
-						callback: function (res) {
```
lizlux: I was able to simplify this section quite a bit by tracing back all the places that set these options and call these functions.

lizlux:
Creating new PRs to handle changes on a more individual level.

PR #6413, opened 2015-02-24
turn off exhibiton for categories with more then 2k results

idradm:


extensions/wikia/CategoryExhibition/CategoryExhibitionSection.class.php
```diff
@@ -5,6 +5,7 @@
  */
 class CategoryExhibitionSection {
 	const CACHE_VERSION = 3;
+	const EXHIBITION_LIMIT = 2000;
```
SebastianMarzjan: This limit could use a word of explanation - what is the rationale behind it being 2000 and not for example 1500, as this knowledge is available neither in code nor in the referenced ticket.

extensions/wikia/CategoryExhibition/services/CategoryDataService.class.php
```diff
@@ -54,11 +54,9 @@ public static function getArticleCount( $sCategoryDBKey, $mNamespace = '', $nega
 
 		$db = wfGetDB( DB_SLAVE );
 
-		// Get a count of articles in a category.  Give at least a very small cache TTL
-		$query = (new WikiaSQL())->cache( 5 )
-			->SELECT( 'count(distinct page_title)' )->AS_( 'count' )
+		$query = (new WikiaSQL())
```
SebastianMarzjan: I'm trying to wrap my head around what why previous version of this query seems so complicated with no apparent reason. Do you happen to know?

idradm: Not really. This code was never used AFAIK.

extensions/wikia/CategoryExhibition/CategoryExhibitionSection.class.php
idradm: It's a guess. No rationale behind. Do you have any other proposals?

SebastianMarzjan: No - I was curious, as will be the next person to touch this :)

SebastianMarzjan:
GTG

PR #6414, opened 2015-02-24
ADEN-1563 Enable Krux on Mobile site-wide

nandy-andy:
After a testing period we're ready to enable Krux on mobile (WikiaMobile skin) site-wide.

Changes below do this and additionally they disable Krux when `noexternals=1` is passed in query params.


gabrys:
LGTM

PR #6421, opened 2015-02-24
CONCF-163 remove race condition from global nav scripts

bognix:


vforge:
LGTM

extensions/wikia/GlobalNavigation/scripts/GlobalNavigationHubsMenu.js
```diff
@@ -6,8 +6,8 @@ require(
 function ($, w, GlobalNavLazyLoad, menuAim, browserDetect, delayedHover, dropdowns) {
 	'use strict';
 
-	var $entryPoint = $('#hubsEntryPoint'),
-		$hubs = $('#hubs'),
+	var $entryPoint,
```
hakubo: Should they be 'global'? I don't see if they are used anywhere else.

PR #6426, opened 2015-02-24
SOC-119 | Allow user to change facebook url

kvas-damian:
https://wikia-inc.atlassian.net/browse/SOC-119
Additionaly all QQQs for user profile page were updated


extensions/wikia/UserProfilePageV3/UserProfilePage.i18n.php
```diff
@@ -132,38 +133,164 @@
...
  */
 $messages['qqq'] = array(
-	'userprofilepage-question-caption' => 'Parameters:
+	'userprofilepage-edit-modal-header' => 'Header for edit profile modal on user profile page.',
+	'userprofilepage-question-caption' => 'Currently not in use!
+	Parameters:
 * $1 is the number of the current question
 * $2 is the total number of questions in the interview.',
-	'userprofilepage-question-list-title' => 'Parameters:
+	'userprofilepage-lightbox-interview-cancel' => "Currently not in use!",
```
kvas-damian: Unused messages are going to be cleaned up in separate ticket: https://wikia-inc.atlassian.net/browse/SOC-416

extensions/wikia/UserProfilePageV3/UserProfilePage.i18n.php
```diff
@@ -132,38 +133,164 @@
...
+	'user-identity-box-avatar-choose-avatar' => 'Label that encourages user to choose one of default avatar pictures on avatar picture edit modal.',
+	'user-identity-box-avatar-save' => "Label for save button on avatar picture edit modal.",
+	'user-identity-box-avatar-cancel' => 'Label for cancel button on avatar picture edit modal. Click on this button close modal or displays confirmation message if any changes where already made.',
+	'user-identity-box-avatar-anon-user-error' => 'Error message that is displayed when anon user tries to change his avatar. It tells that user needs to login before he change his avatar.',
+	'user-identity-box-hide-edits-wikis' => 'Label for checkbox on user profile edit modal. When this field is checked user favorite wikis are not displayed on user profile page.',
+	'user-identity-box-banned-from-chat' => 'Label for tag displayed just after user login on user profile page which tell that used is banned for using Wikia chat.',
+
+	'user-identity-box-aka-label' => "Message that is displayed below user name on user profile page that shows user real name. In English there is 'aka - also known as' in HTML span tag (it's required for styling).
```
kvas-damian: HTML in messages will be removed in: https://wikia-inc.atlassian.net/browse/SOC-417

extensions/wikia/UserLogin/js/spec/UserBaseAjaxForm.spec.js
```diff
@@ -0,0 +1,54 @@
+/* global UserBaseAjaxForm */
+
+describe('Base class for User Ajax Forms', function () {
+	'use strict';
+
+	var instance, $el, options,
+		proto = UserBaseAjaxForm.prototype,
+		$userNameInput = $('<input type="text" name="username">'),
```
rafalkalinski: I would rather wrote it like that:
```javascript
$('<input>', {
  type: 'text',
  name: 'username'
})
```
It looks like much more complicated. And It has also the advantages: No likely that one could tell "Use some template for that markup, mate" :), and IMO it is more convenient to maintenance later.

kvas-damian: Ha! You found some code which wasn't touched by me :)
I need to investigate what happened there.

rafalkalinski: ok - now I see what module is this. So probably - nevermind :)

rafalkalinski:
LGTM

extensions/wikia/UserLogin/js/spec/UserBaseAjaxForm.spec.js
kvas-damian: Fixed after dev merge.

rafalkalinski:
still LGTM

rafalkalinski:
LGTM again :)

extensions/wikia/UserProfilePageV3/css/UserProfilePage_modal.scss
```diff
@@ -15,13 +16,22 @@
...
 			width: 100%;
 		}
+
+		.tweet-at, .fb-start {
+			float: left;
+			margin-top: 10px;
+			width: 110px;
+		}
+
+		input[name="fbPage"], input[name="twitter"] {
```
jsutterfield: I think this css needs to be tweaked a little bit. When I go to enter my facebook profile link the cursor starts to the left of the trailing /.
screen shot 2015-03-02 at 11 18 37 am

extensions/wikia/UserProfilePageV3/templates/UserProfilePage_renderUserIdentityBox.php
```diff
@@ -4,162 +4,164 @@
...
 	<div class="masthead-info">
 		<hgroup>
-			<? if( !empty($user['realName']) ): ?>
-				<h1 itemprop="name"><?= htmlspecialchars($user['name']); ?></h1>
-				<h2><?= wfMsg('user-identity-box-aka-label', array('$1' => $user['realName']) ); ?></h2>
-			<? else: ?>
-				<h1 itemprop="name"><?= htmlspecialchars($user['name']); ?></h1>
+			<h1 itemprop="name"><?= htmlspecialchars( $user['name'] ); ?></h1>
+			<? if ( !empty( $user['realName'] ) ): ?>
+				<h2><?= wfMessage( 'user-identity-box-aka-label', $user['realName'] )->plain(); ?></h2>
```
jsutterfield: This should probably be an `escaped()` call, rather than a `plain()` call.

extensions/wikia/UserProfilePageV3/templates/UserProfilePage_renderUserIdentityBox.php
```diff
@@ -4,162 +4,164 @@
...
 		<div class="masthead-info-lower">
 			<div class="tally">
-				<? if( !empty($user['registration']) ): ?>
-					<? if( !empty($user['edits']) || (empty($user['edits']) && !empty($user['registration'])) ): ?>
+				<? if ( !empty( $user['registration'] ) ): ?>
+					<? if ( !empty( $user['edits'] ) || ( empty( $user['edits'] ) && !empty( $user['registration'] ) ) ): ?>
 						<em><?= $user['edits'] ?></em>
 						<span>
-							<?= wfMsg('user-identity-box-edits-since-joining', array( $user['registration'] ) ) ?>
+							<?= wfMessage( 'user-identity-box-edits-since-joining', $user['registration'] )->plain() ?>
```
jsutterfield: `plain()` -> `escaped()`

extensions/wikia/UserProfilePageV3/templates/UserProfilePage_renderUserIdentityBox.php
```diff
@@ -4,162 +4,164 @@
...
 						</span>
 					<? else: ?>
-						<?php if( $user['edits'] >= 0 ): ?>
-							<?= wfMsg('user-identity-box-edits', array( '$1' => $user['edits'] ) ); ?>
+						<?php if ( $user['edits'] >= 0 ): ?>
+							<?= wfMessage( 'user-identity-box-edits', $user['edits'] )->plain(); ?>
```
jsutterfield: `plain()` -> `escaped()`

extensions/wikia/UserProfilePageV3/templates/UserProfilePage_renderUserIdentityBox.php
```diff
@@ -4,162 +4,164 @@
...
 						<?php else: ?>
-							<br />
+							<br/>
 						<?php endif; ?>
 					<?php endif; ?>
 				<? else: ?>
-					<?php if( $user['edits'] >= 0 ): ?>
-						<?= wfMsg('user-identity-box-edits', array( '$1' => $user['edits'] ) ); ?>
+					<?php if ( $user['edits'] >= 0 ): ?>
+						<?= wfMessage( 'user-identity-box-edits', $user['edits'] )->plain(); ?>
```
jsutterfield: `plain()` -> `escaped()`

extensions/wikia/UserProfilePageV3/templates/UserProfilePage_renderUserIdentityBox.php
```diff
@@ -4,162 +4,164 @@
...
+					<? if ( $user['showZeroStates'] && ( $isUserPageOwner || $canEditProfile ) ): ?>
+						<li><?= wfMessage( 'user-identity-box-zero-state-location' )->escaped(); ?></li>
 					<? endif; ?>
 				<? endif; ?>
-				
-				<? if( !empty($user['birthday']) && intval( $user['birthday']['month'] ) > 0 && intval( $user['birthday']['month'] ) < 13 ): ?>
-					<li><?= wfMsg('user-identity-box-was-born-on', array( '$1' => F::app()->wg->Lang->getMonthName( intval($user['birthday']['month']) ), '$2' => $user['birthday']['day'] )); ?></li>
+
+				<? if ( !empty( $user['birthday'] ) && intval( $user['birthday']['month'] ) > 0 && intval( $user['birthday']['month'] ) < 13 ): ?>
+					<li><?= wfMessage( 'user-identity-box-was-born-on', F::app()->wg->Lang->getMonthName( intval( $user['birthday']['month'] ) ), $user['birthday']['day'] )->plain(); ?></li>
```
jsutterfield: `plain()` -> `escaped()`

extensions/wikia/UserProfilePageV3/templates/UserProfilePage_renderUserIdentityBox.php
```diff
@@ -4,162 +4,164 @@
...
 				<? endif; ?>
-				
-				<? if( !empty($user['occupation']) ): ?>
-					<li><?= wfMsg('user-identity-box-occupation', array( '$1' => $user['occupation'] )); ?></li>
-				<? elseif( !empty($user['showZeroStates']) ): ?>
-					<? if( $user['showZeroStates'] && ($isUserPageOwner || $canEditProfile) ): ?>
-					<li><?= wfMsg('user-identity-box-zero-state-occupation'); ?></li>
+
+				<? if ( !empty( $user['occupation'] ) ): ?>
+					<li><?= wfMessage( 'user-identity-box-occupation', $user['occupation'] )->plain(); ?></li>
```
jsutterfield: `plain()` -> `escaped()`

extensions/wikia/UserProfilePageV3/templates/UserProfilePage_renderLightbox.php
```diff
@@ -81,36 +82,46 @@
...
+					<label for="twitter" class="tweet-at">twitter.com/</label>
+					<input type="text" id="twitter" name="twitter" value="<?= $user['twitter']; ?>">
 				</div>
 				<div class="input-group">
-					<label for="twitter"><?= wfMsg('user-identity-box-about-tweet'); ?></label>
-					<span class="tweet-at">@</span>
-					<input type="text" name="twitter" value="<?= $user['twitter']; ?>">
+					<label for="fbPage"><?= wfMessage( 'user-identity-box-about-fb-page' )->escaped(); ?></label>
+					<label for="fbPage" class="fb-start">facebook.com/</label>
+					<input type="text" id="fbPage" name="fbPage" value="<?= $user['fbPage']; ?>">
```
jsutterfield: This isn't code you touched as part of this ticket, but we should probably update [this line](https://github.com/Wikia/app/blob/dev/extensions/wikia/UserProfilePageV3/UserIdentityBox.class.php#L127) to be:
```php
$data[ $key ] = htmlspecialchars( strip_tags( $data[ $key ] ), ENT_QUOTES );
```
Without `ENT_QUOTES`, this is only escaping the double quotes. I couldn't identify any places where this would currently be an issues, but if the above line were ever changed from `value="<?= $user['fbPage']; ?>"` to `value='<?= $user['fbPage']; ?>'`, the user could input `'onmouseover='alert(0);` and get it to execute. `ENT_QUOTES` makes sure to escape single quotes as well.

Props to @grunny and his security talk. Made me think to look for this kind of thing :)

extensions/wikia/UserProfilePageV3/templates/UserProfilePage_renderUserIdentityBox.php
kvas-damian: We can't escape messages contining HTML. This will available when we finish: https://wikia-inc.atlassian.net/browse/SOC-417

extensions/wikia/UserProfilePageV3/templates/UserProfilePage_renderUserIdentityBox.php
kvas-damian: HTML in meesages

extensions/wikia/UserProfilePageV3/templates/UserProfilePage_renderUserIdentityBox.php
kvas-damian: HTML in messages

extensions/wikia/UserProfilePageV3/templates/UserProfilePage_renderUserIdentityBox.php
kvas-damian: and again html in messages

extensions/wikia/UserProfilePageV3/templates/UserProfilePage_renderUserIdentityBox.php
kvas-damian: For sure we'll do that after we finish cleanup of translations for UPP.

extensions/wikia/UserProfilePageV3/templates/UserProfilePage_renderUserIdentityBox.php
kvas-damian: Yes, I promise ;)

extensions/wikia/UserProfilePageV3/css/UserProfilePage_modal.scss
kvas-damian: @jsutterfield are you sure you used up to date code?
I think we change ```@``` to ```tweeter.com``` already.

extensions/wikia/UserProfilePageV3/templates/UserProfilePage_renderLightbox.php
kvas-damian: Escaping single quote sounds ok for links, on the other hand this methos is escaping other attributes like ```realName``` which imo, should be able to display single quotes.
@jsutterfield : What do you think?

extensions/wikia/UserProfilePageV3/templates/UserProfilePage_renderUserIdentityBox.php
jsutterfield: Haha :)

extensions/wikia/UserProfilePageV3/templates/UserProfilePage_renderLightbox.php
Grunny: Just for reference, the single quotes will still display fine to the user, they'll just be HTML encoded (as `&#039;`) in the source, so there's no problem doing that for all this data if that simplifies things. If you're escaping the content of attributes, you can make use of the helper method [`Sanitizer::encodeAttribute`](https://github.com/Wikia/app/blob/5690ef15aa2/includes/Sanitizer.php#L1120-L1133) which does this and cleans up whitespace a bit. Also, `wfMessage()->escaped()` always [escapes single quotes](https://github.com/Wikia/app/blob/5690ef15aa2/includes/Message.php#L405), so you don't have to worry if you see that in an attribute either. :)

extensions/wikia/UserProfilePageV3/css/UserProfilePage_modal.scss
jsutterfield: Oops, looks like you're correct. I probably had an old branch. I'm seeing the new forms that you mention. That was my mistake.

extensions/wikia/UserProfilePageV3/templates/UserProfilePage_renderLightbox.php
jsutterfield: Awesome, thanks for the clarification @Grunny. I hadn't considered @kvas-damian point considering the validity of apostrophe's in non-link attributes, but I'm glad to hear that those cases will still be handled properly. These values aren't only used as attributes, they're sometimes found inside of `<span>` tags, so I think using ` Sanitizer::encodeAttribute` might be slightly misleading but it's good to know for the future.

Damian, it sounds like we'd be fine to add that `ENT_QUOTES` option here and it should make this code more secure. Since we're at the tail end of the sprint and it'd be great if Rebekah could QA this today, I'm going to make that change and move this into QA. If you have any objections or concerns we can certainly revert that change, but I think it makes sense given the circumstances.

extensions/wikia/UserProfilePageV3/templates/UserProfilePage_renderUserIdentityBox.php
jsutterfield: Awesome :+1:

jsutterfield:
LGTM!

extensions/wikia/UserProfilePageV3/templates/UserProfilePage_renderLightbox.php
kvas-damian: Thanks @jsutterfield for your change!
@Grunny great explanation. 
What do you think about creating best practises for escaping?
Having nice rule could increase our efficiency in this area. For example when displaying data, sanitizing should happen as late as possible - in our case it's template. Quickly I can enumerate some advantages of this approach:
* we won't escape data 2 times by mistake
* we don't need to go back and check if data was escpaed already or not
* we could use proper sanitizing methods for different places
* data that is sent though API is not escaped in controller and by API again

The same rule probably can be applied for data that is going to be saved (in model). What do you think about it?

Grunny: @kvas-damian absolutely :) those points match what I mentioned in the presentation of "escape as close to the output as possible", which is a standard best practice for secure code. It reduces reviewer anxiety and is in general the best practice for escaping and writing provably secure code.

I've added those notes from the presentation here: https://one.wikia-inc.com/wiki/Security/Cross-site_scripting#Best_practices_for_escaping But we can possibly also put the recommendation for moving HTML output escaping to the templates where they belong in the guidelines repo or somewhere else as well. Any thoughts on that?

kvas-damian: That's great.
Did you consider moving articles about different vulnerabilities (and how to avoid them) into guidelines repo and linke them from Wikia One?
New security section could be beneficial also for developers that do not have access to Wikia One.

Grunny: Sounds like a plan, I'll write up a PR for it soon. :)

kvas-damian: :+1:

PR #6428, opened 2015-02-24
MAIN-3949 Suggested way to implement an addData method

garthwebb:
This would be a way to set several keys in the response data without replacing everything (as with setData) or repeated individual sets (as with setVal)

https://wikia-inc.atlassian.net/browse/MAIN-3949

### DIFF
--- a/includes/wikia/nirvana/WikiaResponse.class.php
+++ b/includes/wikia/nirvana/WikiaResponse.class.php
@@ -191,6 +191,14 @@ class WikiaResponse {
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


Grunny:
This can already be done with [`WikiaResponse::setValues`](https://github.com/Wikia/app/blob/0b1fff161bad/includes/wikia/nirvana/WikiaResponse.class.php#L379-L381). Possibly needs renaming, but the functionality is the same.

garthwebb:
Ah, yah, we didn't notice that method.  Thanks!

PR #6429, opened 2015-02-25
Soc 163 Adding form validation to facebook singup form

lizlux:
Adding ajax validation to the facebook signup form. 

Previous PRs include https://github.com/Wikia/app/pull/6294 and https://github.com/Wikia/app/pull/6411. Those PRs were too big and tried to handle too much code cleanup at once. I'll try to address all of that code cleanup in separate PRs.


extensions/wikia/UserLogin/js/UserSignupAjaxValidation.js
```diff
@@ -14,26 +14,20 @@
 	var UserSignupAjaxValidation = function (options) {
 		this.wikiaForm = options.wikiaForm;
 		this.inputsToValidate = options.inputsToValidate || [];
-		this.notEmptyFields = options.notEmptyFields || [];
 		this.submitButton = $(options.submitButton);
-
-		this.activateSubmit();
-	};
-
-	UserSignupAjaxValidation.prototype.activateSubmit = function () {
```
lizlux: We're no longer disabling the submit button until a form validates on the front end. We're just going to let users submit whenever they want, since back end validation will pick it up.

extensions/wikia/UserLogin/js/FacebookLogin.js
```diff
@@ -287,6 +287,30 @@
...
+
+			validator = new UserSignupAjaxValidation({
+				wikiaForm: wikiaForm,
+				inputsToValidate: inputsToValidate,
+				submitButton: inputs.submit
+			});
+
+			// Add validation on blur event for all inputs to validate
+			inputsToValidate.forEach(function (inputName) {
+				inputs[inputName].on('blur', validator.validateMappedInput.bind(validator));
```
rafalkalinski: I guess that 'inputs' is an array of jQuery objects, so isn't there a chance to bind this validation to blur event only once? I mean, not like in this loop, but once using prepared, filtered jQuery collection.
It could be much more effective from performance point of view.

extensions/wikia/UserLogin/js/UserSignupAjaxValidation.js
```diff
@@ -14,26 +14,20 @@
...
+	/**
+	 * Call back end for validation of input where user action occurred
+	 * @param {Event} e Browser event like input blur
+	 */
 	UserSignupAjaxValidation.prototype.validateInput = function (e) {
 		var el = $(e.target),
 			paramName = el.attr('name'),
 			params = this.getDefaultParamsForAjax();
 
+		console.log(typeof e);
```
rafalkalinski: should this be here? :)

extensions/wikia/UserLogin/js/UserSignupAjaxValidation.js
```diff
@@ -44,14 +38,41 @@
...
+	 * @todo This was copied from validateInput with some added input name mapping. Clean this up in a separate ticket.
+	 * @param {Event} e Browser event like input blur
+	 */
+	UserSignupAjaxValidation.prototype.validateMappedInput = function (e) {
+		var el = $(e.target),
+			paramName = el.attr('name'),
+			params = this.getDefaultParamsForAjax(),
+			mappedParamName;
+
+		if (paramName === 'username') {
```
rafalkalinski: Probably it would be better to create an object with mapped params names and then just use something like jQuery.extend. It may not be shorter than these ifs, but for longer terms it would be much easier to maintain.

armonr:
Other than Rafał's comments, LGTM

extensions/wikia/UserLogin/js/FacebookLogin.js
lizlux: I donno, maybe I just can't think of a good way to do this, but it's awkward b/c the second parameter in `$.fn.on` is a string selector, not a jQuery object. So I could theoretically do something like:

```javascript
selector = '[name=' + inputsToValidate.join('], [name=') + ']';
```
Then I could pass that `selector` string as the second argument like:

```javascript
wikiaForm.el.on('blur', selector, validate));
```
That's the only way I can see to make it more performant, b/c I would be attaching the event listener to the parent element as opposed to the inputs themselves. 

To me though, that code look brittle and unclear to the next person that come across it. In theory it would be nice to attach only one event listener but in practice I think it's fine to leave this as is. @rafalkalinski, could you think of a better way to handle this?

extensions/wikia/UserLogin/js/UserSignupAjaxValidation.js
lizlux: Gah! No.

extensions/wikia/UserLogin/js/UserSignupAjaxValidation.js
lizlux: Good call. Updated.

extensions/wikia/UserLogin/js/FacebookLogin.js
rafalkalinski: Yeah, this joining selectors - although it's easy, it is also unreadable in this particular case.
And maybe sth like that:
```javascript
filtered = $();
inputsToValidate.forEach(function (inputName) {
   filtered.add (inputs[inputName]);
})
filtered.on ('blur', validate);
```

Loop is similar, but only one binding.

lizlux: Cool, yeah, logically that makes sense, so I did a test, and indeed, it is faster to bind once. http://jsperf.com/event-bindings-with-jquery. Will fix!

rafalkalinski:
LGTM

PR #6430, opened 2015-02-25
SOC-333 Remove last bits of the old FB SDK

garthwebb:
https://wikia-inc.atlassian.net/browse/SOC-333


lizlux:
Looks like `extensions/wikia/GameGuides/assets/GameGuidesAssets.json` and `extensions/wikia/CuratedContent/assets/CuratedContentAssets.json` still reference the old sdk.

extensions/wikia/UserLogin/UserLoginSpecialController.class.php
```diff
@@ -60,7 +60,7 @@ private function initializeTemplate() {
 		$this->response->addAsset(
 				( $this->wg->request->getInt( 'recover' ) === 1 || empty( $this->wg->EnableFacebookClientExt ) ) ?
 					'userlogin_js_wikiamobile' :
-					'userlogin_js_wikiamobile_fbconnect'
+					'userlogin_facebook_js_wikiamobile'
```
lizlux: The `userlogin_js_wikiamobile` group was being included in `userlogin_js_wikiamobile_fbconnect`, but it's not in `userlogin_js_wikiamobile_fbconnect`. The result is that `/extensions/wikia/UserLogin/js/UserLoginSpecial.wikiamobile.js` isn't being loaded on the login page anymore. 

I think the logic here should be changed to always include `userlogin_js_wikiamobile`, since that's what was happening before anyway, and optionally include `userlogin_facebook_js_wikiamobile`.

Or you could rearrange the assets manager groups to reflect this logic.

jsutterfield:
LGTM. Just need to remove those sdk references from `GameGuidesAssets.json` and `CuratedContentAssets.json` and this should be good to go.

armonr:
Shouldn't those two places be replaced with sdk.js rather than removed? And [SOC-421](https://wikia-inc.atlassian.net/browse/SOC-421) should cover localization of them.

jsutterfield:
Ah, yes @armonr, that's definitely what I meant. Good catch :)

PR #6436, opened 2015-02-25
DAT-2562 limit phase one release of Hero Image to english wikis only

RafLeszczynski:


extensions/wikia/CreateNewWiki/CreateWiki.php
```diff
@@ -1189,4 +1189,13 @@ public function getWikiInfo($key) {
...
 		return $ret;
 	}
+
+	/**
+	 * sets initial value for wgEnableNjordExt for new created wiki
+	 * TODO: for first phase of prototype set to true for 10% of new created english wikis only
+	 * @return bool
+	 */
+	private function setInitialNjordExtValue() {
+		return intval( $this->mNewWiki->city_id ) % 10 === 1 && $this->mNewWiki->language === 'en' ? true : false;
```
SebastianMarzjan: In this pull request only wikis with city_id ending with 1 that are in EN will have Hero Image enabled.
You might want to consider using rand(0,9) instead of $this->mNewWiki->city_id to ensure fair treatment of all IDs.

SebastianMarzjan: Ok, correlation coefficient is 0.000000000011 so there IS no correlation ;-)

RafLeszczynski: ok I'm changing the logic

SebastianMarzjan:
LG

PR #6441, opened 2015-02-25
AvatarService: fix PHP Strict Standards

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-855

```
PHP Strict Standards: Only variables should be passed by reference in /usr/wikia/slot1/3338/src/includes/wikia/services/AvatarService.class.php on line 271
```

@wladekb

### DIFF
--- a/includes/wikia/services/AvatarService.class.php
+++ b/includes/wikia/services/AvatarService.class.php
@@ -269,8 +269,7 @@ class AvatarService extends Service {
 				$url = self::buildVignetteUrl( $width, $bucket, $relativePath, $timestamp, false );
 			}
 		} else { // default avatar
-			$defaultAvatars = $masthead->getDefaultAvatars( 'thumb/' );
-			$legacyDefaultUrl = array_shift( $defaultAvatars );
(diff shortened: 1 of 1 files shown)


includes/wikia/services/AvatarService.class.php
```diff
@@ -269,7 +269,8 @@ public static function getVignetteUrl( Masthead $masthead, $width, $timestamp )
 				$url = self::buildVignetteUrl( $width, $bucket, $relativePath, $timestamp, false );
 			}
 		} else { // default avatar
-			$legacyDefaultUrl = array_shift( $masthead->getDefaultAvatars( 'thumb/' ) );
+			$defaultAvatars = $masthead->getDefaultAvatars( 'thumb/' );
+			$legacyDefaultUrl = array_shift( $defaultAvatars );
```
wladekb: how about using reset() if we don't use the rest of the array?

macbre: Or we can actually call `[0]` on the result of `getDefaultAvatars`. I'll change that tomorrow.

wladekb:
lgtm

PR #6445, opened 2015-02-25
VOLDEV-123 Unhardcode a hardcoded message on Special:Version

ty-a:
@Grunny

### DIFF
--- a/includes/specials/SpecialVersion.php
+++ b/includes/specials/SpecialVersion.php
@@ -344,7 +344,7 @@ class SpecialVersion extends SpecialPage {
 		}
 
 		if ( count( $wgFileExtensions ) ) {
-			$out .= $this->openExtType('File extensions allowed for upload');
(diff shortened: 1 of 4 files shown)


includes/specials/SpecialVersion.php
```diff
@@ -344,7 +344,7 @@ function getExtensionCredits() {
 		}
 
 		if ( count( $wgFileExtensions ) ) {
-			$out .= $this->openExtType('File extensions allowed for upload');
+			$out .= $this->openExtType( wfMsg( 'version-file-extensions-allowed' ) );
```
Grunny: I know the rest of the file is probably using the deprecated `wfMsg` methods, but can you switch this to the newer i18n method: `wfMessage( 'version-file-extensions-allowed' )->plain()`? `->plain()` is safe output format here since it is being escaped when it is output.

Grunny:
:+1:

PR #6447, opened 2015-02-25
Fix SQL query in WikiFactory::getGroups

wladekb:
Small difference and 23k+ fewer unnecessary rows returned from db.

https://wikia-inc.atlassian.net/browse/PLATFORM-915

/cc @macbre


macbre:
LGTM

PR #6448, opened 2015-02-25
Soc 391 Cookie Policy Extension Translations

lizlux:
Downloading translations from crowdin for CookiePolicy extension. 

Also adding the conf file b/c it wasn't added previously.


armonr:
:shipit:

PR #6449, opened 2015-02-26
Captcha display on userSignup should respect wgUserSignupDisableCaptcha

armonr:
Two fixes:
- Hook being disabled on user creation now points to the new Captcha extension
- On displaying captcha on the userSignup page, `wgUserSignupDisableCaptcha` is respected now.

IDE made some stubborn whitespace fixes. for easier read use: https://github.com/Wikia/app/pull/6449/files?w=1

https://wikia-inc.atlassian.net/browse/SOC-287

### DIFF
--- a/extensions/wikia/UserLogin/UserSignupSpecialController.class.php
+++ b/extensions/wikia/UserLogin/UserSignupSpecialController.class.php
@@ -65,7 +65,7 @@ class UserSignupSpecialController extends WikiaSpecialPageController {
 	 * @responseParam string msg - result message
 	 * @responseParam string errParam - error param
 	 */
-	public function signupForm () {
+	public function signupForm() {
 		$this->wg->Out->setPageTitle( wfMessage( 'usersignup-page-title' )->plain() );
 		$this->response->addAsset( 'extensions/wikia/UserLogin/css/UserSignup.scss' );
 
@@ -109,7 +109,7 @@ class UserSignupSpecialController extends WikiaSpecialPageController {
 		// template params
 		$this->pageHeading = wfMessage( 'usersignup-heading' )->escaped();
 		$this->createAccountButtonLabel = wfMessage( 'createaccount' )->escaped();
-		if( $this->byemail ) {
+		if ( $this->byemail ) {
 			$this->pageHeading = wfMessage( 'usersignup-heading-byemail' )->escaped();
 			$this->createAccountButtonLabel = wfMessage( 'usersignup-createaccount-byemail' )->escaped();
 		}
@@ -175,10 +175,17 @@ class UserSignupSpecialController extends WikiaSpecialPageController {
 
 	public function captcha() {
 		$this->rawHtml = '';
-		$captchaObj = $this->getCaptchaObj();
-		if ( !empty( $captchaObj ) ) {
-			$this->rawHtml = $captchaObj->getForm();
-			$this->isFancyCaptcha = ( class_exists( 'Captcha\Module\FancyCaptcha' ) && $captchaObj instanceof Captcha\Module\FancyCaptcha );
+		$this->isFancyCaptcha = false;
+
+		if ( $this->shouldDisplayCaptcha() ) {
+			$captchaObj = $this->getCaptchaObj();
+			if ( !empty( $captchaObj ) ) {
+				$this->rawHtml = $captchaObj->getForm();
+				$this->isFancyCaptcha = (
+					class_exists( 'Captcha\Module\FancyCaptcha' ) &&
+					$captchaObj instanceof Captcha\Module\FancyCaptcha
+				);
+			}
 		}
 	}
 
@@ -232,7 +239,7 @@ class UserSignupSpecialController extends WikiaSpecialPageController {
 			$ret = $signupForm->addNewAccount();
 		}
 
-		$this->result = ( $signupForm->msgType == 'error' ) ? $signupForm->msgType : 'ok' ;
+		$this->result = ( $signupForm->msgType == 'error' ) ? $signupForm->msgType : 'ok';
 		$this->msg = $signupForm->msg;
 		$this->errParam = $signupForm->errParam;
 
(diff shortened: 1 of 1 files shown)


extensions/wikia/UserLogin/UserSignupSpecialController.class.php
```diff
@@ -621,36 +634,52 @@ public function formValidation() {
...
-		$isMobile = $this->app->checkSkin( 'wikiamobile' );
-		$isAutomatedTest = in_array( $this->wg->Request->getIP(), $this->wg->AutomatedTestsIPsList );
-		$isNoCaptchaTest = $this->wg->Request->getInt( 'nocaptchatest' ) == 1;
-
-		//Disable captcha for automated tests and wikia mobile and sites disabling it
-		if ( $isMobile || ( $isAutomatedTest && $isNoCaptchaTest ) || $this->wg->UserSignupDisableCaptcha ) {
-			//Switch off global var
-			$this->wg->WikiaEnableConfirmEditExt = false;
-			//Remove hook function
-			$hookArrayKey = array_search( 'ConfirmEditHooks::confirmUserCreate', $wgHooks['AbortNewAccount'] );
```
armonr: `ConfirmEditHooks::confirmUserCreate` -> `Captcha\Hooks::confirmUserCreate`

extensions/wikia/UserLogin/UserSignupSpecialController.class.php
```diff
@@ -175,10 +175,17 @@ public function signupForm () {
 
 	public function captcha() {
 		$this->rawHtml = '';
-		$captchaObj = $this->getCaptchaObj();
-		if ( !empty( $captchaObj ) ) {
-			$this->rawHtml = $captchaObj->getForm();
-			$this->isFancyCaptcha = ( class_exists( 'Captcha\Module\FancyCaptcha' ) && $captchaObj instanceof Captcha\Module\FancyCaptcha );
+		$this->isFancyCaptcha = false;
+
+		if ( $this->shouldDisplayCaptcha() ) {
```
armonr: Check if captcha should be displayed

owend:
Setup sequence of constructor -> disableCapcha() -> shouldDisplayCapcha()  is confusing.  However, the fix works and we need to merge asap, so maybe we can refactor that later.

extensions/wikia/UserLogin/UserSignupSpecialController.class.php
```diff
@@ -621,36 +634,52 @@ public function formValidation() {
...
 	}
 
 	/**
 	 * Disables User Signup Captcha for automated tests, mobile skin, and sites such as internal
-	 *
-	 * @throws MWException
 	 */
 	private function disableCaptcha() {
+		// Disable captcha for automated tests and wikia mobile and sites disabling it, e.g. Internal
+		if ( $this->shouldDisplayCaptcha() ) {
```
lizlux: I think it would be more clear if this check happened before calling `disableCaptcha`, inside the constructor. 

```php
if ( !$this->shouldDisplayCaptcha() ) {
    disableCaptcha();
}
```

extensions/wikia/UserLogin/UserSignupSpecialController.class.php
```diff
@@ -313,33 +320,39 @@ public function sendConfirmationEmail() {
...
-						$this->msg = $response->getVal( 'msg','' );
-						$this->heading = wfMessage( 'usersignup-confirmation-heading-email-resent' )->escaped();
-					} else if( $this->result == 'error' ) {
-						$this->msgEmail = $response->getVal( 'msg','' );
-						$this->errParam = $response->getVal( 'errParam', '' );
-					} else if ( $this->result == 'confirmed' ) {
-						$this->heading = wfMessage( 'usersignup-confirm-page-heading-confirmed-user' )->escaped();
-						$this->subheading = wfMessage( 'usersignup-confirm-page-subheading-confirmed-user' )->escaped();
-						$this->msg = $response->getVal( 'msg','' );
+				} else {
```
lizlux: what's the point of splitting this into two lines?

armonr: This part was just whitespace fixes.

extensions/wikia/UserLogin/UserSignupSpecialController.class.php
armonr: Good point. This is somewhat of an interesting code flow with constructor calling this off the bat. To be cleaned in the next captcha refactoring round : )

extensions/wikia/UserLogin/UserSignupSpecialController.class.php
```diff
@@ -175,10 +175,17 @@ public function signupForm () {
...
-			$this->rawHtml = $captchaObj->getForm();
-			$this->isFancyCaptcha = ( class_exists( 'Captcha\Module\FancyCaptcha' ) && $captchaObj instanceof Captcha\Module\FancyCaptcha );
+		$this->isFancyCaptcha = false;
+
+		if ( $this->shouldDisplayCaptcha() ) {
+			$captchaObj = $this->getCaptchaObj();
+			if ( !empty( $captchaObj ) ) {
+				$this->rawHtml = $captchaObj->getForm();
+				$this->isFancyCaptcha = (
+					class_exists( 'Captcha\Module\FancyCaptcha' ) &&
```
garthwebb: You don't need to check if a class exists before using ```instanceof```

extensions/wikia/UserLogin/UserSignupSpecialController.class.php
garthwebb: I think Liz is asking why you've split the ```else if````'s into separate ```else``` and ```if```` blocks.  Doing this produces unnecessarily deeply indented code.  It also turns what reads easily as a case-like structure into something less obvious.

extensions/wikia/UserLogin/UserSignupSpecialController.class.php
garthwebb: This function was already named poorly given that it was called ```disableCaptcha``` but then had several tests for when to not disable captcha.  It definitely violates the principle of least surprise.

I agree that the check should move to the constructor and I think it should also be renamed to what its actually doing; checking if we should *disable* captcha (given that it starts with the assumption that captcha is enabled and then tests two cases that would break that assumption.)  That way we wouldn't have to test a negative and the constructor would read more clearly:
```php
if ( $this->shouldDisableCaptcha ) {
    $this->disableCaptcha();
}
```
Also, I think when this is refactored ```disableCaptcha``` should probably move out of the ```UserSignupSpecialController``` and into ```CaptchaController```.

extensions/wikia/UserLogin/UserSignupSpecialController.class.php
armonr: Yeah, intellij did this automatically after I added google styling for java... sigh!

extensions/wikia/UserLogin/UserSignupSpecialController.class.php
armonr: this was another whitespace fix; I will address them in another ticket along with refactor

PR #6457, opened 2015-02-26
CONCF-135 check if caption not empty on video file page

bognix:
Hide caption with number of views on video file page if number of views is not a number (null).
It is caused by a bug with inserting videos to video_info table.
Ticket for investigation of this issue was created - https://wikia-inc.atlassian.net/browse/CONCF-176


vforge:
LGTM

PR #6466, opened 2015-02-26
SOC-230 Use FancyCaptcha as a fallback if reCaptcha fails

jsutterfield:
Currently we're using reCaptcha as our default captcha. We want to make sure however, if reCaptcha fails to load (eg, the user is in China and Google is blocked), we have a fallback version. For that we're going to use FancyCaptcha. This ticket adds FancyCaptcha as that fallback.

Ticket: https://wikia-inc.atlassian.net/browse/SOC-230

ping @inez


extensions/VisualEditor/modules/ve-mw/init/targets/ve.init.mw.ViewPageTarget.js
```diff
@@ -576,9 +576,29 @@ ve.init.mw.ViewPageTarget.prototype.onSaveErrorNewUser = function ( isAnon ) {
  * @method
  */
 ve.init.mw.ViewPageTarget.prototype.onSaveErrorCaptcha = function () {
-	// Wikia change: Only support reCAPTCHA
+	// Wikia change: Only support reCAPTCHA and FancyCaptcha
 	this.captchaResponse = null;
 	this.saveDialog.$captcha.empty();
+
+	// reCaptcha loaded without any problems
+	if (this.saveDialog.frame.$element[0].contentWindow.grecaptcha !== undefined) {
```
jsutterfield: Captcha.js takes care of loading FancyCaptcha and swapping it out in the DOM for all captchas outside of VE. If reCaptcha fails to load when inside of VE however, `this.saveDialog.frame.$element[0].contentWindow.grecaptcha` is undefined and an error will be thrown here. Additionally, Captcha.js expects the reCaptcha form to have been injected into the page when falling back to FancyCaptcha. This isn't the case for VE so some additional logic had to be added.

extensions/VisualEditor/modules/ve-mw/init/targets/ve.init.mw.ViewPageTarget.js
```diff
@@ -587,12 +607,24 @@ ve.init.mw.ViewPageTarget.prototype.onSaveErrorCaptcha = function () {
...
+ * load Fancy Captcha and render it instead.
+ *
+ * @method
+ */
+ve.init.mw.ViewPageTarget.prototype.loadAndRenderFancyCaptcha = function () {
+	$.nirvana.sendRequest({
+		controller: 'CaptchaController',
+		method: 'getFancyCaptcha',
+		type: 'GET',
+		callback: function (data) {
```
lizlux: VE actually follows the MW whitespace guidelines, so there should be spaces in all these parentheses.

extensions/wikia/Captcha/CaptchaController.class.php
```diff
@@ -5,31 +5,33 @@
...
-	public function __construct() {
-		$this->captcha = Captcha\Factory\Module::getInstance();
-		parent::__construct();
-	}
-
 	/**
-	 * Displays a captcha image
+	 * Displays a captcha image. This is used exclusively by FancyCaptcha.
 	 */
 	public function showImage() {
```
lizlux: Is this getting called anywhere?

extensions/wikia/Captcha/scripts/Captcha.js
```diff
@@ -0,0 +1,65 @@
...
+			if (this.useCaptcha) {
+				$.loadReCaptcha().fail(
+					this.handleCaptchaLoadError.bind(this)
+				);
+			}
+		},
+
+		/**
+		 * Captcha is required for signup, so if it fails to load (possibly b/c google is blocked in China)
+		 * inform the user with a modal. Note, this is different from when a user fails the captcha test itself.
```
lizlux: Comment should be updated to reflect new behavior.

extensions/wikia/Captcha/scripts/Captcha.js
```diff
@@ -0,0 +1,65 @@
...
+
+	/**
+	 * JS for handling captchas We default to using reCaptcha, however if we encounter any problems, will
+	 * fall back to using Fancy Captcha instead.
+	 */
+	var Captcha = {
+		/**
+		 * WikiaMobile, Wikia One, and some automated tests do not use captcha
+		 */
+		useCaptcha: !window.wgUserSignupDisableCaptcha,
```
lizlux: This global is actually specific to the user signup page, and it only really gets set from the user signup controller. I think it shouldn't matter in any user-facing way, b/c if you're not on the Special:UseSignup page, this global var won't be set, but maybe the comment should reflect the fact that it's specific to user signup, just so people don't get confused in the future.

extensions/wikia/Captcha/scripts/Captcha.js
```diff
@@ -0,0 +1,65 @@
...
+		},
+
+		insertFancyCaptcha: function (data) {
+			$('.g-recaptcha').replaceWith(data.form);
+		},
+
+		trackFailedReCaptcha: function () {
+			Wikia.Tracker.track({
+				action: Wikia.Tracker.ACTIONS.ERROR,
+				category: 'user-sign-up',
```
lizlux: This category is no longer accurate since this JS affects more than just user signup now. Maybe talk to Cristina to see if changing it would throw off her numbers.

extensions/wikia/Captcha/CaptchaHooks.class.php
```diff
@@ -58,4 +58,12 @@ public static function APIGetAllowedParams( &$module, &$params ) {
 	public static function APIGetParamDescription( &$module, &$desc ) {
 		return Factory\Module::getInstance()->APIGetParamDescription( $module, $desc );
 	}
+
+	// Temporarily load captcha library on every page until we have on demand loading set up.
+	// Remove with https://wikia-inc.atlassian.net/browse/SOC-288
```
lizlux: I think the wording in SOC-288 should be updated to mention the new plan, which is to load the `captcha_js` AM group server side via those hooks as well as on the UserSignup page.

lizlux:
This is interesting: 

screen shot 2015-02-26 at 5 23 16 pm

That's the view when the user submits the form without passing the fancy captcha challenge. The difference is that the surrounding div has a class of `fancy-captcha`, rather than `captcha`. I think the screenshot is how it was intended to look, although I wouldn't say it looks awesome either way.

inez:
Overall looks great, but let me review more carefully tomorrow.

Anyways, very happy that I'm not the only person touching VisualEditor code :) And happy that you are able to work on it without "training", "knowledge sharing", etc.

extensions/VisualEditor/modules/ve-mw/init/targets/ve.init.mw.ViewPageTarget.js
```diff
@@ -937,7 +969,10 @@ ve.init.mw.ViewPageTarget.prototype.getSaveFields = function () {
 
 	ve.extendObject( fields, {
 		'wpSummary': this.saveDialog ? this.saveDialog.editSummaryInput.getValue() : this.initialEditSummary,
-		'g-recaptcha-response': this.captchaResponse
+		'g-recaptcha-response': this.captchaResponse, // reCaptcha
+		'wpCaptchaClass': this.saveDialog.$('#wpCaptchaClass').val(), // FancyCaptcha (fallback if reCaptcha fails to load)
```
inez: Same here - missing spaces

extensions/VisualEditor/modules/ve-mw/init/ve.init.mw.Target.js
```diff
@@ -667,8 +667,8 @@ ve.init.mw.Target.prototype.onSaveError = function ( doc, saveData, jqXHR, statu
 	// "question" or "fancy" type of captcha. They all expose differently named properties in the
 	// API for different things in the UI. At this point we only support the FancyCaptha which we
 	// very intuitively detect by the presence of a "url" property.
-	// Wikia-only change: support only reCAPTCHA
-	if ( editApi && editApi.captcha && editApi.captcha.type === 'recaptcha' ) {
+	// Wikia-only change: support only reCAPTCHA and FancyCaptcha
+	if ( editApi && editApi.captcha && (editApi.captcha.type === 'recaptcha' || editApi.captcha.type ==='fancyCaptcha')) {
```
inez: Missing spaces

extensions/wikia/Captcha/CaptchaHooks.class.php
```diff
@@ -58,4 +58,12 @@ public static function APIGetAllowedParams( &$module, &$params ) {
...
+	public static function onSkinAfterBottomScripts( $skin, &$text ) {
```
inez: Ok, this could sacrifice site performance and generate unnecessary load. Is there a solid reason why to introduce this temporary change (loading captcha always) instead of loading in on demand right away (within this PR)?

extensions/wikia/Captcha/CaptchaController.class.php
garthwebb: Yup:

https://github.com/Wikia/app/pull/6466/files#diff-eef94858475e8db5460aaa39c8f81b82R67

extensions/wikia/Captcha/Factory/Module.class.php
```diff
@@ -18,19 +21,25 @@ class Module {
...
+		return self::$instance;
+	}
 
-			// Check to see if this captcha can be used
-			if ( self::$instance->isValid() ) {
-				break;
-			}
+	/**
+	 * Make sure the Captcha class we are about to instantiate is what we expect. Either the default or fallback
+	 * version. This is especially important since 'wpwpCaptchaClass' is passed in by the client.
```
garthwebb: Too many ```wp```'s here

extensions/wikia/Captcha/CaptchaHooks.class.php
jsutterfield: Fair point @inez. No particular reason other than we had them ticketed out separately since they fell into somewhat discrete tasks. Loading them on demand wasn't too hard to implement though so I went ahead and added it to this PR.

jsutterfield:
Good catch on that css different @lizlux. I'm making sure to always apply the css stylings to Fancy Captcha now, but made a few edits which I think makes it easier to read and far less awkward. It looks like this now.
screen shot 2015-03-02 at 5 54 49 pm

extensions/VisualEditor/modules/ve-mw/init/targets/ve.init.mw.ViewPageTarget.js
```diff
@@ -587,12 +607,30 @@ ve.init.mw.ViewPageTarget.prototype.onSaveErrorCaptcha = function () {
...
+/**
+ * If reCaptcha failed to load (eg, the user is in China and google is blocked),
+ * load Fancy Captcha and render it instead.
+ *
+ * @method
+ */
+ve.init.mw.ViewPageTarget.prototype.loadAndRenderFancyCaptcha = function () {
+	$.when(
+		$.getResources([
+			$.getSassCommonURL('extensions/wikia/Captcha/styles/FancyCaptcha.scss')
```
jsutterfield: This won't currently get applied since the save dialogue is in an iframe, however @inez reports that the iframes are going away in the next 3-4 weeks at which point these styles will start working. Note the FancyCaptcha.scss are getting applied everywhere outside of VE.

inez:
@jsutterfield Just a heads up that I'm currently working in another branch on VisualEditor upstream sync - porting new VisualEditor from WMF to Wikia from scratch - which means I will need to port your changes to current/old VisualEditor as well.

jsutterfield:
Sounds good @inez, thanks for the heads up.

extensions/VisualEditor/wikia/modules/ve/ui/styles/ve.ui.Dialog.scss
```diff
@@ -43,7 +43,8 @@
 	max-height: 480px;
 }
 
-#ve-ui-mwSaveDialog-captcha > div {
+#ve-ui-mwSaveDialog-captcha > div,
+#ve-ui-mwSaveDialog-captcha > p {
```
lizlux: Is this still needed or can we leave it at div since you updated the fancy captcha template to have a {{div}} wrapper?

extensions/wikia/Captcha/scripts/Captcha.js
```diff
@@ -0,0 +1,73 @@
+$(function () {
+	'use strict';
+
+	/**
+	 * JS for handling captchas We default to using reCaptcha, however if we encounter any problems, will
```
lizlux: Period after "JS for handling captchas"

extensions/wikia/UserLogin/js/UserSignup.js
```diff
@@ -22,7 +22,6 @@
 			this.wikiaForm = new WikiaForm('#WikiaSignupForm');
 			this.submitButton = this.wikiaForm.inputs.submit;
 
-			this.loadCaptcha();
```
lizlux: Looks like L13-16 can be removed.

extensions/wikia/UserLogin/js/UserSignup.js
```diff
@@ -30,47 +29,6 @@
...
-			});
-		},
-
-		createCaptchaLoadErrorModal: function (uiModal) {
-			var modalConfig = {
-				vars: {
-					id: 'catchaLoadErrorModal',
-					classes: ['captcha-load-error-modal'],
-					size: 'medium',
-					title: $.msg('usersignup-page-captcha-load-fail-title'),
```
lizlux: Are these messages still needed for anything? If not, let's remove them from the i18n files and from the setup file that exports them to JS.

extensions/VisualEditor/wikia/modules/ve/ui/styles/ve.ui.Dialog.scss
jsutterfield: Nope. Good catch.

extensions/wikia/UserLogin/js/UserSignup.js
jsutterfield: Good call. They're not. I removed the messages.

garthwebb:
Looks good @jsutterfield, that FancyCaptcha is quasi pimpin!

jsutterfield:
LOL, took me a second to figure out where that came from @garthwebb :)

inez:
@jsutterfield , @garthwebb just a reminder go get this merged

PR #6467, opened 2015-02-26
Handle missing databases gently.

ghost:
A connection to a language specific database is required to check
whether a given SharedHelp article exists. There is no guarantee that
such database exists for the given language. The check itself is of
minor importance. Non-existing database implies that the article
does not exists, too. Log it and let the check to return false.


lizlux:
LGTM (after CI tests finish)

PR #6473, opened 2015-02-27
ADEN-1792 Turned off skyscrapers on pages without responsive Oasis

nandy-andy:
71media skyscrapers look bad on hub pages which aren't responsive and have fixed width. We decided to turn them off.

I also fixed small JSLint issues.


gabrys:
LGTM

PR #6480, opened 2015-02-27
Helios code-level improvements

wladekb:
I went through the Helios code and found a few improvements that can be made so I'm suggesting them. Please have a look and comment.

/cc @michalroszka @ArturKlajnerok @macbre

### DIFF
--- a/extensions/wikia/Helios/HelperController.class.php
+++ b/extensions/wikia/Helios/HelperController.class.php
@@ -26,14 +26,15 @@ class HelperController extends \WikiaController
 
 		// Allow the given user name if the AntiSpoof extension is disabled.
 		if ( empty( $this->wg->EnableAntiSpoofExt ) ) {
-			return $this->response->setVal( 'allow', true );
+			$this->response->setVal( 'allow', true );
+			return;
 		}
 
 		$oSpoofUser = new \SpoofUser( $sName );
-		$aConflicts = $oSpoofUser->getConflicts();
 
 		// Allow the given user name if it is legal and does not conflict with other names.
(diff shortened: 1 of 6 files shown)


extensions/wikia/Helios/HelperController.class.php
```diff
@@ -26,14 +26,15 @@ public function checkAntiSpoof()
 
 		// Allow the given user name if the AntiSpoof extension is disabled.
 		if ( empty( $this->wg->EnableAntiSpoofExt ) ) {
-			return $this->response->setVal( 'allow', true );
+			$this->response->setVal( 'allow', true );
+			return;
 		}
 
 		$oSpoofUser = new \SpoofUser( $sName );
-		$aConflicts = $oSpoofUser->getConflicts();
```
wladekb: AFAIU this is redundant because the same method is called again a few lines below while the $aConflicts is never used.

extensions/wikia/Helios/HelperController.class.php
```diff
@@ -81,7 +83,7 @@ public function sendConfirmationEmail()
 
 		$sName = $this->getVal( 'username' );
 
-		$this->wf->WaitForSlaves( $this->wg->ExternalSharedDB );
+		wfWaitForSlaves( $this->wg->ExternalSharedDB );
```
wladekb: We no longer need to use $app->wf-> (the only exception is $app->wg-> )

extensions/wikia/Helios/User.class.php
```diff
@@ -78,7 +80,6 @@ public static function authenticate( $sUserName, $sPassword )
 
 		// Authenticate with username and password.
 		try {
-			global $wgHeliosLoginShadowMode;
```
wladekb: unused

macbre:
LGTM

PR #6484, opened 2015-02-27
(CE-1394) Fix loading of some assets in SMW 1.8

Grunny:
Fix asset path for SMW 1.8 as it is located in the extensions/wikia
directory. Also, fix jquery.qtip error with trying to use amd in a way
that isn't compatible.

/cc @lgarczewski


lgarczewski:
LGTM.

PR #6486, opened 2015-02-27
CON-174 | Global Navigation search submit button hover state fix

Warkot:


extensions/wikia/GlobalNavigation/styles/GlobalNavigationColors.scss
```diff
@@ -15,3 +15,7 @@ $global-navigation-background-color: rgb(230, 235, 242);
 $search-button-background: rgb(238, 241, 244);
 $search-button-active-background: rgb(85, 85, 85);
 $search-button-hover-background: rgb(9, 33, 64);
+$search-input-background-color: rgb(242, 246, 250);
+$search-input-border-color: rgb(197, 206, 217);
+$search-input-text-color: rgba(101, 110, 120, .75);
+$autocomplete-border-bottom: rgb(230, 235, 250);
```
bognix: why did you move those variables here if they are only used in GlobalNavigationSearch.scss file?

Warkot: @bognix I moved even more colors that are used only in the GlobalNavigatgionSearch.scss file.

bognix:
LGTM

PR #6497, opened 2015-03-02
CONCF-175 empty search prevention

dianafa:
https://wikia-inc.atlassian.net/browse/CONCF-175

On submit key press the searchInput value is checked. If value is empty, event is prevented.

### DIFF
--- a/extensions/wikia/GlobalNavigation/scripts/GlobalNavigationSearch.js
+++ b/extensions/wikia/GlobalNavigation/scripts/GlobalNavigationSearch.js
@@ -23,11 +23,11 @@ require(['jquery', 'wikia.browserDetect', 'GlobalNavigationiOSScrollFix'], funct
 		}
 
 		//Disables search button until the user has entered at least one character.
-		$searchInput.keyup(function () {
+		$searchInput.keydown(function () {
 			if ($searchInput.val()) {
-				$searchSubmit.removeAttr('disabled');
+				$searchSubmit.prop('disabled', false);
 			} else {
-				$searchSubmit.attr('disabled', true);
+				$searchSubmit.prop('disabled', 'disabled');
 			}
 		})
 	}
--- a/extensions/wikia/GlobalNavigation/scripts/GlobalNavigationSearch.js
+++ b/extensions/wikia/GlobalNavigation/scripts/GlobalNavigationSearch.js
@@ -2,6 +2,7 @@ require(['jquery', 'wikia.browserDetect', 'GlobalNavigationiOSScrollFix'], funct
 	'use strict';
 	var $selectElement = $('#searchSelect'),
 		$searchInput = $('#searchInput'),
+		$searchSubmit = $('#searchSubmit'),
 		$autocompleteObj;
 
 	/**
@@ -10,6 +11,7 @@ require(['jquery', 'wikia.browserDetect', 'GlobalNavigationiOSScrollFix'], funct
 	function setFormOptions() {
 		var $selectedOption = $selectElement.find('option:selected'),
 			isLocalSearchDisabled = !$selectElement.length;
+
 		$searchInput.attr('placeholder', $selectedOption.data('placeholder'));
 		$('#searchForm').attr('action', $selectedOption.attr('data-search-url'));
 		//Setting reference to jQuery search autocomplete object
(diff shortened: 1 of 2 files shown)


extensions/wikia/GlobalNavigation/scripts/GlobalNavigationSearch.js
```diff
@@ -20,6 +21,15 @@ require(['jquery', 'wikia.browserDetect', 'GlobalNavigationiOSScrollFix'], funct
 		} else {
 			setPropertiesOnInput(true);
 		}
+
+		//Disables search button until the user has entered at least one character.
+		$searchInput.keyup(function () {
+			if ($searchInput.val()) {
+				$searchSubmit.removeAttr('disabled');
```
Warkot: Have thought of using `prop()`? http://stackoverflow.com/a/13626565/3370036

Warkot:
I was able to go around it. How? Write something, then delete it by pressing `delete`, hold it and click on submit or press enter. *Et voila!* Two ways out of it I see are:
* swap `keyup` for `keydown` or `keypress`,
* add an `onsubmit` check and if there's no input, `return false` or `e.preventDefault` and `e.stopPropagation`.

Both can be added, but I'm more into the latter. :shipit:

extensions/wikia/GlobalNavigation/scripts/GlobalNavigationSearch.js
Warkot: [Boolean attributes](http://www.w3.org/TR/html4/intro/sgmltut.html#h-3.3.4.2) don't have `true`/`false` value, in their full form they go like `disabled="disabled"`.

extensions/wikia/GlobalNavigation/scripts/GlobalNavigationSearch.js
```diff
@@ -20,6 +21,15 @@ require(['jquery', 'wikia.browserDetect', 'GlobalNavigationiOSScrollFix'], funct
 		} else {
 			setPropertiesOnInput(true);
 		}
+
+		//Disables search button until the user has entered at least one character.
+		$searchInput.keydown(function () {
```
bognix: Move this to function which is triggered after DOM is loaded. On Special Pages all scripts are loaded in <head> and it may introduce race conditions.

extensions/wikia/GlobalNavigation/templates/GlobalNavigation_searchIndex.php
```diff
@@ -14,7 +14,7 @@
 				</div>
 			</div>
 			<input id="searchInput" accesskey="f" autocomplete="off" class="search-input" name="search" placeholder="<?= Sanitizer::encodeAttribute($defaultSearchPlaceholder); ?>" type="text" value="<?= Sanitizer::encodeAttribute( $query ); ?>"/>
-			<input type="submit" id="searchSubmit" class="search-button" value="" />
+			<input type="submit" id="searchSubmit" class="search-button" value="" disabled/>
```
bognix: If you perform a search in global navigation you are redirected to Special:Search and search input is populated with search query. With this approach button would be disabled even though value in input is set.

extensions/wikia/GlobalNavigation/scripts/GlobalNavigationSearch.js
```diff
@@ -40,6 +41,14 @@ require(['jquery', 'wikia.browserDetect', 'GlobalNavigationiOSScrollFix'], funct
 	$(function () {
 		var $globalNav = $('#globalNavigation');
 
+		//Checks whether the searchInput is empty and prevents any action if true
+		$searchForm.submit(function (e) {
+			if (!$searchInput.val()) {
+				e.preventDefault();
+				e.stopPropagation();
```
bognix: you can do "return false" instead of calling preventDefault() and stopPropagation()

extensions/wikia/GlobalNavigation/scripts/GlobalNavigationSearch.js
```diff
@@ -2,6 +2,7 @@ require(['jquery', 'wikia.browserDetect', 'GlobalNavigationiOSScrollFix'], funct
 	'use strict';
 	var $selectElement = $('#searchSelect'),
```
bognix: please move assignment to those variables to function which is called after DOM loaded in order to avoid race conditions.

hakubo:
LGTM

PR #6498, opened 2015-03-02
bucky.resourceTiming: use weppy.store() instead of send()

macbre:
Use `weppy.store` that is a replacement for `bucky.send`.

@wladekb


wladekb:
lgtm, I think it could go to production outside of the normal weekly release cycle

macbre:
Most definitely :)

macbre:
Backported to production in 7f5fee073450384f181862e2257a963e2b839fb1

PR #6500, opened 2015-03-02
CE-1381: Ensure that user id exists

lukaszkonieczny:
https://wikia-inc.atlassian.net/browse/CE-1381

### DIFF
--- a/includes/wikia/services/WikiService.class.php
+++ b/includes/wikia/services/WikiService.class.php
@@ -480,7 +480,12 @@ class WikiService extends WikiaModel {
 	public function getMostActiveAdmins($wikiId, $avatarSize) {
 		$edits = $ids = $lastRevision = [];
 		$admins = $this->getWikiAdmins($wikiId, $avatarSize);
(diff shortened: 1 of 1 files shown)


includes/wikia/services/WikiService.class.php
```diff
@@ -480,7 +480,12 @@ public function getMostLinkedPages( $wikiId = 0 ) {
 	public function getMostActiveAdmins($wikiId, $avatarSize) {
 		$edits = $ids = $lastRevision = [];
 		$admins = $this->getWikiAdmins($wikiId, $avatarSize);
-		$ids = array_map(function( Array $item ) { return $item['userId']; }, $admins);
+
+		foreach ( $admins as $admin ) {
+			if ( isset( $admin['userId'] ) ) {
+				$ids[] = $admin['userId'];
+			}
+		}
```
macbre: LGTM, should this check be moved to `getWikiAdmins` method instead?

lukaszkonieczny: After conversation, we leave this check as it is.

macbre:
Thanks! Please backport this fix to the preview branch.

PR #6501, opened 2015-03-02
CONCF-4 | any wiki can be added to the Corporate Footer dropdown (wgDisableWAMOnHubPages is irrelevant)

Warkot:
Added an independent wg variable (`wgIncludeWikiInCorporateFooterDropdown`) to determine if a wiki is supposed to act as a corporate wiki, thus added to the Corporate Footer (the one on Wikia Home Pages) dropdown (the one with flags). Still only one wiki per language is permitted as they would overwrite each other.

Previously `wgDisableWAMOnHubPages` was determining if a page is a corporate page on hub template. Now WAM can be disabled on any page, and any page can be set as corporate.

### DIFF
--- a/includes/wikia/DefaultSettings.php
+++ b/includes/wikia/DefaultSettings.php
@@ -1676,6 +1676,11 @@ $wgProfilerSendViaScribe = true;
  */
 $wgDisableWAMOnHubs = false;
 
+/* @name wgTreatAsCorporate
+ * Treat a wiki as if it was a corporate wiki
+ */
+$wgTreatAsCorporate = false;
+
 /**
  * Force ImageServing to return an empty list
  * see PLATFORM-392
--- a/includes/wikia/DefaultSettings.php
+++ b/includes/wikia/DefaultSettings.php
@@ -1676,10 +1676,10 @@ $wgProfilerSendViaScribe = true;
  */
 $wgDisableWAMOnHubs = false;
 
-/* @name wgTreatAsCorporate
- * Treat a wiki as if it was a corporate wiki
+/* @name wgIncludeWikiInCorporateFooterDropdown
+ * Include link to this wiki in the Corporate Footer dropdown (the one with flags).
  */
-$wgTreatAsCorporate = false;
+$wgIncludeWikiInCorporateFooterDropdown = false;
 
 /**
  * Force ImageServing to return an empty list
--- a/extensions/wikia/CityVisualization/helpers/WikiaHomePageHelper.class.php
+++ b/extensions/wikia/CityVisualization/helpers/WikiaHomePageHelper.class.php
@@ -17,7 +17,7 @@ class WikiaHomePageHelper extends WikiaModel {
 	const VIDEO_GAMES_SLOTS_VAR_NAME = 'wgWikiaHomePageVideoGamesSlots';
 	const ENTERTAINMENT_SLOTS_VAR_NAME = 'wgWikiaHomePageEntertainmentSlots';
 	const LIFESTYLE_SLOTS_VAR_NAME = 'wgWikiaHomePageLifestyleSlots';
-	const CORPORATE_ON_HUB_ENABLED = 'wgTreatAsCorporate';
+	const IS_WIKI_INCLUDED_IN_CORPORATE_FOOTER_DROPDOWN_VAR_NAME = 'wgIncludeWikiInCorporateFooterDropdown';
 	const SLOTS_IN_TOTAL = 17;
 
 	const SLOTS_BIG = 2;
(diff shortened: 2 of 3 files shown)


includes/wikia/DefaultSettings.php
```diff
@@ -1676,6 +1676,11 @@
  */
 $wgDisableWAMOnHubs = false;
 
+/* @name wgTreatAsCorporate
+ * Treat a wiki as if it was a corporate wiki
+ */
+$wgTreatAsCorporate = false;
```
bognix: By adding another value you introduce inconsistency with https://github.com/Wikia/config/blob/dev/CommonExtensions.php#L3064 . So please either rename the variable in config to more descriptive name or add your variable to config too.

Warkot: I'm just looking at the line you're quoting (pasted beneath) and I think I must make few things clear.
```
$wgEnableWikiaStatsExt = $wgEnableWikiaStatsExt || $wgEnableWikiaHomePageExt || ($wgEnableWikiaHubsV3Ext && $wgDisableWAMOnHubs);
```
I'd say that the new variable has nothing to do with the old ones. Actually my intent is to make those variables independent. You can set `$wgTreatAsCorporate`, so the page is added to the *Corporate Footer*. You can independently set `$wgDisableWAMOnHubs` on hubs (`$wgEnableWikiaHubsV3Ext = true`), so you can swap WAM module for *Wikia Stats* (`$wgEnableWikiaStatsExt`).

What do you think? @bognix @kvas-damian

kvas-damian: Lets step back for a moment. Our main problem is: which wikis should be displayed in language chooser in footer.
We already found out that putting there corporate pages is not enough, because we have pt-br hub like corporate page. That is why we added pseudohub there.
Now we see that not all pseudohubs should be displayed there. HudlerHub is the reason.

Maybe we should create completely separate logic to put wiki in language chooser? Separate variable and logic completely separated from existing variables. What do you think about it?

Warkot: Isn't what I did here? :neutral_face:
> Separate variable and logic completely separated from existing variables.

This separates logic of 'adding to the corporate dropdown' from 'having a Wikia Stats instead of WAM'. The new variable is checked only while including any other wiki than the ones with `$wgEnableWikiaStatsExt` in the *Corporate Footer* language chooser dropdown—the old variable is removed from this check. Imagine it's renamed from `$wgTreatAsCorporate` to `$wgDisplayInCorporateFooterDropdown` (or something similar). Are you following? Am I following you correctly? :sweat_smile:

kvas-damian: Not exactly, we're adding this list to coporate pages list: https://github.com/Wikia/app/blob/CONCF-4/extensions/wikia/CorporateFooter/CorporateFooterController.class.php#L12

Warkot: IMHO, it's still just semantics. This:
```
$corporateWikis = array_merge( $corporateWikis, $corporateHubWikis );
```
can go like this:
```
$wikisToShowInTheCorporateFooterLanguageDropdown = array_merge( $corporateWikis, $wikisWithTheSpecificVariableEnabledToBeIncluded );
```

kvas-damian: So what I suggested initialy was to have variable that is the only source of truth. Not merging that with any other list.

Warkot: So that implies that you could create a *Wikia Home (//Corp) Page* (or at least a wiki that looks like one) and it doesn't need to be added to the *Corporate Footer* dropdown. Am I right? And this idea actually appeals to me. :+1:

kvas-damian: Exactly. On the oposite edge, you can create regular wiki which will be displayed there.

Warkot: :+1: It is even possible with this implementation. But should there still be only possible one for each language? (My answer: yes.)

Warkot: OK guys, I had today another idea. What if we added in the Corporate Footer dropdown those wikis which have this footer enabled (`$wgEnableCorporateFooterExt`)? It's strange that wiki can be included in the dropdown on other pages, but actually it doesn't have this dropdown itself. What do you think? It would also lead to enabling Corporate Footer on pt-br hub/corp page.

Warkot: And I was also thinking about resolving [this TODO](https://github.com/Wikia/app/blob/dev/extensions/wikia/GlobalFooter/scripts/GlobalFooter.js#L43).

kvas-damian: Your option is also valid. You need to decide if this limitation (you have to display lang dropdown if you want to put wiki there) won't be painful in future.

bognix:
LGTM

hakubo:
LGTM

tomasznapieralski:
LGTM

PR #6505, opened 2015-03-02
Execute non-internal tasks over http

nmonterroso:
@owend 
Currently, tasks created on sandbox, preview, and verify execute over ssh. This was done because of complications with forwarding something like `tasks.sandbox-s3.wikia.com` to the right place. Now, though, we're seeing some errors when executing these tasks over ssh, so we should move these to execute over http. Intentionally not updating `WIKIA_ENV_INTERNAL` because we can't hit `proxy.php` the same way we do in other environments.


owend:
Yeah, this looks good to me (based on offline conversation about a couple of things).

PR #6508, opened 2015-03-03
SOC-377 Additional checks to keep this from dying, including:

garthwebb:
- Check for revision before calling getTimestamp
- Check for title before calling getArticleID

Plus:
- Removed unused functions/variables
- Additional type hinting
- Fix method header comments
- Updated wfMsg
- Changed globals to F::app()->wg
- Adhering to style guide
- Changed array() to []

Jira: https://wikia-inc.atlassian.net/browse/SOC-377

### DIFF
--- a/extensions/wikia/PartnerFeed/PartnerFeed.body.php
+++ b/extensions/wikia/PartnerFeed/PartnerFeed.body.php
@@ -130,6 +130,7 @@ class PartnerFeed extends SpecialPage {
 						/** @var Revision $revision */
 						$firstRevision = $levelOne->mFirstRevision;
 						if ( !$firstRevision instanceof Revision ) {
+							$tmpArticleComment['mFirstRevision'] = $firstRevision;
 							$this->logError( 'Comment has no first revision', $tmpArticleComment );
 							break;
 						}
--- a/extensions/wikia/PartnerFeed/PartnerFeed.body.php
+++ b/extensions/wikia/PartnerFeed/PartnerFeed.body.php
@@ -8,6 +8,8 @@
  * Returns formated RSS / Atom Feeds
  */
 
+use Wikia\Logger\WikiaLogger;
+
 class PartnerFeed extends SpecialPage {
 	var $mName, $mPassword, $mRetype, $mReturnto, $mCookieCheck, $mPosted;
 	var $mAction, $mCreateaccount, $mCreateaccountMail, $mMailmypassword;
@@ -15,37 +17,31 @@ class PartnerFeed extends SpecialPage {
 	var $err;
 
 	function  __construct() {
-		parent::__construct( "PartnerFeed" , '' /*restriction*/);
-	}
-	private function shortenString($sString, $length){
-		return substr($sString,0,$length);
+		parent::__construct( "PartnerFeed" , '' /*restriction*/ );
 	}
 
 	function execute() {
-		global $wgLang, $wgRequest, $wgOut;
+		$req = $this->getRequest();
 
-		$this->mName = $wgRequest->getText( 'wpName' );
-		$this->mRealName = $wgRequest->getText( 'wpContactRealName' );
-		$this->mWhichWiki = $wgRequest->getText( 'wpContactWikiName' );
-		$this->mProblem = $wgRequest->getText( 'wpContactProblem' );
-		$this->mProblemDesc = $wgRequest->getText( 'wpContactProblemDesc' );
-		$this->mPosted = $wgRequest->wasPosted();
-		$this->mAction = $wgRequest->getVal( 'action' );
-		$this->mEmail = $wgRequest->getText( 'wpEmail' );
-		$this->mBrowser = $wgRequest->getText( 'wpBrowser' );
-		$this->mCCme = $wgRequest->getCheck( 'wgCC' );
+		$this->mName = $req->getText( 'wpName' );
+		$this->mRealName = $req->getText( 'wpContactRealName' );
+		$this->mWhichWiki = $req->getText( 'wpContactWikiName' );
+		$this->mProblem = $req->getText( 'wpContactProblem' );
+		$this->mProblemDesc = $req->getText( 'wpContactProblemDesc' );
+		$this->mPosted = $req->wasPosted();
+		$this->mAction = $req->getVal( 'action' );
+		$this->mEmail = $req->getText( 'wpEmail' );
+		$this->mBrowser = $req->getText( 'wpBrowser' );
+		$this->mCCme = $req->getCheck( 'wgCC' );
 
-		$feed = $wgRequest->getText( "feed", false );
-		$feedType = $wgRequest->getText ( "type", false );
-		if (	$feed
-			&& $feedType
-			&& in_array( $feed, array( "rss", "atom" ) )
-		) {
+		$feed = $req->getText( "feed", false );
+		$feedType = $req->getText ( "type", false );
+		if ( $feed && $feedType && in_array( $feed, [ "rss", "atom" ] ) ) {
 			// Varnish cache controll. Cache max for 12h.
-			header( "Cache-Control: s-maxage=".( 60*60*12 ) );
-			header( "X-Pass-Cache-Control: max-age=".( 60*60*12 ) );
+			header( "Cache-Control: s-maxage=" . ( 60 * 60 * 12 ) );
+			header( "X-Pass-Cache-Control: max-age=" . ( 60 * 60 * 12 ) );
 			$isFeed = true;
-			switch( $feedType ){
+			switch( $feedType ) {
 				case 'AchivementsLeaderboard':
 					$this->FeedAchivementsLeaderboard( $feed );
 				break;
@@ -56,7 +52,7 @@ class PartnerFeed extends SpecialPage {
 					$this->FeedRecentBadges( $feed );
 				break;
 				case 'HotContent':
-					$this->FeedHotContent( $feed, true );
+					$this->FeedHotContent( $feed );
 				break;
 				case 'RecentBlogPosts':
 					$this->FeedRecentBlogPosts( $feed );
@@ -73,7 +69,7 @@ class PartnerFeed extends SpecialPage {
 				break;
 			}
 			if ( $isFeed ) {
-				header("Content-Type: application/rss+xml");
+				header( "Content-Type: application/rss+xml" );
 			}
 		} else {
 			$this->showMenu();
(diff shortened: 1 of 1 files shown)


extensions/wikia/PartnerFeed/PartnerFeed.body.php
```diff
@@ -15,37 +17,31 @@ class PartnerFeed extends SpecialPage {
 	var $err;
 
 	function  __construct() {
-		parent::__construct( "PartnerFeed" , '' /*restriction*/);
-	}
-	private function shortenString($sString, $length){
```
garthwebb: This function was unused.

extensions/wikia/PartnerFeed/PartnerFeed.body.php
```diff
@@ -56,7 +52,7 @@ function execute() {
 					$this->FeedRecentBadges( $feed );
 				break;
 				case 'HotContent':
-					$this->FeedHotContent( $feed, true );
+					$this->FeedHotContent( $feed );
```
garthwebb: This arg wasn't being used.

extensions/wikia/PartnerFeed/PartnerFeed.body.php
```diff
@@ -81,100 +77,105 @@ function execute() {
...
 			// local settings
 			$maxNumberOfBlogComments = 10;
 			$userAvatarSize = 48;
 
-			$sBlogPost = $wgRequest->getText ( "blogpost", false );
-			$oTitle = Title::newFromText( $sBlogPost , 500);
-			if ( $oTitle->getArticleID() > 0 ){
+			$sBlogPost = $wg->Request->getText ( "blogpost", false );
+			$oTitle = Title::newFromText( $sBlogPost , 500 );
+			if ( $oTitle && $oTitle->getArticleID() > 0 ) {
```
garthwebb: Check for $oTitle to be present.

extensions/wikia/PartnerFeed/PartnerFeed.body.php
```diff
@@ -81,100 +77,105 @@ function execute() {
...
-						$tmpArticleComment = $commentPage[ 'level1' ]->getData();
 
-						$feedArray[ ] = array(
+					/** @var ArticleComment $levelOne */
+					$levelOne = $commentPage['level1'];
+
+					// make sure all data is loaded
+					if ( $levelOne->load() ) {
+						$tmpArticleComment = $levelOne->getData();
+
```
garthwebb: Below checks to make sure we get a revision.  If we don't get one, we log it in case there is something interesting about this situation.

extensions/wikia/PartnerFeed/PartnerFeed.body.php
```diff
@@ -230,152 +231,151 @@ private function FeedRecentBlogPosts ( $format ){
...
-				),
+				],
 				false,
-				array(
+				[
 					"ORDER BY" => "img_timestamp DESC",
 					"LIMIT" => $maxImagesNumber
-				)
+				]
 		);
```
garthwebb: $thumbSize is never used.

extensions/wikia/PartnerFeed/PartnerFeed.body.php
```diff
@@ -386,158 +386,66 @@ private function FeedRecentImages ( $format ){
...
 	}
 
-/**
- * @author Jakub Kurcek
- * @param hubId integer
- * @param forceRefresh boolean - if true clears the cache and creates new one.
- *
- * Returns data for feed creation. If no cache - creates one.
- */
-	private function PrepareHotContentFeed ( $hubId, $forceRefresh = false ){
```
garthwebb: This method is not used.

extensions/wikia/PartnerFeed/PartnerFeed.body.php
```diff
@@ -386,158 +386,66 @@ private function FeedRecentImages ( $format ){
...
-		# Class WikiaStatsAutoHubsConsumer was removed, so this method should be reimplemented
-		return array();
+	/**
+	 * @author Jakub Kurcek
+	 *
+	 * Returns array of accepted hubs.
+	 */
+	public function allowedHubs () {
+		return F::app()->wg->HubsPages['en'];
 	}
```
garthwebb: None of the private methods below were ever used.

jsutterfield:
LGTM!

extensions/wikia/PartnerFeed/PartnerFeed.body.php
```diff
@@ -81,100 +77,105 @@ function execute() {
...
+					/** @var ArticleComment $levelOne */
+					$levelOne = $commentPage['level1'];
+
+					// make sure all data is loaded
+					if ( $levelOne->load() ) {
+						$tmpArticleComment = $levelOne->getData();
+
+						/** @var Revision $revision */
+						$firstRevision = $levelOne->mFirstRevision;
+						if ( !$firstRevision instanceof Revision ) {
```
lizlux: Should we do this checking inside ArticleComment::load? We're not supposed to reach this point if L127 returns false, so maybe we should check to make sure that function doesn't return true if this condition isn't met. Could save some debugging time later.

extensions/wikia/PartnerFeed/PartnerFeed.body.php
```diff
@@ -81,100 +77,105 @@ function execute() {
...
-						\Wikia\Logger\WikiaLogger::instance()->error('Comment page loading failed', $commentPage);
+						$this->logError( 'Comment page loading failed', $commentPage );
 					}
 				}
 
-				$this->showFeed( $format , wfMsg( 'feed-title-blogcomments', $oTitle->getFullText() ),  $feedArray);
-
+				$this->showFeed(
+					$format,
+					wfMessage( 'feed-title-blogcomments', $oTitle->getFullText() )->text(),
```
lizlux: escaped?

extensions/wikia/PartnerFeed/PartnerFeed.body.php
```diff
@@ -182,46 +183,46 @@ private function FeedRecentChanges ( $format ){
...
+				'otherTags' => [
 					'image' => AvatarService::getAvatarUrl( $val['user'], $userAvatarSize )
-				)
-			);
+				]
+			];
 		}
 
-		$this->showFeed( $format , wfMsg('feed-title-recentchanges'), $feedArray);
+		$this->showFeed( $format, wfMessage( 'feed-title-recentchanges' )->text(), $feedArray );
```
lizlux: escaped?

extensions/wikia/PartnerFeed/PartnerFeed.body.php
```diff
@@ -230,152 +231,151 @@ private function FeedRecentBlogPosts ( $format ){
...
 			$oBlogListing = new CreateBlogListingPage;
-			$oBlogListing->setFormData('listingAuthors', '');
-			$oBlogListing->setFormData('tagContent', '');
-			if ( !empty( $sListing ) ){
+			$oBlogListing->setFormData( 'listingAuthors', '' );
+			$oBlogListing->setFormData( 'tagContent', '' );
+			if ( !empty( $sListing ) ) {
 				$oBlogListing->parseTag( urldecode( $sListing ) );
-				$subTitleName = wfMsg('blog-posts-from-listing', $sListing);
+				$subTitleName = wfMessage( 'blog-posts-from-listing', $sListing )->text();
```
lizlux: escaped?

extensions/wikia/PartnerFeed/PartnerFeed.body.php
```diff
@@ -230,152 +231,151 @@ private function FeedRecentBlogPosts ( $format ){
...
 			} else {
-				$oBlogListing->setFormData('listingCategories', '');
-				$subTitleName = wfMsg('all-blog-posts');
+				$oBlogListing->setFormData( 'listingCategories', '' );
+				$subTitleName = wfMessage( 'all-blog-posts' )->text();
```
lizlux: escaped?

extensions/wikia/PartnerFeed/PartnerFeed.body.php
```diff
@@ -230,152 +231,151 @@ private function FeedRecentBlogPosts ( $format ){
...
-				);
+					'otherTags' => [
+						'image' => AvatarService::getAvatarUrl( $val['username'], $userAvatarSize )
+					]
+				];
 			}
-			$this->showFeed( $format , wfMsg('feed-title-blogposts').' - '.$subTitleName, $feedArray);
+			$this->showFeed(
+				$format,
+				wfMessage( 'feed-title-blogposts' )->text() . ' - ' . $subTitleName,
```
lizlux: escaped?

extensions/wikia/PartnerFeed/PartnerFeed.body.php
```diff
@@ -230,152 +231,151 @@ private function FeedRecentBlogPosts ( $format ){
...
+					]
+				];
 
-				if ( $count++ >= $maxBadgesToDisplay ){
+				if ( $count++ >= $maxBadgesToDisplay ) {
 					break;
 				}
 			}
-			$this->showFeed( $format , wfMsg('feed-title-recent-badges'),  $feedArray);
+			$this->showFeed( $format, wfMessage( 'feed-title-recent-badges' )->text(), $feedArray );
```
lizlux: escaped?

extensions/wikia/PartnerFeed/PartnerFeed.body.php
```diff
@@ -386,158 +386,66 @@ private function FeedRecentImages ( $format ){
...
+				'otherTags' => [
 						'image' => $testImage
-					)
-			);
+				]
+			];
 		}
 
-		$this->showFeed( $format , wfMsg( 'feed-title-recent-images' ),  $feedArray);
+		$this->showFeed( $format, wfMessage( 'feed-title-recent-images' )->text(), $feedArray );
```
lizlux: escaped?

extensions/wikia/PartnerFeed/PartnerFeed.body.php
```diff
@@ -386,158 +386,66 @@ private function FeedRecentImages ( $format ){
...
 		$allowedHubs = $this->allowedHubs();
 
-		if ( isset( $allowedHubs[ $hubTitle ] ) && !is_array( $allowedHubs[ $hubTitle ] ) ){
+		if ( isset( $allowedHubs[ $hubTitle ] ) && !is_array( $allowedHubs[ $hubTitle ] ) ) {
 			$oTitle = Title::newFromText( $hubTitle, 150 );
 		} else {
 			$oTitle = Title::newFromText( $defaultHubTitle, 150 );
 		}
-		$this->showFeed( $format, wfMsg( 'feed-title-hot-content', $oTitle->getText() ), array() );
+		$this->showFeed( $format, wfMessage( 'feed-title-hot-content', $oTitle->getText() )->text(), [] );
```
lizlux: escaped?

extensions/wikia/PartnerFeed/PartnerFeed.body.php
```diff
@@ -573,40 +481,48 @@ private function FeedAchivementsLeaderboard ( $format ) {
...
-						'media:thumbnail' =>AvatarService::getAvatarUrl( $rankedUser->getName(), $userAvatarSize ),
-					)
-				);
+					'otherTags' => [
+						'media:thumbnail' => AvatarService::getAvatarUrl( $rankedUser->getName(), $userAvatarSize ),
+					]
+				];
 			}
-			$this->showFeed( $format , wfMsg('feed-title-leaderboard'),  $feedArray);
+			$this->showFeed( $format, wfMessage( 'feed-title-leaderboard' )->text(), $feedArray );
```
lizlux: escaped?

extensions/wikia/PartnerFeed/PartnerFeed.body.php
```diff
@@ -573,40 +481,48 @@ private function FeedAchivementsLeaderboard ( $format ) {
...
 	private function showFeed( $format, $subtitle, $feedData ) {
-		global $wgOut, $wgRequest, $wgParser, $wgMemc, $wgTitle, $wgSitename;
-
 		wfProfileIn( __METHOD__ );
 		$sFeedName = self::getFeedClass( $format );
-		$feed = new $sFeedName( wfMsg('feed-main-title'),  $subtitle, $wgTitle->getFullUrl() );
+
+		/** @var ChannelFeed $feed */
+		$feed = new $sFeedName(
+			wfMessage( 'feed-main-title' )->text(),
```
lizlux: escaped?

extensions/wikia/PartnerFeed/PartnerFeed.body.php
jsutterfield: My own thinking was that since this is an RSS feed, rather than being inserted into HTML it would be fine. That could be incorrect though.

extensions/wikia/PartnerFeed/PartnerFeed.body.php
garthwebb: There are actually so many checks in ```load``` for ```mFirstRevId``` and ```mFirstRevision``` that I don't know how this is happening.  Given this, I'd rather not make a global change and risk breaking something that relies on this undefined behavior, and instead catch the error I know is occurring.  I noticed that mFirstRevision isn't returned by the ```getData``` call however, so I'm going to drop that into the debug statement as well.

extensions/wikia/PartnerFeed/PartnerFeed.body.php
garthwebb: Yah, RSS and Atom feeds use CDATA blocks to escape content which means we don't need to escape these for transport.  The lines that include user content (like this one) are ok because the content is all titles which are already sanitized for special characters.

PR #6510, opened 2015-03-03
HubsMenu content columns wrap

vforge:
https://wikia-inc.atlassian.net/browse/CONCF-149

### DIFF
--- a/extensions/wikia/GlobalNavigation/templates/GlobalNavigation_hubsMenuSections.php
+++ b/extensions/wikia/GlobalNavigation/templates/GlobalNavigation_hubsMenuSections.php
@@ -3,16 +3,16 @@
 		<div class="hub-menu-columns">
 			<? if ( array_key_exists( 'children', $sections ) ) : ?>
 				<? foreach( $sections['children'] as $category ): ?>
-					<strong><?= $category['text'] ?></strong>
-					<? if ( array_key_exists( 'children', $category ) ) : ?>
-						<ul>
+					<ul>
+						<li><strong><?= $category['text'] ?></strong></li>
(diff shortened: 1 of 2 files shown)


extensions/wikia/GlobalNavigation/templates/GlobalNavigation_hubsMenuSections.php
```diff
@@ -3,16 +3,16 @@
 		<div class="hub-menu-columns">
 			<? if ( array_key_exists( 'children', $sections ) ) : ?>
 				<? foreach( $sections['children'] as $category ): ?>
-					<strong><?= $category['text'] ?></strong>
-					<? if ( array_key_exists( 'children', $category ) ) : ?>
-						<ul>
+					<ul>
+						<li><strong><?= $category['text'] ?></strong></li>
```
Warkot: I'm just wondering if adding a wrapper wouldn't be more semantically correct. Now the higher level list item/header becomes the same level as subitems. What do you think about using `<dl>`/`<dt>`/`<dd>`? It's still not perfectly expressing the semantics, but might be a way out.

vforge: It was like this before, I'm just partially reverting change that was implemented. I don't want to complicate things at this point. It would probably make more sense to replace UL with DIV or SECTION tags (as this is one section inside menu), but it's too complicated for quick fix.

Warkot: Sure, it's not a big deal, it still LGTM.

hakubo:
LGTM

PR #6511, opened 2015-03-03
CE-1381: Ensure that user id exists

lukaszkonieczny:
https://wikia-inc.atlassian.net/browse/CE-1381


macbre:
LGTM (as in #6500 :)

PR #6523, opened 2015-03-03
ADEN-1770 Amazon 2.0 params can overload gpt

gabrys:


hakubo:
LGTM

extensions/wikia/AdEngine/js/AmazonMatch.js
```diff
@@ -96,16 +115,34 @@ define('ext.wikia.adEngine.amazonMatch', [
...
-		};
+	function getSlotParams(slotName) {
+		log(['getSlotParams'], 'debug', logGroup);
+
+		var ret = {};
+
+		// No Object.keys on IE8
+		if (!Object.keys) {
+			log(['filterSlots()', 'no Object.keys (IE8?)'], 'error', logGroup);
+			return {};
```
nandy-andy: What do you think about having just one `return` in function? Here it'd just require to change `if` body with what's below, add `else` and only log the error there - the value of `ret` will be returned and it will be empty in case of IE8.

gabrys: Early returns are good, they tell better stories :-)

PR #6528, opened 2015-03-03
Soc 316 validating passwords browser side

lizlux:
This PR circumvents password validation so it stays in the browser as opposed to being sent over HTTP in plain text. 

I also did some code cleanup.

### DIFF
--- a/extensions/wikia/UserLogin/UserLogin.setup.php
+++ b/extensions/wikia/UserLogin/UserLogin.setup.php
@@ -51,6 +51,8 @@ JSMessages::registerPackage('UserLogin', ['userlogin-login-*']);
 JSMessages::registerPackage('UserSignup', [
 	'usersignup-page-captcha-load-fail-title',
 	'usersignup-page-captcha-load-fail-text',
+	'usersignup-error-password-length',
+	'userlogin-error-wrongpasswordempty',
 ]);
 
 // special pages
--- a/extensions/wikia/UserLogin/UserLogin.setup.php
+++ b/extensions/wikia/UserLogin/UserLogin.setup.php
@@ -61,7 +61,7 @@ JSMessages::registerPackage('UserSignup', [
  */
 $wgResourceModules['ext.userLogin'] = [
 	'localBasePath' => __DIR__ . '/scripts',
-	'remoteExtPath' => 'wikia/CookiePolicy/scripts',
+	'remoteExtPath' => 'wikia/UserLogin/js',
 	'messages' => [
 		'usersignup-error-password-length',
 		'userlogin-error-wrongpasswordempty',
--- a/extensions/wikia/UserLogin/UserLogin.setup.php
+++ b/extensions/wikia/UserLogin/UserLogin.setup.php
@@ -40,6 +40,9 @@ $wgHooks['SavePreferences'][] = 'UserLoginHooksHelper::onSavePreferences';
 $wgHooks['ConfirmEmailComplete'][] = 'UserLoginHooksHelper::onConfirmEmailComplete';
 $wgHooks['WikiaMobileAssetsPackages'][] = 'UserLoginHooksHelper::onWikiaMobileAssetsPackages';
 $wgHooks['AbortNewAccount'][] = 'UserLoginHooksHelper::onAbortNewAccount';
+// Add the JavaScript messages to the output
+$wgHooks['BeforePageDisplay'][] = "UserLoginHooksHelper::onBeforePageDisplay";
+
 
 // i18n mapping
 $wgExtensionMessagesFiles['UserLogin'] = $dir . 'UserLogin.i18n.php';
@@ -51,9 +54,20 @@ JSMessages::registerPackage('UserLogin', ['userlogin-login-*']);
 JSMessages::registerPackage('UserSignup', [
 	'usersignup-page-captcha-load-fail-title',
 	'usersignup-page-captcha-load-fail-text',
-	'usersignup-error-password-length',
-	'userlogin-error-wrongpasswordempty',
 ]);
+
+/**
+ * Use ResourceLoader to load the JavaScript module
+ */
+$wgResourceModules['ext.userLogin'] = [
+	'localBasePath' => __DIR__ . '/scripts',
+	'remoteExtPath' => 'wikia/CookiePolicy/scripts',
+	'messages' => [
+		'usersignup-error-password-length',
+		'userlogin-error-wrongpasswordempty',
+	],
+];
+
 
 // special pages
 $wgSpecialPages['Userlogin'] = 'UserLoginSpecialController';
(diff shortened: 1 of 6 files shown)


extensions/wikia/UserLogin/UserSignupSpecialController.class.php
```diff
@@ -607,18 +607,17 @@ public function formValidation() {
 
 		switch( $field ) {
 			case 'userloginext01' :
-				$response = $signupForm->initValidationUsername();
```
jsutterfield: High-five for getting ride of these unused `$response` variables.

jsutterfield:
It seems like a lot of work is being done to account for the username and password inputs having the names 'userloginext01' and 'userloginext02' on Special:UserSignup, but 'username' and 'password' in the FB Signup modal. What are you thoughts on just renaming those to 'userloginext01' and 'userloginext02' [here](https://github.com/Wikia/app/blob/dev/extensions/wikia/UserLogin/templates/FacebookSignup_modal.php#L8-L17)?

I know those pretty awful input names, but doing so should simplify the logic quite a bit. You could get rid of lines [here](https://github.com/Wikia/app/pull/6528/files#diff-3cf607c0b9d53efede246a5e56ec8188R83), [here](https://github.com/Wikia/app/pull/6528/files#diff-aa100088555a14e1c334d5ea2f3c2ec1R20), [here](https://github.com/Wikia/app/pull/6528/files#diff-aa100088555a14e1c334d5ea2f3c2ec1R32), and this if block [here](https://github.com/Wikia/app/pull/6528/files#diff-aa100088555a14e1c334d5ea2f3c2ec1R39). Also, those inputs were given those names from a signup spam attack, it may make sense to use that obfuscation defense for the new form FB Form as well just to be careful.

extensions/wikia/UserLogin/js/UserSignupAjaxValidation.js
```diff
@@ -9,12 +9,15 @@
 	 * - wikiaForm: instance of WikiaForm
 	 * - inputsToValide: array of input names to be ok'ed before submission
 	 * - submitButton: pointer to main submit button of the form
+	 * - passwordInputName: input name attribute for form's password input
 	 * @constructor
 	 */
 	var UserSignupAjaxValidation = function (options) {
 		this.wikiaForm = options.wikiaForm;
 		this.inputsToValidate = options.inputsToValidate || [];
```
jsutterfield: Hey, are these being used for anything?

extensions/wikia/UserLogin/js/UserSignupAjaxValidation.js
```diff
@@ -22,111 +25,109 @@
...
 	};
 
-	UserSignupAjaxValidation.prototype.checkFieldEmpty = function (field) {
-		return field && ((field.is('input') && field.val() === '') || (field.is('select') && field.val() === -1));
+	/**
+	 * Shared AJAX helper method
+	 * @param {object} data
+	 * @returns {jQuery} Promise
+	 */
+	UserSignupAjaxValidation.prototype.sendRequest = function (data) {
```
jsutterfield: How about giving this a slightly more descriptive name? Like `sendValidationRequest` or `validateInputOnServer`? Since we're hitting the same URL every time but just changing the inputs we want to validate.

lizlux:
@jsutterfield, in response to your comments about username and password field aliases, that's basically what I tried to do here https://github.com/Wikia/app/pull/6411. That was a very long and ultimately un-merged PR that tried to normalize field names and clean up this, until recently, unmaintained extension. 

In the end, I was unable to merge the PR b/c I ran into too many existing bugs that had to be handled before I could refactor and know that I wasn't introducing any new regressions. I shifted my solutions to the front end where I had more control over what was happening, even though I don't really like the end result, and agree with you that it could be better. 

In my last couple UserLogin-related tickets I've been slowly re-introducing some of the cleanup I did in that PR, so that work, along with your upcoming efforts of cleaning up this extension, might make it easier to implement the changes you're suggesting. In the mean time, I'd rather go with this somewhat sub-optimal code.

extensions/wikia/UserLogin/js/UserSignupAjaxValidation.js
lizlux: Oh yeah, I got rid of the functions that were using this array. Looks like we can get rid of that option too.

jsutterfield:
@lizlux, ah, looks like you were way ahead of me on this one :)

This solution works perfectly fine, sticking with it sounds good to me. I agree this extension was(/still kinda is) a mess. We can keep in mind trying to consolidate those input names down the line as we hack away at the tech debt, but this definitely works as an alternative until we get to that point.

extensions/wikia/UserLogin/UserLogin.setup.php
```diff
@@ -53,6 +56,19 @@
 	'usersignup-page-captcha-load-fail-text',
 ]);
 
+/**
+ * Use ResourceLoader to load the JavaScript module
+ */
+$wgResourceModules['ext.userLogin'] = [
+	'localBasePath' => __DIR__ . '/scripts',
+	'remoteExtPath' => 'wikia/CookiePolicy/scripts',
```
armonr: Why is this pointing to CookiePolicy?

extensions/wikia/UserLogin/js/UserSignupAjaxValidation.js
```diff
@@ -22,111 +23,109 @@
...
-	};
+	UserSignupAjaxValidation.prototype.validatePassword = function (password) {
+		var pwLength = password.length,
+			msg = '',
+			inputName = this.passwordInputName,
+			response = {};
+
+		// check password isn't too short (defaults to 1 char)
+		if (pwLength < wgMinimalPasswordLength) {
+			msg = mw.message('userlogin-error-wrongpasswordempty').escaped();
```
armonr: If the minimum password length to more than 1, user gets error saying "your password is empty" rather than "your password is too short". We probably need to change this message or make a new one. Either way, it requires translation :(

extensions/wikia/UserLogin/UserLogin.setup.php
lizlux: B/c of copy and paste :wink:. I'll fix.

extensions/wikia/UserLogin/js/UserSignupAjaxValidation.js
lizlux: Let's handle that in a separate ticket. I'll inform Cristina.

lizlux: Noted here https://wikia-inc.atlassian.net/browse/SOC-468

armonr: :+1:

lizlux:
Unit tests are passing on my devbox so I'm going to merge this.

PR #6530, opened 2015-03-04
SOC-375 PHP Fatal Error: Call to a member function doPurge() on a non-object RelatedForumDiscussionController.class.php

jsutterfield:
Add a check inside of the purgeCache method of the RelatedForumDiscussionController that we are operating on an object before calling doPurge() on it. Also added logging for when this fails, a dashboard to display those failures, and a README with a link to the dashboard.

Ticket: https://wikia-inc.atlassian.net/browse/SOC-375

### DIFF
--- a/extensions/wikia/Forum/RelatedForumDiscussionController.class.php
+++ b/extensions/wikia/Forum/RelatedForumDiscussionController.class.php
@@ -1,4 +1,7 @@
 <?php
+
+use Wikia\Logger\WikiaLogger;
+
 class RelatedForumDiscussionController extends WikiaController {
 
 	/** Expiry time to use in cache requests */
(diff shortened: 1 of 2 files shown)


extensions/wikia/Forum/RelatedForumDiscussionController.class.php
```diff
@@ -77,14 +80,19 @@ public function checkData() {
...
 
-		foreach($ids as $id) {
+		foreach( $ids as $id ) {
 			$key = wfMemcKey( __CLASS__, 'getData', $id );
-			WikiaDataAccess::cachePurge($key);
+			WikiaDataAccess::cachePurge( $key );
 			// VOLDEV-46: Update module by purging page, not via AJAX
-			WikiPage::newFromID( $id )->doPurge();
+			$wikiaPage = WikiPage::newFromID( $id );
+			if ( !empty( $wikiaPage ) ) {
```
armonr: Or simply `if ( $wikiaPage ) {`

extensions/wikia/Forum/RelatedForumDiscussionController.class.php
```diff
@@ -77,14 +80,19 @@ public function checkData() {
...
+				$wikiaPage->doPurge();
+			} else {
+				self::logError( "doPurge() attempted on non-object", [ "articleID" => $id ] );
```
armonr: "doPurge() attempted" might be misleading because it has not. Maybe something similar to "Found a null related wikipage on thread purge"?
I think threadId would also be very helpful in this log.

extensions/wikia/Forum/RelatedForumDiscussionController.class.php
```diff
@@ -101,4 +109,8 @@ public static function getData( $articleId ) {
 			return $messages;
 		});
 	}
+
+	private static function logError( $message, array $param = [] ) {
+			WikiaLogger::instance()->error( 'RelatedForumDiscussionController: ' . $message, $param );
```
armonr: white space

jsutterfield: white space?

armonr: Sorry, need to be more specific :) The indentation seems larger than what it should be

jsutterfield: Ah, now I see it :)

lizlux:
LGTM!

PR #6535, opened 2015-03-04
CreateWiki: improve profiling of cleanupStarter.sql

macbre:
``` php
function sourceFile( $filename, $lineCallback = false, $resultCallback = false, $fname = false )
```

Pass a `__METHOD__` to improve profiling and performance monitoring of `cleanupStarter.sql` import.

@michalroszka


michalroszka:
:+1:

PR #6537, opened 2015-03-04
Concf 145

bognix:


kvas-damian:
What do you think about adding unit tests for this piece of functionality?
It'd reguire some magic to mock static calls but IMO it could be valuable.

bognix:
I was thinking about adding those but not at this point because I still don't know if this is the way we are going to do this (and if we are going to do this).

includes/wikia/services/HubService.class.php
```diff
@@ -61,6 +81,31 @@ public static function getCurrentWikiaVerticalName() {
...
+	 * For Lifestyle and Gaming return their names.
+	 * For rest of values return Entertainment.
+	 * @param $cityId
+	 * @return Boolean|String
+	 */
+	public static function getVerticalNameForComscore($cityId) {
+		$verticalId = WikiFactoryHub::getInstance()->getVerticalId($cityId);
+		$canonicalVerticalId = self::getCanonicalVerticalID($verticalId);
+
+		switch ($canonicalVerticalId) {
```
owend: getVerticalId doesn't return a field which maps to the CATEGORY_ID

See here: https://github.com/Wikia/app/blob/dev/extensions/wikia/WikiFactory/Hubs/WikiFactoryHub.php#L22

And actually, I'm not sure if we have that mapping defined anywhere, but we should probably have one.  We need a mapping of vertical to "legacy category" for comscore, but I think we also need to fall back on the old category if the vertical_id is 0.  

I think we also need to double check the names, as they are different than what is defined in HubService. 

Also, there's a fallback in the getCategoryInfoForCity() function that looks for a comscore override tag, and I think we need to preserve that logic (line 102 of HubService)

bognix: I added the logic to fallback to LIFESTYLE if vertical is not set.
What do you mean by double checking the names?
I would like to simplify the logic as much as possible.
From what I understood from Amber's comment, we don't want to fallback to old category.
It is either GAMING (for gaming hub), ENTERTAINMENT (for everything beside gaming and lifestyle) or LIFESTYLE (for lifestyle or unset vertical).

includes/wikia/services/HubService.class.php
```diff
@@ -44,6 +44,27 @@ public static function getCanonicalCategoryId($categoryId) {
...
 	/**
+	 * Return varticalId.
+	 * For Lifestyle, Gaming and Corporate return passed id.
+	 * If vertical is not set - return Lifestyle.
+	 * For rest return Entertainment id.
+	 *
+	 * @param $verticalId
+	 * @return int
+	 */
+	public static function getCanonicalVerticalID($verticalId) {
```
hakubo: should this be private?

vforge:
Code LGTM (with stuff we've said about it being dated).

includes/wikia/services/HubService.class.php
```diff
@@ -61,6 +61,41 @@ public static function getCurrentWikiaVerticalName() {
...
+					return false;
+				} else {
+					return 'lifestyle';
+				}
+			default:
+				return 'entertainment';
+		}
+	}
+
+	private static function isWikiCorporate( $cityId ) {
```
owend: Having a generic utility function that can get the vertical name for any wiki is useful.  Perhaps that could be moved to the WikiFactoryHub class?  

It looks like this function is always called in the context of the current wiki (wgCityId) so you could also check the wgEnableWikiaHomePageExt variable directly instead of doing a query here, right?

kvas-damian: If there is no param needed they you could use this one: https://github.com/Wikia/app/blob/dev/includes/wikia/WikiaPageType.class.php#L255

BTW
```getVerticalNameForComscore``` won't be probably used in other places than Comscore, that's why I suggest moving it there. Logic is related to comscore only, strings are hardcoded and not standard ones. What do you think about it?

bognix: I agree with Damian. This function is not simply returning vertical for given wikia but applies some mapping specific for Comscore.

I renamed HUB to VERTICAL and used WikiaPageType method.

ping @kvas-damian  @owend

owend:
This LGTM :+1:

hakubo:
LGTM

PR #6540, opened 2015-03-04
Soc 351 Fixing ajaxy change password flows

lizlux:
The username and login token weren't being handled correctly with the change password flows that existed in a modal or as part of the create new wiki flow. The Special:UserLogin version was working fine. 

This PR makes sure that username, loginToken, and editToken are all where they should be for changing passwords via ajax. 

Also includes some code cleanup for this still fairly outdated and under-maintained extension.


extensions/wikia/UserLogin/js/UserLoginModal.js
```diff
@@ -82,6 +82,7 @@
 
 					UserLoginModal.loginAjaxForm = new window.UserLoginAjaxForm($loginModal, {
 						ajaxLogin: true,
+						modal: loginModal,
```
jsutterfield: Is this being used for something?

extensions/wikia/UserLogin/UserLoginSpecialController.class.php
```diff
@@ -602,18 +631,26 @@ public function changePassword() {
...
+		$returnto = $this->request->getVal( 'returnto', '' );
+
+		$this->response->setValues([
+			'username' => $username,
+			'password' => $password,
+			'newpassword' => $newPassword,
+			'retype' => $retype,
+			'editToken' => $this->wg->User->getEditToken(),
+			'loginToken' => $loginToken,
+			'returnto' => $returnto,
```
jsutterfield: How about adding line 631 to this `response->setValues` party?

armonr:
LGTM (beside comments from James and conflicts to be resolved.)

extensions/wikia/UserLogin/templates/UserLoginSpecial_changePassword.php
```diff
@@ -69,9 +63,9 @@
 		)
 	);
 
-	$form['isInvalid'] = !empty($result) && empty($errParam) && !empty($msg);
-	$form['errorMsg'] = !empty($msg) ? $msg : '';
+	$form['isInvalid'] = !empty( $result ) && $result == 'error';
```
armonr: `!empty()` check is superfluous

extensions/wikia/UserLogin/js/UserLoginAjaxForm.js
```diff
@@ -47,32 +47,63 @@
...
+	};
+
+	/**
+	 * Replace form's content based on HTML sent from the server.
+	 * @param {string} html
+	 */
+	UserLoginAjaxForm.prototype.setupChangePasswordForm = function (html) {
+		var content, form, heading, modal, contentBlock, duration;
+
+		content = $('<div>').hide().append(html);
```
armonr: Are `.append()` and `.html()` safe in terms of escaping?

lizlux: This should be safe since we're not getting the HTML from the DOM, where the user could have manipulated it prior to retrieval. We're getting the HTML from a server request, and the HTML is generated from templates where all the UCG has been escaped. So we should be fine here.

extensions/wikia/UserLogin/js/UserLoginModal.js
lizlux: Yeah it's being used for the reset password functionality when it's done inside a modal.

extensions/wikia/UserLogin/UserLoginSpecialController.class.php
lizlux: done, although I kept it where it was since it's `pageHeading` and it's grouped with `pageTitle`

jsutterfield:
LGTM

PR #6544, opened 2015-03-05
OPS-5563: New purging back end

owend:
- Added a generic Celery Task worker, teach SquidUpdate to use it
- Also renamed AsyncBackendTask to AsynNLPTask because it was not generic

@nmonterroso @artursitarski


nmonterroso:
LGTM

PR #6548, opened 2015-03-05
PLATFORM-981: Get rid of wfDBLightMode

macbre:
It's not being used often and is unreliable (features have to use it explicitly)

Additionally, it was making 2.5mm queries daily (that returns zero rows + no caching) - 22% of all queries hitting shared DB from task / cron machines.

``` sql
SELECT /* WikiFactory::getWikiByDB  */  *  FROM `city_list`  WHERE city_dbname = 'wikicities'  LIMIT 1
```

@owend / @wladekb / @drozdo


drozdo:
LGTM

wladekb:
lgtm

PR #6549, opened 2015-03-05
ADEN-1725 Interstitial as a custom ad

gabrys:


extensions/wikia/AdEngine/js/template/interstitial.js
```diff
@@ -0,0 +1,60 @@
...
+	'wikia.window'
+], function (doc, log, uiFactory, win) {
+	'use strict';
+
+	var logGroup = 'ext.wikia.adEngine.template.interstitial',
+		modalId = 'ext-wikia-adEngine-template-interstitial',
+		modalPadding = 20,
+		pollingInterval = 1000; // 1s
+
+	function show(params) {
```
lizlux: It would be nice to include a doc block with some of the required properties of `params`, like `slot` and `elem`.

extensions/wikia/AdEngine/js/template/interstitial.js
```diff
@@ -0,0 +1,60 @@
...
+			uiFactory.init('modal').then(function (uiModal) {
+				uiModal.createComponent(modalConfig, function (modal) {
+					var width = params.width || modal.$content[0].querySelector(selector).clientWidth;
+					modal.$element.width(width + 2 * modalPadding);
+					modal.show();
+				});
+			});
+		}
+
+		function waitForElement() {
```
lizlux: Have you considered using [`MutationObserver`](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) where available instead of polling? TBH, I haven't had a chance to use it myself but it might be a performance improvement. We'd still need the polling for IE<11.

extensions/wikia/AdEngine/js/template/interstitial.js
```diff
@@ -0,0 +1,60 @@
...
+		pollingInterval = 1000; // 1s
+
+	function show(params) {
+		log(['show', params], 'debug', logGroup);
+
+		var iframe = doc.getElementById(params.slot).querySelector('iframe'),
+			iframeBody = iframe.contentWindow.document.body,
+			selector = params.elem;
+
+		function onFound(element) {
```
lizlux: I'd prefer to define these functions outside of the `show` function so that they don't need to be redefined every time show is called. Looks like you could pass `params` to `onFound` or set `params` to a local variable scoped outside of the `show` function in order to pass that value around.

gabrys: Show will be called at max once per PV and the current approach gives me easy access to the context.

lizlux: Okay that sounds fine then.

lizlux:
LGTM after messages are updated and comments are addressed.

extensions/wikia/AdEngine/js/template/interstitial.js
gabrys: The element will be there in most of the cases, since the ad event fires after iframe's onload, so the polling is there only to support some edgy ads, where they load asynchronously.

I think the MutationObserver could come handy when updating the floating rail ad, so it reacts to stuff added to rail.

PR #6550, opened 2015-03-05
CE-1401: Fix automated edit summaries

lukaszkonieczny:
https://wikia-inc.atlassian.net/browse/CE-1401


extensions/wikia/EditPageLayout/EditPageLayout.class.php
```diff
@@ -448,7 +448,8 @@ public function renderSummaryBox() {
 			'name' => $this->mSummaryBox['name'],
 			'placeholder' => $this->mSummaryBox['placeholder'],
 			'tabindex' => '1', // BugId:5327
-		), $this->summary, false /* $allowShortTag */);
+			'value' =>  $this->summary
+		), '', false /* $allowShortTag */);
```
Grunny: Short tag shouldn't be disallowed on an `input` tag (at the moment it results in the output being `<input ...></input>` rather than just `<input ... />`. So, the last two parameters here can be removed to just use the default.

PR #6553, opened 2015-03-05
DatabaseMysqlBase: sampled log of mysql pings

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-949

Log (with 1% sampling) when MW app pings mysql server

@michalroszka / @drozdo


drozdo:
LGTM

PR #6401, opened 2015-02-23
PLATFORM-908 mysqli DB driver for MediaWiki

macbre:
As a part of the upgrade to PHP v5.6 we want to switch to `mysqli` extension to talk to our database (`mysql` extension is no longer maintained and became deprecated). MW 1.22 provides the support for both `mysql` & `mysqli` extensions. This PR is a set of cherry picks from [MediaWiki repository](https://github.com/wikimedia/mediawiki/tree/master/includes/db).

@Wikia/platform-team

PR #6564, opened 2015-03-06
Fixes after switching to mysqli

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-982

After moving from `mysql` to `mysqli` PHP extension (#6401) `mysql_*` functions do not work anymore. For instance [`mysql_real_escape_string`](http://pl1.php.net/manual/en/function.mysql-real-escape-string.php#refsect1-function.mysql-real-escape-string-notes) will now return false, because there's no active DB connection created by `mysql` extension. Use MW build-in escaping features instead of calling these functions directly.

@wladekb / @michalroszka / @bognix


wladekb:
lgtm

PR #6568, opened 2015-03-06
(CE-1411) Escape alt attribute for thumbnails

Grunny:
Escape alt attribute for thumbnails as it is user controlled and causes
an XSS vulnerability. The attribute used to be escaped in the controller
but that was removed while leaving it unescaped in the template (where the
escaping should occur).

/cc @bognix


bognix:
LGTM

PR #6570, opened 2015-03-06
WikiFactory: improve profiling of DB queries

macbre:
Add "submethod" to `$fname` argument when logging SQL queries

@michalroszka

### DIFF
--- a/extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
+++ b/extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
@@ -290,7 +290,7 @@ class WikiFactoryLoader {
 							"city_dbname"
 						),
 						($this->mCityID) ? array( "city_list.city_id" => $this->mCityID ) : array( "city_list.city_dbname" => $this->mCityDB ),
-						__METHOD__
+						__METHOD__ . '::domaindb'
 					);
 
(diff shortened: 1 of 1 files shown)


extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
```diff
@@ -290,7 +290,7 @@ public function execute() {
 							"city_dbname"
 						),
 						($this->mCityID) ? array( "city_list.city_id" => $this->mCityID ) : array( "city_list.city_dbname" => $this->mCityDB ),
-						__METHOD__
+						__METHOD__ . '::domaindb'
```
wladekb: I guess the "db" suffix doesn't hold any meaning here as well as in 2 places below. Can we get rid of it?

macbre: It simply matches the naming of profiler sections (see line 271)

wladekb: I'd rename them too :) but I'll let you make a final call.

macbre: I'd simply keep it as it is :)

PR #6571, opened 2015-03-06
(CE-1349) Remove double escaping on title attribute of thumbnails

Grunny:
The title attribute of thumbnails was moved to be escaped in the template
(where the escaping should happen) in:
https://github.com/Wikia/app/commit/c668ed3283063a#diff-2

But, since it was already being escaped in the helper, it was being double
escaped leading to the title displaying the HTML entities to users.


hakubo:
LGTM

PR #6572, opened 2015-03-06
CE-1397: Fix CreateNewWiki's layout in RTL languages

lukaszkonieczny:
https://wikia-inc.atlassian.net/browse/CE-1397

After CSSJanus update, this workaround breaks code so we can remove it


adamkarminski:
LGTM :+1:

PR #6575, opened 2015-03-06
(SOC-469) Only run this hook if Wall is enabled

Grunny:
Message Wall and Forum are intertwined right now so that if Forum is
enabled but Wall is disabled, then Message Wall's setup file and hooks
are still loaded, so we need this check in the hook method after all
as a quick fix...

/cc @jsutterfield @lizlux

### DIFF
--- a/extensions/wikia/Wall/WallHooksHelper.class.php
+++ b/extensions/wikia/Wall/WallHooksHelper.class.php
@@ -1,5 +1,12 @@
 <?php
 
+/**
+ * Hooks for Message Wall.
+ *
+ * All of these hooks will be executed when Forums are enabled
+ * even if Message Wall is disabled, so appropriate checks are
+ * needed if the hook should only take effect if Wall is enabled.
+ */
 class WallHooksHelper {
 	const RC_WALL_COMMENTS_MAX_LEN = 50;
 	const RC_WALL_SECURENAME_PREFIX = 'WallMessage_';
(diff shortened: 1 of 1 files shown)


armonr:
:shipit:

lizlux:
What a PITA! Thanks for fixing @Grunny

extensions/wikia/Wall/WallHooksHelper.class.php
```diff
@@ -2218,7 +2218,11 @@ public static function onArticleCommentGetSquidURLs( $title, &$urls ) {
 	 * @return bool
 	 */
 	public static function onGetTalkPage( Title $title, Title &$talkPageTitle ) {
-		if ( !$title->isSubpage() && $title->getNamespace() == NS_USER ) {
+		global $wgEnableWallExt;
```
lizlux: I generally opt for `F::app()->wg` rather than the global.

lizlux:
Also, might be worth adding a comment that these hooks are called for wall and forum, regardless of whether wall is enabled.

jsutterfield:
I agree with Liz that adding some docs as to why this logic is added would be good. Otherwise, LGTM. Thanks @Grunny!

extensions/wikia/Wall/WallHooksHelper.class.php
Grunny: I've never had a strong preference for that if the code isn't in a Nirvana controller, since `F::app()` is still a global ;), but switched since it's consistent with everywhere else in Wall. :)

Grunny:
Added a comment at the top, wasn't really sure where best to put it. :)

PR #6577, opened 2015-03-06
axWFactoryDomainQuery: add a LIMIT to the query

macbre:
This function makes queries that can return thousands of rows that are then transfered to the front-end (where only six first items are visible).

``` sql
SELECT /* axWFactoryDomainQuery Raylan13 */  city_id,city_domain  FROM `city_domains`  WHERE (city_domain not like 'www.%') AND (city_domain not like '%.wikicities.com') AND (city_domain  LIKE '%supe%' )
```
- set a proper response content type

@Grunny

### DIFF
--- a/extensions/wikia/WikiFactory/SpecialWikiFactory_ajax.php
+++ b/extensions/wikia/WikiFactory/SpecialWikiFactory_ajax.php
@@ -636,7 +636,10 @@ function axWFactoryDomainQuery() {
 				"city_domain not like '%.wikicities.com'",
 				"city_domain {$cityDomainLike}"
 			],
-			__METHOD__
+			__METHOD__,
+			[
+				'LIMIT' => 15
+			]
 		);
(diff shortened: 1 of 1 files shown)


extensions/wikia/WikiFactory/SpecialWikiFactory_ajax.php
```diff
@@ -636,7 +636,10 @@ function axWFactoryDomainQuery() {
 				"city_domain not like '%.wikicities.com'",
```
macbre:
```sql
>select count(*) from city_domains where city_domain like '%.wikicities.com';
+----------+
| count(*) |
+----------+
|     1092 |
+----------+
1 row in set (0.29 sec)
```

:)

Grunny:
:+1: LGTM.

wladekb:
lgtm

PR #6579, opened 2015-03-07
DAT-2591 Really use DBname for Latest Activity API

sebastianmarzjan:
The WikiaDataAccess used the same key for different wikis. Despite the 5s TTL,
it was still noticed by users. This REALLY uses the dbname.
Also, we do not use filter method for cache (not needed)


jacekjursza:
LGTM

PR #6587, opened 2015-03-09
Return empty array if no redirects are find for a given pageType

hakubo:
Previously if you wanted redirects for 'hubsv3' and it was not in the RedirectPages,
it would return the whole redirects list.

### DIFF
--- a/includes/wikia/services/RedirectService.class.php
+++ b/includes/wikia/services/RedirectService.class.php
@@ -73,6 +73,7 @@ class RedirectService extends WikiaService {
 	public function getRedirects() {
 		if ( empty( $this->redirects ) ) {
 			global $wgRedirectPages;
+
 			if ( !empty( $wgRedirectPages ) ) {
 				$this->redirects = $wgRedirectPages;
 			}
@@ -82,11 +83,17 @@ class RedirectService extends WikiaService {
 			$this->redirects = $this->prepareRedirects();
 		}
 
-		if ( isset( $this->pageType )
-			&& isset( $this->redirects[ $this->pageType ] )
-			&& is_array( $this->redirects[ $this->pageType ]) )
-		{
-			return $this->redirects[ $this->pageType ];
+
+		if ( isset( $this->pageType ) ) {
+			if (
+				isset( $this->redirects[ $this->pageType ] )
+				&& is_array( $this->redirects[ $this->pageType ])
+			) {
+				return $this->redirects[ $this->pageType ];
+			} else {
+				//If we don't find redirects for this pageType
+				$this->redirects = [];
+			}
 		}
 
 		return $this->redirects;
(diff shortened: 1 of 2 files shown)


hakubo:
This should actually brake the tests, 
who knows how this should behave?

hakubo:
@kvas-damian @SebastianMarzjan do you know how this should behave?
This fixes problem with Title.php, but will it brake anything else?

includes/wikia/services/RedirectService.class.php
```diff
@@ -82,11 +83,17 @@ public function getRedirects() {
...
-		if ( isset( $this->pageType )
-			&& isset( $this->redirects[ $this->pageType ] )
-			&& is_array( $this->redirects[ $this->pageType ]) )
-		{
-			return $this->redirects[ $this->pageType ];
+
+		if ( isset( $this->pageType ) ) {
+			if (
+				isset( $this->redirects[ $this->pageType ] )
+				&& is_array( $this->redirects[ $this->pageType ])
```
bognix: you first check if key is set and then you check if this is an array, why not other way around? If an array, then check the key.

kvas-damian:
@hakubo redirects format is decribed here: https://github.com/Wikia/app/blob/dev/includes/wikia/services/RedirectService.class.php#L3

If there is no ```pageType``` key then we use entire array as ```old_url => new_url``` mapping

includes/wikia/services/RedirectService.class.php
kvas-damian: @bognix you need to make sure that this key is present to check if it's value is an array.

bognix: aaa sorry! I misread...

hakubo: if I check ['a'][1] this will give me a notice: Notice: Undefined offset

hakubo:
@kvas-damian does that sound good now?

includes/wikia/services/RedirectService.class.php
```diff
@@ -82,11 +78,17 @@ public function getRedirects() {
 			$this->redirects = $this->prepareRedirects();
 		}
 
-		if ( isset( $this->pageType )
-			&& isset( $this->redirects[ $this->pageType ] )
-			&& is_array( $this->redirects[ $this->pageType ]) )
-		{
-			return $this->redirects[ $this->pageType ];
+
+		if ( !empty( $this->pageType ) ) {
```
kvas-damian: Does it still make sense to keep this condition separately?

hakubo: I think it improves readability

kvas-damian: ok, lets keep it then.

kvas-damian:
LGTM

tomasznapieralski:
All automated WamPageTest passed

PR #6596, opened 2015-03-09
adding option for frame query param and test

nmonterroso:
@drsnyder 
https://wikia-inc.atlassian.net/browse/PLATFORM-984

PHP implementation of `frame` query option added in https://github.com/Wikia/vignette/pull/79


drsnyder:
:sheep: it

PR #6597, opened 2015-03-09
Allow purging vignette urls via surrogate key

owend:
Forgot to add this functionality last week.  Gating behind a separate config global so we can phase it in.  If that becomes the default, we can always remove the global var.  

This code is so not functional style it's painful.  :/  

@nmonterroso @drsnyder

### DIFF
--- a/includes/cache/SquidUpdate.php
+++ b/includes/cache/SquidUpdate.php
@@ -124,28 +124,26 @@ class SquidUpdate {
 
 		global $wgPurgeSquidViaCelery, $wgPurgeVignetteUsingSurrogateKeys;
 		if ( $wgPurgeSquidViaCelery == true ) {
-			if ( $wgPurgeVignetteUsingSurrogateKeys == true ) {
-				// Filter urls into buckets based on service backend
-				$buckets = array_reduce($urlArr, function($carry, $item) {
-					if ( VignetteRequest::isVignetteUrl($item) ) {
-						$carry['vignette'][] = $item;
-					} elseif ( strstr($item, 'MercuryApi') !== false ) {
-						$carry['mercury'][] = $item;
-					} else {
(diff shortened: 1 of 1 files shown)


includes/cache/SquidUpdate.php
```diff
@@ -122,13 +122,33 @@ static function purge( $urlArr ) {
...
-					->setArgs( $urlArr, [] )
-					->setPriority( PurgeQueue::NAME )
-					->queue();
+			if ( $wgPurgeVignetteUsingSurrogateKeys == true ) {
+				// Filter array for vignette urls and split them into separate task
+				$vignetteUrls = [];
+				foreach ( $urlArr as $index => $url ) {
+					if ( VignetteRequest::isVignetteUrl($url) ) {
+						$vignetteUrls[] = $url;
+						unset($urlArr[$index]);
```
nmonterroso: I should probably know this, but does `unset()` during a forloop on the array which you're `unset`ing change the iteration of the for loop?

owend: It's fine, it does leave "gaps" in the index like 0, 1, 3, 4   But in this case we don't care.

drsnyder: Minor: Another way to do this might look like:

```PHP
$sets = array_reduce($urlArr, function($carry, $item) {
   if ( VignetteRequest::isVignetteUrl($url) ) {
      $carry['vignette'][] = $url;
   } else {
      $carry['other'][] = $url;
   }
}, array('vignette' => array(), 'other' => array());
```

Then

```php
if ( !empty($sets['vignette']) ) {
...
}
```

owend: Since I had to add a third type (mercury) this approach makes a lot more sense.

PR #6601, opened 2015-03-10
CE-1283 Make poweruser a global group

adamkarminski:
[See CE-1283](https://wikia-inc.atlassian.net/browse/CE-1283)

Making _poweruser_ a global group makes more sense and will make its management easier.

Ping @Wikia/community-engineering @macbre (as promised :) )


lgarczewski:
LGTM.

PR #6607, opened 2015-03-10
VE-1825: Port support for new ReCaptcha and FancyCaptcha

inez:


extensions/VisualEditor/wikia/modules/ve/init/ve.init.wikia.ViewPageTarget.js
```diff
@@ -358,3 +367,64 @@ ve.init.wikia.ViewPageTarget.prototype.maybeShowDialogs = function () {
...
+	$.nirvana.sendRequest( {
+		controller: 'CaptchaController',
+		method: 'getFancyCaptcha',
+		type: 'GET',
+		callback: function ( data ) {
+			getFancyCaptchaDeferred.resolve( data );
+		}
+	} );
+
+	$.when(
```
inez: @jsutterfield You might want to take a look at this.

jsutterfield: Ah, thanks @inez! Yea, I had been looking through the code base to figure out the best way to load that scss asynchronously, but this makes a lot more sense than the solution I came up with. I'm book marking it to use as a reference. Thanks for pinging me on it!

PR #6610, opened 2015-03-11
Fix overlapping i18n ExtensionMessagesFiles keys

owend:
@inez


inez:
LGTM! Thanks @owend !

PR #6547, opened 2015-03-05
CONCF-191 enable new GlobalFooter on Oasis

bognix:

PR #6611, opened 2015-03-11
VE-1854: Revert change to the box-sizing that break buttons in VE

inez:
This change is reverting box-sizing applied to all elements (*) that was introduced recently within this change https://github.com/Wikia/app/pull/6547 and it breaks buttons in VE (https://wikia-inc.atlassian.net/browse/VE-1854).

### DIFF
--- a/skins/oasis/css/core/body.scss
+++ b/skins/oasis/css/core/body.scss
@@ -7,11 +7,6 @@ html {
 
 }
 
-*, *::before, *::after {
-	box-sizing: inherit;
-}
-
-
 body {
 	color: $color-text;
 	font-family: Helvetica, Arial, sans-serif;


vforge:
@inez Can you please tell me what's the reasoning behind this change? With ```box-sizing: content-box;``` on ```html``` element to implement different ```box-sizing``` you would need to override on all levels of your markup. We want to maintain compatibility with existing Oasis styles and still have possibility to change it as simply as possible; please reference https://css-tricks.com/inheriting-box-sizing-probably-slightly-better-best-practice/ .
Can you think of different way of doing what we intended?
Pinging @lizlux @kenkouot and @RafLeszczynski for their opinion on this topic.

lizlux:
It seems that the changes made [here](https://github.com/Wikia/app/commit/6a5c29ed3ae1bf72249cff9cf70f775ce0129e9f#diff-fbc5af87ea8ea9a35796e23c2148b4c5R5) are superfluous and add unnecessary levels of specificity. The default for `box-sizing` is `content-box`, so there's no need to set it again unless you're overriding some other setting, which I don't think we are. 

@vforge, the example you linked to is a best practice for when you want to override the default box-sizing to `border-box`, not keep it the same with `content-box`.

I haven't confirmed this with the VE code (mainly b/c I'm not sure exactly how to get to the screens listed in https://wikia-inc.atlassian.net/browse/VE-1854) but my assumption is the added specificity is overriding the `box-sizing` values for those buttons. 

@inez it would be helpful if you provided a description in this PR per the [PR guidelines](https://github.com/Wikia/guidelines-internal/blob/master/Process/PullRequests.md). 

In general I would be extremely wary of making global changes to properties like `box-sizing`, given the complexity of our app and the impossibility of visually testing every last detail.

inez:
Perhaps there are different solutions to this problem, however change that this PR is reverting was introduced very recently here: https://github.com/Wikia/app/pull/6547 and that PR does not contain information why box-sizing was changed (or at least I can't find it) - that's why I reached to @bognix (the is the author of particular change to box-sizing).

bognix:
While porting extensions from Venus skin (where everything was styled based on box-sizing: border-box) we wanted to avoid creating multiple costly selectors in every extension we are moving (for now: GlobalNavigation, GlobalFooter, LocalNavigation)
```
*, *::after, *::before
```
We decided to introduce this change globally and if some component of page would need different box-sizing than content-box it would be enough to change it's value only on root element of this component (without taking care of all children and pseudo elements because they will already inherit).

inez:
@bognix This change potentially could a have a lot of negative impact on many UI elements - as it had in case of VE. Did you or someone else already performed some sort of an audit or test - not only for mentioned GlobalNavigation, GlobalFooter, LocalNavigation - but for other elements that could be affected?

bognix:
We made a mistake with not communicating it properly.
We assumed if some component is designed in some box-sizing (border-box let say) then all elements of this component are based on this box-sizing.
This PR would break GlobalNavigation, GlobalFooter and LocalNavigation.
I would create another one to remove the global change and add this change on the extension level.

Although I am going to collect numbers and check what performance impact it has. If it would be meaningful we can come up with some solution.

inez:
Thanks @bognix. And closing this one.

PR #6613, opened 2015-03-11
CONCF-119 RightRail restyle

dianafa:
https://wikia-inc.atlassian.net/browse/CONCF-119

Chat component doesn't work on sandboxes/devboxes (to test chat user hover behavior we need multiple users on one chat), thus I enabled chat (see ChatEntryPoint.js) component styling only for one of my articles on mediawiki119.

The .wikia-rail covers all properties of old .WikiaRail class, so the latter can be safety removed, and style doesn't change (BUT the other dependences i.e. JS directives based on WikiaRail class are not replaced! )


extensions/wikia/Chat2/js/ChatEntryPoint.js
```diff
@@ -93,6 +93,13 @@ var ChatEntryPoint = {
 			var $e = $(this);
 			$e.html($.msg($e.data('msg-id'), $e.data('msg-param')));
 		});
+
+        //enables live-chat styling only on one page -
+        //chat couldn't be tested on devbox/sandbox
+        // so we'll see how it behaves on production
+        if (window.wgTitle == 'Chat test diana') {
```
Warkot: Why not `===`?

extensions/wikia/Rail/styles/RailElement.scss
```diff
@@ -0,0 +1,269 @@
+@import "skins/shared/color";
```
Warkot: You might want to install pre-commit hooks from the https://github.com/Wikia/guidelines repo. Those would run `scss-lint` and tell you immediately that single quotes are preferred. But what is more important:
```
extensions/wikia/Rail/styles/RailElement.scss:55 [W] SelectorDepth: Selector should have depth of applicability no greater than 4, but was 5
extensions/wikia/Rail/styles/RailElement.scss:116 [W] SelectorDepth: Selector should have depth of applicability no greater than 4, but was 5
extensions/wikia/Rail/styles/RailElement.scss:200 [W] SelectorDepth: Selector should have depth of applicability no greater than 4, but was 5
extensions/wikia/Rail/styles/RailElement.scss:209 [W] SelectorDepth: Selector should have depth of applicability no greater than 4, but was 5
extensions/wikia/Rail/styles/RailElement.scss:225 [W] SelectorDepth: Selector should have depth of applicability no greater than 4, but was 5
extensions/wikia/Rail/styles/RailElement.scss:233 [W] SelectorDepth: Selector should have depth of applicability no greater than 4, but was 6
extensions/wikia/Rail/styles/RailElement.scss:236 [W] SelectorDepth: Selector should have depth of applicability no greater than 4, but was 7
extensions/wikia/Rail/styles/RailElement.scss:240 [W] SelectorDepth: Selector should have depth of applicability no greater than 4, but was 7
extensions/wikia/Rail/styles/RailElement.scss:247 [W] SelectorDepth: Selector should have depth of applicability no greater than 4, but was 8
extensions/wikia/Rail/styles/RailElement.scss:255 [W] SelectorDepth: Selector should have depth of applicability no greater than 4, but was 5
extensions/wikia/Rail/styles/RailElement.scss:259 [W] SelectorDepth: Selector should have depth of applicability no greater than 4, but was 7
```
Deep nesting of classes is lowering the page rendering performance. Maybe there's a way to lower the indentation depth? I'll look into that too and try to suggest solutions.

Warkot:
Swap spaces for tabs, please.

extensions/wikia/Chat2/js/ChatEntryPoint.js
bognix: I would vote for not merging this to dev.
For test purposes - ok. But still we can achieve the same result  by adding class to WikiaRail in browser.
With this thing you need to remember to remove this after it is tested.

extensions/wikia/Rail/styles/RailElement.scss
bognix: lint would also fix the order of properties - it should be alphabetical.

extensions/wikia/Rail/styles/RailElement.scss
```diff
@@ -0,0 +1,269 @@
...
+@if $is-dark-wiki {
+    $color-rail-background: mix($color-page, $white, 90%);
+}
+
+
+@mixin wikia-rail-element() {
+    background-color: $color-rail-background;
+    padding: 20px;
+
+    zoom: 1;
```
Warkot: Maybe scss-lint doesn't say that about mixins, but attributes should be listed in an alphabetical order.

skins/oasis/css/oasis.scss
```diff
@@ -74,6 +74,8 @@
 
 // CSS (needs leading slash and file extension)
 @import '/extensions/wikia/CreatePage/css/CreatePage.css';
+//Site Optimizationsgit
```
bognix: remove git

extensions/wikia/Rail/styles/RailElement.scss
```diff
@@ -0,0 +1,269 @@
...
+                        }
+
+                        li {
+                            border-bottom: $color-divider solid thin;
+                            cursor: pointer;
+                            height: 44px;
+                            margin: 0;
+                            white-space: nowrap;
+
+                            .label {
```
bognix: instead of 
```
 li:hover {
   background-color: $color-buttons;
}

 li {
 border-bottom: $color-divider solid thin;
 cursor: pointer;
 height: 44px;
 margin: 0;
 white-space: nowrap;
}
```
you can do 
```
li {
    ...
    &:hover  {
        ...
    }
}

File: extensions/wikia/Rail/styles/RailElement.scss [hunk 6]
```diff
@@ -0,0 +1,269 @@
...
+    padding: 20px;
+
+    zoom: 1;
+    margin-bottom: 10px;
+    position: relative;
+    word-wrap: break-word;
+}
+
+.wikia-rail {
+
```
Warkot: Don't add additional empty lines after class declarations, they are only needed before it.

extensions/wikia/Rail/styles/RailElement.scss
```diff
@@ -0,0 +1,269 @@
...
+
+        .thumbnails {
+
+            a:first-of-type {
+                color: $color-links;
+                margin-bottom: 0;
+            }
+
+            a:first-of-type:hover {
```
bognix: you can use '&' like I describe in comment below

extensions/wikia/Chat2/js/ChatEntryPoint.js
dianafa: hmm. ok. I'll remove it before merging to dev.

extensions/wikia/Rail/styles/RailElement.scss
```diff
@@ -0,0 +1,269 @@
...
+            text-shadow: initial;
+        }
+
+        .activity-heading {
+            font-size: 24px;
+            font-weight: bold;
+            margin-bottom: 20px;
+        }
+
+        li em a {
```
Warkot: Maybe we can drop something here, simplify this selector?
And the other thing, you could place this rule earlier or later in comparison to the class/pseudo-class selectors, for better readability. But be aware that's not always applicable as attributes in CSS are overwriting themselves.

extensions/wikia/Rail/styles/RailElement.scss
```diff
@@ -0,0 +1,269 @@
...
+            @include wikia-rail-element;
+
+            p {
+                color: mix($color-text, $color-page, 60%);
+                font-size: 12px;
+            }
+
+            .chat-headline {
+
+                .chat-live {
```
Warkot: Is there anything using `.chat-live` and `.chat-total` classes outside the `.chat-headline` class? Maybe you can drop the `.chat-headline` selector and decrease the nesting?

extensions/wikia/Rail/styles/RailElement.scss
```diff
@@ -0,0 +1,269 @@
...
+                button {
+                    background-image: none;
+                    border: 0;
+                    border-radius: 0;
+                    color: $white;
+                    height: 100%;
+                    text-transform: uppercase;
+                }
+
+                button:before {
```
Warkot: As @bognix already mentioned, you can write it with the `&`:
```
button {
	…
	
	&:before {
		…
	}
}
```

extensions/wikia/Rail/styles/RailElement.scss
```diff
@@ -0,0 +1,269 @@
...
+            .avatar:hover {
+                opacity: 1;
+            }
+
+            .chat-whos-here {
+                border: 0;
+                margin: 20px 0;
+                padding: 0;
+
+                .arrow-left {
```
Warkot: Any `.arrow-left` and latter classes outside the `.chat-whos-here`? Maybe it's possible to de-nest it by removing the `.chat-whos-here` selector? Maybe similar can be applied to the latter classes?

extensions/wikia/Rail/styles/RailElement.scss
```diff
@@ -0,0 +1,269 @@
...
+                .arrow-right {
+                    padding-right: 0;
+
+                    img {
+                        border-color: transparent;
+                        border-left-color: $color-links;
+                    }
+                }
+
+                .arrow-left.disabled, .arrow-right.disabled {
```
Warkot: This could be split into two lines.
I was also thinking about making a mixin out of this and then adding it with `&.disabled` in  the respective `.arrow-…` classes, but I'm not sure which is better. @vforge?

extensions/wikia/Rail/styles/RailElement.scss
dianafa: the more class I add on the path the stronger the condition will be and will override old .WikiaRail class.

extensions/wikia/Rail/styles/RailElement.scss
dianafa: As above: 
the more class I add on the path the stronger the condition will be and will override old .WikiaRail class.

extensions/wikia/Rail/styles/RailElement.scss
dianafa: I'm using githooks ;) and deep selectors are mostly caused by overriding existing properties and selectors.

extensions/wikia/Rail/styles/RailElement.scss
```diff
@@ -0,0 +1,262 @@
...
+}
+
+
+@mixin wikia-rail-element() {
+    background-color: $color-rail-background;
+    margin-bottom: 10px;
+    padding: 20px;
+    position: relative;
+    word-wrap: break-word;
+    zoom: 1;
```
bognix: As far as I remember zoom: 1 was some sort of a hack for IE. Do we still need it?

extensions/wikia/Rail/styles/RailElement.scss
```diff
@@ -0,0 +1,262 @@
...
+
+.wikia-rail {
+    float: right;
+    overflow: hidden;
+    padding: 0 10px 10px 0;
+    width: 300px;
+
+    h1 {
+        padding-left: 0;
+        font-weight: bold;
```
bognix: h1 is by default bold.
http://www.w3.org/TR/html-markup/h1.html

extensions/wikia/Rail/styles/RailElement.scss
```diff
@@ -0,0 +1,264 @@
...
+        &:before {
+            display: none;
+        }
+}
+
+@mixin wikia-rail-element() {
+    background-color: $color-rail-background;
+    margin-bottom: 10px;
+    padding: 20px;
+    position: relative;
```
bognix: do we need this styling? I removed it at it looks the same for RWA and chat

extensions/wikia/Rail/styles/RailElement.scss
```diff
@@ -0,0 +1,264 @@
...
+    word-wrap: break-word;
```
bognix: do we need this styling? I removed it at it looks the same for RWA and chat

extensions/wikia/Rail/styles/RailElement.scss
```diff
@@ -0,0 +1,264 @@
...
+                    overflow: hidden;
+                    text-overflow: ellipsis;
+                    white-space: nowrap;
+                    word-wrap: normal;
+                }
+            }
+        }
+    }
+
+    &.new-chat {
```
bognix: some parts of chat module are still styled based on old styles (.ChatModule .chat-contents and .ChatModule .chat-joins)

dianafa: Hmm. OK, I was thinking we don't want only old .WikiaRail class. If we don't want ANY of old styles, ok I can change that.

bognix: I don't know either :) It is up to you.

dianafa: well, these are hundreads of code to replace. I'm not sure if it makes sense.

bognix:
LGTM

PR #6617, opened 2015-03-11
CE-1392: Highlight wikitext syntax if article starts with text

lukaszkonieczny:
https://wikia-inc.atlassian.net/browse/CE-1392


adamkarminski:
LGTM!

PR #6621, opened 2015-03-11
ADEN-1401 Floor ad

gabrys:


extensions/wikia/AdEngine/js/template/floor.js
```diff
@@ -0,0 +1,67 @@
...
+	'wikia.iframeWriter',
+	'wikia.window'
+], function ($, log, doc, iframeWriter, win) {
+	'use strict';
+
+	var logGroup = 'ext.wikia.adEngine.template.floor',
+		footerHtml = '<div id="ext-wikia-adEngine-template-footer">' +
+			'<div class="background"></div>' +
+			'<div class="ad"></div>' +
+			'<a class="close" title="Close" href="#"><span>Close</span></a>' +
```
nandy-andy: It doesn't matter for us if it's English or Polish, does it? Everyone understands "Close" in advertisments :japanese_ogre:

nandy-andy:
LGTM :+1: I've got one small i18n concern but we can discuss it offline and it shouldn't be a big blocker before doing QA.

gabrys:
It's the label, so I don't see urging need for translating that. I'm 
welcome to accept a bug for that later ;-).

PR #6622, opened 2015-03-11
PLATFORM-1012: Log writable SQL queries in readonly mode

macbre:
And log when we switch to read-only mode because of lagged slaves

@wladekb / @owend / @drozdo


owend:
Excellent.  I've wanted to investigate that slave-lag behavior for a while now, this is the first step!

PR #6623, opened 2015-03-11
FilePageController: do not treat this query as a writable one

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-1012

MediaWiki fallbacks to database read-only mode when all slaves are lagged. When in read-only mode a simple check is performed on every query sent to the database. It assumes that the query does not contain leading spaces.

Trying to find the root cause of read-only fallback in #6622

@wladekb / @garthwebb


wladekb:
lgtm

PR #6625, opened 2015-03-11
ADEN-1800 Adding Turtle

gabrys:


extensions/wikia/AdEngine/AdEngine2ContextService.class.php
```diff
@@ -71,6 +71,7 @@ public function getContext( Title $title, $skinName ) {
 				'forceProviders' => $this->filterOutEmptyItems( [
 					'directGpt' => $wg->AdDriverForceDirectGptAd,
 					'liftium' => $wg->AdDriverForceLiftiumAd,
+					'turtle' => !!$wg->AdDriverForceTurtleAd,
```
nandy-andy: Why are we forcing it to be `Boolean` and other (above) aren't casted? Shouldn't we cast all of them just in matter of consistency?

nandy-andy:
Except my curiosity connected to small inconsistency the changes look good :+1:

extensions/wikia/AdEngine/AdEngine2ContextService.class.php
gabrys: Ha! Good catch. No reason for that. We can either cast them or not cast them. There's no point to do both. Sorry about it. The reason is I moved the line from the block where we cast vars to bools to the block where we don't.

nandy-andy:
LGTM :+1:

PR #6628, opened 2015-03-11
VE-1854

bognix:
Revert changes in box-sizing made on "html" level and introduce them on "extension" level.


inez:
LGTM. I will give a chance for other people to comment on it (@lizlux?) and eventually merge it this afternoon.

lizlux:
Yeah, I think this is a good approach. It would be nice to have enough confidence in our system to be able to make the more sweeping change, but it's just not feasible. 

Thanks for taking care of this!

PR #6630, opened 2015-03-11
VE-1857: Invalidate RL startup whenever new VE is turned on or off

inez:


owend:
LGTM, based on review and questions answered in offline conversation.  needs to be tested in production. :)

PR #6631, opened 2015-03-11
Update WikiaView to allow for finer grain template locations/names

garthwebb:
This allows controller methods to define a @template annotation which will determine what base template name will be used to render that method's output, e.g.:

``` php
class FooController {
  const DEFAULT_TEMPLATE_ENGINE = \WikiaResponse::TEMPLATE_ENGINE_MUSTACHE;

  /**
   * @template myView
   */
  public function bar() {
    // code
  }
}
```

The view will look for a series of possible template names, falling back until it finds one that exists.  So in the above example it will check:
- extensions/wikia/Foo/templates/myView.mustache
- extensions/wikia/Foo/templates/Foo_bar.mustache
- extensions/wikia/Foo/templates/FooController_bar.mustache

This pull request also allows the controller to define an alternate template path via a `getTemplateDir` method.  If this method exists, the return value will be used as the template directory:

``` php
class FooController {
  public static function getTemplateDir() {
    return dirname( __FILE__ ) . '/subfolder/templates';
  }
}
```

In this case the search paths will look like:
- extensions/wikia/Foo/subfolder/templates/myView.mustache
- extensions/wikia/Foo/subfolder/templates/Foo_bar.mustache
- extensions/wikia/Foo/subfolder/templates/FooController_bar.mustache

Note if the directory given does not exist, it will fall back to the default location.

@Wikia/platform-team

### DIFF
--- a/includes/wikia/nirvana/WikiaView.class.php
+++ b/includes/wikia/nirvana/WikiaView.class.php
@@ -170,17 +170,9 @@ class WikiaView {
 	public function getTemplateOptions( $controllerClass, $methodName ) {
 		$templates = [];
 
-		// Make sure this method exists, otherwise the call to getMethod crashes PHP
-		// so badly it can't even log that a problem occurred.
-		if ( method_exists( $controllerClass, $methodName ) ) {
-			// See if there is a @template annotation for the method we're generating a view for
-			$reflection = new ReflectionClass( $controllerClass );
-			$method = $reflection->getMethod( $methodName );
-
-			$comment = $method->getDocComment();
-			if ( preg_match( '/@template ([^ ]+)/', $comment, $matches ) ) {
-				$templates[] = trim( $matches[ 1 ] );
-			}
+		$fromAnnotation = $this->getTemplateAnnotation( $controllerClass, $methodName );
+		if ( !empty( $fromAnnotation ) ) {
+			$templates[] = $fromAnnotation;
 		}
 
 		// Add variations on the controller name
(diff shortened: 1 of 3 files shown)


includes/wikia/nirvana/WikiaView.class.php
```diff
@@ -31,16 +31,12 @@ function __construct( WikiaResponse $response = null ) {
 	 * @param string $methodName
 	 * @param array $data
 	 * @param string $format
+	 *
+	 * @return WikiaView
 	 */
-	public static function newFromControllerAndMethodName( $controllerName, $methodName, Array $data = array(), $format = WikiaResponse::FORMAT_HTML ) {
-		$app = F::app();
+	public static function newFromControllerAndMethodName( $controllerName, $methodName, Array $data = [], $format = WikiaResponse::FORMAT_HTML ) {
 		// Service classes must be dispatched by full name otherwise we default to a controller.
```
garthwebb: The following 6 lines were functionally duplicated later in code so replaced by common method ```normalizeControllerClass```

includes/wikia/nirvana/WikiaView.class.php
```diff
@@ -114,30 +109,22 @@ protected function buildTemplatePath( $controllerClass, $methodName, $forceRebui
 			}
 
 			// Service classes must be dispatched by full name otherwise we default to a controller.
```
garthwebb: Lines from here up to L126 were rolled into common function ```normalizeControllerClass```

includes/wikia/nirvana/WikiaView.class.php
```diff
@@ -114,30 +109,22 @@ protected function buildTemplatePath( $controllerClass, $methodName, $forceRebui
...
-				$controllerClass = $app->getServiceClassName( $controllerBaseName );
-			} else {
-				$controllerClass = $app->getControllerClassName( $controllerBaseName );
-			}
-
-			if( empty( $wgAutoloadClasses[$controllerClass] ) ) {
-				throw new WikiaException( "Invalid controller or service name: {$controllerClass}" );
-			}
-
-			// First we look for BaseName_MethodName
```
garthwebb: Put these changed lines into a loop looking at each fallback option.  The options are generated by ```getTemplateOptions```

owend:
This LGTM.

includes/wikia/nirvana/WikiaView.class.php
```diff
@@ -114,36 +109,106 @@ protected function buildTemplatePath( $controllerClass, $methodName, $forceRebui
...
+	 * @param $methodName
+	 *
+	 * @return array
+	 */
+	public function getTemplateOptions( $controllerClass, $methodName ) {
+		$templates = [];
+
+		// Make sure this method exists, otherwise the call to getMethod crashes PHP
+		// so badly it can't even log that a problem occurred.
+		if ( method_exists( $controllerClass, $methodName ) ) {
```
wladekb: I'd add an request-wide cache here for cases when you render many items separately so this doesn't add a lot of overhead (reflection used to be slow).

includes/wikia/nirvana/WikiaView.class.php
```diff
@@ -114,36 +109,106 @@ protected function buildTemplatePath( $controllerClass, $methodName, $forceRebui
...
+
+	/**
+	 * See if the controller defines a custom template directory, otherwise use the default directory
+	 *
+	 * @param $controllerClass
+	 *
+	 * @return string
+	 */
+	public function getTemplateDir( $controllerClass ) {
+		if ( method_exists( $controllerClass, 'getTemplateDir' ) ) {
```
wladekb: I'd rather go with a dummy function returning null defined in base class so you can override it in a subclass if you need.

wladekb:
left 2 comments, look good overall

garthwebb:
@wladekb, can you take another look?  I've updated the PR

garthwebb:
Ok, merging this PR after making the suggested changes.

PR #6634, opened 2015-03-11
SOC-450 First working version of email extension

garthwebb:
This is a first pass at the Email extension.  The 'forgot password' email was used as an example email in order to exhibit the Email extensions workings.

https://wikia-inc.atlassian.net/browse/SOC-450


extensions/wikia/Email/EmailController.class.php
```diff
@@ -0,0 +1,279 @@
...
+
+		if ( $user->getId() == 0 ) {
+			throw new Fatal( wfMessage( 'emailext-error-empty-user' )->escaped() );
+		}
+	}
+
+	/**
+	 * @throws \Email\Fatal
+	 */
+	public function assertUserHasEmail() {
```
jsutterfield: Is there a reason why we wouldn't want to check this inside of the this base class's `assertCanEmail()` method? Right now it's being called explicitly by the `ForgotPasswordController` subclass. Making sure the user has an email seems like something we'd always want to check.

extensions/wikia/Email/EmailController.class.php
```diff
@@ -0,0 +1,279 @@
...
+		$email = $this->targetUser->getEmail();
+		if ( empty( $email ) ) {
+			throw new Fatal( wfMessage( 'emailext-error-noemail' )->escaped() );
+		}
+	}
+
+	/**
+	 * @throws \Email\Fatal
+	 */
+	public function assertHasIP() {
```
jsutterfield: Why do we need to make sure they have an IP?

extensions/wikia/Email/EmailController.class.php
```diff
@@ -0,0 +1,279 @@
...
+
+	/**
+	 * Get the 'reply-to' address for the send of this email, if different than the 'from' address.
+	 *
+	 * @return \MailAddress|null
+	 */
+	protected function getReplyToAddress() {
+		$replyAddr = null;
+		if ( !empty( $this->wg->NoReplyAddress ) ) {
+			$name = wfMessage('emailext-no-reply-name' )->escaped();
```
jsutterfield: White space here, as well as lines 229, 264, 265, and 276.

extensions/wikia/Email/EmailController.class.php
jsutterfield: Ah, looks like it might be for the hook run in `ForgotPasswordController`? Can we move this method definition into `ForgotPasswordController` there then, or do we expect this to be used in other places?

includes/wikia/nirvana/WikiaView.class.php
```diff
@@ -114,36 +109,106 @@ protected function buildTemplatePath( $controllerClass, $methodName, $forceRebui
...
+		// Make sure this method exists, otherwise the call to getMethod crashes PHP
+		// so badly it can't even log that a problem occurred.
+		if ( method_exists( $controllerClass, $methodName ) ) {
+			// See if there is a @template annotation for the method we're generating a view for
+			$reflection = new ReflectionClass( $controllerClass );
+			$method = $reflection->getMethod( $methodName );
+
+			$comment = $method->getDocComment();
+			if ( preg_match( '/@template ([^ ]+)/', $comment, $matches ) ) {
+				$templates[] = trim( $matches[ 1 ] );
```
jsutterfield: Don't need the padding inside of `matches`, should be 
```php
$templates[] = trim( $matches[1] );
```

extensions/wikia/Email/EmailController.class.php
```diff
@@ -0,0 +1,279 @@
...
+			$this->assertCanEmail();
+
+			$toAddress = $this->getToAddress();
+			$fromAddress = $this->getFromAddress();
+			$replyToAddress = $this->getReplyToAddress();
+			$subject = $this->getSubject();
+			$body = [ 'html' => $this->getBody() ];
+
+			if ( !$this->test ) {
+				\UserMailer::send( $toAddress, $fromAddress, $subject, $body, $replyToAddress );
```
jsutterfield: Should we be checking the Status object returned by `\UserMailer::send` before returning an ok response?

extensions/wikia/Email/EmailController.class.php
```diff
@@ -0,0 +1,279 @@
...
+	public static function getTemplateDir() {
+		return dirname( __FILE__ ) . '/templates';
+	}
+
+	public function init() {
+		try {
+			$this->assertCanAccessController();
+
+			$this->currentUser = $this->findUserFromRequest( 'currentUser', $this->wg->User );
+			$this->targetUser = $this->findUserFromRequest( 'targetUser', $this->wg->User );
```
jsutterfield: What's the reason for having a currentUser and targetUser again? Also, in `assertCanEmail` of this base class we're checking if the current user wants email and is not blocked. It seems like we'd want to be checking that for the target user instead since they're the one receiving the email, right?

extensions/wikia/Email/EmailController.class.php
```diff
@@ -0,0 +1,279 @@
...
+	 * @template Email_preview
+	 *
+	 * @return bool
+	 *
+	 * @throws \MWException
+	 */
+	public function handle() {
+		// If something previously has thrown an error (likely 'init') don't continue
+		if ( $this->hasErrorResponse ) {
+			return true;
```
jsutterfield: Do we need to return a boolean from here?

includes/wikia/nirvana/WikiaView.class.php
jsutterfield: Also, :+1: again. This is super cool.

extensions/wikia/Email/EmailController.class.php
garthwebb: At least in the case of forgot email, these can be (and are treated differently) different users.  There's the user who is making the request (likely a non-logged in anonymous user but could be a real user) and then there's the user from whatever 'username' is sent to us.

We want to check current user in the for being blocked below because that is the user who is actually logged in currently requesting to take an action.

So current user is whoever is taking the action, and target user is whoever the email is being sent to.  Most often they are the same.

extensions/wikia/Email/templates/ForgotPassword_body.mustache
```diff
@@ -0,0 +1,11 @@
+{{> Email_header}}
```
lizlux: It might be easier to have one main layout template that has a header, footer, and `{{{content}}}` variable as opposed to including the header and footer partials in each template. This would reduce the duplicate code of always having to include the header and footer partials. This is how mercury has been structuring their templates and I think it's a good idea.

garthwebb: I would consider this template (and Email_header/Email_footer) a proof of concept.  I only intended to show how this could work and I fully expect that when we do the first real email a much better template will be created ;)

lizlux: :+1:

extensions/wikia/Email/EmailController.class.php
garthwebb: Well, its a little bit of the tail wagging the dog, but it was because I wanted to return on L98 directly from the ```setErrorResponse call```.  PHPStorm complains if ```setErrorResponse``` returns ```null``` and I try to return that, and if I return a value there, the other two returns have to return something.  I can just make L98 call ```setErrorResponse``` and then put the return on the line below.  Thanks for making bloat the code! ;)

extensions/wikia/Email/EmailController.class.php
garthwebb: Sounds like a good idea to me!

extensions/wikia/Email/EmailController.class.php
garthwebb: Seems reasonable.

extensions/wikia/Email/EmailController.class.php
garthwebb: Yah, the hook needs it.  My expectation is that other emails may need it as well.  I'd say it would be fine to move into ForgotPassword *after* we've finished updating the emails and can see whether its only used there or not.

extensions/wikia/Email/EmailController.class.php
```diff
@@ -0,0 +1,292 @@
...
+		}
+
+		throw new Fatal( wfMessage( 'emailext-error-restricted-controller' )->escaped() );
+	}
+
+	/**
+	 * Allow child classes to set some initial values.  Added here so its always defined
+	 * but is kept blank so does not need to be called via parent::initEmail()
+	 */
+	public function initEmail() {
```
lizlux: Should this be an abstract function?

includes/wikia/tests/WikiaViewTest.php
```diff
@@ -39,7 +39,6 @@ public function testCreatingNewFromControllerAndMethod( $controllerName, $method
 	public function setTemplateDataProvider() {
 		return array(
 		 array( true, 'Test' ),
-		 array( true, 'Oasis' ), // tmp, while modules still exists
```
lizlux: Just curious why this was removed.

extensions/wikia/Email/EmailController.class.php
garthwebb: If it was an abstract function it would *have* to be defined in the child class, even if the child class didn't need to add any thing.

includes/wikia/tests/WikiaViewTest.php
garthwebb: The unit tests were failing because of this line.  The test in question was purposefully faking out controller entries in ```$wgAutoloadClasses``` for some unknown reason here:

  https://github.com/Wikia/app/blob/SOC-450/includes/wikia/tests/WikiaViewTest.php#L49-L62

Owen and I tried to figure out what this god awful test was doing and failed.  Line 41 above works because they defined a fake controller in their fake test path, but line 42 failed because there was no ```OasisController.class.php``` at the fake path.  Given that it doesn't test anything useful, we removed it.

extensions/wikia/Email/EmailController.class.php
```diff
@@ -0,0 +1,292 @@
...
+	}
+
+	/**
+	 * This checks the 'unsubscribed' user option which is linked to the 'Disable all emails from Wikia'
+	 * email preference
+	 *
+	 * @throws \Email\Check
+	 */
+	public function assertUserWantsEmail() {
+		if ( $this->currentUser->getBoolOption( 'unsubscribed' ) ) {
```
lizlux: If a user is 'unsubscribed' does that mean they don't want to receive a password reminder email? What does 'unsubscribed' mean in this case?

Also, to Jame's earlier point, this is being checked on the current user, not the target user. Shouldn't it be the target user?

extensions/wikia/Email/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+These classes should inherit from `Email\EmailController`.  The methods this controller can
+override/implement are:
+
+* assertCanAccessController : This determines who can access this controller.  The defaults, staff and internal
+requests should be sufficient for almost all cases.
+* assertCanEmail : Determines whether the email should be sent given current business logic.
+* initEmail : Any initialization should be done here.  This includes error checking and/or loading request values.
+* getToAddress : Supplies the `To` address.  Default is current user.
+* getFromAddress : Supplies the `From` address.  Defaults to `$wgPasswordSender` and `$wgPasswordSenderName`
```
lizlux: This variable name, `$wgPasswordSender` is misleading for a default "from" address that may or may not have to do with sending a password. A quick search through the code shows it's being used a bunch of times though, so it may be the kind of tech debt that's too hard to dig out of. Just wondering if you considered changing this name to something more semantic.

extensions/wikia/Email/EmailController.class.php
garthwebb: The ```unsubscribed``` option is the option that "Disable all emails from Wikia (overrides all other options)" maps to from the preferences.

I can update this to run against ```targetUser```

extensions/wikia/Email/README.md
garthwebb: I hadn't really considered changing this, but I expect when we're finished moving things into this extension, many of those uses will be eliminated.  I'd vote to updating this name afterward.  I created a ticket for this: https://wikia-inc.atlassian.net/browse/SOC-550

extensions/wikia/Email/EmailController.class.php
jsutterfield: Sounds good to me.

extensions/wikia/Email/EmailController.class.php
jsutterfield: That's what I'm here for :P

extensions/wikia/Email/EmailController.class.php
```diff
@@ -0,0 +1,292 @@
...
+		try {
+			$this->assertCanAccessController();
+
+			$this->currentUser = $this->findUserFromRequest( 'currentUser', $this->wg->User );
+			$this->targetUser = $this->findUserFromRequest( 'targetUser', $this->wg->User );
+			$this->test = $this->getRequest()->getVal( 'test', false );
+
+			$this->initEmail();
+		} catch ( ControllerException $e ) {
+			return $this->setErrorResponse( $e );
```
jsutterfield: I don't think you need to be returning `$this->setErrorResponse( $e );`. Simply calling it should be enough.

extensions/wikia/Email/EmailController.class.php
```diff
@@ -0,0 +1,292 @@
...
+			return;
+		}
+
+		$this->response->setData( [
+			'result' => 'ok',
+			'subject' => $subject,
+			'body' => $body['html'],
+		] );
+
+		return;
```
jsutterfield: Don't think you need this return. Or the one on line 128.

extensions/wikia/Email/EmailController.class.php
jsutterfield: Ah, makes sense. Thanks for clarifying.

extensions/wikia/Email/EmailController.class.php
```diff
@@ -0,0 +1,292 @@
...
+			'body',
+			$this->request->getParams()
+		);
+		return $html;
+	}
+
+	protected function findUserFromRequest( $paramName, \User $default = null ) {
+		$userName = $this->getRequest()->getVal( $paramName );
+		if ( empty( $userName ) ) {
+			return $default;
```
jsutterfield: Should we make sure that `$default` is a user object before returning it? It might be nice to check and throw an exception like we do for `$paramName` since `$default` could be null. Or, we could make the default argument `$this->wg->User`. That could work as well.

garthwebb: The function signature requires that ```$default``` be a ```User``` object or ```null```.  Also, ```null``` may be acceptable for optional arguments so I don't want to assume that the caller will want ```$wgUser```

includes/User.php
```diff
@@ -2685,6 +2685,14 @@ public function isAnon() {
 	}
 
 	/**
+	 * Whether this user is Wikia staff or not
+	 * @return bool
+	 */
+	public function isStaff() {
+		return in_array( 'staff', $this->getEffectiveGroups() );
```
armonr: It may be time to dedicate a constant to staff, say, `GROUP_STAFF`? Imagine not having to grep for `"staff"/'staff'` anymore!

includes/wikia/nirvana/WikiaView.class.php
```diff
@@ -114,36 +109,106 @@ protected function buildTemplatePath( $controllerClass, $methodName, $forceRebui
...
 			}
 
-			// First we look for BaseName_MethodName
-			$dirName = dirname( $wgAutoloadClasses[$controllerClass] );
-			$basePath = "{$dirName}/templates/{$controllerBaseName}_{$methodName}";
-			$templatePath = "{$basePath}.$extension";
-			$templateExists = file_exists( $templatePath );
+			$this->setTemplatePath( $templatePath );
+		}
+		wfProfileOut(__METHOD__);
```
armonr: Might as well remove profiling

includes/wikia/nirvana/WikiaView.class.php
```diff
@@ -114,36 +109,106 @@ protected function buildTemplatePath( $controllerClass, $methodName, $forceRebui
...
 			// Service classes must be dispatched by full name otherwise we default to a controller.
-			$controllerBaseName = $app->getBaseName( $controllerClass );
-			if ($app->isService($controllerClass)) {
-				$controllerClass = $app->getServiceClassName( $controllerBaseName );
-			} else {
-				$controllerClass = $app->getControllerClassName( $controllerBaseName );
+			$controllerClass = $this->normalizeControllerClass( $controllerClass );
+
+			$templateExists = false;
+			$templatePath = '';
```
armonr: Not your code, but `$templateFileName` would be a better name.

includes/wikia/nirvana/WikiaView.class.php
```diff
@@ -114,36 +109,106 @@ protected function buildTemplatePath( $controllerClass, $methodName, $forceRebui
...
+
+	/**
+	 * Generates a list of possible template names ordered by preference
+	 *
+	 * @param $controllerClass
+	 * @param $methodName
+	 *
+	 * @return array
+	 */
+	public function getTemplateOptions( $controllerClass, $methodName ) {
```
armonr: Should this be public?

includes/wikia/nirvana/WikiaView.class.php
```diff
@@ -114,36 +109,106 @@ protected function buildTemplatePath( $controllerClass, $methodName, $forceRebui
...
+	}
+
+	/**
+	 * See if the controller defines a custom template directory, otherwise use the default directory
+	 *
+	 * @param $controllerClass
+	 *
+	 * @return string
+	 */
+	public function getTemplateDir( $controllerClass ) {
```
armonr: Same, this need not be public

includes/wikia/nirvana/WikiaView.class.php
```diff
@@ -114,36 +109,106 @@ protected function buildTemplatePath( $controllerClass, $methodName, $forceRebui
...
+		}
+
+		return $controllerClass;
+	}
+
+	/**
+	 * Generates a list of possible template names ordered by preference
+	 *
+	 * @param $controllerClass
+	 * @param $methodName
```
armonr: Please add types

includes/wikia/nirvana/WikiaView.class.php
```diff
@@ -114,36 +109,106 @@ protected function buildTemplatePath( $controllerClass, $methodName, $forceRebui
 			}
 
 			// Service classes must be dispatched by full name otherwise we default to a controller.
-			$controllerBaseName = $app->getBaseName( $controllerClass );
-			if ($app->isService($controllerClass)) {
-				$controllerClass = $app->getServiceClassName( $controllerBaseName );
-			} else {
-				$controllerClass = $app->getControllerClassName( $controllerBaseName );
+			$controllerClass = $this->normalizeControllerClass( $controllerClass );
```
armonr: `self::`

includes/wikia/nirvana/WikiaView.class.php
```diff
@@ -114,36 +109,106 @@ protected function buildTemplatePath( $controllerClass, $methodName, $forceRebui
...
+	 * For legacy reasons we sometimes get a $controllerClass name that doesn't end with
+	 * 'Controller' or 'Service'.  Normalize to this form.
+	 *
+	 * @param $controllerClass
+	 *
+	 * @return string
+	 *
+	 * @throws WikiaException
+	 */
+	public static function normalizeControllerClass( $controllerClass ) {
```
armonr: private?

extensions/wikia/Email/EmailExceptions.class.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace Email;
+
+class ControllerException extends \Exception {
+	public $errorType = 'error';
```
armonr: Maybe replace with `function getErrorType()`, and derived classes override this method.

extensions/wikia/Email/EmailController.class.php
```diff
@@ -0,0 +1,289 @@
...
+	/**
+	 * Return a user object from a username
+	 *
+	 * @param string $username
+	 *
+	 * @return \User
+	 * @throws \Email\Fatal
+	 */
+	protected function getUserFromName( $username ) {
+		if ( is_null( $username ) ) {
```
armonr: Consider `!$username` which also catches the empty case.

extensions/wikia/Email/EmailController.class.php
```diff
@@ -0,0 +1,289 @@
...
+namespace Email;
+
+abstract class EmailController extends \WikiaController {
+	const DEFAULT_TEMPLATE_ENGINE = \WikiaResponse::TEMPLATE_ENGINE_MUSTACHE;
+
+	/** @var \User The user associated with the current request */
+	protected $currentUser;
+
+	/** @var \User The user this email will be sent to */
+	protected $targetUser;
```
armonr: Consider adding getters for `$currentUser` and `$targetUser`. Maybe I've been overdosing on Java lately :)

armonr:
Looks so well-organized already! Nicely done @garthwebb!

jsutterfield:
LGTM!

lizlux:
LGTM once @armonr's comments are addressed.

includes/User.php
garthwebb: This starts to pull the thread to a lot of code ... I'm not sure I want to expand this PR for that many potential changes ;)

includes/wikia/nirvana/WikiaView.class.php
garthwebb: Hmm, not sure if I agree.  "Filename" could mean the full path or just the name portion of a path, so its a bit ambiguous.

includes/wikia/nirvana/WikiaView.class.php
garthwebb: Given this is a pretty core function of our Nirvana framework and Macbre says he still uses this for somethings, I'm hesitant to remove this.

extensions/wikia/Email/EmailController.class.php
garthwebb: I think I'll leave these.  They don't need public but immutable access and I've never been a huge fan of getters for the sake of getters.

includes/User.php
armonr: I was not suggesting replacing all the instances, but just starting the pattern. I think we eventually need to do that for all such user groups. But, sure, it is not for this PR :)

armonr:
:shipit:

PR #6635, opened 2015-03-12
Mail Diff hackathon resolving conflicts

owend:
Quick hack to add change diffs to mail.  Uses existing mediawiki diff engine and some random css file from a 3rd party extension.  

From what I remember the only problem with this approach is that lots of mail clients don't support external css so we might need to transform it to inline.  

DO NOT MERGE, this is just a hackathon project that might be productized soon.  

@garthwebb


wladekb:
Is this going to be eventually merged or will stay as PoC?

owend:
This was turned into a "real" project so I think this can be deleted, but I'll check with that team.

PR #6637, opened 2015-03-12
ADEN-1829 AmazonMatch to support 300x600 bids

gabrys:


nandy-andy:
LGTM :gem: Should we create a task ticket for removing old names after... a month? ;)

PR #6641, opened 2015-03-12
CE-1405: Tidy editor toolbar

lukaszkonieczny:
https://wikia-inc.atlassian.net/browse/CE-1405

### DIFF
--- a/extensions/wikia/EditPageLayout/css/core/modules.scss
+++ b/extensions/wikia/EditPageLayout/css/core/modules.scss
@@ -103,7 +103,7 @@ $color-icon: mix($color-text, #fff, 90%);
 			text-align: center;
 		}
 
-		#wpDiff.diff-link {
+		.diff-link {
 			color: $color-links;
 			display: inline-block;
 			font-size: 12px;
--- a/extensions/wikia/EditPageLayout/css/core/modules.scss
+++ b/extensions/wikia/EditPageLayout/css/core/modules.scss
@@ -128,10 +128,9 @@ $color-icon: mix($color-text, #fff, 90%);
 		}
 
 		.wikia-menu-button {
-			//float: left;
 			*z-index: 2;
-			width: 110px; // preview has border 1px wide
 			height: 21px;
+			width: 110px; // preview has border 1px wide
 			& > a {
 				display: inline-block;
 				width: 75px;
--- a/extensions/wikia/EditPageLayout/css/core/modules.scss
+++ b/extensions/wikia/EditPageLayout/css/core/modules.scss
@@ -131,7 +131,7 @@ $color-icon: mix($color-text, #fff, 90%);
 			//float: left;
 			*z-index: 2;
 			width: 110px; // preview has border 1px wide
-			height: 19px;
+			height: 21px;
 			& > a {
 				display: inline-block;
 				width: 75px;
--- a/extensions/wikia/EditPageLayout/css/core/wide.scss
+++ b/extensions/wikia/EditPageLayout/css/core/wide.scss
@@ -194,6 +194,7 @@
 
 					#wpSummary {
 						height: 20px;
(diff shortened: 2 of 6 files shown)


extensions/wikia/EditPageLayout/js/plugins/Widescreen.js
```diff
@@ -17,10 +17,34 @@
...
+		initNodes: function() {
+			this.$rail = $('#EditPageRail');
+			this.$editpage = $('#EditPage');
+			this.$toolbar = $('#EditPageToolbar .cke_toolbar_source');
+			this.expandedToolbarBreakpoint =  this.$toolbar.width() + this.$rail.width();
+			this.initialized = true;
+		},
+
+		sizeChanged: function() {
+			if ( this.initialized && this.editor.mode == 'source' && this.wide ) {
```
kamilkoterba: === instead of ==

extensions/wikia/EditPageLayout/css/core/modules.scss
```diff
@@ -131,7 +131,7 @@ $color-icon: mix($color-text, #fff, 90%);
 			//float: left;
 			*z-index: 2;
 			width: 110px; // preview has border 1px wide
-			height: 19px;
+			height: 21px;
```
kamilkoterba: alphabetical order

extensions/wikia/EditPageLayout/css/core/wide.scss
```diff
@@ -263,6 +241,35 @@
...
+
+			h3 {
+				margin-top: 8px;
+			}
+
+			.preview_mobile {
+				padding-top: 3px;
+			}
+
+			.preview_desktop {
```
kamilkoterba: don't you think we can simplify these nested selectors
#EditPage.toolbar-expanded .preview_box .preview_mobile
by moving preview_mobile, preview_desktop and h3 level up, and possibly adding a class to h3?

up to you.

extensions/wikia/EditPageLayout/css/core/wide.scss
```diff
@@ -263,6 +241,35 @@
...
+				padding-top: 5px;
+			}
+		}
+
+		#wpDiff.diff-link {
```
kamilkoterba: It looks like you're overriding some css selectors here. Would be great if we don't have to load style definitions we don't use.

As well per coding conventions it's advised to define css selectors rather with classes than ids.

extensions/wikia/EditPageLayout/js/plugins/Widescreen.js
```diff
@@ -17,10 +17,34 @@
...
+				if ( this.$editpage.width() > this.expandedToolbarBreakpoint ) {
```
kamilkoterba: spaces around brackets not needed

extensions/wikia/EditPageLayout/js/plugins/Widescreen.js
```diff
@@ -55,6 +81,10 @@
 				this.editor.ck.fire('resize');
 			}
 
+			if (!this.initialized && this.editor.mode == 'source' && this.wide) {
```
kamilkoterba: === instead of ==

extensions/wikia/EditPageLayout/css/core/wide.scss
lukaszkonieczny: Good idea :)

extensions/wikia/EditPageLayout/css/core/wide.scss
lukaszkonieczny: Yep, that's tricky but ```toolbar-expanded``` class is toggle via javascript after screen resize so we need to load these css definitions.

kamilkoterba:
LGTM! Thanks for addressing all comments :)

PR #6642, opened 2015-03-12
Re-add required $optionsWithArgs array

jsutterfield:
During the refactor of VideoIngestion done a few months ago (https://github.com/Wikia/app/pull/5339), a bug was introduced which borked the way command line arguments were being processed. This meant that when you passed any unix time stamp for start time or end time, it would always evaluate to 1.

The problem was the removal of [this line](https://github.com/Wikia/app/pull/5339/files#diff-e0718dbca917544223928e9a3ebc229fL14). That `$optionsWithArgs` array is used by the included `commandLine.inc` file below it. This PR re-adds that line to get the command line arguments working again.


bognix:
LGTM

PR #6643, opened 2015-03-12
MAIN-4105 Make sure to only load Captha.js once

jsutterfield:
There was a bug where if a hook to get a Captcha was called multiple times, we would load the Captcha.js library multiple times as well. This fix makes sure we only load the Captcha.js library once.

Ticket: https://wikia-inc.atlassian.net/browse/MAIN-4105

### DIFF
--- a/extensions/wikia/Captcha/Factory/Module.class.php
+++ b/extensions/wikia/Captcha/Factory/Module.class.php
@@ -48,6 +48,10 @@ class Module {
 	 * Load Captcha.js on demand when we're getting an instance of a Captcha.
 	 */
 	public static function addCaptchaJS() {
-		\Wikia::addAssetsToOutput( 'captcha_js' );
+		// Make sure we only load FancyCaptcha library once
+		if ( !\F::app()->wg->CaptchaLibraryLoaded ) {
+			\Wikia::addAssetsToOutput( 'captcha_js' );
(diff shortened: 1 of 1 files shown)


armonr:
LGTM

extensions/wikia/Captcha/Factory/Module.class.php
```diff
@@ -48,6 +48,10 @@ public static function verifyCaptchaClass( $captchaClass ) {
 	 * Load Captcha.js on demand when we're getting an instance of a Captcha.
 	 */
 	public static function addCaptchaJS() {
-		\Wikia::addAssetsToOutput( 'captcha_js' );
+		// Make sure we only load FancyCaptcha library once
+		if ( !\F::app()->wg->CaptchaLibraryLoaded ) {
```
lizlux: Looks like a php [static variable](http://php.net/manual/en/language.variables.scope.php#language.variables.scope.static) could be used here instead of a global

jsutterfield: Ah, good idea @lizlux. I took this approach since I remember we did something similar for the Videos Module (https://github.com/Wikia/app/blob/dev/extensions/wikia/VideosModule/VideosModuleHooks.class.php#L55), but the static variable would probably be a better solution. Let's try and update that the next time we touch this file.

PR #6647, opened 2015-03-13
Ce 1427 make all update fallback create

kamilkoterba:
https://wikia-inc.atlassian.net/browse/CE-1427

First part of fallback to create changes. Fallback to create for user data changes.

Aim is to make all ExactTarget updates fallback to create due to sync process concept between Wikia and ExactTarget. Once ET clears DB and waits for data dump from Wikia we want to send hot data and create missing records.

@Wikia/community-engineering


gabrys:
LGTM :-)

PR #6648, opened 2015-03-13
wfShellExec: log missing executable files as well

macbre:
To make it easier to find the root cause of https://wikia-inc.atlassian.net/browse/PLATFORM-1026 and similar issues.

@michalroszka


michalroszka:
:+1:

PR #6652, opened 2015-03-13
CE-1388: Fix highlighting when switching between visual and source mode

lukaszkonieczny:
https://wikia-inc.atlassian.net/browse/CE-1388


adamkarminski:
LGTM :+1:

PR #6653, opened 2015-03-13
CE-1395: Fix Lua console when syntax highlighting is enabled

lukaszkonieczny:
https://wikia-inc.atlassian.net/browse/CE-1395


adamkarminski:
LGTM :+1:

PR #6660, opened 2015-03-13
Load BannerNotifications only when needed

bkoval:
@lizlux Could you take a look? This way we make sure to require BannerNotifications after the DOM is initialized. This should fix the Uncaught again (I was able to reproduce it in about 1/10 - 1/20 of the cases, really weird stuff).

I'm starting to feel this AMD support for BannerNotifications is to shaky, so maybe we should consider withdrawing it at some point and leaving just the global constructor...


lizlux:
Let's try https://github.com/Wikia/app/pull/6661/files instead.

PR #6661, opened 2015-03-13
SOC-528 fixing race condition with JS asset loading

lizlux:
This make sure we're using the same asset loading mechanism for the CookiePolicy extension and the BannerNotification extension. This should get rid of the race condition. 

Ticket: https://wikia-inc.atlassian.net/browse/SOC-528

Note that we're still using ResourceLoader for enabling wikitext parsing in JS messages.


bkoval:
@lizlux Looks good, thanks! Let's use it and then come back to the discussion about removing AMD support for BannerNotification :)

armonr:
LGTM!

PR #6663, opened 2015-03-13
Soc 442 app

SeanColombo:
Two small changes related to the front-end of chat.

The first one is just deleting a method-definition that was literally a duplicate. The benefit is really small, just somewhat-smaller filesize and one less potential area for disaster (if you changed the first method definition, the second method would actually still be overwriting it, so your changes wouldn't work). ;) #Cleanliness.

The second change is to upgrade the client-side to use the latest socket.io: 1.3.5.  The server is already using 1.3.5 to great success... this will put them on the same version.  The only expected changes are that the socket.io team claims there are some performance improvements (those might have all been server-side though, because the server is doing great now).

### DIFF
--- a/extensions/wikia/Chat2/js/models/models.js
+++ b/extensions/wikia/Chat2/js/models/models.js
@@ -52,14 +52,6 @@ var STATUS_STATE_AWAY = 'away';
 		}
 	});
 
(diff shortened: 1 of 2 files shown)


owend:
LGTM

PR #6666, opened 2015-03-16
Add tests for our specific PHP extensions

macbre:
Prevent regressions during next PHP upgrade:
- https://wikia-inc.atlassian.net/browse/PLATFORM-1026
- https://wikia-inc.atlassian.net/browse/PLATFORM-1029

Check that `wikidiff2` and `mustache` PHP extensions are loaded..Our code & MW core depends on them and does not work well when applying a fallback.

@michalroszka


michalroszka:
LGTM :+1:

PR #6668, opened 2015-03-16
PHP error in ArticlesApiController.class.php fix

dianafa:
https://wikia-inc.atlassian.net/browse/CONCF-216

The error appeared only on one article page. Turned out that the issue was category name: "User:". Because of colon, function getAsJson() treated 'User' as category namespace and the empty string as category title. Then, it wanted to call getText() on empty string.

After considering three options I decided for first solution for how to treat (and return from API) category like "User:" or "Template:" so name of category which consists of wiki namespace and colon.

1) Check if the Title::newFromText doesn't return an empty $categoryTitle (it does for "User:" etc. and do not proceed with them (users on Mercury will not be able to see this category, on desktop will be fine)

2) Remove ":" if it is the last sign in category title.
if ( substr($t -> mDbkeyform, -1 ) == ':' ) {
   $t->mDbkeyform = substr($t -> mDbkeyform, 0, -1);
}
But if user creates category with colon at the end he probably would like to have this colon at the end...

3)  If it is a string like "User:" function secureAndSplit() treat it like proper name, not like the namespace with an empty title. 
if ( strcmp($this->mDbkeyform, $p.":") && ( ... ) {...}

This hovever caused fatal error in LightboxController : "File:" `File:` is not a valid file title.
It was caused by the issue that the same method secureAndSplit() is called for almost all elements on the wiki pages.

After consideration we decided for the first solution as is the least invasive. However it is not a good solution to not display categories like "User:" on Mercury.


includes/Title.php
```diff
@@ -147,6 +147,11 @@ public static function newFromText( $text, $defaultNamespace = NS_MAIN ) {
 
 		$t = new Title();
 		$t->mDbkeyform = str_replace( ' ', '_', $filteredText );
+		# Prevent error when title ends with colon and title
+		# is interpreted as namespace:[empty title]
+		if ( substr($t -> mDbkeyform, -1 ) == ':' ) {
+			$t->mDbkeyform = substr($t -> mDbkeyform, 0, -1);
+		}
 		$t->mDefaultNamespace = $defaultNamespace;
```
bognix: This method occurs in our code more than 1000 times. I'm not sure if changing its logic is safe. For sure it will break for articles which end with ":" or articles which contain only ":".

bognix:
LGTM

tomasznapieralski:
LGTM

PR #6671, opened 2015-03-16
ADEN-1831 do not load liftium if there is noexternals set to 1

nandy-andy:
When we put `?noexternals=1` in the URL in network tab we can still see a call send to liftium. This behaviour is changed with code changes below.


gabrys:
LGTM :+1:

PR #6674, opened 2015-03-16
SOC-528 fixing race condition with message not loading

lizlux:
I was see the cookie policy message showing up like `<cookie-policy-notification-message>`, even though the message was loading later on and was available when you type `mw.message('cookie-policy-notification-message').parse()` into the browser console. This will make sure the message is available before showing the banner. 

Also renaming `ext.cookiePolicy` to `ext.cookiePolicyMessages` to better describe what it is.


bkoval:
LGTM :)

PR #6676, opened 2015-03-17
(MAIN-4139) Fix message output format

Grunny:
Fix message output format to prevent stored XSS issue.

The link is HTML and so the contents shouldn't be further processed and
the output format should be `escaped()`.

/cc @owend


owend:
LGTM

PR #6678, opened 2015-03-17
Ce 1427 fallback to create wiki updates

kamilkoterba:
https://wikia-inc.atlassian.net/browse/CE-1427

Second (last) part of fallback to create changes. Fallback to create for wiki data changes.

Aim is to make all ExactTarget updates fallback to create due to sync process concept between Wikia and ExactTarget. Once ET clears DB and waits for data dump from Wikia we want to send hot data and create missing records.

@Wikia/community-engineering


lgarczewski:
LGTM.

PR #6679, opened 2015-03-17
PLATFORM-1028: improve sessions handling around log in and log out actions

macbre:
This pull request tries to improve the handling of log in and log out actions and avoid race conditions problems caused by long running requests.

https://wikia-inc.atlassian.net/browse/PLATFORM-1028
## Part 1:

Clear `wsUserName` session entry and the cookie when logging out. There's no need to keep the user name (in both session and cookie) when you log out.
## Part 2

Regenerate session ID on user login and logout (and delete the memcache entry for the old session)

@michalroszka / @Grunny  / @garthwebb (as the member of the team that owns the login)

### DIFF
--- a/includes/User.php
+++ b/includes/User.php
@@ -3026,6 +3026,13 @@ class User {
 		$this->clearCookie( 'UserID' );
 		$this->clearCookie( 'Token' );
 
+		// Wikia change - begin (@see PLATFORM-1028)
+		// @author macbre
+		// There's no need to keep the user name (in both session and cookie) when you log out
+		$this->getRequest()->setSessionData( 'wsUserName', null );
+		$this->clearCookie( 'UserName' );
+		// Wikia change - end
+
 		# Remember when user logged out, to prevent seeing cached pages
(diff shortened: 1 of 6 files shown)


michalroszka:
:+1: looks good! I am wondering whether we could simply gedt rid of the cookie, in further steps.

includes/User.php
```diff
@@ -3026,6 +3026,13 @@ public function doLogout() {
 		$this->clearCookie( 'UserID' );
 		$this->clearCookie( 'Token' );
 
+		// Wikia change - begin (@see PLATFORM-1028)
+		// @author macbre
+		// There's no need to keep the user name (in both session and cookie) when you log out
+		$this->getRequest()->setSessionData( 'wsUserName', null );
+		$this->clearCookie( 'UserName' );
```
michalroszka: The cookie name is `wikicitiesUserName` and the prefix seems not be added automatically by `setcookie` methods called later.

macbre: `setcookie` method in WebResponse class does use `$wgCookiePrefix`.

PR #6682, opened 2015-03-17
ADEN-1800 Adjust leaderboard once an ad loaded from Turtle

nandy-andy:
If Turtle had loaded higher ad we didn't make the slot higher. Changes below fixes it.


gabrys:
LGTM

PR #6683, opened 2015-03-17
DAT-2606 change message copy for error on addting article URL while creating POI ...

RafLeszczynski:
...on map


SebastianMarzjan:
lgtm

PR #6685, opened 2015-03-18
SOC-394 Loading FB sdk conditionally

lizlux:
Loading FB sdk conditionally if there's XFBML tags in the content. We were loading it on every page until the parser cache cleared, and it's cleared!


macbre:
:+1:

jsutterfield:
LGTM.

PR #6691, opened 2015-03-18
refactor ExternalUser_Wikia for integration with Helios

ghost:
The INSERT in ExternalUser_Wikia::addToDatabase has been divided into two pieces:
- an INSERT with a subset of data, which is going to be replaced with
  a call to Helios.
- a series of assignments and a call to User::saveSettings() to add
  data Helios does not add.

### DIFF
--- a/includes/wikia/ExternalUser_Wikia.php
+++ b/includes/wikia/ExternalUser_Wikia.php
@@ -251,37 +251,44 @@ class ExternalUser_Wikia extends ExternalUser {
 			$dbw = wfGetDB( DB_MASTER, array(), $wgExternalSharedDB );
 			$User->setToken();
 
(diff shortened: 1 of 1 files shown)


includes/wikia/ExternalUser_Wikia.php
```diff
@@ -251,37 +251,44 @@ protected function addToDatabase( $User, $password, $email, $realname ) {
 			$dbw = wfGetDB( DB_MASTER, array(), $wgExternalSharedDB );
 			$User->setToken();
 
+            // The following INSERT is similar (column-wise) to one done by Helios and is going
```
wladekb: Indentation is off in the changes in this file

PR #6694, opened 2015-03-18
MAIN-4151 set memc key instead of reading it only

bognix:


extensions/wikia/GlobalFooter/GlobalFooterController.class.php
```diff
@@ -80,6 +80,8 @@ private function getGlobalFooterLinks() {
 			}
 		}
 
+		$wgMemc->set($memcKey, $parsedLinks);
```
macbre: Per our Slack discussion:

* add `$ttl` argument
* spacing :)

macbre: Please add error reporting in line 66 if we treat such case as an error.

macbre:
Jolly good! Thanks for fixing it so quick :)

extensions/wikia/GlobalFooter/GlobalFooterController.class.php
```diff
@@ -80,6 +86,8 @@ private function getGlobalFooterLinks() {
 			}
 		}
 
+		$wgMemc->set( $memcKey, $parsedLinks, self::MEMC_EXPIRY );
```
Warkot: There's a slight possibility that we cache an empty response i.e. when:
* there is no message in Memcache
* AND `getMessageAsArray()` finds a message, but its content is empty (`is_null($var)` is equal to `$var === NULL`, so an empty response is still not a `NULL`—lines 65-66).
But that may be also intended. :confused:

bognix: I believe if message key exists we should display its value despite what it contains and if it contains anything. As you said that might be an error but on the other hand that may be intended. Messages are controlled by us so I believe we can assume if key exists then we should show whatever value it contains.

Warkot: Agreed, let's keep it this way. Thanks for the reply!

Warkot:
LGTM :shipit:

PR #6697, opened 2015-03-18
CONCF-296 Make Mercury siteName consistent with wgSiteName

dianafa:
https://wikia-inc.atlassian.net/browse/CONCF-296

MercuryApi.class.php function getWikiVariables() was setting siteName from WF variable (wgSiteName) OR individualized SITENAME_MSG_KEY. This was introduced in Dec 2014 - as Bert said- to improve SEO and to be consistent with wikiamobile skin.

This PR divides this two things placing SITENAME_MSG_KEY in meta tags and siteName in top of the page. 

Second part of this ticket can be found: https://github.com/Wikia/mercury/pull/624


extensions/wikia/MercuryApi/models/MercuryApi.class.php
```diff
@@ -127,19 +125,12 @@ public function getWikiVariables() {
...
 	public function getSiteName() {
-		$siteName = F::app()->wg->Sitename;
-		$msg = wfMessage( static::SITENAME_MSG_KEY )->inContentLanguage();
-
-		if( !$msg->isDisabled() ) {
-			$siteName = $msg->text();
-		}
-
-		return $siteName;
+		return F::app()->wg->Sitename;
```
Warkot: What do you think about moving this to line [122/120](https://github.com/Wikia/app/pull/6697/files#diff-21bab35e67dfc42ffdd18afcf2c13b6cR120)? `getWikiVariables()` already has cached `$wg = F::app()->wg;`.

bognix:
Could you re-add the removed unit tests and change the name of the method which is called to new one?

bognix:
LGTM

bognix:
ping @hakubo

hakubo:
LGTM

bognix:
I changed the logic slightly. 
Instead of fallbacking to wgSitename for disabled siteMessage I return false and handle it inside Mercury. I checked Mercury's code and the only usage of siteMessage is in keywords so no need to duplicate them in case siteMassage is disabled (siteName is already in keywords) 

ping @hakubo @rogatty @dianafa

extensions/wikia/MercuryApi/models/MercuryApi.class.php
```diff
@@ -106,40 +106,37 @@ public function getNumberOfUserContribForArticle( $articleId, $userId ) {
...
-		$siteName = F::app()->wg->Sitename;
+	public function getSiteMessage() {
 		$msg = wfMessage( static::SITENAME_MSG_KEY )->inContentLanguage();
-
-		if( !$msg->isDisabled() ) {
-			$siteName = $msg->text();
-		}
-
-		return $siteName;
+		return !$msg->isDisabled() && !empty( $msg->text() ) ? $msg->text() : false;
```
rogatty: Are we already using PHP 5.6 in production? `empty( $msg->text() )` won't work if we're not.

extensions/wikia/MercuryApi/models/MercuryApi.class.php
```diff
@@ -106,40 +106,40 @@ public function getNumberOfUserContribForArticle( $articleId, $userId ) {
...
-	public function getSiteName() {
-		$siteName = F::app()->wg->Sitename;
+	public function getSiteMessage() {
 		$msg = wfMessage( static::SITENAME_MSG_KEY )->inContentLanguage();
-
-		if( !$msg->isDisabled() ) {
-			$siteName = $msg->text();
+		if ( !$msg->isDisabled() ) {
+			$msgText = $msg->text();
+			return !empty( $msgText ) ? $msgText : false;
```
rogatty: This should be outside the condition block.

extensions/wikia/MercuryApi/models/MercuryApi.class.php
```diff
@@ -106,40 +106,40 @@ public function getNumberOfUserContribForArticle( $articleId, $userId ) {
 	 * @return mixed
 	 */
 	public function getWikiVariables() {
-		$wg = F::app()->wg;
+		global $wgSitename, $wgCacheBuster, $wgDBname, $wgDefaultSkin,
```
Warkot: I am curious, why are you using explicit globals here, how is it better than the previous solution?

rogatty:
LGTM

extensions/wikia/MercuryApi/models/MercuryApi.class.php
dianafa: It's faster than calling F::app()->wg

tomasznapieralski:
LGTM

PR #6699, opened 2015-03-18
PLATFORM-997: improve user settings cache handling

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-997

- set the shared `user_touched` key to the current timestamp on miss
- when invalidating the cache update `user_touched` shared key to the current timestamp instead of deleting it
- when invalidating the cache store the updated data in memcache, instead of deleting the key (reduce the number of misses on `*:user:id:*` keys by ~66% and avoid slave lag issues).

@michalroszka / @adamkarminski

### DIFF
--- a/includes/User.php
+++ b/includes/User.php
@@ -371,6 +371,8 @@ class User {
 		$key = wfMemcKey( 'user', 'id', $this->mId );
 		global $wgMemc;
 		$wgMemc->set( $key, $data );
+
+		wfDebug( "User: user {$this->mId} stored in cache\n" );
 	}
 
 	/** @name newFrom*() static factory methods */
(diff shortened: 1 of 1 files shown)


michalroszka:
:+1: LGTM. As a side not, I am only concerned about extensive use of wfDebug() as IIRC "we" decided to get rid of it in favour of ELK. But I might be wrong about that.

macbre:
`wfDebug` is used solely for debugging purposes. It makes no sense to log these cache misses as we have a different mechanism to collect such metrics.

macbre:
> After this fix went live we got a huge decrease of queries from `User::loadOptions` - from 16 to 2,8k a minute.

PR #6704, opened 2015-03-18
SOC-564 Add logging for the public interfaces to chat

garthwebb:
@owend What do you think about this?  I realized we don't really have much/any logging coming in from chat.  Would have been nice for last nights mysterious outage.

I tried to log all "interesting" functions.  This means any public function that performs an action, but not including mundane actions such as logging.

https://wikia-inc.atlassian.net/browse/SOC-564

### DIFF
--- a/extensions/wikia/Chat2/NodeApiClient.class.php
+++ b/extensions/wikia/Chat2/NodeApiClient.class.php
@@ -28,9 +28,18 @@ class NodeApiClient {
 			$cityId = $cityData->{'cityId'};
 		} else {
 			// FIXME: How should we handle it if there is no cityId?
+			ChatHelper::info( __METHOD__ . ': Method called - no cityId', [
+				'roomId' => $roomId,
+			] );
 		}
 
 		wfProfileOut( __METHOD__ );
+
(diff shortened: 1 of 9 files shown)


armonr:
:shipit:

extensions/wikia/Chat2/NodeApiClient.class.php
```diff
@@ -110,7 +127,9 @@ static private function makeRequest($params){
 		global $wgReadOnly;
 		wfProfileIn( __METHOD__ );
 		$response = "";
-		
+
+		ChatHelper::debug( __METHOD__ . ': Method called ', $params );
```
jsutterfield: This is the only place we're doing a `debug` call, all the others are `info`. Is there a reason to have this distinction? Can we just make this an `info` call and get rid of that extra method since it's only being used here?

garthwebb: I don't have a problem including a few extra lines to capture this.  The difference between debug and info is fairly subjective and I could see switching other logging calls in this PR to debug in the future if it turns out they produce a lot of output that is not always needed.

jsutterfield:
LGTM.

PR #6705, opened 2015-03-18
Swagger mercury api

nmonterroso:
@drsnyder @owend 

Add swagger annotations to `nirvana.json` for the Mercury API. This is a requirement for https://wikia-inc.atlassian.net/browse/SERVICES-139, since that needs to access data that's calculated when in the context of a wiki.


includes/wikia/DefaultSettings.php
```diff
@@ -1550,6 +1550,7 @@
 	],
 	'SearchSuggestionsApiController' => ApiAccessService::WIKIA_NON_CORPORATE,
 	'TvApiController' => ApiAccessService::WIKIA_CORPORATE,
+	'MercuryApiController' => ApiAccessService::ALWAYS_HIDDEN,
```
drsnyder: Does this have any practical implications?

drsnyder:
I'm not sure what the constants mean but OTW, LGTM.

includes/wikia/DefaultSettings.php
nmonterroso: Yeah, it's to hide stuff on http://www.wikia.com/api/v1

PR #6710, opened 2015-03-19
CE-1340: Change user for wiki dumps

lukaszkonieczny:
This is a quick fix
https://wikia-inc.atlassian.net/browse/CE-1340

Proper solution will be done in https://wikia-inc.atlassian.net/browse/CE-1495


macbre:
LGTM

PR #6713, opened 2015-03-19
Sanitize user signup behaviour.

ghost:
Starting from UserSignupSpecialController::signup(), all the methods called during the account creation process must meet the same contract and either return boolean false (on failure) or an instance of User (on success) so we can detect and handle errors.

This is necessary for easy integration with Helios.


extensions/wikia/UserLogin/UserLoginForm.class.php
```diff
@@ -52,24 +52,30 @@ function load() {
 		$this->wpUserBirthDay = strtotime( $this->wpBirthYear . '-' . $this->wpBirthMonth . '-' . $this->wpBirthDay );
 	}
 
-	// add new account
+	/**
+	 * Adds a new user account and sends a confirmation email.
+	 *
+	 * @return bool|User an instance of User on success; boolean false otherwise.
```
armonr: I wonder if returning `null` on failure makes the interface a bit simpler; result is still truthy.

extensions/wikia/UserLogin/UserLoginForm.class.php
```diff
@@ -205,6 +211,13 @@ public function initValidationRegsPerEmail() {
 		return $result;
 	}
 
+	/**
+	 * Adds a user account going through all MW built-in checks.
+	 *
+	 * @return bool|User an instance of User on success, boolean false otherwise.
```
armonr: Same with this method: null vs false

armonr:
LGTM. Thanks @michalroszka

lizlux:
LGTM. I checked out and ran this code and things are looking good.

extensions/wikia/UserLogin/UserSignupSpecialController.class.php
```diff
@@ -239,15 +239,19 @@ public function signup() {
...
-		$this->result = ( $signupForm->msgType == 'error' ) ? $signupForm->msgType : 'ok';
-		$this->msg = $signupForm->msg;
-		$this->errParam = $signupForm->errParam;
-
 		// pass and ID of created account for FBConnect feature
 		if ( $ret instanceof User ) {
 			$this->userId = $ret->getId();
 			$this->userPage = $ret->getUserPage()->getFullUrl();
 		}
+
```
garthwebb: We've been using ```setData``` or ```setValues``` whenever we update or create new controllers to make it more explicit that we're passing something to the reponse, e.g.:
```php
$this->response->setValues( [
    'result' => ( $signupForm->msgType == 'error' ) ? $signupForm->msgType : 'ok',
    'msg' => $signupForm->msg,
    'errParam' => $signupForm->errParam,
] );
```

extensions/wikia/UserLogin/UserSignupSpecialController.class.php
```diff
@@ -239,15 +239,19 @@ public function signup() {
 			$ret = $signupForm->addNewAccount();
 		}
 
-		$this->result = ( $signupForm->msgType == 'error' ) ? $signupForm->msgType : 'ok';
-		$this->msg = $signupForm->msg;
-		$this->errParam = $signupForm->errParam;
-
 		// pass and ID of created account for FBConnect feature
```
garthwebb: I don't think lines 243-246 are used by anything anymore.

includes/ExternalUser.php
```diff
@@ -129,20 +129,23 @@ public static function newFromUser( $user ) {
 	/**
 	 * add new User object to database
 	 * 
-	 * @param $user User
-	 * @return mixed User or false
+	 * @param $User User an instance of User to add
```
garthwebb: Not sure if its worth it for this (dying) class, but the variable name and type are out of order (http://www.stack.nl/~dimitri/doxygen/manual/commands.html#cmdparam - see last part of this section for PHP info)

includes/ExternalUser.php
```diff
@@ -191,10 +194,13 @@ protected function initFromCookie() {
 	/**
 	 * Add user to the central database
 	 *
-	 * @param $user object
-	 * @return bool
+	 * @param $User User an instance of User to add
```
garthwebb: param / type ordering again

includes/wikia/ExternalUser_Wikia.php
```diff
@@ -232,49 +232,46 @@ public function mapToUser() {
...
 			$dbw->insert(
 				'`user`',
-				array(
+				[
 					'user_id' => null,
 					'user_name' => $User->mName,
 					'user_password' => $User->mPassword,
 					'user_newpassword' => $User->mNewpassword,
-					'user_newpass_time' => $dbw->timestamp( $User->mNewpassTime ),
+					'user_newpass_time' => $dbw->timestamp($User->mNewpassTime),
```
garthwebb: Actually, our current style guidelines include spaces around function args; here and L262,266

extensions/wikia/UserLogin/UserLoginForm.class.php
michalroszka: It does make sense to return `null` instead of `false` on failure, but it implies considerable refactoring effort as most if not all methods called within the two return `false`. Taking the time pressure into account, `false` seems good enough.

extensions/wikia/UserLogin/UserLoginForm.class.php
michalroszka: Again, it implies considerable refactoring effort as most if not all methods called within the two return `false`. Taking the time pressure into account, `false` seems good enough.

includes/wikia/ExternalUser_Wikia.php
michalroszka: Yeah. I must have missed it.

includes/ExternalUser.php
michalroszka: Fixed in 43c6130.

includes/ExternalUser.php
michalroszka: Fixed in 43c6130.

extensions/wikia/UserLogin/UserSignupSpecialController.class.php
michalroszka: I do not know this code that well. And it is difficult to catch what else might be using that. So I think I will just leave it for now.

PR #6714, opened 2015-03-19
CONCF-227 patch for an error when replacing a video

bognix:
When replacing a video, new file was displayed as an image.

**Root Cause**
When replacing a video method \WikiaLocalFileShared::afterSetProps is called twice. First time when loading metadata for file, second time when upload is made. In afterSetProps method property forceMime is set to false. Because of that when afterSetProps is called for the second time it is not updating mime and media fields for file and it remains stored as BITMAP.

**Patch**
Assumption: when replacing a video article with given title already exists, we only want to update it's content. 
Given the assumption I do following check:
- check if article with given title exists
- if article exists - restore it's forceMime value after calling afterSetProps for the first time

**What could break**
We need to test:
- uploading new video
- uploading video with title which already exists
- replacing video
- video ingestion script for videos which are not on video.wikia.com
- video ingestion script for videos which are already stored on video.wikia.com

Ping @garthwebb @jsutterfield


extensions/wikia/VideoHandlers/VideoFileUploader.class.php
```diff
@@ -159,7 +159,17 @@ public function upload( &$oTitle ) {
 			$file->forceMetadata( serialize( $this->getNormalizedMetadata() ) );
 		}
 
+
+		$forceMime = $file->forceMime;
+
 		$file->getMetadata();
+
+		//In case of video replacement - Title already exists - preserve forceMime value
+		//By default it is changed to false in WikiaLocalFileShared::afterSetProps method
```
jsutterfield: Can you specify that WikiaLocalFileShared::afterSetProps is called by `$file->getMetadata();` above? At first glance it looks like we're just reassigning `$file->forceMime` to itself.

jsutterfield:
One minor doc comment, but otherwise LGTM. Great job on the PR description :)

hakubo:
LGTM

tomasznapieralski:
LGTM

PR #6716, opened 2015-03-19
CE-1425 Track first page views of power users

adamkarminski:
[See CE-1425](https://wikia-inc.atlassian.net/browse/CE-1425)

We need to know more about our Power Users. One of the questions we need an answer to is: "Where the heck do they even start?!" Do they use Special:RecentChanges? Do they use Special:WikiActivity? Or maybe jump right into favorite articles? Let's find out!

The code works like this:
- If a PowerUser visits us - there is a JS tracking asset added to the output
- If there is no cookie that counts their page views - it sets one up with a value '1'
- If the cookie's value is lower than a maximum count (5 at the moment) - it launches an internal tracking mechanism
- The cookie exists for `time() + 3 hours` and is renewed with every page view. However, if a user is not active for _3h+_ it will treat his visit as a 'new session' and set up the cookie and launch the tracking.

### DIFF
--- a/extensions/wikia/PowerUser/js/PowerUserPageViewsTracking.js
+++ b/extensions/wikia/PowerUser/js/PowerUserPageViewsTracking.js
@@ -1,68 +1,63 @@
 require(
 	[
-		'jquery',
 		'wikia.window',
 		'wikia.cookies',
 		'wikia.tracker'
 	],
-	function ($, w, c, tracker) {
+	function (w, c, tracker) {
+		'use strict';
+
 		var maxCounter = 5,
-			cookieName = 'puPageViews';
+			cookieName = 'puspvs',
+			cookieDefaultValue = 1,
+			/**
+			 * If a user is not active for 3 hours
+			 * we treat his return as a new session
+			 * @type {number}
+			 */
+			cookieExpires = 1000 * 60 * 60 * 3;
 
 		function init() {
-			if (pvIsCookieSet() && !pvRemoveCookie()) {
-				pvLaunchTracking();
+			if (!pvIsCookieSet()) {
+				pvSetCookieCounter(cookieDefaultValue);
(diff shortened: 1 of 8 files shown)


includes/wikia/traits/PowerUserTrait.php
```diff
@@ -10,7 +10,7 @@
 
 trait PowerUserTrait {
 	// Required base-class methods
-	abstract function getBoolOption();
+	abstract function getBoolOption( $oname );
```
gabrys: sName + docs?

extensions/wikia/PowerUser/js/PowerUserPageViewsTracking.js
```diff
@@ -0,0 +1,65 @@
...
+			cookieExpires = 1000 * 60 * 60 * 3;
+
+		function init() {
+			if (!pvIsCookieSet()) {
+				pvSetCookieCounter(cookieDefaultValue);
+			}
+
+			if (pvGetCookieCounter() <= maxCounter) {
+				pvLaunchTracking();
+				pvIncreaseCookieCounter();
```
gabrys: increase always?

adamkarminski: D'oh! True.

includes/wikia/traits/PowerUserTrait.php
macbre: `public function getBoolOption( $oname )` is how this method is defined in `User` class. Let's keep consistent :)

adamkarminski:
unitTests_js_branches have been (rerun with a success)[http://qa-s2.wikia-prod:8080/job/unitTests_js_branches/10062/testReport/]

gabrys:
LGTM

PR #6717, opened 2015-03-19
PLATFORM-949: remove unnecessary DB pings

macbre:
`updateSpecialPages.php` was pinging all DB connections before issuing a `COMMIT`.

This PR changes the behavior of updateSpecialPages script to:
- run `$dbw->commit` first and check it's status (modified MW core to return the result of a commit)
- if commit fails, try to reconnect

Add debug logging for skipped commits, i.e. when not inside a transaction.

@michalroszka / @drozdo


includes/db/Database.php
```diff
@@ -2976,11 +2976,19 @@ function begin( $fname = 'DatabaseBase::begin' ) {
...
 		if ( $this->mTrxLevel ) {
-			$this->query( 'COMMIT', $fname );
+			// Wikia change - begin
+			$res = $this->query( 'COMMIT', $fname );
 			$this->mTrxLevel = 0;
+			return $res === true;
+			// Wikia change - end
+		}
+		else {
+			wfDebug( __METHOD__ . ": skipped\n" );
```
michalroszka: For the sake of consistency, I would do the same for `rollback`, too.

macbre: Good point, PR updated

michalroszka:
:+1:

drozdo:
BTW please remember that this reconnect have to be on the same database host for commit to succeed

drozdo:
So Im not sure if PingAll is the prope way to go, pinging only the previous host should be ok

macbre:
Merged to the release in a96d4934474c159004ca410475748d65bf8dbefc

PR #6718, opened 2015-03-19
MAIN-4162 | fix for "Call to undefined method WikiaHomePageHelper::getVisualizationWikisData()"

Warkot:


macbre:
LGTM, thanks for the fix!

PR #6723, opened 2015-03-19
CE-1498 Use groups only on local wikias

adamkarminski:
[See CE-1498](https://wikia-inc.atlassian.net/browse/CE-1498)

The Fatal error happened on wikias that are not included in our shared user rights (uncyclopedia, wowwiki etc.)

Ping: @macbre


macbre:
LGTM

PR #6724, opened 2015-03-19
SERVICES-184 Log user_properties lookups for usage analysis

drsnyder:
Log `user_properties` lookups for usage analysis. We want to know which properties are actually being used.

I'm aware that this probably isn't 100% coverage because of direct lookups into `user_properties` but it should give us some data on the "legitimate" usage via the user object.

https://wikia-inc.atlassian.net/browse/SERVICES-184

/cc @nmonterroso @garthwebb

### DIFF
--- a/includes/User.php
+++ b/includes/User.php
@@ -2311,6 +2311,7 @@ class User {
 	 * @see getIntOption()
 	 */
 	public function getOption( $oname, $defaultOverride = null, $ignoreHidden = false ) {
+		\Wikia\Logger\WikiaLogger::instance()->info( 'user_properties lookup' , array( 'name' => $oname ) );
 		global $wgHiddenPrefs;
 		$this->loadOptions();
 


nmonterroso:
LGTM

frankfarmer:
This instantaneously started generating 80 million logs per hour, overwhelming elasticsearch

frankfarmer:
https://kibana.wikia-inc.com/index.html#dashboard/temp/AUxWkoibptbsRG8ZboAz

Compared to yesterday's log volume, these new stubthreshold logs singlehandedly double log volume.

frankfarmer:
reverted in b84ade227499a4b74e617be0e4c3691d9017e13b

Can we reintroduce this sampled at 1%?

frankfarmer:
screenshot 2015-03-26 08 08 25

Now that these values are well known, perhaps we could log only when the value doesn't match one of these?

PR #6725, opened 2015-03-19
allow wikia.com domains in dev env

owend:
Devboxes were returning 404 for any curl/get/post request with a production domain in the Host:  
This allows for more flexible local testing.  

@nmonterroso


nmonterroso:
LGTM

PR #6728, opened 2015-03-20
moved purge functions to separate class, allowing batching of urls

owend:
Refactored celery purge code into separate class (modeled on ScribePurge).
This allows a static list of urls to be built up during the request, fixing the problem I noticed where it was sending multiple purge Tasks on one request.  

@nmonterroso @macbre


includes/cache/wikia/CeleryPurge.class.php
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	static function onRestInPeace() {
+		global $wgCityId;
+
+		if (empty(CeleryPurge::$buckets)) return true;
+
+		// log purges using Kibana (BAC-1317)
+		$context = [
+			'city' => $wgCityId,
```
macbre: `city_id` is reported as a part of default context (`@fields.city_id`), no need to duplicate it :)

owend: Oh, that was copied from the ScribePurge client.  I'll just remove it from both places...

PR #6733, opened 2015-03-20
Bucky: purge the queue before sending resource timing data

macbre:
Avoid `HTTP 756` response code (Too long request string)` (we were sending 4k+ long URLs)  that will skew the RUM metrics - see #6498

@wladekb


macbre:
Backported to the upcoming release in 94f3ff6

PR #6736, opened 2015-03-20
PLATFORM-965: async Recent Changes update task

macbre:
Avoid deleting ~1k rows when saving an edit (that can take up to 200 ms). Queue an async task instead and clean up the recent changes off-line

Backport `RecentChangesUpdateJob` class and  `DatabaseBase::selectFieldValues` from MediaWiki 1.25.

@michalroszka / @nmonterroso

### DIFF
--- a/includes/db/Database.php
+++ b/includes/db/Database.php
@@ -1220,6 +1220,48 @@ abstract class DatabaseBase implements DatabaseType {
 	}
 
 	/**
+	 * A SELECT wrapper which returns a list of single field values from result rows.
+	 *
+	 * Usually throws a DBQueryError on failure. If errors are explicitly
+	 * ignored, returns false on failure.
+	 *
+	 * If no result rows are returned from the query, false is returned.
+	 *
+	 * @param string|array $table Table name. See DatabaseBase::select() for details.
+	 * @param string $var The field name to select. This must be a valid SQL
+	 *   fragment: do not use unvalidated user input.
+	 * @param string|array $cond The condition array. See DatabaseBase::select() for details.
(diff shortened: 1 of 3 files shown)


includes/db/Database.php
```diff
@@ -1220,6 +1220,48 @@ function selectField( $table, $var, $cond = '', $fname = 'DatabaseBase::selectFi
...
+	 * @param string|array $options The query options. See DatabaseBase::select() for details.
+	 *
+	 * @return bool|array The values from the field, or false on failure
+	 * @since 1.25
+	 */
+	public function selectFieldValues(
+		$table, $var, $cond = '', $fname = __METHOD__, $options = array()
+	) {
+		if ( $var === '*' ) { // sanity
+			throw new DBUnexpectedError( $this, "Cannot use a * field: got '$var'" );
```
michalroszka: The exception's message will resolve to `Cannot use a * field: got '*'` — not very viking, if you asked me. :-)

includes/db/Database.php
```diff
@@ -1220,6 +1220,48 @@ function selectField( $table, $var, $cond = '', $fname = 'DatabaseBase::selectFi
...
+		}
+
+		if ( !is_array( $options ) ) {
```
michalroszka: Have we decided, whether to use `[]` or `array()`?

michalroszka:
Just minor comments. Other than that, looks good to me. :+1:

includes/db/Database.php
macbre: True, true. However this commit is a backport from MW core - I'd keep it as it is.

PR #6740, opened 2015-03-20
SOC-471 updating loadExternalLibrary to respect noexternals

lizlux:
I wanted to make sure external libraries weren't loaded when `noexternals=1`, but also to clean up this code a bit, as $.loadExternalLibrary shouldn't be called directly. 

We may want to split `loadExternalLibrary` out further into an AMD module that can be required by files like loadCaptcha and loadFacebook, but I think this is fine for now.

### DIFF
--- a/resources/wikia/libraries/jquery/loadLibrary/jquery.wikia.loadExternalLibrary.js
+++ b/resources/wikia/libraries/jquery/loadLibrary/jquery.wikia.loadExternalLibrary.js
@@ -3,30 +3,19 @@
 
 	/**
 	 * Generic loading for external JS libraries, including success and failure handling
-	 * Originally adopted from facebook's developer pages but modified for reuse
 	 * @param {string} name Name of the library being loaded (used for logging)
 	 * @param {string} url URL of library file to be loaded
 	 * @param {*} typeCheck Expression to check if library is already loaded
 	 * @returns {jQuery} Returns a jQuery deferred object
 	 */
-	$.loadExternalLibrary = function (name, url, typeCheck) {
-		var scriptTag,
-			firstScriptTag,
-			$deferred = $.Deferred();
+	function loadExternalLibrary(name, url, typeCheck) {
+		var $deferred = $.Deferred();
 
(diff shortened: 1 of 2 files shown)


jsutterfield:
LGTM

resources/wikia/libraries/jquery/loadLibrary/jquery.wikia.loadExternalLibrary.js
```diff
@@ -37,7 +26,34 @@
 		}
 
 		return $deferred;
-	};
+	}
+
+	/**
+	 * Load a library via script tag injection.
+	 * Originally adopted from facebook's developer pages but modified for reuse
+	 * Note that it relies on an arbitrary script tag already being in the DOM (which is always the case in our stack)
```
bkoval: Hey, is there a reason why it needs the already existing script tag to work? Maybe the new script could just be inserted into the document.head of without further querying?

bkoval:
I triggered the build of Jenkins again as the first one failed with no particular reason (socket timeout for PhantomJS). 

Besides one comment LGTM. :jack_o_lantern:

resources/wikia/libraries/jquery/loadLibrary/jquery.wikia.loadExternalLibrary.js
lizlux: This script is based off of Facebook's script loading mechanism, which I think works nicely. Facebook used to recommend using the `<head>` tag but they've since changed it to the first `<script>` tag. TBH, I'm not sure why they made the change other than to offer flexibility for sites that only load scripts at the bottom of the page. [Here's](https://stackoverflow.com/questions/8578617/inject-a-script-tag-with-remote-src-and-wait-for-it-to-execute/8578840#8578840) a Stack Overflow answer about it. I couldn't find any further info into why they made the change. I just figured I'd go along with it.  

TBH I don't have the exact ansewr

lizlux:
JS unit tests are passing locally so the failure is probably similar to what Bart found earlier. I'd say this is okay to merge.

PR #6742, opened 2015-03-21
fixing fatal due to static keyword

lizlux:
Fixing a fatal: `Fatal error: Cannot make non static method WikiaController::getTemplateDir() static in class Email\EmailController in /usr/wikia/source/app/extensions/wikia/Email/EmailController.class.php on line 5`
@garthwebb


garthwebb:
LGTM

PR #6755, opened 2015-03-23
SOC-568 Move default definiton of getTemplateDir up two parents

garthwebb:
This takes care of weird "Service" classes that are not under the WikiaController inheritance tree.

https://wikia-inc.atlassian.net/browse/SOC-568


macbre:
LGTM

PR #6759, opened 2015-03-23
Soc 569 Force showing notification about deleted article instead of a comment

bkoval:
Since we are allowing for only one backend confirmation, sometimes there are conflicts and two methods want to use confirmation at the same time.

This pull request adds a 'force' flag to addConfirmation method, so the more important confirmations show instead of the less important ones.

It's a fix for a P2 listed here: https://wikia-inc.atlassian.net/browse/SOC-569

### DIFF
--- a/extensions/wikia/BannerNotifications/BannerNotificationsController.class.php
+++ b/extensions/wikia/BannerNotifications/BannerNotificationsController.class.php
@@ -108,12 +108,11 @@ class BannerNotificationsController extends WikiaController {
 			self::addConfirmation(
 				wfMessage(
 					'movepage-moved',
-					array('parseinline'),
 					$oldLink,
 					$newLink,
 					$oldText,
 					$newText
-				)->inContentLanguage()->escaped()
+				)->inContentLanguage()->parse()
 			);
 
 			// redirect to page with new title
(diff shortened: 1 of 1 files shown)


jsutterfield:
Should this be pulled into Sprint 9? I don't know what our policy is (whether a ticket we work on always has to be in the current sprint).

jsutterfield:
LGTM

extensions/wikia/BannerNotifications/BannerNotificationsController.class.php
```diff
@@ -24,8 +24,10 @@ public function init() {
 	/**
 	 * Add confirmation message to the user session (so it persists between redirects)
 	 */
-	public static function addConfirmation( $message, $type = self::CONFIRMATION_CONFIRM ) {
-		if ( !empty( $message ) ) {
+	public static function addConfirmation( $message, $type = self::CONFIRMATION_CONFIRM, $force = false ) {
+		//Add confirmation if there was none set yet or if it's forced
+		if ( !empty( $message ) &&
+			( empty( $_SESSION[self::SESSION_KEY] ) || $force === true ) ) {
```
armonr: Minor: this can be simplified a bit more:
```php
		if ( $message && ( empty( $_SESSION[self::SESSION_KEY] ) || $force ) ) {
```

extensions/wikia/BannerNotifications/BannerNotificationsController.class.php
```diff
@@ -24,8 +24,10 @@ public function init() {
 	/**
 	 * Add confirmation message to the user session (so it persists between redirects)
 	 */
```
armonr: It'd be nice to have updated doc here :)

armonr:
Minor comments, but non-blockers, LGTM

PR #6763, opened 2015-03-23
Soc 499 setting up email template tools

lizlux:
This PR sets up the tools we'll need to create HTML templates, including a css inlining tool that takes external css and turns it into inline style tags. 

The templates styles are a proof of concept for now. This PR is really for https://wikia-inc.atlassian.net/browse/SOC-499, since that is the investigative ticket. 

Note that there's one `todo` in this PR. I'm going to fiddle with it more in https://wikia-inc.atlassian.net/browse/SOC-498, and if I can't come up with a good solution I'll create a separate ticket for it.


extensions/wikia/Email/EmailController.class.php
```diff
@@ -175,11 +187,21 @@ protected function getReplyToAddress() {
...
 	 */
 	protected function getBody() {
-		$html = $this->app->renderView(
+		$moduleBody = $this->app->renderView(
 			get_class( $this ),
 			'body',
 			$this->request->getParams()
 		);
+
+		// todo: stop grunt-email-builder from adding these extraneous tags in to the partials
```
lizlux: Opened this issue https://github.com/Automattic/juice/issues/110 to see if it can be addressed at the Juice level.

jsutterfield:
LGTM

extensions/wikia/Email/templates/compiled/forgotPassword.mustache
```diff
@@ -0,0 +1,11 @@
+
+
+
+
```
jsutterfield: What's up with this whitespace here?

extensions/wikia/Email/templates/src/forgotPassword.mustache
```diff
@@ -0,0 +1,16 @@
+<link rel="stylesheet" href="../../styles/forgotPassword.css" />
+
+<style>
+	p {
+		text-decoration: underline;
+		font-weight: bold;
+	}
```
jsutterfield: Should this style be add to forgotPassword.css?

extensions/wikia/Email/EmailController.class.php
```diff
@@ -175,11 +187,18 @@ protected function getReplyToAddress() {
 	 * @return string
 	 */
 	protected function getBody() {
-		$html = $this->app->renderView(
+		$moduleBody = $this->app->renderView(
 			get_class( $this ),
 			'body',
 			$this->request->getParams()
 		);
+
```
jsutterfield: This method looks **much** better :)

extensions/wikia/Email/templates/src/forgotPassword.mustache
lizlux: These are just dummy styles for now to show that we can use `style` tags as well as external style sheets. I can remove it though since we shouldn't ever be adding styles this way.

jsutterfield: cool :+1:

PR #6764, opened 2015-03-24
SOC-460 Send error logging if error when querying for weekly digest

jsutterfield:
Add additional error logging to Kibana. A check will be made in Kibana to see if any errors are being reported and, if so, the Social Team will be notified. This will replaced the failing nagios check which the weekly digest used before. The Kibana check is going to be done by OPs in a separate ticket which can be found in the jira ticket below:

Ticket: https://wikia-inc.atlassian.net/browse/SOC-460

### DIFF
--- a/extensions/wikia/GlobalWatchlist/GlobalWatchlistBot.class.php
+++ b/extensions/wikia/GlobalWatchlist/GlobalWatchlistBot.class.php
@@ -40,7 +40,7 @@ class GlobalWatchlistBot {
 					$userIDs[] = $row->gwa_user_id;
 				} );
 		} catch ( Exception $e ) {
-			WikiaLogger::instance()->info( 'Weekly Digest Error', [
+			WikiaLogger::instance()->error( 'Weekly Digest Error', [
 				'exception' => $e->getMessage(),
 			] );
 		}
--- a/extensions/wikia/GlobalWatchlist/GlobalWatchlistBot.class.php
+++ b/extensions/wikia/GlobalWatchlist/GlobalWatchlistBot.class.php
@@ -24,17 +24,26 @@ class GlobalWatchlistBot {
 	}
 
 	/**
-	 * Return all users in the global_watchlist table
+	 * Return all users in the global_watchlist table. If there's a problem with the query
+	 * (eg, timing out), log the error. We have a Kibana check which will send out an alert
+	 * if any "Weekly Digest Error" messages are sent.
 	 * @return array
 	 */
 	private function getUserIDs() {
 		$db = wfGetDB( DB_SLAVE, [], \F::app()->wg->ExternalDatawareDB );
-		$userIDs = ( new WikiaSQL() )
-			->SELECT()->DISTINCT( GlobalWatchlistTable::COLUMN_USER_ID )
-			->FROM( GlobalWatchlistTable::TABLE_NAME )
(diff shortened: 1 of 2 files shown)


armonr:
Question: this is only one point of failure, that is gathering all users that must be sent digest emails. Do we plan on adding logging for other points that can be blocked on errors? For instance, if either of clear functions, [clearLocalWatchlists](https://github.com/Wikia/app/blob/SOC-460/extensions/wikia/GlobalWatchlist/GlobalWatchlistBot.class.php#L67) or [clearGlobalWatchlistAll](https://github.com/Wikia/app/blob/SOC-460/extensions/wikia/GlobalWatchlist/GlobalWatchlistBot.class.php#L104) fail.

This change LGTM

jsutterfield:
Initially I only thought to add logging to the query which gets all the users since that's what was failing in the first place, but that's a fair point that it wouldn't hurt to catch errors thrown in other parts of the Weekly Digest as well.

I'm going to move this try/catch higher up into the cronjob so that we can catch any error that might get thrown.

extensions/wikia/GlobalWatchlist/GlobalWatchlistBot.class.php
```diff
@@ -79,7 +78,6 @@ private function getWikisWithWatchedPagesForUser( $userID ) {
 			->SELECT()->DISTINCT( GlobalWatchlistTable::COLUMN_CITY_ID )
 			->FROM( GlobalWatchlistTable::TABLE_NAME )
 			->WHERE( GlobalWatchlistTable::COLUMN_USER_ID )->EQUAL_TO( $userID )
-			->AND_( GlobalWatchlistTable::COLUMN_TIMESTAMP )->IS_NOT_NULL()
```
jsutterfield: This query was incorrect, we don't insert rows in the `global_watchlist` table with a null timestamp.

armonr:
:+1:

maintenance/wikia/cronjobs/sendWeeklyDigest.php
```diff
@@ -30,13 +32,27 @@ public function __construct() {
...
 		$message = "sendWeeklyDigest script run at " . date( "F j, Y, g:i a" ) . "\n";
 		$this->output( $message );
 	}
+
+	/**
+	 * If there's a problem when sending the weekly digest (ie, timing out), log the error. We have
+	 * a Kibana check which will send out an alert if any "Weekly Digest Error" messages are sent.
+	 */
+	private function logError( Exception $exception ) {
+		WikiaLogger::instance()->error( 'Weekly Digest Error', [
```
garthwebb: FYI, I just added this as the first query on this Kibana board:

https://kibana.wikia-inc.com/index.html#/dashboard/elasticsearch/Error%20Investigation

I feel like we add a lot of these logging statements and I'm not always sure we follow up.

garthwebb:
LGTM

maintenance/wikia/cronjobs/sendWeeklyDigest.php
jsutterfield: Awesome, thanks @garthwebb. I also included it as part of the Weekly Digest dashboard: https://kibana.wikia-inc.com/#/dashboard/elasticsearch/Weekly%20Digest, but having our error logging in one place will make it easier to catch problems across our different products.

PR #5893, opened 2015-01-07
Ce 1020 exacttarget user wiki edits

kamilkoterba:
Adding maintenance script for incrementally updating user edits number in ExactTarget.
Maintenance script fetches data from DB and updateUsersEdits task sends an incremental update.

https://wikia-inc.atlassian.net/browse/CE-1020

@Wikia/community-engineering, @SQLReview (sql code in comment below)

PR #6767, opened 2015-03-24
Ce 1528 fix user edits updates

kamilkoterba:
https://wikia-inc.atlassian.net/browse/CE-1528
User edits updates data doesn't arrive to ExactTarget,
data consistency for ExactTaget extension is highly important at the moment due to upcoming big sendouts.

Review of previous version of SQL querries
https://github.com/Wikia/app/pull/5893

@Wikia/community-engineering

### DIFF
--- a/extensions/wikia/ExactTargetUpdates/api/ExactTargetApiDataExtension.php
+++ b/extensions/wikia/ExactTargetUpdates/api/ExactTargetApiDataExtension.php
@@ -3,6 +3,8 @@ namespace Wikia\ExactTarget;
 
 class ExactTargetApiDataExtension extends ExactTargetApi {
 
+	const OBJECTS_PER_REQUEST_LIMIT = 2500;
+
 	/**
 	 * An entry point for DataExtension Create requests
 	 * @param  Array  $aApiCallParams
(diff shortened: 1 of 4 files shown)


kamilkoterba:
## QUERY to statsdb

```
SELECT user_id, wiki_id, sum( edits ) + sum( creates ) AS editcount
	 FROM rollup_wiki_user_events
	 WHERE time_id > '2015-03-23 11:50:13'
	 AND period_id = '1'
	 AND user_id <> '0' GROUP BY wiki_id, user_id"
```
time_id condition is always 24h back in time.

EXPLAIN

```
+----+-------------+-------------------------+-------+-----------------+---------+---------+------+--------+----------------------------------------------+
| id | select_type | table                   | type  | possible_keys   | key     | key_len | ref  | rows   | Extra                                        |
+----+-------------+-------------------------+-------+-----------------+---------+---------+------+--------+----------------------------------------------+
|  1 | SIMPLE      | rollup_wiki_user_events | range | PRIMARY,time_id | PRIMARY | 8       | NULL | 107474 | Using where; Using temporary; Using filesort |
+----+-------------+-------------------------+-------+-----------------+---------+---------+------+--------+----------------------------------------------+
```

Worst case scenario is all our users edited on all our wikis in last 24 hours

@SQLReview

lgarczewski:
LGTM from code side.

wladekb:
SQL seems legit

but I have another question: The split of the work is a bit confusing for me. Fetching the list of work to do is done in a maintenance script and processing is done via Job Queue tasks. What is the reason for that flow? Could both stages be done using the same execution mechanism?

kamilkoterba:
All updates are implemented as tasks and I'm using this mechanism here. I could run task directly without queuing it but this way I'm loosing logs in Celery. So to be consistent with a way we run all updates this one is also queued in Job Queue. Also update tasks are not responsible for gathering data, hence it's separate.

wladekb:
lgtm

PR #6773, opened 2015-03-24
User: do not be so futuristic when there is a miss on "user_touched"

macbre:
Improve #6699

@michalroszka


michalroszka:
Let's try.

macbre:
`user_touched` shared memcache keys need to invalidated as they have the current timestamp instead of user's `mTouched` value.

michalroszka:
:+1:

PR #6776, opened 2015-03-24
Fix for making sure we get ResultWrapper from query().

harnash:
In some rare caese Database::query() returs false and it was causing
errors when trying to use reuturned value as an instance of ResultWrapper.
Added some safe checks to detect such cases.

### DIFF
--- a/extensions/wikia/ImageServing/drivers/ImageServingDriverMainNS.class.php
+++ b/extensions/wikia/ImageServing/drivers/ImageServingDriverMainNS.class.php
@@ -137,14 +137,14 @@ class ImageServingDriverMainNS extends ImageServingDriverBase {
 				__METHOD__
 			);
 
-			if ($result instanceof ResultWrapper && $result->numRows() > 0) {
-				foreach ($result as $row) {
+			if ( $result instanceof ResultWrapper && $result->numRows() > 0 ) {
+				foreach ( $result as $row ) {
 					/* @var mixed $row */
-					if ($row->img_height >= $this->minHeight && $row->img_width >= $this->minWidth) {
-						if (!in_array($row->img_minor_mime, self::$mimeTypesBlacklist)) {
+					if ( $row->img_height >= $this->minHeight && $row->img_width >= $this->minWidth ) {
+						if ( !in_array( $row->img_minor_mime, self::$mimeTypesBlacklist ) ) {
 							$imageDetails[$row->img_name] = $row;
 						} else {
-							wfDebug(__METHOD__ . ": {$row->img_name} - filtered out because of {$row->img_minor_mime} minor MIME type\n");
+							wfDebug( __METHOD__ . ": {$row->img_name} - filtered out because of {$row->img_minor_mime} minor MIME type\n" );
 						}
 					}
 				}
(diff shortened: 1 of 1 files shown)


extensions/wikia/ImageServing/drivers/ImageServingDriverMainNS.class.php
```diff
@@ -137,17 +137,20 @@ protected function loadImageDetails( $imageNames = array() ) {
...
-						wfDebug( __METHOD__ . ": {$row->img_name} - filtered out because of {$row->img_minor_mime} minor MIME type\n" );
+			if ($result instanceof ResultWrapper && $result->numRows() > 0) {
+				foreach ($result as $row) {
+					/* @var mixed $row */
+					if ($row->img_height >= $this->minHeight && $row->img_width >= $this->minWidth) {
+						if (!in_array($row->img_minor_mime, self::$mimeTypesBlacklist)) {
+							$imageDetails[$row->img_name] = $row;
+						} else {
+							wfDebug(__METHOD__ . ": {$row->img_name} - filtered out because of {$row->img_minor_mime} minor MIME type\n");
+						}
```
macbre: Your IDE got rid of all whitespaces that do follow MediaWiki / our conventions.

michalroszka:
Does `false` indicate an error? Perhaps we should log it.

macbre:
We actually do via `DBQueryError` exception handling.

michalroszka:
@macbre but for some reason we still end up with `loadImageDetails` causing errors?

extensions/wikia/ImageServing/drivers/ImageServingDriverMainNS.class.php
harnash: Fixed in: 7a3c6506bbf31faf9c9037495e0f37aca689f346

macbre:
LGTM

extensions/wikia/ImageServing/drivers/ImageServingDriverMainNS.class.php
```diff
@@ -137,17 +137,20 @@ protected function loadImageDetails( $imageNames = array() ) {
...
 			);
 
-			foreach ( $result as $row ) {
-				/* @var mixed $row */
-				if ( $row->img_height >= $this->minHeight && $row->img_width >= $this->minWidth ) {
-					if ( !in_array( $row->img_minor_mime, self::$mimeTypesBlacklist ) ) {
-						$imageDetails[$row->img_name] = $row;
-					} else {
-						wfDebug( __METHOD__ . ": {$row->img_name} - filtered out because of {$row->img_minor_mime} minor MIME type\n" );
+			if ( $result instanceof ResultWrapper && $result->numRows() > 0 ) {
```
wladekb: with ->numRows() == 0 we probably still want to call ->free()

wladekb:
->selectField() can also return false when query succeeded but no row was found

extensions/wikia/ImageServing/drivers/ImageServingDriverMainNS.class.php
harnash: Fixed in  8b169f0

PR #6788, opened 2015-03-25
Explicitly set wikiamobile as the skin for WikiaMobileController

rogatty:
https://wikia-inc.atlassian.net/browse/CONCF-222

https://github.com/Wikia/app/blob/dev/extensions/wikia/Blogs/BlogArticle.php#L438 was sometimes getting skin other than `wikiamobile` while called by `WikiaMobileCategoryViewer::WikiaMobileCategoryViewer` hook. This caused PHP warnings as `WikiaMobileCategoryService_getBatch.php` was getting a string instead of an array.

This should never happen and it's hard to find why it does. With @hakubo we've decided it's best to set `wikiamobile` skin explicitly for `WikiaMobileController`.


hakubo:
LGTM

PR #6791, opened 2015-03-25
SOC-525 | Cleanup in CaptchaController

kvas-damian:
https://wikia-inc.atlassian.net/browse/SOC-525

### DIFF
--- a/extensions/wikia/Captcha/CaptchaController.class.php
+++ b/extensions/wikia/Captcha/CaptchaController.class.php
@@ -5,18 +5,12 @@
  */
 class CaptchaController extends WikiaController {
 
-	const DEFAULT_TEMPLATE_ENGINE = WikiaResponse::TEMPLATE_ENGINE_MUSTACHE;
-
 	/**
 	 * Displays a captcha image. This is used exclusively by FancyCaptcha.
 	 */
 	public function showImage() {
 		$fancyCaptcha = new \Captcha\Module\FancyCaptcha();
-		if ( !$fancyCaptcha->showImage() ) {
-			$this->response->setData( [
-				'error' => wfMessage( 'captcha-no-image' )->escaped(),
-			] );
-		}
+		$fancyCaptcha->showImage();
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


extensions/wikia/Captcha/CaptchaController.class.php
```diff
@@ -5,18 +5,12 @@
...
 	 * Displays a captcha image. This is used exclusively by FancyCaptcha.
 	 */
 	public function showImage() {
 		$fancyCaptcha = new \Captcha\Module\FancyCaptcha();
-		if ( !$fancyCaptcha->showImage() ) {
-			$this->response->setData( [
-				'error' => wfMessage( 'captcha-no-image' )->escaped(),
-			] );
-		}
+		$fancyCaptcha->showImage();
```
kvas-damian: We don't need to put additional error message. It's already handled: https://github.com/Wikia/app/blob/dev/extensions/wikia/Captcha/Module/FancyCaptcha.class.php#L260

extensions/wikia/Captcha/CaptchaController.class.php
```diff
@@ -29,11 +23,4 @@ public function getFancyCaptcha() {
 			[ 'form' => $fancyCaptcha->getForm() ]
 		);
 	}
-
-	/**
-	 * Display information about how this captcha works
-	 */
-	public function showHelp() {
```
kvas-damian: I don't think we use this method. When I tried to call it, I got empty page

bkoval:
LGTM

extensions/wikia/Captcha/CaptchaController.class.php
```diff
@@ -5,17 +5,17 @@
...
-		if ( !$fancyCaptcha->showImage() ) {
-			$this->response->setData( [
-				'error' => wfMessage( 'captcha-no-image' )->escaped(),
-			] );
+		if ( $fancyCaptcha->showImage() ) {
+			/*
+			* Stop immediately when captcha is sent to browser.
+			* We don't want to send additional headers and other nirvana stuff
+			*/
+			exit();
```
rafalkalinski: I don't think there are good grounds for exit like that.
I know that this is quick solution for PHP Warning we currently have, but we should at least create a new ticket for fixing/changing the controller to be able not sending headers when it should not.

kvas-damian: https://wikia-inc.atlassian.net/browse/SOC-583

PR #6793, opened 2015-03-26
DAT-2660 Make sure memcache can be used in the service

sebastianmarzjan:
Using GlobalStateWrapper is hardly the most straightforward approach; However, this service is used only by the Indexer, which by itself sets wgAllowMemcacheWrites to false. This is unwanted in this context and the information returned by WikiViews does not need to be perfectly accurate, so it's fine.


jacekjursza:
LGTM

PR #6795, opened 2015-03-26
ADEN-1859: Fix two ads in one slote (Evolve)

nandy-andy:
Evolve is now returning the correct hops. Therefore, we don't need custom logic which is causing now two ads in one slow.

Once I removed the custom logic I started receiving only one ad in a slot in New Zealand and Canada. There is still remaining issue with not hiding slots after Evolve's hop but I'll follow working on it in `dev` branch. From business reasons it's very important to fix the two ads in one slot ASAP.


gabrys:
LGTM

gabrys:
LGTM

PR #6797, opened 2015-03-26
WikiFactoryHub:getWikiCategories: cache even if there are no results

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-1087

`WikiFactoryHub:getWikiCategories` makes 100mm queries daily to shared DB (~20% of all queries). Most of them can and should be cached. However, `WikiaSQL` by default does **not** cache results with empty rows set.

There are lots of memcache misses and just a few sets:

```
site> select bucket, key, sum(sum) from Mediawiki_App_Memcache_Keys_Stats_60m where time > now() - 6h and key =~ /getWikiCategories/ group by bucket;
┌─────────────────┬────────┬────────┐
│       time      │ sum    │ bucket │
├─────────────────┼────────┼────────┤
│ 1/1/70 01:00:00 │ 65993  │ hit    │
│ 1/1/70 01:00:00 │ 133478 │ miss   │
│ 1/1/70 01:00:00 │ 298    │ set    │
└─────────────────┴────────┴────────┘
```

@owend / @wladekb / @drozdo

### DIFF
--- a/extensions/wikia/WikiFactory/Hubs/WikiFactoryHub.php
+++ b/extensions/wikia/WikiFactory/Hubs/WikiFactoryHub.php
@@ -155,7 +155,7 @@ class WikiFactoryHub extends WikiaModel {
 			->JOIN ( "city_cat_mapping" )->USING( "cat_id" )
 			->WHERE( "city_id" )->EQUAL_TO( $city_id )
 			->AND_( "cat_deprecated" )->EQUAL_TO ( 1 )  // always return the "old" category
-			->cache( $this->cache_ttl, wfSharedMemcKey( __METHOD__, $city_id ) )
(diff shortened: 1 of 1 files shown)


extensions/wikia/WikiFactory/Hubs/WikiFactoryHub.php
```diff
@@ -275,6 +275,9 @@ public function getWikiCategories( $city_id, $active = 1 ) {
 			$city_id = 11;
 		}
 
+		// invalidated in clearCache method
```
wladekb: please update the clearCache() method to call WikiaDataAccess::cachePurge()

owend:
This LGTM

michalroszka:
Go on!

PR #6798, opened 2015-03-26
CE-1513 + CE-1506 Create-new-wikia updates

gabrys:
CE-1513 Change "Wiki" to "Wikia" in SITENAME when a new wikia is created
CE-1506 Switch placement of "wiki" in createwiki for VI


lukaszkonieczny:
LGTM

PR #6799, opened 2015-03-26
CONCF-372 | Move WikiaMobile setup to DefaultSettings, fix initialization of i18n in WikiaMobile

vforge:
Ping @igor @hakubo 
See https://github.com/Wikia/config/pull/1024

### DIFF
--- a/includes/wikia/DefaultSettings.php
+++ b/includes/wikia/DefaultSettings.php
@@ -630,6 +630,7 @@ include_once( "$IP/extensions/wikia/Rail/Rail.setup.php" );
  * Includes that are needed for Mercury to work
  */
 
+include_once( "$IP/extensions/wikia/WikiaMobile/WikiaMobile.setup.php" );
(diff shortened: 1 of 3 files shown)


includes/wikia/DefaultSettings.php
```diff
@@ -630,6 +630,7 @@
  * Includes that are needed for Mercury to work
  */
 
+include_once( "$IP/extensions/wikia/WikiaMobile/WikiaMobile.setup.php" );
```
rogatty: Not sure if this should be just below the comment above.

hakubo: yup, sounds right

Warkot: But doesn't Mercury rely on WikiaMobile to remove the inline styles from the markup?

rogatty: @Warkot yes, but WikiaMobile doesn't rely on Mercury ;)

rogatty:
LGTM

mklucsarits:
LGTM

PR #6801, opened 2015-03-26
ADEN-1874: Call evolve only once

nandy-andy:
For a weird reason when Evolve was used for an ad skin the code sent two calls. We don't want it, so this is the simplest way to fix it.


nandy-andy:
We're waiting for decision from Evolve.

PR #6810, opened 2015-03-27
Soc 498 messages

lizlux:
Adding message keys for the watched article emails and the common footer of all emails. @jsutterfield


jsutterfield:
LGTM. The failing unit tests are the result of https://github.com/Wikia/config/pull/1024. Macbre pinged them on that ticket so should hopefully be resolved soon.

PR #6812, opened 2015-03-30
ADEN-1875: removed mobile slots from turtle map

nandy-andy:
Turtle was enabled and tested on all devices. But for now we want it to be filling only desktop skins. Therefore changes below: we removed mobile slots from :turtle:'s map.


gabrys:
LGTM

nandy-andy:
The broken unit tests are connected to our DFS and not to the changes above. There is [a ticket for fixing the DFS](https://wikia-inc.atlassian.net/browse/OPS-6007). After it's closed the test should start passing.

PR #6814, opened 2015-03-30
ADEN-1878 Use post scribe for 71Media

nandy-andy:
We are using GhostWriter currently on german wikias where we should use PostScribe instead. Changes below may not be too elegant but are fixing the issue without touching code in many places.

### DIFF
--- a/extensions/wikia/AdEngine/AdEngine2Hooks.class.php
+++ b/extensions/wikia/AdEngine/AdEngine2Hooks.class.php
@@ -45,7 +45,7 @@ class AdEngine2Hooks {
 		$wgEnableKruxTargeting = !$wgAdEngineDisableLateQueue && !$wgNoExternals && $wgEnableKruxTargeting;
 		$wgEnableKruxOnMobile = $request->getBool( 'enablekrux', $wgEnableKruxOnMobile && !$wgNoExternals );
 
-		// use post scribe with 71Media - check scriptwriter.js:35
+		// use PostScribe with 71Media - check scriptwriter.js:35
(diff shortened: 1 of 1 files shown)


nandy-andy:
Unit tests are failing because of wrong DB configuration :roller_coaster: Not connected to the changes above :japanese_ogre:

extensions/wikia/AdEngine/AdEngine2Hooks.class.php
```diff
@@ -104,6 +109,9 @@ public static function onWikiaSkinTopScripts( &$vars, &$scripts ) {
 		// GA vars
 		$vars['wgGaHasAds'] = isset($adContext['opts']['showAds']);
 
+		// 71Media
+		$vars['wgUsePostScribe'] = $wg->UsePostScribe;
+
```
gabrys: Let's use wgUsePostScribe and wgTitle

extensions/wikia/AdEngine/AdEngine2Hooks.class.php
```diff
@@ -45,6 +45,11 @@ public static function onAfterInitialize( $title, $article, $output, $user, WebR
 		$wgEnableKruxTargeting = !$wgAdEngineDisableLateQueue && !$wgNoExternals && $wgEnableKruxTargeting;
 		$wgEnableKruxOnMobile = $request->getBool( 'enablekrux', $wgEnableKruxOnMobile && !$wgNoExternals );
 
+		// use post scribe with 71Media - check scriptwriter.js:35
```
gabrys: PostScribe

gabrys:
lgtm

PR #6815, opened 2015-03-30
Fix for enabling Universal Analytics on mobile devices.

harnash:
Due to global variable whitelisting on mobile devices wgGaStaging
was not visible. Unified mechanism for detecting production environment.

@macbre: please take a look at this

### DIFF
--- a/extensions/wikia/AnalyticsEngine/AnalyticsProviderGoogleUA.php
+++ b/extensions/wikia/AnalyticsEngine/AnalyticsProviderGoogleUA.php
@@ -30,7 +30,7 @@ class AnalyticsProviderGoogleUA implements iAnalyticsProvider {
 			return true;
 		}
 
-		global $wgGAUserIdSalt;
+		global $wgDevelEnvironment, $wgGAUserIdSalt, $wgStagingEnvironment;
 
 		$app = F::app();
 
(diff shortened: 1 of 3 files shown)


michalroszka:
LGTM, but @gabrys might want to take a look, as well?

extensions/wikia/AnalyticsEngine/AnalyticsProviderGoogleUA.php
```diff
@@ -48,6 +48,12 @@ static public function onWikiaSkinTopScripts( &$vars, &$scripts, $skin ){
 			$vars['wgGAUserIdHash'] = md5($user->getId() . $wgGAUserIdSalt);
 		}
 
+		global $wgDevelEnvironment, $wgStagingEnvironment;
```
nandy-andy: I always prefer having global in one place of function body. But it's my personal opinion :japanese_ogre: I don't think our guidelines cover it - feel free to change it or not :smiley_cat:

gabrys:
LGTM. Agree with @nandy-andy's concern.

extensions/wikia/AnalyticsEngine/AnalyticsProviderGoogleUA.php
harnash: Good point - my mistake. Fixed in 09f44a5424b7c1eba2f2a634a09deb1f042a4692

nandy-andy:
LGTM :)

macbre:
LGTM as well

PR #6816, opened 2015-03-30
CONCF-377 modify scss to change md5 of file and add TTL to sass cache

bognix:
- SASS cache didn't have any TTL. I changed it to 1 day
- One of assets on Wikia Home Page was cached on sandbox-venus and was served from it to production. Modification inside this file should remove wrong version of asset from cache

ping @vforge


vforge:
lgtm

PR #6823, opened 2015-03-30
ADEN-1785 Always call DART clean up

gabrys:


extensions/wikia/AdEngine/js/AdConfigMobile.js
```diff
@@ -32,14 +32,14 @@ define('ext.wikia.adEngine.adConfigMobile', [
...
 
-		if (instantGlobals && instantGlobals.wgSitewideDisableGpt) {
-			return [];
-		}
-
 		if (context.providers.turtle) {
 			return [turtle, remnantGptMobile];
 		}
 
+		if (instantGlobals && instantGlobals.wgSitewideDisableGpt) {
```
nandy-andy: Is there any other reason of moving this `if` block than alphabetical order? :smile_cat:

gabrys: just wgSitewideDisableGpt shouldn't disable turtle

nandy-andy: Oh, that one is more important, I guess ;)

extensions/wikia/AdEngine/AdEngine2Hooks.class.php
```diff
@@ -262,18 +228,18 @@ public static function onSkinAfterContent( &$text ) {
 		return true;
 	}
 
-	public static function onSkinAfterBottomScripts(Skin $skin, &$text) {
+	public static function onSkinAfterBottomScripts( Skin $skin, &$text ) {
 		// TODO: Check whether this works also on Oasis!
-		if ($skin->getSkinName() === 'venus') {
+		if ( $skin->getSkinName() === 'venus' ) {
```
nandy-andy: Is venus still somewhere on production? I heard rumours it was removed.

extensions/wikia/AdEngine/js/run/desktop.run.js
```diff
@@ -0,0 +1,91 @@
...
+
+	// Everything starts after content and JS
+	window.wgAfterContentAndJS.push(function () {
+		// Ads
+		adTracker.measureTime('adengine.init', 'queue.desktop').track();
+		window.adslots2 = window.adslots2 || [];
+		adEngine.run(adConfigDesktop, window.adslots2, 'queue.desktop');
+
+		// Krux
+		krux.load(kruxSiteId);
```
nandy-andy: I think I see why you removed `wikia.log`. If we still want to use it it'll be used only here, to inform us that the code is about to load `krux`. I think, I agree and we don't need `wikia.log` for this - we can see if `krux` was loaded by checking `localStorage`.

extensions/wikia/AdEngine/js/spec/config.desktop.spec.js
```diff
@@ -0,0 +1,188 @@
...
+	});
+
+	it('NZ, Evolve slot: Evolve, Remnant, Liftium', function () {
+		spyOn(mocks.geo, 'getCountryCode').and.returnValue('NZ');
+		spyOn(mocks.providers.evolve, 'canHandleSlot').and.returnValue(true);
+		expect(getProviders('foo')).toEqual('evolve,remnant,liftium');
+	});
+
+	it('NZ, not Evolve slot: Direct, Remnant, Liftium', function () {
+		spyOn(mocks.geo, 'getCountryCode').and.returnValue('NZ');
```
nandy-andy: We don't spy on `mocks.providers.evolve`'s `canHandleSlot()` method because "by default" it's `noop()` call - is it correct assumption? I wonder if it won't be better to explicitly "show" that the results depends mostly on evolve provider's `canHandleSlot()` result. So, my proposal is to add line:
```js
spyOn(mocks.providers.evolve, 'canHandleSlot').and.returnValue(false);
```

gabrys: The idea was to not introduce any lines that wouldn't change the test 
result. Otherwise we would need to spy on all mocks to be 100% explicit.

tests/karma/js-unit.conf.js
```diff
@@ -97,11 +97,12 @@ module.exports = function (config) {
 			'extensions/wikia/AdEngine/js/EventDispatcher.js',
 			'extensions/wikia/AdEngine/js/EvolveHelper.js',
 			'extensions/wikia/AdEngine/js/EvolveSlotConfig.js',
-			'extensions/wikia/AdEngine/js/lookup/*.js',
 			'extensions/wikia/AdEngine/js/MessageListener.js',
 			'extensions/wikia/AdEngine/js/WikiaDartHelper.js',
 			'extensions/wikia/AdEngine/js/WikiaDartVideoHelper.js',
 			'extensions/wikia/AdEngine/js/WikiaGptAdDetect.js',
+			'extensions/wikia/AdEngine/js/config/*.js',
+			'extensions/wikia/AdEngine/js/lookup/*.js',
```
nandy-andy: It's an interesting way of ordering: alphabetical but within directory :japanese_ogre: I think it can stay like that because the whole fire seems to have "sub-directories sections".

nandy-andy:
Quite good clean up, I would say. Good job :+1: 

I didn't find anything big, so I shared my thoughts. The code changes looks good for me. What's left is to merge `origin/dev`, resolve conflicts and update [page parameters documentation](https://one.wikia-inc.com/wiki/List_of_URL_Parameters) since few parameters were removed :sunny:

tests/karma/js-unit.conf.js
gabrys: It is alphabetical: letters A-Z come before a-z, right? ;-)

nandy-andy:
LGTM :baby_chick:

PR #6825, opened 2015-03-30
ignore duplicate key error which happens when removing a group from a us...

owend:
...er twice


macbre:
Backported to dev (with a comment) in ce3a7e8

macbre:
Backported to the current release in 63d7ff4

PR #6826, opened 2015-03-30
Soc 498

lizlux:
Adding and updating templates and styles for the watched page HTML email template. Also updating tools and docs for development flow.


lizlux:
Note, PHP unit tests are failing in Jenkins but passing on my devbox.

extensions/wikia/Email/scripts/Gruntfile.js
```diff
@@ -9,7 +9,7 @@ module.exports = function (grunt) {
 	 * Use Juice to inline css from an external style sheet
 	 * @see https://github.com/Automattic/juice
 	 */
-	grunt.registerTask('inlineCSS', 'some logging', function () {
+	grunt.registerTask('inlineCSS', 'Replaces css selector attributes with style attributes', function () {
```
bkoval: Did you mean 'CSS selector styles' here?

extensions/wikia/Email/styles/common.css
```diff
@@ -0,0 +1,59 @@
+/**
+ * Include this file in all templates to maintain consistant common styles
+ */
+
+* {
+	color: #656e78;
+	font-family: "Helvetica Neue", Arial, sans-serif;
```
bkoval: I know you'll hate me for that, but how about single quotes here? :)

extensions/wikia/Email/styles/main.css
```diff
@@ -1,6 +1,81 @@
...
+.body-wrap {
+	width: 100%;
+}
+
+.main-wrapper {
+	background-color: #E6EBF2;
+}
+
+.main-content {
+	background-color: #FFF;
```
bkoval: Again, you'll hate me for that, but do we have a rule for upper / lower-casing all the hex values for colors? I can see they're a bit mixed here :walking:

extensions/wikia/Email/styles/main.css
```diff
@@ -1,6 +1,81 @@
...
+	border-top: 1px solid #c5ced9;
+	padding: 50px 40px 80px 40px;
+}
+
+.container {
+	display: block !important;
+	max-width: 600px !important;
+	margin: 0 auto !important;
+	clear: both !important;
+}
```
bkoval: Wow, are these '!important' statements essential here (even after inlining)? If so, maybe we could add a comment what do they need to override?

extensions/wikia/Email/styles/main.css
```diff
@@ -1,6 +1,81 @@
...
+
+.head-wrap {
```
bkoval: Maybe we could add a .footer-wrap selector after a comma here?

extensions/wikia/Email/styles/watchedPage.css
```diff
@@ -0,0 +1,21 @@
...
+	font-size: 14px;
+}
+
+.message .subtle {
+	font-size: 12px;
+}
+
+.links {
+	margin-top: 70px;
+}
```
bkoval: Let's add a new line at the end of the file :)

extensions/wikia/Email/templates/compiled/emailPreview.mustache
```diff
@@ -1,4 +1,6 @@
-{{#body}}{{{body}}}{{/body}}{{^body}}
+{{{body}}}
+
```
bkoval: Will this produce two blank lines in html when 'body' var is empty? Can't remember if Mustache trims it, probably it does :)

extensions/wikia/Email/templates/src/emailPreview.mustache
```diff
@@ -1,4 +1,6 @@
-{{#body}}{{{body}}}{{/body}}{{^body}}
+{{{body}}}
+
```
bkoval: Will this produce two blank lines in html when 'body' var is empty? Can't remember if Mustache trims it, probably it does :)

bkoval:
Hey, looks awesome! Left just a couple of comments / questions.

Jenkins is unhappy with some unit tests for WikiaMobile, maybe merging newest dev here would help?
http://qa-s2.wikia-prod:8080/job/unitTests_php_branches/13394/console

Also, do we consider adding SCSS support to the email Front-End, maybe through grunt plugin for compiling it that runs before juice task? Just asking if it makes sense for the future, I strongly believe this is way out of scope for that pull request :)

extensions/wikia/Email/templates/src/emailPreview.mustache
lizlux: There's no user-facing difference, and the empty line makes this code more clear.

lizlux:
@bkoval, I considered adding SCSS support but I'm not sure we need it, and it might complicate things. I was hesitant to add another SCSS processor on top of the one we're using in Oasis, thus introducing two different systems for doing the same thing. We're already doing enough stuff out of the ordinary here. 

That being said, if we decide that development would be sped up by adding SCSS support, it should be straight-forward to do so.

extensions/wikia/Email/styles/main.css
lizlux: I don't hate you, I love coding conventions! It did look wonky with mix-matched cases in one file, so i made them all lower case. We don't have a rule per se but the [guidelines](https://github.com/Wikia/guidelines/blob/master/SCSS/CodingConventions.md#hexformat) use lower case, so I changed it to that.

extensions/wikia/Email/styles/watchedPage.css
lizlux: done

jsutterfield:
LGTM. Merging into SOC-588 so we can start to consolidate our branches.

PR #6828, opened 2015-03-31
SOC-419 adding japan to special use case countries

lizlux:
We'll treat Japan like Canada when it comes to marketing email opt in.

### DIFF
--- a/extensions/wikia/UserLogin/js/MarketingOptIn.js
+++ b/extensions/wikia/UserLogin/js/MarketingOptIn.js
@@ -12,21 +12,21 @@ define('usersignup.marketingOptIn', ['wikia.geo'], function (geo) {
 		var $optInField = wikiaForm.inputs.wpMarketingOptIn,
 			$optInGroup = wikiaForm.getInputGroup('wpMarketingOptIn'),
 			isEurope = geo.getContinentCode() === 'EU',
-			isCanada = geo.getCountryCode() === 'CA';
+			isCanada = geo.getCountryCode() === 'CA',
+			isJapan = geo.getCountryCode() === 'JP';
 
 		if (!$optInField.length || !$optInGroup.length) {
 			throw 'Wikia Form must contain a field called wpMarketingOptIn';
 		}
 
-		if (!isCanada) {
+		if (!isCanada && !isJapan) {
 			$optInField.attr('checked', true);
 		}
 
-		if (isEurope || isCanada) {
+		if (isEurope || isCanada || isJapan) {
 			$optInGroup
 				.add($optInField)
 				.removeClass('hidden');
-
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


extensions/wikia/UserLogin/js/MarketingOptIn.js
```diff
@@ -12,21 +12,21 @@ define('usersignup.marketingOptIn', ['wikia.geo'], function (geo) {
 		var $optInField = wikiaForm.inputs.wpMarketingOptIn,
 			$optInGroup = wikiaForm.getInputGroup('wpMarketingOptIn'),
 			isEurope = geo.getContinentCode() === 'EU',
-			isCanada = geo.getCountryCode() === 'CA';
+			isCanada = geo.getCountryCode() === 'CA',
+			isJapan = geo.getCountryCode() === 'JP';
```
kvas-damian: Instead of creating new variable for new country, what do you think about changing approach to ```enabled``` and ```isChecked```? This way we'll manipulate on array of countries instead of whole method.

jsutterfield:
LGTM. Either approach for solving this is fine. I feel like the current solution is clear and we (hopefully) shouldn't have to add many more countries to this so keeping the variables separate is probably fine. Either works for me though :)

extensions/wikia/UserLogin/js/MarketingOptIn.js
lizlux: I agree the way I did this isn't very sophisticated and wouldn't stand up against adding a bunch of countries and a bunch of rules. I could have mentioned this in the PR description, but we're going to revisit (and ultimately replace) this code with https://wikia-inc.atlassian.net/browse/SOC-560, so I didn't want to spend too much effort optimizing just yet.

extensions/wikia/UserLogin/js/MarketingOptIn.js
```diff
@@ -12,21 +12,21 @@ define('usersignup.marketingOptIn', ['wikia.geo'], function (geo) {
...
 
 		if (!$optInField.length || !$optInGroup.length) {
 			throw 'Wikia Form must contain a field called wpMarketingOptIn';
 		}
 
-		if (!isCanada) {
+		if (!isCanada && !isJapan) {
```
armonr: @lizlux Sorry, I'm late to the party as this ticket passed me by :)
At this point, it'd be nice to have a separate functions for this and for line 26. For instance, I'd suggest `shouldCheckOptIn()`, and `shouldDisplayOptIn()`, for instance.
Having said that, it can wait for now :)

lizlux: Yep, I even started doing exactly that yesterday, then asked myself why, if we're just going to reformat it into a more mercury style later.

PR #6829, opened 2015-03-31
SOC-463 Create a command line interface for testing the Email extension

garthwebb:
https://wikia-inc.atlassian.net/browse/SOC-463


maintenance/wikia/email.php
```diff
@@ -0,0 +1,184 @@
...
+/**
+ * Email
+ *
+ * This sends an email type (defined in extensions/wikia/Email) to the email address of your
+ * choice
+ *
+ */
+
+ini_set( 'display_errors', 'stderr' );
+ini_set( 'error_reporting', E_NOTICE );
```
kvas-damian: Was it intended to change error displaying setting in this file?

maintenance/wikia/email.php
```diff
@@ -0,0 +1,184 @@
...
+			$this->assertControllerExists( $controllerClass );
+
+			$params = $this->argStringToArray( $this->argString );
+			$resp = F::app()->sendRequest( $controllerClass, 'handle', $params, true );
+
+			$this->assertSuccessfulEmail( $resp );
+
+			echo "Email sent as '$this->to' successfully!\n";
+			exit( 0 );
+		} catch ( Exception $e ) {
```
kvas-damian: I know it's command line tool, but when I see statement like this I think only about one thing:
Pokémon Exception Handling
http://blog.codinghorror.com/new-programming-jargon/

This is big block of code and many unexpected exceptions can be thrown. Do you think it's best idea to limit output only to error message? Personally when running CLI script and something unexpected goes wrong I would like to know where exactly problem lays, but maybe I'm strange ;)

maintenance/wikia/email.php
```diff
@@ -0,0 +1,184 @@
...
+	protected $verbose = false;
+	protected $test = false;
+	protected $to = '';
+	protected $type = '';
+	protected $argString = '';
+	protected $list = false;
+
+	public function __construct() {
+		parent::__construct();
+		$this->mDescription = "Pre-populate LVS suggestions";
```
kvas-damian: C&P? ;)

maintenance/wikia/email.php
```diff
@@ -0,0 +1,184 @@
...
+		} catch ( Exception $e ) {
+			echo "An error occurred:\n";
+			echo $e->getMessage() . "\n";
+			exit( 1 );
+		}
+	}
+
+	protected function initOptions() {
+		$this->test = $this->hasOption( 'test' );
+		$this->verbose = $this->hasOption( 'verbose' );
```
kvas-damian: Don't know what others think but we could think of more descriptive variable names for flags, eg. ```isTest```. What do you think?

maintenance/wikia/email.php
```diff
@@ -0,0 +1,184 @@
...
+		if ( !empty( $this->to ) && !empty( $this->type ) ) {
+			return;
+		}
+
+		throw new OptionException( 'Either --list or both --to and --type options are required' );
+	}
+
+	protected function setupUser() {
+		// We either have a user ID or a user name
+		if ( preg_match( '/^\d+$/', $this->to ) ) {
```
kvas-damian: What do you think about using ```filter_var``` instead of regex?

BTW: there is an edge case when username contains only digitis.

maintenance/wikia/email.php
```diff
@@ -0,0 +1,184 @@
...
+
+		$this->addOption( 'to', 'Address to send email to',
+			Maintenance::PARAM_OPTIONAL,
+			Maintenance::PARAM_HAS_ARG
+		);
+		$this->addOption( 'type', 'Type of email to send',
+			Maintenance::PARAM_OPTIONAL,
+			Maintenance::PARAM_HAS_ARG
+		);
+		$this->addOption( 'args', 'Arguments for email type',
```
kvas-damian: It'd nice to inform users what is args format.

maintenance/wikia/email.php
```diff
@@ -0,0 +1,184 @@
...
+
+		F::app()->wg->User = $userObject;
+	}
+
+	protected function argStringToArray( $argString ) {
+		$paramPairs = explode( ',', $argString );
+
+		$param = [];
+		foreach ( $paramPairs as $pair ) {
+			list( $key, $value ) = explode( '=', $pair );
```
kvas-damian: What do you think about input data validation before exploding?

maintenance/wikia/email.php
```diff
@@ -0,0 +1,184 @@
...
+	public function execute() {
+		try {
+			$this->initOptions();
+
+			if ( $this->list ) {
+				$this->showValidEmailTypes();
+				exit( 0 );
+			}
+
+			// The controller class should end with 'Controller'
```
kvas-damian: Maybe we could add info that ```type``` should not ends with ```Controller```?

kvas-damian: Forget about it, I missed ```list``` param ;)

maintenance/wikia/email.php
garthwebb: Yes.  I do everything in my power to make PHP behave like a normal language on the command line, which means outputting errors, all errors, to STDERR rather than sending them to the log or silently eating them.  This STILL fails however, if I miss a semicolon.  Apparently that error is just too difficult to recover from and give any output ... >:(

maintenance/wikia/email.php
garthwebb: Yah, my intent was to only catch the errors I specifically threw.  I updated the exceptions to inherit from a common exception class for this script and will catch only that.

maintenance/wikia/email.php
garthwebb: These are direct mappings to the CLI arguments so I don't really see much value in altering them unless the CLI argument names are unclear and changing both.  In this case, --test and --verbose are pretty standard.

maintenance/wikia/email.php
garthwebb: Looks useful, I'll use ```filter_var``` instead.  As far as the edge case, this will see pretty limited usage so I don't expect that to be a problem, but its easy enough to fix.

maintenance/wikia/email.php
garthwebb: This code doesn't really know much about what its supposed to get, but it could probably check for null ```$pair```s and null ```$value```s

maintenance/wikia/email.php
```diff
@@ -0,0 +1,196 @@
...
+		$paramPairs = explode( ',', $argString );
+
+		$param = [];
+		foreach ( $paramPairs as $pair ) {
+			if ( empty( $pair ) ) {
+				continue;
+			}
+
+			list( $key, $value ) = explode( '=', $pair );
+			if ( empty( $value ) ) {
```
kvas-damian: I'm not sure if empty string as a value should trigger warning, but empty ```$key``` should be handled somehow.

garthwebb: I felt the check on L142 would cover missing keys but I suppose someone could pass 'foo=bar,=apple'

maintenance/wikia/email.php
```diff
@@ -0,0 +1,200 @@
...
+			Maintenance::PARAM_IS_FLAG,
+			't'
+		);
+		$this->addOption( 'verbose', 'Show extra debugging output',
+			Maintenance::PARAM_OPTIONAL,
+			Maintenance::PARAM_IS_FLAG,
+			'v'
+		);
+
+		$this->addOption( 'user', 'Masquerade as this user when sending email. Can be username or ID.s',
```
jsutterfield: Is that trailing "s" a typo? I think this should be `Masquerade as this user when sending email. Can be username or ID`, yea?

garthwebb: @jsutterfield yup, you're correct.

maintenance/wikia/email.php
```diff
@@ -0,0 +1,200 @@
...
+				$this->showValidEmailTypes();
+				exit( 0 );
+			}
+
+			// The controller class should end with 'Controller'
+			$controllerClass = 'Email\\Controller\\' . $this->type . 'Controller';
+			$this->assertControllerExists( $controllerClass );
+
+			$params = $this->argStringToArray( $this->argString );
+			$resp = F::app()->sendRequest( $controllerClass, 'handle', $params, true );
```
jsutterfield: You shouldn't need to pass `true` here for the 4th argument, it defaults to `true` in the function signature.

maintenance/wikia/email.php
```diff
@@ -0,0 +1,200 @@
...
+		return $param;
+	}
+
+	protected function assertControllerExists( $controllerClass ) {
+		if ( !class_exists( $controllerClass ) ) {
+			throw new OptionException( "Can't find $controllerClass to handle this email type\n" );
+		}
+	}
+
+	protected function assertSuccessfulEmail( WikiaResponse $resp = null ) {
```
jsutterfield: Is there a reason this needs a default argument? This is only called in one place and always passes in an argument.

maintenance/wikia/email.php
```diff
@@ -0,0 +1,200 @@
...
+
+		$param = [];
+		foreach ( $paramPairs as $pair ) {
+			if ( empty( $pair ) ) {
+				continue;
+			}
+
+			list( $key, $value ) = explode( '=', $pair );
+			if ( empty( $value ) ) {
+				echo "WARNING: argument '$key' had no value\n";
```
jsutterfield: The `Maintenance` class which is being extended has a method `error` which writes to stderr which might be useful for cases like this. What about switching these to use that method?

maintenance/wikia/email.php
```diff
@@ -0,0 +1,200 @@
...
+	protected function assertRequiredOptionsExist() {
+		if ( $this->list ) {
+			return;
+		}
+
+		if ( !empty( $this->user ) && !empty( $this->type ) ) {
+			return;
+		}
+
+		throw new OptionException( 'Either --list or both --to and --type options are required' );
```
jsutterfield: This exception, as well as the one on L131 are missing a newline.

maintenance/wikia/email.php
garthwebb: That's part of this check; did we even get a response back from the call?  The next line examines this and throws an error if we didn't get anything back.

maintenance/wikia/email.php
garthwebb: Actually, I'm going to take out the newlines on the other throws that include it. The needed newline is added on L86.

jsutterfield:
LGTM

PR #6833, opened 2015-03-31
CONCF-385 check if file has method getTitle and add some logging otherwise

bognix:
check if file has method getTitle and add some logging otherwise

ping @hakubo @garthwebb @jsutterfield


includes/wikia/services/VideoService.class.php
```diff
@@ -49,9 +52,21 @@ public function addVideo( $url ) {
 				$file = wfFindFile( $videoTitle );
 			}
 
-			// Add a default description if available and one doesn't already exist
-			$vHelper = new VideoHandlerHelper();
-			$vHelper->addDefaultVideoDescription( $file );
+			if ( !method_exists( $file, 'getTitle' ) ) {
```
hakubo: you're checking if getTitle method exists but it is not called here anywhere
should we check if file is instance of File class maybe?

hakubo: I would also add type hinting inside: addDefaultVideoDescription so it accepts only File as a parameter

garthwebb: Agreed on the last two notes.  Also, we should probably return an error here rather than letting this pass through.

bognix:
unit tests are failures are not related with this code -> https://wikia-inc.atlassian.net/browse/OPS-6007

hakubo:
LGTM

includes/wikia/services/VideoService.class.php
```diff
@@ -49,9 +52,23 @@ public function addVideo( $url ) {
 				$file = wfFindFile( $videoTitle );
 			}
 
-			// Add a default description if available and one doesn't already exist
-			$vHelper = new VideoHandlerHelper();
-			$vHelper->addDefaultVideoDescription( $file );
+			if ( $file instanceof File )  {
```
Warkot: I might be a n00b, so I'll ask: what is wrong with file being a file object, why there is an error thrown?

bognix: aaa what a mistake! thank you @Warkot !

Warkot: Yay! I'm helping! :smile: :tada:

Warkot:
LGTM :shipit:

PR #6836, opened 2015-03-31
PLATFORM-1065: Fix fatal error in ApiQueryCategoryInfo

wladekb:
Fix for fatal error occuring in ApiQueryCategoryInfo when a category redirect points to another category which matches the query. Due to the way how it is solved internally in ApiPageSet and then iinterpreted in ApiQueryCategoryInfo it causes a call to getDBkey() on a non-object (please refer to inline docs for further info).

https://wikia-inc.atlassian.net/browse/PLATFORM-1065

/cc @macbre


macbre:
LGTM

PR #6837, opened 2015-03-31
PLATFORM-1099: Add logging for fatal error in Wikia::log()

wladekb:
This fatal error is hard to track down without additional information.

https://wikia-inc.atlassian.net/browse/PLATFORM-1099

/cc @macbre


macbre:
LGTM

PR #6838, opened 2015-03-31
PLATFORM-1093: Exit early if data is incorrect in SendGridPostbackV3Controller

wladekb:
Code flow didn't exit when data is empty or non-array. Logging and flow was fixed and should no longer throw warnings.

https://wikia-inc.atlassian.net/browse/PLATFORM-1093

/cc @macbre

### DIFF
--- a/includes/wikia/api/SendGridPostBackApiV3Controller.class.php
+++ b/includes/wikia/api/SendGridPostBackApiV3Controller.class.php
@@ -24,6 +24,13 @@ class SendGridPostbackV3Controller extends WikiaApiController {
 		if (empty($events)) {
 			Wikia::log(__METHOD__, false, "No data to process", true);
 			wfProfileOut( __METHOD__ );
+			return;
+		}
+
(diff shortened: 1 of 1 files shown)


includes/wikia/api/SendGridPostBackApiV3Controller.class.php
```diff
@@ -24,6 +24,13 @@ public function log() {
 		if (empty($events)) {
 			Wikia::log(__METHOD__, false, "No data to process", true);
 			wfProfileOut( __METHOD__ );
+			return;
+		}
+
+		if (!is_array($events)) {
+			Wikia::log(__METHOD__, false, "Supplied data is not an array (JSON begins with): " . @substr($data,0,100), true);
+			wfProfileOut( __METHOD__ );
+			return;
```
macbre: As we're here why not replace ``Wikia::log`` calls with `WikiaLogger` (via `Loggable` trait) and log with context.

PR #6839, opened 2015-03-31
PLATFORM-1064: Add logging for warning in WikiFactory::getWikisByID() and fail early

wladekb:
Warning was produces infrequently in WikiFactory::getWikisByID() when the argument is not a valid array. This change logs the backtrace and returns an empty result early.

https://wikia-inc.atlassian.net/browse/PLATFORM-1064

/cc @macbre

### DIFF
--- a/extensions/wikia/WikiFactory/WikiFactory.php
+++ b/extensions/wikia/WikiFactory/WikiFactory.php
@@ -1262,7 +1262,7 @@ class WikiFactory {
 	 *
 	 * @return array an array of objects, keys are wikis ids.
 	 */
-	static public function getWikisByID( $ids, $master = false ) {
+	static public function getWikisByID( array $ids, $master = false ) {
 		if ( !self::isUsed() ) {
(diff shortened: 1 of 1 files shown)


extensions/wikia/WikiFactory/WikiFactory.php
```diff
@@ -1273,6 +1273,15 @@ static public function getWikisByID( $ids, $master = false ) {
...
 
+		if ( !is_array( $ids ) ) {
+			WikiaLogger::instance()->debug( "WikiFactory::getWikisByID supplied with a non-array argument", [
+				'exception' => new Exception(),
+				'arguments_ids' => $ids,
+			]);
+
+			return [];
+		}
+
```
macbre: Let's fail early and force `Array` as `$ids` argument type.

wladekb: It won't give us the full stack trace when failed. Shall I still change it?

macbre: PHP Catchable Fatal will give us the caller

PR #6841, opened 2015-03-31
CONCF-80: support two endpoints for mercury api purges

owend:
Purge both internal and external cache for mercury api calls. 
@bognix @Warkot


bognix:
LGTM

PR #6842, opened 2015-03-31
Dat 2677

jacekjursza:
MediaWiki conncetor for our microservices.

@pchojnacki @drsnyder @SebastianMarzjan


includes/wikia/services/ConsulService.class.php
```diff
@@ -0,0 +1,134 @@
...
+	 */
+	public function getServiceTag()	{
+		return $this->serviceTag;
+	}
+
+	/**
+	 * @param $serviceTag
+	 * @return $this
+	 */
+	public function setServiceTag( $serviceTag )	{
```
drsnyder: The spacing here looks off. Running the [stylize command](https://github.com/Wikia/guidelines/blob/master/PHP/CodingConventions.md#mediawiki-php-style-helper) should fix it.

includes/wikia/services/ExternalServicesQueryService.class.php
```diff
@@ -0,0 +1,87 @@
...
+			'{query_path}' => $queryPath
+		) );
+		return $url;
+	}
+
+	protected function request( $method, $queryPath, $additionalOptions = [] ) {
+		$url = $this->getUrl( $queryPath );
+		$result = $this->real_request( $method,
+										$url,
+										array_merge( $this->getHttpOptions(), (array)$additionalOptions ) );
```
drsnyder: Same issue here regarding the spacing.

includes/wikia/DefaultSettings.php
```diff
@@ -351,6 +351,8 @@
 $wgAutoloadClasses['FormBuilderService']  =  $IP.'/includes/wikia/services/FormBuilderService.class.php';
 $wgAutoloadClasses['LicensedWikisService']  =  $IP.'/includes/wikia/services/LicensedWikisService.class.php';
 $wgAutoloadClasses['ArticleQualityService'] = $IP.'/includes/wikia/services/ArticleQualityService.php';
+$wgAutoloadClasses['ConsulService'] = $IP.'/includes/wikia/services/ConsulService.class.php';
+$wgAutoloadClasses['ExternalServicesQueryService'] = $IP.'/includes/wikia/services/ExternalServicesQueryService.class.php';
```
drsnyder: I think this code should reside in [`app/lib/Wikia`](https://github.com/Wikia/app/tree/dev/lib/Wikia/src) instead of `app/includes/wikia/services`. That way, it will comply with the [PHP autoloading standards](http://www.php-fig.org/psr/psr-4/) and you won't have to do the manual autoloading.

For the name/ns, I would suggest `Wikia/Util/Consul` or `Wikia/Service/Util/Consul`.

includes/wikia/services/ConsulService.class.php
```diff
@@ -0,0 +1,134 @@
+<?php
+class ConsulService extends Service {
+
+	const DATA_CENTER_SJC = "sjc";
+	const DATA_CENTER_POZ = "poz";
+	const DATA_CENTER_RES = "res";
```
drsnyder: I don't think we should be defining these here. They should be injected via config / the environment. Same with the tags below.

includes/wikia/services/ConsulService.class.php
```diff
@@ -0,0 +1,134 @@
...
+	protected $serviceTag;
+	protected $serviceName;
+
+	public function __construct( $serviceName, $serviceTag=null ) {
+
+		$this->serviceName = $serviceName;
+
+		if ( $serviceTag === null ) {
+			$serviceTag = $this->isDevelopmentEnv() ? self::TAG_TESTING : self::TAG_PRODUCTION;
+		}
```
drsnyder: I wouldn't accept null for the `$serviceTag`. Make this a requirement at object creation. Then this logic and the dependency on `F:app()` below can fall away.

includes/wikia/services/ConsulService.class.php
```diff
@@ -0,0 +1,134 @@
...
+
+		$this->serviceTag = $serviceTag;
+		return $this;
+	}
+
+	/**
+	 * @return bool
+	 */
+	protected function isDevelopmentEnv() {
+		return F::app()->wg->DevelEnvironment === true;
```
drsnyder: Generally speaking, we are trying to eliminate the coupling (adhesion if you will) between libraries in `lib/Wikia` and anything in core mediawiki. It makes them easier to test, easier to reason about, and easier to properly open source.

includes/wikia/services/ConsulService.class.php
```diff
@@ -0,0 +1,134 @@
...
+	/**
+	 * Build consul service host name <{tag}.{service}.service.{dataCenter}.consul>
+	 * @return string
+	 */
+	public function getConsulServiceName() {
+		$name = strtr(self::SERVICE_NAME_SCHEMA, array(
+			'{tag}' => $this->getServiceTag(),
+			'{service}' => $this->getServiceName(),
+			'{dataCenter}' => $this->getDataCenter()
+		));
```
drsnyder: You may want to do some sanitization of these strings to make sure they are valid host names.

includes/wikia/services/ConsulService.class.php
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+
+	/**
+	 * Get all host/port for the service.
+	 * @param $consulServiceHost
+	 * @return array
+	 */
+	public function getResolvedRecords( $consulServiceHost ) {
+		$records = dns_get_record( $consulServiceHost, DNS_SRV );
```
drsnyder: If this returns `false` will we get an error or a warning in the loop below? How can we test this?

includes/wikia/services/ExternalServicesQueryService.class.php
```diff
@@ -0,0 +1,87 @@
...
+	 * @param array $httpOptions
+	 * @return $this
+	 */
+	public function setHttpOptions( $httpOptions )	{
+		$this->httpOptions = $httpOptions;
+		return $this;
+	}
+
+	protected function getUrl( $queryPath ) {
+		$target = $this->consulService->resolve();
```
drsnyder: Will we get errors below if `$target` is false? Should we check that first and exit early? Same below in the caller `$this->request()`.

includes/wikia/services/ExternalServicesQueryService.class.php
```diff
@@ -0,0 +1,87 @@
...
+			$logger = \Wikia\Logger\WikiaLogger::instance();
+			$logger->error( "ExternalServiceQuery fail", [ "serviceName" => $this->consulService->getServiceName(),
+														 	"serviceTag" => $this->consulService->getServiceTag(),
+														 	"url" => $url ] );
+		}
+		return $result;
+	}
+
+	protected function real_request( $method, $url, $options ) {
+		return Http::request( $method, $url, $options );
```
drsnyder: I think we should move away from the MW HTTP library especially in `lib/Wikia`. There are better [3rd party libraries out there](http://docs.guzzlephp.org/en/latest/).

lib/Wikia/src/Util/Consul/ConsulConfig.php
```diff
@@ -0,0 +1,85 @@
+<?php
+namespace Wikia\Util\Consul;
+
+class ConsulConfig {
+
+	protected $urlSchema = "{tag}.{service}.service.{dataCenter}.consul";
```
owend: Tag and data center are optional so perhaps we should construct this with an implode?  In our case, we probably use data center almost everywhere but tag is less frequent.

lib/Wikia/src/Util/Consul/ConsulService.php
```diff
@@ -0,0 +1,62 @@
...
+		$resolved = [ ];
+		if ( is_array( $records ) && !empty( $records ) ) {
+			foreach ( $records as $record ) {
+				$resolved[ ] = [ 'host' => $record[ 'target' ], 'port' => $record[ 'port' ] ];
+			}
+		}
+		return $resolved;
+	}
+
+	protected function doResolve( $consulServiceHost ) {
```
owend: This is only called once, and it's protected, so I'm not sure if we need a helper function here.  Not every PHP core function call needs a wrapper.

lib/Wikia/src/Util/Consul/ExternalServicesQuery.php
```diff
@@ -0,0 +1,40 @@
+<?php
+namespace Wikia\Util\Consul;
+use Wikia\Logger\WikiaLogger;
+
+/**
+ * Class for querying Wikia Microservices provided by Consul
+ * TODO: do actual request functionality
```
owend: What's the plan for this TODO?  Something more like a general REST client?

lib/Wikia/src/Util/Consul/ConsulService.php
drsnyder: Well... testing is difficult if you don't have a way to mock this. You can mock the private method but not the native PHP function.

lib/Wikia/src/Util/Consul/ExternalServicesQuery.php
```diff
@@ -0,0 +1,40 @@
...
+		$target = $this->consulService->resolve();
+
+		$url = strtr( self::URL_SCHEMA, array(
+			'{host}' => $target[ 'host' ],
+			'{port}' => $target[ 'port' ],
+			'{query_path}' => $queryPath ) );
+		return $url;
+	}
+
+	public function logFail( $url ) {
```
SebastianMarzjan: Where are we using this method?

lib/Wikia/src/Util/Consul/ExternalServicesQuery.php
```diff
@@ -0,0 +1,40 @@
...
+		$logger = WikiaLogger::instance();
+		$logger->error( "ExternalServiceQuery fail", [
+			"serviceName" => $this->consulService->getConfig()->getServiceName(),
```
SebastianMarzjan: I'd suggest going with Law of Demeter here and either logging the url only or exposing getServiceName() from consulService

lib/Wikia/src/Util/Consul/ConsulConfig.php
jacekjursza: In current setup i believe we should keep "tag" required, because in other case we get services with "production" tag and "testing" tag.

lib/Wikia/src/Util/Consul/ExternalServicesQuery.php
jacekjursza: We probably need something that will handle the communication between MW and services (logging, retrying, fail handling). As this topic is much broader that our small project - I decided to not implement the functionality at the moment.

SebastianMarzjan:
In the current scope LGTM. It'll need to be expanded in future for sure, but covers what is needed for our use case at the same time allowing for extension.

lib/Wikia/src/Util/Consul/ConsulService.php
```diff
@@ -0,0 +1,63 @@
...
+
+	/**
+	 * Get single host/port for the service.
+	 * @return array ['host'=>, 'port'=>],
+	 */
+	public function resolve() {
+		$consulServiceHost = $this->config->getConsulServiceName();
+		$records = $this->getResolvedRecords( $consulServiceHost );
+		if ( !empty( $records ) ) {
+			// always first, consul is rotating the set
```
drsnyder: Consul might be rotating but I'm not sure about the different caching layers that might be involved. There could be PHP/apache caching, libc, os, and name server. I don't think it's important for this PR but we do want to understand more about the consequences of always taking the first.

drsnyder:
Some of the tests are still failing. OTW, LGTM. Thanks for refactoring!

SebastianMarzjan:
No action for over a quarter, closing since AFAIK this is parallel to work currently happening.

PR #6846, opened 2015-04-01
Soc 498 using PHP plugin instead of JS for CSS inlining. 

lizlux:
We ran into issues with the order in which we were inlining our css into style attributes. The problem was that at run time, wikitext links would be parsed into anchor tags, and those tags wouldn't have inline styles applied. We decided to go with a PHP plugin instead of the JS plugin so we could inline CSS after parsing mustache and wikitext. 

This actually simplifies the process quite a bit, so in the end I think this was a good move in any case. 

To do: 
- [x] Look into going back to external stylesheets as opposed to style tags within templates. The code structure would be cleaner that way. 

I've also made the template for watched article emails more generic, renaming it to `avatarLayout.mustache` so it can be used in multiple email types. 

One more thing - I removed i18n keys for error messages meant only for devs, and replaced them with the english strings.


extensions/wikia/Email/Controller/WatchedPageController.class.php
```diff
@@ -25,7 +25,9 @@ public function getSubject() {
 
 	// TODO setup error handling here if minorEdit not passed in
 	public function initEmail() {
-		$this->title = $this->request->getVal( 'title' );
+		//$this->title = $this->request->getVal( 'title' );
```
bkoval: I guess we don't need it anymore :)

extensions/wikia/Email/templates/main.mustache
```diff
@@ -1,8 +1,82 @@
 <html>
 <head>
 	<title>Email!</title>
-	<link rel="stylesheet" href="../../styles/common.css" />
-	<link rel="stylesheet" href="../../styles/main.css" />
+	<!--<link rel="stylesheet" href="../../styles/common.css" />-->
```
bkoval: Do we need to keep these comments? :)

lib/composer/composer/ClassLoader.php
```diff
@@ -54,9 +54,15 @@ class ClassLoader
     private $useIncludePath = false;
     private $classMap = array();
 
+    private $classMapAuthoritative = false;
```
bkoval: Should this file follow Mediawiki's PHP formatting guidelines?

extensions/wikia/Email/Controller/WatchedPageController.class.php
lizlux: Thanks, yeah, I submitted this in a bit of a hurry last night. I'll clean up the weirdness.

extensions/wikia/Email/Email.i18n.php
```diff
@@ -8,7 +8,6 @@
 
 	// General errors
 	'emailext-error-restricted-controller' => 'Access to this controller is restricted',
-	'emailext-error-noname' => 'Required username has been left empty',
```
jsutterfield: What happened to these?

extensions/wikia/Email/EmailController.class.php
```diff
@@ -34,6 +38,7 @@ public function init() {
 			$this->currentUser = $this->findUserFromRequest( 'currentUser', $this->wg->User );
 			$this->targetUser = $this->findUserFromRequest( 'targetUser', $this->wg->User );
 			$this->test = $this->getRequest()->getVal( 'test', false );
+			$this->marketingFooter = $this->request->getVal( 'marketingFooter' );
```
jsutterfield: If no `marketingFooter` is passed in this will return `null`, we probably want a boolean here yea?

What about making `$marketingFooter = false;` on L17 a class constant like 
```php
const MARKETING_FOOTER_DEFAULT = false
``` 
and changing this line to 
```php
$this->marketingFooter = $this->request->getVal( 'marketingFooter', self::MARKETING_FOOTER_DEFAULT );
```

extensions/wikia/Email/EmailController.class.php
```diff
@@ -189,22 +194,45 @@ protected function getSubject() {
...
 
+		$html = $this->inlineStyles( $html, $css );
+
+		return $html;
+	}
+
+	/**
+	 * Renders the content unique to each email. Should be overridden in child classes.
+	 */
+	protected function getContent() {
```
jsutterfield: I tried changing `getSubject` to an abstract method in my branch and it worked as expected, you can still call it in the base class even though it's the child class that has to implement it.

How about we make `getSubject` and `getContent` abstract methods and make EmailController an abstract class?

extensions/wikia/Email/EmailController.class.php
lizlux: We could, but mustache doesn't care if a value is false, null, 0, undefined, etc. If it's falsey it won't render the conditional content. I'm inclined to leave it as is but if you'd prefer to make typing more explicit I suppose we could update this.

extensions/wikia/Email/templates/main.mustache
```diff
@@ -0,0 +1,96 @@
...
+
+<div class="main-wrapper">
+	<table class="header-wrap">
+		<tr>
+			<td></td>
+			<td>
+				<div class="content">
+					<div class="header content-inner">
+						<a href="http://www.wikia.com" class="logo">
+							<img src="http://vignette3.wikia.nocookie.net/wikianewsletter/images/8/89/Wikia.gif/revision/latest?cb=20150330185243" alt="Wikia">
```
jsutterfield: Should we be constructing these URLs at runtime using the Vignette library? Do you know if there's a danger of these urls ever not working?

extensions/wikia/Email/EmailController.class.php
jsutterfield: That would make me happy :) At the very least because our default value of `$marketingFooter = false;` is lost as soon as we make this `$this->marketingFooter = $this->request->getVal( 'marketingFooter' );` call. I do also like the explicitness.

extensions/wikia/Email/templates/main.mustache
lizlux: We'll leave as is for now and see if the integration tests start giving us any issues.

extensions/wikia/Email/Email.i18n.php
lizlux: There's a few messages inside `EmailController` that are being needlessly i18n'ed, so I started removing them and replacing them with English strings inside the `throw` statements. Didn't want to look at every one, but in general we can start removing these message keys that are meant only for developers to see.

extensions/wikia/Email/EmailController.class.php
lizlux: at one point EmailController was abstract, but we had to change that. It looks like we might not need that change anymore so I'll see if I can make it abstract again.

extensions/wikia/Email/Email.i18n.php
jsutterfield: Ah, Garth, Noelle, and I actually had a conversation about these yesterday since Noelle starting translating them.

Garth actually asked that we do have these translated if it wasn't too much trouble (Noelle said it wouldn't be). We agreed that right now only developers should see them, but just in case there's a use case for bubbling these up in the future it wouldn't hurt to have them translated. So we can leave these as is.

jsutterfield:
LGTM.

PR #6850, opened 2015-04-01
Concf 403

hakubo:
JSMessages called in setup is too early unless I would change logic in DefaultSettings
JSMessages called in init of WikiaMobileService is too late to work

including it in CommonExtensions is way to go, we have everything at this stage to work properly

@rogatty @bognix @vforge @SebastianMarzjan - will that work?


extensions/wikia/WikiaMobile/WikiaMobile.setup.php
```diff
@@ -76,6 +76,28 @@
  */
 $wgExtensionMessagesFiles['WikiaMobile'] = "{$dir}/WikiaMobile.i18n.php";
 
+// initialize i18ns
+JSMessages::registerPackage( 'WkMbl', array(
```
Warkot: I would swap `array()` for `[]` to keep it consistent.

bognix:
LGTM

vforge:
LGTM

tomasznapieralski:
LGTM

PR #6478, opened 2015-02-27
PLATFORM-783: get the list of wikis where the given article exists

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-783

`MultiTaskDelete`, when not provided with the list of wikis, loops on all active wikis and executes the maintenance script.

This pull request changes this behavior. We now get the list of wikis where the given article exists. This will greatly speed up the `MultiTaskDelete` execution and avoid the issue with deploy tools garbage collecting the old checkouts causing long running task to fail to execute maintenance scripts.

``` sql
>select * from pages where page_title_lower = 'rynek_jeżycki' and page_namespace = 0;
+---------------+---------+----------------+------------------+----------------+-------------+-----------------+------------------+------------+-------------+---------------------+-------------------+
| page_wikia_id | page_id | page_namespace | page_title_lower | page_title     | page_status | page_is_content | page_is_redirect | page_edits | page_latest | page_last_edited    | page_last_indexed |
+---------------+---------+----------------+------------------+----------------+-------------+-----------------+------------------+------------+-------------+---------------------+-------------------+
|          5915 |   29445 |              0 | rynek_jeżycki   | Rynek_Jeżycki |           0 |               1 |                0 |          2 |       96155 | 2015-02-26 17:57:16 |              NULL |
+---------------+---------+----------------+------------------+----------------+-------------+-----------------+------------------+------------+-------------+---------------------+-------------------+
1 row in set (0.01 sec)

>explain select * from pages where page_title_lower = 'tomyvilu' and page_namespace = 2;
+----+-------------+-------+------+-----------------------------------------------+------------------------+---------+-------+------+-------------+
| id | select_type | table | type | possible_keys                                 | key                    | key_len | ref   | rows | Extra       |
+----+-------------+-------+------+-----------------------------------------------+------------------------+---------+-------+------+-------------+
|  1 | SIMPLE      | pages | ref  | page_title_lower_wikia,page_title_lower_edits | page_title_lower_wikia | 258     | const |   47 | Using where |
+----+-------------+-------+------+-----------------------------------------------+------------------------+---------+-------+------+-------------+
1 row in set (0.00 sec)
```

``` php
> print_r(Wikia\Tasks\Tasks\MultiTask::getWikisWherePageExists(Title::newFromText('Forum')))
Array
(
    [0] => 15
    [1] => 19
    [2] => 32
    [3] => 38
    [4] => 51
    [5] => 58
    [6] => 71
    [7] => 74
    [8] => 84
[...]

PR #6851, opened 2015-04-01
MultiTask::delete - do not run sitewide if we can not find wikis by provided title

macbre:
http://celery-s1:5555/task/mw-1EB39647-27F3-458C-81ED-546A85E1D0A0

Follow up of #6478

@michalroszka / @wladekb / @harnash / @owend

### DIFF
--- a/includes/wikia/tasks/Tasks/MultiTask.class.php
+++ b/includes/wikia/tasks/Tasks/MultiTask.class.php
@@ -105,6 +105,12 @@ class MultiTask extends BaseTask {
 		// using the provided article's title
 		if ( $params['selwikia'] === 0 ) {
 			$params['selwikia'] = $this->getWikisWherePageExists( $page );
+
+			// PLATFORM-783 / PLATFORM-1106: do not run when the list of wikis is empty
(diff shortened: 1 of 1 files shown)


includes/wikia/tasks/Tasks/MultiTask.class.php
```diff
@@ -214,6 +220,12 @@ private function getWikisWherePageExists( \Title $title ) {
 			$wikis[] = intval( $row->wiki_id );
 		}
 
+		if ( empty( $wikis ) ) {
+			$this->error( 'No wikis were found that have the provided page', [
```
wladekb: Is it an error?
It doesn't seem to be an error for me if I enter a non-existent page name however if we cannot fetch the list of wikis it is the error then. Can we somehow make sure these two cases are reported without any misleading?

macbre: In case of an DB error `DBQueryError` exception will be thrown. I'll change the message level to warning.

wladekb:
lgtm

PR #6852, opened 2015-04-01
ADEN-1880 Load Krux on special pages on wikiamobile

gabrys:


nandy-andy:
Oh, not only didn't we have Krux on special pages but the `window.onload` handler wasn't attached either. The changes LGTM :+1:

PR #6853, opened 2015-04-01
CONCF-370 | Remove Venus traces from GlobalNavigation and introduce support for breakpoints

vforge:
https://wikia-inc.atlassian.net/browse/CONCF-370

@warkot @bognix


extensions/wikia/GlobalNavigation/styles/GlobalNavigationUserLoginDropdown.scss
```diff
@@ -1,28 +1,28 @@
...
+@import 'GlobalNavigationColors';
+@import 'GlobalNavigationVariables';
 
 .AccountNavigation.active .UserLoginDropdown {
 	display: block;
 }
 
+$user-login-content-padding: 19px;
+
 .UserLoginDropdown {
```
Warkot: I remember that those class names where left as is (using CamelCase) to be backwards compatible with some features from the old nav. Can they be changed to dash-case?

hakubo:
LGTM

bognix:
LGTM

bognix:
@vforge please move to QA after addressing @Warkot's comment

extensions/wikia/GlobalNavigation/styles/GlobalNavigationUserLoginDropdown.scss
vforge: True, but it would require some additional changes to (i.e. tracking) and also would require extensive QA. I've decided not to do that this time around - we can create seaparate ticket for that.

Warkot:
BTW, LGTM. :ox:

PR #6861, opened 2015-04-01
SOC-498 installing css inliner via composer

lizlux:
This was done as part of the HTML email work, but in order to simplify our diff, I'm adding this library in a separate PR. 

Ticket is https://wikia-inc.atlassian.net/browse/SOC-499


jsutterfield:
LGTM.

PR #6862, opened 2015-04-02
SOC-558 Update Watched Page edit emails to use new extension and templates

jsutterfield:
This is the first email which we're porting over to use the new Email extension and new email templates. It combines both the backend work of SOC-588 and front-end work of SOC-498.

Included in these tickets are creating a foundational email template (main.mustache and avatarLayour.mustache), including the "CssToInlineStyles" PHP library to inline css for the emails, and adjusting the backend logic so that watched page edits are handled by the new email extension.

Ticket: https://wikia-inc.atlassian.net/browse/SOC-558


includes/UserMailer.php
```diff
@@ -434,95 +433,154 @@ protected static function quotedPrintableCallback( $matches ) {
...
+	* @param string $timestamp : Edit timestamp
+	* @param string $summary : Edit summary
+	* @param bool $minorEdit
+	* @param int $currentRevId : Revision ID
+	* @param int $previousRevId : Revision ID
+	* @param string $action (Wikia)
+	* @param array $otherParam
 	 */
-	protected $editor;
+	public function __construct( $editor, $title, $timestamp, $summary, $minorEdit, $currentRevId = 0, $previousRevId = 0, $action = '', $otherParam = [] ) {
```
armonr: This is a great candidate for the Builder pattern. There may be a slight upfront cost, but the advantages are maintainability, clarity for client and readability of the client usage.
If there are more parameters being added to this, it'll just get more error prone and harder to maintain.

includes/UserMailer.php
```diff
@@ -434,95 +433,154 @@ protected static function quotedPrintableCallback( $matches ) {
...
 		}
 
-		# <Wikia>
-		if( !wfRunHooks( 'AllowNotifyOnPageChange', array( $editor, $title ) ) ) {
-			return false;
-		}
-		
-		if(!empty($otherParam['watchers'])) {
-			$watchers = $otherParam['watchers'];
+		if( !wfRunHooks( 'AllowNotifyOnPageChange', array( $this->editor, $this->title ) ) ) {
```
armonr: `[]` :)

includes/UserMailer.php
```diff
@@ -434,95 +412,154 @@ protected static function quotedPrintableCallback( $matches ) {
...
+			$notificationTimeoutSql = $this->getTimeOutSql();
+			$watchers = $this->getWatchersToNotify( $notificationTimeoutSql );
 			if ( $watchers ) {
-				// Update wl_notificationtimestamp for all watching users except
-				// the editor
-				$wl = WatchedItem::fromUserTitle( $editor, $title );
-				$wl->updateWatch( $watchers, $timestamp );
+				$this->updateWatchedItem( $watchers );
 			}
+			wfRunHooks( 'NotifyOnSubPageChange', array ( $watchers, $this->title, $this->editor, $notificationTimeoutSql ) );
```
armonr: `[]`

includes/UserMailer.php
```diff
@@ -434,95 +412,154 @@ protected static function quotedPrintableCallback( $matches ) {
...
+			$watchers[] = $row->wl_user;
 		}
 
+		return $watchers;
+	}
+
+	/**
+	 * Update wl_notificationtimestamp for all watching users except the editor
+	 */
+	private function updateWatchedItem( $watchers ) {
```
armonr: Is this an array? Please typehint if so, and add to docstring.

includes/UserMailer.php
```diff
@@ -434,95 +412,154 @@ protected static function quotedPrintableCallback( $matches ) {
...
+	 * Check if there are either 1.) users watching the page, or 2.) the page which was
+	 * edited is a User Talk page and the owner of that page wants to receive notifications
+	 * when their User Talk page is changed.
+	 * @param array $watchers
+	 * @return bool
+	 */
+	private function shouldSendEmail( array $watchers ) {
+		global $wgUsersNotifiedOnAllChanges, $wgEnotifMinorEdits, $wgEnotifUserTalk;
+
 		$sendEmail = true;
```
armonr: This method is unnecessarily complicated. It can be simplified to the following:
```php
		global $wgUsersNotifiedOnAllChanges, $wgEnotifMinorEdits, $wgEnotifUserTalk;

		// If nobody is watching the page, and there are no users notified on all changes
		// don't bother creating a job/trying to send emails
		// $watchers deals with $wgEnotifWatchlist
		return ( count( $watchers ) && count( $wgUsersNotifiedOnAllChanges ) ) &&
			// Only send notification for non minor edits, unless $wgEnotifMinorEdits
			( !$this->minorEdit || ( $wgEnotifMinorEdits && !$this->editor->isAllowed( 'nominornewtalk' ) ) ) &&
			( $wgEnotifUserTalk && $this->isUserTalkPage() && $this->canSendUserTalkEmail() );
```
Also note, as the function stands now, the first conditional smells wrong; I suspect, `||` is more appropriate, as in
`if ( !count( $watchers ) || !count( $wgUsersNotifiedOnAllChanges ) ) {`

includes/UserMailer.php
```diff
@@ -550,126 +583,104 @@ public function notifyOnPageChange( $editor, $title, $timestamp, $summary, $mino
...
 		global $wgEnotifWatchlist;
 		global $wgEnotifMinorEdits, $wgEnotifUserTalk;
 
-		wfProfileIn( __METHOD__ );
+		$this->setReplyToAndFromAddresses();
 
 		# The following code is only run, if several conditions are met:
 		# 1. EmailNotification for pages (other than user_talk pages) must be enabled
 		# 2. minor edits (changes) are only regarded if the global flag indicates so
+		if ( !$this->minorEdit || ( $wgEnotifMinorEdits && !$this->editor->isAllowed( 'nominornewtalk' ) ) ) {
```
armonr: This conditional can be encapsulated in a separate method. It can also be used by `shouldSendEmail()`

includes/UserMailer.php
```diff
@@ -550,126 +583,104 @@ public function notifyOnPageChange( $editor, $title, $timestamp, $summary, $mino
...
-		$this->other_param = $otherParam;
-		
-		$userTalkId = false;
-
-		if ( !$minorEdit || ( $wgEnotifMinorEdits && !$editor->isAllowed( 'nominornewtalk' ) ) ) {
-
-			if ( $wgEnotifUserTalk && $isUserTalkPage && $this->canSendUserTalkEmail( $editor, $title, $minorEdit ) ) {
-				$targetUser = User::newFromName( $title->getText() );
+			$userTalkId = 0;
+			if ( $wgEnotifUserTalk && $this->isUserTalkPage() && $this->canSendUserTalkEmail() ) {
```
armonr: Same with this conditional, can go into a function and also used in `shouldSendEmail()`

includes/UserMailer.php
```diff
@@ -550,126 +583,104 @@ public function notifyOnPageChange( $editor, $title, $timestamp, $summary, $mino
...
-					wfDebug( __METHOD__ . ": sending talk page update notification\n" );
-					return true;
-				} else {
-					wfDebug( __METHOD__ . ": talk page owner doesn't have validated email\n" );
-				}
-			} else {
-				wfDebug( __METHOD__ . ": talk page owner doesn't want notifications\n" );
+	private function canSendUserTalkEmail() {
+		$targetUser = User::newFromName( $this->title->getText() );
+		if ( $targetUser instanceof User && // user exists
```
armonr: This is already boolean expression, ready to be returned, no if required! :dollar:

includes/UserMailer.php
```diff
@@ -550,126 +583,104 @@ public function notifyOnPageChange( $editor, $title, $timestamp, $summary, $mino
...
+			!$targetUser->isAnon() && // and they not anonymous
+			$targetUser->getId() != $this->editor->getId() && // and they are not the user who made the edit
+			$targetUser->getOption( 'enotifusertalkpages' ) && // and they want to be notified about talk pages changes
+			( !$this->minorEdit || $targetUser->getOption( 'enotifminoredits' ) ) && // and this is not a minor edit, or they want to know about minor edits
+			$targetUser->isEmailConfirmed() /* and their email is confirmed */ ) {
```
armonr: Minor: move this 2 or 3 lines up for better readability.

includes/UserMailer.php
```diff
@@ -816,28 +793,65 @@ private function composeCommonMailtext() {
...
+				'summary' => $this->summary,
+				'currentRevId' => $this->currentRevId,
+				'previousRevId' => $this->previousRevId,
+				'replyToAddress' => $this->replyto,
+				'fromAddress' => $this->from->address,
+				'fromName' => $this->from->name
+			]);
+	}
+
+	private function sendUsingUserMailer( $user ) {
```
armonr: Typehint

includes/UserMailer.php
```diff
@@ -849,16 +863,16 @@ function sendMails() {
...
 		//     The mail command will not parse this properly while talking with the MTA.
 		$to = new MailAddress( $watchingUser );
 
-		# $PAGEEDITDATE is the time and date of the page change
-		# expressed in terms of individual local time of the notification
-		# recipient, i.e. watching user
+		// $PAGEEDITDATE is the time and date of the page change
+		// expressed in terms of individual local time of the notification
+		// recipient, i.e. watching user
 		$body = str_replace(
```
armonr: This logic is pretty much repeated twice, in a single function. It'd be nice to factor it out as well.

includes/UserMailer.php
```diff
@@ -677,61 +688,48 @@ private function canSendUserTalkEmail( $editor, $title, $minorEdit ) {
...
+		$keys = [];
+		$postTransformKeys = [];
+
+		if ( $this->isNewPage() ) {
+			// watchlist link tracking
+			list ( $keys['$NEWPAGE'], $keys['$NEWPAGEHTML'] ) = wfMsgHTMLwithLanguageAndAlternative (
+				'enotif_lastvisited',
+				'enotif_lastvisited',
+				$wgLanguageCode,
+				array(),
```
armonr: `[]`

extensions/wikia/Email/templates/main.mustache
```diff
@@ -0,0 +1,96 @@
...
+
+<div class="main-wrapper">
+	<table class="header-wrap">
+		<tr>
+			<td></td>
+			<td>
+				<div class="content">
+					<div class="header content-inner">
+						<a href="http://www.wikia.com" class="logo">
+							<img src="http://vignette3.wikia.nocookie.net/wikianewsletter/images/8/89/Wikia.gif/revision/latest?cb=20150330185243" alt="Wikia">
```
armonr: As discussed, it would be preferable to use Vignette lib to do this. This is brittle and can break if vignette's internal structuring changes at some point.

armonr: It can possibly be done in a new ticket

extensions/wikia/Email/Controller/UserLogin.class.php
```diff
@@ -53,15 +61,15 @@ public function body() {
 		$targetUser->saveSettings();
 
 		$this->response->setData( [
-			'greeting' => wfMessage( 'emailext-password-email-greeting', $targetUser->getName() )->inLanguage( $lang )->text(),
-			'content' => wfMessage( 'emailext-password-email-content', $tempPass )->inLanguage( $lang )->text(),
-			'signature' => wfMessage( 'emailext-password-email-signature' )->inLanguage( $lang )->text(),
+			'greeting' => wfMessage( 'emailext-password-email-greeting', $targetUser->getName() )->inLanguage( $this->getTargetLang() )->text(),
+			'content' => wfMessage( 'emailext-password-email-content', $tempPass )->inLanguage( $this->getTargetLang() )->text(),
+			'signature' => wfMessage( 'emailext-password-email-signature' )->inLanguage( $this->getTargetLang() )->text(),
```
armonr: save `$this->getTargetLang()` in a var

extensions/wikia/Email/Email.setup.php
```diff
@@ -27,6 +27,8 @@
 $wgAutoloadClasses['Email\Fatal'] =  $dir . 'EmailExceptions.class.php';
 $wgAutoloadClasses['Email\Check'] =  $dir . 'EmailExceptions.class.php';
 $wgAutoloadClasses['Email\Controller\ForgotPasswordController'] =  $dir . 'Controller/UserLogin.class.php';
+$wgAutoloadClasses['Email\Controller\WatchedPageController'] =  $dir . 'Controller/WatchedPageController.class.php';
+$wgAutoloadClasses['CssToInlineStyles'] = $dir . '../../../../lib/composer/tijsverkoyen/css-to-inline-styles/src/CssToInlineStyles.php';
```
armonr: Autoload should take care of this. No need to explicitly load this

extensions/wikia/Email/templates/main.mustache
lizlux: Ticket is https://wikia-inc.atlassian.net/browse/SOC-622, and it's in this sprint :dancers:

armonr: Great! :+1:

includes/UserMailer.php
jsutterfield: Agreed, it is unnecessarily complicated. The logic you laid out isn't exactly correct though. We want to check if there are any users explicitly watching the page, `count( $watchers ) || count( $wgUsersNotifiedOnAllChanges )`, *or* we want to see if the page is edited is a User Talk page, then do a bunch of checks with regards to that. I'll update this method to see if I can make it more clear.

includes/UserMailer.php
jsutterfield: True, this could be done as a simple return. Considering how many checks this method makes you juggle in your head, I kind of like the explicit `return true;` and `return false`. For me it just slightly  more clear. I'm willing to sacrifice a couple extra lines for that :)

includes/UserMailer.php
jsutterfield: That would be nice. I think I've hit my limit when it comes to refactoring all this bad code though, how about we leave that for the next time we touch this file? The rest of this class is at least 2000% better than it was before.

includes/UserMailer.php
jsutterfield: I couldn't find a method name which I thought accurately and succinctly describe what these 3 checks are doing, so I split them into 3 methods with descriptive names. Hopefully that should make it more clear. I'm open to combining all 3 if you can think of a better way to do it though.

includes/UserMailer.php
armonr: It is much better, indeed :)
Sure, maybe a todo comment will do then.

armonr:
Thanks, @jsutterfield. LGTM

includes/UserMailer.php
jsutterfield: deal!

includes/UserMailer.php
```diff
@@ -222,56 +214,46 @@ public static function send( $to, $from, $subject, $body, $replyto = null, $cont
...
 		$headers['X-Msg-Category'] = $category;
 		if ( $priority ) {
 			$headers['X-Priority'] = $priority;
 		}
-		# </Wikia>
 
-		$ret = wfRunHooks( 'AlternateUserMailer', array( $headers, $to, $from, $subject, $body
-		/*<Wikia>*/
+		$ret = wfRunHooks( 'AlternateUserMailer', [ $headers, $to, $from, $subject, $body
 			, $priority, $attachements
```
garthwebb: As part of cleaning up, might as well put this on the same line as above and eliminate this weird stray leading comma.

extensions/wikia/Email/Controller/WatchedPageController.class.php
```diff
@@ -0,0 +1,215 @@
...
+		] );
+	}
+
+	/**
+	 * @return String
+	 * Get rendered html for content unique to this email
+	 * @todo We may want to make this available more generically for other emails to use the avatar layout.
+	 */
+	protected function getContent() {
+		$css = file_get_contents( __DIR__ . '/../styles/avatarLayout.css' );
```
garthwebb: As discussed, this could be a constant in this class e.g.:
```php
const LAYOUT_CSS = 'avatarLayout.css'
```
with a default constant defined in ```EmailController```.  Then  getContent could move back to the parent class and do:
```php
$css = file_get_contents( __DIR__ . '/styles/' . static::LAYOUT_CSS );
```

extensions/wikia/Email/Controller/WatchedPageController.class.php
```diff
@@ -0,0 +1,215 @@
...
+		if ( $this->currentUser->isLoggedIn() )	 {
+			return $this->currentUser->getName();
+		}
+		return wfMessage( "emailext-watchedpage-anonymous-editor" )->inLanguage( $this->getTargetLang() )->text();
+	}
+
+	/**
+	 * @return String
+	 */
+	private function getEditorAvatarUR() {
```
garthwebb: I think this method name is missing an ending ```L```

garthwebb: or

ur missing a L, lol!

extensions/wikia/Email/Controller/WatchedPageController.class.php
```diff
@@ -0,0 +1,215 @@
...
+	 */
+	private function getCompareChangesLabel() {
+		return wfMessage( 'emailext-watchedpage-diff-button-text' )->inLanguage( $this->getTargetLang() )->text();
+	}
+
+	/**
+	 * @return String
+	 */
+	private function getCompareChangesLink() {
+		return $this->title->getFullUrl( 'diff=' . $this->currentRevId . '&oldid=' . $this->previousRevId );
```
garthwebb: I don't actually have an opinion this but I thought I'd point out that you could do this as well:
```php
return $this->title->getFullUrl( [
    'diff' => $this->currentRevId,
    'oldid' => $this->previousRevId
] );
```

extensions/wikia/Email/Email.i18n.php
```diff
@@ -1,22 +1,10 @@
 <?php
-$messages = [];
+$messages = array();
```
garthwebb: This change smells like crowdin ...

extensions/wikia/Email/EmailController.class.php
```diff
@@ -13,9 +15,15 @@ class EmailController extends \WikiaController {
 
 	protected $hasErrorResponse = false;
 
+	/** @var bool Whether to show the icons over the hubs links */
+	protected $marketingFooter;
+
 	/** @var bool Whether or not to actually send an email */
 	protected $test;
 
+	/**	var string The language to send the email in. Used mostly while testing */
```
garthwebb: Looks like the ```@``` got replaced by a tab character

extensions/wikia/Email/EmailController.class.php
```diff
@@ -135,10 +154,21 @@ protected function setErrorResponse( ControllerException $e ) {
 	}
 
 	protected function getTargetLang() {
+		if ( !empty( $this->targetLang ) ) {
+			return $this->targetLang;
+		}
```
garthwebb: Maybe now that we have ```$this->targetLang```, we should assign it to the default if we get here.  We call ```getTargetLang``` quite a bit.

extensions/wikia/Email/i18n/crowdin-watchedpage.conf
```diff
@@ -0,0 +1,5 @@
+[crowdin.project]
+projectid = "ext-Email-WatchedPage"
+format = "mediawiki"
+source.file = "WatchedPage.i18n.php"
+repo.branch = SOC-558
```
garthwebb: I don't think we should add this.  The only thing this does is allow crowdin to automatically check out the specified branch and commit the changes to it.  If we really used this we'd have to change and commit it every time we updated the file. Better to only download from crowdin, and do the commit and push ourselves.

extensions/wikia/Email/styles/common.css
```diff
@@ -0,0 +1,54 @@
...
+}
+
+table {
+	border-spacing: 0;
+	border-collapse: separate;
+}
+
+button {
+	border-width: 0;
+	background-color: #1a5eB8;
```
garthwebb: Maybe want to lowercase that ```B``` at the end to be consistent.

includes/UserMailer.php
```diff
@@ -434,114 +412,189 @@ protected static function quotedPrintableCallback( $matches ) {
  *
  */
 class EmailNotification {
```
garthwebb: We talked about moving this logic into the Email extension.  Is this a separate ticket?  Does it have a ticket yet?

includes/UserMailer.php
```diff
@@ -434,114 +412,189 @@ protected static function quotedPrintableCallback( $matches ) {
...
-			}
+	/**
+	 * Add a timeout to the watchlist email block
+	 */
+	private function getTimeOutSql() {
+		global $wgEnableWatchlistNotificationTimeout, $wgWatchlistNotificationTimeout;
+
+		if ( !empty( $this->otherParam['notisnull'] ) ) {
+			$notificationTimeoutSql = "1";
+		} elseif ( !empty( $wgEnableWatchlistNotificationTimeout ) && isset( $WatchlistNotificationTimeout ) ) {
```
garthwebb: Looks like we lost a ```wg``` on ```$WatchlistNotificationTimeout```.  Its funny, PHPStorm doesn't flag this for some reason, maybe because its being used within an ```isset```.  Maybe we can upgrade all of these to ```F::app()->wg->```

includes/UserMailer.php
```diff
@@ -434,114 +412,189 @@ protected static function quotedPrintableCallback( $matches ) {
...
 
-		if ( !$sendEmail ) {
-			return;
+		return $notificationTimeoutSql;
+	}
+
+	private function getWatchersToNotify( $notificationTimeoutSql ) {
+		$watchers = [];
+		$dbw = wfGetDB( DB_MASTER );
+		$res = $dbw->select( [ 'watchlist' ],
```
garthwebb: Maybe this and ```__METHOD__``` below could be on their own line like the other params?

includes/UserMailer.php
```diff
@@ -434,114 +412,189 @@ protected static function quotedPrintableCallback( $matches ) {
...
+	private function thereAreUsersWatchingPage( array $watchers ) {
+		return count( $watchers ) || count( \F::app()->wg->UsersNotifiedOnAllChanges );
+	}
+
+	/**
+	 * @return bool
+	 */
+	private function canSendUserTalkEmail() {
+		$targetUser = User::newFromName( $this->title->getText() );
+		if ( !empty( \F::app()->wg->EnotifUserTalk ) && // we want to notify users when they're user talk page is changed
```
garthwebb: Monster of a conditional!  Given this logic is the only thing this function does, I'd consider splitting these out to their own checks, and moving the explanation above them e.g.:
```php
// Should we notify users when their user talk page is changed?
if ( empty( \F::app()->wg->EnotifUserTalk ) ) {
  return false;
}
// The editor doesn't want users to be notified when they make minor edits on discussion pages
if ( $this->editor->isAllowed( 'nominornewtalk' ) ) {
  return false;
}
...
return true;
```
Reversing the logic like this also allows several negations to be removed, which can be easier to read.

includes/UserMailer.php
```diff
@@ -550,203 +603,148 @@ public function notifyOnPageChange( $editor, $title, $timestamp, $summary, $mino
...
-				/* Wikia change end */
 			}
 
 			if ( $wgEnotifWatchlist ) {
 				// Send updates to watchers other than the current editor
 				$userArray = UserArray::newFromIDs( $watchers );
 
+				/* @var $watchingUser User */
 				foreach ( $userArray as $watchingUser ) {
 					if ( $watchingUser->getOption( 'enotifwatchlistpages' ) &&
```
garthwebb: Another round about ```if```.  These guys sure loved length conditionals!  This would be another good candidate for its own function, with each conditional spelled out as above.  However could happen when this logic is moved to the Email extension.  Maybe for now just moving the ```&&``` on line 641 up to the end of the previous line for consistency sake is enough.

includes/UserMailer.php
```diff
@@ -760,53 +758,34 @@ private function composeCommonMailtext() {
 
 		$keys['$PAGEEDITOR_WIKI'] = $this->editor->getUserPage()->getCanonicalUrl();
 
-		# <Wikia>
-		// RT #1294 Bartek 07.05.2009, use the language of the wiki
-		$summary = ($this->summary == '') ? wfMessage( 'enotif_no_summary' )->inContentLanguage()->plain() : '"' . $this->summary . '"';
-		# </Wikia>
-		
-		# Replace this after transforming the message, bug 35019
+		$summary = ( $this->summary == '' ) ? wfMessage( 'enotif_no_summary' )->inContentLanguage()->plain() : '"' . $this->summary . '"';
```
garthwebb: Parens are unnecessary here.  I realize you just added spacing, but its a pet peeve ;)

includes/UserMailer.php
```diff
@@ -760,53 +758,34 @@ private function composeCommonMailtext() {
...
+
+		// Replace this after transforming the message, bug 35019
```
garthwebb: Do you know what this comment means?  That bug number is from when we had RT, two bug systems ago.  If we don't know what this does, we should probably just remove the comment.

includes/UserMailer.php
```diff
@@ -849,61 +865,70 @@ function sendMails() {
...
-				array( wfMessage( 'enotif_impersonal_salutation' )->inContentLanguage()->text(),
-					$wgContLang->date( $this->timestamp, false, false ),
-					$wgContLang->time( $this->timestamp, false, false ) ),
-				$this->body );
+	/**
+	 * Returns whether this Wikia is configured to notify users on minor edits.
+	 * @return bool
+	 */
+	private function notifyUsersOnMinorEdits() {
+		return !empty(\F::app()->wg->EnotifMinorEdits );
```
garthwebb: Space after opening paren of ```empty```

extensions/wikia/Email/Controller/WatchedPageController.class.php
jsutterfield: LOL

image

extensions/wikia/Email/Controller/WatchedPageController.class.php
jsutterfield: I think this second way reads better. I'll update.

extensions/wikia/Email/Email.i18n.php
jsutterfield: Sneaky crowdin-ses, not to be trusted.

garthwebb:
LGTM!

PR #6864, opened 2015-04-02
Platform 387 api post logging

harnash:
Adding some extra parameters to WikiaLogger context when dealing with API calls.

### DIFF
--- a/includes/api/ApiBase.php
+++ b/includes/api/ApiBase.php
@@ -24,6 +24,8 @@
  * @file
  */
 
+use Wikia\Logger\WikiaLogger;
+
 /**
  * This abstract class implements many basic API functions, and is the base of
  * all API classes.
@@ -61,6 +63,11 @@ abstract class ApiBase extends ContextSource {
 
 	private $mMainModule, $mModuleName, $mModulePrefix;
 	private $mParamCache = array();
+
+	/**
+	 * @var array List of parameters being added to log context.
+	 */
+	protected $mLogContextVariables = ['action', 'prop'];
 
 	/**
 	 * Constructor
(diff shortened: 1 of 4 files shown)


includes/api/ApiQuery.php
```diff
@@ -262,10 +262,12 @@ public function execute() {
 			$params = $module->extractRequestParams();
 			$cacheMode = $this->mergeCacheMode(
 				$cacheMode, $module->getCacheMode( $params ) );
+			$module->setupLogContext($params);
 			$module->profileIn();
 			$module->execute();
 			wfRunHooks( 'APIQueryAfterExecute', array( &$module ) );
 			$module->profileOut();
+			$module->destroyLogContext();
```
macbre: Let's mark these changes with ``// WIkia cnange`` as well

harnash: Fixed in 9b4b592139273d73771dd228c381faedc262297e.

macbre:
LGTM

includes/api/ApiBase.php
```diff
@@ -1568,4 +1579,38 @@ public static function debugPrint( $value, $name = 'unknown', $backtrace = false
...
+	 * Sets up the WikiaLogger context for all log entries for current API call.
+	 * Adds extra information about module which handles the call and parameters
+	 * defined in mLogContextVariables array (if passed to a call).
+	 *
+	 * @param $params array Current query parameters
+	 */
+	protected function setupLogContext($params) {
+		$logContext = [];
+
+		foreach ( $params as $paramName => $paramValue) {
```
wladekb: how about iterating over $this->mLogContextVariables and checking if they are defined using eg. array_key_exists() ?

lib/Wikia/src/Logger/WebProcessor.php
```diff
@@ -62,7 +62,13 @@ public function popContext($type=self::RECORD_TYPE_CONTEXT) {
 	 * @return array list of context objects for the given $type
 	 */
 	private function prepareContext($type) {
-		$this->sharedContext[$type] = call_user_func_array('array_merge', array_reverse($this->contextStack[$type]));
+		$stack = array_reverse($this->contextStack[$type]);
```
wladekb: maybe we should use push this reverse logic into pushContext() and popContext() and use:
* [] operator instead of array_unshift()
* array_pop() instead of array_shift()
?

PR #6868, opened 2015-04-03
Soc 558 liz extras

lizlux:
Fixes after sprint demo and code review of watched page emails.


extensions/wikia/Email/EmailController.class.php
```diff
@@ -221,8 +231,13 @@ protected function getBody() {
 	 * @param string $css
 	 * @return string
 	 * @throws \TijsVerkoyen\CssToInlineStyles\Exception
+	 * @see https://github.com/tijsverkoyen/CssToInlineStyles
 	 */
 	protected function inlineStyles( $html, $css ) {
+		// fix known issue in library with character encoding.
+		$charsetMeta = '<meta http-equiv="Content-Type" content="text/html; charset=utf-8" />';
```
jsutterfield: Can we just put this in the template itself?

lizlux: The reason I put it here is b/c I don't want every template to have to include that line, and what if someone forgets to put it in a new template. This ensures it's always there, since I think it would be easy to forget.

jsutterfield:
LGTM.

PR #6870, opened 2015-04-03
CONCF-420 small design fixes in global footer and breakpoints

bognix:


bognix:
ping @Warkot @dianafa

rogatty:
LGTM

dianafa:
LGTM

Warkot:
Are those two variables that you changed (or maybe even more) linked somehow? If yes, maybe it would be a good idea to reflect that relationship in the code with something like `margin: 40px 0 $padding-width-large - 10px 0;`?

bognix:
@Warkot - no they are not. One is the margin below the footer, seconds is spacing on the sides of an article.

Warkot:
Then LGTM. :chicken:

PR #6871, opened 2015-04-03
CONCF-368 Fix encoding of category titles in WikiaMobileCategoryService

rogatty:
https://wikia-inc.atlassian.net/browse/CONCF-368
https://wikia-inc.atlassian.net/browse/CONCF-411

Use `rawurlencode` instead of `urlencode` as we want `%20` and not `%2B`.

http://php.net/manual/en/function.urlencode.php
http://php.net/manual/en/function.rawurlencode.php

/cc @hakubo @bognix @Warkot


hakubo:
LGTM

bognix:
LGTM

PR #6873, opened 2015-04-03
CONCF-386 add logging when Article is null while uploading a video

bognix:
- When uploading a video and video with given title exists article sometimes is "null" instead of Article instance
- when such case happens -> log info to Kibana and throw an exception

ping @hakubo @rogatty @garthwebb

### DIFF
--- a/extensions/wikia/VideoHandlers/VideoFileUploader.class.php
+++ b/extensions/wikia/VideoHandlers/VideoFileUploader.class.php
@@ -137,9 +137,8 @@ class VideoFileUploader {
 			// before generating upload (for now this only works for edits)
 			$articleId = $oTitle->getArticleID();
 			$article = Article::newFromID( $articleId );
-			//In case Article is null
-			//Log more info and throw an exception
-			if ( !( is_null( $article ) ) ) {
+			// In case Article is null log more info and throw an exception
+			if ( is_null( $article ) ) {
 				$exception = new VideoUploadFailedException( 'Video upload failed');
 				WikiaLogger::instance()->error('Video upload: title exists but article is null', [
 					'Title object' => $oTitle,
--- a/extensions/wikia/VideoHandlers/VideoFileUploader.class.php
+++ b/extensions/wikia/VideoHandlers/VideoFileUploader.class.php
@@ -1,6 +1,7 @@
 <?php
 use Wikia\Logger\WikiaLogger;
 
+class VideoUploadFailedException extends Exception {}
 /**
  * Class VideoFileUploader
  */
@@ -135,6 +136,20 @@ class VideoFileUploader {
 			// if video already exists make sure that we are in fact changing something
 			// before generating upload (for now this only works for edits)
 			$article = Article::newFromID( $oTitle->getArticleID() );
+			//In case Article is null
+			//Log more info and throw an exception
+			if ( !( $article instanceof Article ) ) {
+				$articleId = $oTitle->getArticleID();
+				$exception = new VideoUploadFailedException( 'Video upload failed');
+				WikiaLogger::instance()->error('Video upload: title exists but article is null', [
+					'Title object' => $oTitle,
+					'Video title' => $this->getNormalizedDestinationTitle(),
+					'Article ID' => $articleId,
+					'Title from ID' => Title::newFromID($articleId),
+					'exception' => $exception
+				] );
+				throw $exception;
+			}
 			$content = $article->getContent();
 			$newcontent = $this->getDescription();
 			if ( $content != $newcontent ) {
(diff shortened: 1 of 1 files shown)


extensions/wikia/VideoHandlers/VideoFileUploader.class.php
```diff
@@ -135,6 +136,20 @@ public function upload( &$oTitle ) {
 			// if video already exists make sure that we are in fact changing something
 			// before generating upload (for now this only works for edits)
 			$article = Article::newFromID( $oTitle->getArticleID() );
+			//In case Article is null
+			//Log more info and throw an exception
+			if ( !( $article instanceof Article ) ) {
+				$articleId = $oTitle->getArticleID();
+				$exception = new VideoUploadFailedException( 'Video upload failed');
+				WikiaLogger::instance()->error('Video upload: title exists but article is null', [
```
rogatty: `$article` is not `instanceof Article` but not necessarily null. Maybe we should log it's value too?

extensions/wikia/VideoHandlers/VideoFileUploader.class.php
```diff
@@ -135,6 +136,20 @@ public function upload( &$oTitle ) {
 			// if video already exists make sure that we are in fact changing something
 			// before generating upload (for now this only works for edits)
 			$article = Article::newFromID( $oTitle->getArticleID() );
+			//In case Article is null
+			//Log more info and throw an exception
+			if ( !( $article instanceof Article ) ) {
+				$articleId = $oTitle->getArticleID();
```
rogatty: `$oTitle->getArticleID()` is present in line 138, put it in a variable.

extensions/wikia/VideoHandlers/VideoFileUploader.class.php
bognix: We can add this to logs however all errors where happening when article was equal to null.

rogatty: May be, but it's not what the code says :)
`if ( !( $article instanceof Article ) ) { ... }`

bognix: I'll change the if statement

extensions/wikia/VideoHandlers/VideoFileUploader.class.php
```diff
@@ -134,7 +135,21 @@ public function upload( &$oTitle ) {
 			// @TODO
 			// if video already exists make sure that we are in fact changing something
 			// before generating upload (for now this only works for edits)
-			$article = Article::newFromID( $oTitle->getArticleID() );
+			$articleId = $oTitle->getArticleID();
+			$article = Article::newFromID( $articleId );
+			//In case Article is null
+			//Log more info and throw an exception
+			if ( !( is_null( $article ) ) ) {
```
Warkot: This states the opposite to the condition in the comment, shouldn't it be `if ( is_null( $article ) ) {…` (if article is NULL → throw an exception)?

Warkot: I'll just fix it. :wrench:

Warkot:
Now it LGTM. :space_invader:

extensions/wikia/VideoHandlers/VideoFileUploader.class.php
```diff
@@ -134,7 +135,20 @@ public function upload( &$oTitle ) {
...
+				WikiaLogger::instance()->error('Video upload: title exists but article is null', [
+					'Title object' => $oTitle,
+					'Video title' => $this->getNormalizedDestinationTitle(),
+					'Article ID' => $articleId,
+					'Title from ID' => Title::newFromID($articleId),
+					'exception' => $exception
+				] );
+				throw $exception;
+			}
 			$content = $article->getContent();
```
rogatty: We're trying to use method from Article class here. What if `$article` is not null but object other than Article? ;)

Warkot: What I can read from the code, `$article` can only be an `Article` instance or `null`. An additional check with `&& instanceof` could be added like it was before. I'm not sure about that, but well, it's better to be safe than sorry. :water_buffalo:

rogatty: If it's true that
>`$article` can only be an `Article` instance or `null`

then it would be nice to have it in a comment. I wouldn't ask those questions ;)

bognix: There is no point for checking both: is $article an instance of an Article and checking if it's null.
I would go with checking if $article is null because it was causing errors. When $article would be something else than null even though the outcome of an error would the same the root cause of it would be different and it should be logged and investigated on its own.

PR #6874, opened 2015-04-03
Adding unit tests for Banner Notifications

bkoval:
@lizlux @jsutterfield Please, take a look


bkoval:
There was 1 failure:

1) ImageServingCroppingTest::testCropping
<http://vignette.wikia-dev.com/firefly/images/8/89/Wiki-wordmark.png/revision/latest/window-crop/width/50/x-offset/94/y-offset/0/window-width/66/window-height/65?cb=20110304235605> should return HTTP 200
Failed asserting that false is true.

This is a php unit test, looks totally seperate but maybe let's not merge this branch until it gets fixed :)

lizlux:
Looks good! Nice work. That test has been failing on other branches, so I wouldn't hold up this PR because of it.

jsutterfield:
LGTM.

PR #6875, opened 2015-04-03
Concf 414 check browser language and base on this set share networks for user

bognix:


bognix:
ping @Warkot @rogatty 

After CR please move ticket to DONE and merge the branch. It's going to be QAed as part of the whole story (CONCF-189)

extensions/wikia/PageShare/PageShareHelper.class.php
```diff
@@ -17,6 +17,25 @@ public static function getIcon( $service ) {
...
+	 * For logged in user use user's language.
+	 * Both values can be overwritten by ?uselang parameter.
+	 *
+	 * @return String language
+	 */
+	public static function getLangForPageShare() {
+		global $wgLang, $wgRequest, $wgUser;
+
+		if ( $wgUser->isAnon() ) {
+			$lang = substr($_SERVER['HTTP_ACCEPT_LANGUAGE'], 0, 2);
```
rogatty: What if `$_SERVER['HTTP_ACCEPT_LANGUAGE']` is empty? Let's put some default here.

rogatty: Can we use http://php.net/http_negotiate_language? Do we have a list of supported languages?
Here is some example http://stackoverflow.com/questions/6038236/http-accept-language

rogatty:
Fix formatting, please. Missing spaces etc. :see_no_evil:

extensions/wikia/PageShare/PageShareHelper.class.php
bognix: As far as I know we don't have such list. For page share we only need the two-chars language's code so even if we have such list we would have to take substr from the result. Right now Page Share displays all of social networks and includes/excludes some base on language.

extensions/wikia/PageShare/PageShareHelper.class.php
```diff
@@ -17,6 +20,29 @@ public static function getIcon( $service ) {
...
+	 * For logged in user use user's language.
+	 * Both values can be overwritten by ?uselang parameter.
+	 *
+	 * @return String language
+	 */
+	public static function getLangForPageShare() {
+		global $wgLang, $wgRequest, $wgUser;
+
+		if ( $wgUser->isAnon() ) {
+			if ( array_key_exists( 'HTTP_ACCEPT_LANGUAGE', $_SERVER ) ) {
```
rogatty: How about `isset`?

http://stackoverflow.com/questions/3210935/difference-between-isset-and-array-key-exists
>array_key_exists will definitely tell you if a key exists in an array, whereas isset will only return true if the key/variable exists and is not null

Warkot: So it behaves the same way as would `empty()` on a simple variable?

Warkot: I checked on my machine in the command line client (`PHP 5.5.9-1ubuntu4.7`) and what is crucial `isset($arr['key'])` returns `true` even if the `'key'` has a `false`-y value (`''`, `0`, `[]` etc., **but except** `NULL`) as is expected. Whereas `empty($arr['key'])` would return `true` for non-existing `'key'`, `'key'` with empty value and if `'key'` was `NULL`. I'd suggest using `!empty(…)` here.

rogatty:
@bognix I know you're gonna love it! What do you think about writing unit test for `getLangForPageShare` method? Let's confirm our assumptions about `array_key_exists`/`isset`/`empty`

bognix:
@rogatty I loved the idea. Starting the work on it right away :)

nandy-andy:
I enjoy reading you're so positive about writing unit test :+1: Here's something to thank you: :bouquet:

rogatty:
LGTM

Warkot:
LGTM :shipit:

PR #6876, opened 2015-04-03
SOC-603 | Store filtered mOptions in user memcache object

kvas-damian:
https://wikia-inc.atlassian.net/browse/SOC-603

Fix for user account creation.
During account creation `mOptions` are saved to database, but not to memcache. When sending confirmation email new user object is created based on data stored in memcache: https://github.com/Wikia/app/blob/dev/extensions/wikia/UserLogin/UserLoginHelper.class.php#L240
and when saving changes on `User` object `mOptions` are overwritten by empty array in DB.

This finally allows user to login without email confirmation because there is no flag that is required for this check.

It's connected with better memcache caching for `User` merged last week: https://github.com/Wikia/app/commit/5bfcdf701acc7c527c731cef2b5de8e936b2db92

Ping @wladekb @michalroszka

### DIFF
--- a/includes/User.php
+++ b/includes/User.php
@@ -331,7 +331,6 @@ class User {
 			}
 			# /Wikia
 		}
-
 		if ( !$data || $isExpired ) { # Wikia
 			wfDebug( "User: cache miss for user {$this->mId}\n" );
 			# Load from DB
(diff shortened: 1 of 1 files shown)


wladekb:
lgtm

macbre:
Thanks for the fix

PR #6879, opened 2015-04-03
SOC-498 adding messages for social networks

lizlux:
These company names show up in alt tags in the html emails, so we should give them the chance to be translated if appropriate.


armonr:
:

PR #6881, opened 2015-04-03
SOC-621 Start updating FB data

garthwebb:
- Migrate rows with no App ID to a new row with an app ID
- Add a business token to all new rows and update existing on user login

https://wikia-inc.atlassian.net/browse/SOC-621

### DIFF
--- a/extensions/wikia/FacebookClient/FacebookClient.class.php
+++ b/extensions/wikia/FacebookClient/FacebookClient.class.php
@@ -371,11 +371,18 @@ class FacebookClient {
 		}
 	}
 
+	/**
+	 * Update the FacebookMapModel object with the correct business token
+	 *
+	 * @param FacebookMapModel $map
+	 *
+	 * @throws FacebookMapModelDbException
+	 */
 	public function updateBizTokenMapping( FacebookMapModel $map ) {
 		$token = $this->getBizToken();
 
 		if ( $token ) {
-			WikiaLogger::instance()->notice( 'Updating business token', [
+			WikiaLogger::instance()->info( 'Updating business token', [
 				'wikiaUserId' => $map->getWikiaUserId(),
 				'facebookUserId' => $map->getFacebookUserId(),
 				'appId' => $map->getAppId(),
--- a/extensions/wikia/FacebookClient/FacebookClient.class.php
+++ b/extensions/wikia/FacebookClient/FacebookClient.class.php
@@ -281,6 +281,19 @@ class FacebookClient {
 			return null;
 		}
 
+		// Create a new mapping that includes the app ID.  Leave the default App ID behind
+		// as there is no way to tell what other apps this user had connected, and removing this
+		// default mapping will force the user to reconnect on those apps.
+		if ( $map->isDefaultAppId() ) {
+			$this->migrateMapping( $map );
+		}
+
+		// Update the business token for this user if not already set.  The business token
+		// gives us a unique ID across all apps that we can use to reference a user.
+		if ( !$map->getBizToken() ) {
+			$this->updateBizTokenMapping( $map );
+		}
+
 		$wikiUserId = $map->getWikiaUserId();
 		if ( !$wikiUserId ) {
 			return null;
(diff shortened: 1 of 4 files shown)


extensions/wikia/FacebookClient/FacebookClient.class.php
```diff
@@ -281,6 +281,19 @@ public function getWikiaUser( $fbId = null ) {
 			return null;
 		}
```
garthwebb: This process leaves behind mapping rows with App ID == 0.  We can never quite be sure when we can delete those because we don't know what other apps user's if these old style FB IDs might be connected to.  There is logging in ```migrateMapping``` however, and we should see calls to this dwindle to zero over time.  At some point we can decide we've done enough and delete this code and the rows with App ID == 0.  If there are users who still haven't upgraded a row for an app they use, they'll just have to re-connect.

extensions/wikia/FacebookClient/FacebookClient.class.php
```diff
@@ -281,6 +281,19 @@ public function getWikiaUser( $fbId = null ) {
...
 
+		// Create a new mapping that includes the app ID.  Leave the default App ID behind
+		// as there is no way to tell what other apps this user had connected, and removing this
+		// default mapping will force the user to reconnect on those apps.
+		if ( $map->isDefaultAppId() ) {
+			$this->migrateMapping( $map );
+		}
+
```
garthwebb: This process starts updating existing user's Business Token.  This token is not currently used, but tells us which users are the same users across FB Apps now that FB User IDs are unique only within an App.

extensions/wikia/FacebookClient/FacebookClientFactory.php
```diff
@@ -27,7 +27,12 @@ public function connectToFacebook( $wikiaUserId, $fbUserId ) {
 				$status->setResult( false );
 				$status->error( 'fbconnect-error-already-connected' );
 			} else {
```
garthwebb: Start adding the business token whenever we create new mappings.  The App ID is already added automatically (has been since we updated the FB code).  Since its a global variable, its not passed in like the business token is.

extensions/wikia/FacebookClient/FacebookMapModel.class.php
```diff
@@ -127,13 +142,6 @@ public static function getUserMapping( $wikiaUserId, $facebookId ) {
 			throw new FacebookMapModelException( '', self::ERROR_WIKIA_USER_ID_MISMATCH );
 		}
```
garthwebb: This is overly constrained and is the specific fix for this ticket.  Look at this page for reasons why there can be more than one Wikia ID in the mapping table:

https://one.wikia-inc.com/wiki/UConn/Product/Facebook_Connect/Mapping_Changes#Examples

extensions/wikia/FacebookClient/FacebookClient.class.php
```diff
@@ -281,6 +281,19 @@ public function getWikiaUser( $fbId = null ) {
...
+		// Update the business token for this user if not already set.  The business token
+		// gives us a unique ID across all apps that we can use to reference a user.
+		if ( !$map->getBizToken() ) {
+			$this->updateBizTokenMapping( $map );
```
armonr: Should the the exception from `updateBizToken()` bubble up here?

extensions/wikia/FacebookClient/FacebookClient.class.php
```diff
@@ -304,6 +317,81 @@ public function getWikiaUser( $fbId = null ) {
...
+		if ( !$newMapping ) {
+			WikiaLogger::instance()->warning( 'Unable to migrate mapping', [
+				'wikiaUserId' => $map->getWikiaUserId(),
+				'facebookUserId' => $map->getFacebookUserId(),
+				'appId' => $map->getAppId()
+			] );
+		}
+	}
+
+	public function updateBizTokenMapping( FacebookMapModel $map ) {
```
armonr: Please add docstring

extensions/wikia/FacebookClient/FacebookClient.class.php
```diff
@@ -304,6 +317,81 @@ public function getWikiaUser( $fbId = null ) {
...
+		$token = $this->getBizToken();
+
+		if ( $token ) {
+			WikiaLogger::instance()->notice( 'Updating business token', [
```
armonr: Minor: `info()` may be slightly better, but makes very little difference

armonr:
Other than comments/questions above, LGTM

extensions/wikia/FacebookClient/FacebookClient.class.php
garthwebb: Nope, ```FacebookMapModel::updateBizToken``` was copied from that classes ```saveToDatabase```.  It doesn't need to throw an error. It should just log that a problem occurred and move on.  I'll remove that ```throw```.

armonr:
:ship: it

extensions/wikia/FacebookClient/FacebookMapModel.class.php
```diff
@@ -29,6 +29,12 @@ class FacebookMapModel {
 	const ERROR_FACEBOOK_USER_ID_MISMATCH = 1;
 	const ERROR_WIKIA_USER_ID_MISMATCH = 2;
 
+	// This is the App ID given to records created before we started collecting the App ID
+	const APP_ID_DEFAULT_V1 = 0;
```
jsutterfield: Since we have this handy constant now, can we swap out where were just using "0" before? [Here](https://github.com/Wikia/app/blob/dev/extensions/wikia/FacebookClient/FacebookMapModel.class.php#L192), and [here](https://github.com/Wikia/app/blob/dev/extensions/wikia/FacebookClient/FacebookMapModel.class.php#L384).

extensions/wikia/FacebookClient/FacebookClientFactory.php
```diff
@@ -27,7 +27,12 @@ public function connectToFacebook( $wikiaUserId, $fbUserId ) {
...
-				$map = \FacebookMapModel::createUserMapping( $wikiaUserId, $fbUserId );
+				$bizToken = FacebookClient::getInstance()->getBizToken();
```
jsutterfield: Should this have the leading `\` for the namespace?

garthwebb: There's no namespace defined in this class, and none of the classes currently with ```\``` define special namespaces so its superfluous.  Its not part of our style guide but the rest of the file uses it, so I guess its better to keep consistent.

extensions/wikia/FacebookClient/FacebookMapModel.class.php
garthwebb: Yah, that was my intent!  I guess I got distracted...

jsutterfield:
LGTM!

PR #6883, opened 2015-04-06
window-crop negative offsets

nmonterroso:
@drsnyder 
https://wikia-inc.atlassian.net/browse/PLATFORM-1018

The legacy thumbnailer ignored negative values for `x-offset` and `y-offset`, turning the mode into `scale-to-width`. Since Vignette supports negative offsets, this changes `ImageServing` to explicitly request `scale-to-width` when a negative value is calculated.

### DIFF
--- a/extensions/wikia/ImageServing/imageServing.class.php
+++ b/extensions/wikia/ImageServing/imageServing.class.php
@@ -313,14 +313,21 @@ class ImageServing {
 	private function getVignetteUrl(URLGeneratorInterface $image, $width, $height) {
 		list($top, $right, $bottom, $left) = $this->getCutParams($width, $height);
 
-		return $image->getUrlGenerator()
-			->windowCrop()
-			->width($this->width)
-			->xOffset($left)
-			->yOffset($top)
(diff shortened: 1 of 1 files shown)


extensions/wikia/ImageServing/imageServing.class.php
```diff
@@ -313,14 +313,21 @@ public function getUrl( $name, $width = 1, $height = 1 ) {
...
-			->windowCrop()
-			->width($this->width)
-			->xOffset($left)
-			->yOffset($top)
-			->windowWidth($right - $left)
-			->windowHeight($bottom - $top)
-			->url();
+		$generator = $image->getUrlGenerator()
+			->width($this->width);
+
```
drsnyder: Minor: should we comment the failure mode we are trying to prevent here?

drsnyder:
Minor comment, OTW :sheep: :it:

extensions/wikia/ImageServing/imageServing.class.php
nmonterroso: added in a27ecf1

PR #6884, opened 2015-04-06
(CE-1417) Upgrade SemanticMaps to version 2.0.1

Grunny:
Upgrade SemanticMaps to version 2.0.1 which is compatible with SMW 1.8
which we are now running.


Grunny:
Unit test failure is unrelated.

PR #6885, opened 2015-04-06
SOC-586 Error reporter clean up

garthwebb:
https://wikia-inc.atlassian.net/browse/SOC-586

### DIFF
--- a/extensions/wikia/AjaxPoll/AjaxPoll_body.php
+++ b/extensions/wikia/AjaxPoll/AjaxPoll_body.php
@@ -20,14 +20,6 @@ class AjaxPollClass {
 	public static $mCount = 0; # macbre: count ajax polls on the page
 
 	/**
-	 * __construct
-	 */
-	public function __construct() {
-
-	}
-
-
-	/**
 	 * renderFromTag
 	 *
 	 * static constructor/callback function
@@ -36,9 +28,11 @@ class AjaxPollClass {
 	 * @static
 	 * @author Krzysztof Krzyżaniak <eloy@wikia.com>
 	 *
-	 * @param string $input: Text from tag
-	 * @param array $params: atrributions
-	 * @param Object $parser: Wiki Parser object
+	 * @param string $input : Text from tag
+	 * @param array $params : atrributions
+	 * @param Object $parser : Wiki Parser object
+	 *
+	 * @return string
 	 */
 	static public function renderFromTag( $input, $params, $parser ) {
 		/**
@@ -67,7 +61,9 @@ class AjaxPollClass {
 	 * @static
 	 * @author Krzysztof Krzyżaniak <eloy@wikia.com>
 	 *
-	 * @param string $poll_id	poll identifier in database
+	 * @param string $poll_id poll identifier in database
+	 *
+	 * @return AjaxPollClass
 	 */
 	static public function newFromId( $poll_id ) {
 		global $wgTitle, $wgParser;
@@ -136,9 +132,11 @@ class AjaxPollClass {
 	 *
 	 * get HTML for votes
 	 *
-	 * @return array: votes for this poll
+	 * @param bool $isSubmit
+	 *
+	 * @return array : votes for this poll
 	 */
-	public function getVotes($isSubmit = false) {
+	public function getVotes( $isSubmit = false ) {
 		global $wgLang, $wgMemc;
 
 		if ( is_null( $this->mId ) ) {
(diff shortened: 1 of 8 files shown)


extensions/wikia/AjaxPoll/AjaxPoll_body.php
```diff
@@ -20,14 +20,6 @@ class AjaxPollClass {
 	public static $mCount = 0; # macbre: count ajax polls on the page
```
garthwebb: Changes in this class are related to [SOC-410](https://wikia-inc.atlassian.net/browse/SOC-410)

extensions/wikia/AjaxPoll/AjaxPoll_body.php
```diff
@@ -149,54 +147,48 @@ public function getVotes($isSubmit = false) {
...
 		}
-		if (!empty($votes)) {
-			foreach($votes as $ansData){
+		if ( !empty( $votes ) ) {
+			foreach( $votes as $ansData ){
 				$total += $ansData["value"];
 			}
 		} else {
-			$votes = array();
 			$dbr = wfGetDB( $isSubmit ? DB_MASTER : DB_SLAVE );
```
garthwebb: The primary improvement of this change is to do a ```group by``` in SQL rather than pulling all rows back and doing a "code" group by in PHP.

extensions/wikia/AjaxPoll/AjaxPoll_body.php
```diff
@@ -149,54 +147,48 @@ public function getVotes($isSubmit = false) {
...
+				}, [] );
 
 			/**
 			 * count percentage of answers
 			 */
-			foreach( $votes as $nr => $vote ) {
-				$percent = $vote[ "value" ] / $total * 100;
-				$votes[ $nr ][ "percent" ] = round($percent, 2);
-				$votes[ $nr ][ "pixels" ] = $this->percent2pixels( $percent );
-
```
garthwebb: Note, the front end wasn't using this ```title``` key.

extensions/wikia/Blogs/BlogTemplate.php
```diff
@@ -258,6 +258,7 @@ class BlogTemplateClass {
 	private static $aOptions	= array( );
```
garthwebb: The changes in this file are for [SOC-380](https://wikia-inc.atlassian.net/browse/SOC-380)

extensions/wikia/Blogs/BlogTemplate.php
```diff
@@ -265,7 +266,7 @@ class BlogTemplateClass {
         '/(<table[^>]*>|<\/table>)/i',
         '/(<tr[^>]*>|<\/tr>)/i',
         '/<td[^>]*>(.*?)<\/td>/i',
-        '/<th[^>]*>(.*?)<\/th>/ie',
+        '/<th[^>]*>(.*?)<\/th>/i',
```
garthwebb: The replacement wasn't actually doing anything that required the ```/e``` flag.  Since the flag is deprecated (and throwing errors in dev) removed it.

extensions/wikia/Forum/Forum.i18n.php
```diff
@@ -67,6 +67,8 @@
 	'forum-board-new-message-heading' => 'Start a Discussion',
```
garthwebb: Changes in this file related to [SOC-590](https://wikia-inc.atlassian.net/browse/SOC-590).

extensions/wikia/Forum/ForumSpecialController.class.php
```diff
@@ -1,4 +1,7 @@
 <?php
+
```
garthwebb: [SOC-590](https://wikia-inc.atlassian.net/browse/SOC-590).

extensions/wikia/MyHome/data/ActivityFeedAPIProxy.php
```diff
@@ -25,7 +25,7 @@ public function get($limit, $start = null) {
 		$this->APIparams['rclimit'] = $limit;
 		$api = new ApiMain(new FauxRequest($this->APIparams));
 		$api->execute();
```
garthwebb: Changes in this file related to [SOC-452](https://wikia-inc.atlassian.net/browse/SOC-452)

extensions/wikia/Wall/WallHistoryController.class.php
```diff
@@ -1,5 +1,7 @@
 <?php
```
garthwebb: Changes in this file related to [SOC-578](https://wikia-inc.atlassian.net/browse/SOC-578)

extensions/wikia/WikiaRSS/WikiaRssHelper.class.php
```diff
@@ -7,7 +7,7 @@ class WikiaRssHelper {
 	 *
```
garthwebb: Changes in this file related to [SOC-522](https://wikia-inc.atlassian.net/browse/SOC-522)

extensions/wikia/AjaxPoll/AjaxPoll_body.php
```diff
@@ -149,54 +147,48 @@ public function getVotes($isSubmit = false) {
...
-				/* @var Language $wgLang */
-				$percent = $wgLang->formatNum(round($percent, 2));
-				$votes[ $nr ][ "title" ] = wfMsg("ajaxpoll-percentVotes", $percent);
```
kvas-damian: Did you remove this line on purpose?
I found one place where we use this title: https://github.com/Wikia/app/blob/dev/extensions/wikia/AjaxPoll/js/AjaxPoll.js#L19

extensions/wikia/Blogs/BlogTemplate.php
```diff
@@ -943,19 +944,18 @@ public static function getCategoryNames() {
 	}
 
 	public static function getResultsCount() {
-		global $wgLang;
-    	wfProfileIn( __METHOD__ );
-    	/* main query */
-    	$aResult = array();
-    	$aFields = array( 'distinct(page_id) as page_id' );
 		$res = self::$dbr->select(
```
kvas-damian: You could use ```selectField()``` method to make this code simpler.

extensions/wikia/Forum/ForumSpecialController.class.php
```diff
@@ -210,6 +213,14 @@ public function removeBoardModal() {
 		$boardId = $this->getVal( 'boardId', -1 );
 
 		$board = ForumBoard::newFromId( $boardId );
+		if ( empty( $board ) ) {
+			WikiaLogger::instance()->error( 'Error reporter: failed to find board', [
+				'jiraTicket' => 'SOC-590',
+				'boardId' => $boardId,
+				'method' => __METHOD__
+			]);
+			throw new ErrorPageError( 'forum-board-error-title', 'forum-board-nonexistent' );
```
kvas-damian: What do you think about returning 404 when ```$board``` doesn't exist?

extensions/wikia/Wall/WallHistoryController.class.php
```diff
@@ -217,8 +219,18 @@ private function getFormatedHistoryData( $history, $threadId = 0 ) {
 				$history[$key]['msgurl'] = $messagePageUrl;
 
 				$msgUser = $wm->getUser();
+				$msgPage = Title::newFromText( $msgUser->getName(), $ns );
+				if ( empty( $msgPage ) ) {
```
kvas-damian: @garthwebb did you check why ```WallHostory::get``` returns array that contains objects different than title object?

garthwebb: Yup, the message causing the error in this case, is from a ```$msgUser``` who is an anonymous user.  For some reason, someone ran a Special:Block on this IP address an included the HideUser option which hides user attribution (which doesn't make much sense for an IP).  This is what I suspect anyway, because the ```revision``` row loaded has the ```DELETED_USER``` flag set on it.

The call to ```$msgUser->getName()``` therefore returns an empty string rather than the IP address an anonymous user usually would have because of the HideUser restriction.  This means an empty "title" is used as an argument to ```Title::newFromText``` which returns an empty object.

I spent about 2.5 hours getting as close as I could to the root cause.  The only errors in production are for this wiki, and for this exact thread ID:

https://kibana.wikia-inc.com/index.html#dashboard/temp/AUyVOsKZrWhzbAyo4bmn

 Given this, and that most of the 138 instances of this error over the last month are likely from people at Wikia testing that URL, I can't say that this was time well spent.

extensions/wikia/AjaxPoll/AjaxPoll_body.php
garthwebb: Good catch, I removed it because it wasn't being used in the template, but I didn't see this usage.  I'll add it back.

extensions/wikia/Forum/Forum.i18n.php
```diff
@@ -67,6 +67,8 @@
...
 
 	'forum-no-board-selection-error' => '← Please select a board to post to',
+	'forum-board-error-title' => 'An error occurred',
+	'forum-board-nonexistent' => 'This forum board does not exist',
```
jsutterfield: Did we want to i18n these? If so, is there a ticket for that?

extensions/wikia/AjaxPoll/AjaxPoll_body.php
```diff
@@ -149,54 +147,48 @@ public function getVotes($isSubmit = false) {
...
+			$votes = ( new WikiaSQL() )
+				->SELECT( 'poll_answer' )
+					->COUNT( '*' )->AS_( 'count' )
+				->FROM( 'poll_vote' )
+				->WHERE( 'poll_id' )->EQUAL_TO( $this->mId )
+				->GROUP_BY( 'poll_answer' )
+				->runLoop( $dbr, function( &$votes, $row ) use ( &$total ) {
+					$votes[$row->poll_answer]['value'] = $row->count;
+					$total += $row->count;
+				}, [] );
```
jsutterfield: I don't believe you need to pass this empty array as the third parameter. That's the default value in the function signature.

extensions/wikia/Forum/ForumSpecialController.class.php
garthwebb: Yah, sounds like a better idea.

extensions/wikia/Wall/WallHistoryController.class.php
kvas-damian: Thanks for explanation. 
Having that knowledge I don't think this set of logging parameters will help us to track this issue in future. Instead I suggest checking if HideUser restriction is applied to user and make that clear in the code that we skip revisions that are made by users of that kind. 
What do you think?

garthwebb: I don't think this is the right place to make changes like this.  ```ArticleComment``` is the code that works with ```Revision``` objects which implement this feature.  It manipulates the ```User``` object it loads by setting the ```User```'s ```mName``` field directly (according to Tor) to preserve the original author of the thread:
https://github.com/Wikia/app/blob/d54b8898c9b02f32ca9aa5a292567ae9202369fc/extensions/wikia/ArticleComments/classes/ArticleComment.class.php#L217-L218

Now that we understand this better I'm going to remove the logging, add defaults and leave a comment behind.

I'm going to close out this ticket.  We're in the middle of replacing this system and putting more effort into an edge case no user has reported is not a good use of our resources that are already stretched.  The return on investment in this ticket is extremely low and is only getting smaller.

PR #6886, opened 2015-04-06
(CE-1420) Fix fatal error due incorrect case of class name in SMW

Grunny:
We switched MediaWiki's autoloader to be case sensitive a while ago.
In one instance, SemanticMediaWiki was using incorrect capitalisation
for the FormatJson class. Fixes the following fatal error:

```
PHP Fatal error: Class 'FormatJSON' not found in /extensions/wikia/
SemanticMediaWiki/includes/queryprinters/SMW_QP_JSONlink.php on line 355
```


macbre:
LGTM

PR #6888, opened 2015-04-07
(CE-1563) Backport patch for API vulnerability when running HHVM

Grunny:
The MW 1.24.2 security release included a fix for a vulnerability in
the MW API when running on HHVM:
https://phabricator.wikimedia.org/T85851

It wasn't backported as MediaWiki before 1.24 didn't officially support
HHVM. As we are testing HHVM, just in case we ever run the app on HHVM
on production, probably better to just backport it so we don't forget.

/cc @owend

### DIFF
--- a/includes/api/ApiFormatWddx.php
+++ b/includes/api/ApiFormatWddx.php
@@ -39,13 +39,7 @@ class ApiFormatWddx extends ApiFormatBase {
 	}
 
 	public function execute() {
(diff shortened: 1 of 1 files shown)


Grunny:
Looks like it's only an issue with older versions of HHVM, so as long as we make sure we're running an up to date version of HHVM we should be fine.

PR #6892, opened 2015-04-07
CONCF-80 | purge cache for Mercury Hapi API (/api/v1/article)

Warkot:
Can you have a look at this, please? @macbre @owend @bognix @rogatty

### DIFF
--- a/extensions/wikia/MercuryApi/MercuryApiHooks.class.php
+++ b/extensions/wikia/MercuryApi/MercuryApiHooks.class.php
@@ -88,9 +88,12 @@ class MercuryApiHooks {
 		global $wgServer;
 
 		if( $title->inNamespaces( NS_MAIN ) ) {
-			// Mercury API call from Ember to Hapi
+			// Mercury API call from Ember.js to Hapi.js e.g.
+			// http://elderscrolls.wikia.com/api/v1/article/Morrowind
+			// To access it, you have to set your client to be directed to the Mercury machines.
 			$urls[] = $wgServer . self::SERVICE_API_BASE . self::SERVICE_API_ARTICLE . $title->getPartialURL();
-			// Mercury API call from Hapi to MediaWiki
+			// Mercury API call from Hapi.js to MediaWiki e.g.
+			// http://elderscrolls.wikia.com/wikia.php?controller=MercuryApi&method=getArticle&title=Morrowind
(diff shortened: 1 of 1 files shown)


macbre:
PR title is a bit misleading, this actually enables cache invalidation :)

extensions/wikia/MercuryApi/MercuryApiHooks.class.php
```diff
@@ -85,9 +85,13 @@ public static function onArticleRollbackComplete( WikiPage $wikiPage, User $user
 	 * @return bool
 	 */
 	static public function onTitleGetSquidURLs( Title $title, Array &$urls ) {
+		global $wgServer;
+
 		if( $title->inNamespaces( NS_MAIN ) ) {
-			// Mercury API call from service
-			$urls[] = MercuryApiController::getUrl( 'getArticle', ['title' => $title->getPartialURL()] );
+			// Mercury API call from Ember to Hapi
+			$urls[] = $wgServer . self::SERVICE_API_BASE . self::SERVICE_API_ARTICLE . $title->getPartialURL();
```
macbre: Is there a helper for building such URLs? String concatenation looks a bit scary here.

bognix: Could you also provide in comments how those URLs should look like just to have the rough idea in the future what we should expect here

rogatty: There is no existing helper and I see no point in writing a new method that would do the same thing (concatenate). This is temporary solution anyway as we want to replace Hapi API with a service at some point.

Warkot: What I found that could be used here is `wfExpandUrl()`. What do you think about that?

macbre: `wfExpandUrl` would only add `$wgServer`. As @bognix wrote - please add an example URL in the comment above.

Making a method and unit tests for it will be beneficial when migration to the service takes place. I understand that it's a temporary solution, but let's not introduce technical dept when it can be easily avoided.

vforge:
LGTM

rogatty:
LGTM

hakubo:
LGTM

bognix:
LGTM

owend:
LGTM

PR #6894, opened 2015-04-08
Platform 1116 - adding Mercury App support for Universal Analytics

harnash:
Since Mercury has no simple way to query WikiFactory variables
for a specific wiki, new parameter is being sent with WikiVariables
response to allow enabling UA per wiki. It is temporary solution and
should be removed in next migration phases.

### DIFF
--- a/extensions/wikia/MercuryApi/models/MercuryApi.class.php
+++ b/extensions/wikia/MercuryApi/models/MercuryApi.class.php
@@ -128,7 +128,7 @@ class MercuryApi {
 			'theme' => SassUtil::getOasisSettings(),
 			'wikiCategories' => WikiFactoryHub::getInstance()->getWikiCategoryNames( $wgCityId ),
 			'analytics' => [
-				'UAEnabled' => !empty( $wgAnalyticsProviderUseUA ) ? $wgAnalyticsProviderUseUA : false,
+				'UAEnabled' => !empty( $wgAnalyticsProviderUseUA ),
 			],
 		];
 	}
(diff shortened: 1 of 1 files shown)


extensions/wikia/MercuryApi/models/MercuryApi.class.php
```diff
@@ -125,6 +127,9 @@ public function getWikiVariables() {
 			'mainPageTitle' => Title::newMainPage()->getPrefixedDBkey(),
 			'theme' => SassUtil::getOasisSettings(),
 			'wikiCategories' => WikiFactoryHub::getInstance()->getWikiCategoryNames( $wgCityId ),
+			'analytics' => [
+				'UAEnabled' => $wgAnalyticsProviderUseUA,
+			],
```
macbre: `$wgAnalyticsProviderUseUA` does not have a default value defined. Please use `!empty($wgAnalyticsProviderUseUA)` instead.

harnash: Fixed in 28b58b0.

extensions/wikia/MercuryApi/models/MercuryApi.class.php
```diff
@@ -125,6 +127,9 @@ public function getWikiVariables() {
 			'mainPageTitle' => Title::newMainPage()->getPrefixedDBkey(),
 			'theme' => SassUtil::getOasisSettings(),
 			'wikiCategories' => WikiFactoryHub::getInstance()->getWikiCategoryNames( $wgCityId ),
+			'analytics' => [
+				'UAEnabled' => !empty( $wgAnalyticsProviderUseUA ) ? $wgAnalyticsProviderUseUA : false,
```
macbre: `!empty` is enough, we need a boolean here, right?

harnash: Fixed in a7f6750.

macbre:
LGTM

harnash:
Merged

PR #6895, opened 2015-04-08
PLATFORM-1099: Fix broken call to Wikia::log()

wladekb:
Status object cannot be converted to string implicitly. Feed the textual version of Status to Wikia::log().

https://wikia-inc.atlassian.net/browse/PLATFORM-1099

/cc @macbre


macbre:
LGTM

macbre:
MediaWiki upstream defines ``__toString`` method of `Status` class. Worth backporting?

PR #6904, opened 2015-04-09
preserve original order

jacekjursza:


includes/wikia/api/ArticlesApiController.class.php
```diff
@@ -670,6 +670,7 @@ protected function getArticlesDetails( $articleIds, $articleKeys = [], $width =
 		$collection = [];
 		$resultingCollectionIds = [];
 		$titles = [];
+		$this->wg->Memc->set( self::getCacheKey( '135', self::DETAILS_CACHE_ID ), null );
```
SebastianMarzjan: I believe you don't need this here :)

includes/wikia/api/ArticlesApiController.class.php
```diff
@@ -776,19 +777,24 @@ protected function getArticlesDetails( $articleIds, $articleKeys = [], $width =
 		$collection = $this->appendMetadata( $collection );
 
 		$thumbnails = null;
-		//if strict return to original ids order
+		$collection = $this->preserveOriginalOrder( $articleIds, $collection );
```
SebastianMarzjan: What's the rationale behind moving the behaviour responsible for preserving the order from inside of the loop to outside of it? (all vs strict = true)

includes/wikia/api/ArticlesApiController.class.php
```diff
@@ -776,19 +777,24 @@ protected function getArticlesDetails( $articleIds, $articleKeys = [], $width =
...
-			foreach( $articleIds as $id ) {
-				if ( !empty( $collection[ $id ] ) ) {
-					$result[] = $collection[ $id ];
-				}
-			}
-			return $result;
+			return array_values( $collection );
 		}
-
 		return $collection;
```
SebastianMarzjan: Since keys are numeric, could you explain the case on which the return value needs to be different here and in L783?

includes/wikia/api/ArticlesApiController.class.php
jacekjursza: Argh!... thanks for being watchful!

includes/wikia/api/ArticlesApiController.class.php
jacekjursza: This is protected method used in 4 places of this class. Some API-s like (getList, getMostLinked) need ordered array, others - like getDetails - need dictionary. If we want to keep the output of those APIs unchanged - we need to have the possibility of returning results in both formats (dict and array). However we still can sort the dict version to keep the responses consistent.

includes/wikia/api/ArticlesApiController.class.php
jacekjursza: Not sure if I understand the quetion (which loop?), but hope that my second comments will clarify my approach.

SebastianMarzjan: Sorry, mistype - I meant $strict condition

SebastianMarzjan:
:+1:

includes/wikia/api/ArticlesApiController.class.php
```diff
@@ -776,19 +775,31 @@ protected function getArticlesDetails( $articleIds, $articleKeys = [], $width =
...
+		//if strict - return array instead of associative array (dict)
 		if ( $strict ) {
-			foreach( $articleIds as $id ) {
-				if ( !empty( $collection[ $id ] ) ) {
-					$result[] = $collection[ $id ];
-				}
-			}
-			return $result;
+			return array_values( $collection );
 		}
```
SebastianMarzjan: It makes sense to me to use if - else here for readability

includes/wikia/api/ArticlesApiController.class.php
```diff
@@ -776,19 +775,31 @@ protected function getArticlesDetails( $articleIds, $articleKeys = [], $width =
 		$collection = $this->appendMetadata( $collection );
 
 		$thumbnails = null;
-		//if strict return to original ids order
+		//The collection can be in random order (depends if item was found in memcache or not)
+		//lets preserve original order even if we are not using strict mode:
+		//to keep things consistent over time (some other APIs that are using sorted results are using
+		//ArticleApi::getDetails to fetch info about articles)
+		$orderedIdsFromTitles = array_diff( array_keys( $collection ), $articleIds );
+		$orderedIds = array_merge( (array)$articleIds, (array)$orderedIdsFromTitles );
```
SebastianMarzjan: What forces us to use typecasting here?

SebastianMarzjan:
Go for it!

PR #6905, opened 2015-04-09
CONCF-432 | Support articles with title "0"

rogatty:
https://wikia-inc.atlassian.net/browse/CONCF-432

`empty('0')` returns true which caused following error while trying to access http://fallout.wikia.com/wikia.php?controller=MercuryApi&method=getArticle&title=0:

```
exception: {
  message: "Bad request",
  code: 400,
  details: "You need to pass title or id of an article"
}
```


rogatty:
Ping @nandy-andy @Warkot @bognix

bognix:
LGTM

Warkot:
LGTM :cat2:

extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -158,11 +158,11 @@ private function getTitleFromRequest(){
 		$articleId = $this->request->getInt( self::PARAM_ARTICLE_ID, NULL );
 		$articleTitle = $this->request->getVal( self::PARAM_ARTICLE_TITLE, NULL );
 
-		if ( !empty( $articleId ) && !empty( $articleTitle ) ) {
+		if ( !empty( $articleId ) && !is_null( $articleTitle ) && trim( $articleTitle ) !== '' ) {
```
nandy-andy: If my thinking is correct you could extract the new part to a public method:
```php
public function isTitleStringValid( $title ) {
  return !is_null( $title ) && trim( $title ) !== '';
}
```
then re-use this in both conditions:
```php
if ( !empty( $articleId ) && isTitleStringValid( $articleTitle ) ) {
  ...
}

if ( empty( $articleId ) && !isTitleStringValid( $articleTitle ) ) {
  ...
}
```
then you can even unit test `isTitleStringValid()`! :japanese_ogre:

Warkot:
@nandy-andy, how does it look to you now?

nandy-andy:
Yeah, API helper makes even more sense since if it was public method in Nirvana controller it'd be accessible from outside. Thanks! :smile_cat: 

How about quick unit tests for it? :japanese_ogre:

extensions/wikia/MercuryApi/MercuryApiHelper.class.php
```diff
@@ -0,0 +1,14 @@
...
+class MercuryApiHelper {
+
+	/**
+	 * Check if a string is a valid article title
+	 *
+	 * @param string $title
+	 * @return bool
+	 */
+	public static function isTitleStringValid( $title ) {
+		return !is_null( $title ) && trim( $title ) !== '';
```
Warkot: Guys, I'm wondering if we should actually do the trimming. Shouldn't we let the MW determine if the just whitespace article title is OK or not? We would get this info from here: https://github.com/Wikia/app/blob/CONCF-432/extensions/wikia/MercuryApi/MercuryApiController.class.php#L170 as the title string is taken care of here https://github.com/Wikia/app/blob/dev/includes/Title.php#L146.

On the other hand, instead of using trim, we could check if the title string consists only of whitespace characters **and underscores**, to return the same error code as MW.
http://elderscrolls.wikia.com/wiki/%20_ → 400 bad request
http://elderscrolls.wikia.com/wikia.php?controller=MercuryApi&method=getArticle&title=%20_ → 404 not found

Or maybe it should be handled in some other way?

Warkot: Or maybe I shouldn't care that much? :wink: 
Ping! @nandy-andy @bognix @hakubo

Warkot: If you're wandering how I got this doubt, it came to me while I was creating this test case: https://github.com/Wikia/app/blob/CONCF-432/extensions/wikia/MercuryApi/tests/MercuryApiHelperTest.php#L50

hakubo: Maybe Title class or other MW class has a method for this?

Warkot: And by the way, those are valid links to the "The Elder Scrolls III: Morrowind" article content:
http://elderscrolls.wikia.com/wiki/%20The%20_%20Elder%20Scrolls%20III:%20Morrowind%20_%20
http://elderscrolls.wikia.com/wikia.php?controller=MercuryApi&method=getArticle&title=%20The%20_%20Elder%20Scrolls%20III:%20Morrowind%20_%20

Warkot:
Hey, I'm in a pinch. @hakubo yesterday mentioned that using helpers is not the best idea, as they quickly become a `bagOfStuff` and I can agree on that. Therefore I moved the `isTitleStringValid` method to the `MercuryApiController` class and set is as private. But how can I unit-test a private method? :frowning:

And a BTW question: should it be converted to a non-static class? :question:

SebastianMarzjan:
@Warkot 
As for validation, would WikiaValidatorString work for your use case?

Warkot:
@SebastianMarzjan, could you take a look, please? Have I done it the right way? Is the test added properly? :octocat:

SebastianMarzjan:
@Warkot 
This part makes sense to me :+1:

Warkot:
Thanks, @SebastianMarzjan! :)

vforge:
LGTM

bognix:
LGTM

nandy-andy:
LGTM (thanks @SebastianMarzjan I forgot about our validators :japanese_ogre:) Great job and communication, @Warkot! :smile_cat:

hakubo:
Nice, thanks

PR #6908, opened 2015-04-10
SOC-582 Track events on Special:UserSignup.

garthwebb:
Also, update the logic around showing/hiding captcha to be a bit more intuitive.

https://wikia-inc.atlassian.net/browse/SOC-582

### DIFF
--- a/extensions/wikia/UserLogin/UserSignupSpecialController.class.php
+++ b/extensions/wikia/UserLogin/UserSignupSpecialController.class.php
@@ -688,7 +688,7 @@ class UserSignupSpecialController extends WikiaSpecialPageController {
 	 *
 	 * @param string $label
 	 */
-	public function track( $label ) {
+	protected function track( $label ) {
 		\Track::event( 'trackingevent', [
 			'ga_category' => 'user-sign-up',
 			'ga_action' => 'request',
--- a/extensions/wikia/UserLogin/UserSignupSpecialController.class.php
+++ b/extensions/wikia/UserLogin/UserSignupSpecialController.class.php
@@ -175,6 +175,7 @@ class UserSignupSpecialController extends WikiaSpecialPageController {
 			$this->track( 'signup-successful' );
 			$this->wg->out->redirect( $redirectUrl );
 		} else {
+			$this->track( 'signup-failed' );
 			$this->response->setData( [
 				'result' => $result,
 				'msg' => $response->getVal( 'msg', '' ),
--- a/extensions/wikia/UserLogin/UserSignupSpecialController.class.php
+++ b/extensions/wikia/UserLogin/UserSignupSpecialController.class.php
@@ -183,28 +183,26 @@ class UserSignupSpecialController extends WikiaSpecialPageController {
 		}
 	}
 
+	/**
+	 * Returns captcha HTML
+	 */
 	public function captcha() {
-		$this->rawHtml = '';
-		$this->isFancyCaptcha = false;
-
 		if ( $this->shouldDisableCaptcha() ) {
 			return;
 		}
 
 		$captchaObj = $this->getCaptchaObj();
 		if ( !empty( $captchaObj ) ) {
-			$this->rawHtml = $captchaObj->getForm();
-			$this->isFancyCaptcha = $captchaObj instanceof Captcha\Module\FancyCaptcha;
+			$this->response->setVal( 'rawHtml', $captchaObj->getForm() );
 		}
 	}
 
 	private function getCaptchaObj() {
-		$captchaObj = null;
-
-		if ( !empty( $this->wg->EnableCaptchaExt ) ) {
-			$captchaObj = Captcha\Factory\Module::getInstance();
+		if ( empty( $this->wg->EnableCaptchaExt ) ) {
+			return null;
 		}
-		return $captchaObj;
+
+		return Captcha\Factory\Module::getInstance();
 	}
 
 	/**
--- a/extensions/wikia/UserLogin/UserSignupSpecialController.class.php
+++ b/extensions/wikia/UserLogin/UserSignupSpecialController.class.php
@@ -688,12 +688,12 @@ class UserSignupSpecialController extends WikiaSpecialPageController {
 	 *
 	 * @param string $label
 	 */
-	public static function track( $label ) {
+	public function track( $label ) {
 		\Track::event( 'trackingevent', [
 			'ga_category' => 'user-sign-up',
 			'ga_action' => 'request',
 			'ga_label' => $label,
-			'beacon' => !empty( F::app()->wg->DevelEnvironment ) ? 'ThisIsFake' : wfGetBeaconId(),
+			'beacon' => !empty( $this->wg->DevelEnvironment ) ? 'ThisIsFake' : wfGetBeaconId(),
 		] );
 	}
 }
--- a/extensions/wikia/UserLogin/UserSignupSpecialController.class.php
+++ b/extensions/wikia/UserLogin/UserSignupSpecialController.class.php
@@ -14,7 +14,10 @@ class UserSignupSpecialController extends WikiaSpecialPageController {
 	public function __construct() {
 		parent::__construct( 'UserSignup', '', false );
 
-		$this->disableCaptcha();
+		// Disable captcha for automated tests and wikia mobile and sites disabling it, e.g. Internal
+		if ( $this->shouldDisableCaptcha() ) {
+			$this->disableCaptcha();
+		}
 	}
 
 	public function init() {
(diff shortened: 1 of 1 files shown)


extensions/wikia/UserLogin/UserSignupSpecialController.class.php
```diff
@@ -122,54 +125,61 @@ public function signupForm() {
 		// process signup
 		$redirected = $this->request->getVal( 'redirected', '' );
 		if ( $this->wg->Request->wasPosted() && empty( $redirected ) ) {
+			$this->handleSignupFormSubmit();
+		} else {
+			$this->track( 'signup-start' );
+		}
+	}
```
garthwebb: Moved the following code down into the ```handleSignupFormSubmit``` function below.  Additional changes to not set 'msg' and 'errParam' unless result is not ```ok```, and to set them via ```this->response->setData```.

Also, added a tracking call after we've successfully created a user.

extensions/wikia/UserLogin/UserSignupSpecialController.class.php
```diff
@@ -177,24 +187,21 @@ public function captcha() {
...
+
+		$captchaObj = $this->getCaptchaObj();
+		if ( !empty( $captchaObj ) ) {
+			$this->rawHtml = $captchaObj->getForm();
+			$this->isFancyCaptcha = $captchaObj instanceof Captcha\Module\FancyCaptcha;
 		}
 	}
 
 	private function getCaptchaObj() {
 		$captchaObj = null;
```
garthwebb: We aren't using ConfirmEdit anymore

extensions/wikia/UserLogin/UserSignupSpecialController.class.php
```diff
@@ -177,24 +187,21 @@ public function captcha() {
...
-		if ( $this->shouldDisplayCaptcha() ) {
-			$captchaObj = $this->getCaptchaObj();
-			if ( !empty( $captchaObj ) ) {
-				$this->rawHtml = $captchaObj->getForm();
-				$this->isFancyCaptcha = (
-					class_exists( 'Captcha\Module\FancyCaptcha' ) &&
-					$captchaObj instanceof Captcha\Module\FancyCaptcha
-				);
-			}
+		if ( $this->shouldDisableCaptcha() ) {
```
armonr: Minor: This can be moved to top of the method.

extensions/wikia/UserLogin/UserSignupSpecialController.class.php
```diff
@@ -177,24 +187,21 @@ public function captcha() {
 		$this->rawHtml = '';
 		$this->isFancyCaptcha = false;
 
-		if ( $this->shouldDisplayCaptcha() ) {
-			$captchaObj = $this->getCaptchaObj();
-			if ( !empty( $captchaObj ) ) {
-				$this->rawHtml = $captchaObj->getForm();
-				$this->isFancyCaptcha = (
-					class_exists( 'Captcha\Module\FancyCaptcha' ) &&
```
armonr: Hm... Why was this check here before and is it still needed?

extensions/wikia/UserLogin/UserSignupSpecialController.class.php
garthwebb: Ah, I assumed ```rawHtml``` and ```isFancyCaptcha``` needed this initialization.

extensions/wikia/UserLogin/UserSignupSpecialController.class.php
garthwebb: We talked about this previously:

https://github.com/Wikia/app/pull/6449/files#diff-fd432f58d5c8919e5fe6b60bacf809ecR185

extensions/wikia/UserLogin/UserSignupSpecialController.class.php
```diff
@@ -122,82 +125,84 @@ public function signupForm() {
...
+			] );
 		}
 	}
 
+	/**
+	 * Returns captcha HTML
+	 */
 	public function captcha() {
-		$this->rawHtml = '';
-		$this->isFancyCaptcha = false;
```
garthwebb: I completely removed this because I don't see it used anywhere.  @jsutterfield , you worked on the FancyCaptcha behavior, can you confirm?

extensions/wikia/UserLogin/UserSignupSpecialController.class.php
garthwebb: ... also this entire line is no longer necessary.

armonr: Right, thanks for the reference.

extensions/wikia/UserLogin/UserSignupSpecialController.class.php
jsutterfield: Correct, that that flag had been used in the UserSignupSpecial_captcha.php template. It was used to determine if we wrapped that captcha in a <div> with a class of "fancy-captcha". I removed it from the template [here](https://github.com/Wikia/app/pull/6466/files#diff-45e7b8c6bc3ee4ab9fe072f18a7356e0L3), and updated getForm on FancyCaptcha to take care of that div, but it looks like I forgot to remove it from here.

extensions/wikia/UserLogin/UserSignupSpecialController.class.php
```diff
@@ -122,82 +125,84 @@ public function signupForm() {
 		// process signup
 		$redirected = $this->request->getVal( 'redirected', '' );
 		if ( $this->wg->Request->wasPosted() && empty( $redirected ) ) {
+			$this->handleSignupFormSubmit();
+		} else {
+			$this->track( 'signup-start' );
```
armonr: Please call track statically `self::track()`

extensions/wikia/UserLogin/UserSignupSpecialController.class.php
```diff
@@ -122,82 +125,84 @@ public function signupForm() {
...
-				$this->wg->out->redirect( $redirectUrl );
+				$params = [
+					'method' => 'sendConfirmationEmail',
+					'username' => $this->username,
+					'byemail' => intval( $this->byemail ),
+				];
+				$redirectUrl = $this->wg->title->getFullUrl( $params );
 			}
 
+			$this->track( 'signup-successful' );
```
armonr: Same here, please call statically.

extensions/wikia/UserLogin/UserSignupSpecialController.class.php
jsutterfield: @garthwebb, is there a reason you made `track()` a static function?

armonr: Well, it is not an instance method, so it makes sense to be static.

jsutterfield: I guess my inclination is to make a method instance first and static if it has to be. Since this is only being called by instance methods, we know it could be instance, and it's not some utility method used outside of the class by other code. But I understand why it makes sense as static too. Just different preferences :)

garthwebb: Quite a conversation!

So to start, why would I always need to call a static function with ```self```?  Its perfectly legal to call a static function with the ```$this``` instance variable.  If ```track``` was defined in a subclass of ```UserSignupSpecialController```, I'd end up running the parent version rather than the child.  If I specifically wanted the parent method, then ```self``` would be the way to go.  Without any other considerations, I'd say calling with this ```$this``` is preferable as it makes default, expected assumptions about inheritance.

The opposite, of course, is not true;  you don't want to call an instance method statically.  Even if it currently does not make use of ```$this``` it could, and PHP strict standards rightly scolds you for this.

To the next point, why did I make this static?  Copy and paste.  Generally I don't make things static unless there's a reason to do so.  I don't need to call this statically, nor do I see a reason why I would in the future.

So, I pulled the keyword and now its dynamic.  I also made it protected since there's no reason anyone outside this class and possible children who should be calling it.

extensions/wikia/UserLogin/UserSignupSpecialController.class.php
```diff
@@ -122,82 +125,84 @@ public function signupForm() {
...
+			$this->wg->out->redirect( $redirectUrl );
+		} else {
+			$this->response->setData( [
+				'result' => $result,
+				'msg' => $response->getVal( 'msg', '' ),
+				'errParam' => $response->getVal( 'errParam', '' ),
+			] );
```
armonr: I know it is not part of the acceptance criteria, but it would be interesting to see failed attempts logged too. It may give us more insight into why signups fail.

extensions/wikia/UserLogin/UserSignupSpecialController.class.php
armonr: All good and valid points! I think the older versions of PHP, not using `this` or `static` for static methods was problematic or at least discouraged. Now it is not the case, of course.

armonr:
LGTM

jsutterfield:
LGTM

extensions/wikia/UserLogin/UserSignupSpecialController.class.php
garthwebb: @armonr, seems reasonable.

PR #6916, opened 2015-04-10
In PHP 5.6 mcrypt_decrypt does not accept invalid keys, adding \0 is the...

ghost:
... quick workaround.


macbre:
So PHP :)

LGTM

PR #6920, opened 2015-04-13
MAIN-4321

penguinstyles:


penguinstyles:
@owend

extensions/wikia/SpecialEditHub/SpecialEditHub.setup.php
```diff
@@ -46,7 +46,5 @@
 
 $wgGroupPermissions['*']['edithub'] = false;
 $wgGroupPermissions['staff']['edithub'] = true;
-$wgGroupPermissions['sysop']['edithub'] = true;
```
owend: CSS changes look fine, the other Permissions change looks fine, can you just remind me why the 'edithub' permission would be removed here?

timmyquivy: @owend Right now edithub allows local sysops to change marketing/core hub pages. We want to have volunteer sysops to allow for general curation, but it may just be cleaner in the long run not to give them this power. If it proves later that we want to change course, we can just put it back in.

owend: Sounds good, thanks for the clarification!

PR #6930, opened 2015-04-13
(CE-1531) Upgrade SemanticResultFormats to version 1.8

Grunny:
Upgrade SemanticResultFormats to version 1.8 which is compatible with SMW 1.8
which we are now running.


adamkarminski:
LGTM (Looks Grunny To Me)

PR #6932, opened 2015-04-14
CONCF-407 set WikiaTopsAds width to 1064 so it fits leaderboard and topbutton

bognix:
ping @vforge @Warkot [@gabrys | @nandy-andy]

**DON'T MERGE BEFORE 'OK' FROM ADENG TEAM**


vforge:
LGTM

Warkot:
LGTM :+1:

extensions/wikia/AdEngine/js/AdLogicPageDimensions.js
```diff
@@ -77,6 +77,7 @@ define('ext.wikia.adEngine.adLogicPageDimensions', [
 			mediaQueriesToCheck.twoColumns = 'screen and (min-width: 1064px)';
```
hakubo: how about using 
```js
$.extend(mediaQueriesToCheck, {
...
});
```

bognix: jQuery is not included here. What is more I remember there was an experiment for running Ads in Head and as far as I understand adding jQuery as a dependency wouldn't be optimal solution

PR #6933, opened 2015-04-14
PLATFORM-1133: Extend MediaWiki/Helios integration to recognize the authentication token via cookie

ghost:
### extensions/wikia/Helios/Client.class.php

MediaWiki's `MWHttpRequest` class heavily relies on Messaging API (`wfMessage()`). Normally `$wgLang` is set after `$wgUser`. On per-request authentication with an access token we make an HTTP request to Helios. In other words we use `MWHttpRequest` during the process of `wgUser` creation. We need  `$wgLang` to be present. `GlobalStateWrapper` makes it easy to achieve and enables us to set the global variable in the local, function's scope.
### extensions/wikia/Helios/User.class.php
- `getAccessToken` method was added — it extracts access token from HTTP request data. The value from the cookie takes precedence over the value from the HTTP header. The name of the cookie is defined as class constant: `ACCESS_TOKEN_COOKIE`.
- `newFromToken` method was modified to use the above method.
### extensions/wikia/Helios/tests/UserTest.php

Tests are good.
## Code Review

@wladekb @macbre 
## Stakeholders

@gbenson-wikia @garthwebb @ArturKlajnerok

### DIFF
--- a/extensions/wikia/Helios/Client.class.php
+++ b/extensions/wikia/Helios/Client.class.php
@@ -32,7 +32,6 @@ class Client
 		// Crash if we cannot make HTTP requests.
 		\Wikia\Util\Assert::true( \MWHttpRequest::canMakeRequests() );
 
-
 		// Add client_id and client_secret to the GET data.
 		$getParams['client_id'] = $this->clientId;
 		$getParams['client_secret'] = $this->clientSecret;
--- a/extensions/wikia/Helios/Client.class.php
+++ b/extensions/wikia/Helios/Client.class.php
@@ -52,6 +52,15 @@ class Client
 
 		$options = array_merge( $options, $extraRequestOptions );
 
+		/*
+		 * MediaWiki's MWHttpRequest class heavily relies on Messaging API
+		 * (wfMessage()) which happens to rely on the value of $wgLang.
+		 * $wgLang is set after $wgUser. On per-request authentication with
+		 * an access token we use MWHttpRequest before wgUser is created so
+		 * we need $wgLang to be present. With GlobalStateWrapper we can set
+		 * the global variable in the local, function's scope, so it is the
+		 * same as the already existing $wgContLang.
+		 */
 		global $wgContLang;
 		$wrapper = new GlobalStateWrapper( [ 'wgLang' => $wgContLang ] );
 
--- a/extensions/wikia/Helios/Client.class.php
+++ b/extensions/wikia/Helios/Client.class.php
@@ -1,6 +1,7 @@
 <?php
 namespace Wikia\Helios;
 use Wikia\Util\RequestId;
+use Wikia\Util\GlobalStateWrapper;
 
 /**
  * A client for Wikia authentication service.
@@ -31,6 +32,7 @@ class Client
 		// Crash if we cannot make HTTP requests.
 		\Wikia\Util\Assert::true( \MWHttpRequest::canMakeRequests() );
 
+
 		// Add client_id and client_secret to the GET data.
 		$getParams['client_id'] = $this->clientId;
 		$getParams['client_secret'] = $this->clientSecret;
@@ -51,9 +53,15 @@ class Client
 
 		$options = array_merge( $options, $extraRequestOptions );
 
+		global $wgContLang;
+		$wrapper = new GlobalStateWrapper( [ 'wgLang' => $wgContLang ] );
+
 		// Request execution.
 		/** @var \MWHttpRequest $request */
-		$request = \Http::request( $options['method'], $uri, $options );
+		$request = $wrapper->wrap( function() use ( $options, $uri ) {
+			return \Http::request( $options['method'], $uri, $options );
+		} );
+
 		$status = $request->status;
 
 		// Response handling.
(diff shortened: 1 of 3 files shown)


michalroszka:
The tests are passing now.

extensions/wikia/Helios/User.class.php
```diff
@@ -27,35 +29,57 @@ public static function debugLogin( $password, $callingMethod )
 	}
 
 	/**
-	 * Creates a MediaWiki User object based on the token given in the HTTP request.
+	 * Extracts access token from HTTP request data.
+	 *
+	 * @param \WebRequest $request the HTTP request data as an object
+	 * @return String access token
 	 */
```
macbre: `getAccessToken` can return `null` as well.

extensions/wikia/Helios/tests/UserTest.php
```diff
@@ -17,32 +17,66 @@ public function setUp()
...
+			->method( 'getCookie' )
+			->willReturn( null );
+
+		$this->webRequestMock->expects( $this->once() )
+			->method( 'getHeader' )
+			->with( 'AUTHORIZATION' )
+			->willReturn( "Bearer $token" );
+
+		$this->assertEquals( User::getAccessToken( $this->webRequestMock ), $token );
+	}
```
macbre: Would it make sense to test the case when both cookie and header are set, but with a different token (i.e. test that the cookie is checked before HTTP header)?

extensions/wikia/Helios/Client.class.php
```diff
@@ -51,9 +52,15 @@ public function request( $resourceName, $getParams = [], $postData = [], $extraR
 
 		$options = array_merge( $options, $extraRequestOptions );
 
+		global $wgContLang;
+		$wrapper = new GlobalStateWrapper( [ 'wgLang' => $wgContLang ] );
```
wladekb: can you add a bit of insight in the comment why is that needed here?

extensions/wikia/Helios/User.class.php
wladekb: ...or to be more precise: an empty string

extensions/wikia/Helios/User.class.php
```diff
@@ -27,35 +29,57 @@ public static function debugLogin( $password, $callingMethod )
...
-						return \User::newFromId( $tokenInfo->user_id );
-					}
+	/**
+	 * Creates a MediaWiki User object based on the token given in the HTTP request.
+	 *
+	 * @param \WebRequest $request the HTTP request data as an object
+	 *
+	 * @return \User on successful authentication
+	 */
+	public static function newFromToken( \WebRequest $request )
```
wladekb: this is rather a "newFromWebRequest" method

michalroszka: there is a bunch of `newFrom` and `initFrom` methods that could be renamed to `newFromWebRequest`, too, namely `newFromSession`, `newFromConfirmationCode`, `initFromCookie`, etc. I'll leave it as it is.

extensions/wikia/Helios/tests/UserTest.php
michalroszka: It would. I have added the test.

extensions/wikia/Helios/User.class.php
michalroszka: The method now returns a string or `null`.

extensions/wikia/Helios/Client.class.php
michalroszka: I have added a verbose comment.

extensions/wikia/Helios/Client.class.php
```diff
@@ -1,6 +1,7 @@
 <?php
 namespace Wikia\Helios;
 use Wikia\Util\RequestId;
+use Wikia\Util\GlobalStateWrapper;
```
drsnyder: Name spaced classes should live in `/lib/Wikia`. They aren't really PSR-* compliant under `/extensions`.

extensions/wikia/Helios/Client.class.php
```diff
@@ -51,9 +52,24 @@ public function request( $resourceName, $getParams = [], $postData = [], $extraR
...
+		 */
+		global $wgContLang;
+		$wrapper = new GlobalStateWrapper( [ 'wgLang' => $wgContLang ] );
+
 		// Request execution.
 		/** @var \MWHttpRequest $request */
-		$request = \Http::request( $options['method'], $uri, $options );
+		$request = $wrapper->wrap( function() use ( $options, $uri ) {
+			return \Http::request( $options['method'], $uri, $options );
+		} );
```
drsnyder: I think a better approach would be to inject an HTTP client object. This would mean moving this functionality to a wrapper class but result would be a much cleaner `\Wikia\Helios\Client::request`. 

It would also mean that you could decouple this class from MW which would make it eligible for inclusion in `/lib/Wikia` and make it easier to test. 

At some point, and this illustrates why, we'll need to move away from `Http:request` and use an [HTTP client](http://docs.guzzlephp.org/en/latest/) that we can inject.

PR #6938, opened 2015-04-14
SOC-665 pass namespace to watchedpagecontroller

jsutterfield:
Watched Page emails aren't being sent out properly for articles not in the main namespace. For Talk pages, this results incorrect links, and for Category and File pages, this was resulting in no email being sent out at all. This ticket sends the namespace with the title name so we can create the correct title object.

Ticket: https://wikia-inc.atlassian.net/browse/SOC-665

### DIFF
--- a/extensions/wikia/Email/Controller/WatchedPageController.class.php
+++ b/extensions/wikia/Email/Controller/WatchedPageController.class.php
@@ -24,7 +24,8 @@ class WatchedPageController extends EmailController {
 	}
 
 	public function initEmail() {
-		$this->title = \Title::newFromText( $this->request->getVal( 'title' ) );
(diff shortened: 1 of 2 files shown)


extensions/wikia/Email/Controller/WatchedPageController.class.php
```diff
@@ -24,7 +24,8 @@ public function getSubject() {
 	}
 
 	public function initEmail() {
-		$this->title = \Title::newFromText( $this->request->getVal( 'title' ) );
+		$nameSpace = $this->request->getVal( 'nameSpace', NS_MAIN );
```
armonr: minor: `getInt()` might be a better option

armonr:
LGTM

PR #6939, opened 2015-04-15
(CE-1532) Make sure we always have a parameter array

Grunny:

```
PHP Catchable fatal error: Argument 2 passed to
SMWQueryProcessor::addThisPrintout() must be of the type array,
null given, called in extensions/wikia/SemanticMediaWiki/
includes/api/ApiSMWQuery.php on line 50 and defined in
extensions/wikia/SemanticMediaWiki/includes/SMW_QueryProcessor.php
on line 234
```


kamilkoterba:
lgtm

PR #6940, opened 2015-04-15
CE-1612 adding user updates on user email confirm hook 

kamilkoterba:
https://wikia-inc.atlassian.net/browse/CE-1612

We need to have user_email_authenticated field up to date.

Reorganized hooks on which we update user data due to email authentication process.

Removed EmailChangeConfirmed that runs on email change confirm, but not on user signup email confirm
Added ConfirmEmailComplete that runs both on email change confirm and user signup email confirm
Added InvalidateEmailComplete that runs when email was changed in options but isn't confirmed yet (user_email_authentication field set to null user_email not changed)

@Wikia/community-engineering


lgarczewski:
LGTM.

PR #6909, opened 2015-04-10
[PLATFORM-1007] Renamed MostimagesInContentPage to MostLinkedFilesInContentPage

jcellary:
https://wikia-inc.atlassian.net/browse/PLATFORM-1007

Renamed special page MostimagesInContentPage to MostLinkedFilesInContentPage created earlier as requested by community support.

The actual class has been temporarily duplicated due to caching of query result once every 24 hours. So the classes will run in parallel for a few days and then I will delete the old unneeded class.

I'm not experienced in making changes in this repo, so please take a good look at my PR ;)

Pinging @macbre @michalroszka @wladekb

PR #6942, opened 2015-04-15
PLATFORM-1007: rename MostimagesInContent to MostLinkedFilesInContent

macbre:
Use a bit different approach than in #6909 - rename only the special page name and i18n messages. Do not touch class names and query pages.

@jcellary


jcellary:
LGTM

macbre:
I need to change the query in `ImageServingDriverMainNS.class.php` to match the new name of the special page.

macbre:
`MostimagesInContent` will still be used "internally":

* when refreshing the data via `updateSpecialPages.php` script
* by ImageServing when querying `querycache` table

wladekb:
lgtm

PR #6946, opened 2015-04-15
CE-1606 Front-end for list subpages and a landing page

adamkarminski:


extensions/wikia/Insights/models/QueryPagesModel.php
```diff
@@ -0,0 +1,83 @@
...
+		parent::__construct( $wikiId );
+	}
+
+	/**
+	 * Get list of article
+	 *
+	 * @param int $limit
+	 * @return array
+	 */
+	public function getList( $limit = 100 ) {
```
adamkarminski: We can use a constant to hold this number.

PR #6948, opened 2015-04-16
wikia.scriptwriter update: adding injectHtml method

gabrys:


resources/wikia/modules/scriptwriter.js
```diff
@@ -118,22 +124,38 @@ define('wikia.scriptwriter', [
...
+	/**
+	 * Queue an HTML injection
+	 *
+	 * @param {Object}      params
+	 * @param {HTMLElement} params.element  Element to inject HTML to
+	 * @param {string}      params.html     HTML to inject
+	 * @param {Function}    params.callback Function to call when done
+	 */
+	function queueHtmlInjection(params) {
+		log(['push', params], 'debug', module);
```
nandy-andy: `'queueHtmlInjection'`

resources/wikia/modules/spec/scriptwriter.spec.js
```diff
@@ -0,0 +1,124 @@
...
+				insertType: 'append',
+				script: {
+					text: 'document.write("<script src=\\\"http://some.url/script.js\\\"></script>");'
+				},
+				done: jasmine.any(Function)
+			}
+		);
+	});
+
+	it('injectScriptByHtml ghostwriter', function () {
```
nandy-andy: `'injectHtml ghostwriter'`

resources/wikia/modules/spec/scriptwriter.spec.js
```diff
@@ -0,0 +1,124 @@
...
+		spyOn(mocks, 'postscribe');
+		getPsModule().injectScriptByUrl('foo', 'http://some.url/script.js');
+		expect(mocks.postscribe).toHaveBeenCalledWith(
+			mocks.domElement,
+			'<script src="http://some.url/script.js"></script>',
+			jasmine.any(Object)
+		);
+	});
+
+	it('injectScriptByHtml postscribe', function () {
```
nandy-andy: `'injectHtml postscribe'`

nandy-andy:
Nothing but small typos. Except that LGTM :+1:

resources/wikia/modules/scriptwriter.js
```diff
@@ -24,201 +24,187 @@
...
-			});
-		}
+		loader(url).done(processQueue);
 	}
 
-	function push(type, args, callback) {
-		log(['push', type, args, callback], 'debug', module);
-
-		queue.push({type: type, args: args, callback: callback});
+	function getElement(elementOrElementId) {
```
nandy-andy: I wonder if we can have variables declaration on the beginning of the function and assignment later:
```js
function getElement(elementOrElementId) {
	var el;

	if (elementOrElementId && elementOrElementId.getElementsByTagName) {
		return elementOrElementId;
	}

	if (typeof elementOrElementId === 'string') {
		el = doc.getElementById(elementOrElementId);
		if (el) {
			return el;
		}
	}
	
	throw 'wikia.scriptwriter: Not a valid element or element id ' + elementOrElementId;
}
```
But it most likely doesn't make sense - few more bytes of code and the JSHint rule is to have only one `var` per function. It doesn't say that the declaration should be at the very beginning. So, ignore this comment and thanks for reading :japanese_ogre: (I'm jet-lagged :stuck_out_tongue_winking_eye:)

PR #6950, opened 2015-04-16
Remove plain text option for user emails

jsutterfield:
Remove the ability for users to opt out of HTML emails.

Ticket: https://wikia-inc.atlassian.net/browse/SOC-497


kvas-damian:
Nice, 0 lines added ;)
One question, will we send html emails after this change in places like this https://github.com/Wikia/app/blob/dev/includes/User.php#L3626 ?
I'm not sure if it's in scope of this ticket, but it'd be great if we remove all checks of this kind.

kvas-damian:
Oh, I see this value set to true at default in config. 
What about users who already changed that setting are we going to default them to html as well?

bkoval:
Yeah, I think that's intentional.

extensions/wikia/UserPreferencesV2/UserPreferencesV2.class.php
```diff
@@ -175,11 +175,6 @@ static public function onGetPreferences( $user, &$defaultPreferences ) {
 			}
 		}
 
-		if ( isset( $defaultPreferences['htmlemails'] ) ) {
```
bkoval: Should we remove the whole statement or just a check (and always invoke the three instructions that were inside the statement body)?

jsutterfield: Yep, the whole statement should be removed.

We had been adding 'htmlemails' to the defaultPreferences array [here](https://github.com/Wikia/app/pull/6950/files#diff-bf6d5d0093a049e88d2096e1be377e07L464), and then this code was called as part of a hook to actually add the checkbox. Now that we're deleting that code from Preferences.php, if we remove this check it this will throw an error since there's no longer an `htmlemails` key. We also just want to get rid of the option altogether, so deleting is the way to go :)

jsutterfield:
@kvas-damian, yep, this ticket helped get my LOC ratio closer to 0 :)

As for your question re: user's who already have the setting, I checked with Cristina and we're going to leave them as is for now. They'll still receive the plain text emails if that's what they opted for. As we convert each email over to the new designs, we'll remove the logic to check for this flag (if there is any) and essentially opt the user's into receiving the new html email.

kvas-damian:
in this case LGTM

bkoval:
LGTM :kimono:

PR #6953, opened 2015-04-16
PLATFORM-1151 Helios registration: we are logging false positives

ghost:
I have fixed the condition so we query user_helios when in shadow mode rather than user. Also, I have changed logging statements so errors are logged as errors not as notices.


macbre:
LGTM

Zlatan says: LGTM

PR #6955, opened 2015-04-16
CONCF-427 | Support for international sharing buttons

vforge:
Ping @warkot @bognix


bognix:
LGTM

extensions/wikia/PageShare/PageShareController.class.php
```diff
@@ -14,14 +14,15 @@ public function index() {
...
+		global $wgMemc, $wgEnablePageShareWorldwide;
 
 		$browserLang = $this->getVal( 'browserLang' );
 		$useLang = $this->getVal( 'useLang' );
+		$title = $this->getVal( 'title' );
 		$shareLang = PageShareHelper::getLangForPageShare( $browserLang, $useLang );
 
 		// If social icons should be enabled for EN users, and language is different than EN return false
-		if ( empty( $wgEnablePageShareWorldwide ) && $shareLang !== PageShareHelper::SHARE_DEFAULT_LANGUAGE ) {
+		if ( empty( $wgEnablePageShareWorldwide ) && ( $shareLang !== PageShareHelper::SHARE_DEFAULT_LANGUAGE ) ) {
```
Warkot: AFAIK, you don't need the inner braces as `!==` precedes `&&` (ref. http://php.net/manual/en/language.operators.precedence.php).

Warkot:
Do we really need points in SVGs defined with six digit precision (digits after decimal point)? It might be optimized, IMHO. :water_buffalo:

extensions/wikia/PageShare/PageShareController.class.php
vforge: It add readability to the code. Are you suggesting removal?

vforge:
As the icons can be very small even smallest change can differ their looks - do you want them checked and see if they can be optimized more?

extensions/wikia/PageShare/PageShareController.class.php
Warkot: Nah, you can leave them. It's readable enough without them for me but it might actually help someone.

Warkot:
You're probably right. And such optimization would save only few bytes, so let's don't overoptimize. ;) LGTM :shipit:

PR #6956, opened 2015-04-16
LoadBalancer: revert the support for multiple masters

macbre:
No longer needed as we use consul now instead of haproxy.

Revert #4629

See https://github.com/Wikia/config/pull/1057 as well

@michalroszka


michalroszka:
:+1:

PR #6957, opened 2015-04-16
Soc 681

SeanColombo:


owend:
LGTM

extensions/wikia/Chat2/Chat.class.php
```diff
@@ -40,7 +39,7 @@ static public function echoCookies(){
 		if( !$wgUser->isLoggedIn() ) {
 			return array("key" => false ) ;
 		}
-		$key = md5( $wgUser->getId() . "_" . time() . '_' .  mt_rand(0, 65535) );
+		$key = sha1( $wgUser->getId() . "_" . microtime() . '_' .  mt_rand() );
 		$wgMemc->set($key, array( "user_id" => $wgUser->getId(), "cookie" => $_COOKIE) , 60*60*48);
```
macbre: @SeanColombo: as we're here, please change the structure of the key to be not just the hash, but to contain some sort of a prefix, eg. `Chat::cookies::<hash>`

SeanColombo: @macbre That script sounds really useful :) I'll make sure to put this memcache-key-prefix change in a different PR, since this PR has already been merged.

PR #6958, opened 2015-04-16
CE-1625 Add subpages model classes

adamkarminski:
@lukaszkonieczny


extensions/wikia/Insights/models/InsightsQuerypageModel.php
```diff
@@ -0,0 +1,68 @@
...
+ */
+abstract class InsightsQuerypageModel extends InsightsModel {
+	private $queryPageInstance,
+		$template = 'subpageList';
+
+	public
+		$offset = 0,
+		$limit = 100;
+
+	abstract function getDataProvider();
```
lukaszkonieczny: Did you consider moving vars and function (between lines 9-16) to InsightModel?

adamkarminski: All of them are part of the subpages based on QueryPage. I'd rather keep the InsightsModel minimal. We can think about refactoring InsightsModel to an Interface since it has only abstract methods.

PR #6959, opened 2015-04-16
Monetization (MON-379): MonetizationModule - updated module width

saipetch:
Monetization (MON-379): MonetizationModule - updated module width for breakpoints change


tedgill:
LGTM

Breakpoints are not going out today...so no need to rush it.

PR #6962, opened 2015-04-16
SOC-605 Implement blog and article comments using the new Email extension

garthwebb:
https://wikia-inc.atlassian.net/browse/SOC-605


extensions/wikia/GlobalWatchlist/GlobalWatchlistBot.class.php
```diff
@@ -9,7 +9,8 @@ class GlobalWatchlistBot {
 	const REPLY_ADDRESS = 'noreply@wikia.com';
 
 	public function __construct() {
-		\F::app()->wg->ExtensionMessagesFiles['GlobalWatchlist'] = dirname( __FILE__ ) . '/GlobalWatchlist.i18n.php';
+		$messageFiles = \F::app()->wg->ExtensionMessagesFiles;
+		$messageFiles['GlobalWatchlist'] = dirname( __FILE__ ) . '/GlobalWatchlist.i18n.php';
```
garthwebb: The line as is was throwing this error in dev:
```php
Indirect modification of overloaded property WikiaGlobalRegistry::$ExtensionMessagesFiles has no effect in /usr/wikia/source/app/extensions/wikia/GlobalWatchlist/GlobalWatchlistBot.class.php on line 12
```
Even though its just a notice, it was somehow randomly killing requests

extensions/wikia/ArticleComments/classes/ArticleComment.class.php
```diff
@@ -13,18 +13,22 @@ class ArticleComment {
 	const CACHE_VERSION = 2;
```
garthwebb: There are a lot of spacing fixes introduced in this class via the stylize.php script.  I also changed array() => [].  Sorry for the length ...

extensions/wikia/ArticleComments/classes/ArticleComment.class.php
```diff
@@ -94,6 +98,52 @@ static public function newFromArticle( Article $article ) {
...
+		if ( empty( $param['useSlave'] ) ) {
+			$dbh = wfGetDB( DB_MASTER );
+			$flags = Title::GAID_FOR_UPDATE;
+		} else {
+			$dbh = wfGetDB( DB_SLAVE );
+			$flags = 0;
+		}
+
+		$titleText = $title->getDBkey();
+		$prefix =  $titleText . '/' . ARTICLECOMMENT_PREFIX;
```
rafalkalinski: this is probably a good place for implode finction

garthwebb: For joining two, possibly three things together I think simple concatenation is more readable than an implode.

extensions/wikia/Email/Email.i18n.php
```diff
@@ -3,6 +3,7 @@
 
 $messages['en'] = array(
 	'emailext-desc' => 'An extension to handle delivering email',
+	'emailext-anonymous-editor' => 'A Wikia fan',
```
garthwebb: This message was moved from WatchedPage.i18n.php since it could be applied to multiple email types.

extensions/wikia/Email/EmailController.class.php
```diff
@@ -28,6 +31,12 @@
 	/**	@var string The language to send the email in */
 	protected $targetLang;
```
garthwebb: The following two members were pulled down from the WatchedPageController since they are common to all emails sent.

extensions/wikia/Email/EmailController.class.php
```diff
@@ -145,13 +173,24 @@ public function main() {
 	/**
 	 * Create an error response for any exception thrown while creating this email
 	 *
-	 * @param ControllerException $e
+	 * @param \Exception $e
 	 *
 	 */
-	protected function setErrorResponse( ControllerException $e ) {
+	protected function setErrorResponse( \Exception $e ) {
```
garthwebb: So it turns out that Nirvana change an exception thrown (say one based on our ```Email\ControllerException```) and re-throw it as its own ```WikiaDispatchedException```.  Given that we use the Nirvana request routing a few times during email creation, we need to catch all errors.

extensions/wikia/Email/templates/avatarLayout.mustache
```diff
@@ -1,6 +1,6 @@
 <div class="greeting">
 	<p>{{salutation}}</p>
-	<p>{{{articleEditedText}}}</p>
+	<p>{{{summary}}}</p>
```
garthwebb: Gave this and the variable below a more generic name, since the current names are somewhat specific to the WatchedPage email.

includes/EmailNotification.php
```diff
@@ -458,30 +455,68 @@ private function emailExtensionEnabled() {
...
+		} elseif ( $this->isBlogComment() ) {
+			$controller = 'Email\Controller\BlogComment';
+		}
+
+		if ( !empty( $controller ) ) {
+			$params = [
 				'targetUser' => $user->getName(),
 				'title' => $this->title->getText(),
-				'nameSpace' => $this->title->getNamespace(),
+				'namespace' => $this->title->getNamespace(),
```
garthwebb: This was a merge conflict.  Kept the all lowercased version since ```namespace``` is one word.

includes/UserMailer.php
```diff
@@ -146,21 +146,32 @@ static function makeMsgId() {
 	 * array of parameters. It requires PEAR:Mail to do that.
 	 * Otherwise it just uses the standard PHP 'mail' function.
 	 *
```
garthwebb: Several fixes to doc strings ...

includes/UserMailer.php
```diff
@@ -146,21 +146,32 @@ static function makeMsgId() {
...
+		MailAddress $to,
+		MailAddress $from,
+		$subject,
+		$body,
+		MailAddress $replyTo = null,
+		$contentType = null,
+		$category = 'UserMailer',
+		$priority = 0,
+		$attachments = []
+	) {
```
garthwebb: Simply split this to several lines since its very long and hard to follow.

includes/UserMailer.php
```diff
@@ -209,14 +220,14 @@ public static function send( $to, $from, $subject, $body, $replyto = null, $cont
 		$headers['From'] = $from->toString();
 		$headers['Return-Path'] = $from->address;
 
-		if ( $replyto && $replyto instanceof MailAddress ) {
-			$headers['Reply-To'] = $replyto->toString();
+		if ( $replyTo && $replyTo instanceof MailAddress ) {
+			$headers['Reply-To'] = $replyTo->toString();
```
garthwebb: Applying casing rules.

includes/UserMailer.php
```diff
@@ -209,14 +220,14 @@ public static function send( $to, $from, $subject, $body, $replyto = null, $cont
...
 		}
 
 		$headers['Date'] = date( 'r' );
 		$headers['MIME-Version'] = '1.0';
 
-		if ( empty( $attachements ) ) {
+		if ( empty( $attachments ) ) {
```
garthwebb: Misspelled var.

extensions/wikia/GlobalWatchlist/GlobalWatchlistBot.class.php
jsutterfield: Oh, hmm. Yea, when the bot is instantiated it's including the message files. Incorrectly at that. Looks like that was my change. Sorry about that. Thanks for fixing it.

But also, no, it doesn't really do anything. 1.) Because we're already including the file [here](https://github.com/Wikia/app/blob/dev/extensions/wikia/GlobalWatchlist/GlobalWatchlist.php#L21) during startup. And 2.) because most of these messages are in the MediaWiki namespace on messaging wiki which trumps this file. I'm going to remove that include from the set-up file since these messages are only need when the bot is run.

includes/context/RequestContext.php
```diff
@@ -327,7 +327,7 @@ public function setSkin( Skin $s ) {
 	/**
 	 * Get the Skin object
 	 *
-	 * @return Skin
+	 * @return Skin|Linker
```
garthwebb: The object returned is always of type ```Skin``` but this code contains a ```__call``` method that sends unmatched method calls to ```Linker``` methods.  Adding this hinting allows PHPStorm to find method definitions properly.  This kind of ```__call``` usage is essentially multiple inheritance.

includes/EmailNotification.php
```diff
@@ -458,30 +455,68 @@ private function emailExtensionEnabled() {
...
 	private function sendUsingEmailExtension( \User $user ) {
-		F::app()->sendRequest( 'Email\Controller\WatchedPage', 'handle',
-			[
+
+		if ( $this->isArticlePageEdit() ) {
+			$controller = 'Email\Controller\WatchedPage';
+		} elseif ( $this->isArticleComment() ) {
+			$controller = 'Email\Controller\ArticleComment';
+		} elseif ( $this->isBlogComment() ) {
+			$controller = 'Email\Controller\BlogComment';
```
jsutterfield: So this doesn't actually cover all cases. Adding a category also goes through this method (`$this->action === "categoryadd"`), as does the email for a create page (`empty( $this->action ) && $this->isNewPage();`) so those emails will get skipped by this current logic. Those are the only other 2 that I know of, but given the wilyness of MediaWiki, there could be others.

We should make sure that the category add and created page emails still flows through the `sendUsingUserMailer()` method, and also maybe log the $action when the controller is empty on L471 so we can see if we're missing anything else?

extensions/wikia/Follow/FollowHelper.class.php
```diff
@@ -220,7 +223,7 @@ static public function watchBlogListing(&$article, &$user, $text, $summary, $min
 				$title = Title::makeTitle( NS_BLOG_LISTING, $title );
 				$related[] = ucfirst($title->getDBKey());
 			}
-			self::emailNotification($article->getTitle(), $related, NS_BLOG_LISTING, $user, "blogpost", wfMsg('follow-bloglisting-summary'));
+			self::emailNotification($article->getTitle(), $related, NS_BLOG_LISTING, $user, self::LOG_ACTION_BLOG_POST, wfMsg('follow-bloglisting-summary'));
```
jsutterfield: Since we're touching this line, how about updating wfMsg?

extensions/wikia/Email/i18n/Comment.i18n.php
```diff
@@ -0,0 +1,18 @@
+<?php
+$messages = [];
+
+$messages['en'] = [
+	'emailext-articlecomment-summary' => 'People are talking about $1 on {{SITENAME}}!',
+	'emailext-blogcomment-summary' => 'People are talking about the $1 post on {{SITENAME}}!',
+	'emailext-comment-salutation' => 'Hi $1,',
+	'emailext-comment-link-label' => 'See this comment',
+	'emailext-comment-view-all' => "[$1 View all comments]",
```
jsutterfield: Do we have a ticket to get these translated?

extensions/wikia/Email/EmailController.class.php
```diff
@@ -344,6 +376,35 @@ private function getUnsubscribeLink() {
 		return $unsubscribeTitle->getFullURL( $params );
 	}
 
+
+	/**
+	 * @return String
+	 */
+	protected function getCurrentProfilePage() {
+		if ( $this->currentUser->isAnon() ) {
```
jsutterfield: Don't we want to check if the currentUser is not anon?

extensions/wikia/Email/EmailController.class.php
```diff
@@ -344,6 +376,35 @@ private function getUnsubscribeLink() {
...
+			return $this->currentUser->getUserPage()->getFullURL();
+		}
+		return "";
+	}
+
+	/**
+	 * @return String
+	 */
+	protected function getCurrentUserName() {
+		if ( $this->currentUser->isAnon() )	 {
```
jsutterfield: Same here, shouldn't this be `!$this->currentUser->isAnon()`?

extensions/wikia/Email/Email.i18n.php
mklucsarits: Can the capitalization issue in https://wikia-inc.atlassian.net/browse/SOC-641 be handled as part of your changes here?

garthwebb: @mklucsarits good point.  I'm on the edge of whether to do it though since it seems like she may want two different behaviors out this string which I'm inclined to handle separately.  I'll see what she says.

extensions/wikia/Email/EmailController.class.php
garthwebb: Yup.  I don't really remember why I changed this from ```isLoggedIn```

extensions/wikia/Email/i18n/Comment.i18n.php
garthwebb: No, I need to do that.

garthwebb: And now, witness the glory that is LOC-28: https://wikia-inc.atlassian.net/browse/LOC-28

extensions/wikia/Follow/FollowHelper.class.php
garthwebb: Sure, and while I'm at it I'll UPDATE SPACING FOR THE WHOLE FILE!  And add some comments.  While I'm at it.

includes/EmailNotification.php
garthwebb: Yah, that was my intent but I think I lost it when moving things around.

includes/EmailNotification.php
```diff
@@ -496,44 +541,45 @@ private function sendUsingUserMailer( \User $user ) {
...
 	private function sendPersonalised( $watchingUser ) {
-		// From the PHP manual:
-		//     Note:  The to parameter cannot be an address in the form of "Something <someone@example.com>".
-		//     The mail command will not parse this properly while talking with the MTA.
 		$to = new MailAddress( $watchingUser );
-
-		// $PAGEEDITDATE is the time and date of the page change
-		// expressed in terms of individual local time of the notification
-		// recipient, i.e. watching user
-		// TODO The logic below is duplicated twice, break this off into a separate method.
```
garthwebb: Actually, it was only duplicated once ...

extensions/wikia/Follow/FollowHelper.class.php
jsutterfield: LOL

jsutterfield:
LGTM

extensions/wikia/Email/Email.i18n.php
garthwebb: @mklucsarits ok, I'm going to let this one be for this ticket and not confuse things.

extensions/wikia/Email/i18n/Comment.i18n.php
garthwebb: Ok, LOC-28 got ... lost in translation.  Here's the new ticket, LOC-32: https://wikia-inc.atlassian.net/browse/LOC-32

includes/EmailNotification.php
garthwebb: Ok, I think this should be addressed by the changes made in the calling function.

extensions/wikia/Follow/FollowHelper.class.php
garthwebb: Hmm, somehow with all this I didn't actually update wfMsg...

PR #6967, opened 2015-04-17
CE-1624 Focus on next page button on notification

kamilkoterba:
https://wikia-inc.atlassian.net/browse/CE-1624

We want next page button on notification be focused on page load

@Wikia/community-engineering


lukaszkonieczny:
lgtm

PR #6968, opened 2015-04-19
Added prefix to key so that memcached hit-rates can be tracked for these...

SeanColombo:
... keys. It'll likely be a pretty low hit-rate (one key per user per chat session).


macbre:
LGTM

macbre:
As for the hit ratio, it's not that low:

```
------------------------------------------------------------------------------------------------------------------------
Key                                                                                                  Hit ratio [%]
------------------------------------------------------------------------------------------------------------------------
*                                                                                                    26.01
```

SeanColombo:
Nice :)

PR #6972, opened 2015-04-20
CONCF-437 don't use flexbox for IE - use divs with set widths

bognix:


bognix:
ping @rogatty @Warkot

Warkot:
I actually removed all of the declarations, but `word-wrap: break-word;` (and `width: 175px;`, but that's just cosmetics) and it's still working for me on my IE 11 on Win 7 in VirtualBox. Is everything else needed? Is it maybe for IE 10?

rogatty:
LGTM

PR #6974, opened 2015-04-20
SOC-656 Load Facebook SDK on wikiamobile's Special:UserLogin

bkoval:
@hakubo @kvas-damian @jsutterfield Please, CR.

Explanation:

This is an attempt to make FB Connect login usable on WIkiaMobile's Special:UserLogin, as we need it for the New Login MVP.

https://wikia-inc.atlassian.net/browse/SOC-656

### DIFF
--- a/extensions/wikia/UserLogin/FacebookButtonController.class.php
+++ b/extensions/wikia/UserLogin/FacebookButtonController.class.php
@@ -16,7 +16,7 @@ class FacebookButtonController extends WikiaController {
 		if ( $this->app->checkSkin( 'wikiamobile' ) ) {
 			JSMessages::registerPackage( 'fblogin' , array( 'wikiamobile-facebook-connect-fail' ) );
 
-			if ( F::app()->wg->Title->isSpecial('Userlogin') ) {
+			if ( $this->wg->Title->isSpecial('Userlogin') ) {
 				JSMessages::enqueuePackage( 'fblogin', 'inline' );
 			}
 
--- a/extensions/wikia/UserLogin/FacebookButtonController.class.php
+++ b/extensions/wikia/UserLogin/FacebookButtonController.class.php
@@ -15,7 +15,7 @@ class FacebookButtonController extends WikiaController {
 
 		if ( $this->app->checkSkin( 'wikiamobile' ) ) {
 			JSMessages::registerPackage( 'fblogin' , array( 'wikiamobile-facebook-connect-fail' ) );
-			JSMessages::enqueuePackage( 'fblogin', 'inline' );
+			JSMessages::enqueuePackage( 'fblogin', JSMessages::INLINE );
 
 			$this->response->setVal( 'requestType', $this->request->getVal( 'requestType' ) );
 			$this->overrideTemplate( 'WikiaMobileIndex' );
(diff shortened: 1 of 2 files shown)


kvas-damian:
LGTM

jsutterfield:
If the user's FB account isn't already connected to Wikia, they'll receive an error message which shows the key 'wikiamobile-facebook-connect-fail', rather than the message. I think it's because you need to enqueue the messages. Maybe [here](https://github.com/Wikia/app/blob/dev/extensions/wikia/UserLogin/FacebookButtonController.class.php#L17)?

lizlux:
Is there a reason you didn't use `$.loadFacebookAPI`? We're trying to centralize all facebook library loading with that function. One of the benefits is that it provides a `fail()` method if the library doesn't load. I'm not sure that loader has that. Either way it's good to funnel all this functionality through one interface.

bkoval:
@lizlux Yes, the loader provides a 'fail' method as well. I used it here to be compliant with the way it's loaded in topbar.js for wikiamobile as it uses loader to download both our js and js directly from connect.facebook.net. I'll investigate how log would it take to refactor to use your way, if it's too much work I'll make another ticket for that.

@jsutterfield right. I'll try downloading the messages by loader :dancer:

lizlux:
@bkoval as long as the fail method is actually called when the resource fails to load (like when FB is blocked in China), then it should be okay for now, although I'd still prefer using the same method for loading this library everywhere. `$.getResources`, for example, pretended to have a `fail` handler but it didn't actually work.

lizlux:
Although I guess we don't need a fail method here because if it fails, we don't do anything. This is just something to keep in mind in the future.

extensions/wikia/UserLogin/FacebookButtonController.class.php
```diff
@@ -15,6 +15,8 @@ public function index() {
 
 		if ( $this->app->checkSkin( 'wikiamobile' ) ) {
 			JSMessages::registerPackage( 'fblogin' , array( 'wikiamobile-facebook-connect-fail' ) );
+			JSMessages::enqueuePackage( 'fblogin', 'inline' );
```
jsutterfield: Hmm, this still isn't working for me. I think it's because this second argument is expected to be an integer, not a string. There's a class constant we should probably be using anyways. If I update this to the following, then it seems work:

```php
JSMessages::enqueuePackage( 'fblogin', JSMessages::INLINE );
```

extensions/wikia/UserLogin/js/UserLoginSpecial.wikiamobile.js
```diff
@@ -1,10 +1,35 @@
...
+	'use strict';
+
+	function loadFacebookLoginScripts() {
+		loader({
+			type: loader.LIBRARY,
+			resources: 'facebook'
+		}, {
+			type: loader.MULTI,
+			resources: {
+				messages: 'fblogin',
```
jsutterfield: This line doesn't seem to be necessary since we're already loading the message in FacebookButtonController.

We could remove `JSMessages::enqueuePackage( 'fblogin', 'inline' );` from the controller and move  `JSMessages::registerPackage( 'fblogin' , array( 'wikiamobile-facebook-connect-fail' ) );` to UserLogin.setup.php so that it would be available asynchronously like this, but then we'd be registering that package for every request to MW. I say we just remove this line.

bkoval:
@jsutterfield  Agreed. I made the changes :)

jsutterfield:
LGTM!

PR #6976, opened 2015-04-20
Monetization (MON-379): MonetizationModule - updated styles for breakpoints change

saipetch:
Monetization (MON-379): MonetizationModule - separate styles for breakpoints change


tedgill:
:ship:

PR #6980, opened 2015-04-21
SOC-628 Make sure to construct the icon URLs we use so we can update and cache bust them

garthwebb:
Also, add an explicit height and width in case (as has happened for some reason) the default image URL gets cached with a larger size.

https://wikia-inc.atlassian.net/browse/SOC-628

### DIFF
--- a/extensions/wikia/Email/EmailImageHelper.class.php
+++ b/extensions/wikia/Email/EmailImageHelper.class.php
@@ -8,6 +8,10 @@ namespace Email;
  * @package Email
  */
 class ImageHelper {
+
+	// Cache icon information for a day
+	const ICON_CACHE_TTL = 86400;
+
 	static public $icons = [
 		'Wikia',
 		'Comics',
@@ -30,19 +34,38 @@ class ImageHelper {
 		$info = [];
 
 		foreach ( self::$icons as $name ) {
+			$fileInfo = self::getFileInfo( $name );
+			$info[$name] = $fileInfo;
+		}
+
+		return $info;
+	}
+
+	protected static function getFileInfo( $name ) {
+		$key = self::getFileKey( $name );
+		$memc = \F::app()->wg->Memc;
+
+		$info = $memc->get( $key );
+		if ( empty( $info ) ) {
 			$file   = \GlobalFile::newFromText( $name . '.gif', \Wikia::NEWSLETTER_WIKI_ID );
 			$width  = $file->getWidth();
 			$height = $file->getHeight();
 			$url    = $file->getUrlGenerator()->url();
 
-			$info[$name] = [
-				'name' => $name,
-				'url' => $url,
+			$info = [
+				'name'   => $name,
+				'url'    => $url,
 				'height' => $height,
-				'width' => $width,
+				'width'  => $width,
 			];
+
+			$memc->set( $key, $info, self::ICON_CACHE_TTL );
 		}
 
 		return $info;
+	}
+
+	protected static function getFileKey( $name ) {
+		return wfSharedMemcKey( 'Email', 'ImageHelper', 'icon', $name );
 	}
 }
--- a/extensions/wikia/Email/EmailImageHelper.class.php
+++ b/extensions/wikia/Email/EmailImageHelper.class.php
@@ -42,25 +42,20 @@ class ImageHelper {
 	}
 
 	protected static function getFileInfo( $name ) {
-		$key = self::getFileKey( $name );
-		$memc = \F::app()->wg->Memc;
+		$info = \WikiaDataAccess::cache(
+			self::getFileKey( $name ),
+			self::ICON_CACHE_TTL,
+			function() use ( $name ) {
+				$file = \GlobalFile::newFromText( $name . '.gif', \Wikia::NEWSLETTER_WIKI_ID );
 
-		$info = $memc->get( $key );
-		if ( empty( $info ) ) {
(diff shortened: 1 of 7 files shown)


extensions/wikia/Email/templates/main.mustache
```diff
@@ -42,44 +42,45 @@
...
 							</a>
 							<a href="{{socialMessages.facebook-link}}" >
-								<img src="http://vignette2.wikia.nocookie.net/wikianewsletter/images/a/ad/Facebook.gif/revision/latest?cb=20150330185226" alt="{{socialMessages.facebook}}">
+                                {{# icons.Facebook}}<img src="{{url}}" width="{{width}}" height="{{height}}" alt="{{socialMessages.facebook}}">{{/ icons.Facebook}}
 							</a>
 							<a href="{{socialMessages.youtube-link}}" >
-								<img src="http://vignette4.wikia.nocookie.net/wikianewsletter/images/1/12/You-Tube.gif/revision/latest?cb=20150330185252" alt="{{socialMessages.youtube}}">
+                                {{# icons.YouTube}}<img src="{{url}}" width="{{width}}" height="{{height}}" alt="{{socialMessages.youtube}}">{{/ icons.YouTube}}
 							</a>
+                            <pre>{{iconURL}}</pre>
```
kvas-damian: I'm not sure why it's here? Is it leftover from debugging?

extensions/wikia/Email/EmailImageHelper.class.php
```diff
@@ -0,0 +1,48 @@
...
+	/**
+	 * Get URL, size and name information about icons created in the newsletter wikia for our emails
+	 *
+	 * @return array
+	 */
+	public static function getIconInfo() {
+		$info = [];
+
+		foreach ( self::$icons as $name ) {
+			$file   = \GlobalFile::newFromText( $name . '.gif', \Wikia::NEWSLETTER_WIKI_ID );
```
kvas-damian: ```GlobalFile``` isn't cached in memcache - this means that every email will perform ~10 additional DB queries. What do you think about caching them?

extensions/wikia/Email/templates/main.mustache
```diff
@@ -42,44 +42,45 @@
...
 							<a href="{{hubsMessages.booksURL}}">
-								{{# marketingFooter}}<img src="http://vignette4.wikia.nocookie.net/wikianewsletter/images/c/c2/Books.gif/revision/latest?cb=20150330185117" class="hub-icon">{{/ marketingFooter}}
+								{{# marketingFooter}}{{# icons.Books}}<img src="{{url}}" width="{{width}}" height="{{height}}" class="hub-icon">{{/ icons.Books}}{{/ marketingFooter}}
 								{{hubsMessages.books}}
 							</a>
 							<a href="{{hubsMessages.tvURL}}">
-								{{# marketingFooter}}<img src="http://vignette1.wikia.nocookie.net/wikianewsletter/images/c/c5/TV.gif/revision/latest?cb=20150330185219" class="hub-icon">{{/ marketingFooter}}
+								{{# marketingFooter}}{{# icons.TV}}<img src="{{url}}" width="{{width}}" height="{{height}}" class="hub-icon">{{/ icons.TV}}{{/ marketingFooter}}
 								{{hubsMessages.tv}}
 							</a>
```
kvas-damian: Those blocks of HTML looks pretty the same. What do you think about creating a loop that will generate all links?

bkoval:
LGTM besides Damian's comments.

extensions/wikia/Email/EmailImageHelper.class.php
garthwebb: Good point.  I'll add that.

extensions/wikia/Email/templates/main.mustache
garthwebb: I have a feeling Front End would prefer things like this stay expanded so they more closely represent the HTML output rather than code.  @lizlux (since you worked on this), @bkoval, thoughts on this?

extensions/wikia/Email/templates/main.mustache
garthwebb: yah :/

extensions/wikia/Email/templates/main.mustache
lizlux: Agreed that this is a lot of repeat code, and could be done with a loop, however our Mustache implementation isn't always the easiest to work with when it comes to partials, which is probably why I didn't bother to do it in the first place.

extensions/wikia/Email/EmailImageHelper.class.php
```diff
@@ -0,0 +1,71 @@
...
+		}
+
+		return $info;
+	}
+
+	protected static function getFileInfo( $name ) {
+		$key = self::getFileKey( $name );
+		$memc = \F::app()->wg->Memc;
+
+		$info = $memc->get( $key );
```
kvas-damian: What do you think about using WikiaDataAccess for caching? 
https://github.com/Wikia/app/blob/dev/includes/wikia/WikiaDataAccess.class.php#L52

extensions/wikia/Email/templates/main.mustache
kvas-damian: Ok, lets keep it as it is.

extensions/wikia/Email/EmailImageHelper.class.php
garthwebb: Sounds good!

PR #6990, opened 2015-04-21
ADEN-1946 Paid Asset Drop for Mercury

gabrys:


extensions/wikia/PaidAssetDrop/js/run/desktop.run.js
```diff
@@ -25,8 +26,8 @@ require([
 			return;
 		}
 
-		if (pad.isNowValid() && win.wgEnableAPI) {
-			pad.injectPAD();
+		if (pad.isNowValid(win.wgPaidAssetDropConfig) && win.wgEnableAPI) {
```
nandy-andy: I wonder if we shouldn't use AdContext here as well. It'll be consistent with mobile and if something's wrong with AdContext it'll be broken on both skins - we'll hear about it as soon as possible :japanese_ogre:

PR #6996, opened 2015-04-22
Prettify JSON data for CollectionView

rogatty:
@Warkot @dianafa @bognix

### DIFF
--- a/extensions/wikia/CollectionView/controllers/CollectionViewParserTagController.class.php
+++ b/extensions/wikia/CollectionView/controllers/CollectionViewParserTagController.class.php
@@ -45,10 +45,35 @@ class CollectionViewParserTagController extends WikiaController {
 
 	protected function saveToParserOutput( \ParserOutput $parserOutput, $raw ) {
 		if ( !empty( $raw ) ) {
-			//TODO we should convert $raw to the format described in CONCF-460
 			$collections = $parserOutput->getProperty( self::COLLECTIONS_PROPERTY_NAME );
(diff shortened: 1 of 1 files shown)


extensions/wikia/CollectionView/controllers/CollectionViewParserTagController.class.php
```diff
@@ -45,10 +45,35 @@ public function renderCollectionView( $text, $params, $parser, $frame ) {
...
+
+		foreach ( $raw as $item ) {
+			if ( !$item[ 'isEmpty' ] ) {
+				switch ( $item[ 'type' ] ) {
+					case 'header':
+						$data[ 'header' ] = $item[ 'data' ][ 'value' ];
+						break;
+					case 'item':
+						$data[ 'items' ][] = $item[ 'data' ];
+						break;
```
Warkot: Maybe you could add `default` like this:
```
default:
	$data[ $item[ 'type' ] ][] = $item[ 'data' ];
```

rogatty: I think it's safer to include only the types we explicitly support.

Warkot: It surely is. The non-readable ones will be dropped this way, so it will save some bandwidth if any occur. ;)

Warkot:
LGTM :heavy_check_mark:

PR #6999, opened 2015-04-22
SOC-675 Don't send recent change notification to self

mklucsarits:
https://wikia-inc.atlassian.net/browse/SOC-675

This prevents email notifications to users about their own edits.

### DIFF
--- a/includes/RecentChange.php
+++ b/includes/RecentChange.php
@@ -228,8 +228,10 @@ class RecentChange {
 		global $wgUseEnotif, $wgShowUpdatedMarker, $wgUser;
 		if( $wgUseEnotif || $wgShowUpdatedMarker ) {
 			// Users
-			if ( $this->mAttribs['rc_user'] && $this->mAttribs['rc_user'] !== $wgUser->getId() ) {
-				$editor = User::newFromID( $this->mAttribs['rc_user'] );
+			if( $this->mAttribs['rc_user'] ) {
+				if ( $wgUser->getId() !== $this->mAttribs['rc_user'] ) {
+					$editor = User::newFromID( $this->mAttribs['rc_user'] );
+				}
 			// Anons
 			} else {
 				$editor = ($wgUser->getName() == $this->mAttribs['rc_user_text']) ?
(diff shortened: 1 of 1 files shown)


mklucsarits:
@garthwebb and @jsutterfield please have a look.

includes/RecentChange.php
```diff
@@ -229,24 +229,28 @@ public function save( $noudp = false ) {
...
-				$editor = ($wgUser->getId() == $this->mAttribs['rc_user']) ?
-					$wgUser : User::newFromID( $this->mAttribs['rc_user'] );
+				if ( $wgUser->getId() !== $this->mAttribs['rc_user'] ) {
+					$editor = User::newFromID( $this->mAttribs['rc_user'] );
+				}
 			// Anons
 			} else {
 				$editor = ($wgUser->getName() == $this->mAttribs['rc_user_text']) ?
 					$wgUser : User::newFromName( $this->mAttribs['rc_user_text'], false );
 			}
```
armonr: Can this be simplified to the following? It is so difficult to read, and it is not your fault, Matt :)
```php
$editor = null;
if ( isset( $this->mAttribs['rc_user'] ) && $wgUser->getId() !== $this->mAttribs['rc_user'] ) {
	$editor = User::newFromID( $this->mAttribs['rc_user'] );
} else if ( isset( $this->mAttribs['rc_user_text'] ) ) {
	// Same expression as before ...
}

if ( $editor ) {
...
```

garthwebb:
@mklucsarits, so this will not call ```$enotif->notifyOnPageChange()``` if the current user is the same as the user who made the change.  The problem with this is that this method handles all notification and we still want to notify everyone else who's watching the page; just not the editor.

I have an idea where this bug may have been introduced.  This class used to be defined in the UserMailer.php file where @jsutterfield cleaned up the ```canSendUserTalkEmail``` method:

https://github.com/Wikia/app/commit/8f36ae64d80eb05445bdcaf76758b306cbd2b872#diff-2b0526d8f9a57469c6b27246054dfceaL650

Later in the code review process I suggested to James that he expand that logic to multiple lines giving:

https://github.com/Wikia/app/commit/1a4779539a17bd35b04b9c1cf8923da076b5afa1#diff-2b0526d8f9a57469c6b27246054dfceaL574

I suspect that between one of those two transformations, we lost the part that keeps us from sending an email to just the user who created the change.

mklucsarits:
Thanks, I found the problem. I'm going to just recreate this branch.

PR #7002, opened 2015-04-23
SOC-674 log which user properties are being used

jsutterfield:
As part of the User Attribute service work, it would be very helpful to know which attributes are currently being used given the user_properties table has 2,535,219 unique entries. Add logging to the getOption method of the User object to see which of the properties are being used, and to what extent.

Ticket: https://wikia-inc.atlassian.net/browse/SOC-674


rafalkalinski:
LGTM

mklucsarits:
LGTM. Good documentation!

macbre:
This floods es at the rate of 3,5mm entries a minute. Reverting

PR #7010, opened 2015-04-23
CONCF-483 Launch social buttons international

dianafa:
- wgEnablePageShareWorldwide set to true 
- show LINK only for touch screen devices (and japan language)
- fix tests

ping @bognix @Warkot


hakubo:
LGTM

bognix:
LGTM

extensions/wikia/PageShare/PageShareHelper.class.php
```diff
@@ -46,7 +47,11 @@ public static function getLangForPageShare( $browserLang, $useLang ) {
 		}
 	}
 
-	public static function isValidShareService( $service, $lang ) {
+	public static function isValidShareService( $service, $lang, $isTouchScreen ) {
+		//Don't display LINE social network (because it's a mobile app) on desktops
+		if ( $isTouchScreen === 'false' && ( $service['name'] === self::LINE_NAME ) ) {
```
Warkot: The inner pair of parentheses is redundant.
```		if ( $isTouchScreen === 'false' && $service['name'] === self::LINE_NAME ) {```
(ref. http://php.net/manual/en/language.operators.precedence.php)

Comment (Warkot) [inline extensions/wikia/PageShare/PageShareHelper.class.php]: And do you really mean **string `'false'`**?

Comment (Warkot) [inline extensions/wikia/PageShare/PageShareHelper.class.php]: If yes, then why not to use `!$isTouchScreen`?

File: extensions/wikia/PageShare/tests/PageShareHelperTest.php [hunk 2]
```diff
@@ -18,6 +18,7 @@ public function IsValidShareServiceProvider() {
 					'languages:exclude' => [],
 				],
 				'language' => 'en',
+				'isTouchScreen' => 'false',
```
Warkot: You're using everywhere here **string `'false'`**, are you sure? It looks at least strange for me. :frowning:

extensions/wikia/PageShare/PageShareHelper.class.php
dianafa: yes, it comes as a param - string. Thus evaluating both 'true' and 'false' strings as a bool values always gives true... so `!$isTouchScreen` was always evaluating to false.

Warkot: Then maybe you could use `$isTouchScreen = $this->request->getBool( 'isTouchScreen' );` from https://github.com/Wikia/app/blob/dev/includes/wikia/nirvana/WikiaRequest.class.php#L105? Any external opinion on that, @bognix?

PR #7011, opened 2015-04-23
CE-1641: Disable create new wiki popup in special Insights

lukaszkonieczny:


extensions/wikia/CreatePage/CreatePage.php
```diff
@@ -89,7 +89,9 @@ function wfCreatePageSetupVars(Array &$vars ) {
 		$contentNamespaces[] = $wgContLang->getNsText( $contentNs );
 	}
 
-	$vars['WikiaEnableNewCreatepage'] = $wgUser->getOption( 'createpagepopupdisabled', false ) ? false : $wgWikiaEnableNewCreatepageExt;
+	if ( !isset( $vars['WikiaEnableNewCreatepage'] ) ) {
```
kamilkoterba: Line of motivation would be welcomed here to be more explicit why we want to omit setting this property

extensions/wikia/Insights/InsightsHooks.class.php
```diff
@@ -60,4 +60,18 @@ public static function onGetLocalURL( &$this, &$url, $query ) {
 
 		return true;
 	}
+
+	/**
+	 * Disable create new page popup
```
kamilkoterba: I think we can be more explicit here as well and explain why we're disabling popup on insights page.

kamilkoterba:
General future comment: please remember about adding Jira issue link, description and motivation of changes.

PR #7012, opened 2015-04-23
SOC-396 Limit large queries for wall thread replies

mklucsarits:
https://wikia-inc.atlassian.net/browse/SOC-396

### DIFF
--- a/extensions/wikia/Wall/WallThread.class.php
+++ b/extensions/wikia/Wall/WallThread.class.php
@@ -11,12 +11,15 @@ class WallThread {
 	private $data;
 
 	public function __construct() {
+		$this->initializeReplyData();
+		$this->mCached = null;
+		$this->mCityId = F::app()->wg->CityId;
+	}
+
+	protected function initializeReplyData() {
 		$this->data = new StdClass();
 		$this->data->threadReplyIds = false;
 		$this->data->threadReplyObjs = false;
-
-		$this->mCached = null;
-		$this->mCityId = F::app()->wg->CityId;
 	}
 
 	/**
@@ -30,6 +33,7 @@ class WallThread {
 
 		return $wt;
 	}
+
 	public function loadIfCached() {
 		if($this->mCached === null) {
 			$this->loadFromMemcache();
@@ -53,9 +57,7 @@ class WallThread {
 
 	public function setReplies( $ids ) {
 		// set and cache replies of this thread
-		$this->data = new StdClass();
-		$this->data->threadReplyIds = false;
-		$this->data->threadReplyObjs = false;
+		$this->initializeReplyData();
 		
 		$this->data->threadReplyIds = $ids;
 
@@ -132,6 +134,8 @@ class WallThread {
 	public function invalidateCache() {
 		// invalidate cache at Thread level (new reply or reply removed in thread)
 		$this->getCache()->delete( $this->getThreadKey() );
+		// Reset data
+		$this->initializeReplyData();
 	}
 
 	private function getThreadKey() {
--- a/extensions/wikia/Wall/WallThread.class.php
+++ b/extensions/wikia/Wall/WallThread.class.php
@@ -1,6 +1,8 @@
 <?php
 
 class WallThread {
+	const DB_LIMIT_REPLIES = 4;
+
 	private $mThreadId = false;
 	private $mCached = null;
 	private $mForceMaster = false;
@@ -61,8 +63,10 @@ class WallThread {
 	}
 
 	private function loadReplyObjs() {
-		if( $this->data->threadReplyIds === false )
+		if( $this->data->threadReplyIds === false ) {
 			$this->loadReplyIdsFromDB();
+		}
+
 		$this->data->threadReplyObjs = array();
 
 		if(empty($this->data->threadReplyIds)) {
@@ -77,31 +81,24 @@ class WallThread {
 		}
 	}
 
-	private function loadReplyIdsFromDB($master = false) {
+	private function getReplyIdsFromDB( $dbr, $afterId = null ) {
 		// this is a direct way to get IDs
 		// the other one is in Wall.class done in a grouped way
 		// (fetch for many threads at once, set with ->setReplies)
 
-		$title = Title::newFromId( $this->mThreadId );
+		$conditions = array( 'parent_comment_id = '.$this->mThreadId );
 
-		if( empty($title) ) {
-			$title = Title::newFromId( $this->mThreadId, Title::GAID_FOR_UPDATE );
+		if ( !empty( $afterId ) ) {
+			array_push( $conditions, 'comment_id > '.$afterId );
 		}
-
-		if( empty($title) ) {
-			return ;
-		}
-
-		$dbr = wfGetDB( $master ? DB_MASTER : DB_SLAVE );
-
-		$threadId = $title->getArticleID();
 
 		$result = $dbr->select(
 				array( 'comments_index' ),
 				array( 'distinct comment_id' ),
-				array( 'parent_comment_id = '.$threadId ),
+				$conditions,
 				__METHOD__,
-				array( 'ORDER BY' => 'comment_id ASC' )
+				array( 'ORDER BY' => 'comment_id ASC',
+					'LIMIT' => self::DB_LIMIT_REPLIES )
 		);
 
 		$list = array();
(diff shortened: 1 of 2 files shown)


mklucsarits:
@armonr @garthwebb

extensions/wikia/Wall/WallThread.class.php
```diff
@@ -77,31 +81,31 @@ private function loadReplyObjs() {
...
-
 		$result = $dbr->select(
 				array( 'comments_index' ),
 				array( 'distinct comment_id' ),
-				array( 'parent_comment_id = '.$threadId ),
+				$conditions,
 				__METHOD__,
-				array( 'ORDER BY' => 'comment_id ASC' )
+				array( 'ORDER BY' => 'comment_id ASC',
+					'LIMIT' => self::DB_LIMIT_REPLIES )
```
armonr: You might as well use fluent sql. It has good test coverage and less error prone.

extensions/wikia/Wall/WallThread.class.php
```diff
@@ -77,31 +81,31 @@ private function loadReplyObjs() {
...
+			array_push( $conditions, 'comment_id > '.$afterId );
 		}
 
-		if( empty($title) ) {
-			return ;
-		}
-
-		$dbr = wfGetDB( $master ? DB_MASTER : DB_SLAVE );
-
-		$threadId = $title->getArticleID();
```
armonr: `$threadId` used to come from `$title`. This logic probably should remain the same.

extensions/wikia/Wall/WallThread.class.php
```diff
@@ -109,13 +113,25 @@ private function loadReplyIdsFromDB($master = false) {
...
+	}
+
+	private function loadReplyIdsFromDB( $master = false ) {
+		if ( empty( Title::newFromId( $this->mThreadId ) ) ) {
+			return;
+		}
+
+		$dbr = wfGetDB( $master ? DB_MASTER : DB_SLAVE );
+
+		$this->setReplies( $this->getReplyIdsFromDB( $dbr, null ) );
```
armonr: `null` is not needed here.

extensions/wikia/Wall/WallThread.class.php
```diff
@@ -109,13 +113,25 @@ private function loadReplyIdsFromDB($master = false) {
...
 	}
 
 	public function invalidateCache() {
 		// invalidate cache at Thread level (new reply or reply removed in thread)
-		$this->mForceMaster = true;
-		$this->loadReplyIdsFromDB( true );
+		$this->getCache()->delete($this->getThreadKey());
```
armonr: Whitespace before and after arguments.

extensions/wikia/Wall/WallThread.class.php
```diff
@@ -140,10 +156,7 @@ private function loadFromMemcache() {
 	}
 
 	private function saveToMemcache() {
-		$cache = $this->getCache();
-		$key = $this->getThreadKey();
-
-		$cache->set($key, $this->data);
+		$this->getCache()->set($this->getThreadKey(), $this->data);
```
armonr: Whitespace

extensions/wikia/Wall/WallThread.class.php
```diff
@@ -1,6 +1,8 @@
 <?php
 
 class WallThread {
+	const DB_LIMIT_REPLIES = 500;
```
armonr: Does this mean 500 comments do get loaded on the page? If so, is Cristina fine with this number?

Nitpicking here about name of the constant; how about `COMMENT_REPLY_LIMIT`?

mklucsarits: The limit is only the query limit for recursion. All replies will be loaded on the page; the user experience doesn't change.

extensions/wikia/Wall/WallThread.class.php
mklucsarits: I'm not familiar with FluentSQL, and it would take me a couple hours to learn it. And while that's worthwhile, I don't think it's worthwhile for this particular change. Unless you or another reviewer strongly disagree.

extensions/wikia/Wall/WallThread.class.php
mklucsarits: I discovered that $threadId is identical to $this->mThreadId, which contains the title ID. This call to getArticleId() was introduced here https://github.com/Wikia/app/commit/e5458fbb088cb519d82ac4eb3e90ba53877b7e29#diff-e8f2ddbfe2b36cb814f220f442ce868dL197 which I believe was done by mistake.

extensions/wikia/Wall/WallThread.class.php
```diff
@@ -1,6 +1,8 @@
 <?php
 
 class WallThread {
+	const DATABASE_LIMIT = 500;
```
mklucsarits: I'm still open to suggestions for a better name for this constant. I'm not sure if there's a standard convention for naming constants at Wikia.

extensions/wikia/Wall/WallThread.class.php
```diff
@@ -109,13 +113,25 @@ private function loadReplyIdsFromDB($master = false) {
 			$list[] = $row->comment_id;
 		}
 
-		$this->setReplies( $list );
+		$lastId = end( $list );
+
+		return empty( $lastId ) ? $list :
+			array_merge( $list, $this->getReplyIdsFromDB( $dbr, $lastId ) );
```
kvas-damian: Please avoid recursion when possible.
Usually it consumes a lot more of memory.

extensions/wikia/Wall/WallThread.class.php
```diff
@@ -77,31 +81,31 @@ private function loadReplyObjs() {
...
 
-	private function loadReplyIdsFromDB($master = false) {
+	/**
+	 * Fetches reply IDs for the thread, using a limit to control large queries
+	 *
+	 * @param DatabaseBase $dbr Database resource
+	 * @param integer $afterId The last reply ID after which the next set is selected
+	 * @return array List of reply IDs
+	 */
+	private function getReplyIdsFromDB( $dbr, $afterId = null ) {
```
kvas-damian: Did you consider using ```limit``` and ```offset```?
IMO ```$afterId``` isn't standard solution.

garthwebb: Limit and offset for use with paginating results are not good for performance, check out http://use-the-index-luke.com/no-offset

extensions/wikia/Wall/WallThread.class.php
```diff
@@ -77,31 +81,31 @@ private function loadReplyObjs() {
...
 		// this is a direct way to get IDs
 		// the other one is in Wall.class done in a grouped way
 		// (fetch for many threads at once, set with ->setReplies)
 
-		$title = Title::newFromId( $this->mThreadId );
+		$conditions = array( 'parent_comment_id = '.$this->mThreadId );
```
garthwebb: We prefer the ```[``` and ```]``` syntax for arrays.

extensions/wikia/Wall/WallThread.class.php
garthwebb: Style-wise, I think this constant is fine; upper case and underscore separated.  I think it could probably be more descriptive though.  Since its a limit on the number of replies, maybe ```FETCHED_REPLY_LIMIT```?

extensions/wikia/Wall/WallThread.class.php
mklucsarits: With a limit of 500 rows, this will very rarely ever enter recursion, and when it does, it will be less than 10 recursive calls. I don't think this particular case is anything to worry about.

extensions/wikia/Wall/WallThread.class.php
garthwebb: Yah, ```$title``` was constructed via ```newFromId( ID )``` where ```ID``` is an article ID.  Since its constructed with```$threadId``` we're doing a useless query to return us what we already had to begin with.

The only reason it was even instantiated before was to call ```getDBkey``` and ```getNamespace``` to use as values in the query.  The diff that @mklucsarits pasted above correctly optimizes this out since the ID is the same as a DB key + namespace value.

extensions/wikia/Wall/WallThread.class.php
```diff
@@ -109,13 +113,25 @@ private function loadReplyIdsFromDB($master = false) {
...
+		$dbr = wfGetDB( $master ? DB_MASTER : DB_SLAVE );
+
+		$this->setReplies( $this->getReplyIdsFromDB( $dbr ) );
 	}
 
 	public function invalidateCache() {
 		// invalidate cache at Thread level (new reply or reply removed in thread)
-		$this->mForceMaster = true;
-		$this->loadReplyIdsFromDB( true );
+		$this->getCache()->delete( $this->getThreadKey() );
```
garthwebb: Calling ```loadReplyIdsFromDB``` would eventually call ```setReplies``` which would do some initialization on ```data```:
```php
$this->data = new StdClass();
$this->data->threadReplyIds = false;
$this->data->threadReplyObjs = false;
```
We should probably do that here just to make sure the local copy of this now out of date data is removed.

garthwebb:
LGTM!

extensions/wikia/Wall/WallThread.class.php
kvas-damian: According to @mklucsarits note it'll be very rarely the case that we use parameter: https://github.com/Wikia/app/pull/7012/files#r29190827
We always have to find right balance between performance optimisations vs code readability. In this case both ways are fine with me.

extensions/wikia/Wall/WallThread.class.php
kvas-damian: :+1:

PR #7014, opened 2015-04-24
WikiService::getTopEditors - improve DB profiling

macbre:
Log `WikiService::getTopEditors` instead of `WikiService::{closure}` as a method that sent the query.
- use `$wgSpecialsDB` instead of a hardcoded string

@owend / @michalroszka


wladekb:
lgtm

PR #7017, opened 2015-04-24
Portable infoboxes parse fix

RafLeszczynski:
fixes issues with unneeded escaping of html entities while parsing portable infobox data


RafLeszczynski:
LGTM

PR #7019, opened 2015-04-24
SOC-683 Use unset() directly on global var

mklucsarits:
Currently it's not possible to unset an array value using $this->wg->Global from a controller, because it doesn't have a reference to the actual $wgGlobal var. I didn't see any applicable function in WikiaGlobalRegistry to handle a case like this, so I changed the code to directly reference the global variable.

https://wikia-inc.atlassian.net/browse/SOC-683


mklucsarits:
@jsutterfield @garthwebb

garthwebb:
D'oh, good catch @mklucsarits!  LGTM.

jsutterfield:
LGTM

PR #7022, opened 2015-04-24
ADEN-1977 + ADEN-1980 Paid Asset Drop updates

gabrys:


extensions/wikia/PaidAssetDrop/js/spec/paidAssetDrop.spec.js
```diff
@@ -90,8 +91,29 @@ describe('ext.wikia.paidAssetDrop.paidAssetDrop', function () {
...
-				return '1';
+			getVal: function (param) {
+				return param === 'forcepad' ? '1' : undefined;
+			}
+		});
+		expect(getModule().isNowValid(config)).toEqual(true);
+	});
+
+	it('Disaster recovery', function () {
+		var config = false;
```
nandy-andy: What do you think about passing here a valid configuration? My idea is that if somebody instant globals condition this test will fail with valid config.

nandy-andy:
LGTM :+1:

PR #7026, opened 2015-04-27
Put CuratedContent into Mercury API getArticle

bognix:
ping @vforge @rogatty

### DIFF
--- a/extensions/wikia/MercuryApi/MercuryApiController.class.php
+++ b/extensions/wikia/MercuryApi/MercuryApiController.class.php
@@ -1,5 +1,7 @@
 <?php
 
+use Wikia\Logger\WikiaLogger;
+
 class MercuryApiController extends WikiaController {
 
 	const PARAM_ARTICLE_ID = 'id';
@@ -10,10 +12,6 @@ class MercuryApiController extends WikiaController {
 	const DEFAULT_PAGE = 1;
 
 	const WIKI_VARIABLES_CACHE_TTL = 60;
-
-	const MEMCACHE_TTL = 86400;
-	const MEMCACHE_PREFIX = 'CuratedContent_MercuryApi';
-	const MEMCACHE_VERSION = 1;
 
 	private $mercuryApi = null;
 
@@ -153,19 +151,6 @@ class MercuryApiController extends WikiaController {
 		}
 	}
 
-	private function getCuratedContentSections() {
-		try {
-			$response = $this->sendRequest( 'CuratedContent', 'getList' )->getData();
-			if ( !empty( $response['sections'] ) ) {
-				return $response['sections'];
-			} else {
-				return false;
-			}
-		} catch ( NotFoundApiException $ex ) {
-			return false;
-		}
-	}
-
 	/**
 	 * @return Title Article Title
 	 * @throws NotFoundApiException
@@ -286,7 +271,6 @@ class MercuryApiController extends WikiaController {
 	 * @throws BadRequestApiException
 	 */
 	public function getArticle() {
-		global $wgMemc;
 		try {
 			$title = $this->getTitleFromRequest();
 			$articleId = $title->getArticleId();
@@ -309,15 +293,12 @@ class MercuryApiController extends WikiaController {
 			}
 
 			if ( $title->isMainPage() ) {
-				$curatedContent = $wgMemc->get( $memcKey );
-				if ( empty( $curatedContent ) ) {
-					$curatedContentSections = $this->getCuratedContentSections();
-					if ( !empty( $curatedContentSections ) ) {
-						$curatedContent = $this->getItemsForSections( $curatedContentSections );
-						$wgMemc->set( $memcKey, $curatedContent, self::MEMCACHE_TTL );
-					}
+				try {
+					$curatedContentData = $this->sendRequest( 'CuratedContent', 'getList' )->getData();
+					$data[ 'curatedContent' ] = $this->mercuryApi->getCuratedContent($curatedContentData);
+				} catch (NotFoundApiException $ex) {
+					WikiaLogger::instance()->error('Curated content and categories are empty');
 				}
-				$data[ 'curatedContent' ] = $curatedContent;
 			}
 		} catch ( WikiaHttpException $exception ) {
 			$this->response->setCode( $exception->getCode() );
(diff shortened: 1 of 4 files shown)


extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -149,6 +149,19 @@ private function getRelatedPages( $articleId, $limit = 6 ){
 		}
 	}
 
+	private function getCuratedContentSections() {
+		try {
+			$response = $this->sendRequest( 'CuratedContent', 'getList' )->getData();
+			if ( !empty( $response['sections'] ) ) {
+				return $response['sections'];
```
rogatty: What about `featured` field? According to spec we should display its items on the top of the main page.

Featured section isn't listed in `sections` so we would lose this information. For example http://gameofthrones.wikia.com/wikia.php?controller=CuratedContent&method=getList

bognix: We need to make sure what is the first scope of our iteration. I assumed we just want to display a list of sections and items at the beginning. I'll reach out to our PO to be 100% sure.

extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -280,7 +281,7 @@ public function getArticle(){
 				'topContributors' => $this->getTopContributorsDetails(
 					$this->getTopContributorsPerArticle( $articleId )
 				),
-				'article' => $articleAsJson
+				'article' => $articleAsJson,
```
rogatty: Nope :no_good:

extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -288,7 +289,12 @@ public function getArticle(){
 			if ( !empty( $relatedPages ) ) {
 				$data[ 'relatedPages' ] = $relatedPages;
 			}
-		} catch (WikiaHttpException $exception) {
+
+			if ( $title->isMainPage() ) {
```
rogatty: Should we check here if current wiki is a part of our experiment? It's not very nice to slow down all the main pages if not needed.

extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -310,6 +316,22 @@ public function getArticle(){
 		$this->response->setVal( 'data', $data );
 	}
 
+	public function getItemsForCuratedContentSection() {
+		$sectionName = $this->getVal( 'sectionName' );
+		if ( empty( $sectionName ) ) {
+			$this->response->setVal( 'items', false );
+		} else {
+			$sectionItems = $this->sendRequest( 'CuratedContent', 'getList', ['section' => $this->getVal( 'sectionName' )] )->getData();
```
rogatty: Use `$sectionName`

extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -310,6 +316,22 @@ public function getArticle(){
...
+			$items = [];
+			foreach ( $sectionItems['items'] as $item ) {
+				$item['article_local_url'] = Title::newFromID( $item['article_id'] )->getLocalURL();
```
rogatty: Add a comment explaining why are we doing this.

extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -322,4 +344,15 @@ public function getSearchSuggestions() {
...
 		);
 	}
+
+	private function getMainPageData() {
+		$mainPageData = [];
+		try {
+			$curatedContentData = $this->sendRequest( 'CuratedContent', 'getList' )->getData();
+			$mainPageData['curatedContent'] = $this->mercuryApi->getCuratedContent( $curatedContentData );
+		} catch ( NotFoundApiException $ex ) {
+			WikiaLogger::instance()->error( 'Curated content and categories are empty' );
```
rogatty: Is it really an error? Just an empty wiki, no need for drama IMO.

extensions/wikia/MercuryApi/models/MercuryApi.class.php
```diff
@@ -240,28 +240,27 @@ private function clearUsers() {
 	}
 
 	/**
-	 * Get categories titles
-	 *
-	 * @param array $articleCategories
-	 * @return array
-	 */
-	private function getArticleCategoriesTitles( Array $articleCategories ) {
```
rogatty: Let me just say that this method is not used anywhere.

extensions/wikia/MercuryApi/models/MercuryApi.class.php
```diff
@@ -240,28 +240,27 @@ private function clearUsers() {
...
 	}
+
+	public function getCuratedContent($data) {
+		$sections = $this->getCuratedContentSections($data);
+		return [
+			'sections' => $sections
+		];
+	}
+
+	protected function getCuratedContentSections($data) {
```
rogatty: Both methods are now obsolete as we want to return featured items too. Right?

bognix: There is separate story to implement "featured items"

extensions/wikia/MercuryApi/MercuryApiController.class.php
bognix: We would pick the communities via optimizely which works only on client side. We can add a wg variable but we would need to change it's value in WF manually. So we would need to sync it in two places. We can measure if it's slowing down mercury a lot but my guess would be it won't because we have caching enabled.

extensions/wikia/MercuryApi/MercuryApiController.class.php
bognix: true. If curated content is empty then Mercury won't display it. No need for drama.

rogatty: How about logging it but not as an error?

extensions/wikia/MercuryApi/models/MercuryApi.class.php
rogatty: Discussed IRL, let's remove what's not needed.

bognix:
Closing this PR. Another branch created instead - curated-content.

PR #7028, opened 2015-04-27
SERVICES-292 CuratedContent should return allCategories if wgCuratedContent is empty

sqreek:
fix bug, that no categories was return when wgCuratedContent was empty
change categories result format for same as CuratedContent format ({title, image_id, image_url})

//cc @ukaka


pchojnacki:
LGTM

PR #7039, opened 2015-04-27
Remove memcache from social icons

bognix:
We can't cache it for now because cached value vary for each URL


vforge:
LGTM :heavy_check_mark:

PR #7043, opened 2015-04-27
(CE-1534) Add undeleted images back to the image review queue

Grunny:
When a deleted file is undeleted, it should be readded to the image review
queue. This adds the image data with the undeletion and fixes a bug in the
undeletion event due to incorrectly checking for an instance of Article.

Tested on sandbox.


lukaszkonieczny:
LGTM

PR #7057, opened 2015-04-28
CONCF-470 Video uploading fix

dianafa:
**TLDR;** The failed video upload issue was caused by wgIsGhostVideo which was set in [VideoInfoHooksHelper.class.php](https://gist.github.com/dianafa/898d5e2186d247d78162#file-videoinfohookshelper-class-php-L329) and was causing the error while uploading video.

At first try, before the video was not yet in the DB it set wgIsGhostVideo as true and thus after, when actually everything went correctly, onCheckGhostFile file was set to null and there was a failure. After the page reload everything appeared correctly (it was correctly in fact, only the error caused by wgIsGhostVideo check was misleading).

Function wfFindFile( $videoTitle ); was actually a wrapper for RepoGroup::singleton()->findFile( $videoTitle ); which contained unuseful in case of video uploading wgIsGhostVideo check and caused error.

ping: @bognix @vforge

### DIFF
--- a/extensions/wikia/SpecialEditHub/js/EditHub.js
+++ b/extensions/wikia/SpecialEditHub/js/EditHub.js
@@ -20,10 +20,16 @@
 			$('.module-popular-videos').on('click', '.remove', $.proxy(this.popularVideosRemove, this));
 			$('#edit-hub-removeall').click($.proxy(this.popularVideosRemoveAll, this));
 			editHubMain.find('.vet-show').each(function () {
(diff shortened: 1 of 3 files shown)


vforge:
LGTM :heavy_check_mark:

bognix:
ghost video is a video from www.video.wikia.com which doesn't have a record on local wiki. Please make sure while testing that uploading this videos still works and they are playable.

bognix:
LGTM

PR #7058, opened 2015-04-28
Special:Insights beta version release

adamkarminski:
## Special:Insights beta release

This is a new extension introduced by the Community Engineering team. As we want to help Power Users to be productive, this feature will allow them to work on the pages that need their attention the most.
### Release plan

Parts of the code have been reviewed in the development process. If the extension passes all of the QA tests we're shipping it with the release on 30-04-2015 for a limited amount of communities as a beta version.


extensions/wikia/InsightsBlogpostRedirect/InsightsBlogpostRedirectController.class.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+class InsightsBlogpostRedirectController extends WikiaSpecialPageController {
+
+	public function __construct() {
+		parent::__construct( 'Insights', 'insights', true );
+	}
+
+	public function index() {
+		$this->wg->Out->redirect( wfMessage('insights-blogpost-url')->escaped() );
```
Grunny: I'd use `->useDatabase( false )` here. We shouldn't let admins modify a URL we redirect a user to.

Also, `wfMessage` needs to be spacey per conventions.

extensions/wikia/Insights/InsightsHooks.class.php
```diff
@@ -0,0 +1,87 @@
...
+			$extraQuery .= 'insights=' . $subpage;
+		}
+
+		return true;
+	}
+
+	/**
+	 * Add insights param to edit page form to keep information about insights flow
+	 */
+	public static function onGetLocalURL( &$this, &$url, $query ) {
```
Grunny: The first parameter should be `Title $title` not `&$this`.

PR #7059, opened 2015-04-28
SERVICES-292 CuratedContent should return allcategories in same format as from wgCuratedContent

sqreek:
When there is no data in wgCuratedContent, CuratedContent should return allcategories form mwapi with same structure as in CuratedContent


pchojnacki:
LGTM

PR #7062, opened 2015-04-28
HG-600: Add information about section indexes to WikiaMobile and ArticleAsJson output

inez:
Given the fact that WikiaMobile and ArticleAsJson are to be retired - and latter to be replaced by Article Service - this seems like a right solution

This change is also unifying output for ArticleAsJson so it does not depend on user state (logged in or out).


owend:
Yep, this LGTM

kenkouot:
This looks good to me too.

inez:
@hakubo @kenkouot already merged it, but as the original plan was to have you review it as well - please take a look when you get a chance :) Thanks!

hakubo:
LGTM

PR #7066, opened 2015-04-29
PLATFORM-1151 refresh the db state after Helios call

ghost:
In the scope of a DB connection, once the first read occurs, MediaWiki
will operate on a frozen state of the DB and will not see changes made
by external services. I added an explicit commit that will force
refreshing the state after Helios call.


michalroszka:
@wladekb @macbre et al. — I am sorry :-)

macbre:
Commit this commit to production

PR #7068, opened 2015-04-29
PLATFORM-1192 batch refresh links for template throttling

macbre:
- decrease the value of `$wgUpdateRowsPerJob` to 50 - throttle tasks that parse the articles latest revision after a template edit
- improve logging around tasks queue pushing and execution

@michalroszka / @nmonterroso

### DIFF
--- a/includes/wikia/tasks/Tasks/BatchRefreshLinksForTemplate.class.php
+++ b/includes/wikia/tasks/Tasks/BatchRefreshLinksForTemplate.class.php
@@ -88,7 +88,12 @@ class BatchRefreshLinksForTemplate extends BaseTask {
 
 	protected function batchEnqueue( array $tasks ) {
 		if ( !empty( $tasks ) ) {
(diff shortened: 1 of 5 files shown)


includes/wikia/tasks/Tasks/BatchRefreshLinksForTemplate.class.php
```diff
@@ -83,12 +83,20 @@ public function readyRefreshLinksForTitleTask( \Title $title ) {
 		$task->title( $title );
 		$task->call( 'refresh' );
 		$task->wikiId( $this->getWikiId() );
+
+		// TODO: delay the tasks - see PLATFORM-1192
+
```
macbre: Let's try the current approach and introduce the delay for these tasks when the current fix is not enough.

macbre:
Slack LGTM from @michalroszka

macbre:
Backported to the release in c0ca634

PR #7070, opened 2015-04-29
SOC-701 Results count needs to be an integer

mklucsarits:
This fixes https://wikia-inc.atlassian.net/browse/SOC-701

The BlogTemplate::getResultsCount() function was recently changed, and as a result returned a string instead of an integer. An integer is needed to be passed to Paginator::paginate() to calculate the number of pages. When a string was being passed, Paginator simply didn't attempt to create any pagination.

@armonr @jsutterfield Please review.


armonr:
LGTM
Test has restarted. It should definitely pass for this change :)

jsutterfield:
LGTM!

PR #7071, opened 2015-04-29
ADEN-1779 Amazon on Mercury

nandy-andy:
In order to have Amazon enabled on Mercury we need to change our `ext.wikia.adEngine.lookup.amazonMatch` module and add assets in AssetManager config.


extensions/wikia/AdEngine/js/lookup/amazonMatch.js
```diff
@@ -139,26 +148,49 @@ define('ext.wikia.adEngine.lookup.amazonMatch', [
 		return amazonCalled;
 	}
 
+	function isValidSlot(slotName, validSlotNames) {
```
gabrys: Replace with a simple indexOf

extensions/wikia/AdEngine/js/lookup/amazonMatch.js
```diff
@@ -139,26 +148,49 @@ define('ext.wikia.adEngine.lookup.amazonMatch', [
...
+
+			if (!isValid) {
+				isValid = slotName === validSlotName;
+			}
+		});
+
+		return isValid;
+	}
+
+	function checkIfObjectKeysNotSupported(functionName, retval) {
```
gabrys: Since this module is useless without Object.keys anyways, let's return the full interface, but without any implementation if there's no Object.keys and then assume Object.keys is supported throughout the module.

extensions/wikia/AdEngine/js/spec/AmazonMatch.spec.js
```diff
@@ -49,6 +49,7 @@ describe('Method ext.wikia.adEngine.lookup.amazonMatch', function () {
 		// Single values
 		{input: {'a1x6p14': 1}, expected: {skyscraper: ['a1x6p14']}},
 		{input: {'a3x2p14': 1}, expected: {medrec: ['a3x2p14']}},
+		{input: {'a3x5p14': 1}, expected: {mobile: ['a3x5p14']}},
```
gabrys: add a test with both mobile leaderboard and mobile incontent

nandy-andy: done :+1:

extensions/wikia/AssetsManager/config.php
```diff
@@ -709,13 +709,15 @@
 		'//extensions/wikia/AdEngine/js/AdEngine2.js',
 		'//extensions/wikia/AdEngine/js/AdLogicPageParams.js',
 		'//extensions/wikia/AdEngine/js/AdTracker.js',
+		'//extensions/wikia/AdEngine/js/config/mobile.js',
```
gabrys: sorting alphabetically with uppercase before the lowercase.

PR #7072, opened 2015-04-30
CE-1683 Page views display implementation

adamkarminski:
This code implements displaying number of views of pages in the Special:Insights feature. It is also a backend infrastructure for sorting and pagination.

@Wikia/community-engineering


extensions/wikia/Insights/models/InsightsQuerypageModel.php
```diff
@@ -37,17 +52,132 @@ public function getData() {
...
+			$this->offset = intval( $params['offset'] );
+		}
+
+		if ( isset( $params['limit'] ) ) {
+			$this->limit = intval( $params['limit'] );
+		}
+	}
+
+	public function fetchArticlesData() {
+		$cacheKey = wfSharedMemcKey( $this->getInsightType(), 'articlesData' );
```
lukaszkonieczny: Please add wiki id to memcache key. What do you think about having method which returns the key?

extensions/wikia/Insights/models/InsightsQuerypageModel.php
```diff
@@ -37,17 +52,132 @@ public function getData() {
...
+		$this->cacheTtl = WikiaResponse::CACHE_STANDARD * 3;
```
lukaszkonieczny: I think using raw number will be more descriptive

includes/wikia/services/DataMartService.class.php
```diff
@@ -838,6 +838,32 @@ public static function getTopTagsWikisByPageviews ($tagId, $startDate, $endDate,
...
 	}
 
+	/**
+	 * Gets page views for given articles
+	 *
+	 * @param array $articlesIds
+	 * @param datetime $timeId
+	 * @return array
+	 */
+	public static function getPageViewsForArticles( Array $articlesIds, $timeId ) {
```
lukaszkonieczny: While this method is not strictly conected with Insights, what do you think about passing more arguments like period and wiki id?

extensions/wikia/Insights/models/InsightsQuerypageModel.php
```diff
@@ -37,17 +52,132 @@ public function getData() {
...
+
+	public function assignPageViewsData( $articlesData, $pageViewsData ) {
+		$sortingData = [];
+
+		foreach ( $articlesData as $articleId => $data ) {
+			$pv = [
+				intval( $pageViewsData[0][ $articleId ] ),
+				intval( $pageViewsData[1][ $articleId ] ),
+				intval( $pageViewsData[2][ $articleId ] ),
+				intval( $pageViewsData[3][ $articleId ] ),
```
lukaszkonieczny: These values could not be set if there is no page view for given article,

extensions/wikia/Insights/models/InsightsQuerypageModel.php
```diff
@@ -37,17 +52,132 @@ public function getData() {
...
+			];
+
+			$pv28 = array_sum( $pv );
+			if ( $pv[1] != 0 ) {
+				$pvDiff = ( $pv[1] - $pv[2] ) / $pv[2];
```
lukaszkonieczny: Posible division by zero

extensions/wikia/Insights/models/InsightsQuerypageModel.php
```diff
@@ -37,17 +52,132 @@ public function getData() {
...
+		if ( isset( $params['sort'] ) && isset( $this->sorting[ $params['sort'] ] ) ) {
+			$this->sortingArray = $wgMemc->get( wfSharedMemcKey( $this->getInsightType(), $params['sort'] ) );
+		}
+
+		if ( isset( $params['offset'] ) ) {
+			$this->offset = intval( $params['offset'] );
+		}
+
+		if ( isset( $params['limit'] ) ) {
+			$this->limit = intval( $params['limit'] );
```
lukaszkonieczny: What do you think about some max limit value?

extensions/wikia/Insights/models/InsightsQuerypageModel.php
lukaszkonieczny: I think we also need memcache version variable (const)

Grunny: For reference, just switching to using `wfMemcKey` rather than `wfSharedMemcKey` will make it include the wiki ID automatically.

extensions/wikia/Insights/models/InsightsQuerypageModel.php
```diff
@@ -37,17 +52,132 @@ public function getData() {
...
+		$articlesData = WikiaDataAccess::cache( $cacheKey, $this->cacheTtl, function () {
+			$res = $this->queryPageInstance->doQuery();
+
+			if ( $res->numRows() > 0 ) {
+				$articlesData = $this->prepareData( $res );
+			}
+
+			if ( $this->arePageViewsRequired() ) {
+				$articlesIds = array_keys( $articlesData );
+				$pageViewsData = $this->getPageViewsData( $articlesIds );
```
lukaszkonieczny: We should support situation when list is empty (there is no article ids)

extensions/wikia/Insights/models/InsightsQuerypageModel.php
adamkarminski: ``intval()``` evaluates ```false``` and ```null``` to 0.

extensions/wikia/Insights/models/InsightsQuerypageModel.php
adamkarminski: Right, I forgot to change the condition to `$pv[2]`. Thanks!

extensions/wikia/Insights/models/InsightsQuerypageModel.php
adamkarminski: :+1:

extensions/wikia/Insights/models/InsightsQuerypageModel.php
adamkarminski: :+1:

includes/wikia/services/DataMartService.class.php
adamkarminski: Makes sense! :+1:

extensions/wikia/Insights/models/InsightsQuerypageModel.php
adamkarminski: :+1:

extensions/wikia/Insights/models/InsightsQuerypageModel.php
adamkarminski: I'll use ```wfMemcKey``` and a cache version. Thanks!

extensions/wikia/Insights/models/InsightsQuerypageModel.php
lukaszkonieczny: That's true, but there are warnings or notices added to logs

adamkarminski: There shouldn't be. If a page has 0 views, there is just no entry in the db. In this case evaluating false to 0 with `intval()` is not an unexpected behavior that should be logged.

lukaszkonieczny: If there is no entry in db, ```$pageViewsData[X][ $articleId ]``` is not set so PHP notice ```undefined offset``` will be logged.

adamkarminski: Ok, I'll check if it's set before.

adamkarminski:
unitTests fail unrelated... (as usually :rage:)

lukaszkonieczny:
lgtm

PR #7075, opened 2015-04-30
Revert "SOC-674 log which user properties are being used"

macbre:
Reverts Wikia/app#7002

@jsutterfield / @rafalkalinski


macbre:
Backported to the release in 59e4aea

PR #7085, opened 2015-04-30
(CE-1714) Fix removal of fixed items from Insights

Grunny:
Because of the way our DB transactions are handled, the removal of
fixed items from the Insights list was failing due to the changes
not being committed. This fixes it by both explicitly committing the
changes and switching the request to a POST (which would auto-commit
at the end of the request anyway) as it performs a write action.

Also fixes a small bug where the deletion did not include the namespace
of the article.


owend:
LGTM

PR #7086, opened 2015-05-01
SOC-747 Displaying the correct comment now.  Also hopefully fixed race condition

garthwebb:
In addition to showing and linking to the wrong comment (always the latest rather than the one left) sometimes the comment text would not appear.  This is likely a race condition with lagged master-slave replication.

Both should are addressed in this PR.

https://wikia-inc.atlassian.net/browse/SOC-747

### DIFF
--- a/extensions/wikia/Email/Controller/CommentController.class.php
+++ b/extensions/wikia/Email/Controller/CommentController.class.php
@@ -14,11 +14,24 @@ abstract class CommentController extends EmailController {
 	/** @var \Title */
 	protected $latestComment;
 
+	/** @var \Title */
+	protected $commentTitle;
+
 	public function initEmail() {
 		$titleText = $this->request->getVal( 'title' );
 		$titleNamespace = $this->request->getVal( 'namespace' );
 
 		$this->title = \Title::newFromText( $titleText, $titleNamespace );
+
+		// The summary here is the contents of the comment
+		$commentRevID = $this->getVal( 'currentRevId', false );
+		if ( $commentRevID ) {
+			$rev = \Revision::newFromId( $commentRevID, \Revision::USE_MASTER_DB );
+
+			if ( $rev ) {
+				$this->commentTitle = $rev->getTitle( \Revision::USE_MASTER_DB );
+			}
+		}
 
 		$this->assertValidParams();
 	}
@@ -28,6 +41,7 @@ abstract class CommentController extends EmailController {
 	 */
 	private function assertValidParams() {
 		$this->assertValidTitle();
+		$this->assertValidCommentTitle();
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


extensions/wikia/Email/Controller/CommentController.class.php
```diff
@@ -14,12 +14,25 @@
 	/** @var \Title */
 	protected $latestComment;
 
+	/** @var \Title */
+	protected $commentTitle;
+
 	public function initEmail() {
 		$titleText = $this->request->getVal( 'title' );
```
garthwebb: Note that ```title``` here is the article being commented on, whereas ```currentRevId``` below is the latest revision ID for the comment that was left.

includes/db/Database.php
```diff
@@ -260,7 +260,7 @@ public function getServerInfo() {
 	 *   - false to disable debugging
```
garthwebb: All the changes in this file are to update the doc comments, with the exception of one unused global variable I removed below.

includes/db/Database.php
```diff
@@ -889,7 +894,7 @@ public function query( $sql, $fname = '', $tempIgnore = false ) {
 		}
 
 		# <Wikia>
-		global $wgProfiler, $wgDBReadOnly;
+		global $wgDBReadOnly;
```
garthwebb: $wgProfiler didn't seem to be used in this function

includes/db/Database.php
```diff
@@ -1893,8 +1903,9 @@ function update( $table, $values, $conds, $fname = 'DatabaseBase::update', $opti
 
 	/**
 	 * Makes an encoded list of strings from an array
-	 * @param $a Array containing the data
-	 * @param $mode int Constant
+	 *
+	 * @param Array $a containing the data
+	 * @param int $mode int Constant
```
mklucsarits: There's an extra "int" here.

includes/db/Database.php
```diff
@@ -1893,8 +1903,9 @@ function update( $table, $values, $conds, $fname = 'DatabaseBase::update', $opti
 
 	/**
 	 * Makes an encoded list of strings from an array
-	 * @param $a Array containing the data
```
mklucsarits: What a great variable name and description! :panda_face:

mklucsarits:
Ship it!

includes/Revision.php
```diff
@@ -32,15 +32,21 @@ class Revision {
 	const FOR_THIS_USER = 2;
 	const RAW = 3;
 
+	// These constants are used with methods that accept a true/false 'useMaster' parameter
+	const USE_SLAVE_DB = false;
```
armonr: This looks unused and duplicate

extensions/wikia/Email/Controller/CommentController.class.php
```diff
@@ -14,12 +14,25 @@
 	/** @var \Title */
 	protected $latestComment;
```
armonr: As discussed, we don't need this now

includes/wikia/services/ArticleService.class.php
```diff
@@ -155,7 +155,7 @@ function() use ( $service, $fname ) {
 					wfProfileIn( $fname . '::CacheMiss' );
 
 					$content = '';
-					if ( !empty( $this->wg->SolrMaster ) ) {
+					if ( !$this->wg->DevelEnvironment && !empty( $this->wg->SolrMaster ) ) {
```
garthwebb: Added this extra logic so that we don't query Solr in the dev environment and instead fall back to the code on line 162 below.  We have no dev Solr instance so this code will query production.  Its very likely that production will already have an article indexed with the ID just created here in dev and will return incorrect content.

armonr:
LGTM

PR #7089, opened 2015-05-04
(CE-1679) Remove Phalanx block from the rename process

Grunny:
When the local rename part of the rename process fails at any point
the user is regularly left blocked until we can clean up the rename.
This is unnecessary as there is no reason to stop the user making use
of the new name after the global rename has happened but before all
the edits and so forth have been reattributed.

This removes the Phalanx block as when it is put in place, the main
rename that updates the user table and creates a fake user at the old
name has already occurred.

/cc @Wikia/community-engineering @michalroszka


lgarczewski:
LGTM.

PR #7091, opened 2015-05-04
LoadBalancer: log 1% of connections

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-1194

Trying to debug connections spikes on `db-archive` machines.

@michalroszka / @drozdo


macbre:
Backported to the release in dfa8567

PR #7095, opened 2015-05-04
footer

jacekjursza:


extensions/wikia/PortableInfobox/tests/PortableInfoboxRenderServiceTest.php
```diff
@@ -28,7 +28,13 @@ public function testRenderInfobox( $input, $expectedOutput, $description ) {
 		$actualDOM->loadXML($actualOutput);
 		$expectedDOM->loadXML($expectedOutput);
 
-		$this->assertEquals( $expectedDOM->saveXML(), $actualDOM->saveXML(), $description );
+		// lets sanitize expected and actual HTML to not deal with whitespaces
```
SebastianMarzjan: This looks scary. A link to source explaining why this is needed (in the light of DOMDocument attributes set above) should be added here.

SebastianMarzjan:
Some more context is needed in the replace block of the test.
Other than that, LGTM

extensions/wikia/PortableInfobox/tests/PortableInfoboxRenderServiceTest.php
jacekjursza: without sanitizing, I had something like this:
--- Expected
+++ Actual
@@ @@
 '<?xml version="1.0"?>
 <aside class="portable-infobox">
-  <footer class="portable-infobox-footer portable-infobox-item-margins portable-infobox-header-background portable-infobox-header-font">Footer value</footer>
+  <footer class="portable-infobox-footer portable-infobox-item-margins portable-infobox-header-background portable-infobox-header-font">
+	Footer value
+</footer>
 </aside>
 '

SebastianMarzjan:
Go for it :+1:

PR #7097, opened 2015-05-04
Revision: cache blobs in memcache

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-1194

Set `$wgRevisionCacheExpiry` to an hour

Try to make the load on `db-archive` a bit smaller.

@drozdo / @michalroszka / @wladekb / @harnash

### DIFF
--- a/includes/DefaultSettings.php
+++ b/includes/DefaultSettings.php
@@ -1422,7 +1422,7 @@ $wgDefaultExternalStore = false;
  *
  * Set to 0 to disable, or number of seconds before cache expiry.
  */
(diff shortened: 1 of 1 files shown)


wladekb:
Theoretically blobs are immutable so we could use much longer TTL if and only if the cache is not used when the blob is not found (eg. it didn't get to the slave yet during read).

macbre:
True, but I'd like to verify if caching here makes any sense at all first. Then (after taking a look at memcache hit ratio stats) we can tweak the TTL.

macbre:
Backported to production in 8f5a753

PR #7103, opened 2015-05-05
DAT-2760 Backwards compatibility in <image> tag for portable infoboxes

sebastianmarzjan:


jacekjursza:
LGTM

SebastianMarzjan:
Merging to feature branch

extensions/wikia/PortableInfobox/services/Helpers/ImageFilenameSanitizer.php
```diff
@@ -0,0 +1,60 @@
+<?php
+
+namespace Wikia\PortableInfobox\Helpers;
+
+class ImageFilenameSanitizer {
```
wladekb: It looks like a general-interest functionality. Maybe we have an existing implementation somewhere else or we should put it in some common place?

SebastianMarzjan: Will check on next iteration on this, thanks for the input!

PR #7114, opened 2015-05-05
Curated Main Page for Mercury - featured section

rogatty:
https://wikia-inc.atlassian.net/browse/CONCF-510

/cc @bognix @Warkot @vforge 

It will need some adjustments on the Mercury side. While cleaning, I've renamed some things to make them shorter and consistent with the rest of code:
- `getItemsForCuratedContentSection` to `getCuratedContentSection`
- `sectionName` to `section`


extensions/wikia/MercuryApi/models/MercuryApi.class.php
```diff
@@ -249,11 +249,40 @@ public function getAdsContext( Title $title ) {
...
+	/**
+	 * @desc Mercury can't open article using ID - we need to create a local link
+	 * FIXME talk to platform team about performance cost of this operation (it queries DB)
+	 *
+	 * @param $item
+	 * @return mixed
+	 */
+	private function addArticleUrlToCuratedContentItem( $item ) {
+		if ( !empty( $item[ 'article_id' ] ) ) {
+			$item[ 'article_local_url' ] = Title::newFromID( $item[ 'article_id' ] )->getLocalURL();
```
vforge: Question: Can we mcache this in some way? (just an idea)

rogatty: Yup, we should. There is `FIXME` in the method doc.

PR #7116, opened 2015-05-06
error handling

jacekjursza:


extensions/wikia/PortableInfobox/controllers/PortableInfoboxParserTagController.class.php
```diff
@@ -40,6 +47,14 @@ public function renderInfobox( $text, $params, $parser, $frame ) {
 		return [ $renderedValue, 'markerType' => 'nowiki' ];
 	}
 
+	private function renderUnimplementedTagErrorMesssage( $tagName ) {
+		$renderedValue = '<strong class="error"> '
+			. wfMessage( 'unimplemented-infobox-tag' ) . ': '
```
SebastianMarzjan: You should pass $tagName as a message param as depending on the language the order of words may be different

SebastianMarzjan: I'm also wondering whether this should be rendered server-side or client-side, as running this will make the message display in the editors language regardless of viewers language (parser cache)

extensions/wikia/PortableInfobox/PortableInfobox.i18n.php
```diff
@@ -3,9 +3,11 @@
...
 	'portable-infobox-desc' => 'Create portable infoboxes which can be rendered using clean semantic HTML markup on
-	any skin / platform using using easy to understand powerful XML-like markup'
+	any skin / platform using using easy to understand powerful XML-like markup',
+	'unimplemented-infobox-tag' => 'Unimplemented infobox tag: <$1>'
 ];
 
 $messages[ 'qqq' ] = [
-	'portable-infobox-desc' => 'Portable Infobox extension description'
+	'portable-infobox-desc' => 'Portable Infobox extension description',
+	'unimplemented-infobox-tag' => 'Error message for using unimplemented infobox tag: <$1>'
```
SebastianMarzjan: What do you think about phrasing it like: "Error message for using unimplemented infobox tag; $1 is the tag name without pointy braces"?

SebastianMarzjan:
One short suggestion at your discretion; LGTM

PR #7123, opened 2015-05-06
Dat 2746 - add lightbox support for infobox images and move thumbnail creation to php

RafLeszczynski:


jacekjursza:
LGTM, but please check the tests

PR #7128, opened 2015-05-06
SOC-770 skip getBanInformation for anon's

garthwebb:
Also construct memckeys with the global wfSharedMemKey method and some additional cleanup.

https://wikia-inc.atlassian.net/browse/SOC-770

### DIFF
--- a/extensions/wikia/Chat2/Chat.class.php
+++ b/extensions/wikia/Chat2/Chat.class.php
@@ -14,12 +14,6 @@ class Chat {
 	// Cache ban info for 5m
 	const BAN_INFO_TTL = 300;
 
-	var $chatId;
-
-	public function __construct( $chatId ) {
-		$this->chatId = $chatId;
-	}
-
 	/**
 	 * The return value of this method gets passed to Javascript as the global wgChatKey.  It then becomes the 'key'
 	 * parameter sent with every chat request to the Node.js server.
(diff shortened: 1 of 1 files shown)


extensions/wikia/Chat2/Chat.class.php
```diff
@@ -290,6 +280,10 @@ protected static function getTimeLabel( $time ) {
 	 * @throws MWException
 	 */
 	public static function getBanInformation( $cityID, User $banUser ) {
+		if ( $banUser->isAnon() ) {
+			return false;
+		}
+
```
garthwebb: Since ban rows in the DB require a user ID (and there no anonymous chatting) there's no reason to go any further if we've got an anonymous user).

mklucsarits:
LGTM

extensions/wikia/Chat2/Chat.class.php
```diff
@@ -346,11 +340,12 @@ public static function clearBanInfoCache( $wikiID, $userID ) {
 	}
 
 	private static function getBanInfoCacheKey( $wikiID, $userID ) {
-		return wfMemcKey( 'chat-baninfo', $wikiID, $userID );
+		// Using shared mem key, but adding in the WikiID ourselves since its possible
+		// to call these functions with an alternate wiki ID.
+		return wfSharedMemcKey( 'chat-baninfo', $wikiID, $userID );
```
macbre: Thanks

extensions/wikia/Chat2/Chat.class.php
```diff
@@ -658,12 +649,10 @@ public static function addConnectionLogEntry() {
...
 		}
 
-		wfProfileOut( __METHOD__ );
 	}
 
-
-	static protected function getUserIPMemcKey( $userId, $address ) {
-		return $userId . '_' .  $address . '_v1';
+	protected static function getUserIPMemcKey( $userId, $address ) {
+		return wfSharedMemcKey( $userId, $address, 'v1' );
```
macbre: Please add a prefix to this key so it's easier to identify it: eg. `wfSharedMemcKey( 'Chat', 'userIP', $userId, $address, 'v1' )`

extensions/wikia/Chat2/Chat.class.php
```diff
@@ -674,33 +669,28 @@ static protected function getUserIPMemcKey( $userId, $address ) {
...
-		if ( Chat::getBanInformation( $wgCityId, $userObject ) !== false ) {
+		if ( Chat::getBanInformation( F::app()->wg->CityId, $userObject ) !== false ) {
 			return false;
 		}
 
-		return ( $userObject->isLoggedin() && $userObject->isAllowed( 'chat' ) );
+		return $userObject->isAllowed( 'chat' );
 	}
 
-	static public function getBanTimeFactors() {
```
garthwebb: Moved down below the function that calls this and made private.

extensions/wikia/Chat2/Chat.class.php
jsutterfield: I still don't understand, why are we using `wfSharedMemcKey` here? If we're passing in the `$wikiID` every time, isn't that essentially the same thing as `wfMemcKey`?

macbre: Well, that depends on whether `$cityId` equals `$wgCityId`.

jsutterfield:
LGTM

extensions/wikia/Chat2/Chat.class.php
```diff
@@ -11,14 +11,8 @@ class Chat {
...
-	// Cache ban info for 5m
-	const BAN_INFO_TTL = 300;
-
-	var $chatId;
-
-	public function __construct( $chatId ) {
-		$this->chatId = $chatId;
-	}
+	// Cache ban info for 24h, 24*60*60 = 86400
+	const BAN_INFO_TTL = 86400;
```
macbre: :+1:

PR #7129, opened 2015-05-06
CONCF-525 | Create 1024px breakpoint

vforge:
Ping @bognix @Warkot @hakubo

### DIFF
--- a/skins/oasis/css/core/breakpoints-layout.scss
+++ b/skins/oasis/css/core/breakpoints-layout.scss
@@ -90,9 +90,7 @@
 		.wikinav2.oasis-one-column .WikiaPageHeader {
 			padding: 0;
 		}
-	}
 
-	@media #{$breakpoint-small-only} {
 		.WikiHeader > nav.WikiNav .navbackground {
 			left: -24px;
 			width: 772px;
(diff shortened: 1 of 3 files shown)


skins/oasis/css/core/breakpoints-layout.scss
```diff
@@ -219,4 +217,44 @@
...
+
+	// "special case" for 1024px tablets (think landscape iPad)
+	@media #{$breakpoint-small-plus} {
+		.WikiaPage {
+			padding: 0 $padding-width-small-plus;
+			width: $page-width-small-plus;
+		}
+
+		.wikia-main-page-tag-lcs, .main-page-tag-lcs-exploded {
+			@include width-calc("100% - 320px");
```
bognix: from where is the 320px value? it's right rail width? If so please add comment

skins/oasis/css/core/breakpoints-layout.scss
```diff
@@ -219,4 +217,44 @@
...
+		}
+
+		.main-page-tag-rcs {
+			width: 300px;
+		}
+
+		.WikiHeader {
+			min-height: 69px;
```
bognix: is this value not set before somewhere? if not, why do we need to set it here?

vforge: It's being overriden by small breakpoint, so we want to change it.

vforge:
This branch is no longer needed - it's merged with CONCF-477 into CONCF-477-525.

Deleting this one.

PR #7130, opened 2015-05-06
CONCF-511 | fix social media sharing icons

Warkot:
:information_source: https://github.com/Wikia/config/pull/1090 is complementary to this PR.

ping for CR → @vforge @bognix @dianafa


extensions/wikia/PageShare/scripts/PageShare.js
```diff
@@ -19,48 +13,44 @@ require(['wikia.window', 'wikia.tracker', 'jquery'], function(win, tracker, $) {
 		var service = $(event.target).closest('a'),
 			url = service.prop('href'),
 			title = service.prop('title'),
-			h = (win.innerHeight / 2 | 0), // round down
-			w = (win.innerWidth / 2 | 0);  // round down
+			h = Math.floor(win.innerHeight / 2),
+			w = Math.floor(win.innerWidth / 2),
+			trackFunc = tracker.buildTrackingFunction({
```
bognix: you're building track function on each click.

extensions/wikia/PageShare/scripts/spec/PageShare.spec.js
```diff
@@ -0,0 +1,150 @@
...
+						},
+						useLang: 'pl'
+					},
+					expectedResult: 'pl'
+				}
+			];
+
+		testDataSet.forEach(function (testData) {
+			pageShareModule = modules['wikia.pageShare'](testData.mock.window, _, _);
+			expect(pageShareModule.getShareLang(testData.mock.useLang)).toBe(testData.expectedResult);
```
bognix: consider moving "it" inside foreach  loop. This way if one test case fails only one would fail.

extensions/wikia/PageShare/scripts/PageShare.js
Warkot: Swapped.

extensions/wikia/PageShare/scripts/spec/PageShare.spec.js
Warkot: Done.

bognix:
LGTM

vforge:
:heavy_check_mark: lgtm

Warkot:
Commit https://github.com/Wikia/app/commit/158517066ae54afe02678b9ee715fbf0cd0fb2b8 fixes bugs [CONCF-566](https://wikia-inc.atlassian.net/browse/CONCF-566) and [CONCF-604](https://wikia-inc.atlassian.net/browse/CONCF-604).

Commit https://github.com/Wikia/app/commit/05056497248eb335c548cfbd588fc71d57644051 fixes IE ([CONCF-567](https://wikia-inc.atlassian.net/browse/CONCF-567)) and brings some bigger logic changes than just a simple fix—it utilizes Accept-Language header that was sent with a request for the page. I developed this idea and implementation first, but pushed later as it's more complex and it would be easier to revert it if it's bad or not needed. However, I think that's a good base for reading and checking if we support next languages in the header if we don't support the first one. We can send the whole accepted languages list from the client to our backend and easily mitigate the caching issues while taking from client side the burden of looping over this resources.

Another issue is that some day we might want to differentiate regional languages like 'pt-PT' from 'pt-BR'… :water_buffalo:

vforge:
Despite what you've said, LGTM :heavy_check_mark: ;)

extensions/wikia/PageShare/PageShareHooks.class.php
```diff
@@ -0,0 +1,19 @@
...
+	 * Add global JS variable wgAcceptLangList
+	 *
+	 * @param array $vars global variables list
+	 * @return boolean return true
+	 */
+	public static function onMakeGlobalVariablesScript(Array &$vars) {
+		global $wgRequest;
+
+		$wgAcceptLang = $wgRequest->getAcceptLang();
+		$vars['wgAcceptLangList'] = is_array( $wgAcceptLang ) ? array_keys( $wgAcceptLang ) : null;
```
bognix: are you sure it's not gonna be cached?
From what I know we are cashing base on URL.
Imagine situation:
User with header ZH goes to http://muppet.wikia.com -> page gets cached
User with header PL goes to http://muppet.wikia.com -> wgAcceptLangList is cached and is set to ZH

Warkot: And based on skin AFAIK, but If we're not vary-ing cache based on Accept-Language, then you're right.

Warkot: I don't see any cache variationing for wikia pages based on the Accept-Language header both in wikia-vcl repo and after checking on sandbox. I'll revert that. Thanks for spotting @bognix, well done! :+1:

Warkot:
@bognix @vforge I reverted the big change/IE fix. Does it still LGTY?

extensions/wikia/PageShare/PageShareHelper.class.php
```diff
@@ -23,46 +22,41 @@ public static function getIcon( $service ) {
...
 	}
 
 	public static function isValidShareService( $service, $lang, $isTouchScreen ) {
-		// Don't display LINE social network (because it's a mobile app) on desktops
-		if ( !$isTouchScreen && $service['name'] === self::LINE_NAME ) {
+		// Don't display LINE social network or any other mobile app on desktops
+		if (
+			!$isTouchScreen
+			&& array_key_exists( 'displayOnlyOnTouchDevices', $service )
+			&& $service['displayOnlyOnTouchDevices']
```
bognix: simple !empty($service['displayOnlyOnTouchDevices'] wouldn't be sufficient instead of two checks?

extensions/wikia/PageShare/scripts/PageShare.js
```diff
@@ -72,8 +62,40 @@ require(['wikia.window', 'wikia.tracker', 'jquery'], function(win, tracker, $) {
...
+			return win.navigator.languages[0].substr(0, 2);
+		// Safari
+		} else if (win.navigator.language) {
+			return win.navigator.language.substr(0, 2);
+		// Internet Exploder
+		} else if (win.navigator.userLanguage) {
+			return win.navigator.userLanguage.substr(0, 2);
+		// something exotic
+		} else {
+			return null;
```
bognix: we can be extra fancy here and use wikia.log to log language.
Example usage:
```javascript
wikia.log([message], 'debug', page_share)
```
First argument is an array of messages which are concatenated with ","
Second argument is log level (debug in this case is enough)
Third argument is log group.

This way you can debug which language is chosen by appending to URL:
?log_level=9&log_group=pageShare

bognix:
@Warkot one small note from me, other than that, LGTM

extensions/wikia/PageShare/scripts/PageShare.js
Warkot: Awesome idea, thanks! Shall i do it like this?
```
wikia.log(['Couldn't pick a language for Page Share extension.'], 'debug', page_share);
```
Is it sufficient? Or should I add some variables as next messages in the array? Does it even work like that?

vforge:
:heavy_check_mark: LGTM from me

extensions/wikia/PageShare/PageShareHelper.class.php
Warkot: Great idea, I was thinking too much JavaScript where you can't even check if anything is out of index.

Warkot: I'm wondering if it is possible to use this function https://github.com/Wikia/app/blob/0d3b259720799abb42e9cc281d4faef65ffda12c/includes/OutputPage.php#L1817 → we wouldn't need to pass anything, just read the headers on the API call.

PR #7131, opened 2015-05-07
SOC-773 | Fix SQL injection for Wall

kvas-damian:
https://wikia-inc.atlassian.net/browse/SOC-773
ping @macbre

### DIFF
--- a/extensions/wikia/Wall/WallThread.class.php
+++ b/extensions/wikia/Wall/WallThread.class.php
@@ -95,7 +95,7 @@ class WallThread {
 		// the other one is in Wall.class done in a grouped way
 		// (fetch for many threads at once, set with ->setReplies)
 
-		$query = (new WikiaSQL())
+		$query = ( new WikiaSQL() )
 			->SELECT( 'distinct comment_id' )
 			->FROM( 'comments_index' )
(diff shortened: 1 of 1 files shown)


extensions/wikia/Wall/WallThread.class.php
```diff
@@ -95,25 +95,20 @@ private function getReplyIdsFromDB( $dbr, $afterId = null ) {
 		// the other one is in Wall.class done in a grouped way
 		// (fetch for many threads at once, set with ->setReplies)
 
-		$conditions = [ 'parent_comment_id = '.$this->mThreadId ];
```
macbre: Won't 

```php
$conditions = [ 'parent_comment_id' => $this->mThreadId ];
```

be enough here?

harnash:
LGTM but wait for @macbre to verfiy

extensions/wikia/Wall/WallThread.class.php
kvas-damian: Probably it would, but there is also ```'comment_id > '.$afterId``` which adds complexity in escaping for this function.

macbre: Ok, LGTM :+1:

PR #7135, opened 2015-05-07
Wait for OO to be ready before creating a Player

hakubo:
It looks like before ready event not all plugins are registered therfore we do not display AgeGate overlay


michalroszka:
:+1: - per our conversation, all explained.

PR #7137, opened 2015-05-07
PLATFORM-1199 CloseWiki cron: remove obsolete check

ghost:
We have migrated from NFS to DFS and /images is not mounted anymore
which makes is_dir check pointless and prevents the actual logic
from being executed.

As decided, full refactor will be a P4 (because of half a dozen of other urgent things)


macbre:
LGTM

PR #7141, opened 2015-05-07
ADEN-1989 Fix filling dynamic ad slots in Mercury

nandy-andy:
We found out that in Mercury some ad slots aren't being filled if it's not the first page load.

In changes below I fixed this issue. I changed the logic, so:
- we do not define and display ad divs which were already defined,
- we only create and append a div if it doesn't exist.

I also added some additional logging.


gabrys:
LGTM

PR #7147, opened 2015-05-07
SOC-774 Clear out Facebook session data on logout

garthwebb:
https://wikia-inc.atlassian.net/browse/SOC-774


jsutterfield:
LGTM

PR #7148, opened 2015-05-07
SOC-759 Add a access_token cookie and update file to PHP styleguide

garthwebb:
https://wikia-inc.atlassian.net/browse/SOC-759

### DIFF
--- a/extensions/wikia/Helios/User.class.php
+++ b/extensions/wikia/Helios/User.class.php
@@ -2,24 +2,31 @@
 
 namespace Wikia\Helios;
 
+use Wikia\Logger\WikiaLogger;
+
 /**
  * A helper class for dealing with user-related objects.
  */
 class User {
 
-	const ACCESS_TOKEN_COOKIE = 'access_token';
+	const ACCESS_TOKEN_COOKIE_NAME = 'access_token';
+
+	// This is set to 6 months, 6*30*24*60*60 = 15552000
+	const ACCESS_TOKEN_COOKIE_TTL = 15552000;
 
 	private static $authenticationCache = [];
 
 	/**
 	 * Logs character encoding data for the given password.
+	 *
+	 * @param string $password
+	 * @param string $callingMethod
 	 */
-	public static function debugLogin( $password, $callingMethod )
-	{
+	public static function debugLogin( $password, $callingMethod ) {
 		$detectedEncoding = mb_detect_encoding( $password );
 		$internalEncoding = mb_internal_encoding();
 
-		\Wikia\Logger\WikiaLogger::instance()->info( $callingMethod, [
+		WikiaLogger::instance()->info( $callingMethod, [
 			'byte_length'			=> strlen( $password ),
 			'character_length_detected'	=> mb_strlen( $password, $detectedEncoding ),
 			'character_length_internal'	=> mb_strlen( $password, $internalEncoding ),
@@ -36,7 +43,7 @@ class User {
 	 */
 	public static function getAccessToken( \WebRequest $request ) {
 		// A cookie takes precedence over an HTTP header.
-		$token = $request->getCookie( self::ACCESS_TOKEN_COOKIE, '' );
+		$token = $request->getCookie( self::ACCESS_TOKEN_COOKIE_NAME, '' );
 
 		// No access token in the cookie, try the HTTP header.
 		if ( ! $token ) {
@@ -65,8 +72,7 @@ class User {
 	 *
 	 * @return \User on successful authentication
 	 */
-	public static function newFromToken( \WebRequest $request )
-	{
+	public static function newFromToken( \WebRequest $request ) {
 		// Extract access token from HTTP request data.
 		$token = self::getAccessToken( $request );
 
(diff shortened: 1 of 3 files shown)


extensions/wikia/Helios/User.class.php
```diff
@@ -227,83 +240,105 @@ public static function onUserCheckPassword( $id, $username, $hash, $password, &$
...
+	/**
+	 * Called in ExternalUser_Wikia registers a user.
+	 *
+	 * @param string $username The username
+	 * @param string $password The plaintext password the user entered
+	 * @param string $email The user's email
+	 * @param string $birthDate
+	 *
+	 * @return bool true on success, false otherwise
+	 */
```
garthwebb: There were a bunch of spaces that had to be converted to tabs in the next two functions.

extensions/wikia/Helios/User.class.php
```diff
@@ -2,24 +2,31 @@
...
 /**
  * A helper class for dealing with user-related objects.
  */
 class User {
 
-	const ACCESS_TOKEN_COOKIE = 'access_token';
+	const ACCESS_TOKEN_COOKIE_NAME = 'access_token';
+
+	// This is set to 6 months, 6*30*24*60*60 = 15552000
+	const ACCESS_TOKEN_COOKIE_TTL = 15552000;
```
mklucsarits: In Mercury, the cookie TTL is 15778500 seconds, which is 182.621527777777 days. I guess someone wanted 6 months to be (very) precisely half a year. Anyway, I think the expiration for the cookie for the two platforms should be equal, so we should either make it 15778500 here or change it in Mercury.

garthwebb: Huh, so that works out to ```365.243055555554``` days which doesn't match any of the possible year lengths given on Wikipedia: http://en.wikipedia.org/wiki/Year#Summary

I think they should be the same, but calculated so nobody has to look at a wikipedia page to figure out why the value is anywhere from 5 minutes to 2 days off expected.  I'll update the value here to be ```(365/2)*24*60*60``` and ask for it to be updated on Mercury to the same.

mklucsarits: Sounds good to me.

mklucsarits:
The diff for indentation/spacing changes make this a bit confusing to follow, but from what I can see it LGTM.

garthwebb:
@mklucsarits, you can add ```?w=1``` to the URL and it will hide lines that only have spacing changes.  That makes change sets like this a lot easier to read.

includes/WebResponse.php
```diff
@@ -40,11 +42,11 @@ public function header( $string, $replace = true, $http_response_code = null ) {
...
-	 * @param $name String: name of cookie
-	 * @param $value String: value to give cookie
-	 * @param $expire Int: number of seconds til cookie expires
-	 * @param $prefix String: Prefix to use, if not $wgCookiePrefix (use '' for no prefix)
-	 * @param @domain String: Cookie domain to use, if not $wgCookieDomain
+	 * @param string $name : name of cookie
+	 * @param string $value : value to give cookie
+	 * @param int $expire : Expire time in seconds (a unix epoch time)
+	 * @param string $prefix : Prefix to use, if not $wgCookiePrefix (use '' for no prefix)
+	 * @param $domain String : Cookie domain to use, if not $wgCookieDomain
```
mklucsarits: I just noticed this param still has the wrong order for the ```@param``` tag.

garthwebb:
@michalroszka, can you review this change and lmk if we should be doing anything different to set the cookie?  There was some talk between @lizlux and @ArturKlajnerok about copying the cookie directly from the cookie Helios is setting, etc, etc.

jsutterfield:
LGTM.

michalroszka:
LGTM :+1:

ArturKlajnerok:
In my opinion in the end there should be only one place that sets the cookie. And as this is done in [helios](https://github.com/Wikia/helios) we should relay on this specific cookie.  But as we plan to remove `Special:Login` quite soon, this solution is acceptable. 

LGTM :+1:

extensions/wikia/Helios/User.class.php
```diff
@@ -2,24 +2,31 @@
...
 /**
  * A helper class for dealing with user-related objects.
  */
 class User {
 
-	const ACCESS_TOKEN_COOKIE = 'access_token';
+	const ACCESS_TOKEN_COOKIE_NAME = 'access_token';
+
+	// This is set to 6 months,(365/2)*24*60*60 = 15768000
+	const ACCESS_TOKEN_COOKIE_TTL = 15768000;
```
ArturKlajnerok: We still need to sync up about TTL.

lizlux:
@ArturKlajnerok, just to be clear, `Special:Login` will have to stick around until a couple dependencies are met: 
- Custom domains go away
- Wikia One and Uncyclopedia find different solutions

Last I checked there was no set timeline for either of those items, but it's possible we can disable it for the majority of wikias quite soon. 

I agree that we should consider relaying info like cookie name and TTL from the cookie helios sends.

PR #7151, opened 2015-05-08
SOC-763 Remove references to the 'math' option

garthwebb:
A pull request removing the math option.

https://wikia-inc.atlassian.net/browse/SOC-763


garthwebb:
@drsnyder, @nmonterroso, same comment as for #7151

PR #7152, opened 2015-05-08
SOC-763 Remove numberheadings option from code

garthwebb:
Yet another user property that isn't used and doesn't need to make the transition to our attribute/preferences services.

https://wikia-inc.atlassian.net/browse/SOC-763


drsnyder:
lgtm

garthwebb:
@drsnyder, @nmonterroso, same comment as for https://github.com/Wikia/app/pull/7151

PR #7154, opened 2015-05-08
render infobox as marker

idradm:


extensions/wikia/PortableInfobox/controllers/PortableInfoboxParserTagController.class.php
```diff
@@ -4,6 +4,21 @@ class PortableInfoboxParserTagController extends WikiaController {
 	const PARSER_TAG_NAME = 'infobox';
 	const INFOBOXES_PROPERTY_NAME = 'infoboxes';
 
+	private $count = 0;
```
SebastianMarzjan: could you name it more explicitly, like markerCount ?

PR #7158, opened 2015-05-08
PLATFORM-1032 Helios: change handling of non-200 HTTP responses

ghost:
Prior this change any non-200 response from the service was
considered an error and an exception was thrown.

As of now, non-200 responses are considered legit and no exception
is thrown.

The Client class has a new property - status - and the status object
is stored there. A getter method getStatus() has been added.

Handling specific responses in specific cases has to be implemented
outside the Client class.


michalroszka:
The failing test has nothing to do with the change and is failing on other branches, too.

macbre:
LGTM

PR #7160, opened 2015-05-08
CONCF-504 Trending Articles

dianafa:
Mercury: https://github.com/Wikia/mercury/pull/830

@vforge @bognix

### DIFF
--- a/extensions/wikia/MercuryApi/MercuryApiController.class.php
+++ b/extensions/wikia/MercuryApi/MercuryApiController.class.php
@@ -397,6 +397,6 @@ class MercuryApiController extends WikiaController {
 			WikiaLogger::instance()->info( 'Curated content and categories are empty' );
 		}
 
-	return $data;
+		return $data;
 	}
 }
--- a/extensions/wikia/MercuryApi/MercuryApiController.class.php
+++ b/extensions/wikia/MercuryApi/MercuryApiController.class.php
@@ -310,7 +310,7 @@ class MercuryApiController extends WikiaController {
 			$title = $this->wg->Title;
 		}
 
-		$data['adsContext'] = $this->mercuryApi->getAdsContext( $title );
+		$data[ 'adsContext' ] = $this->mercuryApi->getAdsContext( $title );
 
 		$this->response->setFormat( WikiaResponse::FORMAT_JSON );
 		$this->response->setCacheValidity( WikiaResponse::CACHE_STANDARD );
@@ -334,6 +334,7 @@ class MercuryApiController extends WikiaController {
 	private function getMainPageData() {
 		$mainPageData = [];
 		$curatedContent = $this->getCuratedContentData();
+		$trendingArticles = $this->getTrendingArticlesData();
 
 		if ( !empty( $curatedContent[ 'sections' ] ) ) {
 			$mainPageData[ 'curatedContent' ] = $curatedContent[ 'sections' ];
@@ -341,6 +342,10 @@ class MercuryApiController extends WikiaController {
 
 		if ( !empty( $curatedContent[ 'featured' ] ) ) {
 			$mainPageData[ 'featuredContent' ] = $curatedContent[ 'featured' ];
+		}
+
+		if ( !empty( $trendingArticles ) ) {
+			$mainPageData[ 'trendingArticles' ] = $trendingArticles;
 		}
 
 		return $mainPageData;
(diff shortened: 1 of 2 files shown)


vforge:
LGTM, just make it merge-able please :)

extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -352,6 +352,9 @@ private function getMainPageData() {
 		try {
 			$curatedContentData = $this->sendRequest( 'CuratedContent', 'getList' )->getData();
 			$mainPageData['curatedContent'] = $this->mercuryApi->getCuratedContent( $curatedContentData );
+			$params = [ 'expand' => true, 'abstract' => 0, 'limit' => 10, 'namespaces' => 14, 'height' => 50, 'width' => 50];
```
hakubo: can we add some \n around to make it more readable?

extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -352,6 +352,9 @@ private function getMainPageData() {
 		try {
```
hakubo: we should not reuse try's for different calls
as this will make stuff harder to debug
please add another try block if needed

extensions/wikia/MercuryApi/models/MercuryApi.class.php
```diff
@@ -276,6 +274,24 @@ public function processCuratedContent( $data ) {
 		return $data;
 	}
 
+	public function processTrendingArticles( $data ) {
```
hakubo: We are creating here an abstraction here over fairly new API

I would like us to start to think if we can use API data directly, and process them on client side
and if really we believe it is necessary 
maybe we should just consider creating v2 of those resources?

extensions/wikia/MercuryApi/models/MercuryApi.class.php
```diff
@@ -289,11 +305,52 @@ public function processCuratedContent( $data ) {
...
+	 */
+	private function processTrendingArticlesItem( $item ) {
+		global $wgArticlePath;
+
+		$processedItem = [];
+
+		if ( !empty( $item ) && is_array( $item ) ) {
+			$paramsMap = [
+				// params fetched from ArticlesApi => params passed to response
+				'title' => 'label',
```
hakubo: what is the real value of changing those keys?
Can't we just use what is provided?

Warkot: Yeah, we can, but we don't need to send all the data that's provided by the API to the client, so the needed properties are renamed while stripping the excessive ones.

extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -365,12 +370,33 @@ private function getCuratedContentData( $section = null ) {
...
+
+		return $data;
+	}
+
+	private function getTrendingArticlesData() {
+		$params = [
+			'abstract' => false,
+			'expand' => true,
+			'limit' => 10,
+			'namespaces' => '0'
```
rogatty: Should we use `$wgContentNamespaces` instead?

extensions/wikia/MercuryApi/models/MercuryApi.class.php
rogatty: I prefer to pollute this API which is going to be replaced anyway (some day, I hope!) and not our frontend. This way we can just swap this hacky thing to a new service.

IMO we should assume that there is a source with the correct data.

extensions/wikia/MercuryApi/models/MercuryApi.class.php
rogatty: @Warkot: we're not doing this for curated nor featured content, mapping is being done in the Mercury for these two. Would it be more consistent to do the same for trending articles?

Warkot: I can change this and drop the `id` and `type` as we don't need that in response.

vforge: :+1:

Warkot: Changed it here and appropriately in Mercury.

extensions/wikia/MercuryApi/models/MercuryApi.class.php
Warkot: Agreed with @rogatty.

rogatty:
LGTM

vforge:
LGTM

PR #7162, opened 2015-05-08
Portable infobox inner xml

jacekjursza:
Fix for bug found in DAT-2745

### DIFF
--- a/extensions/wikia/PortableInfobox/services/Parser/Nodes/Node.php
+++ b/extensions/wikia/PortableInfobox/services/Parser/Nodes/Node.php
@@ -51,8 +51,11 @@ class Node {
 	}
 
 	protected function getInnerXML( \SimpleXMLElement $node ) {
-		$tag = $node->getName();
-		return preg_replace( '!<'. $tag .'(?:[^>]*)>(.*)</'. $tag .'>!Ums', '$1', $node->asXml() );
+		$innerXML= '';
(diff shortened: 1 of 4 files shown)


extensions/wikia/PortableInfobox/services/Parser/Nodes/Node.php
```diff
@@ -50,6 +50,14 @@ public function isEmpty( $data ) {
 		return !( isset( $data[ 'value' ] ) ) || empty( $data[ 'value' ] );
 	}
 
+	protected function getInnerXML( \SimpleXMLElement $node ) {
```
idradm: looks awesome comparing to regex, have you looked at performance?

idradm:
LG, however I'm not sure if those are all nodes. What about default tag?

jacekjursza:
@idradm Default tag is already handled here: https://github.com/Wikia/app/pull/7162/files#diff-b09ea7c9a303b75f3ed56e8a326e0e0cR74

extensions/wikia/PortableInfobox/services/Parser/Nodes/Node.php
jacekjursza: No.  It shouldn't be relevant in such small XML as infobox template.

idradm:
mkey

PR #7166, opened 2015-05-08
Get items from curated content items response

bognix:


bognix:
Unit tests failure not related with code change.

PR #7167, opened 2015-05-08
SOC-608 Use new email extension for Message Wall notifications

mklucsarits:
https://wikia-inc.atlassian.net/browse/SOC-608

To make sure that our code is compatible with SOC-611 I merged it here ;)
@garthwebb @jsutterfield


extensions/wikia/Email/Controller/CommentController.class.php
```diff
@@ -96,13 +96,6 @@ public function getSubject() {
 
 	abstract protected function getSubjectKey();
 
-	protected function getSalutation() {
```
mklucsarits: Now uses EmailController::getSalutation()

extensions/wikia/Email/Controller/WatchedPageController.class.php
```diff
@@ -98,14 +98,6 @@ public function body() {
 	/**
 	 * @return String
 	 */
-	private function getSalutation() {
```
mklucsarits: Now uses EmailController::getSalutation()

extensions/wikia/WallNotifications/WallNotifications.class.php
```diff
@@ -370,12 +370,11 @@ protected function createKeyForMailNotification( $watcher, WallNotificationEntit
...
-		$text = strip_tags( $notification->data_noncached->msg_text, '<p><br>' );
+		$text = preg_replace( array( '#<br\s*/?>#i', '#</p\s*/?>#i' ), "\n",
+			$notification->data_noncached->msg_text );
+		$text = trim( strip_tags( $text ) );
 		$text = substr( $text, 0, 3000 ) . ( strlen( $text ) > 3000 ? '...' : '' );
-
-		$textNoHtml = preg_replace( '#<br\s*/?>#i', "\n", $text );
-		$textNoHtml = trim( preg_replace( '#</?p\s*/?>#i', "\n", $textNoHtml ) );
-		$textNoHtml = substr( $textNoHtml, 0, 3000) . ( strlen( $textNoHtml ) > 3000 ? '...' : '' );
+		$textNoNewlines = str_replace( "\n", ' ', $text );
```
mklucsarits: I cleaned this up to remove duplicated code and to fix a bug where opening <p> tags remained in the text. Also renamed the variables to better represent what they contain.

extensions/wikia/WallNotifications/WallNotifications.class.php
```diff
@@ -390,8 +389,7 @@ protected function sendEmails( array $watchers, WallNotificationEntity $notifica
 			if ( !empty( $mode ) && $watcher->getId() != 0 && (
 				( $mode == WALL_EMAIL_EVERY ) ||
 				( $mode == WALL_EMAIL_SINCEVISITED && empty( $this->uniqueUsers[$entityKey][$watcher->getId()] ) )
-			)) {
-
+			) && $watcher->getBoolOption( 'unsubscribed' ) !== true ) {
```
mklucsarits: Condition moved here from further down.

extensions/wikia/Email/Controller/WallMessageController.class.php
```diff
@@ -0,0 +1,111 @@
+<?php
+
+namespace Email\Controller;
+
+use Email\EmailController;
+
+class WallMessageController extends EmailController {
+
+	/**
+	 * @template wallMessage
```
garthwebb: You should use the avatarLayout template for this email type rather than define a new one.  Its meant to be reused for all emails that share this format.

extensions/wikia/Email/i18n/WallMessage.i18n.php
```diff
@@ -0,0 +1,54 @@
...
+);
+
+$messages['qqq'] = array(
+	'emailext-wallmessage-owned-subject' => 'Email subject when another user wrote a new message on your wall. $1 is the name of the user who posted the message.',
+	'emailext-wallmessage-following-subject' => "Email subject when a user wrote a new message on another user's wall, and you are following the thread. $1 is the name of the user who posted the message, $2 is the wall owner (user).",
+	'emailext-wallmessage-owned-summary' => 'Text describing that another user wrote a new message on your wall. $1 is the name of the user who posted the message',
+	'emailext-wallmessage-following-summary' => "Text describing that a user wrote a new message on another user's wall. $1 is the name of the user who posted the message, $2 is the wall owner (user).",
+	'emailext-wallmessage-full-conversation' => 'Text for button that, when clicked, navigates to the full message wall conversation',
+	'emailext-wallmessage-recent-messages' => "Call to action to view all recent messages on the user's message wall",
+);
```
garthwebb: I'm not sure if you added the following empty message arrays or if crowdin did, but in case you did, just note that crowdin will handle it for you ;)

extensions/wikia/Email/Controller/WallMessageController.class.php
```diff
@@ -0,0 +1,111 @@
...
+	/**
+	 * Get the email subject line
+	 *
+	 * @return string
+	 */
+	public function getSubject() {
+		if ( $this->isForFollowedThread() ) {
+			return wfMessage( 'emailext-wallmessage-following-subject',
+				$this->request->getVal( 'fromName' ),
+				$this->request->getVal( 'wallUserName' )
```
jsutterfield: So it might be nice to validate these values before we get here. The pattern we've been kind of following is to put these into instance variables inside of initEmail() and then validate them there. That way we can throw an exception if we don't get passed the right values or they're invalid.

It doesn't really come up when the email is sent by the specified user action, but as we're testing emails/sending them manually that kind of feedback is really helpful. Like this: https://github.com/Wikia/app/blob/dev/extensions/wikia/Email/Controller/WatchedPageController.class.php#L26-L67

extensions/wikia/WallNotifications/WallNotifications.class.php
garthwebb: The '<p><br>' string is the list of allowable tags for the ```strip_tags``` function (https://php.net/strip_tags), so it was intentional for the <p> tag to be in there.  Was this causing display problems in the email?

garthwebb: Oh, I see, I missed that the old code then changed those to newlines...

extensions/wikia/WallNotifications/WallNotifications.class.php
```diff
@@ -370,12 +370,11 @@ protected function createKeyForMailNotification( $watcher, WallNotificationEntit
 	}
 
 	protected function sendEmails( array $watchers, WallNotificationEntity $notification ) {
-		$text = strip_tags( $notification->data_noncached->msg_text, '<p><br>' );
+		$text = preg_replace( array( '#<br\s*/?>#i', '#</p\s*/?>#i' ), "\n",
```
garthwebb: Lets use the new array syntax here ```[]```

extensions/wikia/Email/templates/wallMessage.mustache
```diff
@@ -0,0 +1,49 @@
...
+					{{# editorProfilePage}}
+						<a class="username" href="{{editorProfilePage}}">
+					{{/ editorProfilePage}}
+						{{editorUserName}}
+					{{# editorProfilePage}}
+						</a>
+					{{/ editorProfilePage}}
+				</p>
+				<strong>{{wallMessageTitle}}</strong><br><br>
+				<p>{{wallMessageBody}}</p>
```
jsutterfield: Hmm, this is a ton of duplicated code. The only real difference between this file and avatarLayout.mustache that I can see are L30-31. What we could do is put the contents of those 2 lines into the the `{{details}}` variable [here](https://github.com/Wikia/app/blob/dev/extensions/wikia/Email/templates/avatarLayout.mustache#L30) inside of avatarLayout. That way we can just reuse the one template.

Also, I feel like we shouldn't be using the `<br>` tags. Seems like something we should be doing with CSS? The wallMessageTitle should probably be in a `<p>` tag or something anyways, yea?

extensions/wikia/WallNotifications/WallNotifications.class.php
jsutterfield: We're actually doing this check inside of the Email extension [here](https://github.com/Wikia/app/blob/dev/extensions/wikia/Email/EmailController.class.php#L513-L517), so no need to check this.

extensions/wikia/Email/templates/wallMessage.mustache
mklucsarits: Yeah I'll definitely re-use the avatarLayout template instead.

extensions/wikia/WallNotifications/WallNotifications.class.php
garthwebb: The original code didn't strip out the newlines; is there a reason to do that here?  Also, the variable as used still contained the ```<br>``` and ```<p>``` tags.  It seems like ```$textNoHtml``` was probably the right (intended result) name but they just didn't do a good job with the regex since it doesn't account for parameters for the ```<p>``` tag.

Maybe the we should take the original, use the array form of ```preg_replace``` you introduced and then add a strip_tags afterward.

extensions/wikia/WallNotifications/WallNotifications.class.php
garthwebb: This conditional is getting a little unwieldy.  Maybe it would be good to split this up into several, functionally grouped conditionals that issue a ```continue``` of the loop.  That would also reduce the indentation of this block, improving readability.

extensions/wikia/Email/Controller/WallMessageController.class.php
```diff
@@ -0,0 +1,111 @@
...
+	 * @template wallMessage
+	 */
+	public function body() {
+		$this->response->setData( [
+			'salutation' => $this->getSalutation(),
+			'summary' => $this->getSummary(),
+			'editorProfilePage' => $this->getCurrentProfilePage(),
+			'editorUserName' => $this->request->getVal( 'fromName' ),
+			'editorAvatarURL' => $this->getCurrentAvatarURL(),
+			'wallMessageTitle' => $this->request->getVal( 'messageTitle' ),
```
garthwebb: We've been creating new instance variables for any controller parameters that aren't already known by the parent controller in an ```initEmail``` method.  This makes it easy to see what a controller intends on using and provides an opportunity to do error checking before starting.

extensions/wikia/WallNotifications/WallNotifications.class.php
```diff
@@ -404,34 +402,28 @@ protected function sendEmails( array $watchers, WallNotificationEntity $notifica
...
+					$this->sendEmail( $watcher, $data );
+				} else {
+					$fromUser = new MailAddress( $watcher );
+					$params = [
+						'targetUser' => $watcher->getName(),
+						// Wall message title
+						'messageTitle' => $notification->data->thread_title,
+						// Wall message body
+						'messageBody' => $text,
+						'messageType' => $messageType,
```
garthwebb: I don't see ```$messageType``` defined anywhere.  What is this supposed to be?

extensions/wikia/WallNotifications/WallNotifications.class.php
```diff
@@ -404,34 +402,28 @@ protected function sendEmails( array $watchers, WallNotificationEntity $notifica
...
+						'replyToAddress' => $this->replyto,
```
garthwebb: $this->replyto doesn't seem to be defined either ...

extensions/wikia/Email/Controller/WallMessageController.class.php
```diff
@@ -0,0 +1,111 @@
...
+	/**
+	 * Get the localized text describing the recent messages link
+	 *
+	 * @return string
+	 */
+	protected function getRecentMessagesText() {
+		$wallTitleText = $this->isForFollowedThread() ? $this->request->getVal( 'wallUserName' ) :
+			$this->request->getVal( 'targetuser' );
+		return wfMessage( 'emailext-wallmessage-recent-messages',
+			\Wall::newFromTitle( \Title::newFromText( $wallTitleText ) )->getUrl()
```
garthwebb: One note here; you should be careful with the one argument form of ```Title::newFromText```.  The second argument is the namespace which defaults to ```NS_MAIN``` (e.g. /wiki/SomeTitle), so if you're trying to load a title that isn't in the ```NS_MAIN``` namespace (e.g., a file, /wiki/File:SomeTitle), you'll end up either:
1) Loading the wrong Title object if there is a Title with the same name in the ```NS_MAIN``` namespace
2) Loading an empty Title object if there isn't a Title with the same name.
In the case above its actually #2 since you're trying to load a Title in the ```NS_USER_WALL``` namespace (e.g., /wiki/Message_Wall:Garthwebb), however ```Wall::newFromTitle``` *happens* to make calls on the empty Title object that don't make DB requests and  essentially returns it the ```$wallTitleText``` you passed in to construct the proper URL.

extensions/wikia/Email/Controller/WallMessageController.class.php
```diff
@@ -0,0 +1,161 @@
...
+		return $targetName == $this->receiverUserName;
+	}
+
+	/**
+	 * Get the email subject line
+	 *
+	 * @return string
+	 */
+	public function getSubject() {
+		if ( $this->isForTargetUser() ) {
```
kvas-damian: I see couple conditions of this type in this class. What do you think about creating separate classes to handle those cases?

extensions/wikia/Email/Controller/WallMessageController.class.php
```diff
@@ -0,0 +1,161 @@
...
+		$ownerWall = \Title::newFromText( $this->wallUserName, NS_USER_WALL );
+		$wallUrl = $ownerWall->getFullURL();
+
+		return wfMessage( 'emailext-wallmessage-recent-messages', $wallUrl )
+			->inLanguage( $this->targetLang )->parse();
+	}
+
+	protected function getFooterMessages() {
+		$footerMessages = [
+			wfMessage( 'emailext-unfollow-text' )->inLanguage( $this->targetLang )->parse()
```
kvas-damian: Don't we need parameters for this message?

extensions/wikia/WallNotifications/WallNotifications.class.php
```diff
@@ -372,73 +371,81 @@ protected function createKeyForMailNotification( $watcher, WallNotificationEntit
...
+					'threadUserName' => $notification->data->parent_username,
+					'replyUserName' => $notification->data->msg_author_username,
+					'isThread' => $notification->isMain(),
+
+					'messageTitle' => $notification->data->thread_title,
+					'messageBody' => $textNoHtml,
+					'messageUrl' => $notification->data->url
+				];
 
-				if ( !( $watcher->getBoolOption('unsubscribed') === true ) ) {
```
kvas-damian: Don't we need this check?

extensions/wikia/WallNotifications/WallNotifications.class.php
```diff
@@ -372,73 +371,81 @@ protected function createKeyForMailNotification( $watcher, WallNotificationEntit
 	protected function sendEmails( array $watchers, WallNotificationEntity $notification ) {
 		$text = strip_tags( $notification->data_noncached->msg_text, '<p><br>' );
 		$text = substr( $text, 0, 3000 ) . ( strlen( $text ) > 3000 ? '...' : '' );
```
kvas-damian: While we're touching this code, can we use multibyte safe functions here?

extensions/wikia/WallNotifications/WallNotifications.class.php
```diff
@@ -372,73 +371,81 @@ protected function createKeyForMailNotification( $watcher, WallNotificationEntit
...
+
+			if ( !empty( $data ) ) {
+				// The old email delivery has been kept here for forums, which use the hook above.
+				// This logic can be removed when forum emails are migrated to the new email extension
+				// or when forums are retired.
+				$this->sendEmail( $watcher, $data );
+			} else {
+				$params = [
+					'targetUser' => $watcherName,
+					'fromAddress' => $watcher->getEmail(),
```
kvas-damian: Should we user watcher email address as a from address?

extensions/wikia/WallNotifications/WallNotifications.class.php
```diff
@@ -372,73 +371,81 @@ protected function createKeyForMailNotification( $watcher, WallNotificationEntit
...
+
+					'wallUserName' => $notification->data->wall_username,
+					'threadUserName' => $notification->data->parent_username,
+					'replyUserName' => $notification->data->msg_author_username,
+					'isThread' => $notification->isMain(),
+
+					'messageTitle' => $notification->data->thread_title,
+					'messageBody' => $textNoHtml,
```
kvas-damian: shouldn't we use here ```$text```? It includes ```<br>``` and ```<p>``` tags which helps to keep content readable in HTML emails.

extensions/wikia/WallNotifications/WallNotifications.class.php
```diff
@@ -372,73 +371,81 @@ protected function createKeyForMailNotification( $watcher, WallNotificationEntit
...
 		return true;
 	}
 
+	protected function canSendToWatcher( User $watcher, $entityKey ) {
+		if ( $watcher->getId() == 0 ) {
+			return false;
+		}
+
+		// Don't send the same email twice
+		if ( !empty( $this->uniqueUsers[$entityKey][$watcher->getId()] ) ) {
```
jsutterfield: It looks like in the old logic we only checked this if `$mode == WALL_EMAIL_SINCEVISITED`. When `$mode == WALL_EMAIL_EVERY`, we wouldn't do this check.

extensions/wikia/WallNotifications/WallNotifications.class.php
garthwebb: Yes, I removed because @jsutterfield noted we already do this in EmailController but we need it here until we move all emails from this class over to the new system.

extensions/wikia/WallNotifications/WallNotifications.class.php
```diff
@@ -406,39 +417,37 @@ protected function sendEmails( array $watchers, WallNotificationEntity $notifica
...
-						'$MSG_KEY_SUBJECT' => $key,
-						'$MSG_KEY_BODY' => 'mail-notification-body',
-						'$MSG_KEY_GREETING' => 'mail-notification-html-greeting',
-					];
-				}
 
-				if ( !( $watcher->getBoolOption('unsubscribed') === true ) ) {
-					$this->sendEmail( $watcher, $data );
-				}
+				$this->sendEmail( $watcher, $data );
```
garthwebb: Maybe we should put some kibana logging here to see if this code path is used after the release?  I don't think it should be and it would be nice to confirm.  Then we can take this out...

kvas-damian: SOC-457 still uses this one.

extensions/wikia/WallNotifications/WallNotifications.class.php
```diff
@@ -464,9 +473,28 @@ protected function sendEmail( User $watcher, array $data ) {
...
 	}
 
+	protected function getEmailExtensionController( $notification, $watcherName ) {
+		$controller = false;
+
+		if ( !empty( $notification->data->article_title_ns )
+			&& MWNamespace::getSubject( $notification->data->article_title_ns ) == NS_WIKIA_FORUM_BOARD
+			&& $notification->isMain()
+		) {
+			$controller = 'Email\Controller\Forum';
```
garthwebb: I think Nirvana will let you call the controller with or without the ```Controller``` suffix, but we should probably be consistent.  It'd be good to add ```Controller``` here or remove it on L485,487,489

kvas-damian: Done!

extensions/wikia/WallNotifications/WallNotifications.class.php
```diff
@@ -371,29 +372,39 @@ protected function createKeyForMailNotification( $watcher, WallNotificationEntit
...
 	protected function sendEmails( array $watchers, WallNotificationEntity $notification ) {
 		$text = strip_tags( $notification->data_noncached->msg_text, '<p><br>' );
-		$text = substr( $text, 0, 3000 ) . ( strlen( $text ) > 3000 ? '...' : '' );
-
-		$textNoHtml = preg_replace( '#<br\s*/?>#i', "\n", $text );
-		$textNoHtml = trim( preg_replace( '#</?p\s*/?>#i', "\n", $textNoHtml ) );
-		$textNoHtml = substr( $textNoHtml, 0, 3000) . ( strlen( $textNoHtml ) > 3000 ? '...' : '' );
+		$text = mb_substr( $text, 0, self::MAX_ABSTRACT_LENGTH, 'UTF-8' )
+			. ( mb_strlen( $text, 'UTF-8' ) > self::MAX_ABSTRACT_LENGTH ? '...' : '' );
+		$textNoHtml = trim( preg_replace( '#</?(p|br)[^/>]*/?>#i', "\n", $text ) );
```
mklucsarits: Be sure to check the output of this. The original code had a bug that output opening <p> tags with no closing </p> tags.

garthwebb: Can you find example text of where this happens?  The only way I could get this to happen is when the ```<p>``` tag contained parameters, which should be addressed by the updated regex (i.e. ```[^/>]*``` instead of ```\s*```)

mklucsarits: Yeah, I saw it happening only when the tag had attributes.

extensions/wikia/WallNotifications/WallNotifications.class.php
```diff
@@ -464,9 +473,28 @@ protected function sendEmail( User $watcher, array $data ) {
 		return $watcher->sendMail( $data['$MAIL_SUBJECT'], $text, $from, $replyTo, 'WallNotification', $html );
 	}
 
+	protected function getEmailExtensionController( $notification, $watcherName ) {
```
jsutterfield: Can we type hint `$notification` here?

extensions/wikia/WallNotifications/WallNotifications.class.php
```diff
@@ -371,29 +372,39 @@ protected function createKeyForMailNotification( $watcher, WallNotificationEntit
...
-		if ( empty( $this->uniqueUsers[$entityKey] ) ){
+		if ( empty( $this->uniqueUsers[$entityKey] ) ) {
 			$this->uniqueUsers[$entityKey] = [];
 		}
 
-		foreach ( $watchers as $val ){
-			$watcher = $this->getUser( $val );
-			$mode = $watcher->getOption( 'enotifwallthread' );
-
-			if ( !empty( $mode ) && $watcher->getId() != 0 && (
```
jsutterfield: All the logic here got moved into `canSendToWatcher()`, but it doesn't look like it's being called any more. Should we add a call to `canSendToWatcher()` somewhere around here? Like above the call to `getEmailExtensionController()`?

extensions/wikia/WallNotifications/WallNotifications.class.php
```diff
@@ -371,29 +372,39 @@ protected function createKeyForMailNotification( $watcher, WallNotificationEntit
 
 	protected function sendEmails( array $watchers, WallNotificationEntity $notification ) {
 		$text = strip_tags( $notification->data_noncached->msg_text, '<p><br>' );
-		$text = substr( $text, 0, 3000 ) . ( strlen( $text ) > 3000 ? '...' : '' );
-
-		$textNoHtml = preg_replace( '#<br\s*/?>#i', "\n", $text );
-		$textNoHtml = trim( preg_replace( '#</?p\s*/?>#i', "\n", $textNoHtml ) );
-		$textNoHtml = substr( $textNoHtml, 0, 3000) . ( strlen( $textNoHtml ) > 3000 ? '...' : '' );
+		$text = mb_substr( $text, 0, self::MAX_ABSTRACT_LENGTH, 'UTF-8' )
```
jsutterfield: Not crucial, but all this text preparation seems like a good candidate for it's own method.

extensions/wikia/WallNotifications/WallNotifications.class.php
```diff
@@ -406,39 +417,37 @@ protected function sendEmails( array $watchers, WallNotificationEntity $notifica
 				wfRunHooks( 'NotificationGetMailNotificationMessage', [
```
jsutterfield: Is there a way that `$data` above won't be populated after the call to this hook?

kvas-damian: This code will be removed in SOC-457

extensions/wikia/WallNotifications/WallNotifications.class.php
```diff
@@ -406,39 +420,40 @@ protected function sendEmails( array $watchers, WallNotificationEntity $notifica
...
+		// Don't send an email to users that unsubscribed their email address
+		if ( $watcher->getBoolOption( 'unsubscribed' ) === true ) {
+			return false;
+		}
+
+		$mode = $watcher->getOption( 'enotifwallthread' );
+		if ( empty( $mode ) ) {
+			return false;
+		}
+
```
garthwebb: @kvas-damian FYI, I split this logic up a bit because I accidentally started working on it just before you made your latest changes ;)  This is the only difference we had that I wanted to include...

kvas-damian: Looks ok for me.

garthwebb:
Whew, LGTM

extensions/wikia/WallNotifications/WallNotifications.class.php
```diff
@@ -464,9 +479,41 @@ protected function sendEmail( User $watcher, array $data ) {
...
+		} else {
+			$controller = 'Email\Controller\FollowedWallMessage';
+		}
+
+		return $controller;
+	}
+
+	/**
+	 * Get abstract from whole entity text.
+	 * HTML tags are stripped out, expect <p> <br> which are used to make it easier to read.
```
jsutterfield: Should probably be "HTML tags are stripped out, *except* \<p\> \<br\> which are used to make it easier to read."

kvas-damian:
This branch is already in: https://github.com/Wikia/app/pull/7241

extensions/wikia/WallNotifications/WallNotifications.class.php
kvas-damian: fixed here: https://github.com/Wikia/app/commit/7bfde7c3c29857af7af6559c727136979c425850

PR #7111, opened 2015-05-05
SOC-620 | WatchedPaged emails

kvas-damian:
Emails that are sent when action is performed on page that is followed by user:
- [refactored] email on edit
- [new] email on protect
- [new] email on unprotect
- [new] email on rename
- [new] email on delete

PR #7168, opened 2015-05-08
SOC-464 Create Special Page to Allow Sending of Arbitrary Emails

jsutterfield:
This PR creates a new Special page, Special:SendEmail, which presents a list of forms that Wikia staff can use to send off any of our new emails. Each of the forms has inputs for the required parameters, and the page will report back either success, or errors which were encountered when sending the email.

@kvas-damian, I think you and I are going to have to sync up a little due to [these changes](https://github.com/Wikia/app/pull/7111/files#diff-e6ebc490f598ccb0fc53eca04f1174f2R42) from SOC-620. Some of the forms for Special:SendEmail rely on the user being able to pass in arbitrary revision ids. We can talk about a good solution on Monday. We can even talk in _person_ :D


extensions/wikia/Email/SpecialSendEmailController.class.php
```diff
@@ -0,0 +1,138 @@
...
+	private function addCss() {
+		$this->response->addAsset( 'special_email_admin_css' );
+	}
+
+	/**
+	 * @template specialSendEmail
+	 */
+	public function index() {
+
+		if ( $this->wg->request->wasPosted() ) {
```
Grunny: This should also [protect against CSRF](https://one.wikia-inc.com/wiki/Security/Cross-site_request_forgery#How_to_implement_CSRF_protection_on_Wikia.27s_version_of_MediaWiki) using `User::matchEditToken` and having the form contain the user's edit token in a hidden field.

extensions/wikia/Email/styles/specialSendEmail.scss
```diff
@@ -0,0 +1,12 @@
...
+	margin-bottom: 15px;
+
+	input[type="submit"] {
+		margin-top: 5px;
+	}
+
+	.input-group {
+		clear: none;
+	}
+}
```
kvas-damian: New line missing ;)

extensions/wikia/Email/SpecialSendEmailController.class.php
```diff
@@ -0,0 +1,138 @@
...
+			$result = $this->processForm();
+			$this->addBannerNotification( $result );
```
kvas-damian: What do you think about redirecting user to GET after successful email send? This could avoid resending email on page refresh by user.

extensions/wikia/Email/SpecialSendEmailController.class.php
```diff
@@ -0,0 +1,138 @@
...
+	/**
+	 * Get a list of all of our Email Controller classes. Each of these controller
+	 * classes represents one of the types of emails that we send.
+	 * @return array
+	 */
+	public function getEmailControllerClasses() {
+		$allClasses = \F::app()->wg->AutoloadClasses;
+		$emailControllerClasses = [];
+		foreach ( $allClasses as $className => $classPath ) {
+			if ( preg_match( "/^Email\\\\Controller\\\\.+Controller$/", $className, $matches ) ) {
```
kvas-damian: Is it possible to put this pattern in constant and reuse it here: https://github.com/Wikia/app/pull/7168/files#diff-ca8a4d565df752f3feb118ff94fcf028R620 ?

extensions/wikia/Email/styles/specialSendEmail.scss
jsutterfield: So I've seen IDE's complain about that. From what I can tell it's not in our guidelines, is it just an overall best practice for any source code?

extensions/wikia/Email/SpecialSendEmailController.class.php
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	/**
+	 * Makes sure the current user is staff and can therefore access this page. If not, display
+	 * a permission error page.
+	 * @throws \PermissionsError
+	 */
+	private function assertCanAccess() {
+		if ( !$this->wg->User->isStaff() ) {
+			throw new \PermissionsError( self::REQUIRED_PERMISSION );
```
garthwebb: So ```staff``` is actually a group rather than a permission.  Users belong to groups, groups have permission to do certain actions.  The ```PermissionsError``` constructor wants the action that was attempted, something like 'edit' or 'update' etc and constructs a message string from it like ```action-edit``` to use in the error response.

There is actually a ```sendemail``` action defined that you could use for the purpose of this error.  If you wanted to really be by the books, you'd create your own action for this case (e.g. ```testemail```), set it up so that the ```staff``` group has permission to do it, then do ```$user->can( 'testemail' )```.  Could pass ```testemail``` to ```PermissionsError```, and then add a i18n key ```action-testemail```.

Or you could just reuse ```sendemail``` ;)

extensions/wikia/Email/SpecialSendEmailController.class.php
jsutterfield: So I had originally been redirecting to a GET, but the banner notifications weren't working when I did that which is what I'm using for user feedback, both success and errors.

If this were a user facing feature I would've spent the time to find a better solution, but for an internal admin tool I figured it was easiest to just allow for the possibility of re-posting by doing a refresh and have the banner notifications work.

kvas-damian: I remember BannerNotifications working after redirect. They're stored in session. Can we double check that?

extensions/wikia/Email/Controller/CommentController.class.php
```diff
@@ -171,4 +210,27 @@ protected function getSubjectKey() {
...
+			[
+				'type' => 'text',
+				'name' => 'pageTitle',
+				'label' => "Blog Post Title",
+				'tooltip' => "eg 'Gcheung28/New_sharing_options_on_Wikia' (make sure it's on this wikia!)",
+			],
+			[
+				'type' => 'hidden',
+				'name' => 'namespace',
+				'value' => 500
```
garthwebb: You can use ```NS_BLOG_ARTICLE``` here

extensions/wikia/Email/SpecialSendEmailController.class.php
garthwebb: I was actually contemplating the opposite actually; repopulate the form with the last values it had.  Only staff users will be using this and I expect they would want to send several repeat emails with slightly different values or at the very least, a different language from the pulldown.

extensions/wikia/Email/SpecialSendEmailController.class.php
jsutterfield: Good call, @Grunny I'll update.

extensions/wikia/Email/SpecialSendEmailController.class.php
jsutterfield: Okay, I'll have the form be repopulated with values to make it easier to send multiple emails with slightly different values so I won't do the GET redirect. Turns out @kvas-damian, you were right about the banner notifications though, they can work with a GET redirect. That link you shared showed the secret, I had to use `$this->wg->Out->redirect()` as opposed to `$this->response->redirect()`.

extensions/wikia/Email/SpecialSendEmailController.class.php
jsutterfield: Ah, good to know! Obviously I didn't have a full grasp of how permissions worked in Mediawiki, thanks for clarifying. I think I'll just go ahead and create a new right, since sendemail is already being used (and to test out my new Mediawiki access control knowledge :) ).

extensions/wikia/Email/SpecialSendEmailController.class.php
jsutterfield: Yes, it definitely is. Particularly b/c I shamelessly stole that pattern from [here](https://github.com/Wikia/app/blob/dev/maintenance/wikia/email.php#L189), so I can update it there as well.

extensions/wikia/Email/styles/specialSendEmail.scss
kvas-damian: Yeah, I think it's overall best practise.
http://stackoverflow.com/questions/729692/why-should-files-end-with-a-newline

extensions/wikia/Email/Email.setup.php
```diff
@@ -38,3 +44,11 @@
 $wgExtensionMessagesFiles['Email'] = $dir . 'Email.i18n.php';
 $wgExtensionMessagesFiles['WatchedPage'] = $dir . 'i18n/WatchedPage.i18n.php';
 $wgExtensionMessagesFiles['Comment'] = $dir . 'i18n/Comment.i18n.php';
+$wgExtensionMessagesFiles['SpecialSendEmail'] = $dir . 'i18n/specialSendEmail.i18n.php';
+
+
+/**
+ * permissions
+ */
+$wgDefaultUserOptions['access-sendemail'] = false;
```
kvas-damian: I'm not sure if ```$wgDefaultUserOptions``` is correct variable here.
I'd say that ```$wgGroupPermissions['*']['access-sendemail'] = false;``` looks better ;)

kvas-damian:
LGTM

extensions/wikia/Email/EmailController.class.php
```diff
@@ -526,4 +530,106 @@ public function assertUserNotBlocked() {
...
+					'type' => 'hidden',
+					'name' => 'fromName',
+					'value' => \F::app()->wg->PasswordSenderName
+				],
+				[
+					'type' => 'text',
+					'name' => 'targetUser',
+					'label' => 'Target User',
+					'tooltip' => 'User to send the email to',
+					'value' => \F::app()->wg->User->getName()
```
Grunny: The WikiaStyleGuideForm does not escape values by default unfortunately (we should fix that), so if the user has a quote in their username, this will break the HTML of the form.

jsutterfield: Ah, good to know. This is a special page which is intended only for staff to use so I don't think it would come up to much, but it's easy enough to just escape the name. I'll do that.

PR #7169, opened 2015-05-11
(CE-1704,CE-1705) Add DMCA management for submitting to Chilling Effects

Grunny:
Add DMCA request management to allow staff to submit stored DMCA requests
to Chilling Effects.

This is now simpler than the initial version of the submission to CE as we are
redacting more information.


Grunny:
Unit test failure is unrelated.

extensions/wikia/DMCARequest/ChillingEffectsClient.class.php
```diff
@@ -1,21 +1,146 @@
...
 	private $apiToken;
 
-	public function __construct( $apiToken ) {
+	/**
+	 * Create client object.
+	 *
+	 * @param string $baseUrl  API URL.
+	 * @param string $apiToken API token.
+	 */
+	public function __construct( $baseUrl = '', $apiToken = '' ) {
```
lgarczewski: This may be nit-picking, but can the API token and base URL really be empty? :) What happens if they are not given?

lgarczewski:
Small comment above, other than that LGTM.

extensions/wikia/DMCARequest/ChillingEffectsClient.class.php
Grunny: Ah, yeah, that's just left over from me being lazy while testing the methods that didn't need it at the command line. ;)

PR #7179, opened 2015-05-12
Proper Tracker.track() parameter send in legacy code.

harnash:
Leftovers after changing Tracker.track() method parameters.


macbre:
LGTM

PR #7181, opened 2015-05-12
CreateWiki: wait for slaves to catch up

macbre:
First of all: I'm going to burn in hell until the end of time for this commit.

Long story short: CreateWiki process redirect to the newly created wiki which pretty often results with a fatal / blank page due to slave not catching up and missing the wiki database or lacking part of required tables.

Sleep for a second in `wfWaitForSlaves`  when called for wiki from G cluster.

@drozdo / @michalroszka / @harnash / @wladekb / @owend

### DIFF
--- a/includes/GlobalFunctions.php
+++ b/includes/GlobalFunctions.php
@@ -3826,6 +3826,7 @@ function wfWaitForSlaves( $wiki = false ) {
 			# not just the one returned by consul
 			sleep( 1 );
 
+			/* @var MySQLMasterPos $pos */
 			\Wikia\Logger\WikiaLogger::instance()->info( 'wfWaitForSlaves for consul clusters',  [
 				'exception' => new Exception(),
 				'master' => $masterHostName,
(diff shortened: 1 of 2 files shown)


drozdo:
sleep FTW! :)

includes/GlobalFunctions.php
```diff
@@ -3817,6 +3817,21 @@ function wfWaitForSlaves( $wiki = false ) {
 		$dbw = $lb->getConnection( DB_MASTER, array(), $wiki );
 		$pos = $dbw->getMasterPos();
 		$lb->waitForAll( $pos, $wiki );
+
+		// Wikia change - begin
+		// OPS-6313 - sleep-based implementaion for cluster G (it uses consul-powered DB config)
+		if ( $lb->parentInfo()['id'] === 'main-c7' ) {
```
drozdo: could you check if dbhost youre connecting to is in .consul domain or something generic like that? hardcoding cluster name is not the best option IMHO

macbre: Good point, check improved in  7e2943b

macbre:
Backported to the release in 8a53c27

PR #7186, opened 2015-05-12
(CE-1827) Add additional Wikia submitter entity to CE notices

Grunny:
Adds Wikia as the submitter of the CE notice as requested.


lgarczewski:
LGTM.

PR #7187, opened 2015-05-12
SOC-611 | New forum thread email

kvas-damian:
https://wikia-inc.atlassian.net/browse/SOC-611

Most likely we'll need to synchronise with: https://github.com/Wikia/app/pull/7167
It depends which one gets merged first.


extensions/wikia/Email/Controller/ForumController.class.php
```diff
@@ -0,0 +1,109 @@
...
+
+	protected function getButtonLink() {
+		return $this->titleUrl;
+	}
+
+	protected function getDetailsHeader() {
+		return $this->titleText;
+	}
+
+	protected function assertValidParams() {
```
jsutterfield: This is a small thing, but I find it helpful to have functions located close to the places they're called. It makes the code easier to read. Can we move this method up to L29?

extensions/wikia/Email/Controller/ForumController.class.php
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	public function getSubject() {
+		return wfMessage( $this->getSubjectKey(), $this->boardText )
+			->inLanguage( $this->targetLang )
+			->text();
+	}
+
+	protected function getSubjectKey() {
+		return 'emailext-forum-subject';
```
jsutterfield: Is there a reason to have the `getSubjectKey()`, `getDetails()`, `getButtonLink()`, `getDetailsHeader()` methods? All they're doing is just returning the corresponding instance variable. Can we use those instance variable instead inside of the `body()` method?

extensions/wikia/Email/templates/avatarLayout.mustache
```diff
@@ -27,7 +27,10 @@
 						</a>
 					{{/ editorProfilePage}}
 				</p>
-				<p>{{details}}</p>
+				{{# detailsHeader}}
+					<p><strong>{{detailsHeader}}</strong></p>
+				{{/ detailsHeader}}
```
jsutterfield: I'm sure you already know this, but SOC-608 is doing essentially the same thing [here](https://github.com/Wikia/app/pull/7167/files#diff-b467f3bb9d5d158bafee32774e7a8506R32), we're going to have to resolve that conflict.

extensions/wikia/WallNotifications/WallNotifications.class.php
```diff
@@ -387,50 +387,65 @@ protected function sendEmails( array $watchers, WallNotificationEntity $notifica
 			$watcher = $this->getUser( $val );
 			$mode = $watcher->getOption( 'enotifwallthread' );
 
-			if ( !empty( $mode ) && $watcher->getId() != 0 && (
+			if ( !( $watcher->getBoolOption('unsubscribed') === true ) &&
```
jsutterfield: Hmm, looks like there's going to be a lot of conflicts here actually...you guys were working on essentially the same part of code :)

I'd say we could definitely drop this line though, since we check if the user is unsubscribed in email extension.

extensions/wikia/WallNotifications/WallNotifications.class.php
```diff
@@ -387,50 +387,65 @@ protected function sendEmails( array $watchers, WallNotificationEntity $notifica
...
-				}
+					if ( $notification->data->msg_author_username == $notification->data->msg_author_displayname ) {
+						$author_signature = $notification->data->msg_author_username;
+					} else {
+						$author_signature = $notification->data->msg_author_displayname .
+							' (' . $notification->data->msg_author_username . ')';
+					}
+
+					$data = [];
+					wfRunHooks( 'NotificationGetMailNotificationMessage', [
```
jsutterfield: The Forum extension is using this hook, but we're sending the forum emails up above. Do you know if this hook is still needed? If so, we should move it up around L398, otherwise we should probably delete that hook.

extensions/wikia/WallNotifications/WallNotifications.class.php
```diff
@@ -387,50 +387,65 @@ protected function sendEmails( array $watchers, WallNotificationEntity $notifica
...
 				( $mode == WALL_EMAIL_EVERY ) ||
 				( $mode == WALL_EMAIL_SINCEVISITED && empty( $this->uniqueUsers[$entityKey][$watcher->getId()] ) )
 			)) {
 
-				$key = $this->createKeyForMailNotification( $watcher->getId(), $notification );
-				$watcherName = $watcher->getName();
-
-				if ( $notification->data->msg_author_username == $notification->data->msg_author_displayname ) {
-					$author_signature = $notification->data->msg_author_username;
+				$controller = $this->getEmailExtensionController( $notification );
```
jsutterfield: What're your thoughts on updating this method to return a boolean and be something like `isForumPost`? For me when reading this it's not really clear why we're getting the emailExtensionController, like what the intent is. When looking into `getEmailExtensionController()` then it's revealed that we care if this is a forum post or not. I think it may be clearer if we indicate that here, and then we could hardcode the forum controller on L409.

extensions/wikia/WallNotifications/WallNotifications.class.php
```diff
@@ -1051,4 +1066,18 @@ protected function getUser( $userId ) {
 		return $this->cachedUsers[$userId];
 	}
 
+	private function getEmailExtensionController( $notification ) {
```
jsutterfield: Similar to `assertValidParams` inside of ForumController.class.php, what do you think about moving this method up to around L455? It's defined down here but called way higher up on L396.

extensions/wikia/Email/Controller/ForumController.class.php
```diff
@@ -0,0 +1,109 @@
...
+			->inLanguage( $this->targetLang )
+			->parse();
+	}
+
+	protected function getDetails() {
+		return $this->details;
+	}
+
+	protected function getButtonText() {
+		return wfMessage( 'emailext-forum-button-label')
```
garthwebb: Missing space

garthwebb:
I don't have many comments for this PR currently but I'm interested to see how this and the PR for SOC-608 are pulled together.

extensions/wikia/Email/templates/avatarLayout.mustache
kvas-damian: I'm already working on SOC-608, and will merge them together at some point.

extensions/wikia/WallNotifications/WallNotifications.class.php
kvas-damian: We can't drop it until all emails are migrated to new extension: https://github.com/Wikia/app/pull/7167#discussion_r30186204

extensions/wikia/WallNotifications/WallNotifications.class.php
kvas-damian: There is at least one more email that is sent via this hook. @rafalkalinski is working on removing that in SOC-457

extensions/wikia/WallNotifications/WallNotifications.class.php
kvas-damian: This method is prepared to return different controller names to be able to handle different cases. Including SOC-608 and SOC-457.

Do you think we should still refactor this code?

extensions/wikia/Email/Controller/ForumController.class.php
kvas-damian: I wanted to put everything in getters in order to easily override those parts in classes that inherits from this one. I expect SOC-457 using this code.

jsutterfield: Okay, if this is going be subclassed that makes sense. Works for me :+1:

extensions/wikia/WallNotifications/WallNotifications.class.php
jsutterfield: Ah, I see. I thought of this as binary, we either were sending a Forum email or a Wall email, but if we're differentiating those emails even further then having this method makes more sense to me. Leaving as is works for me :+1:

kvas-damian:
While this code is already merged to SOC-608, closing this PR.

PR #7193, opened 2015-05-13
PLATFORM-1219: more logging in create wiki flow

macbre:
- log if `commit()` was actually performed
- log how long the creation process took

@michalroszka / @harnash / @wladekb / @owend


macbre:
Backported to the upcoming release in eba3611

PR #7197, opened 2015-05-13
DAT-2785 and DAT-2784 - Handle InnerXML properly

sebastianmarzjan:
Remove obsolete test


idradm:
LG, have you thought about injection instead?

SebastianMarzjan:
It might be useful for adding an integration test. I'd suggest let's revise this and add them on next iteration.

PR #7200, opened 2015-05-13
fixPageLatest: fix empty articles with page_latest = 0

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-1206

"Fix" articles that have both `page_latest` and `page_len` set to zero by making an empty edit. Revisions are missing (highly likely they've never existed) - "restore"  them by adding empty ones to every affected article.

A script from #876 improved

@wladekb / @harnash / @owend

### DIFF
--- a/maintenance/wikia/fixPageLatest.php
+++ b/maintenance/wikia/fixPageLatest.php
@@ -4,8 +4,10 @@
  * Script that tries to fix page_latest entry in page table
  *
  * @author macbre
- * @file
  * @ingroup Maintenance
+ *
+ * @see BAC-278
+ * @see PLATFORM-1206
  */
 
 require_once( dirname( __FILE__ ) . '/../Maintenance.php' );
@@ -35,6 +37,7 @@ class FixPageLatest extends Maintenance {
 			[
 				'page_id',
 				'page_title',
+				'page_len',
 				#'page_touched',
 				'rev_id'
 			],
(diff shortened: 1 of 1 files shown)


maintenance/wikia/fixPageLatest.php
```diff
@@ -45,44 +74,49 @@ public function execute() {
...
-					['page_id' => $row->page_id],
-					__METHOD__
-				);
-				$this->output(" - page_latest set to {$revId}\n");
-
-				$fixed++;
+			// no revision data - we can generate an empty revision if "page_len" is set to 0 (PLATFORM-1286)
+			if ($revId == 0 &&  $row->page_len == 0 ) {
+				$this->output(" - making an empty edit");
+				$revId = $this->insertEmptyRevision( $row->page_id );
```
harnash: Shouldn't we check if database INSERT was successful? Will it throw exception on failure and if so should we catch it here?

macbre: It will throw an exception. Good point, I'll add a check here.

macbre: Added in e97eb66

macbre:
Reformated the file - please use [this diff](https://github.com/Wikia/app/pull/7200/files?w=1)

harnash:
No more objections. LGTM

PR #7206, opened 2015-05-13
Adding the wikia style guide files to oasis

lizlux:
A best practice we'd like to establish going forward is to create front end components as bower packages and import them into our different apps (i.e. MW and Mercury). This is the first example of this. 

This PR does two things: 

1) Establishes the `/resources/wikia/bower_components/`directory as the place to include all commonly used bower components inside the app repo. See http://bower.io/ for more info on using bower.

2) Imports the style guide (Wikia/style-guide bower component) into the app repo. Note that this PR doesn't actually include any style guide files via resource loader or assets manager. We'll just include files as needed.


dianafa:
is this PR still valid?

lizlux:
Closed this for now. Might re-open when we re-start work on the global nav.

PR #7210, opened 2015-05-13
CONCF-543 check if article exists and only then add it to API response

bognix:
@vforge @rogatty

### DIFF
--- a/extensions/wikia/MercuryApi/models/MercuryApi.class.php
+++ b/extensions/wikia/MercuryApi/models/MercuryApi.class.php
@@ -263,8 +263,12 @@ class MercuryApi {
 		}
 
 		if ( $process ) {
-			foreach ( $data[ $process ] as &$item ) {
-				$item = $this->addArticleUrlToCuratedContentItem( $item );
+			$items = [];
+			foreach ( $data[ $process ] as $item ) {
+				$processedItem = $this->processItem( $item );
+				if (!empty($processedItem)) {
+					$items[] = $processedItem;
+				}
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


rogatty:
LGTM

extensions/wikia/MercuryApi/models/MercuryApi.class.php
```diff
@@ -263,27 +263,37 @@ public function processCuratedContent( $data ) {
...
 	 * Platform Team is OK with hitting DB for MVP (10-15 wikis)
 	 *
 	 * @param $item
 	 * @return mixed
 	 */
-	private function addArticleUrlToCuratedContentItem( $item ) {
+	private function processCuratedContentItem( $item ) {
 		if ( !empty( $item[ 'article_id' ] ) ) {
-			$item[ 'article_local_url' ] = Title::newFromID( $item[ 'article_id' ] )->getLocalURL();
+			$title = Title::newFromID( $item[ 'article_id' ] );
```
hakubo: hmm that is a call to DB for every article,
do we have to check it?
maybe we can handle nulls on the front end?
or maybe we can optimize it?

hakubo:
LGTM

extensions/wikia/MercuryApi/models/MercuryApi.class.php
bognix: there is a comment above to check with platform team when we would go 100% or even for phase 2.

rogatty: Yup, it's stated in the method description that we should fix it.

vforge: We've talked about that with @owend and it's cool for Phase 1

vforge:
LGTM

PR #7215, opened 2015-05-14
Fix for strict error 

kamilkoterba:
Not ticketed. Just fixing what is annoying ;)

'Strict standards: Declaration of AssetsManagerGroupsBuilder::getContent() should be compatible with AssetsManagerBaseBuilder::getContent' 
that was braking assets build because error text was rendered within js assets code

@Wikia/community-engineering, @macbre


macbre:
LGTM

PR #7217, opened 2015-05-14
CONCF-547: Adjust code manipulating page header to recent changes done to page header

inez:
/cc @dianafa @bognix

### DIFF
--- a/extensions/VisualEditor/wikia/modules/ve/ui/widgets/ve.ui.WikiaFocusWidget.js
+++ b/extensions/VisualEditor/wikia/modules/ve/ui/widgets/ve.ui.WikiaFocusWidget.js
@@ -33,7 +33,6 @@ ve.ui.WikiaFocusWidget = function VeUiWikiaFocusWidget( surface ) {
 	this.$body = this.$( this.getElementDocument() ).find( 'body:first' );
 	this.$window = this.$( this.getElementWindow() );
 	this.$surface = surface.$element;
-
 	this.$navBackground = this.$( '.WikiNav .navbackground' );
 	this.$localNavigation = this.$( '#localNavigation' );
 	this.$wikiaAds = this.$( '.hide-to-edit, .hide-for-edit, .wikia-ad, #WikiaAdInContentPlaceHolder' );
--- a/extensions/VisualEditor/wikia/modules/ve/ui/widgets/ve.ui.WikiaFocusWidget.js
+++ b/extensions/VisualEditor/wikia/modules/ve/ui/widgets/ve.ui.WikiaFocusWidget.js
@@ -33,12 +33,9 @@ ve.ui.WikiaFocusWidget = function VeUiWikiaFocusWidget( surface ) {
 	this.$body = this.$( this.getElementDocument() ).find( 'body:first' );
 	this.$window = this.$( this.getElementWindow() );
 	this.$surface = surface.$element;
-	this.$pageHeaderElements = this.$( '#WikiaPageHeader' ).children( ':not( .header-container  )' ).add( '.header-tally' );
 
 	this.$navBackground = this.$( '.WikiNav .navbackground' );
 	this.$localNavigation = this.$( '#localNavigation' );
-	this.$wikiaBarWrapper = this.$( '#WikiaBarWrapper' );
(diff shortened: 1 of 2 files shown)


inez:
Tests are breaking for not related reason - ignore please.

extensions/VisualEditor/wikia/modules/ve/ui/widgets/ve.ui.WikiaFocusWidget.js
```diff
@@ -285,7 +286,8 @@ ve.ui.WikiaFocusWidget.prototype.hideDistractions = function () {
 		mw.config.get( 'WikiaBar' ).hide();
 	}
 	// Visibility property - problem with edit button opening when setting display property
-	this.$pageHeaderElements.css( 'visibility', 'hidden' );
+	//this.$pageHeaderElements.css( 'visibility', 'hidden' );
```
bognix: why this is commented out not removed?

inez: No reason, just forgot

extensions/VisualEditor/wikia/modules/ve/ui/widgets/ve.ui.WikiaFocusWidget.js
```diff
@@ -33,7 +33,8 @@ ve.ui.WikiaFocusWidget = function VeUiWikiaFocusWidget( surface ) {
 	this.$body = this.$( this.getElementDocument() ).find( 'body:first' );
 	this.$window = this.$( this.getElementWindow() );
 	this.$surface = surface.$element;
-	this.$pageHeaderElements = this.$( '#WikiaPageHeader' ).children( ':not( h1 )' );
+	this.$pageHeaderElements = this.$( '#WikiaPageHeader' ).children( ':not( .header-container  )' ).add( '.header-tally' );
```
bognix: why not add some class to WikiaPageHeader when VE is loaded and inside PageHeader.scss style things which should be hidden/visible.
This way all code related with styling PageHeader would be in one place.
Without knowing VE code root cause wasn't obvious.

inez: I don't see a benefit of it, coupling would still remain - as ve would have to know of internals of header (to know which class where to add). Using ve-activated class, as now, at least makes it easy to notice in header code that it styles slightly different for VE.
Ultimately I think it should be handled in a way more manageable way without creating such a strong coupling but I don't have time to handle it now.

bognix: If there is such class added we can add styles inside PageShare.scss and all JS here can be removed.

inez: That class is added for sure, code in this pr depends on it.
I know we didn't use it originally because content in page header wasn't completely predictable, I'm not sure if it changed. (I don't remember details)

dianafa: It doesn't look good, if there is a possibility we should move styling to .css file, especially when we know, that that class is already added.

extensions/VisualEditor/wikia/modules/ve/ui/widgets/ve.ui.WikiaFocusWidget.js
```diff
@@ -306,7 +308,8 @@ ve.ui.WikiaFocusWidget.prototype.showDistractions = function () {
 	if ( this.showWikiaBar ) {
 		mw.config.get( 'WikiaBar' ).show();
 	}
-	this.$pageHeaderElements.css( 'visibility', 'visible' );
+	//this.$pageHeaderElements.css( 'visibility', 'visible' );
+	this.$pageHeaderElements.show();
```
dianafa: the same here

bognix:
ping @inez @vforge @hakubo

hakubo:
LGTM, Thanks everybody :)

dianafa:
LGTM

vforge:
LGTM

inez:
Lgtm

PR #7223, opened 2015-05-18
(CE-1828) Add language code to DMCA submissions

Grunny:
Requested by the team at ChillingEffects.

We have to list out the supported languages manually as CE currently doesn't
provide a list from their API.

Language names are hardcoded rather than translated, as we only
have translations for languages in the native language (i.e. on CreateNewWiki,
interlanguage links, etc.) rather than the names of each language in the current
language. Given this will only be used by ComSup, doesn't seem
necessary to bother translating them in this case.


lgarczewski:
LGTM.

PR #7224, opened 2015-05-18
CreateWiki::waitForSlaves - more logging

macbre:
According to logs we do wait for slaves (which takes ~ 10 seconds), but we still get errors caused by slaves being behind the master.

Log more details when waiting for slaves (including the result of a `commit` method call).

@michalroszka / @harnash / @wladekb


macbre:
Backported to the current release in 04e0fbb

PR #7225, opened 2015-05-18
Fixes for ads in Mercury

gabrys:
Use the $wgEnableAdEngineExt rather than whether the class exists for being more explicit


nandy-andy:
LGTM :+1:

rogatty:
LGTM

PR #7226, opened 2015-05-18
Enable caching of Interwiki::getAllPrefixesDB

macbre:
All calls to `Interwiki::getAllPrefixesDB` were making DB queries. Add a caching here.

Increase `$wgInterwikiExpiry` to a day (from 3 hours).

@drozdo / @michalroszka / @harnash / @wladekb

### DIFF
--- a/includes/interwiki/Interwiki.php
+++ b/includes/interwiki/Interwiki.php
@@ -275,6 +275,7 @@ class Interwiki {
 	 * @since 1.19
 	 */
 	protected static function getAllPrefixesDB( $local ) {
+		global $wgMemc, $wgInterwikiExpiry; // Wikia change
 		$db = wfGetDB( DB_SLAVE );
 
 		$where = array();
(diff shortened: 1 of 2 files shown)


includes/interwiki/Interwiki.php
```diff
@@ -295,6 +307,8 @@ protected static function getAllPrefixesDB( $local ) {
 		foreach ( $res as $row ) {
 			$retval[] = (array)$row;
 		}
+
+		$wgMemc->set( $key, $retval, $wgInterwikiExpiry ); // Wikia change
```
harnash: Shouldn't we also check here if $retval is an Array? Few lines above we do set it to something which is not an Array (else in Memcache get) so if select fails we might get some unmeaningful value. Maybe it would be better not to use $retval as an value for cached data so we don't get it polluted in cache miss scenario?

harnash:
Only one concern from me. Rest: LGTM.

includes/interwiki/Interwiki.php
macbre: `$retval` is initialized as an array few lines above :)

harnash: Oh. Right! This code was collapsed and I didn't noticed. Then it is OK the way it is.

macbre: Ok, merging :)

PR #7232, opened 2015-05-18
DAT-2813 Treat '0' like a non-empty string

dianafa:
https://wikia-inc.atlassian.net/browse/DAT-2813

### DIFF
--- a/extensions/wikia/PortableInfobox/services/Parser/Nodes/Node.php
+++ b/extensions/wikia/PortableInfobox/services/Parser/Nodes/Node.php
@@ -46,8 +46,10 @@ class Node {
 		return [ 'value' => (string)$this->xmlNode ];
 	}
 
+	//note that a '0' value cannot be treated like a null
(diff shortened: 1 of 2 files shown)


extensions/wikia/PortableInfobox/services/Parser/Nodes/Node.php
```diff
@@ -46,8 +46,10 @@ public function getData() {
 		return [ 'value' => (string)$this->xmlNode ];
 	}
 
+	//note that a '0' value cannot be treated like a null
 	public function isEmpty( $data ) {
-		return !( isset( $data[ 'value' ] ) ) || empty( $data[ 'value' ] );
+		$value = $data[ 'value' ];
+		return !( isset( $value ) ) || (empty( $value ) && $value != '0');
```
RafLeszczynski: LGTM but before we merge I'd suggest creating unit tests for this method :-)

RafLeszczynski:
LGTM

PR #7233, opened 2015-05-18
Dat 2814 - portable infobox fix styling issues on monobook

RafLeszczynski:
- reset margin for figure element 
- standardise background color values for monobook


jacekjursza:
LGTM

PR #7237, opened 2015-05-18
SOC-781 Don't send emails for edited or deleted article comments

mklucsarits:
https://wikia-inc.atlassian.net/browse/SOC-781

@garthwebb @jsutterfield

### DIFF
--- a/includes/EmailNotification.php
+++ b/includes/EmailNotification.php
@@ -168,6 +168,11 @@ class EmailNotification {
 	 * @return bool
 	 */
 	private function shouldSendEmail( array $watchers ) {
+		if ( $this->isArticleComment() && $this->previousRevId !== 0 ) {
+			// Don't send email when an article comment is edited or deleted
+			return false;
+		}
+
 		if ( $this->thereAreUsersWatchingPage( $watchers ) ) {
 			return true;
 		}
--- a/includes/EmailNotification.php
+++ b/includes/EmailNotification.php
@@ -168,8 +168,8 @@ class EmailNotification {
 	 * @return bool
 	 */
 	private function shouldSendEmail( array $watchers ) {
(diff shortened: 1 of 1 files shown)


garthwebb:
LGTM

jsutterfield:
LGTM

includes/EmailNotification.php
```diff
@@ -191,24 +195,14 @@ private function thereAreUsersWatchingPage( array $watchers ) {
 	 * @return bool
 	 */
 	private function canSendUserTalkEmail() {
-		$targetUser = User::newFromName( $this->title->getText() );
```
garthwebb: This was moved down so the lightweight global check on ```EnotifUserTalk``` could happen first.

includes/EmailNotification.php
```diff
@@ -191,24 +195,14 @@ private function thereAreUsersWatchingPage( array $watchers ) {
...
-
 		// Should we notify users when their user talk page is changed?
 		if ( empty( F::app()->wg->EnotifUserTalk ) ) {
 			return false;
 		}
 
-		// Is it a minor edit? If so, do we want to notify users about minor edits?
```
garthwebb: All minor edit checks were grouped together and moved down a bit.

includes/EmailNotification.php
```diff
@@ -233,11 +245,6 @@ private function canSendUserTalkEmail() {
 			return false;
 		}
 
-		// Is it a minor edit? If so, does that user want to know about minor edits?
-		if ( $this->isMinorEdit() &&  $targetUser->getOption( 'enotifminoredits' ) )  {
```
garthwebb: Notice this logic changed slightly when it moved up above.  The logic on this line is incorrect because it says "If this is a minor edit AND this user wants to get minor edits THEN don't send an email."

jsutterfield:
LGTM

PR #7238, opened 2015-05-18
SOC-815 Added cache busting and cleaned up the function slightly.

garthwebb:
This PR cache busts the notification list created and stored in MemcacheSync.

https://wikia-inc.atlassian.net/browse/SOC-815


extensions/wikia/WallNotifications/WallNotifications.class.php
```diff
@@ -30,70 +30,75 @@ public function __construct() {
 	/*
 	 * Public Interface
 	 */
-	public function getWikiNotifications( $userId, $wikiId, $readSlice = 5, $countonly = false, $notifyeveryone = false ) {
-		/* if since == null, get all notifications */
-		/* possibly ignore $wiki at one point and fetch notifications from all wikis */
-		wfProfileIn(__METHOD__);
-
```
garthwebb: This first chunk of code in red was moved into its own function, ```getWikiNotificationList```.

extensions/wikia/WallNotifications/WallNotifications.class.php
```diff
@@ -30,70 +30,75 @@ public function __construct() {
...
+			return $list;
 		}
+
 		$read = [];
 		$unread = [];
+		$cacheCleared = false;
 
-		// walk through list of ids
-		// $listval is unique_id field from wall_notification table in DB
-		foreach ( array_reverse( $list['notification'] ) as $listval ) {
```
garthwebb: A few changes here to reduce the deep ```if/else``` nesting.

extensions/wikia/WallNotifications/WallNotifications.class.php
```diff
@@ -30,70 +30,75 @@ public function __construct() {
...
+			$grouped = [];
+			if ( !$countOnly ) {
+				$grouped = $this->groupEntity( $notifyRelation['list'] );
+
+				if ( empty( $grouped ) ) {
+					continue;
 				}
+			}
 
-				if ( !empty( $grouped ) || $countonly ) {
```
garthwebb: The *above* if statement was repeated logic (we already checked ```$countOnly``` once).  Moved the case we care about (no ```$grouped``` even after calling ```groupEntity```) up to line 68 and reduced the nest level by one.

extensions/wikia/WallNotifications/WallNotifications.class.php
```diff
@@ -30,70 +30,75 @@ public function __construct() {
...
+									: $notifyRelation['count']
+					];
+				} elseif ( $readSlice > 0 ) {
+					// so we have more read notifications that we need for display
+					// remove them
+					$this->remNotificationsForUniqueID( $userId, $wikiId, $notifyUniqueId );
+
+					// After removing the notifications above, make sure we don't still
+					// have them in this cache and thus try to remove them again. SOC-815
+					if ( !$cacheCleared ) {
```
garthwebb: This is the only real change to this function.  If we haven't already cleared the cache, do it now.  This should keep subsequent calls from getting stale ```notifyUniqueId```'s that were removed from the DB in previous calls.

kvas-damian:
Please resolve merge conflicts.

extensions/wikia/WallNotifications/WallNotifications.class.php
```diff
@@ -30,70 +30,75 @@ public function __construct() {
...
+	 * @param int $readSlice Max number of already read messages to show
+	 * @param bool $countOnly Set to true to only return a count.  This should be split out to another function
+	 * @param bool $notifyEveryone
+	 *
+	 * @return array|Mixed
+	 */
+	public function getWikiNotifications( $userId, $wikiId, $readSlice = 5, $countOnly = false, $notifyEveryone = false ) {
 
+		$list = $this->getWikiNotificationList( $userId, $wikiId );
 		if ( empty( $list ) ) {
```
mklucsarits: It seems superfluous to check empty($list) both here and in getWikiNotificationList. Why not just "return $list" in getWikiNotificationList and leave the "return []" here?

extensions/wikia/WallNotifications/WallNotifications.class.php
kvas-damian: @garthwebb Can you explain us more thoroughly how this change is going to impact on this code?
We're removing notifications synchronously in a loop, so I'm not sure if it's possible to call delete for the same notification many times. Was that caused by concurrent requests that were removing notifications for that user?

garthwebb: I'm pretty confident this is caused by concurrent requests.

extensions/wikia/WallNotifications/WallNotifications.class.php
garthwebb: @mklucsarits, updated.

kvas-damian:
LGTM

mklucsarits:
LGTM

PR #7245, opened 2015-05-19
SOC-769 remove DISTINCT from query

jsutterfield:
Remove DISTINCT from a query used to find the last admin to edit a wiki.

The problem here had to do with combining DISTINCT and ORDER BY. The exact query in question looks like this

``` SQL
SELECT DISTINCT rev_user
FROM `revision`
WHERE (rev_user IN ('5654073','5654074') )
AND (rev_timestamp > '20150320042619')
ORDER BY rev_timestamp DESC
LIMIT 1

+----------+
| rev_user |
+----------+
|  5654073 |
+----------+
```

What we're trying to ask is "Give me the last admin to make an edit in the last 2 months". What we're really asking is "Pick all the distinct admins who made an edit in the last 2 months, doesn't matter which edit, just one in the last 2 months. After that, sort those results by their time stamps and give me the latest." Because of the way DISTINCT works, MySQL is free to pick any row in the last 2 months before ordering which is resulting in some undesired return values.

As further illustration, if we modify that query above to show the results before the LIMIT is applied we get the following (Note: these are results from jsutterfield wikia on dev):

``` SQL
SELECT DISTINCT rev_user
FROM `revision`
WHERE (rev_user IN ('5654073','5654074') )
AND (rev_timestamp > '20150320042619')
ORDER BY rev_timestamp DESC;

+----------+
| rev_user |
+----------+
|  5654073 |
|  5654074 |
+----------+
```

But if we modify the query again to show us the last edit each admin made in the last 2 months, we can see the order is different:

``` SQL
SELECT DISTINCT rev_user, rev_timestamp
FROM `revision` 
GROUP BY rev_user 
HAVING rev_user IN ('5654073','5654074') 
AND (rev_timestamp > '20150320042619') 
ORDER BY max(rev_timestamp) DESC;

+----------+----------------+
| rev_user | rev_timestamp  |
+----------+----------------+
|  5654074 | 20150519061549 |
|  5654073 | 20150519054549 |
+----------+----------------+
```

Removing `DISTINCT` also fixes the issue, and I think that's the easiest to understand, so that's the approach I took. [Here's](http://www.programmerinterview.com/index.php/database-sql/sql-select-distinct-and-order-by/) a nice article with more info if you're interested.

Ticket: https://wikia-inc.atlassian.net/browse/SOC-769

### DIFF
--- a/extensions/wikia/HAWelcome/HAWelcomeTask.php
+++ b/extensions/wikia/HAWelcome/HAWelcomeTask.php
@@ -220,7 +220,6 @@ class HAWelcomeTask extends BaseTask {
 					__METHOD__,
 					array(
 						'ORDER BY' => 'rev_timestamp DESC',
-						'DISTINCT'
 					)
 				);
 
--- a/extensions/wikia/HAWelcome/HAWelcomeTask.php
+++ b/extensions/wikia/HAWelcome/HAWelcomeTask.php
@@ -219,7 +219,7 @@ class HAWelcomeTask extends BaseTask {
 					),
 					__METHOD__,
 					array(
-						'ORDER BY' => 'rev_timestamp DESC',
+						'ORDER BY' => 'rev_timestamp DESC'
 					)
 				);
 


extensions/wikia/HAWelcome/HAWelcomeTask.php
```diff
@@ -220,7 +220,6 @@ public function setSender() {
 					__METHOD__,
 					array(
 						'ORDER BY' => 'rev_timestamp DESC',
```
kvas-damian: Comma is not required here now ;)

kvas-damian:
LGTM

mklucsarits:
LGTM

extensions/wikia/HAWelcome/HAWelcomeTask.php
jsutterfield: bah! so close! :)

PR #7246, opened 2015-05-19
CE-1841: Remove finish starter

lukaszkonieczny:
Quick fix for https://wikia-inc.atlassian.net/browse/CE-1841

Will be investigate in https://wikia-inc.atlassian.net/browse/CE-1842


Grunny:
:+1: :shipit:

PR #7248, opened 2015-05-19
DAT-2796 PHP Warning with foreach() fix

dianafa:
https://wikia-inc.atlassian.net/browse/DAT-2796

The only change is the `if (is_array($set['data']['value']))` check. Github doesn't show it nice.


SebastianMarzjan:
https://github.com/Wikia/app/pull/7248/files?w=1 ;-)

SebastianMarzjan:
Looks good. What is the case in which the set['data']['value'] is empty, one-sided comparison? Empty comparson?

idradm:
This warning will occur only if isEmpty functionality for comparison is broken, maybe we should fix it there instead?

PR #7249, opened 2015-05-19
CE-1771 Render flags when action=submit after an edit

adamkarminski:
We cannot use `WikiaPageType::isEditPage` because it treats `action=submit` URLs as edit pages. This caused a bug with flags not being rendered when redirected to article view after an edit.

@Wikia/community-engineering


Grunny:
LGTM.

PR #7251, opened 2015-05-19
DAT-2824 Fix warnings when node has no child nodes

sebastianmarzjan:


extensions/wikia/PortableInfobox/services/Helpers/SimpleXmlUtil.php
```diff
@@ -28,8 +28,12 @@ public static function getInstance() {
 	public function getInnerXML( $node ) {
 		$innerXML = '';
 		if ( $node instanceof \SimpleXMLElement ) {
-			foreach ( dom_import_simplexml( $node )->childNodes as $child ) {
-				$innerXML .= $child->ownerDocument->saveXML( $child );
+			$domDocument = dom_import_simplexml( $node );
+
+			if ( $domDocument->childNodes->length ) {
```
idradm: Maybe we could use http://php.net/manual/pl/domnode.haschildnodes.php here?

SebastianMarzjan: I think we can and we will :)
Thanks!

SebastianMarzjan: On second though, we'd need to have a two-sided conditional here
( if $domDocument instanceof DomNode && $domDocument->hasChildNodes() )

What do you think?

idradm:
I'm not fond of using instanceof. Other than that looks good.

PR #7259, opened 2015-05-19
SOC-797 Clearing encrypted access_token cookie on logout

bkoval:
Mercury's backend for the New Login Flow is dependent on hapi-auth-cookie, which authenticates all the routes and controls the session. It has one downside for us - it's encrypting the cookie and this feature cannot be switched off. 

To work in compliance with Mediawiki Login, we need to set the unencrypted access_token cookie anyway, which means we end up setting it two times (one is encrypted, the second one is not). We will get rid of the hapi-auth-cookie support here: https://wikia-inc.atlassian.net/browse/SOC-798 , however, this might take a bit of time on implementation and considering the business value of releasing the New Login as soon as possible we should end up with the safest workaround until the SOC-798 is done.

Since we're using Mediawiki's logout so far, it's just enough to clear the 'sid' cookie (the encrypted one) once the user is logging out, since access_token is already being cleared.

### DIFF
--- a/extensions/wikia/Helios/User.class.php
+++ b/extensions/wikia/Helios/User.class.php
@@ -196,6 +196,22 @@ class User {
 			time() - self::ACCESS_TOKEN_COOKIE_TTL,
 			\WebResponse::NO_COOKIE_PREFIX
 		);
+		self::clearEncryptedAccessTokenCookie($response);
+	}
+
+	/**
+	 * Mercury's backend (Hapi) is setting access_token cookie in an encrypted form, so we need
+	 * to destroy this one as well on UserLogout
+	 * This is a temporary change which will be deleted while implementing SOC-798
+	 * @param $response
+	 */
+	public static function clearEncryptedAccessTokenCookie($response) {
+		$response->setcookie(
+			'sid',
+			'',
+			time() - self::ACCESS_TOKEN_COOKIE_TTL,
+			\WebResponse::NO_COOKIE_PREFIX
+		);
 	}
 
 	/**
--- a/extensions/wikia/Helios/User.class.php
+++ b/extensions/wikia/Helios/User.class.php
@@ -196,7 +196,7 @@ class User {
 			time() - self::ACCESS_TOKEN_COOKIE_TTL,
 			\WebResponse::NO_COOKIE_PREFIX
 		);
-		self::clearEncryptedAccessTokenCookie($response);
+		self::clearEncryptedAccessTokenCookie( $response );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


extensions/wikia/Helios/User.class.php
```diff
@@ -196,6 +196,22 @@ public static function clearAccessTokenCookie() {
 			time() - self::ACCESS_TOKEN_COOKIE_TTL,
 			\WebResponse::NO_COOKIE_PREFIX
 		);
+		self::clearEncryptedAccessTokenCookie($response);
```
mklucsarits: Needs spaces inside parentheses.

bkoval:
I've just rerun Jenkins (it was failing for an unrelated test) and it passed: http://qa-s2.wikia-prod:8080/job/unitTests_php_branches/15109/

extensions/wikia/Helios/User.class.php
```diff
@@ -196,6 +196,22 @@ public static function clearAccessTokenCookie() {
...
+		self::clearEncryptedAccessTokenCookie( $response );
+	}
+
+	/**
+	 * Mercury's backend (Hapi) is setting access_token cookie in an encrypted form, so we need
+	 * to destroy this one as well on UserLogout
+	 * This is a temporary change which will be deleted while implementing SOC-798
+	 * @param $response
+	 */
+	public static function clearEncryptedAccessTokenCookie($response) {
```
mklucsarits: Spaces also needed here :x

ArturKlajnerok:
Considering that we will remove the additional cookie soon, LGTM :+1:

mklucsarits:
LGTM

extensions/wikia/Helios/User.class.php
```diff
@@ -196,6 +196,22 @@ public static function clearAccessTokenCookie() {
...
+
+	/**
+	 * Mercury's backend (Hapi) is setting access_token cookie in an encrypted form, so we need
+	 * to destroy this one as well on UserLogout
+	 * This is a temporary change which will be deleted while implementing SOC-798
+	 * @param $response
+	 */
+	public static function clearEncryptedAccessTokenCookie( $response ) {
+		$response->setcookie(
+			'sid',
```
kvas-damian: Please move this string to constant value.
I see almost the same call few lines above, did you consider extracting it to separate method and call with different parameters?

extensions/wikia/Helios/User.class.php
```diff
@@ -196,6 +196,22 @@ public static function clearAccessTokenCookie() {
 			time() - self::ACCESS_TOKEN_COOKIE_TTL,
```
kvas-damian: @garthwebb do you remember why we use ```time() - self::ACCESS_TOKEN_COOKIE_TTL``` instead of simply ```-1``` to expire immediately?

PR #7260, opened 2015-05-19
Monetization (MON-393): SitemapPage - add controller and model to show list of wikis

saipetch:
Monetization (MON-393):
- add functions to get list of wikis and total number of wikis
- add hook to render sitemap page
- add pagination

### DIFF
--- a/extensions/wikia/SitemapPage/SitemapPageModel.class.php
+++ b/extensions/wikia/SitemapPage/SitemapPageModel.class.php
@@ -21,9 +21,12 @@ class SitemapPageModel extends WikiaModel {
 
 	protected $maxDate;	// maximum date
 
-	public function __construct() {
-		parent::__construct();
-		$this->maxDate = date( 'Ymd' );
+	/**
+	 * Set maximum date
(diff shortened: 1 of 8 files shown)


extensions/wikia/SitemapPage/SitemapPageModel.class.php
```diff
@@ -0,0 +1,258 @@
...
+	];
+
+	/**
+	 * Get limit per page
+	 * @param int $level
+	 * @return int
+	 */
+	public function getLimitPerPage( $level = 0 ) {
+		$base = $this->getLimitPerPageBase();
+		return ( $level == 1 ) ? pow( $base, self::MAX_LEVEL - 1 ) : $base;
```
jogura: What makes `$level == 1` have such a big difference in limit per page? Documentation could be helpful.

jogura:
You may want to comment how the `from` and `to` are computed and sorted at each level. It was not obvious until you explained to me what it was doing.

jogura:
Other than adding additional comments, LGTM

PR #7261, opened 2015-05-19
DAT-2796 tests for isEmpty()

dianafa:
more details here: https://wikia-inc.atlassian.net/browse/DAT-2796

ping @SebastianMarzjan @idradm


idradm:
:shipit: then

PR #7262, opened 2015-05-19
SOC-742 Don't track sign-up events for mirrored requests

mklucsarits:
https://wikia-inc.atlassian.net/browse/SOC-742

@jsutterfield @kvas-damian

### DIFF
--- a/extensions/wikia/Track/Track.php
+++ b/extensions/wikia/Track/Track.php
@@ -81,6 +81,10 @@ SCRIPT1;
 	}
 
 	public static function event ($event_type, $param=null) {
+		if ( !self::shouldTrackEvents() ) {
+			return false;
+		}
+
 		wfProfileIn(__METHOD__);
 
 		$backtrace = debug_backtrace();
@@ -97,6 +101,23 @@ SCRIPT1;
 			wfProfileOut(__METHOD__);
 			return false;
 		}
+	}
+
+	/**
+	 * Check whether events should be tracked
+	 *
+	 * @return bool
+	 */
+	protected static function shouldTrackEvents() {
+		// Check request headers to make sure this is not a mirrored
+		// request, which could result in duplicate tracking events
+		$headers = getallheaders();
+		foreach ( $headers as $name => $value ) {
+			if ( strtolower( $name ) === 'x-mirrored-request' ) {
+				return false;
+			}
+		}
+		return true;
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


extensions/wikia/UserLogin/UserSignupSpecialController.class.php
```diff
@@ -690,8 +690,18 @@ protected function shouldDisableCaptcha() {
 	 * Track an event with a given label with category 'user-sign-up' and action 'request'
 	 *
 	 * @param string $label
+	 * @return void
 	 */
 	protected function track( $label ) {
+		// Check request headers to make sure this is not a mirrored
+		// request, which could result in duplicate tracking events
+		$headers = getallheaders();
```
kvas-damian: What do you think about moving this logic into ```\Track::event```? This would avoid problems like this in future.

Also I'm thinking about extracting those lines into new method.

mklucsarits: Yep, both good ideas. Will do.

extensions/wikia/Track/Track.php
```diff
@@ -100,6 +104,23 @@ public static function event ($event_type, $param=null) {
...
+	 * Check whether events should be tracked
+	 *
+	 * @return bool
+	 */
+	protected static function shouldTrackEvents() {
+		// Check request headers to make sure this is not a mirrored
+		// request, which could result in duplicate tracking events
+		$headers = getallheaders();
+		foreach ( $headers as $name => $value ) {
+			if ( strtolower( $name ) === 'x-mirrored-request' ) {
```
jsutterfield: Can we make 'x-mirrored-request' a class constant?

extensions/wikia/Track/Track.php
```diff
@@ -100,6 +104,23 @@ public static function event ($event_type, $param=null) {
...
 
 	/**
+	 * Check whether events should be tracked
+	 *
+	 * @return bool
+	 */
+	protected static function shouldTrackEvents() {
+		// Check request headers to make sure this is not a mirrored
+		// request, which could result in duplicate tracking events
+		$headers = getallheaders();
```
jsutterfield: So this way works, but we already have a class which abstracts away getting the headers like this. What do you think about updating this method to use
```php
F::app()->wg->request->getHeader()
```
?

mklucsarits: Makes sense, it's a lot easier.

jsutterfield:
LGTM!

kvas-damian:
LGTM

PR #7276, opened 2015-05-20
Do not transition over all properties

hakubo:


extensions/wikia/GlobalNavigation/styles/GlobalNavigation.scss
```diff
@@ -16,6 +16,7 @@ $start-a-wikia-logo-size: 30px;
 	margin-top: -$global-navigation-height - 1px;
 	position: fixed;
 	transition-duration: .5s;
+	transition-property: background-color;
```
Warkot: `box-shadow` is also animated now.

hakubo: But animating it does not add any value to user.

PR #7279, opened 2015-05-20
CE-1845 Handle POST requests from Flags edit form

adamkarminski:
@kamilkoterba @Grunny

### DIFF
--- a/extensions/wikia/Flags/FlagsHelper.class.php
+++ b/extensions/wikia/Flags/FlagsHelper.class.php
@@ -6,11 +6,14 @@ use Flags\Views\FlagView;
 
 class Helper {
 
+	const FLAGS_INPUT_NAME_PREFIX = 'editFlags';
+	const FLAGS_INPUT_NAME_CHECKBOX = 'checkbox';
+
 	public function compareDataAndGetFlagsToChange( $currentFlags, $postData ) {
 		$flagsToAdd = $flagsToRemove = $flagsToUpdate = [];
 
 		foreach ( $currentFlags as $flagTypeId => $flag ) {
-			$keyCheckbox = "editFlags:{$flagTypeId}:checkbox";
+			$keyCheckbox = $this->composeInputName( $flagTypeId, self::FLAGS_INPUT_NAME_CHECKBOX );
 
 			if ( !isset( $flag['flag_id'] ) && !isset( $postData[$keyCheckbox] ) ) {
 				/**
@@ -75,7 +78,7 @@ class Helper {
 			 * This is a protection from users modifying names of parameters in the DOM.
 			 */
 			foreach ( $paramNames as $paramName ) {
-				$key = "editFlags:{$flagTypeId}:{$paramName}";
+				$key = $this->composeInputName( $flagTypeId, $paramName );
 				if ( isset( $postData[$key] ) ) {
 					$flagFromPost['params'][$paramName] = \Sanitizer::escapeHtmlAllowEntities( $postData[$key] );
 				}
@@ -116,5 +119,9 @@ class Helper {
 				'pageId' => $pageId,
 			]
 		)->getData();
+	}
+
+	private function composeInputName( $flagTypeId, $field ) {
+		return self::FLAGS_INPUT_NAME_PREFIX . ":{$flagTypeId}:" . $field;
 	}
 }
--- a/extensions/wikia/Flags/FlagsHelper.class.php
+++ b/extensions/wikia/Flags/FlagsHelper.class.php
@@ -36,7 +36,9 @@ class Helper {
 			} elseif ( $flag['flag_param_names'] !== null ) {
 				/**
 				 * 4. The flag type HAS an instance on this page and WAS posted
-				 * and HAS parameters - update the parameters of the instance
+				 * and HAS parameters - update the parameters of the instance.
+				 * If the flag type has no parameters defined - there is nothing to update
+				 * because the form only allows you to modify parameters.
 				 */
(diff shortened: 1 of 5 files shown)


extensions/wikia/Flags/controllers/FlagsController.class.php
```diff
@@ -79,24 +83,12 @@ public function getFlagsForPageForEdit() {
...
-		 * 2. Get all flag types for a wikia
-		 */
-		$this->model = new FlagType();
-		$flagTypesForWikia = $this->model->getFlagTypesForWikia( $this->params['wikiId'] );
-
-		/**
-		 * 3. Unite the arrays
+		 * 1. Get all flag types with instances for the page
 		 */
-		$responseData = $flagsForPage + $flagTypesForWikia;
```
kamilkoterba: Can we explain here why we can combine flags with types. Or change naming.

extensions/wikia/Flags/FlagsHelper.class.php
```diff
@@ -6,6 +6,85 @@
 
 class Helper {
 
+	public function compareDataAndGetFlagsToChange( $currentFlags, $postData ) {
```
kamilkoterba: Documentation

extensions/wikia/Flags/controllers/FlagsController.class.php
```diff
@@ -240,4 +284,23 @@ public function removeFlagType() {
...
+		/**
+		 * 2. Get all flag types for a wikia
+		 */
+		$flagTypeModel = new FlagType();
+		$flagTypesForWikia = $flagTypeModel->getFlagTypesForWikia( $wikiId );
+
+		/**
+		 * 3. Return the united arrays
+		 */
+		return $flagsForPage + $flagTypesForWikia;
```
kamilkoterba: Can we explain here why we can combine flags with types. Or change naming.

kamilkoterba:
Despite from "Can we explain here why we can combine flags with types. Or change naming" comment
LGTM

extensions/wikia/Flags/controllers/FlagsController.class.php
```diff
@@ -146,6 +138,57 @@ public function getFlagsForPage() {
...
+
+		/**
+		 * Get the current status to compare
+		 */
+		$currentFlags = $this->getAllFlagTypes( $this->params['wikiId'], $this->params['pageId'] );
+
+		$helper = new Helper();
+		$flagsToChange = $helper->compareDataAndGetFlagsToChange( $currentFlags, $this->params );
+
+		$this->model = new Flag();
```
Grunny: Is there a reason the model is in a class property? Right now every method that uses it creates a new instance anyway.

extensions/wikia/Flags/FlagsHelper.class.php
```diff
@@ -38,4 +130,8 @@ private function sendGetFlagsForPageRequest( $pageId ) {
 			]
 		)->getData();
 	}
+
+	private function composeInputName( $flagTypeId, $field ) {
+		return self::FLAGS_INPUT_NAME_PREFIX . ":{$flagTypeId}:" . $field;
```
Grunny: `$field` can be part of the interpolated string as well.

Grunny:
LGTM.

extensions/wikia/Flags/FlagsHelper.class.php
```diff
@@ -6,6 +6,98 @@
...
+			 * Check only for names of parameters defined in the flag_param_names field.
+			 * This is a protection from users modifying names of parameters in the DOM.
+			 */
+			foreach ( $paramNames as $paramName ) {
+				$key = $this->composeInputName( $flagTypeId, $paramName );
+				if ( isset( $postData[$key] ) ) {
+					/**
+					 * Use a value from the form if it is provided
+					 */
+					$flagFromPost['params'][$paramName] = \Sanitizer::escapeHtmlAllowEntities( $postData[$key] );
```
Grunny: Why is this being escaped here, i.e. where is it going from here, and how will this be output in the end? If these parameters end up being parsed as wikitext, they could end up double escaped.

Grunny:
Left some questions, but overall LGTM.

extensions/wikia/Flags/controllers/FlagsController.class.php
adamkarminski: Yep, I think that at this point we can easily say that storing a model as the Controller's property is pointless. Thanks!

extensions/wikia/Flags/FlagsHelper.class.php
adamkarminski: :+1:

extensions/wikia/Flags/models/FlagParameter.class.php
```diff
@@ -36,6 +36,25 @@ public function createParametersForFlag( $flagId, $flagTypeId, $wikiId, $pageId,
 		return $this->status;
 	}
 
+	public function updateParametersForFlag( $flagId, $params ) {
+		$db = $this->getDatabaseForWrite();
+
+		foreach ( $params as $paramName => $paramValue ) {
+			( new \WikiaSQL )
+				->UPDATE( self::FLAGS_PARAMS_TABLE )
+				->SET( $paramName, $paramValue )
```
Grunny: Shouldn't this be `->SET( 'param_name', $paramValue )`?

adamkarminski: Actually, it should be `->SET( 'param_value', $paramValue )` :D

PR #7288, opened 2015-05-20
MAIN-4597: Update GA call from old method gaTrackPageview to new method guaTrackPageview

inez:
This got broken by this: https://github.com/Wikia/app/commit/2ba326ccf897b810faf69efc192855e565638aac


rogatty:
LGTM
Should we search through the code for other instances of old method names?

inez:
@rogatty absolutely. I'm writing e-mail to original author of the breaking change to do that work.

PR #7289, opened 2015-05-21
PLATFORM-1233: Fixed sending custom dimensions to special account.

harnash:
There is an issue when sending custom dimension for special wiki's. Instead of calling UA API directly we are now using wrapper method which automatically sends data to both trackers if needed.


macbre:
LGTM

PR #7290, opened 2015-05-21
CE-1752: Migration script for flags

lukaszkonieczny:
https://wikia-inc.atlassian.net/browse/CE-1752

Script parses wikitext to fetch data about templates to flags and adds it to database.
Templates are not removed.

Script options:
--csv = path to csv file
--section = number of article section to search. If it's not set, all article is searched
--list = just create log (nothing is added to database)


extensions/wikia/Flags/maintenance/MoveNotices.php
```diff
@@ -0,0 +1,510 @@
...
+		while( ( $data = fgetcsv( $csvFile ) ) !== false ) {
+			$csvData[] = $data;
+		}
+
+		fclose( $csvFile );
+
+		$this->output( "Start processing\n" );
+
+		$flag = new Flag();
+		$flagTypes = new FlagType();
```
adamkarminski: Could you name them `$flagType**Model**` or similar? You're using `$flagTypes` array a few lines below and it may be confusing.

extensions/wikia/Flags/maintenance/MoveNotices.php
```diff
@@ -0,0 +1,465 @@
...
+				$flagType = $this->prepareDataForFlagType( $flagTypeModel, $data );
+
+				if ( !($this->flagId = $this->addFlagType( $flagTypeModel, $flagType ) ) ) {
+					continue;
+				}
+
+				$this->log = "Adding flag type: " . json_encode( $flagType ) ."\n";
+			}
+
+			$this->log .= "Start processing template: $this->templateName \n";
```
adamkarminski: What about creating a separate method, let's say `log()` that would handle the `.=` concatenating? I think it is easy to make a mistake if you have to write `.=` and overwrite the logs by accident. :)

extensions/wikia/Flags/maintenance/MoveNotices.php
```diff
@@ -0,0 +1,465 @@
...
+
+			if ( empty( $rows ) ) {
+				$this->log .= "[WARNING] This template is not used \n";
+				$this->log .= "================================================== \n\n\n";
+				fwrite( $this->logFile, $this->log );
+				$this->output( $this->log );
+				continue;
+			}
+
+			if ( !is_null( $section )  ) {
```
adamkarminski: What do you think about making searching only in the `0` section a default option given the strategy we have? Then, you won't run the script for the whole article by accident.

extensions/wikia/Flags/maintenance/MoveNotices.php
```diff
@@ -0,0 +1,465 @@
...
+					$this->log .= "[WARNING] Found more than one ($size) template $this->templateName on page $pageName\n";
+				}
+
+				if ( $size ) {
+					$this->logTemplatesInfo( $templates );
+
+					$flagsToPages = $this->prepareDataForFlagsToPage( $templates[0]['params'] );
+
+					if ( !$list ) {
+						$flagModel->verifyParamsForAdd( $flagsToPages );
```
adamkarminski: I think that you could just use the Flags API to do the insert. Maybe create a private method wrapping the POST request?

extensions/wikia/Flags/maintenance/MoveNotices.php
```diff
@@ -0,0 +1,465 @@
...
+	/**
+	 * Prepare data to add flags to page
+	 */
+	private function prepareDataForFlagsToPage( $params ) {
+		$flagsToPages = [
+			'wiki_id' => $this->wikiId,
+			'page_id' => $this->pageId,
+			'flags' => [
+				[
+					'flag_type_id' => $this->flagId,
```
adamkarminski: If it's the right variable, can we change its name to avoid confusing `$flagTypeId` and `$flagId`?

extensions/wikia/Flags/maintenance/MoveNotices.php
adamkarminski: A second idea - since we output everything that is logged, maybe do both actions - `$this->log .= ` and `$this->output()` in this method?

adamkarminski:
I left a few comments. I think that when they are fixed the script will Look Good Enough For Me for the MVP usage. :)

adamkarminski:
Apply minor code styling fixes and it will Look Good To Me :)

PR #7296, opened 2015-05-21
PLATFORM-1232 force a FromSession user in Piggyback sessions

ghost:
When an access token is provided, the user object for the request
is created based on that. It makes Piggyback useless as Piggyback
relies on session data accessed later during the execution.

We can distinguish a Piggyback session from a regular one by
a simple check and override the user object created from an
access token with one created from session data.

@wladekb @macbre

### DIFF
--- a/includes/context/RequestContext.php
+++ b/includes/context/RequestContext.php
@@ -213,7 +213,7 @@ class RequestContext implements IContextSource {
 		}
 
 		// Replace the user object according to the context, e.g. Piggyback.
-		wfRunHooks( 'RequestContextOverrideUser', [ &$this->user ] );
+		wfRunHooks( 'RequestContextOverrideUser', [ &$this->user, $this->getRequest() ] );
 		return $this->user;
 	}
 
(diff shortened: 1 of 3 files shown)


includes/context/RequestContext.php
```diff
@@ -206,8 +206,9 @@ public function getUser() {
 		if ( $this->user === null && $wgUserForceAnon ) {
 			$this->user = new User();
 		}
+		// replace the FromToken user with a FromSession user when in a Piggyback session
+		if ( $this->user === null || PBLoginForm::isPiggyback() ) {
```
wladekb: is the PBLoginForm class always available when this function can be called?
AFAIR it's our extension that can be at least theoretically disabled, I'd suggest putting
a check if that class exists

michalroszka: Based on the same input from @grunny and my own second thoughts (great timing), I have decided to make the core change more general and move the Piggyback-specific logic to Piggyback. :-)

Grunny:
LGTM :shipit:

Grunny:
LGTM... again. :)

PR #7297, opened 2015-05-21
ADEN-1960 Improve playwire placement

nandy-andy:
We want to introduce a new "in content" slot for videos. We call it `INCONTENT_PLAYER`. The changes include:
- new AMD module to dynamically insert new slot and fill it,
- new `wgIncontentPlayerCountries` InstantGlobal variable.


extensions/wikia/AdEngine/js/run/desktop.run.js
```diff
@@ -14,7 +14,8 @@ require([
 	'ext.wikia.adEngine.slotTracker',
 	'ext.wikia.adEngine.slotTweaker',
 	'wikia.krux',
-	'wikia.window'
+	'wikia.window',
+	'ext.wikia.adEngine.adInContentPlayer'
```
nandy-andy: Actually, if I add `wgEnableIncontentPlayerSlot` I could load it on-demand in Oasis by using our hook. Since `wgIncontentPlayerCountries` is an `InstantGlobal` I don't think it's a good idea to go with a condition:
```php
if( !empty($wgIncontentPlayerCountries) ) {
// add asset group
}
```

fraszczakszymon:
LGTM :+1:

PR #7305, opened 2015-05-22
SOC-812 Email update: article comment 'view all comment' link is broken

rafalkalinski:
https://wikia-inc.atlassian.net/browse/SOC-812


kvas-damian:
LGTM

extensions/wikia/Email/Controller/CommentController.class.php
```diff
@@ -9,7 +9,7 @@
 abstract class CommentController extends EmailController {
 
 	/** @var \Title */
-	protected $title;
+	protected $pageTtitle;
```
garthwebb: This variable is spelled incorrectly (two t's).

extensions/wikia/Email/Controller/CommentController.class.php
```diff
@@ -78,17 +78,17 @@ public function body() {
 			'editorUserName' => $this->getCurrentUserName(),
 			'editorAvatarURL' => $this->getCurrentAvatarURL(),
 			'summary' => $this->getSummary(),
-			'details' => $this->getDetails(),
```
garthwebb: Ok, it took me a long time to figure out *this* was the change that addressed the bug for SOC-812.  It seems that calling ```getDetails``` has a side effect that screws up creating the link in ```getCommentSectionLink```.  My recommendation would be:

1) Document in a comment that the 'details' line is position dependent so others know not to touch it
2) Realize this is a terrible state of things
3) Erase the comment, move the line back up and find the root cause of the problem ;)

I found the issue, details below...

kvas-damian: Actually this line change has nothing to do right now.
It was a problem when we had variable called ```title``` because there is some strange code that is modifying title parameter in a controller. After we renamed it everything should work ok

extensions/wikia/Email/Controller/CommentController.class.php
```diff
@@ -104,7 +104,7 @@ protected function getSalutation() {
 	}
 
 	protected function getSummary() {
-		$articleTitle = $this->title->getText();
+		$articleTitle = $this->pageTtitle->getText();
 
 		return wfMessage( $this->getSummaryKey(), $articleTitle )
 			->inLanguage( $this->targetLang )
```
garthwebb: Ok, I can't comment on this line because its not in the diff, but a few lines below on L117 we call ```newFromTitle```.  This ends up calling (via hooks) ```ArticleCommentList::ArticleFromTitle```.  This sets the fragment ```#comm-XXX``` and _somehow_ this persists and is applied to the next Title object created.  I'm not positive but I think its because it issues a redirect for some strange reason.

The solution is to not use ```newFromTitle``` which requires this request context.  I think this context is what screws things up.  We can also easily get the article via ID like this:
```php
$article = \Article::newFromID( $this->commentTitle->getArticleID() );
```

rafalkalinski: Like Damian said above - the reason of the bug was the $title property name, and now it is fixed.
Since this line was not causing the problem, I'm gonna leave it like it is.

PR #7307, opened 2015-05-22
Dat 2831

jacekjursza:


idradm:
can't we just ignore empty check in set nodes, and then correct isEmpty to handle such use case?

jacekjursza:
@idradm it's not so simple, since we need to change behavior of data nodes, not set nodes themselves. So we need information about parent node inside data nodes (and preferably in all other nodes too). Also, this solution limits 'ignore' behavior only to data tags inside comparison tag (which allows in future use <set> tags in different context).

extensions/wikia/PortableInfobox/services/Parser/Nodes/Node.php
```diff
@@ -39,7 +58,7 @@ public function setExternalParser( ExternalParser $externalParser ) {
 	}
 
 	public function getType() {
-		return $this->xmlNode->getName();
+		return strtolower( $this->xmlNode->getName() );
```
SebastianMarzjan: This could use a comment on why strtolower is needed

SebastianMarzjan:
One small request for comment, other than that LGTM

SebastianMarzjan:
:+1:

PR #7309, opened 2015-05-22
ADEN-2049 Make Exitstitial.js only requested when Ad Engine ext is on

gabrys:


extensions/wikia/AdEngine/js/slot/exitstitial.js
```diff
@@ -47,23 +57,28 @@ require([
...
-				// Show ads
+				// Once the ad loads, show the modal
 				$modalAd.html($('<div></div>').html($ad));
-				window.adslots2.push(adSlot);
+				win.adslots2.push({
+					slotName: adSlot,
+					onError: skipAd,
+					onSuccess: showModal
+				});
+				setTimeout(showModal, 1000); // Show modal after a second even if the ad still loads
```
nandy-andy: You can move the comment to the line above so it's consistent with another `setTimeout()` and its comment below :japanese_ogre:

nandy-andy:
LGTM :+1:

PR #7311, opened 2015-05-22
(INT-82) Add isMainPage data to getArticle response

Grunny:
Add isMainPage data to getArticle response to reliably check if the
current article is the main page in Mercury.

/cc @inez @rogatty


rogatty:
LGTM

PR #7312, opened 2015-05-22
CE-1777 Process flag_params_names…

kamilkoterba:
…to be sure we show empty fields on modal as well
https://wikia-inc.atlassian.net/browse/CE-1777

@Wikia/community-engineering


kamilkoterba:
LGTM

PR #7313, opened 2015-05-22
Update DefaultSettings.php

paladox:
When will Wikia use a newer mediawiki version since mediawiki 1.19 gets discontinued this month. and wikia should use latest mediawiki to get latest features.


onei:
Unfortunately, that's not how an update works at all. It requires updating various code, and, in wikia's case, testing all their customisations and extensions.

wladekb:
@paladox: @onei is right, it's proved to be a very long process in the past taking a few months given a number of our custom features and modifications. AFAIK there's no scheduled time for the next MediaWiki upgrade.

PR #7316, opened 2015-05-25
ADEN-1684 Block BTF slots on premium campaign

fraszczakszymon:
Delay BTF slots load when wgAdDriverDelayBelowTheFold variable is enabled and after ATF slots are loaded, decide if BTF part should be flushed (based on window.ads.runtime.disableBtf variable).

### DIFF
--- a/extensions/wikia/AdEngine/js/provider/directGpt.js
+++ b/extensions/wikia/AdEngine/js/provider/directGpt.js
@@ -130,13 +130,17 @@ define('ext.wikia.adEngine.provider.directGpt', [
 	}
 
 	function fillInSlotWithDelay(slotName, success, hop) {
-		if (context.opts.delayBtf && (!gptFlushed || pendingSlots.length > 0)) {
-			delayBtfSlot(slotName, success, hop);
-		} else if (window.ads.runtime.disableBtf && !gptConfig[slotName]) {
-			blockBtfSlot(slotName, success);
-		} else {
-			provider.fillInSlot(slotName, success, hop);
+		if (context.opts.delayBtf) {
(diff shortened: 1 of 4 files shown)


includes/wikia/DefaultSettings.php
```diff
@@ -1279,6 +1279,12 @@
 $wgEnableAdEngineExt = true;
 
 /**
+ * @name $wgAdDelayBTF
+ * Prevents from loading BTF before ATF ad slots
+ */
+$wgAdDelayBTF;
```
nandy-andy: Do we want it to be by default `null`?

extensions/wikia/AdEngine/js/provider/directGpt.js
```diff
@@ -54,30 +58,88 @@ define('ext.wikia.adEngine.provider.directGpt', [
...
-					gptFlushed = true; // Setting the module-scope var here
+					log(['shouldFlush', slotName, gptFlushed]);
+					return gptFlushed;
 				}
+			}
+		);
 
-				log(['shouldFlush', slotName, gptFlushed]);
-				return gptFlushed;
+	function removePendingSlot(slotName) {
```
nandy-andy: I'd suggest renaming this function or removing the condition connected to pushing delayed queue and placing it within a new function :japanese_ogre:

nandy-andy:
Looks good. I added two minor notes and I wonder if adding some unit tests wouldn't be beneficial now since we added more logic to the file. Let's talk about it offline and proceed with the task.

nandy-andy:
LGTM :+1:

nandy-andy:
Failed unit tests aren't connected to the changes :crying_cat_face:

PR #7317, opened 2015-05-25
CE-1857 View and API controllers separation

adamkarminski:
@Wikia/community-engineering

### DIFF
--- a/extensions/wikia/Flags/Flags.hooks.php
+++ b/extensions/wikia/Flags/Flags.hooks.php
@@ -74,7 +74,7 @@ class Hooks {
 			&& $helper->shouldDisplayFlags()
 			&& !( $wgRequest->getVal( 'action' ) == 'visualeditor' )
 		) {
-			$addText = $helper->getFlagsForPageWikitext( $parser->getTitle()->getArticleID() );
+			$addText = ( new \FlagsController )->getFlagsForPageWikitext( $parser->getTitle()->getArticleID() );
 
 			if ( $addText !== null ) {
 				$mwf = \MagicWord::get( 'flags' );
--- a/extensions/wikia/Flags/controllers/FlagsApiController.class.php
+++ b/extensions/wikia/Flags/controllers/FlagsApiController.class.php
@@ -1,0 +1,312 @@
+<?php
+
+/**
+ * The public interface of the extension and the main entry point for all requests.
+ * It provides a set of CRUD methods to manipulate Flags instances and their types.
+ *
+ * @author Adam Karmiński <adamk@wikia-inc.com>
+ * @copyright (c) 2015 Wikia, Inc.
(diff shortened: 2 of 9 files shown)


extensions/wikia/Flags/Flags.hooks.php
```diff
@@ -74,7 +74,7 @@ public static function onParserBeforeInternalParse( \Parser $parser, &$text, &$s
 			&& $helper->shouldDisplayFlags()
 			&& !( $wgRequest->getVal( 'action' ) == 'visualeditor' )
 		) {
-			$addText = $helper->getFlagsForPageWikitext( $parser->getTitle()->getArticleID() );
+			$addText = ( new \FlagsController )->getFlagsForPageWikitext( $parser->getTitle()->getArticleID() );
```
kamilkoterba: Why FlagsController is not namespaced?

adamkarminski: To be accessible via public api

kamilkoterba:
Looks Great To Me ;)

extensions/wikia/Flags/controllers/FlagsApiController.class.php
```diff
@@ -0,0 +1,336 @@
...
+	private
+		$cache,
+		$params,
+		$status = false;
+
+	/**
+	 * Returns a singleton instance of FlagsCache
+	 * @return FlagsCache
+	 */
+	private function getCache() {
```
Grunny: This is just a suggestion, since we don't have a convention for it. But I generally think it's preferable to order methods by visibility (public methods all first, then protected and private), which is a common convention. It makes it easy to separate the API from the implementation. Right now the public-facing API is split by various private methods.

extensions/wikia/Flags/models/FlagType.class.php
```diff
@@ -72,7 +72,7 @@ public function getFlagTypesForWikia( $wikiId ) {
 	 * @return bool
 	 */
 	public function verifyParamsForAdd( $params ) {
-		$required = [ 'wikiId', 'flagGroup', 'flagName', 'flagView', 'flagTargeting' ];
+		$required = [ 'wiki_id', 'flagGroup', 'flagName', 'flagView', 'flagTargeting' ];
```
lukaszkonieczny: I think we should change camel case naming to underscored

extensions/wikia/Flags/controllers/FlagsApiController.class.php
adamkarminski: Ok, I understand. The methods have been grouped in terms of a related action, but exposing the `public` methods at the top makes more sense. Thank you!

adamkarminski:
Merging per conversations with @Grunny and @lukaszkonieczny :)

PR #7295, opened 2015-05-21
removeTempUser: maintenance script removing TempUser accounts

macbre:
https://wikia-inc.atlassian.net/browse/CE-1182 (required by https://wikia-inc.atlassian.net/browse/PLATFORM-1055)
- remove `TempUser%` accounts from a given database (defaults to `wikicities`, can be used on `uncyclo` as well - via `--db` option)
- check the structure of the user name - `TempUser+<user ID>`
- remove accounts not touched in the last year only
- perform deletes in batches of 1000 users to limit the impact on shared DB (load and slave lag)

**70% of all accounts** on `wikicities` are TempUsers and ~10% on uncyclo. Let's finally clean it up :)

Based on @kamilkoterba's https://github.com/Wikia/community-engineering-misc/pull/8

@Wikia/community-engineering / @michalroszka

PR #7318, opened 2015-05-25
RemoveTempUserAccounts: run per cluster

macbre:
Improves #7295

Run this script on per-cluster basis to avoid causing lag on all clusters by calling `ExternalUser_Wikia::removeFromSecondaryClusters()`.

@michalroszka / @Wikia/community-engineering


michalroszka:
:+1: - looks like with've become extremely vulnerable for batch operations, recently.

PR #7326, opened 2015-05-25
ADEN-1930 Single Request Architecture for Turtle

fraszczakszymon:


extensions/wikia/AdEngine/js/provider/factoryWikiaGpt.js
```diff
@@ -64,12 +56,6 @@ define('ext.wikia.adEngine.provider.factory.wikiaGpt', [
 				hop(adInfo);
 			}
 
-			if (slotTargeting.skipCall) {
```
nandy-andy: Where this logic got moved? Or what's the reason for removing it? :smile_cat:

nandy-andy:
LGTM :+1: And... MOAR unit tests :heart:

extensions/wikia/AdEngine/js/GptSraHelper.js
```diff
@@ -0,0 +1,52 @@
...
+		log(['shouldFlush', slotName, flushed], 'debug', logGroup);
+		return flushed;
+	}
+
+	function pushAd(slotName, slotPath, slotTargeting, success, error, forcedAdType) {
+		if (shouldPush(slotName, slotTargeting)) {
+			gptHelper.pushAd(slotName, slotPath, slotTargeting, success, error, forcedAdType);
+			log(['pushAd', 'Pushed slot', slotName], 'debug', logGroup);
+		} else {
+			success({});
```
gabrys: I think separating the shouldPush and shouldFlush functions makes pushAd less readable

extensions/wikia/AdEngine/js/provider/factoryWikiaGpt.js
```diff
@@ -77,8 +63,12 @@ define('ext.wikia.adEngine.provider.factory.wikiaGpt', [
 				lookups.extendSlotTargeting(slotName, slotTargeting);
 			}
 
-			gptHelper.pushAd(slotName, slotPath, slotTargeting, doSuccess, doHop);
-			flushIfNeeded();
+			if (gptSraHelper && extra.sraEnabled) {
```
gabrys: * why gptSraHelper is optional?
* document extra.sraEnabled in jsDoc

fraszczakszymon: 1. While it is useless for mobile providers, `//extensions/wikia/AdEngine/js/GptSraHelper.js` dependency is added only to **adengine2_desktop_js** group

2. Doc added

extensions/wikia/Venus/VenusController.class.php
```diff
@@ -81,6 +81,7 @@ private function setAds() {
 		$this->adPrefooterLeftBoxad = $this->app->renderView( 'Ad', 'Index', [ 'slotName' => 'PREFOOTER_LEFT_BOXAD', 'includeLabel' => true ] );
 		$this->adPrefooterRightBoxad = $this->app->renderView( 'Ad', 'Index', [ 'slotName' => 'PREFOOTER_RIGHT_BOXAD', 'includeLabel' => true ] );
 		$this->adsBottom = $this->app->renderView( 'Ad', 'Index', [ 'slotName' => 'GPT_FLUSH' ] );
+		$this->adsBottom = $this->app->renderView( 'Ad', 'Index', [ 'slotName' => 'TURTLE_FLUSH' ] );
```
gabrys: Let's not touch Venus at all. In fact I'll create a story for the CONCF team to clean this up.

extensions/wikia/AdEngine/js/provider/factoryWikiaGpt.js
gabrys: Let's then fail visibly when one requests SRA but the module is not loaded (rather than just not using SRA).

extensions/wikia/AdEngine/js/GptSraHelper.js
```diff
@@ -0,0 +1,45 @@
...
+		flushed = false,
+		sraSlots = [
+			'CORP_TOP_LEADERBOARD',
+			'HOME_TOP_LEADERBOARD',
+			'HUB_TOP_LEADERBOARD',
+			'INVISIBLE_SKIN',
+			'TOP_LEADERBOARD'
+		];
+
+	function shouldFlush(slotName) {
```
nandy-andy: At the beginning I liked those small functions which gave us logging and had descriptive names. But since you removed one of them I can see the benefits of removing the other one as well :smiling_imp: 

I think I can see now what @gabrys saw :japanese_ogre: If we remove this method we could get rid of `flushed` variable as well and use the `sraSlots.indexOf(slotName) === -1` condition directly in `pushAd()` function :smile_cat: We'll gain:
* one variable less,
* less code,
* and we won't loose logging since we can always add logging in `pushAd()` :)

extensions/wikia/AdEngine/js/provider/factoryWikiaGpt.js
```diff
@@ -77,8 +63,12 @@ define('ext.wikia.adEngine.provider.factory.wikiaGpt', [
 				lookups.extendSlotTargeting(slotName, slotTargeting);
 			}
 
-			gptHelper.pushAd(slotName, slotPath, slotTargeting, doSuccess, doHop);
-			flushIfNeeded();
+			if (extra.sraEnabled) {
```
nandy-andy: I'm still thinking if we need this condition here. We have two ways of using `GptSraHelper`:
* use it instead of `GptHelper` in a provider (example: `turtle.js`),
* use it conditionally in a provider - how we make it here.

And I wonder if we could choose one of those way and stick to it. So it's more consistent. My suggestion would be using `GptSraHelper` instead of `GptHelper` - then we could remove also `sraEnabled` flag, I guess. But I have the feeling I'm missing something here and we have to leave the condition here because of other partners.

gabrys:
One other small note, we've been using format without the brackets around the ticket number for commit and pull request names:

`[ADEN-1930] Single Request Architecture for Turtle` -> `ADEN-1930 Single Request Architecture for Turtle`

(We can discuss which one is better and why :-) )

fraszczakszymon:
Oh, I thought that JIRA tracks only commits/PRs with brackets around ticket number. Sure, I can switch to non-brackets - it's cleaner.

extensions/wikia/AdEngine/js/GptSraHelper.js
```diff
@@ -0,0 +1,30 @@
...
+		if (sraSlots.indexOf(slotName) === -1) {
```
gabrys: Please add some logging. This can become tricky at one point, since the flushed is per-module var, so it won't work if multiple GPT-based providers would like to use SRA.

extensions/wikia/AdEngine/js/provider/factoryWikiaGpt.js
```diff
@@ -17,7 +17,7 @@ define('ext.wikia.adEngine.provider.factory.wikiaGpt', [
 	 * @param {Object} [extra]      - optional extra params
 	 * @param {function} [extra.beforeSuccess] - function to call before calling success
 	 * @param {function} [extra.beforeHop]     - function to call before calling hop
-	 * @param {function} [extra.shouldFlush]   - should ads be flushed after given slotname
+	 * @param {boolean}  [extra.sraEnabled]    - decide if use SRA
```
gabrys: decide if use SRA -> whether to use SRA

extensions/wikia/AdEngine/js/spec/GptSraHelper.spec.js
```diff
@@ -0,0 +1,31 @@
...
+
+	it('Returns false when SRA slot passed', function () {
+		expect(getModule().shouldFlush('TOP_LEADERBOARD')).toEqual(false);
+	});
+
+	it('Returns always true when at least one slot has been flushed', function () {
+		var sraHelper = getModule();
+
+		expect(sraHelper.shouldFlush('TOP_RIGHT_BOXAD')).toEqual(true);
+		expect(sraHelper.shouldFlush('TOP_LEADERBOARD')).toEqual(true);
```
gabrys: Good tests

PR #7328, opened 2015-05-25
(MAIN-4620) Sanitize custom theme class in PortableInfobox

Grunny:
Sanitize the custom theme class users can provide in Portable Infoboxes.
Sanitizer::escapeClass only allows valid characters in a class name
through.

/cc @idradm @RafLeszczynski @SebastianMarzjan


RafLeszczynski:
LGTM

PR #7339, opened 2015-05-27
PLATFORM-1250 Hook Helios' clearAccessTokenCookie() to UserLogout

ghost:
Calling Helios' methods when Helios is disabled (Uncyclo) results with fatal errors which is why it is better to hook to core events rather than make explicit calls.
- extensions/wikia/Helios/Helios.setup.php - hook Wikia\Helios\User::clearAccessTokenCookie() to UserLogout event
- extensions/wikia/Helios/Helios.setup.php - Wikia\Helios\User::clearAccessTokenCookie() returns true so that wfRunHooks('UserLogout') evalates to true, too.
- includes/User.php - remove explicit call from doLogout() and rely on wfRunHooks('UserLogout') in logout()

### DIFF
--- a/extensions/wikia/Helios/Helios.setup.php
+++ b/extensions/wikia/Helios/Helios.setup.php
@@ -35,3 +35,4 @@ $wgHooks['ExternalUserWikiaAddToDatabase'][] = 'Wikia\\Helios\\User::onRegister'
 
 $wgHooks['UserSaveSettings'][] = 'Wikia\\Helios\\User::onUserSave';
 $wgHooks['UserSaveOptions'][] = 'Wikia\\Helios\\User::onUserSave';
+$wgHooks['UserLogout'][] = 'Wikia\\Helios\\User::clearAccessTokenCookie';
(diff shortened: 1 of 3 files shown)


extensions/wikia/Helios/Helios.setup.php
```diff
@@ -35,3 +35,4 @@
 
 $wgHooks['UserSaveSettings'][] = 'Wikia\\Helios\\User::onUserSave';
 $wgHooks['UserSaveOptions'][] = 'Wikia\\Helios\\User::onUserSave';
+$wgHooks['UserLogout'][] = 'Wikia\\Helios\\User::clearAccessTokenCookie';
```
macbre: Is `Wikia\\Helios\\User::clearAccessTokenCookie` method called only via this hook? if so, let's rename it to `onUserLogout` to match the methods above.

michalroszka:
@garthwebb - per this is per https://github.com/Wikia/app/commit/9da934006954257a131634a7f9770921056df532 and https://wikia-inc.atlassian.net/browse/PLATFORM-1250

extensions/wikia/Helios/Helios.setup.php
michalroszka: Currently it is, although it seems the author's intention was to make a general purpose method.

garthwebb:
LGTM

PR #7340, opened 2015-05-27
CE-1779 Changing way to get resources from server so it works on mono…

kamilkoterba:
…book and adding some styles adjustments

https://wikia-inc.atlassian.net/browse/CE-1779

Changed from global Wikia.getMultiTypePackage to AMD wikia.loader which does the same.

@Wikia/community-engineering

### DIFF
--- a/extensions/wikia/Flags/scripts/FlagsModal.js
+++ b/extensions/wikia/Flags/scripts/FlagsModal.js
@@ -1,4 +1,4 @@
-require(['jquery'], function ($) {
+require(['jquery', 'wikia.loader'], function ($, loader) {
 	'use strict';
 
 	/* Modal component configuration */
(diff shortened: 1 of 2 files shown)


extensions/wikia/Flags/scripts/FlagsModal.js
```diff
@@ -47,17 +47,19 @@ require(['jquery', 'mw'], function ($, mw) {
...
-			callback: handlePackage
-		});
+		loader({
+			type: loader.MULTI,
+			resources: {
+				templates: [{
+					controller: 'Flags',
+					method: 'editForm',
+					params: {
+						'page_id': window.wgArticleId
```
Grunny: I'd suggest adding `wikia.window` to the required modules.

Grunny:
One minor comment, otherwise LGTM.

extensions/wikia/Flags/scripts/FlagsModal.js
```diff
@@ -47,17 +47,19 @@ require(['jquery', 'mw'], function ($, mw) {
...
+					}
```
Grunny: Use `mw.config.get('wgArticleId')` instead of the global.

extensions/wikia/Flags/scripts/FlagsModal.js
Grunny: Use `mw.config.get('wgArticleId')` instead of the global.

PR #7344, opened 2015-05-27
PLATFORM-1055: script for migrating Uncyclopedia users to the shared DB

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-1055

This script goes through all accounts on Uncyclopedia and migrates them to shared DB (and resolves  name conflicts based on edit count).

Adds:
- `maintenance/wikia/renameUser.php` that allows us to run Wikia's RenameUser from Uncyclopedia context (via `wfShellExec`)
- `maintenance/wikia/uncycloUsersMigrator.php` the migration script that performs The Magic

```
Script specific parameters:
    --csv: Generate a report in CSV file
    --dry-run: Don't perform any operations
    --only-rename-global-users: Perform global users rename ONLY
```

Will be run with `--only-rename-global-users` option first to rename shared DB accounts only without affecting Uncyclopedia.

The migration plan is described in [PLATFORM-1240](https://wikia-inc.atlassian.net/browse/PLATFORM-1240).

@michalroszka


maintenance/wikia/renameUser.php
```diff
@@ -0,0 +1,76 @@
...
+		$logger->pushContext( [
+			'old-name' => $oldName,
+			'new-name' => $newName,
+			'reason'   => $reason
+		] );
+
+		$logger->info( __CLASS__ . ': start' );
+
+		$renameProcess = new RenameUserProcess( $oldName, $newName, true, $reason );
+		$res = $renameProcess->run();
```
wladekb: AFAIK 3 years ago (it might've changed in the meantime though) the main task spawns more tasks, one for each wiki the user edited so after this line the process is not yet completed.

maintenance/wikia/uncycloUsersMigrator.php
```diff
@@ -0,0 +1,695 @@
...
+			if ( !$dbw->tableExists( $tableName ) ) {
+				$this->err( 'Table does not exist', [ 'tableName' => $tableName ] );
+				continue;
+			}
+
+			$dbw->update(
+				$entry[ 'table' ],
+				[ $columnName => $newValue ], // SET
+				[ $columnName => $oldValue ], // WHERE
+				__METHOD__
```
wladekb: no "LIMIT" here? are you confident that it won't fail in the biggest case and break in the middle?
how about doing it in batches of, say, 500, and looping until no row is affected?

macbre: Good point, I simply used the approach from the core RenameUser extension. Batches make more sense here.

maintenance/wikia/uncycloUsersMigrator.php
```diff
@@ -0,0 +1,695 @@
...
+		if ( $this->onlyRenameGlobalUsers ) {
+			return $uncycloUser;
+		}
+
+		$user = clone $uncycloUser;
+		$newName = self::PREFIX_RENAME_UNCYCLO . $user->getName();
+
+		// check user name conflicts
+		if ( $this->getGlobalUserByName( $newName ) ) {
+			$newName .= '-1';
```
wladekb: what if this name also exists? it's worth throwing an exception here in such case

maintenance/wikia/uncycloUsersMigrator.php
```diff
@@ -0,0 +1,695 @@
...
+	 * Rename global user
+	 *
+	 * @param User $user global user
+	 */
+	protected function doRenameGlobalUser( User $user ) {
+		$newName = self::PREFIX_RENAME_GLOBAL . $user->getName();
+
+		// check user name conflicts
+		if ( $this->getGlobalUserByName( $newName ) ) {
+			$newName .= '-1';
```
wladekb: check again if that name exists and throw exception in such case

maintenance/wikia/uncycloUsersMigrator.php
```diff
@@ -0,0 +1,695 @@
...
+
+		$cmd = sprintf( 'php %s/renameUser.php --old-username=%s --new-username=%s --reason=%s',
+			__DIR__,
+			escapeshellarg( $user->getName() ),
+			escapeshellarg( $newName ),
+			escapeshellarg( self::RENAME_REASON )
+		);
+
+		$retVal = 0;
+		$output = wfShellExec( $cmd, $retVal );
```
wladekb: As we learned recently changes made by the external script won't be visible afterwards in this script. I'd suggest committing/closing any connection left to shareddb cluster after processing each user.

maintenance/wikia/uncycloUsersMigrator.php
```diff
@@ -0,0 +1,695 @@
...
+
+		// check the shared users database
+		$globalUser = $this->getGlobalUserByName( $user->getName() );
+		$globalEdits = null;
+
+		if ( $globalUser instanceof User ) {
+			// HACK: calling getEmail() on global user will trigger a SQL query on uncyclo database
+			$this->output( sprintf( ' - conflicts with the global user #%d <%s>', $globalUser->getId(), $globalUser->mEmail ?: 'MISSING_EMAIL' ) );
+
+			$globalEdits  = (int) $globalUser->getEditCount();
```
wladekb: Sure that it fetches number of edits of the global user, not the uncyclo one with the same name or id?

maintenance/wikia/uncycloUsersMigrator.php
```diff
@@ -0,0 +1,695 @@
...
+					$fname
+				);
+			});
+
+			// we have a new ID for a global account
+			// update user ID in uncyclo database
+			$this->doChangeUncycloUserId( $user, $extUser->getId() );
+
+			// delete the user and his settings from the uncyclo DB
+			$this->deleteUncycloUser( $user );
```
wladekb: leaving this should be harmless, sure that we want to do it here? maybe just flag them somehow so the next script run won't catch them

maintenance/wikia/uncycloUsersMigrator.php
macbre: Yes, this value is taken from database as the user is initialized via `User::newFromRow`

maintenance/wikia/renameUser.php
macbre: Correct, `UserRenameTask` is scheduled which spawns process for each wiki.

The plan is to run the migration script in `--only-rename-global-users`` mode first and rename all conflicting shared accounts.

maintenance/wikia/uncycloUsersMigrator.php
macbre: The solution for now is to pre-generate shared accounts for uncyclo accounts and update the `uncyclo.user` entries with the new shared ID.

maintenance/wikia/uncycloUsersMigrator.php
macbre: :+1:

maintenance/wikia/uncycloUsersMigrator.php
macbre: MW's `Database::update` does not allow to pass `LIMIT` for the `UPDATE` query.

maintenance/wikia/uncycloUsersMigrator.php
macbre: The user will be processed just once in the script, so there's no need to keep its state up to date after the background task(s) is completed.

More over, each user migration is wrapped in a transaction, so the state will be updated.

wladekb:
@macbre is it going to be merged some time?

michalroszka:
Let's proceed with it. The script makes sense to me, although let us run it with extra caution plus. And backups are mandatory. :)

PR #7350, opened 2015-05-28
Ce 1877 flags empty state modal message

kamilkoterba:
https://wikia-inc.atlassian.net/browse/CE-1877
Adding empty message to modal instead of edit form when there are no flags on wiki.
Also needed to add buttons depending on whether there are flags or not, as we don't want to have save button when there are not flags.

To do that I also abstracted modal piece of code for rendering buttons to separate public method.

@Wikia/community-engineering , @nandy-andy


kamilkoterba:
Have to fix merge dev commit ;/ will be back in a sec ;)

PR #7351, opened 2015-05-28
Ce 1877 flags empty state modal message 2

kamilkoterba:
https://wikia-inc.atlassian.net/browse/CE-1877
Adding empty message to modal instead of edit form when there are no flags on wiki.
Also needed to add buttons depending on whether there are flags or not, as we don't want to have save button when there are not flags.

To do that I also abstracted modal piece of code for rendering buttons to separate public method.

@Wikia/community-engineering , @nandy-andy

### DIFF
--- a/extensions/wikia/Flags/Flags.i18n.php
+++ b/extensions/wikia/Flags/Flags.i18n.php
@@ -11,9 +11,11 @@ $messages = [];
 $messages['en'] = [
 	'flags-description' => 'Portable flags',
 	'flags-edit-form-more-info' => 'More info >',
-	'flags-edit-modal-title' => 'Flags',
-	'flags-edit-modal-done-button-text' => 'Done',
 	'flags-edit-modal-cancel-button-text' => 'Cancel',
+	'flags-edit-modal-close-button-text' => 'Close',
+	'flags-edit-modal-done-button-text' => 'Done',
+	'flags-edit-modal-no-flags-on-community' => 'This community doesn\'t have any flags set up. [[Help:Interwiki_link|Learn more about flags]] or [[Special:Flags|define the flags for this community]].',
+	'flags-edit-modal-title' => 'Flags'
 ];
 
 /**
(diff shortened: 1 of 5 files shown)


extensions/wikia/Flags/Flags.i18n.php
```diff
@@ -11,9 +11,11 @@
 $messages['en'] = [
 	'flags-description' => 'Portable flags',
 	'flags-edit-form-more-info' => 'More info >',
-	'flags-edit-modal-title' => 'Flags',
-	'flags-edit-modal-done-button-text' => 'Done',
 	'flags-edit-modal-cancel-button-text' => 'Cancel',
+	'flags-edit-modal-close-button-text' => 'Close',
+	'flags-edit-modal-done-button-text' => 'Done',
```
nandy-andy: Don't we have such messages in Oasis skin which we could re-use here?

extensions/wikia/Flags/scripts/FlagsModal.js
```diff
@@ -1,38 +1,50 @@
...
+
+	/* Modal close button config*/
+	buttonForEmptyState = [{
+		vars: {
+			value: mw.message('flags-edit-modal-close-button-text').escaped(),
+			data: [
+				{
+					key: 'event',
+					value: 'close'
+				}]
```
nandy-andy: I'm not sure if our guidelines touch it but you can make it consistent in the file and move `]` to the next line just like above :japanese_ogre:

nandy-andy:
The code changes look good for me. I left only few minor notes - it's up to you to address it or not. Have you checked if the unit tests pass? They should IMHO but consider also adding a new test which checks if `renderButtons` is always being called while modal instance creation.

Another feasible approach would be not to change anything in the core `modal.js` but to modify `modalConfig` conditionally before modal creation.

Overall: code looks good but I'd :heart: a unit test :japanese_ogre:

extensions/wikia/Flags/Flags.i18n.php
kamilkoterba: We want to use it for other skins as well. Haven't found any reasonable place to store it.

nandy-andy: Oh, if it's for other skins I think it's fine to keep it in the extension. I recall Oasis reusable messages file but I don't recall one which is shared between skins.

PR #7356, opened 2015-05-28
CE-1889: Initialize params property for flag

lukaszkonieczny:
https://wikia-inc.atlassian.net/browse/CE-1889


lgarczewski:
OKed.

PR #7359, opened 2015-05-28
HG-714 adding weppy to WikiaMobile

lizlux:
We want to start comparing WikiaMobile performance to Mercury, so this enables Weppy/Bucky perf tracking.

### DIFF
--- a/extensions/wikia/Bucky/js/bucky_init.js
+++ b/extensions/wikia/Bucky/js/bucky_init.js
@@ -20,7 +20,7 @@
 		Weppy.setOptions(config);
 		$(context).on('load', function () {
 			setTimeout(function () {
-				Weppy.sendPagePerformance(false);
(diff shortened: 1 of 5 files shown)


extensions/wikia/Bucky/js/bucky_init.js
```diff
@@ -18,7 +18,7 @@
 			'context': context.wgTransactionContext
 		});
 		Weppy.setOptions(config);
-		$(context).load(function () {
+		$(context).on('load', function () {
 			setTimeout(function () {
 				Weppy.sendPagePerformance(false);
```
wladekb: sendPagePerformance() doesn't use any argument so you can remove the argument

wladekb:
lgtm from code perspective, however I didn't test if that actually works

PR #7361, opened 2015-05-29
CE-1830: Added support for actions like add / replace / remove

lukaszkonieczny:
Added options and support for template actions in migration script.


extensions/wikia/Flags/FlagsExtractor.class.php
```diff
@@ -114,10 +174,268 @@ public function getTemplate() {
...
+	 */
+	private function addFlagToPage() {
+		if ( !empty( $this->actionParams )
+			&& isset( $this->actionParams['wiki_id'] )
+			&& isset( $this->actionParams['page_id'] )
+		) {
+			$flagsToPages = $this->prepareParameters();
+
+			if ( !$flagsToPages ) {
+				return false;
```
lgarczewski: This needs some sort of logging, even if it's just an echo to stdout. We'd need to know that this failed.

Up to you weather this would be an exception or just a log about the error

extensions/wikia/Flags/FlagsExtractor.class.php
```diff
@@ -114,10 +174,268 @@ public function getTemplate() {
...
+	 *
+	 * @return array|bool
+	 */
+	private function prepareParameters() {
+		$params = [];
+
+		$template = $this->getLastTemplate();
+
+		if ( !$template ) {
+			return false;
```
lgarczewski: Same here: error logging?

extensions/wikia/Flags/FlagsExtractor.class.php
```diff
@@ -114,10 +174,268 @@ public function getTemplate() {
...
+
+		if ( !empty( $template['params'] ) ) {
+			$params = $template['params'];
+		}
+
+		if ( !isset( $this->actionParams['flag_type_id'] ) || empty( $this->actionParams['flag_type_id'] ) ) {
+			if ( $flagTypeId = $this->getFlagTypeIdByTemplate( $template['name'] ) ) {
+				$this->actionParams['flag_type_id'] =  $flagTypeId;
+			} else {
+				return false;
```
lgarczewski: and probably here: error logging?

lgarczewski:
Added suggestions to add logging in a couple places, once that's done this is good to go.

lgarczewski:
LGTM after additions.

PR #7362, opened 2015-05-29
DAT-2822 Support new breakpoints in Njord Prototype by using negative margins

sebastianmarzjan:
This feature will be either removed or fully productized; In the latter case, moving the wrapper outside of page container may be considered.


RafLeszczynski:
LGTM

RafLeszczynski:
LGTM

PR #7365, opened 2015-05-29
CONCF-598 | Typography update: Bloglist headings are showing up too large

vforge:
/cc @bognix @hakubo @Warkot


Warkot:
LGTM

extensions/3rdparty/tabber/tabber.css
```diff
@@ -102,12 +102,16 @@ ul.tabbernav li.tabberactive a:hover
...
+/*
+Disabled as CONCF-598 fixes needs those back.
+
 .tabberlive .tabbertab h2 {
  display:none;
 }
 .tabberlive .tabbertab h3 {
  display:none;
 }
+*/
```
bognix: can't we just remove it?

extensions/wikia/Blogs/css/oasis.scss
```diff
@@ -60,7 +60,7 @@
 		padding: 0 10px;
 	}
 
-	h1 {
```
bognix: :+1:

extensions/wikia/Blogs/css/oasis.scss
```diff
@@ -261,7 +261,7 @@
 	margin: 4px 2px 0;
 }
 
-wpCategoryInput {
+#wpCategoryInput {
```
bognix: I couldn't find element with this selector in our repo.
We are styling this element couple of times but I don't see where it is created.

vforge: Neither do I, but it was a bug in the CSS file - and based on https://github.com/Wikia/app/pull/7365#discussion-diff-31339417R259 I've fixed declaration. I'm not sure that we use it, but it may be used by our users.

extensions/3rdparty/tabber/tabber.css
vforge: https://github.com/Wikia/app/commit/14a1c936576acc1671bd0ac98479f36b3d75be34

bognix:
LGTM

extensions/wikia/Blogs/css/oasis.scss
hakubo: Since it was not working,
and nobody filed a ticket for that
I would actually remove it
as if we add some styles to it, we might brake something else :)

PR #7366, opened 2015-05-29
CONCF-553 add wikistats to mercury api response

bognix:
https://wikia-inc.atlassian.net/browse/CONCF-553

### DIFF
--- a/extensions/wikia/MercuryApi/MercuryApiController.class.php
+++ b/extensions/wikia/MercuryApi/MercuryApiController.class.php
@@ -336,7 +336,6 @@ class MercuryApiController extends WikiaController {
 	}
 
 	private function getMainPageData() {
-
 		$mainPageData = [];
 		$curatedContent = $this->getCuratedContentData();
 		$trendingArticles = $this->getTrendingArticlesData();
@@ -360,7 +359,7 @@ class MercuryApiController extends WikiaController {
 		}
 
 		if ( !empty( $wikiStats ) ) {
-			$mainPageData['wikiStats'] = $wikiStats;
+			$mainPageData[ 'wikiStats' ] = $wikiStats;
 		}
 
 		return $mainPageData;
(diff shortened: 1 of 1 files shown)


extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -428,4 +434,18 @@ private function getTrendingVideosData() {
...
 	}
+
+	private function getWikiStats() {
+		global $wgCityId;
+
+		$wikiStats = [];
+		$service = new WikiDetailsService();
+		$wikiDetails = $service->getWikiDetails( $wgCityId );
+
+		if ( !empty( $wikiDetails['stats'] ) ) {
```
vforge: As we always want to return some value or empty value is that `if` even nessesary? Wouldn't be better to replace that block with just `return $wikiDetails['stats']`?

bognix: I figured it is better to return empty array instead of undefined so we can avoid errors in: 
```
{{#if model.pages}}
```
because we are referencing key from object when object doesn't exists.

extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -357,6 +359,10 @@ private function getMainPageData() {
 			$mainPageData[ 'trendingVideos' ] = $trendingVideos;
 		}
 
+		if ( !empty( $wikiStats ) ) {
+			$mainPageData['wikiStats'] = $wikiStats;
```
rogatty: Inconsistency in whitespaces.

extensions/wikia/MercuryApi/MercuryApiController.class.php
rogatty: We won't reference key from non existent object after you change https://github.com/Wikia/mercury/pull/881/files#r31415277

vforge:
:heavy_check_mark: LGTM

rogatty:
LGTM

PR #7368, opened 2015-05-29
robots.txt: remove the msnbot delay

ljagiello:
We're suffering for lack of crawling from msnbot and we should not limit that traffic.

/cc @tedgill @macbre


tedgill:
:+1:

owend:
Sure, LGTM

PR #7369, opened 2015-05-30
SOC-831 Make sure to load revisions on the right wiki

garthwebb:
Lots of additional cleanup to make this all work right.


extensions/wikia/Wall/WallBaseController.class.php
```diff
@@ -257,7 +257,7 @@ public function message() {
 			$this->response->setVal( 'editorUrl',  $editorUrl );
 			$this->response->setVal( 'isEdited',  true );
```
garthwebb: Spelling fix

extensions/wikia/Wall/WallMessage.class.php
```diff
@@ -448,7 +448,7 @@ public function setNotifyEveryone( $notifyEveryone ) {
 					wfMessage( 'wall-message-update-highlight-summary' )->inContentLanguage()->text(),
 					false, true );
 				$rev = $this->getArticleComment()->mLastRevision;
```
garthwebb: I'm eliminating the city ID argument from places where we shouldn't need it.  Were we need it are times when the Revision data is on a wiki that is not the current wiki.  Since we already have a revision object created here in ```$rev``` we have what we need and the city ID is not necessary.

extensions/wikia/Wall/WallMessage.class.php
```diff
@@ -498,8 +498,8 @@ public function getMainPageText() {
 		return $wallOwnerName;
 	}
```
garthwebb: The name ```$master``` is vague, ```$useMasterDB`` says what is actually intended.

extensions/wikia/Wall/WallMessage.class.php
```diff
@@ -846,7 +846,7 @@ public function notifyEveryone() {
 			return true;
 		}
```
garthwebb: Don't need a city ID here since we already have a ```Revision``` object.

extensions/wikia/Wall/WallMessage.class.php
```diff
@@ -897,7 +897,7 @@ public function isEdited() {
 		return $this->getArticleComment()->mLastRevId != $this->getArticleComment()->mFirstRevId;
 	}
```
garthwebb: Spelling fix.

extensions/wikia/WallNotifications/WallNotificationEntity.class.php
```diff
@@ -1,61 +1,116 @@
 <?php
 
-class WallNotificationEntity {
+use Wikia\Logger\WikiaLogger;
 
+class WallNotificationEntity {
+	/** @var int */
 	public $id;
```
garthwebb: I have no idea why the data in ```$data_noncached``` is not supposed to be cached (and ```$data_non_cached``` isn't used at all) but most of the data stored in ```$data_noncached``` isn't used anywhere, so I declared the few that are as real class members instead since they don't need to be handled in bulk (as ```$data``` does)

extensions/wikia/WallNotifications/WallNotificationEntity.class.php
```diff
@@ -1,61 +1,116 @@
...
-	 * @param Revision $rev
-	 * @param $wikiId
-	 * @param bool $master
+	 * Create a new object from an existing revision object.
+	 *
+	 * @param Revision $rev A revision object
+	 * @param bool $useMasterDB Whether to query the MASTER DB on Title lookup.
 	 *
 	 * @return WallNotificationEntity
 	 */
```
garthwebb: If you have a ```Revision``` object already you don't need to specify a city ID.  You should be able to get the city from ```$wgCityId```.

extensions/wikia/WallNotifications/WallNotificationEntity.class.php
```diff
@@ -1,61 +1,110 @@
...
-		if ( $wn->loadDataFromRev( $rev, $wikiId, $master ) ) {
-			$wn->save();
+
+		if ( $wn->loadDataFromRev( $rev, $useMasterDB ) ) {
+			$wn->saveToCache();
 			return $wn;
 		}
 
 		return null;
 	}
```
garthwebb: This method wasn't used anywhere

extensions/wikia/WallNotifications/WallNotificationEntity.class.php
```diff
@@ -1,61 +1,110 @@
...
+	 * owns the data.
+	 *
+	 * @param int $revId The revision ID for this notification entity
+	 * @param int $wikiId The wiki where this revision exists
+	 * @param bool $useMasterDB Whether to query the MASTER DB on Title lookup.
+	 *
+	 * @return WallNotificationEntity
+	 */
+	public static function createFromRevId( $revId, $wikiId, $useMasterDB = false ) {
 		$wn = new WallNotificationEntity();
```
garthwebb: The cache lookups are still here, but below where the ```getById``` definition has moved.

extensions/wikia/WallNotifications/WallNotificationEntity.class.php
```diff
@@ -78,174 +127,202 @@ public function getUniqueId() {
 		}
 	}
```
garthwebb: Why have a single getter for ```id``` when none of the other members have getters and this one is not used uniformly?  I played with adding more getters but Wall code liberally accesses the ```data``` member and in ways that PHPStorm didn't recognize.  Eliminating all getters seemed a safer way to achieve consistency.

extensions/wikia/WallNotifications/WallNotificationEntity.class.php
```diff
@@ -78,174 +127,202 @@ public function getUniqueId() {
...
 
-	public function getId() {
-		return $this->id;
+	/**
```
garthwebb: This is the big fix.  Call ```loadDataFromRevId``` on the wiki given by ```$wikiId``` when ```$wikiId``` is not the current wiki.

extensions/wikia/WallNotifications/WallNotificationEntity.class.php
```diff
@@ -78,174 +127,202 @@ public function getUniqueId() {
...
+		return $this->loadDataFromRev( $rev, $userMasterDB );
+	}
 
-		$wm = WallMessage::newFromTitle( $rev->getTitle() ); /* @var $wm WallMessage */
+	/**
+	 * @param Revision $rev
+	 * @param bool $useMasterDB
+	 *
+	 * @return bool
+	 */
```
garthwebb: This method used to be really enormous.  I tried to organize it logically and move some things into smaller methods.  Its still a bit cumbersome however, due to the large number of data members stuffed into ```$data```

extensions/wikia/WallNotifications/WallNotificationEntity.class.php
```diff
@@ -78,174 +127,202 @@ public function getUniqueId() {
...
-			$this->data->thread_title = $wm->getMetaTitle();
-		}
-
-		return true;
-	}
-
-	protected function buildId( $wikiId, $RCid ) {
-
-	}
-
```
garthwebb: I thought ```recreateFromDB``` was a weird name and totally didn't match the other ```createFromXXXX``` methods that do similar things.  So I renamed this ```createFromId``` (since that's what it does given an entity ID) and moved it up next to the other create methods.

extensions/wikia/WallNotifications/WallNotificationEntity.class.php
```diff
@@ -78,174 +127,202 @@ public function getUniqueId() {
...
-			$this->save();
-			return;
+			$data->parent_username = wfMessage( 'oasis-anon-user' )->text();
+			$data->parent_displayname = $data->parent_username;
+			$data->parent_user_id = 0;
 		}
-
-		$this->loadDataFromRev( $rev, $wikiId );
-		$this->save();
 	}
```
garthwebb: I thought ```save``` was misleading since most of the time this implies permanent storage.  Updated to declare that its actually saving to cache.

extensions/wikia/WallNotifications/WallNotificationEntity.class.php
```diff
@@ -1,61 +1,110 @@
...
+	public static function createFromId( $entityId, $useMasterDB = false ) {
+		list( $revId, $wikiId ) = explode( '_', $entityId );
+
+		return self::createFromRevId( $revId, $wikiId, $useMasterDB );
+	}
+
+	/**
+	 * This method attempts to load data from the cache first before
+	 *
+	 * @param $id
```
mklucsarits: int

extensions/wikia/WallNotifications/WallNotificationEntity.class.php
```diff
@@ -78,174 +127,231 @@ public function getUniqueId() {
...
-			$this->data->msg_author_username = $authoruser->getName();
-			if ( $authoruser->getId() > 0 ) {
-				$this->data->msg_author_displayname = $this->data->msg_author_username;
-			} else {
-				$this->data->msg_author_displayname = wfMessage( 'oasis-anon-user' )->text();
-			}
+		// If we're missing either of these, treat as an anonymous user
+		if ( empty( $authorName ) || empty( $userId ) ) {
+			$data->msg_author_id = 0;
+			$data->msg_author_displayname = wfMessage( 'oasis-anon-user' )->text();
```
jsutterfield: `->escaped()`?

extensions/wikia/WallNotifications/WallNotificationEntity.class.php
```diff
@@ -78,174 +127,231 @@ public function getUniqueId() {
...
-		// TODO: double ?
-		$this->data->title_id = $wm->getTitle()->getArticleId();
-
-		$this->data_noncached->title = $wm->getTitle();
+	/**
+	 * This method returns a user object given a user ID.
+	 *
+	 * @param int $userId User ID for the author of this post
+	 *
+	 * @return User
```
jsutterfield: Also could return null.

extensions/wikia/WallNotifications/WallNotificationEntity.class.php
```diff
@@ -78,174 +127,231 @@ public function getUniqueId() {
...
+		$parentUser = $parent->getUser();
+
+		if ( $parentUser instanceof User ) {
+			$data->parent_username = $parentUser->getName();
+			$data->parent_user_id = $parentUser->getId();
+
+			if ( $data->parent_user_id > 0 ) {
+				$data->parent_displayname = $data->parent_username;
 			} else {
+				$data->parent_displayname = wfMessage( 'oasis-anon-user' )->text();
```
jsutterfield: `->escaped()` here, and line 341?

extensions/wikia/WallNotifications/WallNotificationEntity.class.php
```diff
@@ -78,174 +127,231 @@ public function getUniqueId() {
...
-	}
-
-	/**
-	 * @return Memcache
-	 */
-	public function getCache() {
-		global $wgMemc;
-		return $wgMemc;
+	public static function getMemcKey( $id ) {
+		return wfSharedMemcKey( __CLASS__, "v32", $id, 'notification' );
```
jsutterfield: v32! That's a lot of versions.

extensions/wikia/WallNotifications/WallNotificationEntity.class.php
```diff
@@ -78,174 +127,231 @@ public function getUniqueId() {
...
+			'revId' => $revId,
+			'useMasterDB' => $useMasterDB,
+		];
+
+		$response = ApiService::foreignCall( $dbName, $params, ApiService::WIKIA );
+		if ( !empty( $response['status'] ) && $response['status'] == 'ok' ) {
+			$this->parentTitleDbKey = $response['parentTitleDbKey'];
+			$this->msgText = $response['msgText'];
+			$this->threadTitleFull = ['threadTitleFull'];
+			$this->data = ['data'];
```
jsutterfield: Are lines 152 and 153 missing `$response`?

jsutterfield:
LGTM

mklucsarits:
LGTM

PR #7370, opened 2015-05-30
SOC-847 update wall_notification code to use a more sane query

jsutterfield:
Ticket: https://wikia-inc.atlassian.net/browse/SOC-847

See the ticket for an explanation as to what's wrong with the original query and why it was updated.


mklucsarits:
LGTM

garthwebb:
LGTM!

PR #7378, opened 2015-06-01
CE-1822 Adding logging for flags removal

kamilkoterba:
https://wikia-inc.atlassian.net/browse/CE-1822

Adding logging about removed flags to logging table to appear on Special:Log and Special:RecentChanges.

@Wikia/community-engineering


extensions/wikia/Flags/tasks/FlagsLogTask.php
```diff
@@ -8,18 +8,22 @@ class FlagsLogTask extends BaseTask {
...
+	 * invoke ->wikiId( ... ) method before queuing task.
+	 *
 	 * @param array $flags list of flags changed, each item of that list is an array with flag fields as items
-	 * @param int $wikiId ID of wiki where the article is
 	 * @param int $pageId ID of article where flags were changed
 	 * @param string $actionType Type of action performed on flag represented by constants in Flags\Models\Flag class
 	 */
-	public function logFlagChange( array $flags, $wikiId, $pageId, $actionType ) {
+	public function logFlagChange( array $flags, $pageId, $actionType ) {
+		global $wgCityId;
```
lgarczewski: Why are we using a global here?

extensions/wikia/Flags/models/Flag.class.php
```diff
@@ -326,4 +324,17 @@ private function verifyParamsForRemove( $flags ) {
...
 		return true;
 	}
+
+	/**
+	 * Queue task for logging flag change
+	 * @param array $flags list of flags changed, each item of that list is an array with flag fields as items
+	 * @param int $pageId ID of article where flags were changed
+	 * @param string $action Type of action performed on flag represented by constants in Flags\Models\Flag class
+	 */
+	private function logFlagChange( $flags, $pageId, $action ) {
```
adamkarminski: Can you rename `$action` to `$actionType` to be consistent with the `FlagsLogTask::logFlagChange`?

extensions/wikia/Flags/tasks/FlagsLogTask.php
kamilkoterba: beacause 'Task need to be run in wikia context' plus we want to fetch types for wiki to log type name.
I don't want to pass it as parameter to avoid potential difference between actual wiki context and value of cityId passed.

Do you have different idea?

adamkarminski: You can overwrite the `wiki_id` variable when making an API request so I think we should be consistent here and log othe event on a wikia that has been specified in the request (and I consider not using a custom `wiki_id` parameter specifying as well :) ).

kamilkoterba: Good point

PR #7382, opened 2015-06-01
ADEN-1925 Removed PageFair code

nandy-andy:


gabrys:
Also clean up this one: https://github.com/Wikia/app/blob/dev/skins/oasis/modules/templates/Ad_Index.php

gabrys:
:car: LGTM

PR #7384, opened 2015-06-01
CONCF-416 initial work for removing social dropdown from header

bognix:
Remove the old "Share" button from Page Header.

ping @vforge @Warkot


vforge:
https://i.imgflip.com/mbv5e.jpg

LGTM

skins/oasis/js/SharingToolbar/SharingToolbar.js
```diff
@@ -1,189 +0,0 @@
...
-												}
-											};
-
-										uiModal.createComponent( afterShareModalConfig,
-											function ( afterShareEmailModal ) {
-												afterShareEmailModal.bind( 'close', function ( event ) {
-													event.preventDefault();
-
-													if ( result.success ) {
-														window.UserLogin.refreshIfAfterForceLogin();
```
Warkot: 14 tabs… I'm happy that we remove this. :scissors:

extensions/wikia/Lightbox/Lightbox.i18n.php
```diff
@@ -73,11 +73,7 @@
 	'lightbox-share-email-label' => 'Email this image to a friend',
 	'lightbox-share-email-page-label' => 'Email this page to a friend',
 	'lightbox-share-email-page-label-address' => 'Email address',
-	'lightbox-share-email-ok-caption' => 'Confirmation',
 	'lightbox-share-email-ok-content' => '{{PLURAL:$1|Email has|Emails have}} been sent.',
```
Warkot: Why are you not removing this key? Only use of it I found is here: https://github.com/Wikia/app/blob/dev/skins/oasis/modules/SharingToolbarController.class.php#L132 And what about other keys?

Warkot: And I'm not sure but aren't those actually used for lightbox sharing button?

bognix: Regarding first comment: https://github.com/Wikia/app/blob/dev/extensions/wikia/Lightbox/LightboxController.class.php#L367

Regarding second comment: 
I checked the code and couldn't find any usages. I didn't won't to assume this messages keys are somehow concatenated somewhere in the code. I thought we can remove those keys and test lightbox sharing if all messages are displayed correctly. If something is broken, it is fairly easy to bring messages back, on the other hand if everything works fine we leave clean code behind us :)

Warkot: OK, let's do it this way, but I'm still suspicious. I'm just wondering why these messages are in `Lightbox.i18n.php` and not in something like `SharingToolbar.i18n.php`.

And BTW, links like this one `http://elderscrolls.wikia.com/wikia.php?controller=Lightbox&method=lightboxModalContent&format=html&lightboxVersion=1433238455&userLang=qqx` which are sent as XHRs are broken and should be `http://elderscrolls.wikia.com/wikia.php?controller=Lightbox&method=lightboxModalContent&format=html&lightboxVersion=1433238455&uselang=qqx` (`userLang` → `uselang`). I'll add it to the ticket.

Warkot:
LGTM and let's see what QM brings up. :wink:

extensions/wikia/Lightbox/Lightbox.i18n.php
bognix: It is something different than useLang. userLang represents wgUserLanguage variable.

Warkot: Thanks for the info. What I meant is that changing `userLang` param in this query has no visible effect (but don't trust me for sure, I have near-sightedness condition). :see_no_evil:

PR #7385, opened 2015-06-01
Improve EmailIntegrationTest to verify image is present on production

jsutterfield:
The EmailIntegrationTest::testEmailImages was failing frequently in dev. After some investigation (see SOC-816), it was decided this test makes more sense to just run against production every time. This updates that test to do that.

This PR also adds the base Vignette URLs as class constants. Previously they were just hardcoded in the config.

Ticket: https://wikia-inc.atlassian.net/browse/SOC-860

### DIFF
--- a/extensions/wikia/Email/tests/EmailIntegrationTest.php
+++ b/extensions/wikia/Email/tests/EmailIntegrationTest.php
@@ -6,6 +6,9 @@
  * @group Integration
  */
 class EmailIntegrationTest extends WikiaBaseTest {
+
+	const VIGNETTE_BASE_URL_PROD = 'http://vignette<SHARD>.wikia.nocookie.net';
+
 	function setUp() {
 		$this->setupFile = __DIR__ . '/../Email.setup.php';
 		include_once( __DIR__ . '/../../../../includes/HttpFunctions.php');
@@ -34,6 +37,6 @@ class EmailIntegrationTest extends WikiaBaseTest {
 	 */
 	private function setVignetteEnvToProd() {
 		global $wgVignetteUrl;
-		$wgVignetteUrl = \Wikia\Vignette\UrlGenerator::BASE_URL_PROD;
+		$wgVignetteUrl = self::VIGNETTE_BASE_URL_PROD;
 	}
 }
--- a/extensions/wikia/Email/tests/EmailIntegrationTest.php
+++ b/extensions/wikia/Email/tests/EmailIntegrationTest.php
@@ -26,7 +26,7 @@ class EmailIntegrationTest extends WikiaBaseTest {
 			$url = $iconInfo['url'];
 			$name = $iconInfo['name'];
 			$response = HTTP::get( $url );
-			$this->assertTrue( $response !== false, "{$name} should return HTTP 200" );
+			$this->assertTrue( $response !== false, "{$name} should return HTTP 200 -- {$url}" );
 		}
 	}
 
@@ -36,7 +36,6 @@ class EmailIntegrationTest extends WikiaBaseTest {
 	 * part of our emails, we're updating that test to only run in prod. See SOC-860.
 	 */
 	private function setVignetteEnvToProd() {
-		global $wgVignetteUrl;
-		$wgVignetteUrl = self::VIGNETTE_BASE_URL_PROD;
+		$this->mockGlobalVariable('wgVignetteUrl', self::VIGNETTE_BASE_URL_PROD);
 	}
 }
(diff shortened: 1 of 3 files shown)


jsutterfield:
Ha, failed build above because of this unit test. That's why we're doing this ticket :)

The URLs were being cached, I just cleared the cache and re-ran the test. It's passing now: http://qa-s2.wikia-prod:8080/job/unitTests_php_branches/15713/

lib/Wikia/src/Vignette/UrlGenerator.php
```diff
@@ -31,6 +31,10 @@ class UrlGenerator {
 
 	const REVISION_LATEST = 'latest';
 
+	const BASE_URL_PROD = 'http://vignette<SHARD>.wikia.nocookie.net';
```
nmonterroso: We've kept integration with Wikia/Mediawiki to `includes/wikia/vignette` and `lib/Wikia/src/Vignette` as an environment-agnostic library for generating image URLs, so I'd say that `includes/wikia/vignette` is probably a better place for these configurations to live, and even better would be somewhere in the config repo if that's easy to do.

drsnyder: The spirit of keeping these out of `/lib/Wikia` is the hopes that these vignette libraries could be decoupled from MW and opened as separate libraries. If we embed Wikia specific config we can't do that or we'll have to unwind it later.

jsutterfield: Hmm. Okay, well, maybe it makes the most sense to just put 'http://vignette<SHARD>.wikia.nocookie.net' in the test itself. Looking through the rest of the config repo, I don't see it referencing the app repo anywhere else. It does end up repeating some code, but it keeps all of our config in the config repo and doesn't tie to /lib/Wikia to MW.

extensions/wikia/Email/tests/EmailIntegrationTest.php
```diff
@@ -16,15 +19,24 @@ function setUp() {
...
+		$this->setVignetteEnvToProd();
 
-		foreach ( $icons as $info ) {
-			$url = $info['url'];
-			$name = $info['name'];
+		foreach ( Email\ImageHelper::getIconInfo() as $iconInfo ) {
+			$url = $iconInfo['url'];
+			$name = $iconInfo['name'];
 			$response = HTTP::get( $url );
 			$this->assertTrue( $response !== false, "{$name} should return HTTP 200" );
```
macbre: Adding a `$url` in the assert message should help debugging in case of a failed assertion.

extensions/wikia/Email/tests/EmailIntegrationTest.php
```diff
@@ -16,15 +19,24 @@ function setUp() {
...
 	}
+
+	/**
+	 * Set the Vignette base URL to point to production. The testEmailImages test above
+	 * was failing frequently in dev and since prod images are what our user's see as
+	 * part of our emails, we're updating that test to only run in prod. See SOC-860.
+	 */
+	private function setVignetteEnvToProd() {
+		global $wgVignetteUrl;
+		$wgVignetteUrl = self::VIGNETTE_BASE_URL_PROD;
```
macbre: Please mock the global variable instead of polluting the global scope for next tests.

mklucsarits:
LGTM

macbre:
Thanks!

garthwebb:
LGTM!

kvas-damian:
@jsutterfield Do I understand this PR corretly, are we fetching url from dev DB and verifying it exists on production vignette?
IMO we should verify that url generated on production is valid.

jsutterfield:
@kvas-damian it doesn't pull from the database, however it does pull from memcached which should be disabled or we'll often get dev URLs so I'm glad you double checked.

Disabling memcached makes sure that we don't use cache [here](https://github.com/Wikia/app/blob/dev/extensions/wikia/Email/EmailImageHelper.class.php#L45), and then we're overriding  `$wgVignetteUrl` so that when the URL is constructed it's done so as if we're on production.

Here's the updated PR which disabled memcached. https://github.com/Wikia/app/pull/7413 Can you take a look?

kvas-damian:
Thanks for clarification. I thought there is some data required from DB to calculate URL, but it's not true.

The only thing right now that could be different in production and dev environment is ```wgUploadPath``` https://github.com/Wikia/app/blob/dev/includes/wikia/GlobalFile.class.php#L90
I don't expect it changes to often so I think we can go with your solution.

PR #7386, opened 2015-06-02
DAT-2858 Add 'context' to media image

dianafa:
https://wikia-inc.atlassian.net/browse/DAT-2858

### DIFF
--- a/extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
+++ b/extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
@@ -21,7 +21,8 @@ class ArticleAsJson extends WikiaService {
 
 	private static function createMediaObj( $details, $imageName, $caption = '', $link = null ) {
 		wfProfileIn( __METHOD__ );
-
+		
+		$context = '';
 		$media = [
 			'type' => $details['mediaType'],
 			'url' => $details['rawImageUrl'],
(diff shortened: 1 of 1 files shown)


extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
```diff
@@ -118,6 +122,7 @@ public static function onPortableInfoboxNodeImageGetData( $title, &$ref, $alt )
 		wfProfileIn( __METHOD__ );
 		if ( $title ) {
 			$details = WikiaFileHelper::getMediaDetail( $title, self::$mediaDetailConfig );
+			$details['context'] = 'infobox-big';
```
SebastianMarzjan: I wonder if it'd make sense to make possible contexts constants in ArticleAsJson extension. Obviously, this is the first explicit context we are introducing, so it's up to your discretion how to handle it. My two sentiments would be to either extract const or leave a todo to address when another context is added.

extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
```diff
@@ -35,6 +35,10 @@ private static function createMediaObj( $details, $imageName, $caption = '', $li
 			$media['link'] = $link;
 		}
 
+		if ( is_string( $details['context'] ) && $details['context'] !== '' ) {
```
SebastianMarzjan: is_string throws a notice for undefined array index

SebastianMarzjan:
:+1:

PR #7387, opened 2015-06-02
PLATFORM-1258: Removed Universal Analytics auto linker feature.

harnash:
There is a requests to remove Universal Analytics Auto Linker feature since it is making hard for other teams to handle links on wiki pages. It should not affect Universal Analytics stats.


macbre:
Syntactic LGTM

PR #7388, opened 2015-06-02
CE-1848 Add logging of exceptions and cache usage

adamkarminski:
@Wikia/community-engineering

### DIFF
--- a/extensions/wikia/Flags/controllers/FlagsApiController.class.php
+++ b/extensions/wikia/Flags/controllers/FlagsApiController.class.php
@@ -12,8 +12,12 @@
 use Flags\FlagsCache;
 use Flags\Models\Flag;
 use Flags\Models\FlagType;
+use Wikia\Logger\Loggable;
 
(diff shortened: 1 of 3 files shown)


lukaszkonieczny:
lgtm

extensions/wikia/Flags/controllers/FlagsApiController.class.php
```diff
@@ -325,6 +339,9 @@ private function getFlagTypesForWikiaRawData( $wikiId ) {
 			$flagTypesForWikia = $flagTypeModel->getFlagTypesForWikia( $wikiId );
 
 			$flagsCache->setFlagTypesForWikia( $flagTypesForWikia );
+			$this->logCache( 'getFlagsForPageForEdit', 'MISS' );
+		} else {
+			$this->logCache( 'getFlagsForPageForEdit', 'HIT' );
```
macbre: If 1% sampling is fine for you, we do collect memcache stats (number of hits, misses, gets, sets) in Influx DB.

PR #7390, opened 2015-06-02
Ce 1833 admin dashboard buttons

kamilkoterba:
https://wikia-inc.atlassian.net/browse/CE-1833

Fix add image and add page buttons on admin dashboard

add page was broken because name of method was changed to requestDialog
add photo was broken because assets weren't sent

@Wikia/community-engineering


lgarczewski:
LGTM -- but please check CI tests before merging.

kamilkoterba:
Passed after restart

PR #7391, opened 2015-06-02
CE-1754 Rendering Flags based on a user's login status

adamkarminski:
https://wikia-inc.atlassian.net/browse/CE-1754

@Wikia/community-engineering


kamilkoterba:
lgtm

PR #7392, opened 2015-06-02
CONCF 445

hakubo:
Loads of changes so unit tests are failing.
But before I jump in and fix those, I want to be sure that those changes make sense.

@kvas-damian @nandy-andy @SebastianMarzjan 

I could not find how wikia.com loads images from a specific hub. Did it ever work?
More examples you can find in a ticket, here is one:
You can see all images here: http://gameshub.wikia.com/wiki/Games_Hub
but not in the APIs
http://www.wikia.com/api/v1/WikiaHubs/ModuleDataV3?module=1&city=955764

and sometimes they do:
http://www.wikia.com/api/v1/WikiaHubs/ModuleDataV3?module=5&city=955764

I made a change where I always explicitly ask GlobalFile with a correct cityId. But maybe there is something that I am missing there?

I also can not find any fallback for images made with foreignrepo.

I spent way to much time on this right now. Do you think you can help me with this one?


extensions/wikia/WikiaHubsServices/modules/WikiaHubsModuleService.class.php
```diff
@@ -90,8 +91,8 @@ protected function getView($viewName, $data) {
 		return $this->app->getView(get_class($this), $viewName, $data);
 	}
 
-	protected function getImageInfo($fileName, $destSize = 0) {
-		return ImagesService::getLocalFileThumbUrlAndSizes($fileName, $destSize, ImagesService::EXT_JPG);
+	protected function getImageInfo($fileName) {
+		return GlobalFile::newFromText($fileName, $this->cityId);
```
kvas-damian: Please use JPGs in output urls to improve performance.

extensions/wikia/WikiaHubsServices/modules/WikiaHubsModuleFromthecommunityService.class.php
```diff
@@ -259,21 +260,24 @@ public function getStructuredData($data) {
...
 					'wikiUrl' => $data['wikiUrl' . $i],
 					'quote' => $data['quote' . $i],
 					'photoName' => isset($data['photo'.$i]) ? $data['photo'.$i] : ''
 				);
+
+				if( !empty($data['photo' . $i]) ) {
+					$imageData = $this->getImageInfo($data['photo' . $i]);
+
+					if ($imageData) {
+						$entries['imageAlt'] = $imageData->getName();
```
nandy-andy: I don't think this is what you want. You probably want to have `imageAlt` and `imageUrl` set for one specific entry from the `$entries`.

extensions/wikia/WikiaHubsServices/modules/WikiaHubsModuleService.class.php
nandy-andy: Nice. I didn't know `GlobalFile`. It probably works better than `ImagesService` which worked some time ago :smile_cat: 
And it seems it existed back when we worked on hubs services... what a pity we didn't know it.

nandy-andy:
The code overall looks fine - just address one of my notes, please. However, I don't think I'm the right person to write "LGTM" here ;) //ping @kvas-damian

kvas-damian:
@hakubo I double check why some API responses have URLs in place. Most likely it's because data stored in memcache was generated on a Hub wiki and right now we don't need to make those calculations again.

To verify that hypothesis I rebuilt memcache on one working API response and URLs disappeared. Then I rebuild memecahe on Hub page and API response was fixed. 

Looks that this feature never workd correctly before :(
Your solution looks ok, the only thing I'd verify is performance of ```GlobalFile```. Don't know if it's still valid but some time ago we didn't cache it enough.

extensions/wikia/WikiaHubsServices/modules/WikiaHubsModuleService.class.php
```diff
@@ -10,25 +10,36 @@
...
+		$hubWiki = WikiFactory::getWikiByID($cityId);
+		$hubDbName = $hubWiki->city_dbname;
+
+		$wgForeignFileRepos[] = [
+			'class'            => 'WikiaForeignDBViaLBRepo',
+			'name'             => $hubDbName,
+			'transformVia404'  => true,
+			'wiki'             => $hubDbName,
+			'backend'          => "wikia{$hubDbName}-backend"
+		];
```
kvas-damian: Did you check what happen if you add foreign repo to the wiki that you're currently on? eg. render hub page

extensions/wikia/WikiaHubsServices/models/EditHubModel.class.php
```diff
@@ -269,6 +269,9 @@ protected function getDefaultModuleList() {
 		return $out;
 	}
 
+	/**
+	 * @return User
```
rogatty: This is a string: `protected $userClass = 'User';`

extensions/wikia/WikiaHubsServices/models/EditHubModel.class.php
```diff
@@ -299,6 +302,9 @@ public function getModuleUrl( $timestamp, $moduleId ) {
 		);
 	}
 
+	/**
+	 * @return SpecialPage
```
rogatty: This is a string too: `protected $specialPageClass = 'SpecialPage';`

extensions/wikia/WikiaHubsServices/models/EditHubModel.class.php
```diff
@@ -551,9 +557,11 @@ public function getCorporateWikisLanguages() {
 	public function getVideoData ($fileName, $thumbSize) {
 		$videoData = [];
 		$title = Title::newFromText($fileName, NS_FILE);
+
 		if (!empty($title)) {
-			$file = wffindFile($title);
+			$file = wfFindFile($title);
```
rogatty: Wow... What was happening when it got here?

extensions/wikia/WikiaHubsServices/modules/WikiaHubsModuleService.class.php
hakubo: It does not get there, as it'll find the file in local repo.

extensions/wikia/WikiaHubsServices/models/EditHubModel.class.php
hakubo: it... just... worked...

extensions/wikia/WikiaHubsServices/models/EditHubModel.class.php
hakubo: but on like 295 it is used as SpecialPage Class. wtf?

rogatty: http://stackoverflow.com/a/3121559/1050577

It's basically the same as `new $className()`

rogatty:
Fix types in comments, fix tests and it LGTM.

rogatty:
Tests are failing with:
```
Running 'HubRssModelTest'.....
Strict Standards: Declaration of WikiaHubsModuleSliderService::getImageInfo() should be compatible with WikiaHubsModuleService::getImageInfo($fileName, $destSize = 0) in /var/lib/jenkins/workspace/unitTests_php_branches/extensions/wikia/WikiaHubsServices/modules/WikiaHubsModuleSliderService.class.php on line 2
.
Catchable fatal error: Argument 1 passed to Mock_WikiaHubsModuleSliderService_4fc7e173::loadData() must be an instance of EditHubModel, null given, called in /var/lib/jenkins/workspace/unitTests_php_branches/extensions/wikia/HubRssFeed/models/external/HubRssFeedModel.class.php on line 127 and defined in /var/lib/jenkins/workspace/unitTests_php_branches/lib/composer/phpunit/phpunit-mock-objects/src/Framework/MockObject/Generator.php(313) : eval()'d code on line 11
```

so merging dev won't help :P

rogatty:
LGTM

bognix:
LGTM, tested, merging

PR #7394, opened 2015-06-02
CE-1897 Don't include flags on non-existent pages 

kamilkoterba:
plus naming refactoring
https://wikia-inc.atlassian.net/browse/CE-1897

@Wikia/community-engineering

### DIFF
--- a/extensions/wikia/Flags/FlagsHelper.class.php
+++ b/extensions/wikia/Flags/FlagsHelper.class.php
@@ -132,7 +132,7 @@ class FlagsHelper {
 	 * Checks if a request for flags does not come from an edit page
 	 * @return bool
 	 */
-	public function shouldDisplayFlags() {
+	public function shouldInjectFlags() {
 		global $wgTitle, $wgRequest;
 
 		return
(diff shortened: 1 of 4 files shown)


extensions/wikia/Flags/FlagsHelper.class.php
```diff
@@ -143,11 +143,25 @@ public function shouldDisplayFlags() {
...
-				[ 'edit', 'formedit' , 'history' ]
+				[ 'edit', 'formedit' , 'history', 'visualeditor' ]
 			);
 	}
 
 	/**
+	 * Checks if flags can be edited on current page to decide whether include edit modal
+	 * @return bool
+	 */
+	public function shouldDisplayFlags() {
```
lgarczewski: I would change this to "areFlagsEditable" and add one more condition:
* check for protection status using User::isAllowed('edit')

lgarczewski:
LGTM.

extensions/wikia/Flags/FlagsHelper.class.php
```diff
@@ -143,11 +144,39 @@ public function shouldDisplayFlags() {
...
+	 * @return bool
+	 */
+	public function areFlagsEditable() {
+		global $wgUser;
+		return
+			/* Check condition for view */
+			$this->shouldDisplayFlags()
+			/* Don't display flags when user is not allowed to edit */
+			&& $wgUser instanceof \User
+			&& $wgUser->isAllowed( 'edit' );
```
Grunny: I'd consider using `Title::userCan` instead of `User::isAllowed` as `isAllowed` just checks if the user has the edit right, not if they can edit the current page. `Title::userCan` checks if they can edit the current page including checking if it is protected and whether the user has the rights to edit pages at that protection level.

PR #7396, opened 2015-06-02
SOC-632 Wrap email extension messages to allow options for all of them

mklucsarits:
https://wikia-inc.atlassian.net/browse/SOC-632

This change will disable database look-up of messages for emails. This was only intended for emailext-\* message keys, so only those lines containing such messages have been modified.


extensions/wikia/Email/EmailController.class.php
```diff
@@ -644,4 +644,13 @@ private static function getLegendName() {
 
 		return $legendName;
 	}
+
+	/**
+	 * A wrapper for wfMessage() which removes possibility of database lookup for message text
```
jsutterfield: Can we update this comment to something more like `"A wrapper for wfMessage() which removes the possibility of the message being overridden in the MediaWiki namespace."`? That's what the intent of this method is, database lookup is a little vague for what we're actually doing.

mklucsarits: Sure, makes sense.

extensions/wikia/Email/EmailController.class.php
```diff
@@ -62,7 +62,7 @@ public function init() {
 			$this->test = $this->getVal( 'test', false );
 			$this->marketingFooter = $this->request->getBool( 'marketingFooter' );
 
-			$noReplyName = wfMessage( 'emailext-no-reply-name' )->escaped();
+			$noReplyName = $this->getMessage( 'emailext-no-reply-name' )->escaped();
```
garthwebb: This should probably have the ```inLanguage``` flag, so I don't think its an exception to my comment below.

extensions/wikia/Email/EmailController.class.php
```diff
@@ -644,4 +644,13 @@ private static function getLegendName() {
 
 		return $legendName;
 	}
+
+	/**
+	 * A wrapper for wfMessage() which removes the possibility of messages being overridden in the MediaWiki namespace
+	 *
+	 * @return Message
+	 */
+	protected function getMessage() {
```
garthwebb: Could we also add in the ```->inLanguage( $this->targetLang )``` part here as well?  I think that's always added.

extensions/wikia/Email/EmailController.class.php
```diff
@@ -644,4 +644,13 @@ private static function getLegendName() {
...
+		return wfMessage( func_get_args() )->useDatabase( false );
```
jsutterfield: @mklucsarits were you able to get this to work before? I didn't test this, but when I just tried merging it with my code none of the messages which have parameters are coming through.

I think the problem is that wfMessage always expects the first argument to be the message string. By calling `func_get_args()` and passing that result to wfMessage, the first and only argument now becomes an array which is causing problems in [here](https://github.com/Wikia/app/blob/dev/includes/GlobalFunctions.php#L1352-L1358).

mklucsarits: I tested it with the Forgot Password email and it worked. You might be right though. I also may have misread how func_get_args() could be used in the PHP manual. I'll investigate and fix if necessary.

jsutterfield:
LGTM!

extensions/wikia/Email/EmailController.class.php
```diff
@@ -644,4 +644,13 @@ private static function getLegendName() {
...
 		return $legendName;
 	}
+
+	/**
+	 * A wrapper for wfMessage() which removes the possibility of messages being overridden in the MediaWiki namespace
+	 *
+	 * @return Message
+	 */
+	protected function getMessage() {
+		return call_user_func_array( 'wfMessage', func_get_args() )->useDatabase( false )->inLanguage( $this->targetLang );
```
garthwebb: Could you break each arrow bit to a new line now that we have a few?

garthwebb:
LGTM

PR #7398, opened 2015-06-02
Monetization (MON-406): GlobalFooter - add config for crowdin and update massage

saipetch:
GlobalFooter (MON-406):
- add config for crowdin
- update shared-Oasis-footer-wikia-links message from Messaging

@hakubo


hakubo:
LGTM

PR #7401, opened 2015-06-03
CE-1911 Perform Flags edits as WikiaBot

adamkarminski:
@lukaszkonieczny


lukaszkonieczny:
lgtm

extensions/wikia/Flags/maintenance/MoveNotices.php
```diff
@@ -108,6 +108,12 @@ public function execute() {
 
 		$this->output( "Start processing\n" );
 
+		/**
+		 * Perform all edits as WikiaBot and overwrite wgUser so it is
+		 * available in the Flags logs
+		 */
+		$this->app->wg->User = $user = User::newFromName( 'WikiaBot' );
```
Grunny: `$user` is unused, why is it being set?

adamkarminski: Damn, an oversight. Apologies! :(

PR #7404, opened 2015-06-03
PLATFORM-1239: improve caching of DataProvider::GetTopFiveUser

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-1239
- use `WikiaDataAccess::cacheWithLock`
- set the maximum limit value and cache all results + slice them before returning
- stop when all required results are collected
- code cleanup

@michalroszka / @wladekb / @harnash


extensions/wikia/DataProvider/DataProvider.php
```diff
@@ -427,47 +429,78 @@ public static function GetNewlyCreatedArticles($limit = 5, $ns = array(NS_MAIN))
...
+		$limit = min( $limit, self::TOP_USERS_MAX_LIMIT );
 
 		if (empty($wgStatsDB) || empty($wgStatsDBEnabled)) {
-			$result = array();
+			$results = [];
 		}
 		else {
-			$memckey = wfMemcKey("TopFiveUsers", $limit);
-			$results = $wgMemc->get($memckey);
+			$results = WikiaDataAccess::cacheWithLock( wfMemcKey( __CLASS__, 'GetTopFiveUsers' ), WikiaResponse::CACHE_STANDARD, function() use ($fname) {
```
wladekb: AFAIR we used to put __METHOD__ into wfMemcKey in such cases

extensions/wikia/DataProvider/DataProvider.php
```diff
@@ -427,47 +429,78 @@ public static function GetNewlyCreatedArticles($limit = 5, $ns = array(NS_MAIN))
...
+					"user_groups",
+					"ug_user",
+					[ 'ug_group' => [ 'staff', 'bot' ] ],
+					$fname
+				);
+
+				// add more blacklisted user IDs
+				$users_list[] = 0;
+				$users_list[] = 22439; // Wikia
+				$users_list[] = 929702; // CreateWiki script
```
wladekb: Good candidate to put the list of such IDs into single common place like config variable and use it here

extensions/wikia/DataProvider/DataProvider.php
macbre: We do, but this call is within a closure.

extensions/wikia/DataProvider/DataProvider.php
```diff
@@ -427,47 +429,78 @@ public static function GetNewlyCreatedArticles($limit = 5, $ns = array(NS_MAIN))
...
-
-				$res = $dbs->query($dbs->limitResult($query, $limit * 4, 0));
-
-				$results = array();
+				$users_list = $dbr->selectFieldValues(
+					"user_groups",
+					"ug_user",
+					[ 'ug_group' => [ 'staff', 'bot' ] ],
+					$fname
+				);
```
wladekb: use DISTINCT here?

extensions/wikia/DataProvider/DataProvider.php
macbre: Isn't it a bit easier to grep this key name when it's provided as a raw text?

extensions/wikia/DataProvider/DataProvider.php
macbre: Good point, but I'm afraid that we will end up with completely rewritten code here :)

extensions/wikia/DataProvider/DataProvider.php
macbre: Improved in 0578e68

michalroszka:
:+1:

PR #7407, opened 2015-06-03
(CE-1867) Make Browse QueryPage use SMW cluster

Grunny:
This was missing from the previous version too, but we need to override
`QueryPage::reallyDoQuery` so we can set the DB connection to point
to the SMW cluster. This was probably related to the data issues Naruto
has been having with the old version of the extension.


kamilkoterba:
After some discussion you convinced me to this hack. I say LGTM then.

PR #7408, opened 2015-06-03
CONCF-558 move css rule to file which is used on breakpoints

bognix:
https://wikia-inc.atlassian.net/browse/CONCF-558

ping @rogatty @hakubo 

@gabrys - this rule is used to override rule which is added here: https://github.com/Wikia/app/blob/dev/skins/oasis/css/core/page.scss#L87-L90
Maybe we don't need this rule from page.scss in the first place?


rogatty:
The code LGTM but I have no idea if it can break something.

bognix:
@rogatty - it shouldn't after second commit. Now we only target responsive and breakpoints based versions of oasis.

gabrys:
We still give the german wikis oasis-fluid treatment. And they have the 
prefooter ads (one of them in fact). The anticipated date of converting 
them to breakpoints is 1st July according to Esteban.

bognix:
@gabrys - I know, that's why this rule is applied to both old responsive and new breakpoints based layouts.

rogatty:
@gabrys is comment from @bognix satisfying for you? Does it look good to you?

nandy-andy:
I'm think if it would be a good practice to add a follow-up ticket to clean this code after we release Oasis breakpoints on German wikias and `//TODO` comments with the ticket number here. Thanks to that we'd gain even more red numbers in next PR connected to the follow-up ticket :smile_cat: What do you think about it? :japanese_ogre:

hakubo:
LGTM

bognix:
@nandy-andy - added comments as you suggested. Although best idea would be to remove this rule: https://github.com/Wikia/app/blob/dev/skins/oasis/css/core/page.scss#L87-L90

do you know if it still needed? from what I know we don't display prefooters on wikia.com and subpages and this is the only page where grid is enabled.

nandy-andy:
@bognix those lines were added in 2010 from what I can check. It was before I worked here ;) I removed them in my devbox and only spacing around pre-footer ads changed - and does not look bad. I'd remove them if it causes troubles or the code is less cleaner with them :japanese_ogre:

bognix:
@nandy-andy I created a ticket to take care of wikia grid skin. If it is solved then those lines won't matter anymore and will be removed.

nandy-andy:
Thanks for taking care of it! :smile_cat:

PR #7411, opened 2015-06-03
SOC-459 Whitespace/stylize updates

garthwebb:
@jsutterfield could you please confirm this stylize.php only PR and merge when you are satisfied?  There will be another dedicated branch and PR for SOC-459 once this is done.


jsutterfield:
LGTM

PR #7412, opened 2015-06-03
ADEN-2071 FLOOR ADHESION - Desktop floor asset not centered on page correctly

gabrys:


nandy-andy:
Well, it's CSS... and red wins... LGTM :+1:

PR #7417, opened 2015-06-05
Ce 1832 flags extractor unittests

kamilkoterba:
https://wikia-inc.atlassian.net/browse/CE-1832

Added tests for FlagsExtractor 
covered init, getText, findFirstTemplateFromList, isTagAdded, getAllTemplates and getTemplate public methods (all)

@Wikia/community-engineering

### DIFF
--- a/extensions/wikia/Flags/FlagsExtractor.class.php
+++ b/extensions/wikia/Flags/FlagsExtractor.class.php
@@ -553,7 +553,7 @@ class FlagsExtractor {
 	 * @param string $message text
 	 * @param array $params additional parameters
 	 */
-	private function logInfoMessage( $message, $params = [] ) {
+	protected function logInfoMessage( $message, $params = [] ) {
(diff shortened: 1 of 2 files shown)


extensions/wikia/Flags/FlagsExtractor.class.php
```diff
@@ -553,7 +553,7 @@ private function getTemplateParam( $isParamWithName, $paramsCounter, $paramOffse
 	 * @param string $message text
 	 * @param array $params additional parameters
 	 */
-	private function logInfoMessage( $message, $params = [] ) {
+	protected function logInfoMessage( $message, $params = [] ) {
```
Grunny: I think it's better not to be changing the visibility of a private method just to mock it as this is an implementation detail. I'd just mock `info` from the Loggable trait which would be used even if you changed this method and is public and leave this private.

PR #7418, opened 2015-06-05
MAIN-4460 User wsEditToken logging

ghost:
Log the following:
- whenever a new wsEditToken is generated
- whenever wsEditToken mismatch is detected

Initially, I only log 10% of cases, as I do not know how much data
would that change push to ELK. The sampling rate will be changed.


macbre:
:+1:

michalroszka:
@grunny - pinging you since you know the context; just so you are aware of the changes. And once again thank you for helpful tips.

PR #7421, opened 2015-06-05
Implement MemcacheUserBlock for short, per-wiki user blocks

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-1055

As a part of Uncyclopedia migration we need to block user that is currently migrated for a very short time (up to 10 seconds). We can't use DB-based blocks as the migration is performed within a transaction.

**The solution**: use memcache-based, per-wiki blocks.

The API:

``` php
$blocker = new MemcacheUserBlock( $wgUser );

// apply a block for 60 seconds, memcache key will expire in 60 seconds
$blocker->setBlock('The block reason goes here', 60);

// check the block
$blocker->getBlock();

// remove a block
$blocker->removeBlock();

// memcached: MemCache: delete uncyclo:MemcacheUserBlock:119245
```
- this class binds to the MW via `UserIsBlockedFrom` hook.
- migration script will call `setBlock` and `removeBlock` methods

@michalroszka / @wladekb

### DIFF
--- a/extensions/wikia/MemcacheUserBlock/MemcacheUserBlock.class.php
+++ b/extensions/wikia/MemcacheUserBlock/MemcacheUserBlock.class.php
@@ -23,7 +23,8 @@ class MemcacheUserBlock {
 			$this->getMemcacheBlockKey(),
 			[
 				'expiry' => time() + $ttl,
-				'reason' => $reason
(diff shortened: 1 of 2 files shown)


michalroszka:
Seems legit.

PR #7423, opened 2015-06-05
CE-1904 Done button in primary color

kamilkoterba:
Done button in primary color instead of secondary 
plus restoring removed preventDefault

https://wikia-inc.atlassian.net/browse/CE-1904

@Wikia/community-engineering


Grunny:
:+1:

PR #7424, opened 2015-06-05
CE-1759 Move images back to the ImageReview queue

adamkarminski:
@see [CE-1759](https://wikia-inc.atlassian.net/browse/CE-1759)

Current behavior of once deleted images is not acceptable! If they are restored, they don't go back to the ImageReview queue! You can just add an [ugly picture](http://static.idolator.com/uploads/2013/09/09/miley-cyrus-wrecking-ball-video-600x337.jpg), get it deleted, restore it and still enjoy it being displayed. Let's put an end to this outrage!

We detect if an image has ImageReview records. If it does not - build a Scribe package and send it back where it belongs!

### DIFF
--- a/extensions/wikia/ImageReview/ImageReviewImageStatus.php
+++ b/extensions/wikia/ImageReview/ImageReviewImageStatus.php
@@ -8,7 +8,7 @@
 $wgHooks['ImagePageAfterImageLinks'][] = 'efImageReviewDisplayStatus';
 $wgGroupPermissions['staff']['imagereviewstats'] = true;
 
-function efImageReviewDisplayStatus( $imagePage, &$html ) {
+function efImageReviewDisplayStatus( ImagePage $imagePage, &$html ) {
 	global $wgCityId, $wgExternalDatawareDB, $wgUser;
 
 	if ( !$wgUser->isAllowed( 'imagereviewstats' ) ) {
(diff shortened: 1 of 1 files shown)


extensions/wikia/ImageReview/ImageReviewImageStatus.php
```diff
@@ -50,8 +50,37 @@ function efImageReviewDisplayStatus( $imagePage, &$html ) {
...
+			 * If the file is older than 1 hour - send it to ImageReview
+			 * since it's probably been restored, and is not just a fresh file.
+			 */
+			$lastTouched = new DateTime( $imagePage->getRevisionFetched()->getTimestamp() );
+			$now = new DateTime();
+			if ( $lastTouched < $now->modify( '-1 hour' ) ) {
+				$scribeEventProducer = new ScribeEventProducer( 'edit' );
+				$user = $imagePage->getContext()->getUser();
+				$file = $imagePage->getDisplayedFile();
+				if ( $scribeEventProducer->buildEditPackage( $imagePage, $user, null, null, $file ) ) {
```
lgarczewski: Do we know if this will count as an edit in DataMart? It shouldn't.

extensions/wikia/ImageReview/ImageReviewImageStatus.php
```diff
@@ -50,8 +50,37 @@ function efImageReviewDisplayStatus( $imagePage, &$html ) {
...
 		if ( false === $imgCurState ) {
+			/**
+			 * If the file is older than 1 hour - send it to ImageReview
+			 * since it's probably been restored, and is not just a fresh file.
+			 */
+			$lastTouched = new DateTime( $imagePage->getRevisionFetched()->getTimestamp() );
+			$now = new DateTime();
+			if ( $lastTouched < $now->modify( '-1 hour' ) ) {
+				$scribeEventProducer = new ScribeEventProducer( 'edit' );
+				$user = $imagePage->getContext()->getUser();
```
Grunny: This is from the RequestContext, so would be the user viewing the image page, not the user who uploaded it, right? Should we instead get the user who uploaded the file to insert into the ImageReview table?

adamkarminski: Yeah, I've just noticed it when testing the solution. Thanks! ;)

lgarczewski:
LGTM.

PR #7425, opened 2015-06-05
ADEN-2044 Prepare Amazon mobile rollout

nandy-andy:
Before we rollout Amazon on mobile devices we want to have [automation tests for it](https://github.com/Wikia/selenium-tests/pull/665) ready and we want a mechanism which doesn't display Amazon ad once it was displayed in Mercury.

Changes below are supposed to provide such mechanism.

### DIFF
--- a/extensions/wikia/AdEngine/js/GptHelper.js
+++ b/extensions/wikia/AdEngine/js/GptHelper.js
@@ -73,6 +73,23 @@ define('ext.wikia.adEngine.gptHelper', [
 		}
 	}
 
+	function setSlotLevelParams(slot, slotTargeting) {
+		var name,
+			value;
+
+		log(['setSlotLevelParams', slotTargeting], 'debug', logGroup);
+
+		for (name in slotTargeting) {
+			if (slotTargeting.hasOwnProperty(name)) {
+				value = slotTargeting[name];
+				if (value) {
+					log(['defineSlot', 'slot.setTargeting', name, value], 'debug', logGroup);
+					slot.setTargeting(name, value);
+				}
+			}
+		}
+	}
+
 	function registerGptCallback(adDivId, gptCallback) {
 		log(['registerGptCallback', adDivId], 'info', logGroup);
 		gptCallbacks[adDivId] = gptCallback;
@@ -180,9 +197,7 @@ define('ext.wikia.adEngine.gptHelper', [
 		}
 
 		function queueAd() {
-			var name,
-				value,
-				sizes,
+			var sizes,
 				slot,
 				pageLevelParams = adLogicPageParams.getPageLevelParams();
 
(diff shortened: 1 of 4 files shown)


extensions/wikia/AdEngine/js/GptHelper.js
```diff
@@ -73,6 +73,23 @@ define('ext.wikia.adEngine.gptHelper', [
 		}
 	}
 
+	function setSlotLevelParams(slot, slotTargeting) {
+		var name,
+			value;
+
+		log(['setSlotLevelParams', slotTargeting], 'debug', logGroup);
+
+		for (name in slotTargeting) {
```
gabrys: Let's start with clearTargeting so we don't need to clear the amznslots manually https://developers.google.com/doubleclick-gpt/reference#googletag.Slot_clearTargeting

extensions/wikia/AdEngine/js/GptHelper.js
```diff
@@ -211,16 +226,6 @@ define('ext.wikia.adEngine.gptHelper', [
 
 				delete slotTargeting.size;
```
gabrys: delete size should go outside the if I think

extensions/wikia/AdEngine/js/lookup/amazonMatch.js
```diff
@@ -181,7 +203,8 @@ define('ext.wikia.adEngine.lookup.amazonMatch', [
 		trackState: function () {
 			log('fake trackState - module is not supported in IE8', 'debug', logGroup);
 		},
-		wasCalled: wasCalled
+		wasCalled: wasCalled,
+		wasRendered: wasRendered
```
gabrys: do we need this exported by the module?

nandy-andy: Nope, not at all. At one point I was checking it outside and forgot to remove it from here. Good catch, thanks! :smile_cat:

extensions/wikia/AdEngine/js/lookup/amazonMatch.js
```diff
@@ -145,21 +147,30 @@ define('ext.wikia.adEngine.lookup.amazonMatch', [
...
-		});
+				if (validSlotNames.indexOf(slotName) !== -1 && amazonPricePoint) {
+					amznSlots.push('a' + amazonSize + 'p' + amazonPricePoint);
+				}
+			});
 
-		log(['getSlotParams - amznSlots: ', amznSlots], 'debug', logGroup);
+			log(['getSlotParams - amznSlots: ', amznSlots], 'debug', logGroup);
+		} else {
+			log(['getSlotParams - no amznSlots since ads has been already displayed', slotName], 'debug', logGroup);
```
fraszczakszymon: I wonder if this else statement is needed since we have __wasRendered__ method with log, or maybe this method is useless because it's not used outside the module anymore?

fraszczakszymon:
LGTM :+1:

PR #7427, opened 2015-06-05
SOC-877 Add a new group for translators so they can access the email tool

garthwebb:
https://wikia-inc.atlassian.net/browse/SOC-877


extensions/wikia/Email/EmailHelper.class.php
```diff
@@ -0,0 +1,17 @@
...
+namespace Email;
+/**
+ * Class Helper
+ *
+ * Common methods used between controllers and classes
+ *
+ * @package Email
+ */
+class Helper {
+	const REQUIRED_USER_RIGHT = 'access-sendemail';
```
mklucsarits: Can either the SpecialSendEmailController use the constant from this class, or vice-versa, instead of having it in two places?

jsutterfield: +1 to that

garthwebb: Sorry, that was my intention, I didn't dot all the I's ...

mklucsarits:
LGTM

jsutterfield:
LGTM

PR #7428, opened 2015-06-05
Make sure we don't have forms which point to abstract classes

jsutterfield:
Special:SendEmail creates a list of forms which users can use to send off our emails. These forms are created programmatically by finding classes in our Email extension and then pointing those forms to those classes. We also have abstract classes in that extension and need to make sure we don't create forms for those classes.

### DIFF
--- a/extensions/wikia/Email/Controller/WatchedPageController.class.php
+++ b/extensions/wikia/Email/Controller/WatchedPageController.class.php
@@ -173,6 +173,33 @@ abstract class WatchedPageController extends EmailController {
 			$this->getAllChangesText( $this->title ),
 		];
 	}
+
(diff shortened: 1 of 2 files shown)


extensions/wikia/Email/Controller/WatchedPageController.class.php
```diff
@@ -310,31 +337,4 @@ protected function getArticleLinkText() {
 	protected function getAllChangesText( $title ) {
 		return parent::getAllChangesText( $this->newTitle );
 	}
-
-	protected static function getEmailSpecificFormFields() {
```
jsutterfield: This need to get moved into the abstract class since it's code needed by all child classes.

mklucsarits:
LGTM

PR #7440, opened 2015-06-08
PLATFORM-1261: Helios - start the session when initializing the user from the access token

macbre:
@michalroszka / @ArturKlajnerok 
- `access_token` cookie TTL is set to 6 months
- `wikicities_session` is set for browser session only

Hence, the user can stay logged-in even when he's session cookie expires.

### DIFF
--- a/extensions/wikia/Helios/User.class.php
+++ b/extensions/wikia/Helios/User.class.php
@@ -82,6 +82,14 @@ class User {
 			global $wgHeliosBaseUri, $wgHeliosClientId, $wgHeliosClientSecret;
 			$heliosClient = new Client( $wgHeliosBaseUri, $wgHeliosClientId, $wgHeliosClientSecret );
 
+			// start the session if there's none so far
+			// the code is borrowed from SpecialUserlogin
+			// @see PLATFORM-1261
+			if ( session_id() == '' ) {
+				wfSetupSession();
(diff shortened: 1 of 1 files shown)


ArturKlajnerok:
Lgtm :+1:

extensions/wikia/Helios/User.class.php
```diff
@@ -82,6 +82,14 @@ public static function newFromToken( \WebRequest $request ) {
 			global $wgHeliosBaseUri, $wgHeliosClientId, $wgHeliosClientSecret;
 			$heliosClient = new Client( $wgHeliosBaseUri, $wgHeliosClientId, $wgHeliosClientSecret );
 
+			// start the session if there's none so far
+			// the code is borrowed from SpecialUserlogin
+			// @see PLATFORM-1261
+			if ( session_id() == '' ) {
+				wfSetupSession();
+				WikiaLogger::instance()->debug( __METHOD__ . '::startSession' );
+			}
```
drsnyder: I know this was done to resolve the P2 but I don't think this is the right place for this logic. This function in it's current form is difficult to test and now we are in a situation where we are resolving a P2 with no test coverage to ensure that it doesn't happen in the future.

From the outside looking in, this violates the principle of least surprise. When I call `newFromId` I don't expect it to test the current session or set it up. I would only expect this function to do one thing which is give me a user object based on a token.

PR #7443, opened 2015-06-08
CE-1907 + CE-1818 Set up a special page for Flags

adamkarminski:


extensions/wikia/Flags/models/FlagType.class.php
```diff
@@ -33,6 +33,22 @@ class FlagType extends FlagsBaseModel {
 	];
 
 	/**
+	 * The first var was created for the use of the converting maintenance script. This one
+	 * is for user-facing purposes.
+	 * @var array
+	 */
+	public static $flagGroupsFullNames = [
```
adamkarminski: A note to myself - groups should be i18nable :)

lgarczewski:
LGTM except for the i18n as noted above.

PR #7444, opened 2015-06-08
CE-1862 Skip a row if a Title created from it is null

adamkarminski:
**see [CE-1862](https://wikia-inc.atlassian.net/browse/CE-1862)**

@Wikia/community-engineering


Grunny:
LGTM.

PR #7445, opened 2015-06-08
CE-1930 Check if the sorting array exists in cache before updating it

adamkarminski:
see [CE-1930](https://wikia-inc.atlassian.net/browse/CE-1930)

@Wikia/community-engineering


Grunny:
LGTM.

PR #7448, opened 2015-06-08
Monetization (MON-367): MonetizationModule - add functions to add in-content ad and show below category ad

saipetch:
Monetization (MON-367):
- add functions to add in-content ad and show below category
- add event tracking

### DIFF
--- a/extensions/wikia/AdEngine/js/MonetizationServiceHelper.js
+++ b/extensions/wikia/AdEngine/js/MonetizationServiceHelper.js
@@ -38,11 +38,8 @@ define('ext.wikia.adEngine.monetizationsServiceHelper', [
 			num = $(elementName).length,
 			content = '<div id="' + slot + '" class="wikia-ad noprint default-height"></div>';
 
-		// TOC exists. Insert the ad above the 3rd <H2> tag.
-		if ($('#toc').length > 0 && num > 3) {
-			$(elementName).eq(2).before(content);
-		// TOC not exist. Insert the ad above the 2nd <H2> tag.
-		} else if (num > 2) {
+		// Insert the ad above the 2nd <H2> tag. (Insert the ad above the 3rd <H2> tag if TOC exists)
+		if (num >= 2) {
 			$(elementName).eq(1).before(content);
 		// Otherwise, insert to the end of content
 		} else {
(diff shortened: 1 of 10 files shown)


extensions/wikia/AdEngine/js/MonetizationServiceHelper.js
```diff
@@ -0,0 +1,130 @@
...
+			var $container = $('#' + slot).find('.monetization-module');
+
+			trackImpression($container);
+			trackClickEvent($container);
+
+			success();
+		});
+	}
+
+	function addInContentSlot(slot) {
```
gabrys: I would rather have this slot added in extension/wikia/AdEngine/js/slot/monetizationInContent.js just like other dynamically added slots are handled.

extensions/wikia/AdEngine/js/MonetizationServiceHelper.js
```diff
@@ -0,0 +1,130 @@
...
+		// Otherwise, insert to the end of content
+		} else {
+			$('#mw-content-text').append(content);
+			isEndOfContent = true;
+		}
+
+		window.adslots2.push(slot);
+	}
+
+	function trackImpression($container) {
```
gabrys: For tracking we could merge this with GA ad tracking if that makes sense (it causes 1/100 sampling and is GA only, so maybe it's not what we want for Monetization Service?)

extensions/wikia/AdEngine/js/MonetizationServiceHelper.js
```diff
@@ -0,0 +1,130 @@
...
+			trackClickEvent($container);
+
+			success();
+		});
+	}
+
+	function addInContentSlot(slot) {
+		var elementName = '#mw-content-text > h2',
+			num = $(elementName).length,
+			content = '<div id="' + slot + '" class="wikia-ad noprint default-height"></div>';
```
gabrys: Since we're using jQuery anyway, it's better to set the id through $(el).attr to avoid HTML injection

extensions/wikia/AdEngine/js/MonetizationServiceHelper.js
saipetch: We would like to track 100% for Monetization Service.

saipetch:
Merged to MON-375 branch

PR #7449, opened 2015-06-09
PLATFORM-1261: log real edit token mismatches

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-1261

Remove logs added in #7419 and only log in case of a "real" edit token mismatch (i.e. when POST and session data do not match and POST field is not empty).

@michalroszka


michalroszka:
:+1:

PR #7453, opened 2015-06-09
(CE-1928) Fix sending error emails

Grunny:
UserMailer::send doesn't accept an array of addresses after:
https://github.com/Wikia/app/commit/b34bfb51

```
PHP Catchable fatal error: Argument 1 passed to UserMailer::send()
must be an instance of MailAddress, array given, called in
extensions/wikia/ImageReview/ImageReviewTask.class.php on line 122
and defined in includes/UserMailer.php on line 165
```

Avoiding changing UserMailer for this quick fix as I'm not sure if it was
intentional (/cc @garthwebb)


lgarczewski:
LGTM.

PR #7457, opened 2015-06-09
Started caching the 404s that the Lyrics mobile app's API sends. This…

SeanColombo:
… is because of the special nature of Lyrics mobile apps that popular songs that we don't have pages for yet, will be requested hundreds (thousands?) of times per hour.  The caching is a very short duration (5 minutes) which will just unburden the backend (max of 12 requests per hour per 404).  LYR-235

I'm not sure if the preference is there for a CACHE_VERY_SHORT to exist, or if that should be a local var (like in ChatRailController).  I'm considering making another branch & just letting Macbre choose which one he'd rather merge ;)

### DIFF
--- a/extensions/wikia/LyricsApi/LyricsApiController.class.php
+++ b/extensions/wikia/LyricsApi/LyricsApiController.class.php
@@ -2,7 +2,9 @@
 /**
  * Class LyricsApiController
  *
- * @desc Entry point for LyricsAPI
+ * @desc Entry point for LyricsAPI used by the Lyrically app. For the LyricWiki API that runs
+ *       through the standard MediaWiki and is accessible to the public, please use the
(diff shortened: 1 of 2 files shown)


SeanColombo:
Here is a curl to confirm that it's working:
curl -I -x dev-sean:80 "http://lyrics.sean.wikia-dev.com/wik
ia.php?controller=LyricsApi&method=getSong&artist=Jannat&song=%D8%AD%D8%A8+%D8%A
C%D8%A7%D9%85%D8%AF"

SeanColombo:
IMPORTANT: IF YOU'D PREFER THE SHORTER CACHE DURATION TO BE A LOCAL CONST INSTEAD OF ANOTHER WikiaResponse OPTION, THEN MERGE THIS PULL REQUEST INSTEAD: https://github.com/Wikia/app/pull/7458

macbre:
This one looks a bit better than #7458 :)

PR #7458, opened 2015-06-09
Started caching 404s from the lyrically mobile app's API because of t…

SeanColombo:
…he special nature of Lyrics mobile apps that popular songs that we don't have pages for yet, will be requested hundreds (thousands?) of times per hour. The caching is a very short duration (5 minutes) which will just unburden the backend (max of 12 requests per hour per 404). LYR-235

This is the same fix as https://github.com/Wikia/app/pull/7457 ...just another option for how it's coded. PLEASE MERGE JUST ONE OF THEM... WHICHEVER IS PREFERRED :)


SeanColombo:
Merged 7457 instead.

PR #7460, opened 2015-06-09
Ce 1869 flags tracking

kamilkoterba:
https://wikia-inc.atlassian.net/browse/CE-1869

Added clicktracking for flags edit modal for following events:

An impression when the dialog is opened [DONE]
When the dialog is closed and no changes are saved ("x", "cancel," clicking outside the dialog) [DONE]
When "done" is clicked but no changes are actually made
When the dialog is closed and some changes are saved [DONE] note: checking only if form was touched
so if user checks checkbox and unchecks back before save it's stored as changed
When the dialog is submitted [DONE]

When "Flags" is selected from the edit dropdown (similar to how other buttons there are tracked) [DONE]
When a flag is toggled on [DONE]
When a flag is toggled off [DONE]
When "More info -->" is clicked [DONE]

@Wikia/community-engineering


lgarczewski:
LGTM.

PR #7461, opened 2015-06-10
Prepare an A/B Performance Measurement Framework

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-1246

This is a proposal for the micro-framework that will handle performance A/B testing.
### Criteria

Performance experiments can now be enabled based on two criteria:
-  `wikis`: enable on a fraction of wikis (each bucket is 0.1% of all wikis - uses `$wgCityId`)
- `traffic`: enable site-wide on a fraction of traffic (each bucket is 0.1% of all clients - uses `beacon_id`)
- `oasisArticles`: if provided with the value of `true` will cause the experiment to be run on Oasis content namespace articles only

Only up to one experiment can be run for the same request. The first match will be used. The name of the selected experiment will be report to GA (via custom variable) and to InfluxDB (via Transaction parameter - for both backend and frontend related metrics).

``` js
wgTransactionContext.perf_test
"backend_delay_a"
```
### Experiments config

All performance experiments are defined in `$wgABPerformanceTestingExperiments` global array with a single entry as follows:

``` php
$wgABPerformanceTestingExperiments['backend_delay_b'] = [
    'handler' => 'Wikia\\AbPerfTesting\\Experiments\\BackendDelay',
    'params' => [
        'delay' => 50,
    ],
    'criteria' => [
        'wikis' => 2,
    ]
];
```
- `handler` is a class name that will be instantiated when the experiment criteria are matched (`params` array will be passed as constructor arguments).
- `criteria`: criteria to be met
### Experiments implemented
- `BackendDelay` - add a small delay at the end of the request.
- `FrontDelay` - add a small delay before onDOMReady events is handled by jQuery (using a [docReady](https://github.com/jfriend00/docReady) and a small helper JS script - less than 1k of gzipped bytes)

@wladekb / @jcellary / @harnash


extensions/wikia/AbPerfTesting/classes/Hooks.class.php
```diff
@@ -0,0 +1,49 @@
...
+				$reflector->newInstanceArgs( $experiment['params'] ? : [] );
+
+				// mark a transaction with an experiment name
+				\Transaction::getInstance()->set( \Transaction::PARAM_AB_PERFORMANCE_TEST, $name );
+
+				// mark a transaction using UA's custom dimensions
+				global $wgHooks;
+				$wgHooks['WikiaSkinTopScripts'][] = function( Array &$vars, &$scripts ) use ( $name ) {
+					$name = \Xml::encodeJsVar( $name );
+					$scripts .= \Html::inlineScript( "_gaq.push(['_setCustomVar', 50, 'PerfTest', {$name}, 3]);" );
```
macbre: Pinging @harnash

extensions/wikia/AbPerfTesting/js/FrontendDelay.js
```diff
@@ -0,0 +1,81 @@
...
+		weppy.setOptions({
+			context: w.wgTransactionContext
+		});
+
+		// set UA custom dimension
+		// should match the PHP code in /extensions/wikia/AbPerfTesting/classes/Hooks.class.php
+		if (perf_test) {
+			w._gaq.push(
+				['_setCustomVar', 50, 'PerfTest', perf_test, 3]
+			);
```
macbre: Pinging @harnash

extensions/wikia/AbPerfTesting/classes/Hooks.class.php
harnash: https://github.com/Wikia/app/blob/dev/extensions/wikia/AnalyticsEngine/js/universal_analytics.js#L208-L212

macbre: Used `dimension20` custom variable.

extensions/wikia/AbPerfTesting/classes/Hooks.class.php
```diff
@@ -0,0 +1,49 @@
...
+				$reflector->newInstanceArgs( $experiment['params'] ? : [] );
+
+				// mark a transaction with an experiment name
+				\Transaction::getInstance()->set( \Transaction::PARAM_AB_PERFORMANCE_TEST, $name );
+
+				// mark a transaction using UA's custom dimensions
+				global $wgHooks;
+				$wgHooks['WikiaSkinTopScripts'][] = function( Array &$vars, &$scripts ) use ( $name ) {
+					$name = \Xml::encodeJsVar( $name );
+					$scripts .= \Html::inlineScript( "_gaq.push(['set', 'dimension20', {$name}]);" );
```
macbre: @harnash

extensions/wikia/AbPerfTesting/js/FrontendDelay.js
```diff
@@ -0,0 +1,81 @@
...
+
+		weppy.setOptions({
+			context: w.wgTransactionContext
+		});
+
+		// set UA custom dimension
+		// should match the PHP code in /extensions/wikia/AbPerfTesting/classes/Hooks.class.php
+		if (perf_test) {
+			w._gaq.push(
+				['set', 'dimension20', perf_test]
```
macbre: @harnash

extensions/wikia/AbPerfTesting/AbPerfTesting.setup.php
```diff
@@ -0,0 +1,124 @@
...
+$wgExtensionCredits['other'][] = [
+	'name'   => 'A/B Performance Tests',
+	'url'    => 'https://github.com/Wikia/app/tree/dev/extensions/wikia/AbPerfTesting',
+	'author' => [
+		'[http://community.wikia.com/wiki/User:Macbre Maciej Brencz]'
+	],
+];
+
+// generic classes
+$wgAutoloadClasses[ 'Wikia\\AbPerfTesting\\Hooks'                     ] = __DIR__ . '/classes/Hooks.class.php';
```
wladekb: I'd prefer having it namedl "AbPerformanceTests" instead of shortcut-ish "AbPerfTesting" (in all places, not only the namespace but also the master global variable etc.). I'm still open for discussion here.

extensions/wikia/AbPerfTesting/AbPerfTesting.setup.php
```diff
@@ -0,0 +1,124 @@
...
+ * Devbox debugging of traffic criterion:
+ *
+ * 1. Set the beacon cookie: document.cookie = 'wikia_beacon_id=3j-YqSr9BQ'
+ * 2. Set the "traffic" criterion to '158'
+**/
+
+$wgABPerfTestingExperiments['backend_delay_0'] = [
+	'handler' => 'Wikia\\AbPerfTesting\\Experiments\\BackendDelay',
+	'params' => [
+		'delay' => 0, // control group to compare other three groups against
```
wladekb: delay => backend_delay ?

wladekb: nvm, handler defines it

extensions/wikia/AbPerfTesting/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Criteria
+
+Performance experiments can be enabled based on the following criteria:
+
+*  `wikis`: enable on a fraction of wikis (each bucket is 0.1% of all wikis - uses `$wgCityId`)
+* `traffic`: enable site-wide on a fraction of traffic (each bucket is 0.1% of all clients - uses `beacon_id`)
+* `oasisArticles`: if provided with the value of `true` will cause the experiment to be run on Oasis content namespace articles only
+
+Only up to one experiment can be run for the same request. The first match will be used. The name of the selected experiment will be report to Universal Analytics (via custom variable) and to InfluxDB (via Transaction parameter - for both backend and frontend related metrics).
```
wladekb: report => reported

extensions/wikia/AbPerfTesting/README.md
```diff
@@ -0,0 +1,43 @@
...
+];
+```
+
+* `handler` is a class name that will be instantiated when the experiment criteria are matched (`params` array will be passed as constructor arguments).
+* `criteria`: criteria to be met
+
+### Experiments implemented
+
+* `BackendDelay` - add a small delay at the end of the request.
+* `FrontDelay` - add a small delay before onDOMReady events is handled by jQuery
```
wladekb: FrontDelay => FrontendDelay or maybe FrontendInteractiveDelay?

extensions/wikia/AbPerfTesting/classes/Hooks.class.php
```diff
@@ -0,0 +1,49 @@
...
+		global $wgABPerfTestingExperiments;
+		wfDebug( sprintf( "%s - checking experiments (with beacon ID set to '%s')...\n", __METHOD__, wfGetBeaconId() ) );
+
+		// loop through all registered experiments and run those matching criteria
+		foreach ( $wgABPerfTestingExperiments as $name => $experiment ) {
+			if ( Experiment::isEnabled( $experiment ) ) {
+				wfDebug( sprintf( "%s: starting '%s' experiment using %s class with %s params\n",
+					__METHOD__, $name, $experiment['handler'], json_encode( $experiment['params'] ) ) );
+
+				$reflector = new \ReflectionClass( $experiment['handler'] );
```
wladekb: This should also work and will not use Reflection mechanism (but please test it whether it works or not if you decide to switch to this version):
```
$handlerClass = $experiment['handler'];
$handlerObject = new $handlerClass( $experiment['params'] ? : [] );
```

extensions/wikia/AbPerfTesting/classes/Hooks.class.php
wladekb: @macbre please double-check that _gaq will always be defined at this point

extensions/wikia/AbPerfTesting/criteria/OasisArticles.class.php
```diff
@@ -0,0 +1,30 @@
...
+
+	function __construct() {
+		$this->mContext = \RequestContext::getMain();
+	}
+
+	/**
+	 * @param bool $bucket
+	 * @return boolean
+	 */
+	function applies( $bucket ) {
```
wladekb: function matches/test/check/... ?

extensions/wikia/AbPerfTesting/classes/Hooks.class.php
macbre: It's defined in a global variables section in `<head>`

extensions/wikia/AbPerfTesting/classes/Hooks.class.php
macbre: Nope, http://pl1.php.net/manual/en/reflectionclass.newinstanceargs.php - this simplifies the structure of experiments constructor by passing each entry from `params` as a separate argument.

extensions/wikia/AbPerfTesting/README.md
macbre: The class is called `FrontendDelay`. Docs typo. Good catch!

extensions/wikia/AbPerfTesting/js/FrontendDelay.js
```diff
@@ -0,0 +1,81 @@
...
+		w.wgTransactionContext['perf_test'] = perf_test;
+
+		weppy.setOptions({
+			context: w.wgTransactionContext
+		});
+
+		// set UA custom dimension
+		// should match the PHP code in /extensions/wikia/AbPerfTesting/classes/Hooks.class.php
+		if (perf_test) {
+			w._gaq.push(
```
wladekb: In backend tests we put setting the dimension in TopScripts while in frontend tests we put it at the bottom of `<body>`. Maybe it would be better to be consistent in this aspect and choose a single place?

extensions/wikia/AbPerfTesting/js/FrontendDelay.js
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	var delay = w.wgPerfTestFrontEndDelay,
+		perf_test = w.wgTransactionContext['perf_test'];
+
+	// FIXME: perform the bucketing logic here, should be moved to the backend
+	delay = false;
+
+	if (typeof w.beacon_id === 'string') {
+		switch(w.beacon_id.charAt(0)) {
```
wladekb: no crc here?

extensions/wikia/AbPerfTesting/AbPerfTesting.setup.php
macbre: How about `AbPerformanceTesting`? This will make us consistent with `AbTesting` extension.

I'll run the rename when the review is completed (to avoid loosing inline comments).

extensions/wikia/AbPerfTesting/js/FrontendDelay.js
```diff
@@ -0,0 +1,81 @@
...
+			case '0':
```
wladekb: how about something more like instead of the switch:
```
var DELAY_BUCKETS = {
  '0': 0, 
  '1': 100, 
  '2': 200, 
  '3': 300
};
var i = w.beacon_in.charAt(0); // crc?
if ( i in DELAY_BUCKETS ) {
  perf_test += '_' + i;
  delay = DELAY_BUCKETS[i];
}
```

extensions/wikia/AbPerfTesting/js/FrontendDelay.js
macbre: I do agree, I'd like to keep the logic in a single place - i.e. in the backend code. However, CDN caching of front-end stuff does not allow us to do so.

Front-end delays (if handled by the backend script) for the testing group (chosen by modulo of `beacon_id`) would pollute the cache for clients from outside the test group.

extensions/wikia/AbPerfTesting/js/FrontendDelay.js
```diff
@@ -0,0 +1,81 @@
...
+
+	info('delay set to ' + delay + ' ms');
+
+	// @see http://api.jquery.com/jquery.holdready/
+	$.holdReady(true);
+
+	setTimeout(function() {
+		info('hold released');
+		$.holdReady(false);
+	}, delay);
```
wladekb: say, we have 100ms delay and this script is the last but one to execute and the last one takes 20ms to execute
then we'll add a delay of only (100-20) = 80ms
either we need to find another way to add the delay or we need to state it clearly that it is up to 100ms delay (might be actually 0ms if we are waiting for another http request that takes > 100ms)

extensions/wikia/AbPerfTesting/js/FrontendDelay.js
macbre: See the comment from above. It's an ugly hack right now to get the first data. We should definitely discuss the solid solution.

extensions/wikia/AbPerfTesting/classes/Hooks.class.php
wladekb: Reflection used to be slow and not considered a best practice in production code, has anything changed in that matter? (I haven't tracked it thoroughly recently)

extensions/wikia/AbPerfTesting/AbPerfTesting.setup.php
wladekb: Looks good.

wladekb:
I might have missed that but haven't seen a master switch to enable/disable all tests or control the frontend/backend tests. Are these somewhere here?

extensions/wikia/AbPerfTesting/classes/Hooks.class.php
macbre: It's pretty fast.

```
  0.18% 0.003898      1 - Wikia\AbPerfTesting\Hooks::onAfterInitialize
```

But to simplify the code we can just pass the params array as a single argument and let the experiment's constructor handle it.

macbre:
The extension is loaded sitewide via DefaultSettings. Good point, let's introduce a WF variable and load it via CommonExtensions to be able to disable it with a production push - just in case ;)

extensions/wikia/AbPerfTesting/AbPerfTesting.setup.php
```diff
@@ -0,0 +1,124 @@
...
+];
+
+$wgABPerfTestingExperiments['backend_delay_1'] = [
+	'handler' => 'Wikia\\AbPerfTesting\\Experiments\\BackendDelay',
+	'params' => [
+		'delay' => 100,
+	],
+	'criteria' => [
+		'oasisArticles' => true,
+		'traffic' => 1,
```
wladekb: how about support for ranges here, eg. `[ 1, 100 ]` or `'1-10'`? As we spoke we'll probably want to have more than 0.1% sampling over the traffic in a given bucket?

extensions/wikia/AbPerfTesting/classes/Hooks.class.php
```diff
@@ -0,0 +1,80 @@
...
+			wfDebug( __METHOD__ . " - session started\n" );
+
+			// log started sessions
+			global $wgUser;
+			WikiaLogger::instance()->info( __METHOD__, [
+				'experiment' => $experimentName,
+				'session_id' => session_id(),
+				'is_anon' => $wgUser->isAnon()
+			] );
+		}
```
macbre: Pinging @wladekb - we need these to avoid polluting the CDN cache with performance tests tracking data.

Additionally, this approach will allow us to set up the front-end test on the back-end side - ..e. less magic in front-end code.

extensions/wikia/AbPerfTesting/AbPerfTesting.setup.php
macbre: Support for ranges (e.g. `[ 1, 100 ] `) added in 73624e8

extensions/wikia/AbPerfTesting/classes/Hooks.class.php
wladekb: This solves the issue when the request hits backend, but are you sure the first will not get a cached response for the configured beacon bucket?

macbre: It will, both types of tests (for front-end and back-end) will need a request that hits the backend to be triggered. Given a small articles CDN hit ratio (a bit more than 10%) I think we can live with that in MVP phase of the framework.

wladekb: I think the results may then be opinionated towards heavy readers behavior and don't take casual readers into account fully. My best guess would be that casual readers may hit more popular pages where the hit ratio can be much bigger than 10%.

extensions/wikia/AbPerfTesting/js/FrontendDelay.js
macbre: [domReady](https://github.com/jfriend00/docReady) used to start the delay when domReady event is triggered by the browser.

extensions/wikia/AbPerfTesting/AbPerfTesting.setup.php
macbre: Renamed in 5ee6aa8

extensions/wikia/AbPerformanceTesting/classes/Hooks.class.php
```diff
@@ -0,0 +1,80 @@
...
+		new $experimentConfig['handler']( $experimentConfig['params'] ? : [] );
+
+		// mark a transaction with an experiment name
+		\Transaction::getInstance()->set( \Transaction::PARAM_AB_PERFORMANCE_TEST, $experimentName );
+
+		// mark a transaction using UA's custom dimensions
+		global $wgHooks;
+		$wgHooks['WikiaSkinTopScripts'][] = function( Array &$vars, &$scripts ) use ( $experimentName ) {
+			$val = \Xml::encodeJsVar( $experimentName );
+			$scripts .= \Html::inlineScript( "_gaq.push(['set', 'dimension20', {$val}]);" );
```
macbre: @harnash, let's add this variable in UA before this Thursday release.

wladekb:
Global switch allows us to disable the entire extension if something goes wrong which is good. 
However do we always want to run both frontend and backend tests in parallel? If not we'd need to add more fine-grained control over what tests are enabled/disabled... What do you think?

extensions/wikia/AbPerformanceTesting/AbPerformanceTesting.setup.php
```diff
@@ -0,0 +1,165 @@
...
+ *	'params' => [
+ *		'foo' => 25,
+ *		'bar' => 42,
+ *	],
+ *
+ *  # set of criteria to which bucket given experiment should be assigned (all needs to be met)
+ *	'criteria' => [
+ *      # all wikis are split into 1000 buckets (modulo of city ID), pick one here
+ *		'wikis' => 1,
+ *      # all clients are split into 1000 buckets (modulo of beacon_id md5 hash), pick one here
```
wladekb: it's a bit more of crc32 than md5 in the code ;-)

extensions/wikia/AbPerformanceTesting/js/FrontendDelay.js
```diff
@@ -0,0 +1,113 @@
...
+
+		// This is the one public interface
+		// docReady(fn, context);
+		// the context argument is optional - if present, it will be passed
+		// as an argument to the callback
+		return function(callback, context) {
+			// if ready has already fired, then just schedule the callback
+			// to fire asynchronously, but right away
+			if (readyFired) {
+				setTimeout(function() {callback(context);}, 1);
```
wladekb: Just curious, why is it 1 and not 0?

macbre: No idea about the reason. It's a 3rd-party code.

`setTimeout` with 0 is usually delayed by ~20ms. 0 or 1 - should not really make any difference

extensions/wikia/AbPerformanceTesting/criteria/OasisArticles.class.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+namespace Wikia\AbPerformanceTesting\Criteria;
+
+use Wikia\AbPerformanceTesting\Criterion;
+
+/**
+ * Define a "oasisArticles" criterion
```
wladekb: s/a/an/

macbre:
The A/B tests config does not contain any secret data (like API keys etc.) so I'd say - let's keep it all in the same place :)

wladekb:
lgtm

extensions/wikia/AbPerformanceTesting/classes/Hooks.class.php
harnash: we should rather use ```ga('set', 'dimension20', {$val});``` syntax.

macbre: Changed in #7532

harnash: Oh and one more thing. Above code will set dimension on the main property so it will be 10% sampled. If you want to be also collected for spacial wiki's (100% sampled) add this:
```
ga('special.set', 'dimension20', {$val});
```

macbre: Sampling is fine for now

PR #7468, opened 2015-06-10
CONCF 633 | Forum thread subject line is indented, not using full width available

tomasznapieralski:


tomasznapieralski:
@hakubo @rogatty

rogatty:
LGTM

extensions/wikia/Forum/css/Forum.scss
```diff
@@ -287,14 +287,16 @@
...
-	padding-right: 260px;
+	padding-right: 0;
 	position: relative;
 
+	h1 {
+		margin: 0;
+		word-break: normal;
+	}
+
 	.header-right {
```
hakubo: .WikiaPage .WikiaPageHeader this looks like a generic class that touches every page
are we sure that Forum.css is not loaded on other pages? :)

rogatty: /cc @bognix @Warkot

bognix: LGTM as a quick fix. Although the best idea would be to apply markup and styles from article so it becomes "social share icons" friendly.

tomasznapieralski: https://wikia-inc.atlassian.net/browse/CONCF-664

bognix: LGTM

PR #7473, opened 2015-06-10
CONCF-515 Call replaceLinkHolders only when needed

rogatty:
https://wikia-inc.atlassian.net/browse/CONCF-515

Fix timeouts that were happening on articles with a lot of images.

/cc @wladekb @hakubo @tomasznapieralski

### DIFF
--- a/extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
+++ b/extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
@@ -210,13 +210,8 @@ class ArticleAsJson extends WikiaService {
 				}
 			}
 
-			//because we take caption out of main parser flow
-			//we have to replace links manually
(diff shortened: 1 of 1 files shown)


rogatty:
@wladekb I've removed comment to consider caching identical captions which you suggested.

I did a test and they're never identical if there is a link placeholder inside. I created an article with two images, same captions. Then I var_dumped `$media['caption']` before replacing link holders and the result was:
```
string 'Internal link in caption: <!--LINK 0:0-->' (length=41)
string 'Internal link in caption: <!--LINK 0:1-->' (length=41)
```

hakubo:
LGTM, 
but we still have to think of situations where we have tons of captions with links

tomasznapieralski:
LGTM,
we need to think how to optimalize loading images

PR #7474, opened 2015-06-10
SOC-887 WallNotification emails are no longer being properly classified in Sendgrid

jsutterfield:
As part of revamping the new emails, we stopped sending the custom categories used by send grid to categorize our emails. This fixes that.

Ticket: https://wikia-inc.atlassian.net/browse/SOC-887


mklucsarits:
LGTM

PR #7479, opened 2015-06-11
CE-1720 Non-portable infoboxes Insights list

adamkarminski:

This PR brings a new Insights list of non-portable infoboxes to life. It includes:
- A limited version of a `QueryPage` subclass - `UnconvertedInfoboxesPage.class.php` - that is not listed on `Special:SpecialPages`. It handles recognizing the non-portable infoboxes in the `isTitleWithNonportableInfobox()` method.
- An `InsightsQuerypageModel` subclass that is responsible for creating a nice list with an amount of articles linking to the templates.
- Transfering a method that gets articles linking to a given page from `MoveNotice.php` to a new trait - `TitleTrait`. This allows using in different contexts directly from a Title object.
- Minor tweaks to Insights

**IMPORTANT! The wording is still in progress from the PO side so please skip it when reviewing.** :)

Have fun!


extensions/wikia/Insights/specials/UnconvertedInfoboxesPage.class.php
```diff
@@ -0,0 +1,116 @@
...
+						NS_TEMPLATE,
+						$row->title,
+					];
+				}
+			} );
+
+		return $nonportableTemplates;
+	}
+
+	public static function isTitleWithNonportableInfobox( Title $title ) {
```
adamkarminski: @kamilkoterba @Grunny Please, take a close look at this method. :)

extensions/wikia/Insights/specials/UnconvertedInfoboxesPage.class.php
```diff
@@ -0,0 +1,116 @@
...
+		$content = ( new WikiPage( $title ) )->getText();
+
+		$titleRegEx = '/infobox/i';
+		$titleRegExMatch = preg_match( $titleRegEx, $title->getText() );
```
Grunny: No need for a regex here, `stripos` would do the job.

extensions/wikia/Insights/specials/UnconvertedInfoboxesPage.class.php
```diff
@@ -0,0 +1,116 @@
...
+		if ( !empty( $titleRegExMatch ) ) {
+			$portableInfoboxRegEx = '/\<infobox\>/';
+			$portableInfoboxRegExMatch = preg_match( $portableInfoboxRegEx, $content );
```
Grunny: And here `strpos` will do the trick instead of the regex.

extensions/wikia/Insights/specials/UnconvertedInfoboxesPage.class.php
```diff
@@ -0,0 +1,116 @@
...
+
+			// If a portable infobox markup was not found
+			// the $portableInfoboxRegExMatch is empty
+			return empty( $portableInfoboxRegExMatch );
+		} else {
+			$nonportableInfoboxRegEx = '/(class=\".*infobox.*\")|(/i';
```
Grunny: This isn't a valid regex and will create compilation errors, `PHP Warning:  preg_match(): Compilation failed: missing )`.

Grunny: This regex (if the error is removed) also isn't very strict, as there is no guarantee `infobox` appears within the class attribute, and possibly infobox is being used as part of a subtemplate class ("infobox-row-helper"). As a completely unrealistic example, this would match your regex:
`<div class="foobar">This template is used on an infobox <span class="bar"></span></div>`

I'd consider at least changing both `.*` to `[^\"]*` to restrict it to within the attribute.

includes/wikia/traits/TitleTrait.php
```diff
@@ -0,0 +1,160 @@
...
+			'tl_title' => $this->getDBkey(),
+		];
+
+		$ilConds = [
+			'page_id=il_from',
+			'il_to' => $this->getDBkey(),
+		];
+
+		if ( is_array( $wgContentNamespaces ) && !empty( $wgContentNamespaces ) ) {
+			$namespaces = implode( ',', $wgContentNamespaces );
```
Grunny: As a best practice, prefer `DatabaseBase::makeList`, even if you trust the source.

extensions/wikia/Insights/Insights.i18n.php
```diff
@@ -30,6 +30,13 @@
 	'insights-notification-message-inprogress-wantedpages' => 'This page needs to be created. ',
 	'insights-notification-message-fixed-wantedpages' => 'Thanks for making this page!',
 	'insights-notification-next-item-wantedpages' => 'Create another article.',
+
+	'insights-list-subtitle-unconvertedinfoboxes' => 'Unconverted Infoboxes',
+	'insights-list-description-unconvertedinfoboxes' => 'The following infoboxes are built with tables or divs, and should be converted to use the new [[Help:PortableInfoboxes|Infobox markup]] so they render appropriately to all visitors.',
+	'insights-notification-message-inprogress-unconvertedinfoboxes' => 'This infobox needs to be updated with the new infobox markup.',
+	'insights-notification-message-fixed-unconvertedinfoboxes' => 'Awesome! The infobox on this page has been converted.',
+	'insights-notification-next-item-unconvertedinfoboxes' => 'Convert the next infobox.',
```
Grunny: QQQ docs?

adamkarminski: Coming, along with some previously missing ones. :)

includes/wikia/traits/TitleTrait.php
```diff
@@ -0,0 +1,160 @@
...
+				__METHOD__,
+				$options,
+				$joinConds
+			);
+		}
+
+		// Read the rows into an array and remove duplicates
+		// templatelinks comes second so that the templatelinks row overwrites the
+		// pagelinks row, so we get (inclusion) rather than nothing
+		if ( $fetchlinks ) {
```
Grunny: Is there a reason to do these conditionals again rather than just putting these loops after the relevant query inside the same conditional?

includes/wikia/traits/TitleTrait.php
adamkarminski: :+1:

includes/wikia/traits/TitleTrait.php
adamkarminski: I guess that the author wanted to break the code into logical blocks for readability... I'm with you on that matter. :)

extensions/wikia/Insights/specials/UnconvertedInfoboxesPage.class.php
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	public static function isTitleWithNonportableInfobox( $titleText, $contentText ) {
+		$titleNeedle = 'infobox';
+		if ( strripos( $titleText, $titleNeedle ) !== false ) {
+			$portableInfoboxNeedle = '<infobox>';
+
+			// If a portable infobox markup was found
+			// it means that the template doesn't have a non-portable infobox
+			return !( strpos( $contentText, $portableInfoboxNeedle ) !== false );
```
Grunny: Or just `return strpos( $contentText, $portableInfoboxNeedle ) === false;` rather than the double negative.

adamkarminski: <facepalm> of course :)

Grunny:
LGTM after the fixes. :)

extensions/wikia/Insights/models/InsightsQuerypageModel.php
```diff
@@ -82,6 +82,14 @@ public function arePageViewsRequired() {
 		return true;
 	}
 
+	public function isWlhLinkRequired() {
```
kamilkoterba: Abbreviations makes it harder to use and search. Be explicit.
Also please provide a documentation.

kamilkoterba:
Please provide documentation for public methods if possible.

extensions/wikia/Insights/specials/UnconvertedInfoboxesPage.class.php
```diff
@@ -0,0 +1,112 @@
...
+		$dbw = wfGetDB( DB_MASTER );
+
+		/**
+		 * 1. Get the new data first
+		 */
+		$nonportableTemplates = $this->reallyDoQuery();
+		$dbw->begin();
+
+		/**
+		 * 2. Delete the existing records
```
kamilkoterba: It could be more specific.
What do you think about 
'Delete all existing records for current queryPage type'
do I guess that right?

adamkarminski: It is a subclass that overwrites the original `recache` method so the only `queryPage` type here is `unconvertedinfoboxes`

extensions/wikia/Insights/specials/UnconvertedInfoboxesPage.class.php
```diff
@@ -0,0 +1,112 @@
...
+		}
+
+		return $num;
+	}
+
+	public function reallyDoQuery( $limit = false, $offset = false ) {
+		$dbr = wfGetDB( DB_SLAVE, [ $this->getName(), __METHOD__, 'vslow' ] );
+
+		$nonportableTemplates = ( new WikiaSQL() )
+			->SELECT( 'page_title as title' )
```
kamilkoterba: you can use ->AS_( 'title' ) that is Fluent SQL way

includes/wikia/traits/TitleTrait.php
```diff
@@ -0,0 +1,155 @@
...
+	abstract function getDBkey();
+
+	/**
+	 * Returns a list of pages that link to that Title. It is based on the method used in
+	 * Special:WhatLinksHere but returns raw results instead of a ready (and useless) HTML markup.
+	 *
+	 * @param int $level How deep should I go? 0 means to the bottom (which is limited to 2)
+	 * @return array
+	 */
+	public function getIndirectLinks( $level = 0 ) {
```
kamilkoterba: Do you think Special:WhatLinksHere could make use of that method to meet DRY principle?

adamkarminski: It could, although refactoring Special:WLH is definitely not in the scope of this task.

PR #7480, opened 2015-06-11
CE-1842 Adding fi starter to CreateNewWiki\

kamilkoterba:
https://wikia-inc.atlassian.net/browse/CE-1842

Adding fi starter to CreateNewWiki as starter list is duplicaed there from config

@Wikia/community-engineering


adamkarminski:
LGTM

PR #7482, opened 2015-06-11
CE-1979 Fix getDatabase.php script

kamilkoterba:
https://wikia-inc.atlassian.net/browse/CE-1979
Added ability to define cluster to search for a dump, 
added support for cluster numbers from 6 to 9
context for buckets naming
https://github.com/Wikia/chef-repo/blob/master/cookbooks/_mysql_helper/attributes/default.rb#L96

@Wikia/community-engineering @owend  @macbre @wladekb

### DIFF
--- a/maintenance/wikia/getDatabase.php
+++ b/maintenance/wikia/getDatabase.php
@@ -24,21 +24,17 @@ $USAGE =
 	"\t\t--help      show this message\n" .
 	"\t\t-h          Fetch and import to dev db by hostname (short name or fully qualified name ok)\n" .
 	"\t\t-f          Fetch a new database file from s3\n" .
-	"\t\t-c          Use different cluster to look for the dump: c1, c2, etc (optional, fetched automatically by) \n".
 	"\t\t-i          Import a downloaded file to dev db\n" .
 	"\t\t-p          Which dev database to use for target: sjc or poz (optional, defaults to WIKIA_DATACENTER) \n".
 	"\n";
 
-$opts = getopt ("c:h:i:f:p:?::");
+$opts = getopt ("h:i:f:p:?::");
 if( empty( $opts ) ) die( $USAGE );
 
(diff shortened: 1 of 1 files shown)


maintenance/wikia/getDatabase.php
```diff
@@ -69,11 +73,19 @@
...
-			$databaseDirectory = "database-f";
+		// Don't use fetched cluster number if it was provided on script invoke params list
+		if ( empty( $clusterNumberParam ) ) {
+			$clusterNumberParam = $matches[2];
+		}
+		if ( $clusterNumberParam > 9 ) {
+			echo "Clusters higher than 9 are not yet operated by this script. Time to update the script.\n";
+			exit;
+		}
+		$clusterLetter = strtr( $clusterNumberParam, '123456789', 'ABCDEFGHI' );
```
macbre: Maybe `chr(64 + <cluster number>)` instead of "hardcoded" `strtr`?

owend: The script already looks up the cluster from wikicities database, I don't think we ever need to provide it on the command line? 

The FIXME here was related to the directory name.   Cluster A-E use underscore but cluster F uses a dash.  We can also have a mapping of cluster -> separator like:

 $separator = strtr($matches[2], '1234567', ‘______-‘);

kamilkoterba:
I rediscussed possible usage of c param with my team and all we came up with were edge cases, so I removed it.
(primal concept was to use it when wiki was moved to a different cluster and new dump is not yet generated and you want to import that into devbox)

owend:
Aha.  Another edge case can happen...  I've seen it when the central db on dev is out of date, and the cluster lookup fails because the wiki isn't in there yet.  In that case, the proper fix is to update the central db though.  

Fix for cluster numbering/naming weirdness LGTM!

PR #7484, opened 2015-06-11
Revert "Close/maintenance: add --cluster parameter"

macbre:
See #7171

We'd like to re-enable the cron job that removes wikis every week on all clusters.

@michalroszka / @drozdo


drozdo:
LGTM

michalroszka:
:+1:

PR #7485, opened 2015-06-11
CE-1720 Simplify sql params 

kamilkoterba:
Simplify sql params preparation as DatabaseBase class combines it inside anyway.

Can you test that? as I just put it quickly within github editor

Feel free to close it if changes are wrong.

@adamkarminski

### DIFF
--- a/includes/wikia/traits/TitleTrait.php
+++ b/includes/wikia/traits/TitleTrait.php
@@ -34,7 +34,7 @@ trait TitleTrait {
 
 		// Make the query
 		$plConds = [
-			'page_id=pl_from',
+			'page_id' => 'pl_from',
 			'pl_namespace' => $this->getNamespace(),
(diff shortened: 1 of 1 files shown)


includes/wikia/traits/TitleTrait.php
```diff
@@ -34,7 +34,7 @@ public function getIndirectLinks( $level = 0 ) {
 
 		// Make the query
 		$plConds = [
-			'page_id=pl_from',
+			'page_id' => 'pl_from',
```
Grunny: This won't work as it will become `page_id = 'pl_from'` rather than `page_id = pl_from` since the values in key value pairs get run through `addQuotes`.

kamilkoterba:
Moved here https://github.com/Wikia/app/pull/7497

PR #7490, opened 2015-06-12
CE-1974 Update the title and the URL of non-portable infoboxes insights

adamkarminski:
@Wikia/community-engineering


kamilkoterba:
LGTM

PR #7494, opened 2015-06-12
(CE-1975) Ensure user is logged out when account requested closure

Grunny:
When a user who requested the closure of their account, when they attempt
to login and provide the correct credentials, we abort the login before the
cookies are set. With Helios, the access token is now set before that, so
we need to make sure the token is cleared before it can be used to initialise
the session.

/cc @michalroszka


michalroszka:
:+1:

PR #7497, opened 2015-06-12
CE-1720 Simplify sql params

kamilkoterba:
Simplify sql params preparation as DatabaseBase class combines it inside anyway.

Can you test that? as I just put it quickly within github editor

Feel free to close it if changes are wrong.

@adamkarminski


kamilkoterba:
Checked that, It won't work

PR #7499, opened 2015-06-12
SOC-466 Move founder emails to new Email extension

garthwebb:
https://wikia-inc.atlassian.net/browse/SOC-466


extensions/wikia/FounderEmails/events/FounderEmailsEditEvent.class.php
```diff
@@ -32,7 +34,109 @@ public function __construct( Array $data = array() ) {
 		$this->setData( $data );
 	}
 
-	public function enabled ( $wgCityId, User $user ) {
```
garthwebb: This ```enabled``` function was just moved further down.  The diff ended up a little goofy ...

extensions/wikia/FounderEmails/events/FounderEmailsEditEvent.class.php
```diff
@@ -43,151 +147,121 @@ public function enabled ( $wgCityId, User $user ) {
...
 			return false;
 		}
-		if ( $user->getOption( "founderemails-edits-$wgCityId" ) ) {
+
+		if ( $user->getOption( "founderemails-edits-$wikiId" ) ) {
 			return true;
 		}
+
 		return false;
 	}
```
garthwebb: As seen above, the ```process``` method was broken up into smaller methods.  More can be removed once we finish the final email sent via this code (https://wikia-inc.atlassian.net/browse/SOC-767)

extensions/wikia/Email/Controller/FounderController.class.php
```diff
@@ -0,0 +1,211 @@
...
+			'summary' => $this->getSummary(),
+			'buttonText' => $this->getChangesLabel(),
+			'buttonLink' => $this->getChangesLink(),
+			'contentFooterMessages' => [
+				$this->getFooterEncouragement(),
+				$this->getFooterArticleLink(),
+				$this->getFooterAllChangesLink(),
+			],
+			'details' => $this->getDetails(),
+			'hasContentFooterMessages' => true
```
mklucsarits: It seems strange to me to have to explicitly set this attribute. Is there a way we can just check ```count(data['contentFooterMessages']) > 0```?

extensions/wikia/Email/Controller/FounderController.class.php
```diff
@@ -0,0 +1,211 @@
...
+	protected function getFooterEncouragement() {
+		$name = $this->getCurrentUserName();
+		$profileUrl = $this->getCurrentProfilePage();
+
+		return $this->getMessage( $this->getFooterEncouragementKey(), $profileUrl, $name )
+			->parse();
+	}
+
+	protected function getFooterEncouragementKey() {
+		return 'emailext-founder-encourage';
```
mklucsarits: This might be more appropriate as a constant, but just my opinion.

extensions/wikia/FounderEmails/events/FounderEmailsEditEvent.class.php
```diff
@@ -43,151 +147,121 @@ public function enabled ( $wgCityId, User $user ) {
...
-				}
+	private function userCounterNeedsReset( array $counter ) {
+		if ( empty( $counter[F::app()->wg->CityId] ) ) {
+			return true;
+		}
 
-				// quit if the Founder has recieved enough emails today
+		$today = date( 'Y-m-d' );
+		$wikiCounter = $counter[F::app()->wg->CityId];
+		return empty( $wikiCounter['date'] ) || $wikiCounter['date'] !== $today;
```
mklucsarits: Only the second condition is needed here.

extensions/wikia/Email/Controller/FounderController.class.php
garthwebb: I've looked at it several times because I don't like it either.  However its not possible in mustache AFAICT.  The code is this:
```mustache
{{# hasContentFooterMessages}}
	<tr>
		<td colspan="2">
			<div class="content-footer-messages">
				{{#contentFooterMessages}}
					<p>{{{.}}}</p>
				{{/contentFooterMessages}}
			</div>
		</td>
	</tr>
{{/ hasContentFooterMessages}}
```
We only want the ```<tr>``` stuff if we actually have an array of values in ```contentFooterMessages```.  However we only want it once, and mustache only gives you one behavior from ```{{#foo}}...{{/foo}}```; a simple conditional if ```foo``` is a scalar value or a loop if ```foo``` is an array.

extensions/wikia/Email/Controller/FounderController.class.php
garthwebb: We have been overriding functions like this a lot in these templates so each child can return a different key for a common part of the email.  Constants can technically be overridden as well using ```static::CONST_NAME``` but I've never been a huge fan of that as it seems unintuitive.

extensions/wikia/Email/Controller/FounderController.class.php
jsutterfield: Yea, I spent some time on another email ticket trying to pull out this `hasContentFooterMessages` variable, but there's no way in mustache to use the length of an array has a boolean. Or at least when you do, it won't loop over the array.

Damian is doing something similar to what you're suggesting @mklucsarits [here](https://github.com/Wikia/app/blob/dev/extensions/wikia/Email/Controller/WatchedPageController.class.php#L92). I think the difference is that the check is done in the base class, and then some of the sublcasses will have contentFooterMessages and some won't. In this case where we always have contentFooterMessages I think just using `true` makes sense.

extensions/wikia/Email/Controller/FounderController.class.php
```diff
@@ -0,0 +1,211 @@
+<?php
+
+namespace Email\Controller;
+
+use Email\Check;
+use Email\EmailController;
+use Email\Fatal;
+
+abstract class FounderController extends EmailController {
```
jsutterfield: So this class is missing tracking info, which means all of these emails will use the default category of "UserMailer".

The way we have been adding tracking categories is via a class constant like [this](https://github.com/Wikia/app/blob/dev/extensions/wikia/Email/Controller/ForumController.class.php#L11). The problem for this ticket though is that there looks like there's a lot more logic in place for the tracking category (FounderEmailsEditAnonEN versus FounderEmailsEditAnonINT versus FounderEmailsEditUserEN).

Not sure what the best approach would be, maybe change the constants to a method which can be overridden? Or we could check with Cristina to see if these fine-grained categories are even useful any more and, if not, just give all of these a category of "FounderEmail."

extensions/wikia/Email/Controller/FounderController.class.php
```diff
@@ -0,0 +1,211 @@
...
+					'value' => NS_MAIN
+				]
+			]
+		];
+
+		return array_merge_recursive( $formFields, parent::getEmailSpecificFormFields() );
+	}
+}
+
+class FounderEditController extends FounderController { }
```
jsutterfield: What's this class for?

extensions/wikia/Email/EmailController.class.php
```diff
@@ -448,6 +448,11 @@ protected function findUserFromRequest( $paramName, \User $default = null ) {
 			return $default;
 		}
 
+		// Allow an anonymous user to be specified
+		if ( $userName == -1 ) {
```
jsutterfield: What do you think about pulling `-1` into a class constant?

extensions/wikia/FounderEmails/events/FounderEmailsEditEvent.class.php
```diff
@@ -43,151 +147,121 @@ public function enabled ( $wgCityId, User $user ) {
...
+	private function sendUsingEmailExtension( $controller, User $user, array $eventData ) {
+		$params = [
+			'targetUser' => $user->getName(),
+			'currentUser' => $eventData['editorName'],
+			'pageTitle' => $eventData['titleText'],
+			'pageNs' => $eventData['titleNs'],
+			'previousRevId' => $eventData['previousRevId'],
+			'currentRevId' => $eventData['currentRevId'],
+			'fromAddress' => \F::app()->wg->PasswordSender,
+			'fromName' => \F::app()->wg->PasswordSenderName,
```
jsutterfield: I made these same values for `fromAddress` and `fromName` the defaults in the EmailController as part of SOC-530 which is currently in QA. Once that's merged, we don't need to specify these here any more. Doesn't hurt to leave them in, but we should try to remember to remove them once SOC-530 is merged.

extensions/wikia/Email/Controller/FounderController.class.php
mklucsarits: I understand the problem with Mustache, but what I originally meant was, can ```hasContentFooterMessages``` be set using logic in PHP. It looks like WikiaResponse.class.php sets overwrites the data each time ```setData``` is called, so that's out. It also looks like there's nothing like a ```appendData``` function, despite it's potential usefulness.

jsutterfield: Oh, hmm, I think I see what you mean. Like have that logic in one place so we wouldn't have to set it explicitly in each of the controllers (hardcoded or otherwise)? That could be worth looking into. We're actually duplicating quite a bit with all of the `body` methods we keep redefining with similar looking arrays of data. Consolidating that at some point would be nice.

extensions/wikia/Email/Controller/FounderController.class.php
garthwebb: I saw this in the original but I thought some how the ticket didn't specify this level of tracking originally.  I'll look into it ...

extensions/wikia/FounderEmails/events/FounderEmailsEditEvent.class.php
garthwebb: Yah, I remembered this then looked around in vain to try to find where you had done this.  I just gave up figuring you'd point this out when the time came ;)

jsutterfield: Gah! I fell right into your trap!

extensions/wikia/Email/Controller/FounderController.class.php
garthwebb: Ok, fixed

extensions/wikia/Email/Controller/FounderController.class.php
garthwebb: Yah, the irritation of setting this boolean manually is about equal to the irritation I would feel trying to abstract this in a meaningful way.  My vote is to just leave this as is unless someone can think of something clever.

mklucsarits: That's fine with me.

extensions/wikia/Email/Controller/FounderController.class.php
jsutterfield: Looks good. The last thing is I'd been putting the category names in the TrackingCategories class. These should probably either be part of that class, or we could do away with that class and just hardcode the categories as part of each of the controllers. My vote would be for the former.

extensions/wikia/FounderEmails/events/FounderEmailsEditEvent.class.php
garthwebb: I'm glad you trust my initialization code above!  I've been slapped by PHP when I haven't been so rigorous, but I suppose we should always have 'date' defined here...

extensions/wikia/Email/Controller/FounderController.class.php
jsutterfield: :+1:

extensions/wikia/FounderEmails/events/FounderEmailsEditEvent.class.php
mklucsarits: Ah ok. If we're trying to avoid PHP warnings in logs, then I see your point.

mklucsarits:
LGTM

extensions/wikia/Email/Controller/FounderController.class.php
garthwebb: Blerg.  I kind of hate creating a class of constants that is just used to set other constants in different classes.  I can buy that ```WALL_NOTIFICATION``` is used more than once however so we would want to define it only once.  I'll move them over ...

mklucsarits:
Still LGTM ;)

jsutterfield:
LGTM!

garthwebb:
Relocating this pull request to https://github.com/Wikia/app/pull/7554

PR #7500, opened 2015-06-12
SOC-864 Make sure var is of expected type before calling method, othe…

jsutterfield:
…rwise log error

Ticket: https://wikia-inc.atlassian.net/browse/SOC-864


garthwebb:
LGTM

PR #7504, opened 2015-06-15
CloseWikiMaintenance: handle DB exceptions where dropping the wiki database

macbre:
Handle the following error:

```
A database error has occurred.  Did you forget to run maintenance/update.php after upgrading?  See: https://www.mediawiki.org/wiki/Manual:Upgrading#Run_the_update_script
Query: DROP DATABASE `cinemacultura`
Function: CloseWikiMaintenance:execute
Error: 1008 Can't drop database 'cinemacultura'; database doesn't exist (x.x.x.x)
```

@michalroszka 

Relates to #7484


michalroszka:
:+1:

PR #7505, opened 2015-06-15
DAT-2891 block elements will now be parsed correctly, test added

idradm:


extensions/wikia/PortableInfobox/services/Parser/MediaWikiParserService.php
```diff
@@ -21,17 +21,17 @@ public function __construct( \Parser $parser, \PPFrame $frame ) {
 	 */
 	public function parse( $wikitext ) {
 		wfProfileIn( __METHOD__ );
-		if ( substr( $wikitext, 0, 1 ) == "*" ) {
+		if ( in_array( substr( $wikitext, 0, 1 ), [ '*', '#' ] ) ) {
```
SebastianMarzjan: Do you want to add definition lists and terms too?

SebastianMarzjan:
In general LGTM; the dl support can be treated as a functionality increment and done separately if needed

PR #7506, opened 2015-06-15
AbTestingData: improve the profiling of DB queries coming from getEffectiveChangeTime

macbre:
@wladekb 

https://wikia-inc.atlassian.net/browse/ER-6742


wladekb:
lgtm

PR #7512, opened 2015-06-15
CONCF-509 Return the main page data OR the article data

rogatty:
https://wikia-inc.atlassian.net/browse/CONCF-509

/cc @bognix @Warkot @tomasznapieralski

### DIFF
--- a/extensions/wikia/MercuryApi/MercuryApiController.class.php
+++ b/extensions/wikia/MercuryApi/MercuryApiController.class.php
@@ -308,30 +308,29 @@ class MercuryApiController extends WikiaController {
 		try {
 			$title = $this->getTitleFromRequest();
 			$articleId = $title->getArticleId();
-			$articleAsJson = $this->getArticleJson( $articleId, $title );
 
 			$data = [
 				'details' => $this->getArticleDetails( $articleId ),
 				'topContributors' => $this->getTopContributorsDetails(
 					$this->getTopContributorsPerArticle( $articleId )
-				),
-				'article' => $articleAsJson
+				)
 			];
+
+			$isMainPage = $title->isMainPage();
+			$data['isMainPage'] = $isMainPage;
+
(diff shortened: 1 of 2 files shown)


bognix:
LGTM

extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -77,12 +77,46 @@ private function getArticleDetails( $articleId ) {
...
+	 * @throws WikiaException
+	 */
+	private function getArticleDescription( $articleId, $descLength = 100 ) {
+		$article = Article::newFromID( $articleId );
+		$title = $article->getTitle();
+		$sMessage = null;
+
+		if ( $title->isMainPage() ) {
+			// we're on Main Page, check MediaWiki:Description message
+			$sMessage = wfMessage( 'Description' )->text();
```
Warkot: Is it safe to use `->text()`?

Warkot:
If you're sure that moving the method doesn't break anything, then LGTM. Otherwise, why not just make it public and reference it?

extensions/wikia/MercuryApi/MercuryApiController.class.php
rogatty: It's copied from https://github.com/Wikia/app/blob/dev/extensions/wikia/ArticleMetaDescription/ArticleMetaDescription.php#L64

But from what I see here: https://github.com/Wikia/app/blob/dev/includes/GlobalFunctions.php#L1358
`wfMessage` always returns instance of `Message`. Should be safe.

rogatty:
`getArticleDescription` was written (mostly copied from `ArticleMetaDescription` extension ;)) by me and used only for this case. I wouldn't like to spend too much time on thinking where to put it as a public method before it's actually needed elsewhere.

Warkot:
LGTM then. (y)

tomasznapieralski:
LGTM

PR #7514, opened 2015-06-15
(CE-2003) Close modals when the escape key is hit

Grunny:
This adds support for closing modals when the escape button is hit. This
was the default in the old jQuery modals but was missing from the new
modals.


lgarczewski:
LGTM.

kamilkoterba:
LGTM

PR #7516, opened 2015-06-15
CE-2009 Add a control mechanism for displaying Flags 

adamkarminski:

I've added a new ParserOption to control the display of Flags. It's set in `Article::view` for Oasis and Monobook and after an edit (which can only happen in Oasis and Monobook for now).


lgarczewski:
LGTM.

PR #7518, opened 2015-06-15
(CE-1370) Switch allowuserjs/css to useuserjs/css

Grunny:
Switch the URL parameters allowuserjs and allowusercss to useuserjs and
useusercss to be consistent with usesitejs/css. To avoid breaking things
for existing users, it will fallback to allowuserjs/css as aliases.


lgarczewski:
LGTM.

PR #7525, opened 2015-06-15
SOC-781 Use master when getting redirect URL, validate newTitle var

jsutterfield:
This ticket fixes the fatal being thrown as part of https://wikia-inc.atlassian.net/browse/SOC-871. We're now both validating the value of `$newTitle`, as well as using the master database when querying for the redirectTarget in case the slave hasn't caught up yet.

### DIFF
--- a/includes/WikiPage.php
+++ b/includes/WikiPage.php
@@ -591,11 +591,11 @@ class WikiPage extends Page {
 	 * The target will be fetched from the redirect table if possible.
 	 * If this page doesn't have an entry there, call insertRedirect()
 	 *
-	 * @param bool $userMaster
+	 * @param int $useMasterDb
 	 * @return Title|mixed object, or null if this page is not a redirect
 	 */
-	public function getRedirectTarget( $userMaster = false ) {
-		if ( !$this->mTitle->isRedirect() ) {
+	public function getRedirectTarget( $useMasterDb = 0 ) {
+		if ( !$this->mTitle->isRedirect( $useMasterDb ) ) {
 			return null;
 		}
 
@@ -604,7 +604,7 @@ class WikiPage extends Page {
 		}
 
 		# Query the redirect table
-		$dbr = $userMaster ? wfGetDB( DB_MASTER ) : wfGetDB( DB_SLAVE );
+		$dbr = $useMasterDb ? wfGetDB( DB_MASTER ) : wfGetDB( DB_SLAVE );
 		$row = $dbr->selectRow( 'redirect',
 			array( 'rd_namespace', 'rd_title', 'rd_fragment', 'rd_interwiki' ),
 			array( 'rd_from' => $this->getId() ),
--- a/includes/WikiPage.php
+++ b/includes/WikiPage.php
@@ -590,9 +590,11 @@ class WikiPage extends Page {
 	 *
 	 * The target will be fetched from the redirect table if possible.
 	 * If this page doesn't have an entry there, call insertRedirect()
+	 *
+	 * @param bool $userMaster
 	 * @return Title|mixed object, or null if this page is not a redirect
 	 */
-	public function getRedirectTarget() {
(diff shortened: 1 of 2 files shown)


includes/WikiPage.php
```diff
@@ -590,9 +590,11 @@ public function isCountable( $editInfo = false ) {
 	 *
 	 * The target will be fetched from the redirect table if possible.
 	 * If this page doesn't have an entry there, call insertRedirect()
+	 *
+	 * @param bool $userMaster
 	 * @return Title|mixed object, or null if this page is not a redirect
 	 */
-	public function getRedirectTarget() {
+	public function getRedirectTarget( $userMaster = false ) {
 		if ( !$this->mTitle->isRedirect() ) {
```
garthwebb: The ```isRedirect``` method does a DB lookup as well to determine whether the page is a redirect, and it accepts the ```\Title::GAID_FOR_UPDATE``` argument.

includes/WikiPage.php
```diff
@@ -590,9 +590,11 @@ public function isCountable( $editInfo = false ) {
 	 *
 	 * The target will be fetched from the redirect table if possible.
 	 * If this page doesn't have an entry there, call insertRedirect()
+	 *
+	 * @param bool $userMaster
 	 * @return Title|mixed object, or null if this page is not a redirect
 	 */
-	public function getRedirectTarget() {
+	public function getRedirectTarget( $userMaster = false ) {
```
garthwebb: I know its pretty unintuitive, but we should probably use ```\Title::GAID_FOR_UPDATE``` for this kind of stuff like the rest of this class and many other core classes do.

jsutterfield: Yea, it is pretty unintuitive. This file also has a method which accepts the string `"fromdbmaster"` to do the same thing which is one of the uglier things I've seen in a while. Good to be consistent though, I'll update to use `\Title::GAID_FOR_UPDATE`.

includes/WikiPage.php
```diff
@@ -590,10 +590,12 @@ public function isCountable( $editInfo = false ) {
 	 *
 	 * The target will be fetched from the redirect table if possible.
 	 * If this page doesn't have an entry there, call insertRedirect()
+	 *
+	 * @param int $useMasterDb
 	 * @return Title|mixed object, or null if this page is not a redirect
 	 */
-	public function getRedirectTarget() {
-		if ( !$this->mTitle->isRedirect() ) {
+	public function getRedirectTarget( $useMasterDb = 0 ) {
```
garthwebb: Just to be a stickler, you should probably use the same signature as other methods that accept this, e.g.:
```php
public function getRedirectTarget( $flags = 0 )
```
Especially since what you pass to ```isRedirect``` is not a boolean, but this bit field value.  Then on L607 you would check:
```php
$dbr = $flags & \Title::GAID_FOR_UPDATE ? ...
```
If you want to be clear about the only flag you care about you could do your bitwise just prior into a variable:
```php
$useMasterDb = $flags & \Title::GAID_FOR_UPDATE;
```

jsutterfield: Okay, that makes sense. I'll update.

includes/WikiPage.php
```diff
@@ -590,10 +590,12 @@ public function isCountable( $editInfo = false ) {
...
 	 * The target will be fetched from the redirect table if possible.
 	 * If this page doesn't have an entry there, call insertRedirect()
+	 *
+	 * @param int $flags
 	 * @return Title|mixed object, or null if this page is not a redirect
 	 */
-	public function getRedirectTarget() {
-		if ( !$this->mTitle->isRedirect() ) {
+	public function getRedirectTarget( $flags = 0 ) {
+		if ( !$this->mTitle->isRedirect( $flags ) ) {
```
garthwebb: @jsutterfield ok, it would be really great if we could just make this text blue ...

garthwebb:
LGTM

PR #7534, opened 2015-06-16
Rename CONTROL group to something less misleading in A/B tests

wladekb:
Lack of assigned group in A/B test ends up reported as CONTROL group which happens to be misleading in current ads-related use cases. This change should make the name more descriptive and intuitive.

/cc @macbre


macbre:
LGTM

nandy-andy:
LGTM :smiley_cat:

PR #7536, opened 2015-06-16
wfSetupSession: log all calls with 1% sampling

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-1266

We'd like to know how many PHP sessions we start now and compare it with the number of sessions started by performance A/B testing. This will allow us to measure the impact of increased logged-in traffic on various metrics (CDN hit ratio, # of requests hitting the backend, # of database queries. backend response time, # of calls to services such as Helios or Phalanx).

@wladekb / @jcellary / @michalroszka


michalroszka:
:+1:

PR #7539, opened 2015-06-17
Restore the disabled test to ensure we cover the partners api.

sebastianmarzjan:
Next step: move to a separate group


SebastianMarzjan:
(the provider fixed the issue on their end)

nandy-andy:
LGMT :+1:

PR #7540, opened 2015-06-17
wfSetupSession: log session_id and the full backtrace

macbre:
See #7536

Log:
- the full backtrace (`wfGetCaller` was returnings `unknown`)
- does the anon started the session
- did we restart the existing session

@michalroszka


michalroszka:
:+1:

PR #7546, opened 2015-06-17
AbPerformanceTesting: fix a fatal

macbre:
We do not use any arguments passed to this function, do not force types

@michalroszka / @wladekb


wladekb:
E_FATAL_NOT_ACCEPTED ;-)

PR #7548, opened 2015-06-17
CE-2012: Added logging for flags parameters change

lukaszkonieczny:
- Removed unnecessary php templates (we are using mustache)
- Changed post data structure from flat to array


extensions/wikia/Flags/scripts/FlagsModal.js
```diff
@@ -94,9 +94,15 @@ require(
...
 
-				flagsData[0].flags = prepareFlagsData(flagsData[0].flags);
+				if (flagsData[0].flags) {
+					template = res.mustache[0];
+					flagsData[0].flags = prepareFlagsData(flagsData[0].flags);
+				} else if (flagsData[0].emptyMessage) {
+					template = res.mustache[1];
+				} else if (flagsData[0].exceptionMessage) {
+					template = res.mustache[2];
+				}
```
kamilkoterba: This function is growing. Would you agree it's good time to move it and name it to avoid anonymous function?

kamilkoterba:
Except from one comment above it LGTM.
Feel free to merge when ready.

extensions/wikia/Flags/scripts/FlagsModal.js
lgarczewski: My suggestion would be to leave this in place, in this case splitting this off to a separate function this would be the only place that uses that code. Since the logic is very simple, I'd opt to leave it here -- if we end up adding more code we can revisit and potentially break this up as you suggested, Kamil.

lgarczewski:
LGTM.

extensions/wikia/Flags/scripts/FlagsModal.js
kamilkoterba: OK, remember about that one for the future as code in anonymous functions is harder to debug

PR #7550, opened 2015-06-17
CE-2046 Allow merging vars from another ParserOutput object

adamkarminski:
The idea here is to allow other instances of ParserOutput (e.g. coming from portable Flags) being merged into the one of the main article. This allows you to have control over parsing any block of wikitext and still take benefit of `LinksUpdate.php` (Categories, Special:WhatLinksHere etc.).

I'm not handling the `mText` var on purpose to allow flexible content manipulation.

cc: @wladekb @macbre @Grunny We've talked about it offline :)

### DIFF
--- a/includes/parser/ParserOutput.php
+++ b/includes/parser/ParserOutput.php
@@ -148,6 +148,27 @@ class ParserOutput extends CacheTime {
 	public $mPerformanceStats = array();
 	function getPerformanceStats( $k )     { return @$this->mPerformanceStats[$k]; }
 	function setPerformanceStats( $k, $v ) { $this->mPerformanceStats[$k] = $v; }
+
+	/**
+	 * @var Names of vars that we are able to merge from another ParserOutput object
+	 */
+	public static $varsToMerge = [
+		'mCategories',
+		'mLinks',
+		'mTemplates',
+		'mTemplateIds',
+		'mImages',
+		'mFileSearchOptions',
+		'mExternalLinks',
+		'mInterwikiLinks',
+		'mModules',
+		'mModuleScripts',
+		'mModuleStyles',
+		'mModuleMessages',
+		'mWarnings',
+		'mLanguageLinks',
+	];
+
 	# </Wikia>
 
 	const EDITSECTION_REGEX = '#<(?:mw:)?editsection page="(.*?)" section="(.*?)"(?:/>|>(.*?)(</(?:mw:)?editsection>))#';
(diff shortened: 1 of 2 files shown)


macbre:
Worth adding a unit test? :)

adamkarminski:
@macbre :+1: :mag: :santa:

includes/parser/tests/ParserOutputTest.php
```diff
@@ -0,0 +1,382 @@
...
+class ParserOutputTest extends WikiaBaseTest {
+
+	/**
+	 * @dataProvider getParserOutputVars
+	 */
+	public function testMergeExternalParserOutputVars( $sourceVars, $externalVars, $expectedVars ) {
+		$parserOutputSource = new ParserOutput();
+		$parserOutputExternal = new ParserOutput();
+
+		foreach ( $parserOutputSource::$varsToMerge as $var ) {
```
adamkarminski: Just as a note - I am aware of methods like `addModules`, `addCategory` etc., however mocking a lot of `Title`, `Category` and `Revisions` objects, just to get these arrays would be over-engineering in my opinion. :)

includes/parser/tests/ParserOutputTest.php
```diff
@@ -0,0 +1,382 @@
...
+			$this->assertSame( $expectedVars[$var], $parserOutputSource->$var );
+		}
+	}
+
+	public function getParserOutputVars() {
+		return [
+			// 1st test case
+			[
```
wladekb: you can use named array item (specific string as key) you can make it easier to debug (the name appears in the test failure information

macbre:
LGTM

includes/parser/ParserOutput.php
```diff
@@ -466,4 +489,62 @@ function recordOption( $option ) {
...
+			// operator to use values from both arrays.
+			} elseif ( is_array( $source[$key] ) && is_array( $new[$key] ) ) {
+				$result[$key] = $source[$key] + $new[$key];
+
+			// If we get to this point it means that both arrays are non-associative
+			// and we can safely merge them because preserving indexes is not important.
+			} else {
+				$result = array_merge( $source, $new );
+				return $result;
+			}
```
wladekb: `is_array( $source[$key] ) && !is_array( $new[$key] )` is explicitly considered on line 533
while:
`!is_array( $source[$key] ) && is_array( $new[$key] )` implicitly goes to the else clause

is this correct?

adamkarminski: Good catch, I think that whenever type of `$source[$key]` does not match type of `$new[$key]`, the `$source[$key]` is used. Thank you!

adamkarminski: Done!

wladekb:
lgtm

PR #7553, opened 2015-06-17
Removing jserrorslog.wikia.com domain

ljagiello:
I would like to cleaned up VCL and chef-repo from `jserrorslog.wikia.com`. To do that I'm going to disable MW code responsible for generating requests with `jserrorslog.wikia.com` domain.

/cc @gabrys @macbre @owend


owend:
This LGTM.  Would need to merge to release branch if you want it to go out tomorrow.

macbre:
Let's remove the code bound to this hook as well.

ljagiello:
@macbre sure, I'll create separate PR after we will know it's not in use.

macbre:
:+1:

PR #7556, opened 2015-06-18
DAT-2903 remove external api integration test from pr suite, add new …

idradm:
…test group for such cases


SebastianMarzjan:
I think it makes sense to exclude this group also in https://github.com/Wikia/app/pull/7556/files#diff-2231ca366d8aa7366f5594abfd41d66aR48;
What do you think about adding phpunit-external build to take all tests from this group?

SebastianMarzjan:
:+1:

PR #7559, opened 2015-06-18
SwiftSync: use $wgSwiftSyncDB global

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-1287

As we're going to move specials database to a separate cluster we have to make sure that no other databases are accessed via $wgSpecialsDB global.

@wladekb / @michalroszka / @drozdo


michalroszka:
:+1:

PR #7566, opened 2015-06-18
Add Geo.setCountryCode method

gabrys:


nandy-andy:
LGTM :+1:

resources/wikia/modules/geo.js
```diff
@@ -44,6 +44,20 @@
 		}
 
 		/**
+		 * Returns the code for the country
```
fraszczakszymon: It is not up to date and function does not `@return` anything

PR #7570, opened 2015-06-18
SOC-765 As a wikia founder, I want pretty "new member joined" emails

jsutterfield:
Update the new member emails received by admins to use the new templates and email extension.

Ticket: https://wikia-inc.atlassian.net/browse/SOC-765


extensions/wikia/FounderEmails/events/FounderEmailsRegisterEvent.class.php
```diff
@@ -1,101 +1,52 @@
...
-				wfProfileOut( __METHOD__ );
-				$founderEmailObj->notifyFounder( $user, $this, $mailSubject, $mailBody, $mailBodyHTML, $wgCityId, $mailCategory );
-			}
-			return true;
+			F::app()->sendRequest( self::EMAIL_CONTROLLER, 'handle', $emailParams );
 		}
 
-		wfProfileOut( __METHOD__ );
-		return false;
+		return true;
```
mklucsarits: Does this change affect any external calls to process()?

jsutterfield: Nope. You'll see that before editing this method it returned in a few different places. It still returns false if it's an anwersWiki, but otherwise the 3 other checks about whether we send the email or not got moved into the Email extension.

extensions/wikia/FounderEmails/FounderEmailsController.class.php
```diff
@@ -84,10 +84,7 @@ public function executeGeneralUpdate( $params ) {
 		if ( isset( $params['$PAGEURL'] ) ) {
 			$this->buttonUrl = $params['$PAGEURL'];
 		}
-		switch( $this->type ) {
-			case 'user-registered':
-				$this->buttonUrl = $params['$EDITORTALKPAGEURL'];
-				break;
+		switch($this->type) {
```
garthwebb: Spacing

extensions/wikia/FounderEmails/events/FounderEmailsRegisterEvent.class.php
```diff
@@ -1,101 +1,52 @@
...
-				// skip if not enable
-				if ( !$this->enabled( $wgCityId, $user ) ) {
-					continue;
-				}
-				self::addParamsUser( $wgCityId, $user->getName(), $emailParams );
-
-				$langCode = $user->getOption( 'language' );
-				$mailSubject = strtr( wfMsgExt( 'founderemails' . $wikiType . '-email-user-registered-subject', array( 'content' ) ), $emailParams );
-				$mailBody = strtr( wfMsgExt( 'founderemails' . $wikiType . '-email-user-registered-body', array( 'content' ) ), $emailParams );
+		if ( !$this->isThresholdMet( count( $events ) ) ) {
```
garthwebb: In a different PR I eliminated the threshold setting and removed the ```isThresholdMet``` over-engineering.  I would just replace this with:
```php
if ( $count( $events ) > 0 ) {
...
}
```

extensions/wikia/Email/Controller/FounderController.class.php
```diff
@@ -244,3 +244,89 @@ protected function getFooterEncouragement() {
...
+	 * @throws \Email\Check
+	 */
+	public function assertFounderWantsNewMembersEmail() {
+		$wikiId = \F::app()->wg->CityId;
+		if ( !$this->targetUser->getBoolOption( "founderemails-joins-$wikiId"  ) ) {
+			throw new Check( "Founder doesn't want to be emailed about new members joining this wiki" );
+		}
+	}
+
+	protected function getFooterMessages() {
```
garthwebb: You should probably override ```getEmailSpecificFormFields``` and have it explicitly return ```EmailController::getEmailSpecificFormFields``` to bypass the extra fields that the ```FounderController``` provides that you don't need ... or create a ```FounderEditController``` class that only the edit classes inherit from and move the edit fields there.  Or maybe @mklucsarits already did this in a different branch?

jsutterfield: Ah, good catch @garthwebb. I checked with Matt and he just created a separate class for his work, he didn't end up breaking apart the `FounderController` class. I think I'll just make another noop method since it's fast and easy. If we end up needing to extend the `FounderController` class more than we can break it up, but I think after mine and Matt's ticket we should be done with the Founder emails.

mklucsarits:
LGTM

garthwebb:
LGTM

PR #7572, opened 2015-06-18
Proof of concept user preference service interface

drsnyder:
Add a user preference "service" interface that we can use to replace the SQL call sites using the `user_properties` table. For example, the logic [here](https://github.com/Wikia/app/blob/dev/includes/User.php#L4406-4424) would be replaced (approximately) by:

``` php
$service = \Wikia\Services\ServiceRegistry::instance()->getUserPreferenceService();
$res = $service->getPreferences($this->getId())
```

This is associated with [SERVICES-441](https://wikia-inc.atlassian.net/browse/SERVICES-441). See also this [blog post](https://one.wikia-inc.com/wiki/User_blog:Damon_Snyder/Replacing_%22preferences%22_in_MW_with_the_user_preference_service) for the rational.

/cc @Wikia/services-team @garthwebb @cmiller-wikia


sqreek:
:+1: 
minor: I think /Service/User/

drsnyder:
@sqreek I saw you had a comment about "minor: I think /Service/User/" but it looks like it disappeared. What was that in reference to?

sqreek:
It's only suggestion's:

1) I will move Domain\User to Service\Domain\User and keep all related with service stuff in one namespace. All services stuff is in one space. and there will not mess up with Logger, Vighnette or Util.
2) Maybe rename Service to ... ServicesStuffSomething becuse:
3) In ServicesStuffSomething create Gateway, Services, Domain, Consul.
4) If I got it right User/ should contain all services related to User?, So going this way if we want to add service related to Article, it will be Article/ so going this way futher, if some service connect to article of this some user, I need to check it in to places. I would prefer one flat folder structure for each of service keep in separate folder/namespace

Also in other way Consul can be unrelated with services, so it can go out into src/ , also in future domain can be used in other places than services (I don't see it right now).

drsnyder:
@sqreek, this is what I would propose for a namespace hierarchy:

```
Domain/
   User/
       PreferenceValue.php
Service/
   User/
       Preference.php
       PreferenceService.php # interface
   Auth/ # hypothetical
   Article/
Gateway/
   User/
       PreferenceKeyValueGateway.php
Model/
    UserProperties.php # CaMelCaSed after the table name
```

>  I will move Domain\User to Service\Domain\User and keep all related with service stuff in one namespace. All services stuff is in one space. and there will not mess up with Logger, Vighnette or Util.

To me, the "domain" stands along side the the "service". Also, I think consul related logic can probably be part of `Util`. It will be needed by any service gateway, but that's all as far as I can tell.

>  If I got it right User/ should contain all services related to User?, So going this way if we want to add service related to Article, it will be Article/ so going this way futher, if some service connect to article of this some user, I need to check it in to places. I would prefer one flat folder structure for each of service keep in separate folder/namespace

I think the above supports that. An `Article` service would be at the top level of the `Service` directory. See the example structure above. Most of what we are doing now has zero "domain" logic to it since we are starting with the preferences. If things get more complicated I think it will be more consistent to have the "domain" separated from the "service" since any given service might cut across multiple domains. To say that differently, the domain logic should be invariant across services.

lib/Wikia/src/Service/User/Preference.php
```diff
@@ -0,0 +1,62 @@
...
+		if (!is_array($result)) {
+			return [];
+		}
+
+		$preferences = $this->gatewayResultToPreferenceArray( $result );
+
+		return $preferences;
+	}
+
+	public function gatewayResultToPreferenceArray(array $result) {
```
nmonterroso: I think this should be the responsibility of the gateway, since this is assuming a json string coming back. In swagger, for instance, results are de-serialized into domain objects, so `resultToPreferenceArray` there would be something like:

```php
foreach ($result as $p) {
  $preferences[] = new Preference($p->getName(), $p->getValue());
}
```

nmonterroso: Though I guess the `Swagger` stuff implements [`ArrayAccess`](http://php.net/manual/en/class.arrayaccess.php) so technically it would work, but I think the reasoning still stands.

lib/Wikia/src/Service/User/PreferencePersistence.php
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Wikia\Service\User;
+
+interface PreferencePersistence {
+
+	/**
+	 * Save the users preferences.
+	 *
+	 * @param int $userId
+	 * @param array [\Wikia\Domain\User\Preference, ... ]
```
nmonterroso: This can be properly annotated with `Preference[]` to indicate an array of `Preference` objects

lib/Wikia/src/Service/User/Preference.php
```diff
@@ -0,0 +1,62 @@
...
+		$preferences = [];
+		foreach( $result as $index => $row ) {
+			if (isset($row["name"]) && isset($row["value"])) {
+				$preferences[] = new Preference($row["name"], $row["value"]);
```
nmonterroso: shouldn't this be `new PreferenceValue`?

lib/Wikia/src/Service/User/PreferenceService.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace Wikia\Service\User;
+
+interface PreferenceService {
```
nmonterroso: can there also be PHPDoc on these methods?

lib/Wikia/tests/Service/User/PreferenceTest.php
```diff
@@ -0,0 +1,102 @@
...
+			->method( 'get' )
+			->with( $this->userId )
+			->willReturn( [
+			[ $this->testPreference->getName() => $this->testPreference->getValue() ]
+			] );
+
+		$service = new Preference( $this->gatewayMock );
+		$preferences = $service->getPreferences( $this->userId );
+
+		$this->assertTrue( is_array($preferences), "expecting an array" );
```
nmonterroso: I think this should actually iterate over the array to make sure they are `PreferenceValue` objects. See https://github.com/Wikia/app/pull/7572/files#r32981885

lib/Wikia/src/Service/User/Preference.php
sqreek: That's why In my opinion we should use `Preference` as domain object (instead PreferenceValue). It's very missleading

drsnyder: Ah, ok. I changed it in 8089955

lib/Wikia/src/Service/User/PreferenceService.php
drsnyder: done

nmonterroso:
LGTM :ship:

PR #7573, opened 2015-06-18
CE-1378. Restrict which namespaces appear in dropdown for move page.

perjg:
Added hook for filtering restricted Namespaces that should not appear in the dropdown on move page. See list in CE-1378 for details.

Notes:
- Removed unused namespace hook from VideoHandlerHooks (MWNamespace:isMovable) as it was always returning true without modifying any data. Reused the same hook name.
- Namespaces are hardcoded by number, as most of the restricted namespaces are in extensions.

@Wikia/sustaining-team


owend:
I fixed the conflicts.  SUS team will review it this week and decide if we want to do this or some other approach.

includes/wikia/Wikia.php
```diff
@@ -2448,6 +2451,32 @@ public static function getCountryNames( array $countryCodes ) {
...
+	 * @see CE-1378
+	 *
+	 * @param index Index of namespace
+	 * @return bool
+	 */
+	static public function isNamespaceMovable( &$result, $index ) {
+		// namespace ids from extensions are hardcoded here
+		$filter = array(
+			NS_USER_TALK, // 3 - User_talk
+			501, // User_blog_comment
```
ghost, line 2466: `extensions/wikia/Blogs/Blogs.php:define( "NS_BLOG_ARTICLE_TALK", 501 );`

includes/wikia/Wikia.php
```diff
@@ -2448,6 +2451,32 @@ public static function getCountryNames( array $countryCodes ) {
...
+			700, // Top_10_list
```
ghost, line 2467: `extensions/wikia/TopLists/TopLists_setup.php:define( 'NS_TOPLIST', 700 );`

includes/wikia/Wikia.php
```diff
@@ -2448,6 +2451,32 @@ public static function getCountryNames( array $countryCodes ) {
...
+			1200, // Message_Wall
```
ghost, line 2468: `extensions/wikia/Wall/WallNamespaces.php:define( "NS_USER_WALL", 1200 );`

includes/wikia/Wikia.php
```diff
@@ -2448,6 +2451,32 @@ public static function getCountryNames( array $countryCodes ) {
...
+			1201, // Thread
```
ghost, line 2469: `extensions/wikia/Wall/WallNamespaces.php:define( "NS_USER_WALL_MESSAGE", 1201 );`

includes/wikia/Wikia.php
```diff
@@ -2448,6 +2451,32 @@ public static function getCountryNames( array $countryCodes ) {
...
+			2000, // Board
```
ghost, line 2470: `extensions/wikia/Forum/Forum.namespace.setup.php:define( "NS_WIKIA_FORUM_BOARD", 2000 );`

includes/wikia/Wikia.php
```diff
@@ -2448,6 +2451,32 @@ public static function getCountryNames( array $countryCodes ) {
...
+			2001  // Board_Thread
```
ghost, line 2471: `extensions/wikia/Forum/Forum.namespace.setup.php:define( "NS_WIKIA_FORUM_BOARD_THREAD", 2001 );`

includes/wikia/Wikia.php
```diff
@@ -2448,6 +2451,32 @@ public static function getCountryNames( array $countryCodes ) {
...
+	 * Hide certain namespaces from list of movable ones
+	 *
+	 * @author pgroland
+	 * @see CE-1378
+	 *
+	 * @param index Index of namespace
+	 * @return bool
+	 */
+	static public function isNamespaceMovable( &$result, $index ) {
+		// namespace ids from extensions are hardcoded here
```
ghost, line 2463: Is there any reason of having the namespaces hardcoded here?

The only reason I see, is that the named constants can only be used when corresponding extensions are enabled (and the code defining those constants is included).

I would refactor the entire thing so that the extensions in question automatically add their namespaces to the list of movable namespaces (e.g. `wgMovableNamespaces`). But that touches on the topic of the global state - which is kind of evil.

Another option is to have extensions' own functions hooked to `MWNamespace:isMovable` event covering only the relevant extensions' namespaces. It should be simple as the only thing we do here is to modify `$result`. Depending on the context, the execution of the hook could be interrupted once matching value is reached.

If that's too much, at least include named constants as comments, so that they appear in `grep` results. I don't now how IDEs will handle that, though (in terms of 'check usage' function).

owend, line 2463: Good point, Wikia.php is a global "utility class" loaded everywhere and not an extension.  I have seen we usually use the integer values when the extention is conditionally loaded (with a comment saying what it actually is like in this case). 

We could define a global $wgMovableNamespaceFilter that extensions could add to and this function could use.  Using globals "for great good" I suppose?

macbre:
See #12275

PR #7574, opened 2015-06-19
INT-102: Allow image file upload without login if wiki content is Jap…

mari-mcmurtrie:
…anese.


inez:
This looks good.
There was no need for similar change in permanent api?

mari-mcmurtrie:
Thanks Inez. Yes "addmediapermanet" does not require login credential.

PR #7575, opened 2015-06-19
PLATFORM-1237 Vignette — remove unwanted default upsampling

ghost:
# PLATFORM-1237

Two new modes have been added to Vignette: `scale-to-width-down` and `scale-to-height-down`. Now, on the MediaWiki side, we want to use `scale-to-width-down` rather than `scale-to-width` to avoid upsampling (when thumbnails are larger than originals).

The `scale-to-height-down` is now not supported/used on the MW side.
- https://wikia-inc.atlassian.net/browse/PLATFORM-1237
## Changes
- add constants representing the new mode
- add upsampling as an option to `scaleToWidth` and set no upsampling the default behaviour
- update `VignetteRequest`, `VignetteUrlToUrlGenerator` and `UrlGenerator` for `MODE_SCALE_TO_WIDTH_DOWN`
- update unit tests
- finally, fix all the failing tests elsewhere :-)
  - ServiceTest
  - AvatarServiceTest
  - VignetteRequestTest
- actually run Vignette tests on every push, tej
## Code reviewers

@drsnyder @macbre 
## Stakeholders

@RafLeszczynski @SebastianMarzjan

### DIFF
--- a/tests/phpunit.xml
+++ b/tests/phpunit.xml
@@ -12,6 +12,9 @@
 	<testsuite name="Wikia Test Suite">
 		<directory>./unit/</directory>
 	</testsuite>
+	<testsuite name="Wikia Lib Test Suite">
+		<directory>../lib/Wikia/tests/</directory>
+	</testsuite>
 	<testsuite name="Wikia Includes Test Suite">
 		<directory>../includes/wikia/</directory>
 	</testsuite>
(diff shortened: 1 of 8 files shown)


tests/phpunit.xml
```diff
@@ -12,6 +12,9 @@
 	<testsuite name="Wikia Test Suite">
 		<directory>./unit/</directory>
 	</testsuite>
+	<testsuite name="Wikia Lib Test Suite">
+		<directory>../lib/Wikia/tests/</directory>
+	</testsuite>
```
macbre: :+1:

macbre:
LGTM

PR #7576, opened 2015-06-19
DAT-2912 Support empty ParserOutput in ApiParse

sebastianmarzjan:
https://kibana.wikia-inc.com/index.html#/dashboard/elasticsearch/DAT-2912

### DIFF
--- a/includes/api/ApiParse.php
+++ b/includes/api/ApiParse.php
@@ -204,6 +204,18 @@ class ApiParse extends ApiBase {
 			$result_array['redirects'] = $redirValues;
 		}
 
+		/* Wikia Change Starts */
+		if ( !( $p_result instanceof ParserOutput ) ) {
+			\Wikia\Logger\WikiaLogger::instance()->error(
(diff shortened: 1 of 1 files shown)


SebastianMarzjan:
cc @macbre

macbre:
lgtm

macbre:
lgtm

inez:
How is this PR solving the real issue here? `$p_result` is not what code expects it to be but where the idea of setting it to `new ParserOutput('');` comes from?
Was it tested? Clearly after that change API is returning empty response instead of an error but it is still a wrong response.

sebastianmarzjan:
It's an intermediate step (yeah, the comment would be great there) to identify cases in which it happens:
https://kibana.wikia-inc.com/index.html#/dashboard/elasticsearch/DAT-2912

Is your question related to https://wikia-inc.atlassian.net/browse/PLATFORM-1355 ?
What kind of information do you need?

PR #7579, opened 2015-06-19
PLATFORM-1292 use wgSpecialsDB

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-1292
- always connect to a proper database when querying `phalanx_stats` and `multilookup` tables - the fact that `specials` DB is on the same machine as `stats` DB is just the coincidence and a subject to change
- use `$wgSpecialsDB` global instead of hardcoded `"specials"` string

@wladekb / @michalroszka


michalroszka:
:+1:

PR #7580, opened 2015-06-19
ADEN-1127 Update postscribe code to the newest version

gabrys:


resources/wikia/modules/scriptwriter.js
```diff
@@ -68,10 +68,14 @@ define('wikia.scriptwriter', [
 			win.postscribe(
 				params.element,
 				params.html,
-				{beforeWrite: psBeforeWrite, done: params.callback}
+				{
+					autoFix: false,
```
nandy-andy: Is there postscribe documentation somewhere? I wonder if we could leave a comment here explaining what's the `autoFix` is.

nandy-andy:
LGTM :+1:

PR #7583, opened 2015-06-19
CE-2048: Add support for Template:FlagName

lukaszkonieczny:


kamilkoterba:
LGTM

extensions/wikia/Flags/FlagsExtractor.class.php
```diff
@@ -109,6 +111,8 @@ public function getTemplate() {
 		$template = [];
 		$templateParams = [];
 
+		$nsPrefix = $wgContLang->getNsText( NS_TEMPLATE ) . ':';
```
Grunny: Just a consideration, both the localised namespace and "Template:" will work, so while it may be an edge case, might it be better to cover both?

PR #7584, opened 2015-06-19
CE-2048: Added support for common Template namespace

lukaszkonieczny:


extensions/wikia/Flags/FlagsExtractor.class.php
```diff
@@ -531,6 +524,39 @@ public function findTemplatePosition( $templateName, $offset ) {
 		return false;
 	}
 
+	private function getNextTemplateFormat() {
+		global $wgContLang;
+
+		$nsPrefix = $wgContLang->getNsText( NS_TEMPLATE ) . ':';
+		$nsPrefixCommon = 'Template:';
```
adamkarminski: You can probably use `Namespace::getCanonicalName( NS_TEMPLATE )` here. :)

extensions/wikia/Flags/FlagsExtractor.class.php
```diff
@@ -531,6 +524,39 @@ public function findTemplatePosition( $templateName, $offset ) {
 		return false;
 	}
 
+	private function getNextTemplateFormat() {
```
adamkarminski: I found this name confusing, since it's about a current template. What about simply `getTemplateFormat`? :)

adamkarminski:
Besides the 2 comments - LGTM :+1:

PR #7586, opened 2015-06-19
SOC-533 Update Category Add emails to use new templates

jsutterfield:
Ticket: https://wikia-inc.atlassian.net/browse/SOC-533


extensions/wikia/Email/Controller/CategoryAddController.class.php
```diff
@@ -0,0 +1,129 @@
+<?php
+
+namespace Email\Controller;
+
+use \Email\EmailController;
+use Email\Check;
+
+class CategoryAddController extends EmailController {
```
kvas-damian: Please make sure this email is tracked under right category in sendgrid.

extensions/wikia/Email/Controller/CategoryAddController.class.php
```diff
@@ -0,0 +1,129 @@
...
+
+	/** @var \Title */
+	private $pageAddedToCategory;
+
+	public function initEmail() {
+		$pageTitle = $this->request->getVal( 'pageTitle' );
+		$titleNamespace = $this->request->getVal( 'namespace', NS_MAIN );
+		$categoryPageId = $this->getVal( 'childArticleID' );
+
+		$this->pageAddedToCategory = \Title::newFromText( $pageTitle, $titleNamespace );
```
kvas-damian: I'm not sure if it's an issue here, but please check if there isn't possibility to have master/slave replication errors. I suspect this method called in the same request as page is created.

extensions/wikia/Email/Controller/CategoryAddController.class.php
```diff
@@ -0,0 +1,129 @@
...
+	protected function getSubject() {
+		return $this->getMessage( 'emailext-categoryadd-subject', $this->categoryPage->getText() )->parse();
+	}
+
+	protected function getDetails() {
+		return $this->getMessage( 'emailext-categoryadd-details' )->text();
+	}
+
+	protected function getContentFooterMessages() {
+		return $this->getMessage( 'emailext-categoryadd-footer-1',
```
kvas-damian: What ```1``` means in this message key? Can we make it more descriptive?

extensions/wikia/Follow/FollowEmailTask.class.php
```diff
@@ -35,4 +33,16 @@ public function emailFollowNotifications( $initiatingUser, $aWatchers, $iUserId,
...
+	private function logWatchers( $watchers, $action ) {
+		if ( $action == FollowHelper::LOG_ACTION_BLOG_POST ) {
+			$msg = 'WatchlistLogs: Sending bloglisting watchlist updates';
+		} elseif ( $action == FollowHelper::LOG_ACTION_CATEGORY_ADD ) {
+			$msg = 'WatchlistLogs: Sending categoryadd watchlist updates';
+		} else {
+			$msg = 'WatchlistLogs: Sending other watchlist updates';
+		}
+
+		$this->info( $msg, [ 'watchedPages' => $watchers, 'action' => $action ] );
```
kvas-damian: I'm not sure about naming here. ```watchedPages``` vs ```watchers```. I know pages are use for keys but in values we still store watchers.

extensions/wikia/Email/Controller/CategoryAddController.class.php
jsutterfield: It is. The Category Add emails use the "UserMailer" category which is the default inherited from `EmailController`.

extensions/wikia/Email/Controller/CategoryAddController.class.php
jsutterfield: I don't think this should be too much of a problem since I assume the majority of categories will be added after a page has already been created, but I can add a check to the master DB just in case since we have run into that problem with previous emails.

extensions/wikia/Follow/FollowEmailTask.class.php
jsutterfield: Yea, that is kind of a confusing/misleading name. I'll updated to `watchers`.

extensions/wikia/Email/Controller/CategoryAddController.class.php
```diff
@@ -0,0 +1,133 @@
+<?php
+
+namespace Email\Controller;
+
+use \Email\EmailController;
```
mklucsarits: This beginning slash is not needed.

mklucsarits:
LGTM whether or not you want to make that one suggested change.

jsutterfield:
Note, the failed unit is from that annoying GameStarApiWrapper test which fails every now and again. I reran it and everything passed: http://qa-s2.wikia-prod:8080/job/unitTests_php_branches/16669/

extensions/wikia/Email/Controller/CategoryAddController.class.php
kvas-damian: Hmm it's possible to create a page with categories in one request (in old editor for sure), so I think there could be a cases like this.
I see it's already fixed :+1:

extensions/wikia/Email/Controller/CategoryAddController.class.php
kvas-damian: Ohh, I missed that in the ticket.

kvas-damian:
LGTM

PR #7588, opened 2015-06-19
HG-765: Setup session for edits done within API

inez:
This is important fo Mercury - in which edits are done within API, and having session_cookie is important to know, at cache layer, so skip caches and serve fresh content.


kenkouot:
LGTM

PR #7590, opened 2015-06-22
DAT-2915 Pass only real Title object to viewRedirect() method

sebastianmarzjan:
Otherwise, the Article code fails on trying to execute method on a null variable.


idradm:
LG

PR #7591, opened 2015-06-22
PLATFORM-1248 Recursive RequestContext::getLanguage() calls bug

ghost:
https://wikia-inc.atlassian.net/browse/PLATFORM-1248

### DIFF
--- a/includes/context/RequestContext.php
+++ b/includes/context/RequestContext.php
@@ -270,7 +270,7 @@ class RequestContext implements IContextSource {
 				[
 					'caller'    => wfGetAllCallers(),
 					'exception' => new Exception(),
-					'language'  => json_encode( $l )
+					'language'  => serialize( $l )
 				]
 			);
 		}
@@ -315,7 +315,7 @@ class RequestContext implements IContextSource {
 				[
 					'caller'    => wfGetAllCallers(),
 					'exception' => new Exception(),
-					'language'  => json_encode( $this->lang )
+					'language'  => serialize( $this->lang )
 				]
 			);
 		} elseif ( $this->lang === null ) {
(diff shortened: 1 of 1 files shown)


includes/context/RequestContext.php
```diff
@@ -256,6 +262,19 @@ public function setLang( $l ) {
 	 * @since 1.19
 	 */
 	public function setLanguage( $l ) {
+
+		// PLATFORM-1248 Recursive calls of RequestContext::getLanguage() michal@wikia-inc.com
+		if ( ( new Wikia\Util\Statistics\BernoulliTrial( 0.1 ) )->shouldSample() ) {
```
michalroszka: This is just for starters, as I don't know how much data should I expect here.

includes/context/RequestContext.php
```diff
@@ -284,16 +303,29 @@ public function getLang() {
...
 			$code = ( $wgLanguageCode ) ? $wgLanguageCode : 'en';
 			$this->lang = Language::factory( $code );
+
+			// PLATFORM-1248 Recursive calls of RequestContext::getLanguage() michal@wikia-inc.com
+			Wikia\Logger\WikiaLogger::instance()->debug(
+				__METHOD__,
+				[
+					'caller'    => wfGetAllCallers(),
+					'exception' => new Exception(),
+					'language'  => serialize( $this->lang )
```
macbre: Serialized `Language` can be quite long. Won't the following be enough here:

```php
is_object($this->lang) ? get_class($this->lang) : gettype($this->lang)
```

michalroszka: Per our conversation - it won't. But I only need a little sample of data.

PR #7593, opened 2015-06-22
CE-2064 Preserve whitelisted magic words for later processing

adamkarminski:
@Wikia/community-engineering 

See [CE-2064](https://wikia-inc.atlassian.net/browse/CE-2064)

This PR prevents the `__FLAGS__` magic word (and ultimately - any whitelisted one as well) being removed from a content before it can be replaced by portable flags (or processed later).

### DIFF
--- a/includes/MagicWord.php
+++ b/includes/MagicWord.php
@@ -197,7 +197,12 @@ class MagicWord {
 	);
 
 	static public $mObjects = array();
+
 	static public $mDoubleUnderscoreArray = null;
+
+	static public $mWhitelistedIDs = [
+		'flags',
+	];
 
 	/**#@-*/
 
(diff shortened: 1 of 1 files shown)


includes/MagicWord.php
```diff
@@ -266,7 +271,17 @@ static function getCacheTTL( $id ) {
...
 	static function getDoubleUnderscoreArray() {
 		if ( is_null( self::$mDoubleUnderscoreArray ) ) {
-			self::$mDoubleUnderscoreArray = new MagicWordArray( self::$mDoubleUnderscoreIDs );
+			self::$mDoubleUnderscoreArray = new MagicWordArray(
+				/**
+				 * Wikia change begin
+				 *
+				 * Allow different behavior for some magic words from a white list
+				 */
+				array_diff( self::$mDoubleUnderscoreIDs, self::$mWhitelistedIDs )
```
Grunny: This is the only place `$mDoubleUnderscoreIDs` is used, so why not just remove flags from `$mDoubleUnderscoreIDs` which you added earlier https://github.com/Wikia/app/blob/92b28ea61b0/includes/MagicWord.php#L180?

PR #7597, opened 2015-06-22
ADEN-1935 URL params for Mercury

nandy-andy:
As `AdContext` is more widely used (Oasis, Mercury) we decided to move there some URL params logic. Especially, we were interested in `force[provider]=1` logic, so we can test specific providers on Mercury.

The main changes:
- `force[provider]=1` changed to `forcead=[provider]` and moved to `AdContext`,
- cleanup in `AdEngine2Hooks`,
- cleanup in unit tests.

### DIFF
--- a/extensions/wikia/AdEngine/AdEngine2ContextService.class.php
+++ b/extensions/wikia/AdEngine/AdEngine2ContextService.class.php
@@ -73,7 +73,7 @@ class AdEngine2ContextService {
 					'exitstitialRedirectDelay' => $wg->OutboundScreenRedirectDelay,
 					'invisibleHighImpact' => $wg->AdDriverEnableInvisibleHighImpactSlot,
 				] ),
-				'forcedAdProvider' => $this->getForcedAdProvider(),
+				'forcedAdProvider' => $wg->AdDriverForcedProvider
 			];
 		} );
 	}
@@ -86,23 +86,5 @@ class AdEngine2ContextService {
 			}
 		}
 		return $output;
-	}
-
-	private function getForcedAdProvider() {
-		$wg = F::app()->wg;
-
-		$possibleForcedAdProviders = [
-			'AdDriverForceLiftiumAd' => 'liftium',
-			'AdDriverForceOpenXAd' => 'openx',
-			'AdDriverForceTurtleAd' => 'turtle'
-		];
-
-		foreach( $possibleForcedAdProviders as $enableFlag => $provider) {
-			if( $wg->$enableFlag === true ) {
-				return $provider;
-			}
-		}
-
-		return null;
 	}
 }
--- a/extensions/wikia/AdEngine/AdEngine2ContextService.class.php
+++ b/extensions/wikia/AdEngine/AdEngine2ContextService.class.php
@@ -73,7 +73,6 @@ class AdEngine2ContextService {
 					'exitstitialRedirectDelay' => $wg->OutboundScreenRedirectDelay,
 					'invisibleHighImpact' => $wg->AdDriverEnableInvisibleHighImpactSlot,
 				] ),
-				// TODO: make it like forceadprovider=liftium
 				'forceProviders' => $this->filterOutEmptyItems( [
 					'liftium' => $wg->AdDriverForceLiftiumAd,
 					'openX' => $wg->AdDriverForceOpenXAd,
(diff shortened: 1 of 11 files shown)


extensions/wikia/AdEngine/AdEngine2ContextService.class.php
```diff
@@ -92,4 +87,22 @@ private function filterOutEmptyItems( $input ) {
 		}
 		return $output;
 	}
+
+	private function getForcedAdProvider() {
+		$wg = F::app()->wg;
+
+		$possibleForcedAdProviders = [
```
gabrys: Maybe let's use one wg var for this as well? :-)

extensions/wikia/AdEngine/AdEngine2Hooks.class.php
```diff
@@ -20,26 +20,19 @@ class AdEngine2Hooks {
...
 		$wgNoExternals = $request->getBool( 'noexternals', $wgNoExternals );
-		$wgLiftiumOnLoad = $request->getBool( 'liftiumonload', (bool)$wgLiftiumOnLoad );
 
-		$wgAdDriverForceLiftiumAd = $request->getBool( 'forceliftium', $wgAdDriverForceLiftiumAd );
-		$wgAdDriverForceOpenXAd = $request->getBool( 'forceopenx', $wgAdDriverForceOpenXAd );
-		$wgAdDriverForceTurtleAd = $request->getBool( 'forceturtle', $wgAdDriverForceTurtleAd );
-
-		$wgEnableKruxTargeting = !$wgNoExternals && $wgEnableKruxTargeting;
-		$wgEnableKruxOnMobile = $request->getBool( 'enablekrux', $wgEnableKruxOnMobile && !$wgNoExternals );
+		if( $wgNoExternals ) {
```
gabrys: I think you need a space after if

extensions/wikia/AdEngine/js/AdContext.js
```diff
@@ -44,7 +46,7 @@ define('ext.wikia.adEngine.adContext', [
 		context.slots = context.slots || {};
 		context.targeting = context.targeting || {};
 		context.providers = context.providers || {};
-		context.forceProviders = context.forceProviders || {};
+		context.forcedAdProvider = qs.getVal('forcead', null) || context.forcedAdProvider || null;
```
gabrys: `s/forceAdProvider/forceProvider/` ?

extensions/wikia/AdEngine/js/AdContext.js
```diff
@@ -68,10 +70,20 @@ define('ext.wikia.adEngine.adContext', [
 		}
 
 		// Turtle
-		if (context.forceProviders.turtle) {
+		if (context.forcedAdProvider === 'turtle') {
 			context.providers.turtle = true;
 		}
 
+		// OpenX
+		if (context.forcedAdProvider === 'openx') {
```
gabrys: I don't think this is needed at all.

extensions/wikia/AdEngine/AdEngine2ContextService.class.php
nandy-andy: Yeah, it would definitely make things easier ;)

extensions/wikia/AdEngine/js/AdContext.js
nandy-andy: or `s/forcedprovider/forcedadprovider/` in query param - although, it makes sense to keep them short and convenient to use, I guess ;)

extensions/wikia/AdEngine/js/config/desktop.js
```diff
@@ -67,13 +67,21 @@ define('ext.wikia.adEngine.config.desktop', [
...
+			log(['getProvider', slotName, 'OpenX (wgAdDriverForcedProvider)'], 'info', logGroup);
 			return [adProviderOpenX];
 		}
 
-		if (context.forceProviders.liftium) {
+		// Force Liftium
+		if (context.forcedProvider === 'liftium') {
+			log(['getProvider', slotName, 'Liftium (wgAdDriverForcedProvider)'], 'info', logGroup);
 			return [adProviderLiftium];
 		}
```
fraszczakszymon: I wonder if we could force one of all providers in `for` loop instead of multiple `if` statements, then it would be easier to test another providers also, but it was never needed so it's just an idea.

fraszczakszymon:
LGTM :+1:

PR #7600, opened 2015-06-22
DAT-2818 Infoboxes field mutators

dianafa:
https://wikia-inc.atlassian.net/browse/DAT-2818

### DIFF
--- a/extensions/wikia/PortableInfobox/services/Parser/Nodes/Node.php
+++ b/extensions/wikia/PortableInfobox/services/Parser/Nodes/Node.php
@@ -224,13 +224,19 @@ class Node {
 	 *
 	 */
 	protected function extractSourceFromNode( \SimpleXMLElement $xmlNode ) {
-		$source = $this->getXmlAttribute( $xmlNode, self::DATA_SRC_ATTR_NAME );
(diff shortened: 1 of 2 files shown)


extensions/wikia/PortableInfobox/services/Parser/Nodes/Node.php
```diff
@@ -145,7 +146,10 @@ protected function getSourceForChildren() {
 	protected function getValueWithDefault( \SimpleXMLElement $xmlNode ) {
 		$value = $this->extractDataFromSource( $xmlNode );
 		if ( !$value && $xmlNode->{self::DEFAULT_TAG_NAME} ) {
```
idradm: Maybe we could switch places for those two ifs, to make return statement less confusing

idradm:
Could you add test for getSource testing for format tag?

extensions/wikia/PortableInfobox/services/Parser/Nodes/Node.php
dianafa: added return statement

idradm:
good :shipit:

PR #7601, opened 2015-06-22
DAT-2920 clean up thumbnail creating logic for portable infoboxes - special ca…

RafLeszczynski:
…se for Vignette no longer needed


dianafa:
LGTM

PR #7602, opened 2015-06-22
CE-2087 Execute log tasks right away 

kamilkoterba:
- don't queue it due to job queue clog

https://wikia-inc.atlassian.net/browse/CE-2087

@Wikia/community-engineering


lgarczewski:
LGTM.

PR #7604, opened 2015-06-22
SOC-927 Don't try to call FacebookClient code if the extension is not enabled

garthwebb:
https://wikia-inc.atlassian.net/browse/SOC-927

### DIFF
--- a/includes/specials/SpecialUserlogout.php
+++ b/includes/specials/SpecialUserlogout.php
@@ -44,8 +44,11 @@ class SpecialUserlogout extends UnlistedSpecialPage {
 			throw new HttpError( 400, wfMessage( 'suspicious-userlogout' ), wfMessage( 'loginerror' ) );
 		}
 
-		// Clean up any facebook cookies/data
-		FacebookClient::getInstance()->logout();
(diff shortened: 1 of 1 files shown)


mklucsarits:
LGTM

includes/specials/SpecialUserlogout.php
```diff
@@ -44,8 +44,11 @@ function execute( $par ) {
 			throw new HttpError( 400, wfMessage( 'suspicious-userlogout' ), wfMessage( 'loginerror' ) );
 		}
 
-		// Clean up any facebook cookies/data
-		FacebookClient::getInstance()->logout();
+		// Some communities (notably uncyclopedia) do not have FacebookClient enabled
+		if ( F::app()->wg->EnableFacebookClientExt ) {
+			// Clean up any facebook cookies/data
+			FacebookClient::getInstance()->logout();
+		}
```
macbre: Please use a [hook](https://www.mediawiki.org/wiki/Manual:Hooks/UserLogoutComplete) to avoid adding our code (wrapped in an if statement) into MW code.

PR #7585, opened 2015-06-19
SOC-735 Implement Forgot Password request email.

garthwebb:
https://wikia-inc.atlassian.net/browse/SOC-735

PR #7606, opened 2015-06-22
SOC-766 New founder digest emails

mklucsarits:
https://wikia-inc.atlassian.net/browse/SOC-766

This pull request includes some changes to the founder email event classes from SOC-466, to prevent merge conflicts.


extensions/wikia/Email/i18n/FounderDigest.i18n.php
```diff
@@ -0,0 +1,35 @@
+<?php
+$messages = array();
+
+$messages['en'] = array(
+	'emailext-founder-activity-digest-subject' => 'The latest activity on {{SITENAME}}',
+	'emailext-founder-views-digest-subject' => "Today's views on {{SITENAME}}",
+	'emailext-founder-digest-summary' => "It's time for your daily dose of activity from {{SITENAME}}.",
+	'emailext-founder-activity-digest-link-label' => 'Recent Wikia Activity',
+	'emailext-founder-activity-digest-footer' => 'Get more tips on building your wikia and ask questions on [$1 Community Central].',
```
jsutterfield: Garth constructed a similar link in the last email ticket he worked on. Apparently we have to hardcode these community URLS and have the translators do the same in order to have them localized. [Here's](https://github.com/Wikia/app/pull/7585/files#diff-239e1d7b95191079f1350ff59d25599fR18) the discussion about it.

extensions/wikia/Email/i18n/FounderDigest.i18n.php
```diff
@@ -0,0 +1,35 @@
...
+	'emailext-founder-digest-summary' => 'Summary text in body of email for founder digests.',
+	'emailext-founder-activity-digest-link-label' => 'Button label linking to recent wiki activity page',
+	'emailext-founder-activity-digest-footer' => 'Footer text for activity digest that suggests visiting the Community Central wiki for help',
+	'emailext-founder-views-digest-link-label' => 'Button label linking to the wiki inviting the user to add new pages.',
+	'emailext-founder-digest-views-header' => 'Header text for number of pages viewed. $1 -> number of page views',
+	'emailext-founder-digest-views-description-1' => 'Description text for the page views header in the activity digest',
+	'emailext-founder-digest-views-description-2' => 'Alternate description text for the page views header in the page views digest',
+	'emailext-founder-digest-edits-header' => 'Header text for number of edits made',
+	'emailext-founder-digest-edits-description' => 'Description text for the number-of-edits header',
+	'emailext-founder-digest-users-header' => 'Header text for number of new users',
```
jsutterfield: Missing the `$1 -> what-the-$1-stands-for` part of this qqq.

extensions/wikia/FounderEmails/events/FounderEmailsCompleteDigestEvent.class.php
```diff
@@ -33,52 +24,34 @@ public function process ( Array $events ) {
...
-				$mailSubject = strtr( wfMsgExt( 'founderemails-email-complete-digest-subject', array( 'language' => $langCode ) ), $emailParams );
-				$mailBody = strtr( wfMsgExt( 'founderemails-email-complete-digest-body', array( 'language' => $langCode, 'parsemag' ), $emailParams['$UNIQUEVIEWS'], $emailParams['$USEREDITS'], $emailParams['$USERJOINS'] ), $emailParams );
-				$mailBodyHTML = F::app()->renderView( "FounderEmails", "GeneralUpdate", array_merge( $emailParams, array( 'language' => $langCode, 'type' => 'complete-digest' ) ) );
-				$mailBodyHTML = strtr( $mailBodyHTML, FounderEmails::addLink( $emailParams, $links ) );
-				$mailCategory = FounderEmailsEvent::CATEGORY_COMPLETE_DIGEST . ( !empty( $langCode ) && $langCode == 'en' ? 'EN' : 'INT' );
-
-				// Send the e-mail
-				$founderEmailObj->notifyFounder( $user, $this, $mailSubject, $mailBody, $mailBodyHTML, $cityID, $mailCategory );
+				$emailParams['targetUser'] = $user->getName();
+				F::app()->sendRequest( 'Email\Controller\FounderActivityDigest', 'handle', $emailParams );
```
jsutterfield: What do you think about making the Email Controller a class constant?

extensions/wikia/FounderEmails/events/FounderEmailsViewsDigestEvent.class.php
```diff
@@ -38,41 +33,25 @@ public function process ( Array $events ) {
...
-				);
-				$mailSubject = strtr( wfMsgExt( 'founderemails-email-views-digest-subject', array( 'content' ) ), $emailParams );
-				$mailBody = strtr( wfMsgExt( 'founderemails-email-views-digest-body', array( 'content', 'parsemag' ), $emailParams['$UNIQUEVIEWS'] ), $emailParams );
-				$mailBodyHTML = F::app()->renderView( 'FounderEmails', 'GeneralUpdate', array_merge( $emailParams, array( 'language' => 'en', 'type' => 'views-digest' ) ) );
-				$mailBodyHTML = strtr( $mailBodyHTML, FounderEmails::addLink( $emailParams, $links ) );
-				$mailCategory = FounderEmailsEvent::CATEGORY_VIEWS_DIGEST . ( !empty( $langCode ) && $langCode == 'en' ? 'EN' : 'INT' );
 
-				$founderEmailObj->notifyFounder( $user, $this, $mailSubject, $mailBody, $mailBodyHTML, $cityID, $mailCategory );
+				$emailParams['targetUser'] = $user->getName();
+				F::app()->sendRequest( 'Email\Controller\FounderPageViewsDigest', 'handle', $emailParams );
```
jsutterfield: Same question here about potentially making a class constant.

extensions/wikia/FounderEmails/events/FounderEmailsViewsDigestEvent.class.php
```diff
@@ -5,24 +5,19 @@ public function __construct( Array $data = array() ) {
...
-		}
-
 		// If complete digest mode is enabled, do not send views only digest
-		if ( $user->getOption( "founderemails-complete-digest-$wgCityId" ) ) {
+		if ( $user->getOption( "founderemails-complete-digest-$wikiId" ) ) {
 			return false;
 		}
-		if ( $user->getOption( "founderemails-views-digest-$wgCityId" ) ) {
+		if ( $user->getOption( "founderemails-views-digest-$wikiId" ) ) {
 			return true;
```
jsutterfield: It looks like you pulled these 2 checks into the `FounderPageViewsDigestController` but didn't remove them from here?

extensions/wikia/Email/Controller/FounderDigestController.class.php
```diff
@@ -0,0 +1,307 @@
...
+			] 
+		];
+	}
+
+	public function getButtonText() {
+		return $this->getMessage( 'emailext-founder-views-digest-link-label' )->text();
+	}
+
+	public function getButtonLink() {
+		return \GlobalTitle::newFromText( 'CreatePage', NS_SPECIAL, $this->wikiId )->getFullURL( array( 'modal' => 'AddPage' ) );
```
jsutterfield: So we made it a [Wikia coding convention](https://github.com/Wikia/guidelines/blob/master/PHP/CodingConventions.md#array-syntax) to use the new array syntax, ie, `[ 'modal' => 'AddPage' ]`

extensions/wikia/Email/Controller/FounderDigestController.class.php
```diff
@@ -0,0 +1,307 @@
...
+
+class FounderActivityDigestController extends FounderDigestController {
+	const TRACKING_CATEGORY_EN = TrackingCategories::FOUNDER_ACTIVITY_DIGEST_EN;
+	const TRACKING_CATEGORY_INT = TrackingCategories::FOUNDER_ACTIVITY_DIGEST_INT;
+
+	public function initEmail() {
+		$this->wikiId = $this->request->getVal( 'wikiId' );
+		$this->pageViews = $this->request->getVal( 'pageViews' );
+		$this->pageEdits = $this->request->getVal( 'pageEdits' );
+		$this->newUsers = $this->request->getVal( 'newUsers' );
```
jsutterfield: Can we declared these properties up above before using them?

extensions/wikia/Email/Controller/FounderDigestController.class.php
```diff
@@ -0,0 +1,307 @@
...
+class FounderPageViewsDigestController extends FounderDigestController {
+	const TRACKING_CATEGORY_EN = TrackingCategories::FOUNDER_VIEWS_DIGEST_EN;
+	const TRACKING_CATEGORY_INT = TrackingCategories::FOUNDER_VIEWS_DIGEST_INT;
+
+	public function initEmail() {
+		$this->wikiId = $this->request->getVal( 'wikiId' );
+		$this->pageViews = $this->request->getVal( 'pageViews' );
+		$this->assertValidParams();
+
+		$this->pageViews = number_format( $this->pageViews );
```
jsutterfield: What do you think about pulling this `initEmail()` logic into the base class? Both `FounderPageViewsDigestController` and `FounderActivityDigestController` share this same logic and then you can just extend it in `FounderActivityDigestController`.

extensions/wikia/Email/Controller/FounderDigestController.class.php
```diff
@@ -0,0 +1,307 @@
...
+		$this->assertPageViewsSet();
+	}
+
+	/**
+	 * Asserts that the wikiId parameter is valid
+	 *
+	 * @throws \Email\Check
+	 */
+	protected function assertValidWikiId() {
+		if ( (int) $this->wikiId < 1 ) {
```
jsutterfield: I don't believe you need to do this casting here. PHP does implicit casting if you do a comparison against an int like this.

extensions/wikia/Email/Controller/FounderDigestController.class.php
mklucsarits: Yup, that would be fine. I originally had it that way and changed my mind, but I think having it in the base class makes more sense.

extensions/wikia/Email/Controller/FounderDigestController.class.php
```diff
@@ -0,0 +1,312 @@
...
+	const TRACKING_CATEGORY_EN = TrackingCategories::DEFAULT_CATEGORY;
+	const TRACKING_CATEGORY_INT = TrackingCategories::DEFAULT_CATEGORY;
+
+	protected $wikiId;
+
+	protected $pageViews;
+
+	protected $pageEdits;
+
+	protected $newUsers;
```
jsutterfield: Looks good. The last thing is I think it'd probably make more sense for the declarations for `$newUsers`, and `$pageEdits` to be in `FounderActivityDigestController` since that's the class which actually uses them.

mklucsarits: Yup, you're right.

extensions/wikia/Email/Controller/FounderDigestController.class.php
garthwebb: We should use the facilities in the ```Language``` class rather than rely on PHP's ```number_format``` (see  https://one.wikia-inc.com/wiki/Internationalization/Technical#Numbers)

You can use ```\Language::factory( $this->targetLang );``` to get the right language object.

extensions/wikia/Email/Controller/FounderDigestController.class.php
```diff
@@ -0,0 +1,310 @@
...
+		$this->wikiId = $this->request->getVal( 'wikiId' );
+		$this->pageViews = $this->request->getVal( 'pageViews' );
+		$this->assertValidParams();
+
+		$this->pageViews = number_format( $this->pageViews );
+	}
+
+	protected function getSummary() {
+		return $this->getMessage( 'emailext-founder-digest-summary',
+			\WikiMap::getWikiName( $this->wikiId ) )->parse();
```
garthwebb: Ok, a few things here.  I think ```WikiMap``` is some kind of old MediaWiki attempt at hosting multiple wikis, and as such I don't think it works with our setup.  For getting wiki information you should use the methods in ```WikiFactory```.  The commonly seen ones are things like:

* WikiFactory::IDtoDB - maps a wiki ID to a dbname
* WikiFactory::DBtoDomain - maps a dbname to a domain name
* WikiFactory::DBtoUrl - I think you get the point ...

For this line however, there is no easy, single function to get the wiki title.  The best thing I think would be:
```php
$wikiObj = WikiFactory::getWikiByID( $this->wikiId );
$name = empty( $wikiObj ) ? '' : $wikiObj->city_title;
```

The second thing here is that ```emailext-founder-digest-summary``` does not have a ```$1``` to receive this value.  Its currently using ```{{SITENAME}}``` which will always resolve to the current wiki name.

extensions/wikia/Email/Controller/FounderDigestController.class.php
```diff
@@ -0,0 +1,310 @@
...
+		$this->assertPageViewsSet();
+	}
+
+	/**
+	 * Asserts that the wikiId parameter is valid
+	 *
+	 * @throws \Email\Check
+	 */
+	protected function assertValidWikiId() {
+		if ( $this->wikiId < 1 ) {
```
garthwebb: It might make sense to compare this with ```empty```.  This will work fine, but the values this will likely contend with are ```null``` and empty string, rather than an integer less than one.

extensions/wikia/Email/Controller/FounderDigestController.class.php
```diff
@@ -0,0 +1,310 @@
...
+		}
+	}
+
+	/**
+	 * Asserts that the `pageViews` value is passed
+	 *
+	 * @throws \Email\Check
+	 */
+	protected function assertPageViewsSet() {
+		if ( empty( $this->pageViews ) && $this->pageViews !== '0' ) {
```
garthwebb: The check for ```empty``` should cover the case when this is zero, whether its integer zero or string zero.

extensions/wikia/Email/Controller/FounderDigestController.class.php
```diff
@@ -0,0 +1,310 @@
...
+	protected $pageEdits;
+	protected $newUsers;
+
+	public function initEmail() {
+		$this->pageEdits = $this->request->getVal( 'pageEdits' );
+		$this->newUsers = $this->request->getVal( 'newUsers' );
+		// Parent method is called here so that assertion of parameters is done at the right time
+		parent::initEmail();
+		$this->pageEdits = number_format( $this->pageEdits );
+		$this->newUsers = number_format( $this->newUsers );
```
garthwebb: Same comment for number_format as above.  Maybe we should create a method in the parent class that returns the language object from targetLang so we don't have to construct it every time.

extensions/wikia/Email/Controller/FounderDigestController.class.php
```diff
@@ -0,0 +1,310 @@
...
+	}
+
+	protected function getSubject() {
+		return $this->getMessage( 'emailext-founder-activity-digest-subject',
+			\WikiMap::getWikiName( $this->wikiId ) )->parse();
```
garthwebb: Same as above and same as just above; maybe we should create a new method in WikiFactory that returns the wiki title?

extensions/wikia/Email/Controller/FounderDigestController.class.php
```diff
@@ -0,0 +1,310 @@
...
+		$this->assertPageEditsSet();
+	}
+
+	/**
+	 * Asserts that the `newUsers` value is passed
+	 *
+	 * @throws \Email\Check
+	 */
+	protected function assertNewUsersSet() {
+		if ( empty( $this->newUsers ) && $this->newUsers !== '0' ) {
```
garthwebb: Same as above and on line 168

extensions/wikia/Email/Controller/FounderDigestController.class.php
```diff
@@ -0,0 +1,310 @@
...
+
+	/**
+	 * Returns list of details for the digest
+	 *
+	 * @return array
+	 */
+	protected function getDetailsList() {
+		return [
+			[
+				'iconSrc' => \GlobalFile::newFromText( 'Page-views.png', \Wikia::NEWSLETTER_WIKI_ID )->getUrlGenerator()->url(),
```
garthwebb: Oh!  This is why you were asking about ```GlobalFile```, I didn't realize.  You should use ```EmailImageHelper``` for this.  It constructs the proper URL and caches the result.  You can see it in use within ```EmailController::getBody```

extensions/wikia/Email/Controller/FounderDigestController.class.php
```diff
@@ -0,0 +1,310 @@
...
+	}
+
+	public function getButtonLink() {
+		return \GlobalTitle::newFromText( 'WikiActivity', NS_SPECIAL, $this->wikiId )->getFullURL();
+	}
+
+	protected function getCommunityFooterMessage() {
+		return $this->getMessage(
+			'emailext-founder-activity-digest-footer',
+			\GlobalTitle::newMainPage( \Wikia::COMMUNITY_WIKI_ID )->getFullURL()
```
garthwebb: Looks like you're passing in this value, but it the i18n key has it hard coded.  Its probably fine to just leave it hard coded, or if you'd like to be more dynamic, you can use the wikitext shorthand for the community wiki: ```[[w:c:community|Community Wiki]]```

extensions/wikia/Email/Controller/FounderDigestController.class.php
```diff
@@ -0,0 +1,310 @@
...
+	}
+}
+
+class FounderPageViewsDigestController extends FounderDigestController {
+	const TRACKING_CATEGORY_EN = TrackingCategories::FOUNDER_VIEWS_DIGEST_EN;
+	const TRACKING_CATEGORY_INT = TrackingCategories::FOUNDER_VIEWS_DIGEST_INT;
+
+	protected function getSubject() {
+		return $this->getMessage( 'emailext-founder-views-digest-subject',
+			\WikiMap::getWikiName( $this->wikiId ) )->parse();
```
garthwebb: Same as above

extensions/wikia/Email/Controller/FounderDigestController.class.php
```diff
@@ -0,0 +1,310 @@
...
+
+	/**
+	 * Returns list of details for the digest
+	 *
+	 * @return array
+	 */
+	protected function getDetailsList() {
+		return [
+			[
+				'iconSrc' => \GlobalFile::newFromText( 'Page-views.png', \Wikia::NEWSLETTER_WIKI_ID )->getUrlGenerator()->url(),
```
garthwebb: ```EmailImageHelper```` here

extensions/wikia/Email/Controller/FounderDigestController.class.php
mklucsarits: Yes but '0' is a valid value for page views here. That's why I'm checking that it's not '0'.

extensions/wikia/Email/Controller/FounderDigestController.class.php
mklucsarits: OK, I noticed EmailImageHelper only supports GIF files. So I'll have to change it to accept other image types.

extensions/wikia/Email/Controller/FounderDigestController.class.php
mklucsarits: Thanks for the info. I don't know what I was thinking when I added this. I must have forgot that ```{{SITENAME}}``` can be -- and was -- used.

extensions/wikia/Email/Controller/FounderDigestController.class.php
mklucsarits: Yeah, @jsutterfield suggested that the Community URL should be hardcoded in messages, as it was done for another email, and translated for each language. So this parameter can just be removed.

extensions/wikia/Email/Controller/FounderDigestController.class.php
garthwebb: Ah, my mistake!

extensions/wikia/Email/Controller/FounderDigestController.class.php
garthwebb: Oh, I forgot about that.  Yah, sounds good.

garthwebb:
LGTM

jsutterfield:
Closing this to retarget email-sprint-16 branch. New PR can be found here: https://github.com/Wikia/app/pull/7680

PR #7608, opened 2015-06-22
add swagger autoload and commit client + user prefs api

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-453

Configuration changes in Wikia/config#1188

Adding Swagger-generated client for `user-preference` service. Allows us to do stuff like the following:

``` php
        $config = (new \Swagger\Client\Configuration())
            ->setHost('http://localhost:8080')
            ->setApiKey('X-Wikia-UserId', $userId);

        $client = new \Swagger\Client\ApiClient($config);
        $api = new \Swagger\Client\User\Preferences\Api\UserPreferencesApi($client);

        try {
            print_r($api->getUserPreferences($userId));
            $newPrefs = [
                (new \Swagger\Client\User\Preferences\Models\Preference())->setName("language")->setValue("de"),
                (new \Swagger\Client\User\Preferences\Models\Preference())->setName("marketingallowed")->setValue("0")
            ];
            $api->updateUserPreferences(2, $newPrefs);
            $api->updateUserPreference(2, "date", "ydm");
            print_r($api->getUserPreferences($userId));
        } catch (\Swagger\Client\ApiException $e) {
            print_r($e);
        }
```

which works as you'd expect - prefs (defined in \Swagger\UserPreferences\Models\Preference) are dumped out, modified as a batch, modified as a single pref, and new prefs are fetched, which reflect the changes done. Not shown (but work) during an exception is the ability to call `$e->getResponseObject` will return a `Swagger\UserPreferences\Models\Problem` object (since that's what the doc specifies)

TODO:
- [x] Integrate with `Preference` interfaces once #7572 is merged


lib/Swagger/src/User/Preferences/Api/UserPreferencesApi.php
```diff
@@ -0,0 +1,342 @@
...
+  /** @var \Swagger\Client\ApiClient instance of the ApiClient */
+  private $apiClient;
+
+  /**
+   * @param \Swagger\Client\ApiClient|null $apiClient The api client to use
+   */
+  function __construct($apiClient = null) {
+    if ($apiClient == null) {
+      $apiClient = new ApiClient();
+      $apiClient->getConfig()->setHost('https://localhost/user-preference');
```
ArturKlajnerok: should we use 'services.wikia.com' instead of`localhost` as default ?

lib/Swagger/src/User/Preferences/Api/UserPreferencesApi.php
```diff
@@ -0,0 +1,342 @@
...
+      
+      
+      
+      // make the API Call
+      try {
+        $response = $this->apiClient->callAPI($resourcePath, $method,
+                                              $queryParams, $httpBody,
+                                              $headerParams);
+      } catch (ApiException $e) {
+        switch ($e->getCode()) {
```
ArturKlajnerok: Should we add a 500 http status code for cases when there's a server side problem that prevent accessing the preference ?

lib/Swagger/src/User/Preferences/Api/UserPreferencesApi.php
nmonterroso: No. This is autogenerated code, which gets the `basePath` from the swagger documentation, which doesn't specify a domain, so it defaults to `localhost`

lib/Swagger/src/User/Preferences/Api/UserPreferencesApi.php
nmonterroso: We can add that in the Swagger documentation and regenerate, but I think that can be a separate PR since we don't do anything with the status code in the exceptions thrown

ArturKlajnerok:
Lgtm :+1:

nmonterroso:
re-ran the js tests, and they succeeded but didn't update this page http://qa-s2.wikia-prod:8080/job/unitTests_js_branches/14903/

PR #7611, opened 2015-06-23
CE-2092 Check if params names are an array before supplying it to for…

adamkarminski:
…each

@Wikia/community-engineering

### DIFF
--- a/extensions/wikia/Flags/controllers/templates/SpecialFlagsController_index.php
+++ b/extensions/wikia/Flags/controllers/templates/SpecialFlagsController_index.php
@@ -27,9 +27,16 @@
 				</a>
 			</td>
 			<td class="flags-special-list-item-params">
-				<?php foreach ( json_decode( $flag['flag_params_names'] ) as $name => $description ): ?>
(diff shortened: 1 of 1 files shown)


lukaszkonieczny:
lgtm

extensions/wikia/Flags/controllers/templates/SpecialFlagsController_index.php
```diff
@@ -27,9 +27,16 @@
 				</a>
 			</td>
 			<td class="flags-special-list-item-params">
-				<?php foreach ( json_decode( $flag['flag_params_names'] ) as $name => $description ): ?>
-					<?= $name ?> <small><em><?= $description ?></em></small><br>
-				<?php endforeach; ?>
+				<?php
+					$paramsNames = json_decode( $flag['flag_params_names'] );
+					if ( is_array( $paramsNames ) ) :
```
Grunny: `json_decode` without setting true as the second parameter returns an StdClass object, not an array, so this check will always fail and no parameters will be displayed. Fixed in https://github.com/Wikia/app/pull/7652.

PR #7620, opened 2015-06-23
Wikiamobile sections bug

lizlux:
I noticed this error here http://glee.wikia.com/wiki/I_Just_Can%27t_Stop_Loving_You?useskin=wikiamobile. The error is: `Uncaught SyntaxError: Failed to execute 'querySelectorAll' on 'Document': 'h2[id]:not( [id=""] ), h3[id]:not( [id=""] ), h4[id]:not( [id=""] )' is not a valid selector.` This was causing JS to break and images not to load.

It looks like the code hasn't changed in a while, so I'm not sure when this broke. I'm wondering if chrome (where I tested) used to support spaces inside `:not()` selectors and it stopped.

Anyway, I removed the spaces to get rid of the error and also ran js-beautify in a separate commit. 

@bkoval @hakubo

### DIFF
--- a/extensions/wikia/WikiaMobile/js/sections.js
+++ b/extensions/wikia/WikiaMobile/js/sections.js
@@ -23,7 +23,7 @@ define( 'sections', ['jquery', 'wikia.window'], function ( $, window ) {
 	function getHeaders(){
 		//we switch nodeList to Array to use filter / forEach type methods
 		return Array.prototype.slice.apply( d.querySelectorAll(
-			'h2[id]:not( [id=""] ), h3[id]:not( [id=""] ), h4[id]:not( [id=""] )' ) );
+			'h2[id]:not([id=""]), h3[id]:not([id=""]), h4[id]:not([id=""])' ) );
 	}
 
 	sections = getHeaders();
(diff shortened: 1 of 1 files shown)


extensions/wikia/WikiaMobile/js/sections.js
```diff
@@ -20,10 +20,10 @@ define( 'sections', ['jquery', 'wikia.window'], function ( $, window ) {
 	 * @desc grab all headers (with non-empty id's) on the page
 	 * @return Array
 	 */
-	function getHeaders(){
+	function getHeaders() {
 		//we switch nodeList to Array to use filter / forEach type methods
-		return Array.prototype.slice.apply( d.querySelectorAll(
-			'h2[id]:not( [id=""] ), h3[id]:not( [id=""] ), h4[id]:not( [id=""] )' ) );
+		return Array.prototype.slice.apply(d.querySelectorAll(
+			'h2[id]:not([id=""]), h3[id]:not([id=""]), h4[id]:not([id=""])'));
```
lizlux: This was the actual fix.

hakubo:
LGTM

PR #7621, opened 2015-06-24
Platform 1179

owend:
- add Mail2 vendor lib because Mail is not PHP 5.6 compatible
- switch mail backend to SMTP
- remove some old vars (Schwartz mailer, wgForceSendgrid) 
- requires matching platform-1179 branch from config
- do not merge yet, for code review and testing


owend:
This is the PR for app @macbre

PR #7624, opened 2015-06-24
DAT-2917 fix double preprocess breaking inclusion functionality

idradm:


SebastianMarzjan:
Less is more :smiling_imp: 
:shipit:

PR #7629, opened 2015-06-24
Add base gateway to connect with micro services using consul

sqreek:
[SERVICES-451]


lib/Wikia/src/Service/Gateway/GatewayBase.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
+namespace Wikia\Service\Gateway;
+
+use Http;
+
+class GatewayBase {
```
nmonterroso: I wonder if we still need this class, since for now we'll be relying on Swagger generated code to take care of the http plumbing for us

lib/Wikia/src/Service/Gateway/UrlProviderInterface.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+namespace Wikia\Service\Gateway;
+
+
+interface UrlProviderInterface {
```
nmonterroso: Can we change this interface name to `UrlProvider`?

lib/Wikia/src/Service/Gateway/ConsulUrlProvider.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+namespace Wikia\Service\Gateway;
+
+
+use Http;
+
+class ConsulUrlProvider implements UrlProviderInterface {
+
+	private $healthUrlSchema = '{consulUrl}/v1/health/service/{serviceName}?passing&tag={serviceTag}';
```
drsnyder: Is there a way to specify whether or not we care about stale entries? The `golang` client provides [this feature](https://github.com/hashicorp/consul/blob/master/api/api.go#L27).

lib/Wikia/src/Service/Gateway/GatewayBase.php
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	public function delete( $path, $headers = null ) {
+		return $this->doRequest( "DELETE", $path, $headers );
+	}
+
+	private function doRequest( $method, $path, $headers = null, $data = null ) {
+		$url = $this->getUrl( $path );
+		$options = $this->getHttpRequestOption( $headers, $data );
+		$response = Http::Request( $method, $url, $options );
```
drsnyder: We should inject the HTTP client so we don't have a dependency on MW in `/lib/Wikia`.

drsnyder:
Tests?

Minor comments. OTW, LGTM

lib/Wikia/src/Service/Gateway/GatewayBase.php
nmonterroso: using DEPENDENCY INJECTION?!

nmonterroso:
anything else for this pr? or can we merge it?

PR #7631, opened 2015-06-24
Introduce Dependency Injection

nmonterroso:
@Wikia/services-team @owend 
https://wikia-inc.atlassian.net/browse/SERVICES-456

Introduce dependency injection into MediaWiki using [PHP-DI](http://php-di.org/index.html). This will be useful for us to instantiate and swap out implementations of services easily, and could possibly be used in other placed in MediaWiki (Nirvana, for example).

There's a lot of composer additions here, but the real core changes are in `lib/Wikia/src/DependencyInjection`. There's also implementation of a module for instantiating  `PreferencePersistenceMySQL` in `PreferencePersistenceModuleMySQL`, which serves as an example of how we'd add to the injector in other places.

This also requires configuration changes from Wikia/config#1172


lib/Wikia/src/DependencyInjection/InjectorBuilder.php
```diff
@@ -0,0 +1,56 @@
+<?php
+/**
+ * InjectorBuilder
+ *
+ * <insert description here>
```
drsnyder: Minor: does it make sense to add some header documentation to these?

drsnyder:
LGTM. I think the only thing missing is some documentation on the usage (specifically in the case of testing).

lib/Wikia/src/DependencyInjection/InjectorBuilder.php
nmonterroso: Ooops, that's IntelliJ auto generated stuff when adding a new file.

PR #7632, opened 2015-06-24
SOC-251 Stop Chat from interfering with the getIP method.

garthwebb:
We already have the right IP in the X-Forwarded-For header

https://wikia-inc.atlassian.net/browse/SOC-251


owend:
LGTM

PR #7633, opened 2015-06-24
SOC-886 Tie the ban TTL to the remaining ban time

garthwebb:
The TTL for `getBanInformation` is set at an arbitrary 24hrs.  That only works for bans over 1 day, but even then if the data gets cached near the end of the ban day, the user can end up banned for almost one extra day.

This PR changes the TTL to be the time remaining on the ban.

https://wikia-inc.atlassian.net/browse/SOC-886

### DIFF
--- a/extensions/wikia/Chat2/Chat.class.php
+++ b/extensions/wikia/Chat2/Chat.class.php
@@ -291,17 +291,24 @@ class Chat {
 
 		$userID = $banUser->getId();
 		$key = self::getBanInfoCacheKey( $cityID, $userID );
+		$memc = F::app()->wg->Memc;
 
-		$banInfo = WikiaDataAccess::cache( $key, self::BAN_INFO_TTL, function () use ( $cityID, $userID ) {
-			ChatHelper::info( __METHOD__ . ': Cache miss - querying DB', [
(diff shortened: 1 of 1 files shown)


extensions/wikia/Chat2/Chat.class.php
```diff
@@ -291,17 +291,24 @@ public static function getBanInformation( $cityID, User $banUser ) {
...
+		$banInfo = $memc->get( $key );
+		if ( empty( $banInfo ) ) {
+			$banInfo = self::getBanInfoFromDB( $cityID, $userID );
 
-			return self::getBanInfoFromDB( $cityID, $userID );
-		} );
+			// Only cache for as long as we have left in the ban, or a default if the ban has expired
+			if ( empty( $banInfo ) ) {
+				$ttl = self::BAN_INFO_TTL;
+				$banInfo = -1;
```
owend: Add a comment about the -1 here for the benefit of future code spelunkers?

garthwebb: Good call.  Created a constant for this and left a comment.

owend:
LGTM

PR #7634, opened 2015-06-24
[SERVICES-459] Don't authenticate disabled user with helios

ArturKlajnerok:
This is quick fix for problem:
https://wikia-inc.atlassian.net/browse/SERVICES-459
Remote logging out for disabled users don't invalidate access token with helios, so for now each helios authentication of disabled user will fail by default and provided access token cookie will be cleared. Proper solution that will invalidate tokens for disabled user will be introduced with: 
https://wikia-inc.atlassian.net/browse/SERVICES-208

/cc @Wikia/services-team @Grunny @michalroszka


drsnyder:
LGTM. Looks like some of the tests failed though.

Grunny:
LGTM too. :)

ArturKlajnerok:
Merged in dev, so test won't fail on Fatal error: Class 'Wikia\DependencyInjection\InjectorInitializer'

PR #7635, opened 2015-06-25
(CE-2155) Only show right rail on templates with "infobox" in title

Grunny:
Only show right rail on templates with "infobox" in title for now. Also
check the template exists so we don't show "migrate this infobox" on a
non-existent template page.


Grunny:
Unit test failure is unrelated. Caused by changes in config that need up to date dev. We can just merge the dev branch in.

adamkarminski:
looks good, merging. :)

PR #7637, opened 2015-06-25
ADEN-2164 Krux in specific countries

nandy-andy:
We don't want Krux enabled site-wide. We would like to be able to decide in which country Krux should be enabled.


jacekjursza:
LGTM

PR #7639, opened 2015-06-25
(CE-2150) Add simple unit tests for the infobox conversion

Grunny:
Add simple unit tests for the infobox conversion that can be expanded
as the conversion is improved.

### DIFF
--- a/extensions/wikia/TemplateDraft/tests/TemplateConverterTest.php
+++ b/extensions/wikia/TemplateDraft/tests/TemplateConverterTest.php
@@ -1,0 +1,73 @@
+<?php
+
+class TemplateConverterTest extends WikiaBaseTest {
+
+	public function setUp() {
+		$this->setupFile = __DIR__ . '/../TemplateDraft.setup.php';
+		parent::setUp();
+	}
+
+	/**
+	 * @dataProvider convertAsInfoboxProvider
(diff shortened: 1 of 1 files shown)


extensions/wikia/TemplateDraft/tests/TemplateConverterTest.php
```diff
@@ -0,0 +1,73 @@
...
+				"{{{name|}}}\n{{{name|}}}\\n{{{name}}}",
+				"<infobox>\n\t<title source=\"name\"><default>{{PAGENAME}}</default></title>\n</infobox>\n",
+				'Multiples of the same parameter in the wikitext only add one row',
+			],
+			[
+				'',
+				"<infobox>\n</infobox>\n",
+				'Empty content',
+			],
+			/* @todo: Common case to try to account for, uncomment when improving
```
Grunny: This test case fails because it creates a broken data row with `{#if:{{{length` as the source and label. We should account for this type of thing if possible, so leaving it in but commented out as an example for future improvements.

adamkarminski: Great approach!

adamkarminski:
Looks great to me! TDD FTW! :flags:

PR #7640, opened 2015-06-25
PLATFORM-1310: Move user history related tables to specials

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-1310

`user_login_history_summary` is used in joins with `events_local_users` - they both need to be on specials DB.

To keep all related tables in the same place let's move these tables from stats to specials:
- user_history (2.4 mm rows)
- user_login_history (500 mm rows)
- user_login_history_summary (6.7 mm rows)

### DIFF
--- a/extensions/wikia/UserChangesHistory/UserChangesHistory.class.php
+++ b/extensions/wikia/UserChangesHistory/UserChangesHistory.class.php
@@ -70,7 +70,7 @@ class UserChangesHistory {
 							WScribeClient::singleton('trigger')->send($data);
 						}
 						catch( TException $e ) {
-							Wikia\Logger\WikiaLogger::instance()->error( __METHOD__, [
+							Wikia\Logger\WikiaLogger::instance()->error( __METHOD__, ' - scribeClient exception', [
 								'exception' => $e
 							] );
 						}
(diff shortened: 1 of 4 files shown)


extensions/wikia/UserChangesHistory/UserChangesHistory.class.php
```diff
@@ -70,31 +70,35 @@ static public function LoginHistoryHook( $from, $user, $type = false ) {
 							WScribeClient::singleton('trigger')->send($data);
 						}
 						catch( TException $e ) {
-							Wikia::log( __METHOD__, 'scribeClient exception', $e->getMessage() );
+							Wikia\Logger\WikiaLogger::instance()->error( __METHOD__, [
```
wladekb: I'd suggest not removing the message "scribeClient exception"

wladekb:
it makes more sense to keep `user_login_history` and `user_login_history_summary` in the same db, can we move the `user_login_history` to `specials` as well?

macbre:
`user_login_history_summary` has 6 mm and `user_login_history` 500 mm rows. I'd like to move as less data as possible. I'll consult that with @drozdo

macbre:
@wladekb: Drozdo is moving all three tables (listed in PR description) from stats to specials DB. PR updated accordingly.

wladekb:
Cool! Thanks!

michalroszka:
:+1:

macbre:
Thanks for the CR, I'll merge it into dev and push it live when we're done with tables migration (still 2-3 days left).

PR #7642, opened 2015-06-25
CE-1934: Change user for moving main page while create new wiki process

lukaszkonieczny:


includes/wikia/tasks/Tasks/CreateNewWikiTask.class.php
```diff
@@ -146,6 +146,9 @@ private function moveMainPage() {
 					'target' => $targetTitle->getPrefixedText(),
 				];
 				if ( $sourceTitle->getPrefixedText() !== $targetTitle->getPrefixedText() ) {
+					$saveUser = $wgUser;
+					$wgUser = \User::newFromName( 'Wikia' );
```
Grunny: Can we put `'Wikia'` into a constant and use it here and in `protectKeyPages` where the same pattern is used?

Grunny:
Actually, can you use the GlobalStateWrapper Damon made for just this purpose? It's a slightly cleaner way to handle the mess of MW globals. See an example in HAWelcome which had the same issue:
* https://github.com/Wikia/app/blob/dev/extensions/wikia/HAWelcome/HAWelcomeTask.php#L325-L327
* https://github.com/Wikia/app/blob/dev/extensions/wikia/HAWelcome/HAWelcomeTask.php#L334-L336

Grunny:
LGTM.

lukaszkonieczny:
Tested.

PR #7643, opened 2015-06-25
CE-2118 Make flags modal respect colors

kamilkoterba:
:flags: @Wikia/community-engineering 
https://wikia-inc.atlassian.net/browse/CE-2118

Now theme designer colors are respected by modal. It was caching issue and the fix is to make varnish cache and mcache depend on sassParams.

At the same time refactoring code to use sass params sent from frontend not generating that within function. (replaced SassUtil::getOasisSettings with $sassParams)

### DIFF
--- a/extensions/wikia/Flags/scripts/FlagsModal.js
+++ b/extensions/wikia/Flags/scripts/FlagsModal.js
@@ -88,10 +88,7 @@ require(
 					type: loader.MULTI,
 					resources: {
 						mustache: '/extensions/wikia/Flags/controllers/templates/FlagsController_editForm.mustache,/extensions/wikia/Flags/controllers/templates/FlagsController_editFormEmpty.mustache,/extensions/wikia/Flags/controllers/templates/FlagsController_editFormException.mustache',
-						styles: '/extensions/wikia/Flags/styles/EditFormModal.scss',
-						params: {
-							sassParams: window.wgSassParams // Ensure per-theme colors Varnish cache and mcache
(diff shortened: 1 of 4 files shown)


lgarczewski:
Should another team be reviewing this as well as us? This touches AssetsManager so maybe send this to Platform?

Grunny:
@macbre you're probably the best to give this a quick look :).

extensions/wikia/Flags/scripts/FlagsModal.js
```diff
@@ -88,7 +88,10 @@ require(
 					type: loader.MULTI,
 					resources: {
 						mustache: '/extensions/wikia/Flags/controllers/templates/FlagsController_editForm.mustache,/extensions/wikia/Flags/controllers/templates/FlagsController_editFormEmpty.mustache,/extensions/wikia/Flags/controllers/templates/FlagsController_editFormException.mustache',
-						styles: '/extensions/wikia/Flags/styles/EditFormModal.scss'
+						styles: '/extensions/wikia/Flags/styles/EditFormModal.scss',
+						params: {
+							sassParams: window.wgSassParams // Ensure per-theme colors Varnish cache and mcache
+						}
```
macbre: I'd rather pass `sassParams` in `loader` module when SASS assets are requested instead of every place when the loader is used to fetch SASS styles.

lgarczewski:
LGTM and changes suggested by macbre have been made.

PR #7644, opened 2015-06-25
CE-1997 adding documentation and preview to conversion

lgarczewski:
https://wikia-inc.atlassian.net/browse/CE-1997

Known issue
also: quick comment -- the preview DOES NOT work on preview before the first save after all, I must've been viewin g a pre-created Draft ([CE-2158](https://wikia-inc.atlassian.net/browse/CE-2158))

added a "purge" link instead as a provisional solution

@Wikia/community-engineering

### DIFF
--- a/extensions/wikia/TemplateDraft/TemplateConverter.class.php
+++ b/extensions/wikia/TemplateDraft/TemplateConverter.class.php
@@ -4,6 +4,8 @@ class TemplateConverter {
 
 	const TEMPLATE_VARIABLE_REGEX = '/{{{([^\|}]*?)\|?.*}}}/sU';
 
+	var $title; // Title object of the template we're converting
+
 	/**
 	 * Names of variables that should be converted to a <title> tag
 	 * @var array
@@ -23,6 +25,10 @@ class TemplateConverter {
 		'photo',
 		'mainimage',
 	];
+
+	public function __construct( Title $templateTitle ) {
+		$this->title = $templateTitle;
+	}
 
 	/**
 	 * Performs a conversion to a template with a portable infobox.
@@ -46,6 +52,8 @@ class TemplateConverter {
 		}
 
 		$draft .= "</infobox>\n";
+
+		$draft .= $this->generatePreviewSection( $variables );
 
 		return $draft;
 	}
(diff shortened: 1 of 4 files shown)


extensions/wikia/TemplateDraft/TemplateConverter.class.php
```diff
@@ -102,4 +110,30 @@ public function createDataTag( $source, $label = '' ) {
...
+
+		$return = "<noinclude>\n== Usage & preview ==\n";
+
+		$return .= "Type in this:\n<pre>\n$docs</pre>\nto see this:\n$preview\n";
+
+		$return .= "[{{fullurl:PAGENAME}}?action=purge Click here to refresh the preview above]\n";
+
+		$return .= "</noinclude>\n";
+
+		return $return;
```
kamilkoterba: I think you may use a template for that to separate logic from view.

extensions/wikia/TemplateDraft/TemplateConverter.class.php
```diff
@@ -102,4 +110,30 @@ public function createDataTag( $source, $label = '' ) {
 		}
 		return "\t<data source=\"{$source}\"><label>{$label}</label></data>\n";
 	}
+
+	public function generatePreviewSection( $variables ) {
+		$return = "== Usage & preview ==\n";
```
kamilkoterba: i18n

extensions/wikia/TemplateDraft/TemplateConverter.class.php
```diff
@@ -102,4 +110,30 @@ public function createDataTag( $source, $label = '' ) {
...
+		$return .= "<noinclude>\n";
```
Grunny: The `<noinclude>` go before the section heading so the section heading is also removed from transclusions.

Grunny: Actually, you're starting again below where this was done, so these can be removed.

extensions/wikia/TemplateDraft/TemplateConverter.class.php
```diff
@@ -47,6 +53,8 @@ public function convertAsInfobox( $content ) {
 
 		$draft .= "</infobox>\n";
 
+		$draft .= $this->generatePreviewSection( $variables );
```
Grunny: This will break the unit tests for this method. Can you rebase against `CE-templatedraft` so you get the unit tests in this branch?

Alternatively (or additionally), it might be better to separate the method responsibilities for conversion of the infobox text from the preview generation and just call this after `convertAsInfobox` in the controller (will need to account for the variables, but that's a quick change). Would make getting just the infobox conversion reusable rather than always getting the preview text too.

lgarczewski: Agreed, good idea.

extensions/wikia/TemplateDraft/TemplateConverter.class.php
lgarczewski: Oops. Leftover from previous version. ;)

extensions/wikia/TemplateDraft/TemplateConverter.class.php
lgarczewski: I would say in this case this is not needed because the converter's logic operates purely on the infobox code and is a "gluing" class that puts that code together in a different form. There is no logic indepentent from view in this case (at least not as it's currently built).

kamilkoterba:
lgtm

PR #7647, opened 2015-06-25
(CE-2160) Fix interface language always displaying content language

Grunny:
Fix interface language always displaying content language instead of the
user's language. The fix in #7591 removed the unsetting of the recursion
value after a language was returned from the main method (in the elseif block)
leading it to get stuck returning the content language on successive calls
due to always entering the recursion block which should only be called from
within the elseif block of the method.

/cc @macbre @michalroszka @Wikia/community-engineering


macbre:
lgtm :shipit:

PR #7648, opened 2015-06-25
Decouple user preferences, attributes, and flags from getOption()

drsnyder:
This work decouples preferences, attributes, and flags from `User::getOption()`.  See the [ticket](https://wikia-inc.atlassian.net/browse/SERVICES-469) for the details and rationale. 

TODO:
- [x] create the setters for each
- [x] replace the `getOption` call sites. We are using [this spreadsheet](https://docs.google.com/spreadsheets/d/11B8FupwtaahpE2n9GvYeaSMM43HDhmWM-VoCh1B572A/edit#gid=0) to coordinate and track that effort
- [x] replace the `setOption` call sites. Tracking TBD
- [x] add logging to `get`/`set` option so we can log when we miss any

/cc @Wikia/services-team


nmonterroso:
As part of this, should we also do something like
```php
public function setOption(...) {
   log("WHY ARE YOU CALLING SETOPTION? YOU CALLED IT FROM <location in code>");
  $this->setOptionForRealsies(...);
}

private function setOptionForRealsies(...) {
  // logic that used to be in setOption
}
```
That way we can track down set/get calls that are dynamically invoked using `call_user_func` or we missed in our grep. OTW LGTM.

extensions/wikia/EditAccount/SpecialEditAccount_body.php
```diff
@@ -219,10 +219,10 @@ public function execute( $par ) {
 					'userRealName' => $this->mUser->getRealName(),
 					'userId'  => $this->mUser->getID(),
 					'userReg' => date( 'r', strtotime( $this->mUser->getRegistration() ) ),
-					'isUnsub' => $this->mUser->getOption('unsubscribed'),
-					'isDisabled' => $this->mUser->getOption('disabled'),
+					'isUnsub' => $this->mUser->getGlobalFlag('unsubscribed'),
```
drsnyder: @ArturKlajnerok fixed in 8a3e5fa4.

includes/User.php
```diff
@@ -2416,6 +2416,177 @@ public function getIntOption( $oname, $defaultOverride=0 ) {
...
+	 * preferences, attributes or flags should be of the form "{option}-{cityId}"
+	 *
+	 * IF YOU USE $sep, MAKE A PLAN TO NORMALIZE IT TO "-"!
+	 *
+	 * @param string $option
+	 * @param int $cityId [optional]
+	 * @param char $sep the separator between the option and the id.
+	 * @return string
+	 */
+	public static function createLocalOptionName($option, $cityId = null, $sep = '-') {
```
ArturKlajnerok: Shouldn't this be a `private` function ?

drsnyder: But then you can't test it (easily).

sqreek:
[getOption](https://github.com/Wikia/app/pull/7648/files#diff-897fba1dd6b8edc3238762e35d0b7c71R2320) and [setOption](https://github.com/Wikia/app/pull/7648/files#diff-897fba1dd6b8edc3238762e35d0b7c71R2596) are deprecated now

includes/User.php
```diff
@@ -2416,11 +2417,183 @@ public function getIntOption( $oname, $defaultOverride=0 ) {
...
+	 * @param string $preference the preference name
+	 * @param int $cityId the city id
+	 * @param string $sep the separator between the name and the city id
+	 * @return string
+	 */
+	public function getLocalPreference($pref, $cityId = null, $sep = "-") {
+		global $wgCityId;
+		if (!isset($cityId)) {
+			$cityId = $wgCityId;
+		}
```
garthwebb: @drsnyder I think I'm missing something ... shouldn't there be more code here?  Something to actually get the preference value?

drsnyder: @garthwebb it's magic! Actually that work is happening over [here](https://github.com/Wikia/app/pull/7723). That branch was taken off of this one and will be merged in here before we merge to dev. So yes, when we are done something will actually get the preferences :).

ArturKlajnerok:
Lgtm :+1:

PR #7652, opened 2015-06-25
(CE-2166) Fix missing parameters on Special:Flags

Grunny:
In commit be8778f309e486e3 a check for whether or not the JSON decoded
flag names was an array was added; however, by default json_decode will
return an StdClass object for a decode JSON object, not an array. This
makes json_decode return an associative array instead.


adamkarminski:
LGTM, thanks Grunny! :+1:

PR #7656, opened 2015-06-26
CE-2169 Removing merging ParserOutput which is messing with the extra…

adamkarminski:
…cting task

@Wikia/community-engineering


lukaszkonieczny:
lgtm

PR #7659, opened 2015-06-26
CE-2044 adding approveDraft method

kamilkoterba:
https://wikia-inc.atlassian.net/browse/CE-2098
method for approving draft

@Wikia/community-engineering

### DIFF
--- a/extensions/wikia/TemplateDraft/TemplateConverter.class.php
+++ b/extensions/wikia/TemplateDraft/TemplateConverter.class.php
@@ -111,6 +111,23 @@ class TemplateConverter {
 	}
 
 	/**
+	 * Overrides content of parent page with contents of draft page
+	 * @param Title $title
+	 */
+	public function approveDraft( Title $title ) {
+		// Get contents of draft page
+		$article = Article::newFromId( $title->getArticleID() );
+		$draftContent = $article->getContent();
+		// Get parent page
+		$parentTitleText = $title->getBaseText();
+		$parentTitle = Title::newFromText( $parentTitleText, $title->getNamespace() );
+		$article = Article::newFromId( $parentTitle->getArticleID() );
+		$page = $article->getPage();
+		// Save to parent page
+		$page->doEdit( $draftContent, wfMessage( 'templatedraft-approval-summary' )->plain() );
+	}
+
+	/**
 	 * Prepares variables from templates
 	 *
 	 * @param $templateVariables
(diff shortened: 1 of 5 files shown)


lgarczewski:
IMO the method should be in TemplateDraftController, not in the converter. Otherwise looks good.

extensions/wikia/TemplateDraft/TemplateConverter.class.php
```diff
@@ -111,6 +111,27 @@ public function createDataTag( $variable ) {
...
+		if ( $title->userCan( 'edit' ) ) {
+			// Get contents of draft page
+			$article = Article::newFromId( $title->getArticleID() );
+			$draftContent = $article->getContent();
+			// Get parent page
+			$parentTitleText = $title->getBaseText();
+			$parentTitle = Title::newFromText( $parentTitleText, $title->getNamespace() );
+			$page = WikiPage::newFromID( $parentTitle->getArticleID() );
+			// Save to parent page
+			$page->doEdit( $draftContent, wfMessage( 'templatedraft-approval-summary' )->plain() );
```
lgarczewski: This message should use content language.

extensions/wikia/TemplateDraft/TemplateConverter.class.php
```diff
@@ -111,6 +111,27 @@ public function createDataTag( $variable ) {
 	}
 
 	/**
+	 * Overrides content of parent page with contents of draft page
+	 * @param Title $title Title object of sub page (draft)
+	 * @throws PermissionsException
+	 */
+	public function approveDraft( Title $title ) {
+		if ( $title->userCan( 'edit' ) ) {
```
Grunny: Should we instead check the permissions against the parent title? A user may be able to edit the draft, but the parent template may be protected, and this would enable them to edit the parent template when they don't have the appropriate permissions.

extensions/wikia/TemplateDraft/TemplateConverter.class.php
```diff
@@ -111,6 +111,27 @@ public function createDataTag( $variable ) {
...
+			// Get contents of draft page
+			$article = Article::newFromId( $title->getArticleID() );
+			$draftContent = $article->getContent();
+			// Get parent page
+			$parentTitleText = $title->getBaseText();
+			$parentTitle = Title::newFromText( $parentTitleText, $title->getNamespace() );
+			$page = WikiPage::newFromID( $parentTitle->getArticleID() );
```
Grunny: Can we just use the [`TemplateDraftHelper::getParentTitleId`](https://github.com/Wikia/app/blob/4c9040be81/extensions/wikia/TemplateDraft/TemplateDraftHelper.class.php#L28) here?

kamilkoterba: Now I need instance of parent title to check edit permissions anyway, so I don't think getParentTitleId can simplify something.

Grunny:
LGTM.

PR #7660, opened 2015-06-26
CE-2103 mark templates as not infoboxes (and other classification)

lgarczewski:
Basic assumptions:
- we want to classify any template
- a template can be positively classified as being of only one type
- ...but a template can also be negatively classified (i.e. "this is not an infobox)

Notes on future directions:
- should be part of Article service or a similar page-related metadata storage/retrieval mechanism

### DIFF
--- a/extensions/wikia/TemplateDraft/controllers/TemplateClassificationController.class.php
+++ b/extensions/wikia/TemplateDraft/controllers/TemplateClassificationController.class.php
@@ -27,14 +27,14 @@ class TemplateClassificationController extends WikiaController {
 	private function getClassificationProp( $type ) {
 		if ( array_search( $type, $this->templateTypes, true ) ) {
 			return self::TEMPLATE_CLASSIFICATION_DATA_PREFIX . $type;
-		} else {
-			// invalid type!
-			return false;
 		}
+		
+		// invalid type!
+		return false;
 	}
 
-	public function classifyTemplate( String $templateName, $type, bool $value, $actor = self::CLASSIFICATION_ACTOR_HUMAN ) {
-		$title = Title::newFromName( $titleName, NS_TITLE );
+	public function classifyTemplate ($templateName, $type, bool $value, $actor = self::CLASSIFICATION_ACTOR_HUMAN ) {
+		$title = Title::newFromName( $titleName, NS_TEMPLATE );
 
 		if ( is_null( $title ) ) {
 			return false;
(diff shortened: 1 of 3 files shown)


extensions/wikia/TemplateDraft/controllers/TemplateClassificationController.class.php
```diff
@@ -0,0 +1,65 @@
...
+		if ( array_search( $type, $this->templateTypes, true ) ) {
+			return self::TEMPLATE_CLASSIFICATION_DATA_PREFIX . $type;
+		} else {
+			// invalid type!
+			return false;
+		}
+	}
+
+	public function classifyTemplate( String $templateName, $type, bool $value, $actor = self::CLASSIFICATION_ACTOR_HUMAN ) {
+		$title = Title::newFromName( $titleName, NS_TITLE );
```
Grunny: Should be `NS_TEMPLATE`?

extensions/wikia/TemplateDraft/controllers/TemplateClassificationController.class.php
```diff
@@ -0,0 +1,65 @@
...
+	private function getClassificationProp( $type ) {
+		if ( array_search( $type, $this->templateTypes, true ) ) {
+			return self::TEMPLATE_CLASSIFICATION_DATA_PREFIX . $type;
+		} else {
+			// invalid type!
+			return false;
+		}
+	}
+
+	public function classifyTemplate( String $templateName, $type, bool $value, $actor = self::CLASSIFICATION_ACTOR_HUMAN ) {
```
Grunny: This will be a fatal error due to typing.

extensions/wikia/TemplateDraft/controllers/TemplateClassificationController.class.php
```diff
@@ -0,0 +1,65 @@
...
+	/**
+	 * Flags indicating who's taking a classification action
+	 */
+	const CLASSIFICATION_ACTOR_AI = 0;
+	const CLASSIFICATION_ACTOR_HUMAN = 1;
+
+	private function getClassificationProp( $type ) {
+		if ( array_search( $type, $this->templateTypes, true ) ) {
+			return self::TEMPLATE_CLASSIFICATION_DATA_PREFIX . $type;
+		} else {
```
Grunny: Would be cleaner without the else since you return in the if block anyway.

PR #7661, opened 2015-06-26
Fix unit tests as the constructor now requires a Title

Grunny:
#7644 introduced a required Title object to the constructor. Updating the

unit tests so this works again.


lukaszkonieczny:
lgtm

PR #7664, opened 2015-06-26
CE-2159 Fetch flags ParserOutput to LinksUpdate

adamkarminski:
This fix deals with flags ParserOutput not being available in LinksUpdate and Extraction tasks firing in a wrong context.

@Wikia/community-engineering


lgarczewski:
LGTM.

PR #7666, opened 2015-06-26
CONCF-672 | CuratedContent/GameGuides setting/purging cache

Warkot:
It should be compatible with https://github.com/Wikia/mercury/pull/1001, but it seems that mercury is not so keen on caching these days. Nevertheless, stuff has to be purged, if it's cached it's changing afterwards.

Please CR. @Wikia/x-wing @owend @macbre 

:cat2:

### DIFF
--- a/extensions/wikia/CuratedContent/CuratedContentController.class.php
+++ b/extensions/wikia/CuratedContent/CuratedContentController.class.php
@@ -395,11 +395,10 @@ class CuratedContentController extends WikiaController {
 		global $wgServer;
 
 		$content = F::app()->wg->WikiaCuratedContent;
-		$mercuryApiExists = class_exists( 'MercuryApiHooks' );
 
 		( new SquidUpdate( array_unique( array_reduce(
 			$content,
-			function ( $urls, $item ) use ( $wgServer, $mercuryApiExists ) {
+			function ( $urls, $item ) use ( $wgServer ) {
 				if ( $item[ 'title' ] !== '' && empty( $item[ 'featured' ] ) ) {
 					// Purge section URLs using urlencode() (standard for MediaWiki), which uses implements RFC 1738
 					// https://tools.ietf.org/html/rfc1738#section-2.2 - spaces encoded as `+`.
@@ -414,13 +413,11 @@ class CuratedContentController extends WikiaController {
 					// Mercury web app uses this variant - request from Hapi.js to MediaWiki.
 					$urls[ ] = self::getUrl( 'getList' ) . '&section=' . self::encodeURIQueryParam( $item[ 'title' ] );
 					// Mercury web app uses this variant - request from Ember.js to Hapi.js.
-					if ( $mercuryApiExists ) {
-						$urls[ ] =
-							$wgServer .
-							MercuryApiHooks::SERVICE_API_BASE .
-							MercuryApiHooks::SERVICE_API_CURATED_CONTENT .
-							self::encodeURIQueryParam( $item[ 'title' ] );
-					}
+					$urls[ ] =
+						$wgServer .
+						MercuryApiHooks::SERVICE_API_BASE .
+						MercuryApiHooks::SERVICE_API_CURATED_CONTENT .
+						self::encodeURIQueryParam( $item[ 'title' ] );
 				}
 				return $urls;
 			},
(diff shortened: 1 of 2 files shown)


extensions/wikia/CuratedContent/CuratedContentController.class.php
```diff
@@ -66,7 +66,7 @@ public function getPage() {
 		//This will always return json
 		$this->response->setFormat( 'json' );
 
-		$this->cacheResponseFor( 7, self::DAYS );
+		$this->cacheResponseFor( 24, self::HOURS );
```
macbre: Why can't we call `setCacheValidity` here with one of the constant values taken from `WikiaResponse`?

extensions/wikia/CuratedContent/CuratedContentController.class.php
```diff
@@ -400,12 +392,61 @@ function getJsonItem( $titleName, $ns, $pageId ) {
...
+						$wgServer .
+						MercuryApiHooks::SERVICE_API_BASE .
+						MercuryApiHooks::SERVICE_API_CURATED_CONTENT .
+						self::encodeURIQueryParam( $item[ 'title' ] );
+				}
+				return $urls;
+			},
+			// Purge all sections list getter URL - no additional params
+			[ self::getUrl( 'getList' ) ]
+		) ) ) )->doUpdate();
```
macbre: As we're purging four different URLs here maybe it's worth to consider to use surrogate keys here instead.

extensions/wikia/CuratedContent/CuratedContentController.class.php
Warkot: I just left it how it was before, but indeed, it's a good idea. :+1:

extensions/wikia/CuratedContent/CuratedContentController.class.php
```diff
@@ -400,12 +392,62 @@ function getJsonItem( $titleName, $ns, $pageId ) {
...
 	static function onCuratedContentSave() {
-		self::purgeMethod( 'getList' );
+		global $wgServer;
+
+		$content = F::app()->wg->WikiaCuratedContent;
+
+		( new SquidUpdate( array_unique( array_reduce(
+			$content,
+			function ( $urls, $item ) use ( $wgServer ) {
+				if ( $item[ 'title' ] !== '' && empty( $item[ 'featured' ] ) ) {
```
bkoval: Why not to use empty() for the first check?

extensions/wikia/CuratedContent/CuratedContentController.class.php
```diff
@@ -400,12 +392,62 @@ function getJsonItem( $titleName, $ns, $pageId ) {
...
+					// Purge section URLs using urlencode() (standard for MediaWiki), which uses implements RFC 1738
+					// https://tools.ietf.org/html/rfc1738#section-2.2 - spaces encoded as `+`.
+					// iOS apps use this variant.
+					$urls[ ] = self::getUrl( 'getList' ) . '&section=' . urlencode( $item[ 'title' ] );
```
bkoval: Is the space inside empty square brackets forced by the guidelines? :)

extensions/wikia/CuratedContent/CuratedContentController.class.php
Warkot: `empty` would return `true` when `$item[ 'title' ]` is `false`, `null`, empty array etc.

extensions/wikia/CuratedContent/CuratedContentController.class.php
Warkot: Hard to tell, it's not a definition, but it's also not strictly indexing.  https://github.com/Wikia/guidelines/blob/master/PHP/CodingConventions.md#array-syntax
I would rather go without spaces inside, as this construct is closer to indexing. I wrote it this way to be consistent with the rest of the file.

Warkot:
As a reminder — not only API calls have to be purged, but also URLs to the new Curated Main Pages routes like `/main/section/`, `/main/category/`.

Warkot:
It's continued in #7793.

extensions/wikia/CuratedContent/CuratedContentController.class.php
Warkot: Is it possible to purge internal cache using surrogate keys — for responses like from Mercury's Hapi.js to Mercury API in MediaWiki?
We could move our conversation to new PR: #7793.

PR #7668, opened 2015-06-27
CE-2100 convert button

lgarczewski:
This was a fun one. Now any insights page can have an "alternate action" (other than linking to the page).

@Wikia/community-engineering

### DIFF
--- a/extensions/wikia/Insights/models/InsightsQuerypageModel.php
+++ b/extensions/wikia/Insights/models/InsightsQuerypageModel.php
@@ -82,6 +82,14 @@ abstract class InsightsQuerypageModel extends InsightsModel {
 		return true;
 	}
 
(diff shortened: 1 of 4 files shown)


extensions/wikia/Insights/models/InsightsQuerypageModel.php
```diff
@@ -82,6 +82,14 @@ public function arePageViewsRequired() {
 		return true;
 	}
 
+	public function hasAltAction() {
+		return false;
+	}
+
+	public function getAltActionUrl( $title ) {}
+
+	public function altActionLinkMessage() {}
```
lukaszkonieczny: What do you think about returning some values?

lukaszkonieczny:
lgtm

PR #7669, opened 2015-06-27
[SERVICES-422] Helios register should not be called via hook

ArturKlajnerok:
We shouldn't be calling helios registration via hook anymore.

/cc @Wikia/services-team


ljagiello:
:+1:

PR #7670, opened 2015-06-27
[SERVICES-422] Removed unused hook

ArturKlajnerok:
Removing unused hook that was used for helios registration.

/cc @Wikia/services-team


ljagiello:
LGTM

PR #7673, opened 2015-06-29
CONCF-941 | As a developer I would like getCuratedContentSection resource to throw exceptions when there is error

tomasznapieralski:
https://wikia-inc.atlassian.net/browse/CONCF-941


extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -407,9 +407,14 @@ public function getCuratedContentSection() {
...
 		if ( empty( $section ) ) {
 			$this->response->setVal( 'items', false );
-		} else {
-			$data = $this->getCuratedContentData( $section );
-			$this->response->setVal( 'items', $data[ 'items' ] );
+			throw new NotFoundApiException( 'Section is not set' );
+		}
+
+		$data = $this->getCuratedContentData( $section );
+		$this->response->setVal( 'items', $data[ 'items' ] );
```
rogatty: You should do this after checking `if ( empty( $data ) )`

Warkot:
LGTM

bognix:
LGTM

hakubo:
LGTM

PR #7677, opened 2015-06-29
CE-2103 Setting up a JS module and hiding the right rail module

adamkarminski:

This PR sets up a JS module for the TemplateDraft extension. The first functionality is hiding the right rail module when a user clicks the "This is not an infobox" link. Clicking the link also triggers marking the template as a non-infobox one.

In this PR you can also find a few adjustments to hooked methods firing conditions and minor fixes to a  documentation message for Drafts.

### DIFF
--- a/extensions/wikia/TemplateDraft/TemplateDraftHooks.class.php
+++ b/extensions/wikia/TemplateDraft/TemplateDraftHooks.class.php
@@ -2,10 +2,8 @@
 
 class TemplateDraftHooks {
 
-	public static function onSkinAfterBottomScripts( $skin, &$text ) {
-		global $wgTitle, $wgUser;
-
-		if ( $wgUser->isPowerUser() && $wgTitle->getNamespace() === NS_TEMPLATE ) {
+	public static function onSkinAfterBottomScripts( Skin $skin, &$text ) {
+		if ( $skin->getUser()->isPowerUser() && $skin->getTitle()->getNamespace() === NS_TEMPLATE ) {
 			$scripts = AssetsManager::getInstance()->getURL( 'template_draft' );
 
 			foreach ( $scripts as $script ) {
--- a/extensions/wikia/TemplateDraft/TemplateDraftHooks.class.php
+++ b/extensions/wikia/TemplateDraft/TemplateDraftHooks.class.php
@@ -8,7 +8,7 @@ class TemplateDraftHooks {
 		if ( $wgUser->isPowerUser() && $wgTitle->getNamespace() === NS_TEMPLATE ) {
 			$scripts = AssetsManager::getInstance()->getURL( 'template_draft' );
 
-			foreach( $scripts as $script ) {
+			foreach ( $scripts as $script ) {
 				$text .= Html::linkedScript( $script );
 			}
 		}
@@ -63,7 +63,7 @@ class TemplateDraftHooks {
 				 */
 				$controller = new TemplateDraftController();
 				$text = $controller->createDraftContent(
-					$title, // @TODO this is currently taking the *edited* title (with subpage), not the *converted* title  
+					$title, // @TODO this is currently taking the *edited* title (with subpage), not the *converted* title
 					$parentContent,
 					[ $controller::TEMPLATE_INFOBOX ]
 				);
@@ -77,7 +77,7 @@ class TemplateDraftHooks {
 	 * It adds an editintro message with help and links.
 	 *
 	 * @param String $msgName
-	 * @param Array $msgParams 
+	 * @param Array $msgParams
 	 * @param Title $title
 	 * @return bool
 	 */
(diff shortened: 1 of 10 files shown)


extensions/wikia/TemplateDraft/controllers/TemplateDraftController.class.php
```diff
@@ -35,4 +35,23 @@ public function createDraftContent( Title $title, $content, Array $flags ) {
 
 		return $newContent;
 	}
+
+	public function markTemplateAsNotInfobox() {
```
Grunny: This is a state changing action, so should require a post and a token as a best practice.

adamkarminski: A very valid point, thanks!

Grunny: And it should also meet the same permission requirements as the display of the module, so should also check for power user.

adamkarminski: Even more valid point, even more thanks :)

extensions/wikia/TemplateDraft/controllers/TemplateDraftController.class.php
```diff
@@ -35,4 +35,23 @@ public function createDraftContent( Title $title, $content, Array $flags ) {
...
+		$requestParams = $this->getRequest()->getParams();
+
+		if ( empty( $requestParams['pageId'] )
+			|| !is_numeric( $requestParams['pageId'] )
```
Grunny: Instead of using `getParams` when you only have one parameter right now, it might be better to just use `$pageId = this->getRequest()->getInt( 'pageId' );` and checking it isn't `0`.

extensions/wikia/TemplateDraft/scripts/rightRailModule.js
```diff
@@ -0,0 +1,36 @@
...
+		}
+
+		function closeModule() {
+			$('.templatedraft-module').hide();
+
+			$.nirvana.sendRequest({
+				controller: 'TemplateDraftController',
+				method: 'markTemplateAsNotInfobox',
+				data: {
+					'pageId': w.wgArticleId
```
Grunny: In MW 1.19, it's generally preferred to require `mw` and use `mw.config.get('wgArticleId');`.

extensions/wikia/TemplateDraft/TemplateDraftHooks.class.php
```diff
@@ -2,6 +2,20 @@
 
 class TemplateDraftHooks {
 
+	public static function onSkinAfterBottomScripts( $skin, &$text ) {
+		global $wgTitle, $wgUser;
```
Grunny: `$skin` should be a context source, so you should be able to avoid the globals with `$this->getTitle()` and `$this->getUser()`.

Grunny:
LGTM. :)

extensions/wikia/TemplateDraft/controllers/TemplateDraftController.class.php
```diff
@@ -59,4 +59,43 @@ public function approveDraft( Title $title ) {
 		// Save to parent page
 		$page->doEdit( $draftContent, wfMessage( 'templatedraft-approval-summary' )->inContentLanguage()->plain() );
 	}
+
+	public function markTemplateAsNotInfobox() {
+		/**
+		 * First, validate a request
+		 */
+		if ( !$this->wg->User->isPowerUser()
```
lgarczewski: We should tie this to a user right validated via `User::userCan()` (quoting method from memory, please verify) and add it to the following groups:
* powerusers
* staff
* helpers
* vstf
* voldev

extensions/wikia/TemplateDraft/controllers/TemplateDraftController.class.php
```diff
@@ -59,4 +59,43 @@ public function approveDraft( Title $title ) {
...
+		}
+
+		/**
+		 * Wikia::setProps unfortunately fails silently so if we get to this point
+		 * we can set the response's status to true anyway...
+		 */
+		$this->response->setVal( 'status', true );
+
+		Wikia::setProps( $pageId, [
+			self::TEMPLATE_INFOBOX_PROP => 0,
```
lgarczewski: I think we're using two different sets of props, @adamkarminski -- I'd suggest following the one set out in #7660 -- can we get that one reviewed and merged and use TemplateClassificationController here?

adamkarminski: Since #7660 is not ready to be merged yet I'd suggest leaving it as is and ensure consistency when working on #7660 .

lgarczewski: Sure.

PR #7682, opened 2015-06-29
CE-2017: Adding flags to parser cache

lukaszkonieczny:
@Wikia/community-engineering 

PLEASE DONT MERGE!

Hi guys,
I tried to find the root cause why categories and whatlinkshere weren't updated and probably I found it :) 

I think it's here: https://github.com/Wikia/app/blob/dev/extensions/wikia/Flags/controllers/FlagsController.class.php#L88. There was a lag between master and slaves, so sometimes we didn't get just added flags. Also, not always ParserCache was purged.

In my approach I'm using flags from form instead fetching them from db. And this code still needs some work - there are some hacks to check if it works. I'm sure there are things we can do better and more kosher.


lgarczewski:
LGTM.

PR #7683, opened 2015-06-29
Monetization (MON-415): SpecialSitemap - set robots meta tag to noindex for xml sitemap

saipetch:


extensions/wikia/Sitemap/SpecialSitemap_body.php
```diff
@@ -95,6 +95,7 @@ public function execute( $subpage ) {
 			// cache on both CDN and client
 			header( "Cache-Control: s-maxage=86400", true );
 			header( "X-Pass-Cache-Control: public, max-age=86400", true );
+			header( "X-Robots-Tag: noindex" );
```
jimmordino: Do you also need this after line 180?

jimmordino:
:+1:

extensions/wikia/Sitemap/SpecialSitemap_body.php
owend: Nope, that shouldn't be necessary.

PR #7693, opened 2015-06-30
CE-2017: Fix for editing pages with flags

lukaszkonieczny:


Grunny:
LGTM, after an explanatory comment is added. :)

PR #7695, opened 2015-06-30
SOC-994 Add additional space to format docs the way ReflectionClass::…

jsutterfield:
…getDocComment expects

The fix for this is literally adding a single space. The block of code which extracts the template name from the doc comment looks like this:

``` php
$reflection = new ReflectionClass( $controllerClass );
            $method = $reflection->getMethod( $methodName );

            $comment = $method->getDocComment();
            if ( preg_match( '/@template ([^ ]+)/', $comment, $matches ) ) {
                $template = trim( $matches[1] );
```

The problem was that `$method->getDocComment();` expects the comment to be formatted in a certain way, specifically the first line (/**) should be preceded by one tab, and all subsequent lines should be preceded by one tab, followed by one space.

 Since this wasn't formatted properly, `$method->getDocComment();` was returning a string which looked like this `"/**\n\t* @template avatarLayout\n\t* /"`,  rather than this  `"/**\n\t* @template avatarLayout\n\t * /"`. Note in the previous version, there asterisk is immedietly following the `\t` character. This was resulting in `matches[1]` equaling `avatarLayout\n\t*` which wasn't being trimmed b/c of the trailing \* and wreaking all kinds of havoc.

Ticket: https://wikia-inc.atlassian.net/browse/SOC-994


garthwebb:
Ugh, that's terrible.  A poorly formatted doc string shouldn't be able to crash us again though.  Can we fix the regex here?  Lets include ```\n\r\t``` in addition to the single space of that negative character class.

garthwebb:
Actually I just found that PHP supports the Perlish ```\s``` and ```\S``` character classes matching any whitespace character or no whitespace characters, so lets use that instead of the character class, e.g. ```\S+```

jsutterfield:
@garthwebb, what do you think about just matching the word instead? `\w+`?

mklucsarits:
Yeah, ```[^ ]``` seems like overkill when you just need a template that will definitely be matched with ```\w+```.

garthwebb:
Using ```\w``` won't work if the template name has a hyphen or dot in it.  I realize our naming conventions currently don't include these characters and that this platform has limited life.  However, I think we should still try to match as many legal characters as reasonable rather than what meets our current style guide.

garthwebb:
... and by 'reasonable' I mean by unix naming standards.  Folks use periods and dashes but generally avoid spaces or special characters even though they're technically legal when escaped properly.

jsutterfield:
Ah, okay, that makes sense.

mklucsarits:
LGTM

PR #7699, opened 2015-07-01
UncycloUserMigrator: clear the user cache

macbre:
Improve user cache invalidation. So far the cache was only invalidated when the global user was created.

This change will invalidate the user cache when only ID remapping takes place.

@michalroszka / @wladekb


michalroszka:
:+1:

PR #7706, opened 2015-07-01
CE-2175 Edit draft entry point

MariuszCz:


extensions/wikia/TemplateDraft/TemplateDraftHooks.class.php
```diff
@@ -76,24 +81,52 @@ public static function onEditFormPreloadText( &$text, Title $title ) {
...
+						),
+						Xml::element( 'a', [
+							'href' => $base->getFullUrl( [ 'action' => 'edit' ] ),
+							'target' => '_blank'
+						],
+							wfMessage( 'templatedraft-module-view-parent' ) )
+					)->escaped(),
+				];
+			} elseif ( !$helper->isTitleDraft( $title ) ) {
+				$base = Title::newFromText( $title->getBaseText() . '/Draft', NS_TEMPLATE );
```
lgarczewski: Please use the message for the Draft subage name here, instead of hardcoding it. See Helper::isTitleDraft for the message name.

maintenance/app
```diff
@@ -0,0 +1 @@
+/usr/wikia/source/app/
```
Grunny: Unintentional? :)

extensions/wikia/TemplateDraft/TemplateDraftHooks.class.php
```diff
@@ -76,24 +77,51 @@ public static function onEditFormPreloadText( &$text, Title $title ) {
...
+					'action' => 'edit',
+					TemplateConverter::CONVERSION_MARKER => 1,
+				] );
+				$preloads['EditPageIntro'] = [
+					'content' => wfMessage( 'templatedraft-module-editintro-please-convert' )->rawParams(
+						Xml::element( 'a', [
+							'href' => $draftUrl,
+							'target' => '_blank'
+						],
+							wfMessage( 'templatedraft-module-button' ) )
```
Grunny: Should use an explicit output format. Right now, without one, it will default to `->parse()` which is probably unnecessary given the message text, and the contents of `Xml::element` are escaped anyway, so `->plain()` or `->text()` would be fine here. :)

maintenance/app
MariuszCz: Yes, that was something wrong with my devbox :)

extensions/wikia/TemplateDraft/TemplateDraftHooks.class.php
```diff
@@ -76,24 +77,51 @@ public static function onEditFormPreloadText( &$text, Title $title ) {
...
+					)->plain(),
```
Grunny: This should still be `->escaped()`, I meant the `wfMessage( 'templatedraft-module-button' )` above.

MariuszCz: Ok, what about wfMessage( 'templatedraft-module-help' ) and wfMessage( 'templatedraft-module-view-parent' ) ? Also should be plain()?

extensions/wikia/TemplateDraft/TemplateDraftHooks.class.php
```diff
@@ -76,24 +77,51 @@ public static function onEditFormPreloadText( &$text, Title $title ) {
...
+							'target' => '_blank',
+						],
+							wfMessage( 'templatedraft-module-help' )
+						),
+						Xml::element( 'a', [
+							'href' => $base->getFullUrl( [ 'action' => 'edit' ] ),
+							'target' => '_blank'
+						],
+							wfMessage( 'templatedraft-module-view-parent' ) )
+					)->plain(),
```
Grunny: Should be `->escaped()`.

extensions/wikia/TemplateDraft/TemplateDraftHooks.class.php
Grunny: Yep, but `templatedraft-editintro` which wraps them should be `->escaped()` same as this one (which I just added a comment for).

lgarczewski:
LGTM, thanks Mariusz.

PR #7707, opened 2015-07-01
PLATFORM-1308 PLATFORM-1317 invalid curl options, file uploads failures

ghost:
https://wikia-inc.atlassian.net/browse/PLATFORM-1308
https://wikia-inc.atlassian.net/browse/PLATFORM-1317

@macbre 

I need to collect some data on the invalid curl options and the context. Some of the failing results imply file upload failures.


macbre:
LGTM.

Worth pushing outside the release to monitor the log rate?

PR #7710, opened 2015-07-01
DAT-2960 ensure data is an array

idradm:


dianafa:
If we are sure that the empty $infobox[ 'data' ] is caused by not purged articles after the changes with the infobox array, then it's ok to prevent them by this simply check. If it is caused by other reason, it would be nice to find it.

LGTM

SebastianMarzjan:
:+1:

PR #7714, opened 2015-07-01
DAT-2940 rename footer tag to navigation

idradm:


dianafa:
LGTM

extensions/wikia/PortableInfobox/templates/PortableInfoboxItemNavigation.mustache
```diff
@@ -0,0 +1,2 @@
+<nav class="portable-infobox-navigation portable-infobox-item-margins portable-infobox-header-background
```
SebastianMarzjan: While I know it's just a move I wonder if portable-infobox-header-background and  portable-infobox-header-font are the best classes for this - IMO we should rename them so there's no confusion why they are here (i.e. to portable-infobox-secondary-background or something similar)

SebastianMarzjan:
LGTM :+1:

PR #7716, opened 2015-07-01
CONCF-714 | Add validation for images for Special:CuratedContent

vforge:
/cc @Warkot @bognix @rogatty


bognix:
LGTM

extensions/wikia/CuratedContent/js/CuratedContentManagmentTool.js
```diff
@@ -115,51 +115,46 @@ $(function () {
...
+				var $lastSection = $ul.find('.section').last().find('.section-input');
+				if ($lastSection.val().length === 0) {
+					$lastSection.parent().find('.image').removeError();
+				}
+			},
 			checkForm = function () {
 				$save.removeClass();
 
-				checkInputs($ul.find('.section-input'), ['required', 'limit']);
+				checkInputs($ul.find('.section-input'), ['limit', 'checkEmpty']);
```
rogatty: Why?

vforge: It's fix for bug that we've found - it should possible to add optional section (without a name), but only once - those flags were wrong since last code refactor.

rogatty:
LGTM

bognix:
LGTM

extensions/wikia/CuratedContent/js/CuratedContentManagmentTool.js
```diff
@@ -2,416 +2,414 @@
...
+				maxAllowedLength = 48, // it's derived from MAX_LABEL_LENGTH in CuratedContentSpecialController
+
+				setup = function (elem) {
+					(elem || $ul.find('.item-input')).autocomplete({
+						serviceUrl: wgServer + wgScript,
+						params: {
+							action: 'ajax',
+							rs: 'getLinkSuggest',
+							format: 'json'
+							//ns: categoryId
```
Warkot: Shouldn't this be removed?

extensions/wikia/CuratedContent/js/CuratedContentManagmentTool.js
```diff
@@ -2,416 +2,414 @@
...
+						optionCheckEmpty = options.indexOf('checkEmpty') !== -1;
+						optionRequired = options.indexOf('required') !== -1;
+						optionLimit = options.indexOf('limit') !== -1;
+					}
+
+					elements.each(function () {
+						var val = this.value,
+							$this = $(this);
+
+						// check if filed valuer is empty and it's required
```
Warkot: typo

extensions/wikia/CuratedContent/js/CuratedContentManagmentTool.js
```diff
@@ -2,416 +2,414 @@
...
+						if (optionRequired && val === '') {
+							$this.addError(requiredError);
+							return true;
```
Warkot: Those are not needed. http://api.jquery.com/each/ → 'You can stop the loop from within the callback function by returning false.'

vforge: I'll still need to return from function prematurely - I've added `true` just to explicitly say it's not breaking loop.

Warkot: You could also use `else if` as you do in the last check in this `each`.

rogatty:
LGTM

Warkot:
LGTM

vforge:
LGTM

PR #7720, opened 2015-07-01
ADEN-2192 In content ad placed incorrectly when having "display: none"

gabrys:


extensions/wikia/AdEngine/js/slot/inContent.js
```diff
@@ -69,9 +69,14 @@ define('ext.wikia.adEngine.slot.inContent', [
 	 */
 	function getSectionHeight(section) {
 		var $firstElement = section.$ad || section.$firstElement,
-			offsetTop = $firstElement.offset().top,
+			firstElementOffset = $firstElement.offset(),
+			offsetTop = firstElementOffset && firstElementOffset.top,
 			offsetBottom;
 
+		if (offsetTop === undefined) {
```
nandy-andy: I prefer `typeof(offsetTop) === 'undefined'` just-in-case if somebody overwrites `undefined` :japanese_ogre: But I agree it's a lot more chars :crying_cat_face:

nandy-andy:
LGTM :+1:

PR #7722, opened 2015-07-01
remove code for JavaScript Error Logging

owend:
- Feature was already deprecated/disabled
- This PR just removes the unused code

@macbre @ljagiello


lib/Wikia/src/Logger/Hooks.php
```diff
@@ -12,72 +12,6 @@
...
-
-		if ( !$wgDevelEnvironment ) {
-			$onError = $wgIsGASpecialWiki || $wgEnableJavaScriptErrorLogging;
-			$key = "wikialogger-top-script-$onError";
-			$loggingJs = $wgMemc->get( $key );
-
-			if ( !$loggingJs ) {
-				$errorUrl = "//jserrorslog.wikia.com/";
-				$loggingJs = "
-					function syslogReport(priority, message, context) {
```
macbre: This function is called in `resources/wikia/modules/log.js`. The call is "safe" as it's wrapped in an if statement, but let's remove it :)

owend: Oh yeah, there is actually quite a lot of code that references it... I will remove those references too.

owend:
There is a function called veTrack() which is defined in tracker.stub.js and used quite a lot in the VE.  It also calls syslogReport.  I am contacting that team about the best approach to take there.  @inez

hakubo:
lgtm

PR #7723, opened 2015-07-01
Use prefs service

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-442

Replace calls to `getOption`, `setOption` in the `User` class to use the preference service (for preferences). Change others to call `getOptionHelper` and `setOptionHelper` and add logging when calling `getOption`, `setOption`.


includes/User.php
```diff
@@ -228,6 +235,9 @@ class User {
 	 */
 	private $mBlockedFromCreateAccount = false;
 
+	/** @var UserPreferences */
+	private $preferences;
```
drsnyder: Minor: I would prefer to call this something like `$preferenceService`. Something with "service" in the name so it's clear what we are working with.

includes/User.php
```diff
@@ -2481,14 +2504,52 @@ public static function createLocalOptionName($option, $cityId = null, $sep = '-'
...
+		global $wgPreferencesUseService;
+
+		if ($wgPreferencesUseService) {
+			$this->load();
+
+			if ($value) {
+				$value = str_replace("\r\n", "\n", $value);
+				$value = str_replace("\r", "\n", $value);
+				$value = str_replace("\n", " ", $value);
+			}
```
drsnyder: Minor: This could be separated out into a helper function since it's at a lower level of abstraction than the rest of the function.

includes/User.php
```diff
@@ -2566,6 +2629,8 @@ public function getLocalFlag($flag, $cityId = null, $sep = "-") {
 		if (!isset($cityId)) {
 			$cityId = $wgCityId;
 		}
```
drsnyder: Is this block still necessary since it's handled [below](https://github.com/Wikia/app/pull/7723/files#diff-897fba1dd6b8edc3238762e35d0b7c71R2668)?

includes/User.php
```diff
@@ -2585,7 +2650,41 @@ public function getGlobalFlag($flag, $default = null) {
...
+	 * preferences, attributes or flags should be of the form "{option}-{cityId}"
+	 *
+	 * IF YOU USE $sep, MAKE A PLAN TO NORMALIZE IT TO "-"!
+	 *
+	 * @param string $property
+	 * @param int $cityId [optional]
+	 * @param string $sep the separator between the option and the id.
+	 * @return string
+	 */
+	public static function localToGlobalPropertyName($property, $cityId = null, $sep = '-') {
```
drsnyder: Minor: Is `localToGlobalPropertyName` the right name for this? On the surface I would expect this to drop the `$wgCityId` based solely on the name.

includes/wikia/DefaultSettings.php
```diff
@@ -1748,3 +1748,8 @@
  * ONLY UPDATE THROUGH WIKI FACTORY ON COMMUNITY - it's an instant global.
  */
 $wgAdDriverIncontentPlayerSlotCountries = null;
+
+/**
+ * flag for using (possibly external) service for managing a user's preferences
+ */
+$wgPreferencesUseService = true;
```
drsnyder: Should this default to `false` until we add the config changes?

drsnyder:
Minor comments, OTW LGTM. Looks like the tests are [still failing though](http://qa-s2.wikia-prod:8080/job/unitTests_php_branches/17363/).

includes/User.php
```diff
@@ -2514,6 +2575,8 @@ public function getLocalAttribute($attr, $cityId = null, $sep = "-") {
 		if (!isset($cityId)) {
 			$cityId = $wgCityId;
 		}
+
+		return $this->getGlobalAttribute(self::localToGlobalPropertyName($attr, $cityId, $sep));
```
ArturKlajnerok: why not use `getOptionHelper` directly ?

includes/User.php
```diff
@@ -2546,11 +2599,21 @@ public function setGlobalAttribute($attribute, $value) {
 	 * @param string $attribute
 	 * @param string $value
 	 * @param int $cityId
-	 * @param char $sep
+	 * @param string $sep
 	 * @return bool
 	 */
 	public function setLocalAttribute($attribute, $value, $cityId = null, $sep = '-') {
-		return $this->setOption(self::createLocalOptionName($attribute, $cityId, $sep), $value, $sep);
+		$this->setGlobalAttribute(self::localToGlobalPropertyName($attribute, $cityId, $sep), $value);
```
ArturKlajnerok: why not use `setOptionHelper` directly ?

includes/User.php
```diff
@@ -2317,9 +2349,8 @@ public function setRealName( $str ) {
 	 * @return String User's current value for the option
 	 * @see getBoolOption()
 	 * @see getIntOption()
-	 * @deprecated use get(Global|Local)Preference  get(Global|Local)Attribute or get(Global|Local)Flag
 	 */
-	public function getOption( $oname, $defaultOverride = null, $ignoreHidden = false ) {
+	public function getOptionHelper( $oname, $defaultOverride = null, $ignoreHidden = false ) {
```
ArturKlajnerok: should `getOptionHelper` be private ?

nmonterroso: oops. yeah, i thought i made it that way.

includes/User.php
nmonterroso: So when we introduce the attribute service, there's one less thing to worry about :P

includes/User.php
nmonterroso: Good point

includes/User.php
nmonterroso: Good point. done in 6b59238

includes/User.php
nmonterroso: Hm, yeah, I guess it can be seen that way. The way I was looking at it was "Here's a local property for this wiki, how can I reference this globally?"

includes/wikia/DefaultSettings.php
nmonterroso: ALL IN

nmonterroso: what config changes?

drsnyder: I was imagining that we'll set this to `false` by default then tweak the config with something like:

```php
if ( ($wgCityId % 100) < 5) {
 $wgPreferencesUseService = true;
}
```

nmonterroso: Ohh.. ok, I suppose we can ramp that too.

nmonterroso: default to false in 349227f

PR #7725, opened 2015-07-01
SOC-458 Send Facebook disconnect email using new extension

mklucsarits:
https://wikia-inc.atlassian.net/browse/SOC-458


extensions/wikia/Email/Controller/FacebookDisconnectController.class.php
```diff
@@ -0,0 +1,41 @@
...
+	/**
+	 * @template temporaryPassword
+	 */
+	public function body() {
+		$this->response->setData( [
+			'salutation' => $this->getSalutation(),
+			'summary' => $this->getSummary(),
+			'passwordIntro' => $this->getIntro(),
+			'username' => $this->getMessage( 'emailext-fbdisconnect-username',
+				$this->targetUser->getName() )->parse(),
```
garthwebb: You only need `parse` if the message contains wikitext.  For simple substitutions (e.g. `$1`, `$2`), `text()` will suffice.  This has the advantage of not firing up the parser for the string.

extensions/wikia/Email/Controller/FacebookDisconnectController.class.php
```diff
@@ -0,0 +1,41 @@
...
+			'tempPassword' => $this->getMessage( 'emailext-fbdisconnect-password',
+				$this->tempPass )->parse(),
```
garthwebb: Same, can use `text()` here.

extensions/wikia/FacebookClient/FacebookClientController.class.php
```diff
@@ -151,21 +151,18 @@ public function disconnectFromFB() {
...
+		// Create a temporary password for the user
+		$userService = new \UserService();
+		$tempPass = $userService->resetPassword( $user );
+
+		// Send email to user with temp password, telling them their FB account is disconnected
+		$emailParams = [
+			'targetUser' => $user,
+			'tempPass' => $tempPass,
+		];
+		F::app()->sendRequest( 'Email\Controller\FacebookDisconnect', 'handle', $emailParams );
```
garthwebb: This does actually send a status back.  The `sendRequest` method returns a `WikiaResponse` object on which you can call `getData()`.  This will be an array with a `result` key that will either be `ok` or some type of error key.  For errors, the error message will be in `msg`.

We don't use this everywhere since a lot of places that send email didn't check the status of the email anyway (its a fire and forget process that is a side effect to something else the user was doing).  Here though, since this code actually does look for an error, we should probably keep and update it.

extensions/wikia/Email/Controller/FacebookDisconnectController.class.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+namespace Email\Controller;
+
+use Email\EmailController;
+use Email\Tracking\TrackingCategories;
+
+class FacebookDisconnectController extends ForgotPasswordController {
+	const TRACKING_CATEGORY = TrackingCategories::TEMPORARY_PASSWORD;
```
jsutterfield: Don't need to define `TRACKING_CATEGORY` here since it's inherited from `ForgotPasswordController`.

extensions/wikia/Email/templates/temporaryPassword.mustache
```diff
@@ -3,8 +3,11 @@
 	<tr>
 		<td class="widest">
 			<p>{{passwordIntro}}</p>
+			{{# username}}
+				<p><pre>{{username}}</pre></p>
```
jsutterfield: Are username and password supposed to be surrounded in `<pre>` tags? The [designs](https://docs.google.com/document/d/1OM6xL0SNvkCyr7if8bwslFIeZWwwjW1Sdke2Y1A_s5A/edit) don't look like they call for them.

mklucsarits: The forgot-password email design didn't specify a preformatted display for the temp password, but @garthwebb added it. Should I remove the `pre` tags or leave them? @garthwebb what do you think?

https://docs.google.com/document/d/1BCds1YL-olhClDVPBoi43Fmkn6uQlf-7tolRdjCyDo4/edit

jsutterfield: I'd vote for removing them myself since the designs don't call for them.

garthwebb: Well, the designs don't really go to that level of detail so I just added &lt;pre>'s for the password to help it stand out.  I don't really have a strong opinion about it.

jsutterfield:
LGTM

PR #7726, opened 2015-07-01
SOC-859 update created page emails to use edit page email templates

jsutterfield:
Ticket: https://wikia-inc.atlassian.net/browse/SOC-859


extensions/wikia/Email/Controller/WatchedPageController.class.php
```diff
@@ -201,7 +201,7 @@ protected static function getEmailSpecificFormFields() {
 	}
 }
 
-class WatchedPageEditedController extends WatchedPageController {
+class WatchedPageEditedOrCreatedController extends WatchedPageController {
```
mklucsarits: Just my opinion, but I don't think this more verbose name was necessary. I think, in MediaWiki terms, we already know that a created page is considered an "edit" of that page. I don't care either way though what the class is named.

includes/EmailNotification.php
```diff
@@ -357,29 +357,16 @@ private function composeCommonMailtext() {
...
-			list ( $keys['$NEWPAGE'], $keys['$NEWPAGEHTML'] ) = wfMsgHTMLwithLanguageAndAlternative (
-				'enotif_lastvisited',
-				'enotif_lastvisited',
-				F::app()->wg->LanguageCode,
-				[],
-				$this->title->getFullUrl( 's=wldiff&diff=0&previousRevId=' . $this->previousRevId )
-			);
-			$keys['$OLDID']   = $this->previousRevId;
-			$keys['$CHANGEDORCREATED'] = wfMessage( 'changed' )->inContentLanguage()->plain();
+		if ( $action == '' ) {
```
bkoval: Would empty() work here? Just asking, I really have no preference on this, but in the code we're using it more often that the empty string comparison

bkoval:
LGTM :+1:

mklucsarits:
LGTM

PR #7731, opened 2015-07-02
PLATFORM-1320 - AbPerformanceTesting: phase #2

macbre:
- backend: Add a delay of 0 / 100 / 200 / 300 / 400 ms (for 1% of traffic each)
- frontend: Add a delay of 0 / 200 / 400 / 600 / 800 ms (for 1% of traffic each)

And enable on all wikis - https://github.com/Wikia/config/pull/1182

Merging into release branch only as we want to run the test during the upcoming weekend only.

@jcellary / @wladekb


jcellary:
LGTM

PR #7736, opened 2015-07-02
There was an undocumented status code which is a success-code (it app…

SeanColombo:
…ears to mean that they know of the song, they just don't have lyrics for it, which doesn't really affect us since we use our own anyway). When we don't recognize a code, we return a 404 on the tracker, so I've added this code to now be recognized.

### DIFF
--- a/extensions/3rdparty/LyricWiki/LyricFind/LyricFindTrackingService.class.php
+++ b/extensions/3rdparty/LyricWiki/LyricFind/LyricFindTrackingService.class.php
@@ -11,6 +11,10 @@ class LyricFindTrackingService extends WikiaService {
 	const CODE_LRC_IS_AVAILABLE  = 111;
 	const CODE_LYRIC_IS_BLOCKED  = 206;
 
+	// Not documented. The response body says "SUCCESS: NO LYRICS" which I assume means that they
+	// have licensing in place, they just don't have lyrics for the song.
+	const CODE_SUCCESS_NO_LYRICS = 106;
+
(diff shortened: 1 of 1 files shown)


macbre:
LGTM

extensions/3rdparty/LyricWiki/LyricFind/LyricFindTrackingService.class.php
```diff
@@ -136,6 +140,7 @@ public function track($amgId, $gracenoteId, Title $title) {
 					$this->markLyricForRemoval($this->wg->Title->getArticleID());
 					break;
 
+				case self::CODE_SUCCESS_NO_LYRICS:
```
macbre: Worth adding a test case in `LyricFindTrackingServiceTest`?

extensions/3rdparty/LyricWiki/LyricFind/LyricFindTrackingService.class.php
```diff
@@ -213,6 +218,7 @@ public static function isPageBlockedViaApi($amgId="", $gracenoteId="", $pageTitl
 					$isBlocked = true;
 					break;
 
+				case self::CODE_SUCCESS_NO_LYRICS:
```
macbre: Worth adding a test case in `LyricFindTrackingServiceTest`?

SeanColombo: I looked into it... I don't think it would really add anything in this case (we don't exhaustively test every valid & invalid status code & it would be hard to do it right without knowing all of the other codes... since any other bugs would just presumably be undocumented codes: and we'd be writing a test to say what we think the code should return, without knowing what it actually should return).

PR #7737, opened 2015-07-02
Ce 2126 templatedraft clicktracking

kamilkoterba:
https://wikia-inc.atlassian.net/browse/CE-2126

Clicktracking and some conditions simplifications

@Wikia/community-engineering

PR #7739, opened 2015-07-02
(INT-129) Add API method for getting the converted infobox markup

Grunny:
Add API method for getting the converted portable infobox markup.

### DIFF
--- a/extensions/wikia/TemplateDraft/controllers/TemplateDraftController.class.php
+++ b/extensions/wikia/TemplateDraft/controllers/TemplateDraftController.class.php
@@ -58,6 +58,36 @@ class TemplateDraftController extends WikiaController {
 		return true;
 	}
 
+	/**
+	 * Get the converted infobox markup of a given template.
+	 *
+	 * @requestParam string template The name of the template to convert.
+	 */
+	public function getInfoboxMarkup() {
+		$this->response->setFormat( WikiaResponse::FORMAT_JSON );
+
+		$templateName = $this->request->getVal( 'template' );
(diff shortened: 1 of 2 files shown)


extensions/wikia/TemplateDraft/controllers/TemplateDraftController.class.php
```diff
@@ -58,6 +58,36 @@ public function markTemplateAsNotInfobox() {
...
+	 * @requestParam string template The name of the template to convert.
+	 */
+	public function getInfoboxMarkup() {
+		$this->response->setFormat( WikiaResponse::FORMAT_JSON );
+
+		$templateName = $this->request->getVal( 'template' );
+		$title = Title::newFromText( $templateName, NS_TEMPLATE );
+		if ( !$title instanceof Title
+			|| !$title->exists()
+			|| !$title->inNamespace( NS_TEMPLATE )
```
kamilkoterba: I think we want to use [TemplateDraftHelper::allowedForTitle](https://github.com/Wikia/app/blob/CE-2126-templatedraft-clicktracking/extensions/wikia/TemplateDraft/TemplateDraftHelper.class.php#L90)
once its merged (https://github.com/Wikia/app/pull/7737)

Grunny: Parts of it, but for this use case we don't really want or need to check anything to do with the parent or the `templatedraft` permission. IMO there's no reason to restrict just getting the converted wikitext from the API to anything more than `read` which happens automatically for Nirvana controllers, especially as it's aimed at being able to convert a template on one community to be used on another.

PR #7741, opened 2015-07-02
SOC-535 Send username change email using new extension

mklucsarits:
https://wikia-inc.atlassian.net/browse/SOC-535


garthwebb:
LGTM

extensions/wikia/UserRenameTool/UserRenameTask.class.php
```diff
@@ -8,6 +8,8 @@
 use Wikia\Tasks\Tasks\BaseTask;
 
 class UserRenameTask extends BaseTask {
+	const EMAIL_CONTROLLER = 'Email\Controller\UserNameChange';
```
jsutterfield, line 11: Nice! Class constants! :)

extensions/wikia/Email/templates/userNameLayout.mustache
```diff
@@ -0,0 +1,20 @@
...
+		</td>
+		<td class="widest border">
+			<div class="message">
+				<p>
+					{{userNameMessage}}
+				</p>
+			</div>
+		</td>
+	</tr>
+	{{# hasContentFooterMessages}}
```
jsutterfield: Is this template used by a controller which does't have contentFooterMessages? If not, can we get rid of this conditional?

extensions/wikia/Email/styles/userNameLayout.css
```diff
@@ -0,0 +1,36 @@
...
+
+.username {
+	font-size: 12px;
+}
+
+.message {
+	margin: 20px 0;
+}
+
+.message p {
```
jsutterfield, line 28: It's a small thing, but I think our coding conventions call for css/scss to be ordered alphabetically. Can we move these 2 `.message` rules up above `.username`?

extensions/wikia/Email/templates/userNameLayout.mustache
mklucsarits: The controller is outputting contentFooterMessages, although they could be renamed to other parameters for this template, if that's what you're implying.

jsutterfield: What I mean is that I don't think we need to have this `{{# hasContentFooterMessages}}` check in this template since we're always going to display the `{contentFooterMessages}}`.

In the avatarLayout template this check makes sense since it's shared by a number of emails, some of which that show contentFooterMessages and some that don't. If this template is only used by emails which show contentFooterMessages, we don't have to have that check.

jsutterfield:
LGTM

PR #7742, opened 2015-07-02
SOC-766 Update Community Central URLs in translated messages

mklucsarits:
There was a change made to these i18n messages after they were submitted for translation, and the updated i18n file was never uploaded to Crowdin. This fixes the problem caused by that mishap.


garthwebb:
LGTM

jsutterfield:
LGTMx2!

PR #7747, opened 2015-07-02
return correct authentication method from helios

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-439

Allow Helios to modify the return value for authentication in `SpecialUserLogin->authenticateUserData`. This was done via hook to avoid breaking the return values between the various authentication functions and hooks, which expect `true`/`false` as return types. 

~~Wikia/helios#78 should be merged/deployed first~~


ArturKlajnerok:
Lgtm :+1: 
/cc @garthwebb @kvas-damian

PR #7752, opened 2015-07-03
(CE-2214) Fix ArticleAsJson output when a page has Flags

Grunny:
ArticleAsJson modifies the ParserOutput before we do, and converts it to
JSON. As a quick fix, this handles this situation when modifying the
parser output.

We'll soon take flags out of the content for Mercury and have them handled
in their own component, but we'll also follow this up with investigating alternative
places to modify the parser output.

/cc @Wikia/community-engineering @rogatty


lukaszkonieczny:
lgtm

PR #7753, opened 2015-07-03
UncycloUserMigrator: authenticate all accounts that come from Uncyclopedia

macbre:
Avoid Uncyclopedia users be deleted by the cleanup script removing accounts without verified email address (`/extensions/wikia/UserLogin/maintenance.php`).

``` sql
mysql@x.x.x.x[uncyclo]>select count(*) from user where user_email_authenticated is null;
+----------+
| count(*) |
+----------+
|   112559 |
+----------+
1 row in set (0.07 sec)
```

@michalroszka


michalroszka:
:+1:

PR #7754, opened 2015-07-03
(CE-2214) Disable Flags in Mercury for now

Grunny:
It doesn't look great and breaks some pages in Mercury.

/cc @Wikia/community-engineering @rogatty


rogatty:
LGTM

PR #7755, opened 2015-07-03
CONCF-560 | Layout broken for long titles on FF

tomasznapieralski:
https://wikia-inc.atlassian.net/browse/CONCF-560

https://bugzilla.mozilla.org/show_bug.cgi?id=1136818

### DIFF
--- a/skins/oasis/css/core/PageHeader.scss
+++ b/skins/oasis/css/core/PageHeader.scss
@@ -236,8 +236,10 @@
 		&.header-title {
 			@include align-self(flex-start);
 			@include flex(0 1 100%);
-			order: 0;
+			// This is fix for firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1136818
+			// Jira ticket: https://wikia-inc.atlassian.net/browse/CONCF-560
 			min-width: 0;
+			order: 0;
 		}
 
 		&.header-tally {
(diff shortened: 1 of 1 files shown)


tomasznapieralski:
@vforge @bognix

skins/oasis/css/core/PageHeader.scss
```diff
@@ -237,6 +237,7 @@
 			@include align-self(flex-start);
 			@include flex(0 1 100%);
 			order: 0;
+			min-width: 0;
```
Warkot: CSS properties should be in alphabetical order. This rule is common and is also added to our [Guidelines repo](https://github.com/Wikia/guidelines). You can clone it and then install pre-commit hooks from there and/or enable scss-lint plugin in Php/WebStorm and point to config file from the repo.

Second thing is that when you're doing such ~~hacks~~ workarounds, add a comment why it's done. For someone who didn't work on the issue, who has no context, this line looks like something that does nothing and can be deleted. Please add the ticket number in the comment and a link to the Mozilla bug board, few words that it handles text wrapping would be nice, too.

vforge: I second @Warkot comment

Warkot:
LGTM :+1:

rogatty:
LGTM

vforge:
LGTM

PR #7758, opened 2015-07-06
CE-1994: Fetch labels for infobox markup

lukaszkonieczny:


extensions/wikia/TemplateDraft/TemplateConverter.class.php
```diff
@@ -160,6 +181,74 @@ private function prepareVariables( $templateVariables ) {
...
+	 * Prepares content in special format.
+	 * It doesn't parse template variables to extract them later.
+	 *
+	 * @param $content
+	 * @return string
+	 */
+	private function prepareContent( $content ) {
+		global $wgUser;
+
+		$ParserPool = \ParserPool::get();
```
Grunny: The variable name is misleading, as you're not getting a `ParserPool` object, you're getting a `Parser` object. It also should use lowerCamelCase. I'd just rename it to `$parser`.

Grunny:
LGTM.

PR #7762, opened 2015-07-06
VOLDEV-117

kerriamber:
Also fixes this issue for pages with ArticleComments enabled. Original code is from #1060 (BAC-535).

ping @Grunny


mszabo-wikia:
@macbre pinging you for Phalanx

extensions/wikia/ArticleComments/classes/ArticleCommentList.class.php
```diff
@@ -511,17 +511,37 @@ public function blockedPage() {
...
+
+				foreach($groups as $group) {
+					if (in_array($group, $blockerGroups)) {
+						$blockerLink = wfMessage("group-$group")->plain();
+					}
+				}
+			}
+		} else {
+			$blockerLink = '[[' . $wgContLang->getNsText( NS_USER ) . ":{$blockerName}|{$blockerName}]]";
+		}
```
macbre, line 552: Thanks for the fix. LGTM.

While VOLDEV-117 says that we should make this change for comments only, I think that this should be consistent across all places where block information is presented to the user. Given that let's introduce a global helper function (bind to a block-related MW hook?) that will apply this logic for all blocks.

@Wikia/community-engineering

kerriamber, line 552: @macbre: this change is already implemented for all other times when block
information is shown to the user - the current implementation does not work
with article / blog comments, though, and this should rectify that

macbre, line 552: Thanks @sactage, so LGTM

PR #7772, opened 2015-07-07
Ce 2234 conversion link on insights

lgarczewski:
...also fixes the layout issue with the alt action button (Convert) on the insights page

### DIFF
--- a/extensions/wikia/Insights/models/InsightsUnconvertedInfoboxesModel.php
+++ b/extensions/wikia/Insights/models/InsightsUnconvertedInfoboxesModel.php
@@ -43,11 +43,12 @@ class InsightsUnconvertedInfoboxesModel extends InsightsQuerypageModel {
 	}
 
 	public function getAltActionUrl( Title $title ) {
-		if ( !class_exists( 'TemplateConverter' ) ) {
+		$subpage = Title::newFromText( $title->getText() . "/" . wfMessage('templatedraft-subpage')->escaped() , NS_TEMPLATE );
+
+		if ( $subpage === null ) {
+			// something went terribly wrong, quit early
 			return '';
 		}
-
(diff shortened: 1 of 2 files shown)


extensions/wikia/Insights/models/InsightsUnconvertedInfoboxesModel.php
```diff
@@ -43,9 +43,16 @@ public function hasAltAction() {
 	}
 
 	public function getAltActionUrl( Title $title ) {
-		$subpage = Title::newFromText( $title->getText() . "/Draft", NS_TEMPLATE );
+		if ( !class_exists( 'TemplateConverter' ) ) {
+			return '';
```
adamkarminski: Returning an empty string will still cause the button to appear with an empty `href` attribute. What about adding a check [here](https://github.com/Wikia/app/blob/dev/extensions/wikia/Insights/models/InsightsQuerypageModel.php#L410)?

extensions/wikia/Insights/models/InsightsUnconvertedInfoboxesModel.php
```diff
@@ -43,9 +43,16 @@ public function hasAltAction() {
...
+		}
 
-		return $subpage->getFullUrl( [ 'action' => 'edit' ] );
+		$subpage = Title::newFromText( $title->getText() . "/" . wfMessage('templatedraft-subpage')->escaped() , NS_TEMPLATE );
+
+		return $subpage->getFullUrl( [
```
adamkarminski: We had major issues with `Title::newFromText` returning null if a message key was passed to it due to message rebuild issues. Worth a check for `$subpage === null`?

adamkarminski:
LGTM

PR #7778, opened 2015-07-07
fix wrong replacement for setOption to setGlobalAttribute

sqreek:
should be set
https://github.com/Wikia/app/compare/release-309...release-310#diff-6bbd52bf955477230a9e62f57ae13d8dR249


nmonterroso:
LGTM - after tests let's merge

PR #7779, opened 2015-07-07
[SERVICES-487] Extended time to 60 days before unconfirmed accounts are removed

ArturKlajnerok:
Before we provide final solution for user account removal, I'm extending time interval to 60 days, so we won't remove any more unconfirmed accounts that are active.

/cc @Wikia/services-team @garthwebb @kvas-damian @macbre

### DIFF
--- a/extensions/wikia/UserLogin/maintenance.php
+++ b/extensions/wikia/UserLogin/maintenance.php
@@ -110,7 +110,7 @@
 				'user_email_authenticated' => NULL,
 				'up_property' => UserLoginSpecialController::NOT_CONFIRMED_SIGNUP_OPTION_NAME,
 				'up_value' => 1,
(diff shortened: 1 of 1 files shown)


nmonterroso:
LGTM - let's make sure we comment on the ticket that this is the change we're doing, so that we can figure out the right way to go about this in the next (or next-next) sprint

macbre:
@ArturKlajnerok, can we instead disable this script temporarily (and re-enable once the fix is live and all affected accounts are marked properly)?

nmonterroso:
We just spoke with Christina, and here's a related ticket: https://wikia-inc.atlassian.net/browse/SOC-597

So we can/should disable this script indefinitely.

ArturKlajnerok:
Closing this, as the decision on disabling the removal script was already done.

PR #7780, opened 2015-07-07
[SERVICES-487] Removed unconfirmed users removal maintenance.

ArturKlajnerok:
Based on decision made in https://wikia-inc.atlassian.net/browse/SERVICES-487, we should not remove accounts, even if they're not confirmed.

/cc @Wikia/services-team @garthwebb @kvas-damian @macbre

### DIFF
--- a/extensions/wikia/UserLogin/maintenance.php
+++ b/extensions/wikia/UserLogin/maintenance.php
@@ -2,7 +2,6 @@
 
 	/**
 	 * Maintenance script to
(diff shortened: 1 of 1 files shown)


nmonterroso:
LGTM - do we need to disable the calling of this specific cron anywhere?

ArturKlajnerok:
@nmonterroso  Cron job recipe is removed: https://github.com/Wikia/chef-repo/pull/5790, also @ljagiello removed the job from cron-s5.

PR #7781, opened 2015-07-07
Use the UserLogout hook

garthwebb:
Avoid having to check for the wgEnableFacebookExt global directly.  This addresses feedback in a previous PR here: https://github.com/Wikia/app/pull/7604/files#r33021504

### DIFF
--- a/includes/specials/SpecialUserlogout.php
+++ b/includes/specials/SpecialUserlogout.php
@@ -44,12 +44,6 @@ class SpecialUserlogout extends UnlistedSpecialPage {
 			throw new HttpError( 400, wfMessage( 'suspicious-userlogout' ), wfMessage( 'loginerror' ) );
 		}
 
-		// Some communities (notably uncyclopedia) do not have FacebookClient enabled
-		if ( F::app()->wg->EnableFacebookClientExt ) {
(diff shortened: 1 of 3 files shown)


includes/specials/SpecialUserlogout.php
```diff
@@ -44,12 +44,6 @@ function execute( $par ) {
 			throw new HttpError( 400, wfMessage( 'suspicious-userlogout' ), wfMessage( 'loginerror' ) );
 		}
 
-		// Some communities (notably uncyclopedia) do not have FacebookClient enabled
-		if ( F::app()->wg->EnableFacebookClientExt ) {
```
mklucsarits: Why is this check no longer needed?

garthwebb: @mklucsarits because I moved the code to a ```FacebookClient``` file that will only be included if ```EnableFacebookClientExt``` is true.  That's one of the benefits of the hook system (as much as I hate it) which is conditionally included extensions can register for hooks and that code will only be called if the extension has been loaded and you don't have to do extra checks for these ```wg``` variables.

mklucsarits:
LGTM

jsutterfield:
LGTM

PR #7786, opened 2015-07-08
CE-2263 Fix removing flags if there are no parameters assigned

adamkarminski:
@Wikia/community-engineering 

The PR prevents passing a `null` to `compareDataAndGetFlagsToChange` which causes an error.
`null` is sent from the edit form if no flag type on a given wikia has a parameter assigned and no flags are checked in the form. Casting `null` to `array` sends an empty array to the method and everything works fine.


Grunny:
LGTM.

PR #7790, opened 2015-07-08
change right rail view permissions to all users

MariuszCz:


lgarczewski:
Please also correct usnit tests.

lgarczewski:
LGTM.

lgarczewski:
Needs a correction -- should only show to users who can edit the given template page.

PR #7791, opened 2015-07-08
SERVICES-492 - Added profiler for UserPreferences get/set methods.

harnash:
Added additional (sampled at 1%) log to get/set user preferences.


harnash:
@Wikia/services-team

nmonterroso:
I thought that there was already a mechanism in MediaWiki for sending timing stats to grafana? /cc @wladekb

harnash:
After consultations with @wladekb I've decided to use Transactions to send data "directly" to InfluxDB. This will allow us to skip writing specific collector which will send data to InfluxDB.

lib/Wikia/src/Service/User/PreferenceKeyValueService.php
```diff
@@ -24,8 +24,13 @@
 use Wikia\Persistence\User\PreferencePersistence;
 
 class PreferenceKeyValueService implements PreferenceService {
+	use \Wikia\Util\WikiaProfiler;
 
+	/**
+	 * @var PreferencePersistence
+     */
```
wladekb: indentation is off here

wladekb:
lgtm

lib/Wikia/src/Service/User/PreferenceKeyValueService.php
harnash: Fixed in 7422ad68c0e6e4df45feb2dac813fbd9b214d5e6

nmonterroso:
This just feels like we're recreating `wfProfileIn` and `wfProfileOut`.. is this really how we send stuff to InfluxDB? If so I guess LGTM, but I thought we didn't have to do the "start timing", "end timing" anymore

nmonterroso:
LGTM

PR #7792, opened 2015-07-08
CE-1929: Move classes to extract data from template and to classify them

lukaszkonieczny:
Template classification is not only connected with infobox migration so I moved it to more commons place (includes/wikia), as well as part TemplateConverter as TemplateDataExtractor.


includes/wikia/TemplateClassification.class.php
```diff
@@ -33,13 +33,12 @@ class TemplateClassificationController extends WikiaController {
 
 	/**
 	 * A Title object for the page you are classifying.
-	 * @var Title
+	 * @var \Title
```
adamkarminski: Since the class is in the main namespace you don't need the backslashes :)

extensions/wikia/EditPageLayout/EditPageLayoutHelper.class.php
```diff
@@ -195,8 +195,8 @@ static public function isCodeSyntaxHighlightingEnabled( Title $articleTitle ) {
 	static public function isInfoboxTemplate( Title $title ) {
 		$namespace = $title->getNamespace();
 
-		if ( $namespace === NS_TEMPLATE && class_exists( 'TemplateClassificationController' ) ) {
-			$tc = new TemplateClassificationController( $title );
+		if ( $namespace === NS_TEMPLATE && class_exists( 'TemplateClassification' ) ) {
```
adamkarminski: I think that since `TemplateClassification` is included in the `DefaultSettings.php` we can skip this check safely. :)

adamkarminski:
Beside two minor comments it LGTM :)

PR #7794, opened 2015-07-08
SOC-964 update method override to use same signature as parent

jsutterfield:
Ticket: https://wikia-inc.atlassian.net/browse/SOC-964

### DIFF
--- a/includes/WikiFilePage.php
+++ b/includes/WikiFilePage.php
@@ -57,12 +57,13 @@ class WikiFilePage extends WikiPage {
 	}
 
 	/**
+	 * @param int $flags a bit field; may be Title::GAID_FOR_UPDATE to use master db
 	 * @return mixed|null|Title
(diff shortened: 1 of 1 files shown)


mklucsarits:
Well that was simple. LGTM.

includes/WikiFilePage.php
```diff
@@ -57,12 +57,13 @@ protected function loadFile() {
 	}
 
 	/**
+	 * @param int $flags a bit field; may be Title::GAID_FOR_UPDATE to use master db
 	 * @return mixed|null|Title
 	 */
-	public function getRedirectTarget() {
+	public function getRedirectTarget( $flags = 0 ) {
```
garthwebb: Maybe ```0``` should be a constant here?


Just kidding ...

garthwebb:
LGTM

PR #7796, opened 2015-07-08
Hg 754 Add section parsing to article API

lizlux:
This PR updates the mercury article API so that sections can be requested one by one or by 'all', and the response will be an array of sections. If the sections param is not present, the article will be represented as a string (as it has been). 

Given a base URL of  `http://muppet.liz.wikia-dev.com/wikia.php?controller=MercuryApi&method=getArticle&title=kermit`, here's some sample URLs that are now available: 
- `&sections=all`
- `&sections=1`
- `&sections=0,1,2,3`

Note that this will only return top level sections (h2 tags), and any sub-sections (h3-6 tags) will be included in those top-level sections. 

@hakubo @aquilax ;)


extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -127,12 +127,13 @@ private function getArticleDescription( $articleId, $descLength = 100 ) {
 	 * @param int $articleId
 	 * @return array
 	 */
-	private function getArticleJson( $articleId, Title $title ) {
+	private function getArticleJson( $articleId, Title $title, $sections = [] ) {
```
garthwebb: Should add ```$title``` and ```$sections``` to the function doc.

includes/wikia/api/ArticlesApiController.class.php
```diff
@@ -1021,6 +1023,12 @@ public function getAsJson() {
 
 		if ( $parsedArticle instanceof ParserOutput ) {
 			$articleContent = json_decode( $parsedArticle->getText() );
+
+			if ( !empty( $sectionsToGet ) ) {
```
garthwebb: I usually feel like its clearer to test for the positive when an ```if``` statement has an ```else``` that you can flip around.

kenkouot:
As per our discussion, this API should also support encoded JSON arrays as input.

includes/wikia/api/ArticlesApiController.class.php
```diff
@@ -1048,6 +1056,73 @@ public function getAsJson() {
...
+	/**
+	 * Get the HTML for an article section
+	 * @param string $articleAsWikiText
+	 * @param Title $title
+	 * @param integer $section
+	 * @return string
+	 * @todo Supposedly getSection works with HTML, but I could only get it to work with wikitext
+	 */
+	private function getArticleSection( $articleAsWikiText, $title, $section ) {
+		/** @var $wgParser Parser */
```
garthwebb: The datatype should come first here, and since its a single variable declaration, the variable name can actually be left off.

garthwebb:
LGTM

includes/wikia/api/ArticlesApiController.class.php
```diff
@@ -1048,6 +1056,86 @@ public function getAsJson() {
...
+	 * @param Title $title
+	 * @return array
+	 * @throws BadRequestApiException
+	 */
+	private function getArticleSections( $sectionsToGet, ParserOutput $parsedArticle, Article $article, $title ) {
+		$content = [];
+		$articleAsWikiText = $article->getContent();
+		$parsedSections = $parsedArticle->getSections();
+
+		// decode strings like "%5B1,2,3%5D" or "[1,2,3]"
```
lizlux: @kenkouot

includes/wikia/api/ArticlesApiController.class.php
```diff
@@ -1048,6 +1056,86 @@ public function getAsJson() {
...
+		if ( is_string( $sectionsToGet ) ) {
+			$sectionsToGet = json_decode( urldecode( $sectionsToGet ) );
+		}
+		// convert "all" or ["all"] to array of numbers
+		if ( $sectionsToGet === 'all' || in_array( 'all', $sectionsToGet ) ) {
+			$sectionsToGet = range( 0, count( $parsedArticle->getSections() ) );
```
aquilax: `$parsedSections` should already contain `$parsedArticle->getSections()` at this point.

includes/wikia/api/tests/ArticlesApiControllerTest.php
```diff
@@ -166,6 +166,49 @@ public function reorderForLinksDataProvider() {
 		];
 	}
 
+	public function replaceSectionIndexDataProvider() {
+		return [
+			// section text outside of heading tag
+			[
+				'<h2 section="1">Foo</h2><p>section=\'2\'</p>',
+				'3',
```
aquilax: These may better be number as the function documentation states that integer is expected as `$section`;

includes/wikia/api/ArticlesApiController.class.php
```diff
@@ -1048,6 +1056,86 @@ public function getAsJson() {
...
+	 * @return array
+	 * @throws BadRequestApiException
+	 */
+	private function getArticleSections( $sectionsToGet, ParserOutput $parsedArticle, Article $article, $title ) {
+		$content = [];
+		$articleAsWikiText = $article->getContent();
+		$parsedSections = $parsedArticle->getSections();
+
+		// decode strings like "%5B1,2,3%5D" or "[1,2,3]"
+		if ( is_string( $sectionsToGet ) ) {
```
aquilax: It at this point `$sectionsToGet` is 'all', it will get `json_decode`-d and will end being `null` after this block;

lizlux:
thanks @aquilax!

includes/wikia/api/ArticlesApiController.class.php
SebastianMarzjan: What is the current source of requests passing arrays as json-encoded strings? I mean, we have multiple types of passing an array as an argument in our stack already:
* JSON payload - an object of params
* ?sections[]=1&sections[]=2&sections[]=3 (common for directly calling PHP apps)
* ?sections=1,2,3 (most core MW apis and many Nirvana ones) 

And I haven't seen this one here yet - is is used somewhere already, or is it us introducing it here?

rogatty: IMO `?sections=1,2,3` fits the best in here.

includes/wikia/api/ArticlesApiController.class.php
```diff
@@ -1048,6 +1056,86 @@ public function getAsJson() {
...
+			$sectionsToGet = json_decode( urldecode( $sectionsToGet ) );
+		}
+		// convert "all" or ["all"] to array of numbers
+		if ( $sectionsToGet === 'all' || in_array( 'all', $sectionsToGet ) ) {
```
rogatty: Why do we need `sections=all`? Wouldn't it be simpler to just omit this parameter when we want all sections? This parameter has inconsistency right now - it's a string or an array of numbers converted to string.

includes/wikia/api/ArticlesApiController.class.php
```diff
@@ -1048,6 +1056,86 @@ public function getAsJson() {
...
+	 * @param integer $section
+	 * @return string
+	 * @todo Supposedly getSection works with HTML, but I could only get it to work with wikitext
+	 */
+	private function getArticleSection( $articleAsWikiText, $title, $section ) {
+		/** @var Parser $wgParser */
+		global $wgParser;
+
+		$wikitext = $wgParser->getSection( $articleAsWikiText, $section );
+		$html = $wgParser->parse( $wikitext, $title, $wgParser->mOptions )->mText;
```
rogatty: Does it run all the hooks that we need? For example replacing media with placeholders and putting data in the media property?

SebastianMarzjan:
A general idea - would it make sense to start refactoring this API into smaller resources and have fetching in sections as a separate method instead of extending the general one?

includes/wikia/api/ArticlesApiController.class.php
kenkouot: Hi @aquilax

includes/wikia/api/tests/ArticlesApiControllerTest.php
kenkouot: Hi @aquilax

includes/wikia/api/ArticlesApiController.class.php
lizlux: Going to change it to accept only strings - `1,2,3` or `all`. We can change it it further iterations.

includes/wikia/api/ArticlesApiController.class.php
lizlux: It's running the hooks but there's indexing issues with the media array :(. Thanks for bringing this up. I'll fix it.

lizlux:
@SebastianMarzjan yes I think that's a good idea, but will probably be after we test this MVP/proof of concept. Thanks!

kenkouot:
@SebastianMarzjan Good idea. I wonder if this is the right time to start actually just gathering requirements for a real article service. I'd much rather build something that meets our updated standards. We've always considered this API an interim solution, so we have to consider how much we want to keep investing in something that was intended to be replaced.

lizlux:
I've simplified the solution quite a bit by just splitting the article content on H2 tags. I thought it would be nice to use MW's `getSection()` method but it turned out to be too much of a PITA.

includes/wikia/api/ArticlesApiController.class.php
```diff
@@ -1048,6 +1058,53 @@ public function getAsJson() {
 		$this->setResponseData( $result, '', self::SIMPLE_JSON_VARNISH_CACHE_EXPIRATION );
 	}
 
+	/**
+	 * Get all top level sections as an array by splitting on h2 tags
+	 * @param $content
+	 * @return array
+	 */
+	private function splitArticleIntoSections( $content ) {
+		$pattern = '/(\<h2[^\>]*>)/';
```
kenkouot: I feel like this leaves room for too many edge cases? Anyway we can make this stricter? How about h2s with id attributes?

lizlux: Yeah I thought about that too, so I can make it stricter. All section header `<h2>` tags have a `section` attribute, so I'll check for that too.

lizlux: I added the section attribute check. 

It's interesting to note that if you hard-code an `h2` tag in the article source, MW doesn't quite know what to do with it. See http://lizlux.liz.wikia-dev.com/wiki/Short_with_headers. The section is included in the TOC, but there's no edit link. However there is a `section` attribute added to the `h2` tag. Going to `?action=edit&section=X` does not count that section, and will either be empty or show the following section. 

For our purposes, I think this the current approach is fine. It will consider the offending `h2` to be a separate section but it's so edge-casey and the downside is so little that I wouldn't worry about it.

kenkouot:
LGTM

includes/wikia/api/ArticlesApiController.class.php
```diff
@@ -1048,6 +1058,53 @@ public function getAsJson() {
 		$this->setResponseData( $result, '', self::SIMPLE_JSON_VARNISH_CACHE_EXPIRATION );
 	}
 
+	/**
+	 * Get all top level sections as an array by splitting on h2 tags
+	 * @param $content
+	 * @return array
+	 */
+	private function splitArticleIntoSections( $content ) {
```
rogatty: IMO this method could use some unit testing.

lizlux: Good call. I'll add at test.

lizlux: @rogatty https://github.com/Wikia/app/pull/7876

PR #7804, opened 2015-07-09
PLATFORM-1241 - Added new UA tracker for logged-in users.

harnash:
New UA tracker (100% sampled) which is used for all logged-in users (based on UA user-id). Nothing interesting in here. It is just as main/special tracker with all the Custom Dimensions the same.


wladekb:
lgtm

PR #7809, opened 2015-07-09
SOC-1018 Add dedicated subject keys that do not include links

garthwebb:
https://wikia-inc.atlassian.net/browse/SOC-1018


jsutterfield:
LGTM

PR #7810, opened 2015-07-09
Ce 2243

MariuszCz:


lgarczewski:
LGTM. Please had off for testing.

adamkarminski:
@MariuszCz We discussed the visual design at the yesterday sync meeting and the decision is to move the video to the right side of a page. Please implement the change before merging. Thanks!

PR #7811, opened 2015-07-09
SOC-788 Make sure text field always contains most recent comment

mklucsarits:
The mini editor used for article comments pulls its content from a hidden textarea, which is supposed to contain the latest content of a comment text. This change makes sure that textarea is always filled with the most recent edited version of the comment.

https://wikia-inc.atlassian.net/browse/SOC-788

@jsutterfield @garthwebb


jsutterfield:
LGTM

garthwebb:
LGTM

PR #7815, opened 2015-07-10
CE-2298 Ignore redirects and docs pages on non-portable insights

lgarczewski:


extensions/wikia/Insights/specials/UnconvertedInfoboxesPage.class.php
```diff
@@ -128,6 +128,18 @@ public function reallyDoQuery( $limit = false, $offset = false ) {
...
 	public static function isTitleWithNonportableInfobox( $titleText, $contentText ) {
+		// ignore docs pages
+		if ( strpos( $titleText, '/doc' ) ) {
+			return false;
+		}
+
+		$title = Title::newFromText( $titleText, NS_TEMPLATE );
+
+		// ingore redirects
+		if ( $title->isRedirect() ) {
```
Grunny: I'd suggest excluding this in the query (in `reallyDoQuery`) so we don't even iterate over redirects (i.e. `page_is_redirect = 0`).

Grunny:
I think it would be good to add a quick test case to the unit tests for the `/doc` subpage exclusion, i.e. add another item like [this one](https://github.com/Wikia/app/blob/c93bd55fa2b/extensions/wikia/Insights/tests/UnconvertedInfoboxesPageTest.php#L25-L29) that would normally return true but with `/doc` at the end of the title.

extensions/wikia/Insights/specials/UnconvertedInfoboxesPage.class.php
```diff
@@ -128,6 +129,11 @@ public function reallyDoQuery( $limit = false, $offset = false ) {
 	 * @return bool
 	 */
 	public static function isTitleWithNonportableInfobox( $titleText, $contentText ) {
+		// ignore docs pages
+		if ( strpos( $titleText, '/doc' ) ) {
```
Grunny: Just a minor edge-case consideration, but this will match `/doc` anywhere in the sting, not just at the end, so Template:Infoboc/doctor. I think it would be better to have an equivalent of `$title->isSubpage() && $title->getSubpageText() === 'doc'` to be precise about the actual goal.

lgarczewski: Yep, good point, done.

Grunny:
LGTM.

lgarczewski:
Tests are failing because of LyricFind -- unrelated.

PR #7818, opened 2015-07-10
CE-2299 Do not display a fix message for nonportable infoboxes

adamkarminski:

In this PR I introduce a way to modify a behavior of our LoopNotification system. For now, it only includes a parameter that control displaying the "Let's fix it!" message, but can be expanded if needed.

### DIFF
--- a/extensions/wikia/Insights/InsightsController.class.php
+++ b/extensions/wikia/Insights/InsightsController.class.php
@@ -54,7 +54,7 @@ class InsightsController extends WikiaSpecialPageController {
 		 * - getData() - returning all the helping data
 		 * - getTemplate() - returning an overriding template
 		 */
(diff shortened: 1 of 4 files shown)


Grunny:
LGTM.

extensions/wikia/Insights/InsightsController.class.php
```diff
@@ -161,8 +171,8 @@ private function getCongratulationsNotificationParams() {
 	 * Get params for notification template shown when user fix one issue from the insights list
 	 *
 	 * @param String $subpage Insights subpage
-	 * @param String $articleName current article name
-	 * @param InsightsModel $model
+	 * @param $next
```
Grunny: These should have types added.

PR #7831, opened 2015-07-11
[SERVICES-481] Helios confirmation email language cleanup 

ArturKlajnerok:
With new emails language is set based on user preference so there's no point to pass it as a parameter.

/cc @Wikia/services-team

### DIFF
--- a/extensions/wikia/Helios/HelperController.class.php
+++ b/extensions/wikia/Helios/HelperController.class.php
@@ -118,7 +118,7 @@ class HelperController extends \WikiaController
 			return;
 		}
 
-		$mailStatus = $user->sendConfirmationMail( false, 'ConfirmationMail' );
+		$mailStatus = $user->sendConfirmationMail();
 
 		if ( ! $mailStatus->isGood() ) {
 			$this->response->setVal( 'message', 'could not send an email message' );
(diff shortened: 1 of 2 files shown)


extensions/wikia/Helios/HelperController.class.php
```diff
@@ -120,28 +118,7 @@ public function sendConfirmationEmail()
...
-				return $user->sendConfirmationMail(
-					false,
-					'ConfirmationMail',
-					'usersignup-confirmation-email',
-					true,
-					$mailTemplate,
-					$langCode
-				);
-			});
+		$mailStatus = $user->sendConfirmationMail( false, 'ConfirmationMail' );
```
dianafa: Would `sendConfirmationMail( false )` be also good here?
Why do we want to use the "confirmemail_body" instead of "usersignup-confirmation-email_body" message?

ArturKlajnerok: @dianafa In this case `sendConfirmationMail()` is best minimal usage with `ConfirmationMail` as default in : https://github.com/Wikia/app/blob/dev/includes/User.php#L3869 , other params don't matter as execution of https://github.com/Wikia/app/blob/dev/includes/User.php#L3909 depends only on passed `$mailtype` and is returning before any other param would be used.

nmonterroso:
LGTM

ArturKlajnerok:
The failing test are not connected with changes in this  PR.

PR #7842, opened 2015-07-14
(MAIN-4934,MAIN-4935) Make getExamples methods public

Grunny:
Make getExamples methods public to match Wikia core change.

Fixes the following fatal errors:

```
PHP Fatal error: Access level to SFAutocompleteAPI::getExamples()
must be public (as in class ApiBase) in
/extensions/SemanticForms/includes/SF_AutocompleteAPI.php on line 15

PHP Fatal error: Access level to SFAutoeditAPI::getExamples()
must be public (as in class ApiBase) in
/extensions/SemanticForms/includes/SF_AutoeditAPI.php on line 15
```

/cc @macbre


harnash:
LGTM

PR #7847, opened 2015-07-14
add SetUserPreferences to User.php and use it in Preferences

sqreek:
Remove of $wgPreferencesUseService will be in another pull request

//cc @Wikia/services-team

### DIFF
--- a/includes/User.php
+++ b/includes/User.php
@@ -363,7 +363,7 @@ class User {
 			}
 
 			if (isset($data[self::CACHE_PREFERENCES_KEY])) {
-				 $this->userPreferences()->setPreferences($this->mId, $data[self::CACHE_PREFERENCES_KEY]);
+				 $this->userPreferences()->setPreferencesInCache($this->mId, $data[self::CACHE_PREFERENCES_KEY]);
 			}
 		}
 		return true;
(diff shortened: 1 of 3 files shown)


includes/User.php
```diff
@@ -2674,6 +2693,18 @@ public static function localToGlobalPropertyName($property, $cityId = null, $sep
 		return sprintf("%s%s%s", $property, $sep, $cityId);
 	}
 
+	private function sanitizePropertyArray( $array ) {
+		if ( !is_array( $array ) ) {
+			return [ ];
+		}
+
+		foreach ( $array as $key => $value ) {
+			$array[ $key ] = $this->sanitizeProperty( $value );
```
harnash: couldn't we use ```array_map``` here?

harnash:
LGTM

includes/User.php
```diff
@@ -2551,6 +2554,22 @@ public function setGlobalPreference($preference, $value) {
 	}
 
 	/**
+	 * @param array $preferences preferenceName->Value
+	 */
+	public function setGlobalPreferences( $preferences ) {
+		global $wgPreferencesUseService;
+		if ($wgPreferencesUseService) {
+			$this->sanitizePropertyArray($preferences);
+			$this->userPreferences()->setPreferences($this->mId, $preferences);
```
nmonterroso: I guess the naming of the function `setPreferences` is misleading. It doesn't actually make a service call to set a user's preferences. It's probably better named as `loadPreferencesFromCache`, since that's what it's used for in [User.php](https://github.com/Wikia/app/blob/dev/includes/User.php#L363). 

Can we change the name of the `UserPreferences->setPreferences` function and introduce a new function called `setMultiple` that takes the `$userId` and `$preferences` array as input?

nmonterroso: Also maybe the logic from `User->setGlobalPreference` should be incorporated into `User->setGlobalPreferences` and `setGlobalPreference` should call `setGlobalPreferences`, since there is some additional logic/hooks that need to take place

lib/Wikia/src/Service/User/Preferences/UserPreferences.php
```diff
@@ -68,16 +68,29 @@ public function get($userId, $pref, $default = null, $ignoreHidden = false) {
...
 
-		$default = $this->getFromDefault($pref);
-		if ($val == null && isset($default)) {
-			$val = $default;
+	/**
+	 * @param int $userId
+	 * @param array $prefs
+	 */
+
+	public function setMultiply( $userId, $prefs ) {
```
nmonterroso: Do you mean `setMultiple` here? Also whitespace between phpdoc and function signature

includes/User.php
```diff
@@ -2524,29 +2527,31 @@ public function setLocalPreference($preference, $value, $cityId = null, $sep = '
...
-			// Clear cached skin/theme, so the new one displays immediately in Special:Preferences
-			switch ($preference) {
-				case 'skin':
-					unset($this->mSkin);
-					break;
-				case 'theme':
-					unset($this->mTheme);
-					break;
+			$this->sanitizePropertyArray( $preferences );
+			$this->userPreferences()->setPreferencesInCache( $this->mId, $preferences );
```
nmonterroso: Should this call `->set($this->mId, $preferences)`?

sqreek: good catch :+1:

PR #7848, opened 2015-07-14
CONCF Fix navigations for 1024px

vforge:
Fixes for Global Navigation width and Local Navigation's ribbon for 1024px

/cc @bognix @Warkot

### DIFF
--- a/extensions/wikia/GlobalNavigation/styles/GlobalNavigation.scss
+++ b/extensions/wikia/GlobalNavigation/styles/GlobalNavigation.scss
@@ -101,7 +101,7 @@ $start-a-wikia-logo-size: 30px;
 		}
 	}
 
-	@media #{$breakpoint-small-only} {
+	@media #{$breakpoint-below-small-plus} {
 		.global-navigation .global-navigation-container {
 			width: $content-width-small - $scroll-bar-width;
 
@@ -122,6 +122,12 @@ $start-a-wikia-logo-size: 30px;
 					}
 				}
 			}
+		}
+	}
+
+	@media #{$breakpoint-small-plus} {
+		.global-navigation .global-navigation-container {
+			width: $content-width-small-plus;
 		}
 	}
 
--- a/extensions/wikia/GlobalNavigation/styles/GlobalNavigation.scss
+++ b/extensions/wikia/GlobalNavigation/styles/GlobalNavigation.scss
@@ -5,8 +5,6 @@
 @import 'GlobalNavigationColors';
 @import 'GlobalNavigationVariables';
 
-$start-a-wikia-logo-size: 30px;
-
 .global-navigation {
 	background-color: $global-wikia-background-color;
 	box-shadow: 0 0 0 2px rgba(0, 0, 0, .3);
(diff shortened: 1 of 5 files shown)


skins/oasis/css/core/breakpoints-layout.scss
```diff
@@ -282,7 +282,7 @@
 			}
 
 			> nav.WikiNav {
-				@include width-calc('100% - 250px');
+				@include width-calc('100% - 265px');
```
Warkot: :sparkles: Why the :sparkles: magic :sparkles: numbers? :sparkles:

extensions/wikia/GlobalNavigation/styles/GlobalNavigation.scss
```diff
@@ -101,7 +101,7 @@ $start-a-wikia-logo-size: 30px;
 		}
 	}
 
-	@media #{$breakpoint-small-only} {
+	@media #{$breakpoint-below-small-plus} {
```
bognix: why not moving it to breakpoints-layout.scss so all breakpoint related code is the same place. Next time similar issue happens we know where to look for.

skins/oasis/css/core/breakpoints-layout.scss
vforge: It's previous value minus scrollbar width.

skins/oasis/css/core/breakpoints-layout.scss
```diff
@@ -294,7 +294,7 @@
 
 				.navbackground {
 					left: 10px;
-					width: 734px; // it's 100% of content minus wordmark
+					width: 713px;
```
vforge: Same as previous comment

extensions/wikia/GlobalNavigation/styles/GlobalNavigation.scss
vforge: I'm all against it - `breakpoints-layout.scss` is only for site-wide layout changes (like font and margins) - Here we have one extension that looks different in dependence from breakpoints - ultimately all code from `breakpoint-*` files should be moved to proper extensions.

skins/oasis/css/core/breakpoints-layout.scss
rogatty: Please add this as a comment.

skins/oasis/css/core/breakpoints-layout.scss
rogatty: Add a comment to the code, thanks.

extensions/wikia/GlobalNavigation/styles/GlobalNavigation.scss
bognix: On the other hand - when we change from breakpoints to any other layout we would need to grep through all extensions to look for this media queries. Having all code related with breakpoints inside breakpoints makes sense. Maybe instead of putting it inside breakpoints-layout.scss we can create file like breakpoints-globalnavigation.scss inside global nav extension. Thanks to this when changing code related to breakpoints we know we need to look inside breakpoints-* files not foo-*.scss

vforge: :+1:

bognix:
LGTM

tomasznapieralski:
LGTM

Warkot:
LGTM

PR #7850, opened 2015-07-14
remove mOptions references from ouside of User.php

sqreek:
/cc @Wikia/services-team

### DIFF
--- a/includes/User.php
+++ b/includes/User.php
@@ -2551,6 +2551,13 @@ class User {
 	}
 
 	/**
+	 * @param String $preference set to default
+	 * for now its support only old aproach to handle preferences (get/set Option)
+	 */
+	public function removeGlobalPreference($preference){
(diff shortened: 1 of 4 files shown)


includes/User.php
```diff
@@ -2551,6 +2551,13 @@ public function setGlobalPreference($preference, $value) {
 	}
 
 	/**
+	 * @param String $preference set to default
+	 * for now its support only old aproach to handle preferences (get/set Option)
+	 */
+	public function removeGlobalPreference($preference){
+		unset( $this->mOptions[ $preference ] );
```
nmonterroso: can we add a TODO here that says the user-preference service has a task for this: https://wikia-inc.atlassian.net/browse/SERVICES-483

nmonterroso:
Other than minor comment LGTM - as a side note I also debugged `Special:Preferences` and I never hit the `validate()` function calls

PR #7852, opened 2015-07-14
Log sampling

nmonterroso:
@ljagiello @frankfarmer @Wikia/services-team 
Sampling log calls at 10%, since this volume is too much for ES


ArturKlajnerok:
Lgtm :+1:

PR #7853, opened 2015-07-14
Services 552

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-522

Account for 404s when getting preferences from the user preference service, which is possible when asking for a user that doesn't exist.


ArturKlajnerok:
Lgtm :+1:

PR #7854, opened 2015-07-15
(CE-2327) Backport wfRandomString for use in SemanticForms

Grunny:
SemanticForms claims to be compatible with MW 1.19 but is using some
methods from MW 1.20. Backport wfRandomString to make the editing
form work.


lgarczewski:
LGTM.

PR #7855, opened 2015-07-15
INT-133: Fix issue by using mb_pathinfo instead of pathinfo

inez:
PHP pathinfo does not work wiht multibyte strings - it's a known issue that is solved in PHP6. In the meantime I decided to use mb_pathinfo which I found here http://php.net/manual/en/function.pathinfo.php#107461 and for which implementation looks pretty solid.

### DIFF
--- a/includes/wikia/GlobalFunctions.php
+++ b/includes/wikia/GlobalFunctions.php
@@ -1618,10 +1618,18 @@ function wfGetUniqueArrayCI( array $arr ) {
  * Like pathinfo but with support for multibyte - copied from http://php.net/manual/en/function.pathinfo.php#107461
  */
 function mb_pathinfo($filepath) {
-	preg_match('%^(.*?)[\\\\/]*(([^/\\\\]*?)(\.([^\.\\\\/]+?)|))[\\\\/\.]*$%im',$filepath,$m);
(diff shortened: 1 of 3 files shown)


inez:
@Grunny @mari-mcmurtrie @owend

inez:
Before you say: I know, it would be awesome to have units for it.

includes/wikia/GlobalFunctions.php
```diff
@@ -1613,3 +1613,15 @@ function wfGetUniqueArrayCI( array $arr ) {
 	$unique = array_intersect_key( $arr, array_unique( $lower ) );
 	return array_filter( $unique );
 }
+
+/**
+ * Like pathinfo but with support for multibyte - copied from http://php.net/manual/en/function.pathinfo.php#107461
+ */
+function mb_pathinfo($filepath) {
```
Grunny: I know this is copied from elsewhere, but I still think it should use our coding conventions (spacey, always use braces on blocks, etc.).

Grunny:
Other than the coding convention comment, LGTM. So, can be merged once unit tested and any extra manually testing is done.

inez:
@Grunny Ok. This should be ready for review again.

Grunny:
LGTM.

PR #7858, opened 2015-07-15
CE-2241 Basic Flags HQ front-end

adamkarminski:
This PR includes only the basic front-end elements. FlagsEditForm will be included as a separate module. You can use SpecialFlags.js as the main controller of scripts execution.


lukaszkonieczny:
lgtm

PR #7859, opened 2015-07-15
JsonFormatTest: run wikitext parsing with memcache disabled

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-1337

Parser cache is suspected to cause random assertion failures. Introduce `WikiaBaseTest::memCacheDisabledSection` helper method.

@SebastianMarzjan / @lgarczewski


SebastianMarzjan:
It.Should.Work
:+1:

PR #7862, opened 2015-07-15
CONCF-872 | Curated Content Cleanup

vforge:
https://wikia-inc.atlassian.net/browse/CONCF-872

/cc @Wikia/x-wing

### DIFF
--- a/extensions/wikia/CuratedContent/CuratedContentController.class.php
+++ b/extensions/wikia/CuratedContent/CuratedContentController.class.php
@@ -185,11 +185,12 @@ class CuratedContentController extends WikiaController {
 	}
 
 	public function getList() {
+		global $wgWikiaCuratedContent;
+
 		wfProfileIn( __METHOD__ );
 
 		$this->response->setFormat( WikiaResponse::FORMAT_JSON );
-
-		$content = $this->wg->WikiaCuratedContent;
+		$content = $wgWikiaCuratedContent;
 		if ( empty( $content ) ) {
 			$this->getCategories();
 		} else {
@@ -464,11 +465,11 @@ class CuratedContentController extends WikiaController {
 		foreach ( $curatedContent as $curatedContentModule => $items ) {
 			foreach ( $items as $item ) {
 				if ( $item['type'] == 'category' || $curatedContentModule == 'featured' ) {
-					if ( strlen( $item['label'] ) > CuratedContentHelper::VALIDATE_LABEL_MAX_LENGTH ) {
+					if ( strlen( $item['label'] ) > CuratedContentValidator::LABEL_MAX_LENGTH ) {
 						$tooLongTitleCount++;
 					}
 				} else {
-					if ( strlen( $item['title'] ) > CuratedContentHelper::VALIDATE_LABEL_MAX_LENGTH ) {
+					if ( strlen( $item['title'] ) > CuratedContentValidator::LABEL_MAX_LENGTH ) {
 						$tooLongTitleCount++;
 					}
 				}
--- a/extensions/wikia/CuratedContent/CuratedContentController.class.php
+++ b/extensions/wikia/CuratedContent/CuratedContentController.class.php
@@ -320,12 +320,12 @@ class CuratedContentController extends WikiaController {
 	 */
 	private function setSectionItemsResponse( $sectionName, $ret ) {
 		foreach ( $ret as &$value ) {
-			list( $image_id, $image_url ) = CuratedContentHelper::findImage(
+			list( $imageId, $imageUrl ) = CuratedContentHelper::findImageIdAndUrl(
 				$value['image_id'],
 				$value['article_id']
 			);
-			$value['image_id'] = $image_id;
-			$value['image_url'] = $image_url;
+			$value['image_id'] = $imageId;
+			$value['image_url'] = $imageUrl;
 		}
 		$this->response->setVal( $sectionName, $ret );
 	}
@@ -359,7 +359,7 @@ class CuratedContentController extends WikiaController {
 					$ret[] = [
 						'title' => $item['title'],
 						'image_id' => $imageId,
-						'image_url' => CuratedContentHelper::findImage( $imageId )[1]
+						'image_url' => CuratedContentHelper::findImageUrl( $imageId )
 					];
 				}
 
@@ -374,15 +374,15 @@ class CuratedContentController extends WikiaController {
 
 	function getJsonItem( $titleName, $ns, $pageId ) {
 		$title = Title::makeTitle( $ns, $titleName );
-		list( $image_id, $image_url ) = CuratedContentHelper::findImage( 0, $pageId );
+		list( $imageId, $imageUrl ) = CuratedContentHelper::findImageIdAndUrl( null, $pageId );
 
 		return [
 			'title' => $ns . ':' . $title->getFullText(),
 			'label' => $title->getFullText(),
-			'image_id' => $image_id,
+			'image_id' => $imageId,
 			'article_id' => $pageId,
 			'type' => 'category',
-			'image_url' => $image_url
+			'image_url' => $imageUrl
 		];
 	}
 
--- a/extensions/wikia/CuratedContent/CuratedContentController.class.php
+++ b/extensions/wikia/CuratedContent/CuratedContentController.class.php
@@ -319,7 +319,7 @@ class CuratedContentController extends WikiaController {
 	 */
 	private function setSectionItemsResponse( $sectionName, $ret ) {
 		foreach ( $ret as &$value ) {
-			list( $image_id, $image_url ) = CuratedContentSpecialController::findImageIfNotSet(
+			list( $image_id, $image_url ) = CuratedContentHelper::findImageIfNotSet(
 				$value['image_id'],
 				$value['article_id']
 			);
@@ -358,7 +358,7 @@ class CuratedContentController extends WikiaController {
 					$ret[] = [
 						'title' => $item['title'],
 						'image_id' => $imageId,
-						'image_url' => CuratedContentSpecialController::findImageIfNotSet( $imageId )[1]
+						'image_url' => CuratedContentHelper::findImageIfNotSet( $imageId )[1]
 					];
 				}
 
@@ -373,7 +373,7 @@ class CuratedContentController extends WikiaController {
 
 	function getJsonItem( $titleName, $ns, $pageId ) {
 		$title = Title::makeTitle( $ns, $titleName );
-		list( $image_id, $image_url ) = CuratedContentSpecialController::findImageIfNotSet( 0, $pageId );
+		list( $image_id, $image_url ) = CuratedContentHelper::findImageIfNotSet( 0, $pageId );
 
 		return [
 			'title' => $ns . ':' . $title->getFullText(),
@@ -464,11 +464,11 @@ class CuratedContentController extends WikiaController {
 		foreach ( $curatedContent as $curatedContentModule => $items ) {
 			foreach ( $items as $item ) {
 				if ( $item['type'] == 'category' || $curatedContentModule == 'featured' ) {
-					if ( strlen( $item['label'] ) > CuratedContentSpecialController::LABEL_MAX_LENGTH ) {
+					if ( strlen( $item['label'] ) > CuratedContentHelper::VALIDATE_LABEL_MAX_LENGTH ) {
 						$tooLongTitleCount++;
 					}
 				} else {
-					if ( strlen( $item['title'] ) > CuratedContentSpecialController::LABEL_MAX_LENGTH ) {
+					if ( strlen( $item['title'] ) > CuratedContentHelper::VALIDATE_LABEL_MAX_LENGTH ) {
 						$tooLongTitleCount++;
 					}
 				}
--- a/extensions/wikia/CuratedContent/CuratedContentController.class.php
+++ b/extensions/wikia/CuratedContent/CuratedContentController.class.php
@@ -190,16 +190,16 @@ class CuratedContentController extends WikiaController {
 		wfProfileIn( __METHOD__ );
 
 		$this->response->setFormat( WikiaResponse::FORMAT_JSON );
-		$content = $wgWikiaCuratedContent;
-		if ( empty( $content ) ) {
+
+		if ( empty( $wgWikiaCuratedContent ) ) {
 			$this->getCategories();
 		} else {
 			$section = $this->request->getVal( 'section' );
 			if ( empty( $section ) ) {
-				$this->setSectionsInResponse( $content );
-				$this->setFeaturedContentInResponse( $content );
+				$this->setSectionsInResponse( $wgWikiaCuratedContent );
+				$this->setFeaturedContentInResponse( $wgWikiaCuratedContent );
 			} else {
-				$this->setSectionItemsInResponse( $content, $section );
+				$this->setSectionItemsInResponse( $wgWikiaCuratedContent, $section );
 			}
 
 			$this->response->setCacheValidity( WikiaResponse::CACHE_STANDARD );
@@ -320,7 +320,7 @@ class CuratedContentController extends WikiaController {
 	 */
 	private function setSectionItemsResponse( $sectionName, $ret ) {
 		foreach ( $ret as &$value ) {
-			list( $image_id, $image_url ) = CuratedContentHelper::findImageIfNotSet(
+			list( $image_id, $image_url ) = CuratedContentHelper::findImage(
 				$value['image_id'],
 				$value['article_id']
 			);
@@ -359,7 +359,7 @@ class CuratedContentController extends WikiaController {
 					$ret[] = [
 						'title' => $item['title'],
 						'image_id' => $imageId,
-						'image_url' => CuratedContentHelper::findImageIfNotSet( $imageId )[1]
+						'image_url' => CuratedContentHelper::findImage( $imageId )[1]
 					];
 				}
 
@@ -374,7 +374,7 @@ class CuratedContentController extends WikiaController {
 
 	function getJsonItem( $titleName, $ns, $pageId ) {
 		$title = Title::makeTitle( $ns, $titleName );
-		list( $image_id, $image_url ) = CuratedContentHelper::findImageIfNotSet( 0, $pageId );
+		list( $image_id, $image_url ) = CuratedContentHelper::findImage( 0, $pageId );
 
 		return [
 			'title' => $ns . ':' . $title->getFullText(),
@@ -503,12 +503,10 @@ class CuratedContentController extends WikiaController {
 	 * @return bool
 	 */
 	static function onCuratedContentSave() {
-		global $wgServer;
-
-		$content = F::app()->wg->WikiaCuratedContent;
+		global $wgServer, $wgWikiaCuratedContent;
 
 		( new SquidUpdate( array_unique( array_reduce(
-			$content,
+			$wgWikiaCuratedContent,
 			function ( $urls, $item ) use ( $wgServer ) {
 				if ( $item['title'] !== '' && empty( $item['featured'] ) ) {
 					// Purge section URLs using urlencode() (standard for MediaWiki), which uses implements RFC 1738
--- a/extensions/wikia/CuratedContent/CuratedContentHelper.class.php
+++ b/extensions/wikia/CuratedContent/CuratedContentHelper.class.php
@@ -34,7 +34,7 @@ class CuratedContentHelper {
 
 	private function fillItemInfo( &$item ) {
 		$title = Title::newFromText( $item['title'] );
-		if ( $title instanceof Title && $title->exists() ) {
+		if ( !empty($title ) ) {
 			$articleId = $title->getArticleId();
 			$namespaceId = $title->getNamespace();
 			$imageId = (int)$item['image_id'];
--- a/extensions/wikia/CuratedContent/CuratedContentHelper.class.php
+++ b/extensions/wikia/CuratedContent/CuratedContentHelper.class.php
@@ -8,6 +8,7 @@ class CuratedContentHelper {
 	const STR_FILE = 'file';
 	const STR_CATEGORY = 'category';
 	const STR_EMPTY_CATEGORY = 'emptyCategory';
+	const STR_VIDEO = 'video';
 
 	public function processLogic( $sections ) {
 		$processedSections = [ ];
@@ -40,7 +41,7 @@ class CuratedContentHelper {
 				case self::STR_FILE:
 					$info = self::getVideoInfo( $title );
 					if ( !empty( $info ) ) {
-						$item['type'] = 'video';
+						$item['type'] = self::STR_VIDEO;
 						$item['video_info'] = $info;
 					}
 					break;
--- a/extensions/wikia/CuratedContent/CuratedContentHelper.class.php
+++ b/extensions/wikia/CuratedContent/CuratedContentHelper.class.php
@@ -18,7 +18,7 @@ class CuratedContentHelper {
 	}
 
 	private function processLogicForSection( $section ) {
-		$section['image_id'] = (int)$section['image_id'];
+		$section['image_id'] = (int)$section['image_id']; // fallback to 0 if it's not set in request
 
 		foreach ( $section['items'] as &$item ) {
 			$this->fillItemInfo( $item );
@@ -29,7 +29,7 @@ class CuratedContentHelper {
 
 	private function fillItemInfo( &$item ) {
 		$title = Title::newFromText( $item['title'] );
-		if ( !empty( $title ) ) {
+		if ( $title instanceof Title && $title->exists() ) {
 			$articleId = $title->getArticleId();
 			$namespaceId = $title->getNamespace();
 			$imageId = (int)$item['image_id'];
@@ -47,19 +47,21 @@ class CuratedContentHelper {
 
 				case self::STR_CATEGORY:
 					$category = Category::newFromTitle( $title );
-					if ( !empty( $category ) ) {
+					if ( $category instanceof Category && $category->getID() ) {
 						$count = $category->getPageCount();
 						if ( empty( $count ) ) {
 							$item['type'] = self::STR_EMPTY_CATEGORY;
 						}
+					} else {
+						$item['type'] = null;
 					}
 					break;
 			}
 
-			if ( $imageId === 0 ) {
+			if ( empty( $imageId ) ) {
 				$imageTitle = self::findFirstImageTitleFromArticle( $articleId );
 				if ( !empty( $imageTitle ) ) {
-					$articleId = $imageTitle->getArticleId();
+					$imageId = $imageTitle->getArticleId();
 				}
 			}
 			$item['article_id'] = $articleId;
@@ -68,13 +70,13 @@ class CuratedContentHelper {
 	}
 
 	public static function getIdFromCategoryName( $categoryName ) {
-		$cat = Title::newFromText( $categoryName, NS_CATEGORY );
+		$category = Title::newFromText( $categoryName, NS_CATEGORY );
 
-		return ($cat instanceof Title) ? $cat->getArticleID() : 0;
+		return ($category instanceof Title && $category->exists()) ? $category->getArticleID() : 0;
 	}
 
-	public static function getImageUrl( $id ) {
-		$thumbnail = (new ImageServing( [$id], 50, 50 ))->getImages( 1 );
+	public static function getImageUrl( $id, $imageSize = 50 ) {
+		$thumbnail = (new ImageServing( [ $id ], $imageSize, $imageSize ))->getImages( 1 );
 
 		return !empty( $thumbnail ) ? $thumbnail[$id][0]['url'] : '';
 	}
@@ -83,7 +85,7 @@ class CuratedContentHelper {
 		$mediaService = new MediaQueryService();
 		$mediaInfo = $mediaService->getMediaData( $title );
 		if ( !empty( $mediaInfo ) ) {
-			if ( $mediaInfo['type'] === 'video' ) {
+			if ( $mediaInfo['type'] === MediaQueryService::MEDIA_TYPE_VIDEO ) {
 				return [
 					'provider' => $mediaInfo['meta']['provider'],
 					'thumb_url' => $mediaInfo['thumbUrl'],
@@ -94,17 +96,17 @@ class CuratedContentHelper {
 		return null;
 	}
 
-	public static function findImageIfNotSet( $imageId, $articleId = 0 ) {
+	public static function findImage( $imageId, $articleId = 0 ) {
 		$url = '';
-
 		$imageTitle = null;
+
 		if ( empty( $imageId ) ) {
 			$imageId = null;
 			$imageTitle = self::findFirstImageTitleFromArticle( $articleId );
 		} else {
 			$imageTitle = Title::newFromID( $imageId );
 		}
-		if ( !empty( $imageTitle ) ) {
+		if ( $imageTitle instanceof Title && $imageTitle->exists() ) {
 			$url = self::getUrlFromImageTitle( $imageTitle );
 			$imageId = $imageTitle->getArticleId();
 		}
@@ -115,16 +117,16 @@ class CuratedContentHelper {
 	public static function findFirstImageTitleFromArticle( $articleId ) {
 		$imageTitle = null;
 		if ( !empty( $articleId ) ) {
-			$is = new ImageServing( [$articleId] );
-			$image = $is->getImages( 1 );
+			$imageServing = new ImageServing( [ $articleId ] );
+			$image = $imageServing->getImages( 1 );
 			if ( !empty( $image ) ) {
-				$image_title_name = $image[$articleId][0]['name'];
-				if ( !empty( $image_title_name ) ) {
-					$imageTitle = Title::newFromText( $image_title_name, NS_FILE );
+				$imageTitleName = $image[$articleId][0]['name'];
+				if ( !empty( $imageTitleName ) ) {
+					$imageTitle = Title::newFromText( $imageTitleName, NS_FILE );
 				}
 			}
 		}
-		return $imageTitle;
+		return ($imageTitle instanceof Title && $imageTitle->exists()) ? $imageTitle : null;
 	}
 
 	public static function getUrlFromImageTitle( $imageTitle ) {
--- a/extensions/wikia/CuratedContent/CuratedContentHelper.class.php
+++ b/extensions/wikia/CuratedContent/CuratedContentHelper.class.php
@@ -34,7 +34,9 @@ class CuratedContentHelper {
 
 	private function fillItemInfo( &$item ) {
 		$title = Title::newFromText( $item['title'] );
-		if ( !empty($title ) ) {
+		// We cannot check for $title->isEmpty() because it's empty for non-exisitng Articles.
+		// Categories without content are non-existing Articles (their article_id is 0)
+		if ( !empty( $title ) ) {
 			$articleId = $title->getArticleId();
 			$namespaceId = $title->getNamespace();
 			$imageId = (int)$item['image_id'];
--- a/extensions/wikia/CuratedContent/CuratedContentHelper.class.php
+++ b/extensions/wikia/CuratedContent/CuratedContentHelper.class.php
@@ -10,10 +10,12 @@ class CuratedContentHelper {
 	const STR_EMPTY_CATEGORY = 'emptyCategory';
 	const STR_VIDEO = 'video';
 
-	public function processLogic( $sections ) {
+	public function processSections( $sections ) {
 		$processedSections = [ ];
-		foreach ( $sections as $section ) {
-			$processedSections[ ] = $this->processLogicForSection( $section );
+		if ( !empty( $sections ) && is_array( $sections ) ) {
+			foreach ( $sections as $section ) {
+				$processedSections[ ] = $this->processLogicForSection( $section );
+			}
 		}
 		return $processedSections;
 	}
@@ -21,8 +23,10 @@ class CuratedContentHelper {
 	private function processLogicForSection( $section ) {
 		$section['image_id'] = (int)$section['image_id']; // fallback to 0 if it's not set in request
 
-		foreach ( $section['items'] as &$item ) {
-			$this->fillItemInfo( $item );
+		if ( !empty( $section['items'] ) && is_array( $section['items'] ) ) {
+			foreach ( $section['items'] as &$item ) {
+				$this->fillItemInfo( $item );
+			}
 		}
 
 		return $section;
@@ -70,12 +74,6 @@ class CuratedContentHelper {
 		}
 	}
 
-	public static function getIdFromCategoryName( $categoryName ) {
-		$category = Title::newFromText( $categoryName, NS_CATEGORY );
-
-		return ($category instanceof Title && $category->exists()) ? $category->getArticleID() : 0;
-	}
-
 	public static function getImageUrl( $id, $imageSize = 50 ) {
 		$thumbnail = (new ImageServing( [ $id ], $imageSize, $imageSize ))->getImages( 1 );
 
@@ -97,7 +95,13 @@ class CuratedContentHelper {
 		return null;
 	}
 
-	public static function findImage( $imageId, $articleId = 0 ) {
+	public static function findImageUrl( $imageId ) {
+		list( $_imageId, $imageUrl ) = self::findImageIdAndUrl( $imageId );
+
+		return $imageUrl;
+	}
+
+	public static function findImageIdAndUrl( $imageId, $articleId = 0 ) {
 		$url = '';
 		$imageTitle = null;
 
--- a/extensions/wikia/CuratedContent/CuratedContentHelper.class.php
+++ b/extensions/wikia/CuratedContent/CuratedContentHelper.class.php
@@ -10,9 +10,9 @@ class CuratedContentHelper {
 	const STR_EMPTY_CATEGORY = 'emptyCategory';
 
 	public function processLogic( $sections ) {
-		$processedSections = [];
+		$processedSections = [ ];
 		foreach ( $sections as $section ) {
-			$processedSections[] = $this->processLogicForSection( $section );
+			$processedSections[ ] = $this->processLogicForSection( $section );
 		}
 		return $processedSections;
 	}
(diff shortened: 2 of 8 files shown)


extensions/wikia/CuratedContent/CuratedContentHelper.class.php
```diff
@@ -0,0 +1,160 @@
...
+	private function fillItemInfo( &$item ) {
+		$title = Title::newFromText( $item['title'] );
+		if ( !empty( $title ) ) {
+			$articleId = $title->getArticleId();
+			$namespaceId = $title->getNamespace();
+			$imageId = (int)$item['image_id'];
+
+			$item['type'] = self::getTypeFromNamespaceId( $namespaceId );
+
+			switch ( $item['type'] ) {
```
tomasznapieralski, line 46: What if there is no case for that switch? Maybe default case should be there and return some exception or sth else?

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
```diff
@@ -0,0 +1,160 @@
+<?php
+/**
+ * CuratedContent Helper
+ */
+class CuratedContentHelper {
```
tomasznapieralski, line 5: Maybe abstract should be added, because all public methods are static

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
vforge, line 46: We're setting that value one line above; and it's handling only some special cases, we don't want it to handle anything else - we could use double `if`s, but I think `switch` is cleaner.

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
vforge, line 5: To the declaration? Isn't abstract class not supposed to have any methods implemented?

extensions/wikia/CuratedContent/CuratedContentValidator.class.php
```diff
@@ -0,0 +1,131 @@
...
+	private function validateSection ( $section ) {
+		if ( empty( $section['title'] ) ) {
+			if ($this->hadEmptyLabel) {
+				$this->error( $section, 'duplicatedLabel' );
+			} else {
+				$this->hadEmptyLabel = true;
+			}
+		}
+
+		if ( strlen( $section['title'] ) > self::LABEL_MAX_LENGTH ) {
```
tomasznapieralski, line 76: Use >= and change LABEL_MAX_LENGTH to be accurate limit

extensions/wikia/CuratedContent/CuratedContentValidator.class.php
```diff
@@ -0,0 +1,131 @@
...
+			$this->error( $section, 'tooLongLabel' );
+		}
+
+		if ( empty( $section['featured'] ) && $section['title'] !== '' ) {
```
tomasznapieralski: I think `$section['title'] !== ''` should be wrapped in brackets

extensions/wikia/CuratedContent/CuratedContentValidator.class.php
vforge, line 76: What would reasoning behind this change? `LABEL_MAX_LENGTH` is max allowed length right now.

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
rogatty, line 5: > Abstract classes cannot be instantiated, but they can be subclassed.

This shouldn't be an abstract class.

extensions/wikia/CuratedContent/js/CuratedContentManagmentTool.js
```diff
@@ -380,16 +384,38 @@ require(['wikia.window', 'jquery', 'wikia.nirvana', 'wikia.tracker', 'JSMessages
...
-											return false;
+									$sections.each(function () {
+										if (this.value === errTitle) {
+											var $itemWithError;
+
+											switch (errReason) {
+												case 'missingImage':
+													$itemWithError = $(this).parent().find('.image');
+													break;
+												case 'duplicatedLabel':
```
tomasznapieralski: what is this? empty case? remove it or add break

extensions/wikia/CuratedContent/js/CuratedContentManagmentTool.js
```diff
@@ -380,16 +384,38 @@ require(['wikia.window', 'jquery', 'wikia.nirvana', 'wikia.tracker', 'JSMessages
...
+												return false;
+											}
 										}
 										return true;
 									});
 								});
 
 								$save.addClass('err');
-								$save.attr('disabled', true);
+								//$save.attr('disabled', true);
```
tomasznapieralski: if you need to keep that in comment, please write why, otherwise remove it

extensions/wikia/CuratedContent/js/CuratedContentManagmentTool.js
vforge: It's multiple case code block - see http://phpswitch.com/

extensions/wikia/CuratedContent/CuratedContentValidator.class.php
tomasznapieralski: Yes but it means that title has length between 0 and 39 not 40

tomasznapieralski, line 76: Ah sorry i thought about that it is less than your value, not greater than

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
```diff
@@ -0,0 +1,160 @@
...
+	const STR_ARTICLE = 'article';
+	const STR_BLOG = 'blog';
+	const STR_FILE = 'file';
+	const STR_CATEGORY = 'category';
+	const STR_EMPTY_CATEGORY = 'emptyCategory';
+
+	public function processLogic( $sections ) {
```
rogatty: I'm not sure about this method name. It's not clear what it does - processing some logic isn't really self-explanatory.

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
```diff
@@ -0,0 +1,160 @@
...
+		$processedSections = [ ];
+		foreach ( $sections as $section ) {
+			$processedSections[ ] = $this->processLogicForSection( $section );
+		}
+		return $processedSections;
+	}
+
+	private function processLogicForSection( $section ) {
+		$section['image_id'] = (int)$section['image_id'];
```
rogatty: `$section` comes straight from the request, right? What happens here if there is no `$section['image_id']` set?
If the result is 0 then let's add it as a comment.

extensions/wikia/CuratedContent/CuratedContentValidator.class.php
Warkot: I disagree, It's readable and works as intended. http://php.net/manual/en/language.operators.precedence.php

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
```diff
@@ -0,0 +1,160 @@
...
+		foreach ( $section['items'] as &$item ) {
+			$this->fillItemInfo( $item );
+		}
+
+		return $section;
+	}
+
+	private function fillItemInfo( &$item ) {
+		$title = Title::newFromText( $item['title'] );
+		if ( !empty( $title ) ) {
```
rogatty, line 39: Should we check `$title->exists()`?

extensions/wikia/CuratedContent/CuratedContentValidator.class.php
```diff
@@ -0,0 +1,131 @@
...
+				$this->validateImage($item);
+			}
+		} else {
+			if ( empty( $section['featured'] ) ) {
+				$this->error( $section, 'itemsMissing');
+			}
+		}
+
+		if ( strlen( $section['title'] ) ) {
+			$this->titles[ ] = $section['title'];
```
Warkot: `$this->titles[ ]` is an array 'index selector' (new automatic index), so no spaces needed. Use spaces only inside array constructors e.g. `$this->titles = [ ]` or `$this->titles = [ 'foo', 'bar' ]`.

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
```diff
@@ -0,0 +1,160 @@
...
+					$info = self::getVideoInfo( $title );
+					if ( !empty( $info ) ) {
+						$item['type'] = 'video';
+						$item['video_info'] = $info;
+					}
+					break;
+
+				case self::STR_CATEGORY:
+					$category = Category::newFromTitle( $title );
+					if ( !empty( $category ) ) {
```
rogatty: What if it's empty? Should it still be a category type?

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
```diff
@@ -0,0 +1,160 @@
...
+							$item['type'] = self::STR_EMPTY_CATEGORY;
+						}
+					}
+					break;
+			}
+
+			if ( $imageId === 0 ) {
+				$imageTitle = self::findFirstImageTitleFromArticle( $articleId );
+				if ( !empty( $imageTitle ) ) {
+					$articleId = $imageTitle->getArticleId();
```
rogatty: Shouldn't this be `$imageId`?

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
```diff
@@ -0,0 +1,160 @@
...
+				}
+			}
+			$item['article_id'] = $articleId;
+			$item['image_id'] = $imageId;
+		}
+	}
+
+	public static function getIdFromCategoryName( $categoryName ) {
+		$cat = Title::newFromText( $categoryName, NS_CATEGORY );
```
rogatty: `$category =`

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
```diff
@@ -0,0 +1,160 @@
...
+
+		return ($cat instanceof Title) ? $cat->getArticleID() : 0;
```
rogatty: Should we check `$cat->exists()`?

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	public static function getImageUrl( $id ) {
+		$thumbnail = (new ImageServing( [$id], 50, 50 ))->getImages( 1 );
```
rogatty: How about moving this numbers to method arguments? 50 can still be a default.

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
```diff
@@ -0,0 +1,160 @@
...
+
+		return !empty( $thumbnail ) ? $thumbnail[$id][0]['url'] : '';
+	}
+
+	private function getVideoInfo( $title ) {
+		$mediaService = new MediaQueryService();
+		$mediaInfo = $mediaService->getMediaData( $title );
+		if ( !empty( $mediaInfo ) ) {
+			if ( $mediaInfo['type'] === 'video' ) {
```
rogatty: `MediaQueryService::MEDIA_TYPE_VIDEO`

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
```diff
@@ -0,0 +1,160 @@
...
+		$url = '';
+
+		$imageTitle = null;
+		if ( empty( $imageId ) ) {
+			$imageId = null;
+			$imageTitle = self::findFirstImageTitleFromArticle( $articleId );
+		} else {
+			$imageTitle = Title::newFromID( $imageId );
+		}
+		if ( !empty( $imageTitle ) ) {
```
rogatty: Should we check `$imageTitle->exists()`?

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
```diff
@@ -0,0 +1,160 @@
...
+					'provider' => $mediaInfo['meta']['provider'],
+					'thumb_url' => $mediaInfo['thumbUrl'],
+					'videoId' => $mediaInfo['meta']['videoId']
+				];
+			}
+		}
+		return null;
+	}
+
+	public static function findImageIfNotSet( $imageId, $articleId = 0 ) {
```
rogatty: I'm not sure about this method name because it does _things_ even if image is set.

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
```diff
@@ -0,0 +1,160 @@
...
+			$imageId = $imageTitle->getArticleId();
+		}
+
+		return [ $imageId, $url ];
+	}
+
+	public static function findFirstImageTitleFromArticle( $articleId ) {
+		$imageTitle = null;
+		if ( !empty( $articleId ) ) {
+			$is = new ImageServing( [$articleId] );
```
rogatty: `$imageServing`

extensions/wikia/CuratedContent/CuratedContentController.class.php
```diff
@@ -185,11 +185,12 @@ public function renderFullPage() {
...
 
 	public function getList() {
+		global $wgWikiaCuratedContent;
+
 		wfProfileIn( __METHOD__ );
 
 		$this->response->setFormat( WikiaResponse::FORMAT_JSON );
-
-		$content = $this->wg->WikiaCuratedContent;
+		$content = $wgWikiaCuratedContent;
```
hakubo: can't we just use `$wgWikiaCuratedContent`?

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
```diff
@@ -0,0 +1,160 @@
...
+			$image = $is->getImages( 1 );
+			if ( !empty( $image ) ) {
+				$image_title_name = $image[$articleId][0]['name'];
```
rogatty: `$imageTitleName`

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
```diff
@@ -0,0 +1,160 @@
...
+				if ( !empty( $image_title_name ) ) {
+					$imageTitle = Title::newFromText( $image_title_name, NS_FILE );
```
rogatty: Should we check `$imageTitle->exists()` before returning?

extensions/wikia/CuratedContent/CuratedContentController.class.php
```diff
@@ -319,7 +320,7 @@ private function getFeaturedSection( $content ) {
 	 */
 	private function setSectionItemsResponse( $sectionName, $ret ) {
 		foreach ( $ret as &$value ) {
-			list( $image_id, $image_url ) = CuratedContentSpecialController::findImageIfNotSet(
+			list( $image_id, $image_url ) = CuratedContentHelper::findImageIfNotSet(
```
hakubo: Not a fan of Helper functions as they tend to become PutSomeRandomMethodsClass.php
'Memory is cheap' can we divide it into some logical classes?
CuratedContentImage
CuratedContentUrl

maybe even those can live somewhere else and be helpful to other teams?

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
```diff
@@ -0,0 +1,160 @@
+<?php
+/**
+ * CuratedContent Helper
+ */
+class CuratedContentHelper {
+	const STR_ARTICLE = 'article';
```
hakubo, line 6: hmm, hmmmmm don't we have those somewhere already? hmmm

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
hakubo: still better than `doThings` :)

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
```diff
@@ -0,0 +1,160 @@
...
+				$image_title_name = $image[$articleId][0]['name'];
+				if ( !empty( $image_title_name ) ) {
+					$imageTitle = Title::newFromText( $image_title_name, NS_FILE );
+				}
+			}
+		}
+		return $imageTitle;
+	}
+
+	public static function getUrlFromImageTitle( $imageTitle ) {
```
rogatty, line 139: `$imageTitle` has to be of `Title` type, right?

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
hakubo, line 39: also we should also check if it is Title class I think
`$title instanceof Title && $title->exists()`

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
hakubo, line 6: I think there is something like: `wgNamespaceIds`

extensions/wikia/CuratedContent/CuratedContentSpecialController.class.php
```diff
@@ -108,14 +87,15 @@ public function index() {
...
+				$featuredSection = $this->sendSelfRequest( 'featuredSection' );
 			}
 			// prepend featured section
 			$list = $featuredSection . $list;
 
 			$this->response->setVal( 'list', $list );
 		} else {
-			$this->response->setVal( 'featured', $this->sendSelfRequest( self::FEATURED_SECTION_TEMPLATE ) );
+			$this->response->setVal( 'featured', $this->sendSelfRequest( 'featuredSection' ) );
+			$this->response->setVal( 'featured', $this->sendSelfRequest( 'featuredSection' ) );
```
rogatty: This shouldn't be doubled, right?

extensions/wikia/CuratedContent/CuratedContentSpecialController.class.php
```diff
@@ -84,7 +64,7 @@ public function index() {
 		$this->response->setVal( 'item_placeholder', wfMessage( 'wikiacuratedcontent-content-item' ) );
 		$this->response->setVal( 'name_placeholder', wfMessage( 'wikiacuratedcontent-content-name' ) );
 
-		$itemTemplate = $this->sendSelfRequest( self::ITEM_FUNCTION_NAME )->toString();
+		$itemTemplate = $this->sendSelfRequest( 'item' )->toString();
```
rogatty, line 68: Why? I don't hate, just ask ;) I thougt that we try to keep strings like this one in constants.

extensions/wikia/CuratedContent/CuratedContentSpecialController.class.php
```diff
@@ -155,27 +135,23 @@ public function section() {
...
 
 
-		if ( $id == 0 && $item != '' ) {
-			$cat = Title::newFromText( $item, NS_CATEGORY );
-
-			if ( $cat instanceof Title ) {
-				$id = $cat->getArticleID();
-			}
+		if ( empty( $id ) && $item != '' ) {
+			$id = CuratedContentHelper::getIdFromCategoryName( $item );
```
rogatty: Can you add a comment why this is happening?

extensions/wikia/CuratedContent/CuratedContentSpecialController.class.php
```diff
@@ -155,27 +135,23 @@ public function section() {
 	 * referred by ITEM_FUNCTION_NAME
```
rogatty: This isn't true anymore.

extensions/wikia/CuratedContent/CuratedContentSpecialController.class.php
```diff
@@ -187,65 +163,39 @@ public function save() {
...
+		} else {
+			$status = WikiFactory::setVarByName( 'wgWikiaCuratedContent', $this->wg->CityId, $sections );
+			$this->response->setVal( 'status', $status );
 
-			if ( !empty( $thumbnail ) ) {
-				$url = $thumbnail[ $id ][ 0 ][ 'url' ];
+			if ( $status) {
+				wfRunHooks('CuratedContentSave', [$sections]);
 			}
+			return true;
```
rogatty:
```
if (...) {
  ...
  return true;
} else {
  ...
  return true;
}
```

You know what I mean?

extensions/wikia/CuratedContent/CuratedContentSpecialController.class.php
```diff
@@ -187,65 +163,39 @@ public function save() {
...
+			$this->response->setVal( 'error', $errors );
+			return true;
+		} else {
+			$status = WikiFactory::setVarByName( 'wgWikiaCuratedContent', $this->wg->CityId, $sections );
+			$this->response->setVal( 'status', $status );
 
-			if ( !empty( $thumbnail ) ) {
-				$url = $thumbnail[ $id ][ 0 ][ 'url' ];
+			if ( $status) {
+				wfRunHooks('CuratedContentSave', [$sections]);
```
rogatty: Missing spaces here and the line above! :skull:

extensions/wikia/CuratedContent/CuratedContentSpecialController.class.php
```diff
@@ -255,271 +205,24 @@ private function buildSection( $section ) {
...
-		$imageTitle = null;
-		if ( !empty( $articleId ) ) {
-			$is = new ImageServing( [ $articleId ] );
-			$image = $is->getImages( 1 );
-			if ( !empty( $image ) ) {
-				$image_title_name = $image[ $articleId ][ 0 ][ 'name' ];
-				if ( !empty( $image_title_name ) ) {
-					$imageTitle = Title::newFromText( $image_title_name, NS_FILE );
-				}
+			if ( !empty( $img ) && $img instanceof Title ) {
```
rogatty: `&& $img->exists()`?

extensions/wikia/CuratedContent/CuratedContentSpecialController.class.php
```diff
@@ -255,271 +205,24 @@ private function buildSection( $section ) {
...
-	}
 
-	public static function getUrlFromImageTitle( $imageTitle ) {
-		$imageUrl = null;
-		if ( !empty( $imageTitle ) ) {
-			$imageFile = wfFindFile( $imageTitle );
-			if ( !empty( $imageFile ) ) {
-				$imageUrl = $imageFile->getUrl();
-			}
+		if ( $id != 0 ) {
```
rogatty: `!==`?

extensions/wikia/CuratedContent/CuratedContentValidator.class.php
```diff
@@ -0,0 +1,131 @@
...
+	private $titles;
+	private $hadEmptyLabel;
+
+	public function __construct( $data ) {
+		$this->errors = [ ];
+		$this->titles = [ ];
+		$this->hadEmptyLabel = false;
+
+		// validate sections
+		foreach( $data as $section ) {
```
rogatty: Missing space :see_no_evil:

extensions/wikia/CuratedContent/CuratedContentValidator.class.php
```diff
@@ -0,0 +1,131 @@
...
+			if ( !empty( $section['featured'] ) ) {
+				$this->validateFeaturedSection( $section );
+			} else {
+				$this->validateSection( $section );
+			}
+		}
+		// also check section for duplicate title
+		foreach( array_count_values( $this->titles ) as $title => $count ) {
```
rogatty: We put space between block name and bracket. No space between function name and bracket, though.

extensions/wikia/CuratedContent/CuratedContentValidator.class.php
```diff
@@ -0,0 +1,131 @@
...
+			$this->errors[ ] = [ 'title' => $itemWithTitle['title'], 'reason' => $errorString ];
+		}
+	}
+
+	public function getErrors() {
+		return $this->errors;
+	}
+
+	private function validateFeaturedSection( $section ) {
+		foreach ( $section['items'] as $item ) {
```
rogatty: Should we check if `$section['items']` exists at all and is an array?

extensions/wikia/CuratedContent/CuratedContentValidator.class.php
```diff
@@ -0,0 +1,131 @@
...
+		}
+	}
+
+	private function validateImage( $sectionOrItem ) {
+		if ( empty( $sectionOrItem['image_id'] ) ) {
+			$this->error( $sectionOrItem, 'imageMissing' );
+		}
+	}
+
+	private function validateSection ( $section ) {
```
rogatty: I like this space but it's incosistent with the rest of file.

extensions/wikia/CuratedContent/CuratedContentValidator.class.php
```diff
@@ -0,0 +1,131 @@
...
+		if ( empty( $section['title'] ) ) {
+			if ($this->hadEmptyLabel) {
```
rogatty: I like this lack of spaces but it's inconsistent with our conventions.

rogatty: Also let's add a comment that this is "optional block" or section or sth and that there can be only one.

extensions/wikia/CuratedContent/CuratedContentValidator.class.php
rogatty: Let's add a comment with description why we need to validate image when this condition is true.

extensions/wikia/CuratedContent/CuratedContentValidator.class.php
```diff
@@ -0,0 +1,131 @@
...
+			$this->validateImage( $section );
+		}
+
+		if ( !empty( $section['items'] ) && is_array( $section['items'] ) ) {
+			foreach ($section['items'] as $item) {
+				$this->validateCategoryItem($item);
+				$this->validateImage($item);
+			}
+		} else {
+			if ( empty( $section['featured'] ) ) {
```
rogatty, line 92: `} else if ( ... ) {`?
Let's also add comments what is going on here.

extensions/wikia/CuratedContent/CuratedContentValidator.class.php
```diff
@@ -0,0 +1,131 @@
...
+
+		if ( empty( $item['label'] ) ) {
+			$this->error( $item, 'emptyLabel' );
+		}
+
+		if ( strlen( $item['label'] ) > self::LABEL_MAX_LENGTH ) {
+			$this->error( $item, 'tooLongLabel' );
+		}
+
+		if ( $item['type'] == null ) {
```
rogatty: `===`?

extensions/wikia/CuratedContent/CuratedContentValidator.class.php
```diff
@@ -0,0 +1,131 @@
...
+			$this->error( $item, 'notSupportedType' );
+		}
+
+		if ( $item['type'] === 'video' ) {
```
rogatty: Shouldn't we use some constant here?

extensions/wikia/CuratedContent/CuratedContentValidator.class.php
```diff
@@ -0,0 +1,131 @@
...
+
+		if ( self::needsArticleId( $item['type'] ) && empty( $item['article_id'] ) ) {
+			$this->error( $item, 'articleNotFound' );
+		}
+
+		$this->titles[ ] = $item['title'];
+	}
+
+	private static function needsArticleId( $type ) {
+		return $type != CuratedContentHelper::STR_CATEGORY;
```
rogatty: `!==`?

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
vforge, line 6: Those are name of the keys inside `$wgWikiaCuratedContent`.

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
vforge, line 139: Nope - it can be passed as string - `File` class normalize it to title before doing anything else.

extensions/wikia/CuratedContent/CuratedContentSpecialController.class.php
vforge, line 68: It's only function name - we usually just use raw string for that - it's a common pattern in our code.

extensions/wikia/CuratedContent/js/CuratedContentManagmentTool.js
```diff
@@ -354,18 +354,22 @@ require(['wikia.window', 'jquery', 'wikia.nirvana', 'wikia.tracker', 'JSMessages
...
 
 							if (data.error) {
-								var items = $form.find('.item-input, .section-input');
+								var $items = $form.find('.item-input'),
+									$sections = $form.find('.section-input');
 
-								$.each(data.error, function() {
-									var errTitle = this.title,
-										errReason = this.reason,
+								[].forEach.call(data.error, function(err) {
```
rogatty: Would it be difficult to extract this function? There is sooo much indentation here.

extensions/wikia/CuratedContent/js/CuratedContentManagmentTool.js
```diff
@@ -380,16 +384,38 @@ require(['wikia.window', 'jquery', 'wikia.nirvana', 'wikia.tracker', 'JSMessages
...
+													$itemWithError = $(this).parent().find('.image');
+													break;
+												case 'duplicatedLabel':
+												case 'tooLongLabel':
+													$itemWithError = $(this);
+													break;
+											}
+											if ($itemWithError) {
+												$itemWithError.addError(reasonMessage);
+												return false;
```
rogatty: Does it mean that we break from `$sections.each` loop and display only the first error?

extensions/wikia/CuratedContent/js/CuratedContentManagmentTool.js
rogatty: +1

extensions/wikia/CuratedContent/CuratedContentSpecialController.class.php
```diff
@@ -124,58 +103,52 @@ public function index() {
...
 
-		if ( $id == 0 && $item != '' ) {
-			$cat = Title::newFromText( $item, NS_CATEGORY );
-
-			if ( $cat instanceof Title ) {
-				$id = $cat->getArticleID();
-			}
+		if ( empty( $imageId ) && !empty( $item ) ) {
+			// image_id is 0 - that means it should be taken from page
+			$imageId = CuratedContentHelper::getIdFromCategoryName( $item );
```
rogatty: But `getIdFromCategoryName` returns category id, not the image one, right?

extensions/wikia/CuratedContent/js/CuratedContentManagmentTool.js
vforge: It's this loop is iterated for each error, so no.

extensions/wikia/CuratedContent/js/CuratedContentManagmentTool.js
vforge: Done.

rogatty:
LGTM

tomasznapieralski:
It will be nice to add one more line not to leave 999, but LGTM

vforge:
@tomasznapieralski I've also noticed it ;)

extensions/wikia/CuratedContent/CuratedContentController.class.php
```diff
@@ -358,7 +359,7 @@ function ( $ret, $item ) {
 					$ret[] = [
 						'title' => $item['title'],
 						'image_id' => $imageId,
-						'image_url' => CuratedContentSpecialController::findImageIfNotSet( $imageId )[1]
+						'image_url' => CuratedContentHelper::findImage( $imageId )[1]
```
bognix: maybe it's better to return associative array? It would give some idea what are you returning from this function.
Also I'm thinking if it's not worth to create two functions, find image by id (which would return URL) and find image by article id and it would return imageId,Url array?

extensions/wikia/CuratedContent/CuratedContentController.class.php
```diff
@@ -373,7 +374,7 @@ function ( $ret, $item ) {
 
 	function getJsonItem( $titleName, $ns, $pageId ) {
 		$title = Title::makeTitle( $ns, $titleName );
-		list( $image_id, $image_url ) = CuratedContentSpecialController::findImageIfNotSet( 0, $pageId );
+		list( $image_id, $image_url ) = CuratedContentHelper::findImage( 0, $pageId );
```
bognix: should it be 0 or null? I think null would show better that you are not passing imageId. Or maybe you want to find image with id 0 on article with pageId, then it's fine.

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
```diff
@@ -0,0 +1,163 @@
...
+ */
+class CuratedContentHelper {
+	const STR_ARTICLE = 'article';
+	const STR_BLOG = 'blog';
+	const STR_FILE = 'file';
+	const STR_CATEGORY = 'category';
+	const STR_EMPTY_CATEGORY = 'emptyCategory';
+	const STR_VIDEO = 'video';
+
+	public function processLogic( $sections ) {
```
bognix: could you rename this function to processSections or add some doc? ProcessLogic is not very descriptive.

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
```diff
@@ -0,0 +1,163 @@
...
+					} else {
+						$item['type'] = null;
+					}
+					break;
+			}
+
+			if ( empty( $imageId ) ) {
+				$imageTitle = self::findFirstImageTitleFromArticle( $articleId );
+				if ( !empty( $imageTitle ) ) {
+					$imageId = $imageTitle->getArticleId();
```
bognix, line 71: can getArticleId return null/0/false? If so what should we do? Do we need to handle this kind of situation?

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
```diff
@@ -0,0 +1,163 @@
...
+				}
+			}
+			$item['article_id'] = $articleId;
+			$item['image_id'] = $imageId;
+		}
+	}
+
+	public static function getIdFromCategoryName( $categoryName ) {
```
bognix: Is this function used in our code base? I searched the code and found nothing.

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
```diff
@@ -0,0 +1,163 @@
...
+				return self::STR_BLOG;
+				break;
+			case NS_CATEGORY:
+				return self::STR_CATEGORY;
+				break;
+			case NS_FILE:
+				return self::STR_FILE;
+				break;
+			default:
+				return null;
```
bognix, line 165: should the default be null or maybe should we fallback to article?

extensions/wikia/CuratedContent/CuratedContentSpecialController.class.php
```diff
@@ -124,58 +103,52 @@ public function index() {
...
+		$imageId = $this->request->getVal( 'image_id', 0 );
 
-		$this->response->setVal( 'value', $this->request->getVal( 'value' ), '' );
-		$this->response->setVal( 'image_id', $id );
-		$this->response->setVal( 'image_url', $this->getImage( $id ) );
-		if ( $id != 0 ) {
+		$this->response->setVal( 'value', $this->request->getVal( 'value' , '' ) );
+		$this->response->setVal( 'image_id', $imageId );
+		$this->response->setVal( 'image_url', CuratedContentHelper::getImageUrl( $imageId ) );
+		if ( empty( $imageId ) ) {
```
bognix, line 125: you sometimes check if imageId != 0 (https://github.com/Wikia/app/pull/7862/files#diff-514f443fe83103e42561ed6c76b92b66R113) and sometimes if it is not empty. Can we unify it?

extensions/wikia/CuratedContent/CuratedContentSpecialController.class.php
```diff
@@ -187,65 +160,38 @@ public function save() {
...
 
-		$sections = $this->request->getArray( 'sections' );
-		list( $sections, $err ) = $this->processSaveLogic( $sections );
+		$helper = new CuratedContentHelper();
 
-		if ( !empty( $err ) ) {
-			$this->response->setVal( 'error', $err );
-			return true;
-		}
+		$sections = $helper->processLogic( $this->request->getArray( 'sections' ) );
```
bognix: does getArray fallback to empty array or null when nothing is passed in the request? If it's null you can add fallback inside getArray method. This way we won't throw PHP errors about supplying null to foreach

extensions/wikia/CuratedContent/CuratedContentSpecialController.class.php
```diff
@@ -187,65 +160,38 @@ public function save() {
...
-			}
-		}
 
-		if ( $id != 0 ) {
-			$is = new ImageServing( [ $id ], 50, 50 );
-			$thumbnail = $is->getImages( 1 );
+		if ( !empty( $errors ) ) {
+			$this->response->setVal( 'error', $errors );
+		} else {
+			$status = WikiFactory::setVarByName( 'wgWikiaCuratedContent', $this->wg->CityId, $sections );
```
bognix: sometimes you use convention global $wgVar, here you use $this->wg->CityId, could you unify?

extensions/wikia/CuratedContent/CuratedContentSpecialController.class.php
```diff
@@ -187,65 +160,38 @@ public function save() {
 		}
```
bognix, line 160: line 157: you use $this->wg->User, could you do global $wgUser so it is unified with rest of the code

extensions/wikia/CuratedContent/CuratedContentSpecialController.class.php
```diff
@@ -255,271 +201,24 @@ private function buildSection( $section ) {
...
-			$imageTitle = self::findFirstImageTitleFromArticle( $articleId );
-		} else {
-			$imageTitle = Title::newFromID( $imageId );
-		}
-		if ( !empty( $imageTitle ) ) {
-			$url = self::getUrlFromImageTitle( $imageTitle );
-			$imageId = $imageTitle->getArticleId();
-		}
+	public function getImage() {
+		$file = $this->request->getVal( 'file', '' );
```
bognix: maybe you should fallback to null not empty string?

extensions/wikia/CuratedContent/CuratedContentValidator.class.php
```diff
@@ -0,0 +1,136 @@
...
+		foreach ( array_count_values( $this->titles ) as $title => $count ) {
+			if ( $count > 1 ) {
+				$this->error( [ 'title' => $title ], 'duplicatedLabel' );
+			}
+		}
+	}
+
+	private function error( $itemWithTitle, $errorString ) {
+		if ( array_key_exists( 'title', $itemWithTitle ) && !empty( $errorString ) ) {
+			$this->errors[ ] = [ 'title' => $itemWithTitle['title'], 'reason' => $errorString ];
```
bognix: space inside is not needed. You already initiated this array

extensions/wikia/CuratedContent/CuratedContentValidator.class.php
```diff
@@ -0,0 +1,136 @@
...
+	private function validateImage( $sectionOrItem ) {
+		if ( empty( $sectionOrItem['image_id'] ) ) {
+			$this->error( $sectionOrItem, 'imageMissing' );
+		}
+	}
+
+	private function validateSection( $section ) {
+		// check for "optional" section - it has empty label, but there can be only ONE
+		if ( empty( $section['title'] ) ) {
+			if ( $this->hadEmptyLabel ) {
```
bognix: maybe instead of calling this field hadEmptyLabel you should name it hasOptionalSection because it's only how you use this filed (to check if optional is set and if it's only one).

extensions/wikia/CuratedContent/CuratedContentValidator.class.php
```diff
@@ -0,0 +1,136 @@
...
+				$this->error( $section, 'duplicatedLabel' );
+			} else {
+				$this->hadEmptyLabel = true;
+			}
+		}
+
+		if ( strlen( $section['title'] ) > self::LABEL_MAX_LENGTH ) {
+			$this->error( $section, 'tooLongLabel' );
```
bognix: how do you feel about moving error labels to consts?

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
vforge, line 71: It can return 0 - if that's the case we don't have image for article - we still want to save 0 as `image_id`, so validation will get it.

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
vforge, line 165: Default is null - validator checks if it's not null, so null equals wrong namespace.

extensions/wikia/CuratedContent/CuratedContentSpecialController.class.php
vforge, line 125: Missed that one, thanks!

extensions/wikia/CuratedContent/CuratedContentSpecialController.class.php
vforge: :+1:

extensions/wikia/CuratedContent/CuratedContentSpecialController.class.php
vforge, line 160: Done.

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
```diff
@@ -0,0 +1,163 @@
...
+		}
+
+		return [ $imageId, $url ];
+	}
+
+	public static function findFirstImageTitleFromArticle( $articleId ) {
+		$imageTitle = null;
+		if ( !empty( $articleId ) ) {
+			$imageServing = new ImageServing( [ $articleId ] );
+			$image = $imageServing->getImages( 1 );
```
Warkot, line 128: Is this method filtering out small images e.g. currency  and skill icons or icons in notices e.g. stub info and spoiler alert?

vforge, line 128: There _was_ filter for small images, so they won't be found - dunno current status.

bognix:
LGTM

rogatty:
LGTM

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
```diff
@@ -0,0 +1,167 @@
...
+				$this->fillItemInfo( $item );
+			}
+		}
+
+		return $section;
+	}
+
+	private function fillItemInfo( &$item ) {
+		$title = Title::newFromText( $item['title'] );
+		if ( !empty($title ) ) {
```
rogatty: Can you add a comment explaining why we allow titles that don't exist? Also space is missing.

PR #7864, opened 2015-07-15
updating PowerUserTrait to use getGlobalFlag

nmonterroso:
@Wikia/services-team 
Updating a `getBoolOption` callsite that we found on the Kibana dashboard


sqreek:
Is this will be always getGlobalFlag?
is cast for boolean is not needed?

`If (answers is yes, yes(not needed)){
  LGTM();
}`

nmonterroso:
Yeah, it's always `getGlobalFlag`, it checks `if (getGlobalPref($prefName))`, so it does a truthy type test, which if a preference is unset or empty, will work as expected.

PR #7870, opened 2015-07-16
CE-2331

adamkarminski:

This is an initial PR for a Flag create/edit form. For now, it covers displaying of the form for a flag's creation, collecting the data and sending it in an AJAX request.

My main goal is to consult the module's architecture on an early stage.

Please mind the _TODO_ notes :)

CC: @kamilkoterba @lukaszkonieczny

### DIFF
--- a/extensions/wikia/Flags/specials/scripts/FlagEditForm.js
+++ b/extensions/wikia/Flags/specials/scripts/FlagEditForm.js
@@ -118,6 +118,7 @@ define('ext.wikia.Flags.FlagEditForm',
 		}
 
 		function displayFormEdit(content) {
+			/* TODO - Finish this method */
 			modalConfig.vars.content = content;
 			modalConfig.vars.title = mw.message('flags-special-create-form-title-edit').escaped();
 
@@ -127,12 +128,10 @@ define('ext.wikia.Flags.FlagEditForm',
 		function displayModal() {
 			require(['wikia.ui.factory'], function (uiFactory) {
 				/* Initialize the modal component */
-				uiFactory.init(['modal']).then(createComponent);
+				uiFactory.init(['modal']).then(function(uiModal){
+					uiModal.createComponent(modalConfig, processInstance);
+				});
 			});
-		}
-
-		function createComponent(uiModal) {
-			uiModal.createComponent(modalConfig, processInstance);
 		}
 
 		function processInstance(modalInstance) {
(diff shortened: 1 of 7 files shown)


extensions/wikia/Flags/specials/scripts/FlagEditForm.js
```diff
@@ -0,0 +1,273 @@
...
+					buttons: getButtons(),
+					content: '' // content
+				}
+			};
+		}
+
+		function getButtons() {
+			return [
+				{
+					vars: {
```
kamilkoterba: I think these buttons miss i18n, right?
It can be done by adding value property to vars: {}
eg. value: mw.message('msg-name').escaped(),

extensions/wikia/Flags/specials/scripts/FlagEditForm.js
```diff
@@ -0,0 +1,266 @@
...
+				uiFactory.init(['modal']).then(function(uiModal){
+					uiModal.createComponent(modalConfig, processInstance);
+				});
+			});
+		}
+
+		function processInstance(modalInstance) {
+			modalInstance.show();
+			modalInstance.bind('done', saveCreateFlagForm);
+			$('.flags-special-form-params-new').on('click', addNewParameterInput);
```
kamilkoterba: I would advice to keep the scope 
$('#FlagEditForm .flags-special-form-params-new')
so we won't be surprized by duplicated actions in the future.

adamkarminski: Don't you think that the class name is long & unique enough? :)

kamilkoterba: Probably we can either simplify #FlagEditForm .form-params-new or leave this unique class.
I think both approaches are good here. Up to you.

extensions/wikia/Flags/specials/templates/SpecialFlags_createFlagForm.mustache
```diff
@@ -0,0 +1,73 @@
+<form class="flags-special-form">
+	<fieldset>
+		<table class="flags-special-form-table">
```
kamilkoterba: Please don't use tables for non tabular data just for defining layout.

kamilkoterba:
Despite comment with tables LGTM

kamilkoterba:
Please add sub tasks for resolving TODOs so we won't loose track of it

PR #7872, opened 2015-07-16
DAT-3035 article preview typography styles overwrite portable infobox font-size for labels

RafLeszczynski:
- reduce specificity of breakpoints typography css selectors in article preview


dianafa:
LGTM but with testing will be fun :D

PR #7874, opened 2015-07-16
PLATFORM-1163 namespace analytics

harnash:


extensions/wikia/AnalyticsEngine/js/universal_analytics.js
```diff
@@ -287,7 +287,8 @@
         ['set', 'dimension20', window.wgABPerformanceTest || 'not set'],        // Performance A/B testing
         ['set', 'dimension21', String(window.wgArticleId)],                     // ArticleId
         ['set', 'dimension23', window.wikiaIsPowerUserFrequent ? 'Yes' : 'No'], // IsPowerUser: Frequent
-        ['set', 'dimension24', window.wikiaIsPowerUserLifetime ? 'Yes' : 'No']  // IsPowerUser: Lifetime
+        ['set', 'dimension24', window.wikiaIsPowerUserLifetime ? 'Yes' : 'No'], // IsPowerUser: Lifetime
+        ['set', 'dimension25', window.wgNamespaceNumber]                        // Namespace Number
```
macbre: `wgNamespaceNumber` global is a number. Should we cast to it to a string (like we do for `wgArticleId`)?

harnash:
Since we do have problems with this variable being defined much later then the Analytics code is being fired I need to look into once more.

PR #7875, opened 2015-07-16
Platform-1163 namespace analytics

harnash:
I had to move definition of JS variable `$wgNamespaceNumber` before UA code is executed.

Added missing CD to ADS account.


macbre:
LGTM

PR #7876, opened 2015-07-16
Hg 754 adding another test and sections value

lizlux:
This is a follow-up PR to https://github.com/Wikia/app/pull/7796. It adds a test for `splitArticleIntoSections`, fixes an off-by-one error, and returns a string instead of an array of article sections.


rogatty:
> fixes an off-by-one error

Ha, unit tests!

LGTM

kenkouot:
LGTM2

PR #7879, opened 2015-07-16
[SERVICES-525] Making sure user is logged out when resetting password

ArturKlajnerok:
As checking temporary password is now externalized to helios we need to be sure that user is not logged in MW to properly reset password.

/cc @Wikia/services-team  @garthwebb


garthwebb:
LGTM! :+1:

PR #7886, opened 2015-07-17
MAIN-4977 Use the login type to make sure mediawiki does the right thing with wgUser

garthwebb:
This change attempts to mimic what Mediawiki used to do around password resets more closely.

https://wikia-inc.atlassian.net/browse/MAIN-4977


ArturKlajnerok:
Lgtm :+1:

PR #7891, opened 2015-07-20
CE-2330

adamkarminski:

This PR covers creation of a form that enables Flags editing to our users. It does not include an API method that updates DB with the sent data yet.

It also introduces fixes in communication with a user on Special:Flags.

CC: @kamilkoterba


extensions/wikia/Flags/Flags.i18n.php
```diff
@@ -28,7 +28,7 @@
 	'flags-special-create-form-save' => 'Save',
 	'flags-special-create-form-save-success' => 'The flag has been added!',
 	'flags-special-create-form-save-failure' => 'Unfortunately, an error happened. Can you try again?',
-	'flags-special-autoload-delete-confirm' => 'Deleting this flag will also remove it from all articles. Are you sure?',
+	'flags-special-autoload-delete-confirm' => 'Deleting the $1 flag will also remove it from all articles. Are you sure?',
```
kamilkoterba, line 31: Add parameter description in qqq

adamkarminski, line 31: All docs (including `qqq`) upcoming when the final shape of the code is achieved. Is it ok?

extensions/wikia/Flags/specials/scripts/FlagEditForm.js
```diff
@@ -13,7 +13,7 @@ define ('ext.wikia.Flags.FlagEditForm',
 			$.when(getFormResources()).done(function (formResources) {
 				setupForm(formResources);
 
-				if (prefillData == null || prefillData.values == null) {
+				if (prefillData == null) {
```
kamilkoterba, line 16: triple = per coding conventions

adamkarminski, line 16: I used `==` deliberately to check for both `undefined` and `null`.

kamilkoterba, line 16: So explicit is better than implicit.
Use undefined check in statement or write a comment explaining intention.

kamilkoterba:
Not related to these changes but probably you may change it?
https://github.com/Wikia/app/blob/CE-2330/extensions/wikia/Flags/specials/scripts/FlagEditForm.js#L11
this documentation doesn't reflect what it actually does.
General documentation for this extension would be also nice to meet definition of done.

kamilkoterba:
LGTM

extensions/wikia/Flags/Flags.i18n.php
kamilkoterba, line 31: np, hope we won't loose track of it. Maybe a subtask with a checklist of places that needs docs to lookup before dev merge?

PR #7894, opened 2015-07-20
Services-528 harnash

harnash:
Added support for User Properties white lists. Changes in wikia/config can be found on: https://github.com/Wikia/config/pull/1197

White lists are implemented in two contexts:
1. On Media Wiki core side (User.php) where we need to distinguish user preferences from other entities (options, flags, attributes) since User.saveOptions() clears all unnecessary entries in 'user_preferences' table
2. On PreferencePersistenceMySQL we do filter fields being saved by 'service mockup' to avoid overwriting entities managed by Media Wiki.

Since we are using regex'es there are some raw SQL statements added since not all API's allow using REGEXP statement (i.e. FluentSQL).

@Wikia/services-team

### DIFF
--- a/includes/User.php
+++ b/includes/User.php
@@ -4798,7 +4798,7 @@ class User {
 	 * @todo document
 	 */
 	protected function saveOptions() {
-		global $wgAllowPrefChange, $wgPreferencesUseService;
+		global $wgAllowPrefChange, $wgPreferencesUseService, $wgUserPreferenceWhiteList;
 
 		$extuser = ExternalUser::newFromUser( $this );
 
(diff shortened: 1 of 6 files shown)


includes/User.php
```diff
@@ -4841,7 +4841,29 @@ protected function saveOptions() {
 			}
 		}
 
-		$dbw->delete( 'user_properties', array( 'up_user' => $this->getId() ), __METHOD__ );
+		// kinda ghetto, but :(
+		if ($wgPreferencesUseService) {
+			$regex = implode ('|', $wgUserPreferenceWhiteList['regexes']);
```
nmonterroso: I don't think that `User.php` needs to know about the whitelist. Since `PreferencePersistenceMySQL` knows about the whitelist and filters based on that

``` php
            $preferenceNames = array_keys($this->userPreferences()->getPreferences($this->getId()));

            (new WikiaSQL())
                ->DELETE('user_properties')
                ->WHERE('up_user')->EQUAL_TO($this->getId())
                    ->AND_('up_property')->NOT_IN($preferenceNames)
                ->run($dbw);

            $insert_rows = array_reduce($insert_rows, function($result, $current) use ($preferenceNames) {
                if (!in_array($current['up_property'], $preferenceNames)) {
                    $result[] = $current;
                }

                return $result;
            }, []);
```

Should produce only whitelisted preference names

harnash: Well I agree on the concept but getPreferences() would return only preferences set for the user (at a given time) not the current list of preferences managed by user-preferences service. I would go into making getter for this in userPreferences so it will return the list of known preferences... Still there is a case where we should handle regex'es.

nmonterroso: In which case does this need to handle regular expressions? I think the goal of this piece of functionality is "delete entries in `user_properties` for this user that are not preferences", which can be achieved without the use of the whitelist, since the entries for a user are stored by the preference service. It also doesn't need to know about regular expressions IMO because it will do a straight forward `NOT IN (some-pref-12345, some-pref-23456)` for those properties.

harnash: Right. I forgot that it will be handled in the same request so there will be no race conditions (which could be problematic when saving user preference in different request/thread). Am I right? There will be no case when we try to save in thread user_pref1 and it will not be returned by getPreferences() in other thread, but it will be saved before delete on 'user_preference' and thus lost at the end?

`getPreferences() [thread1] -> savePreferences('user-pref1') [thread2] -> delete('user_preferences') [thread2]`

nmonterroso: That problem does exist, but it exists in the current, non-service implementation since a thread will forcibly delete all a user's rows in the `user_properties` table

ArturKlajnerok:
Lgtm :+1:

PR #7895, opened 2015-07-20
CE-2360 

adamkarminski:

This method binds an actual update of a flag type with its edit form on Special:Flags.
Includes a new API method and some JS tweaks.

CC: @kamilkoterba


kamilkoterba:
LGTM, sql looks fine as well.

PR #7896, opened 2015-07-21
CE-2323 Move Title creation from API to places of its usage

adamkarminski:

This is a permanent fix for [CE-2323](https://wikia-inc.atlassian.net/browse/CE-2323). I've removed Title object creation from API entirely. It is now created only when needed. I also use the `Linker` class to create Links since it deals with `Title === null` well by creating a red link.

CC: @lgarczewski @kamilkoterba


lgarczewski:
LGTM, @MariuszCz please test.

PR #7902, opened 2015-07-21
CE-2349 Validate the form before sending

adamkarminski:
With this PR we will check:
1. _flag_name_ for being empty or already used
2. _flag_view_ for being empty
3. _Names in flags_params_names_ for being empty

Any of the cases above will trigger a BannerNotification with an appropriate error message.

Additionally, if you try to save a flag without making any changes - you'll be warned that nothing was actually done.

CC: @lgarczewski @MariuszCz


lgarczewski:
LGTM.

MariuszCz:
If you try to edit and save a flag there is notification that "The name of the flag is already used. Please, choose another one."

MariuszCz:
And if you try delete all parameters in edit mode there is notification "Please enter an appropriate names for all parameters."

PR #7903, opened 2015-07-21
CE-2351 get flag parameters from template on type creation

lgarczewski:
@Wikia/community-engineering  when reviewing think about:
- is this the right place for this code?
- can we assume that we fetch from template always on type creation for now?
- how the hell do I verify if there are no params passed from API? the `$params == '{}'` hack is ugly!


extensions/wikia/Flags/FlagsParamsComparison.class.php
```diff
@@ -12,7 +12,7 @@ class FlagsParamsComparison {
 	 * @param $flagVariables
 	 * @return bool|mixed|null
 	 */
-	public function compareTemplateVariables( $title, $oldText, $newText, $flagVariables ) {
+	public function compareTemplateVariables( $title, $oldText, $newText, $flagVariables = array() ) {
```
adamkarminski: Short array notation `[]`

extensions/wikia/Flags/specials/SpecialFlagsController.class.php
```diff
@@ -40,6 +40,67 @@ public function index() {
 		$this->hasAdminPermissions = $this->wg->User->isAllowed( 'flags-administration' );
 	}
 
+	public function addFlagType() {
```
Grunny: This method and the `updateFlagType` method below will bypass CSRF protections in the API as the `FlagsApiController::processRequest` method checks if the request is internal, which it now will be due to using `sendRequest`. You can set it as not internal in the `sendRequest` call, but that probably wouldn't account for the POST requirement, so you'd need to duplicate CSRF protections here...

extensions/wikia/Flags/specials/SpecialFlagsController.class.php
```diff
@@ -40,6 +40,67 @@ public function index() {
...
+		$response = $this->sendRequest( 'FlagsApiController',
+			'updateFlagType',
+			$data
+		)->getData();
+
+		$this->response->setValues( $response );
+	}
+
+	/**
+	 * @param string name of template treated as view and source of params
```
Grunny: `@param string $template ...`

PR #7904, opened 2015-07-21
[MAIN-4983] Checking if helios is enabled before calling extension code.

ArturKlajnerok:
This is causing error when helios is disabled on a Wikia

@Wikia/services-team @garthwebb


garthwebb:
LGTM

PR #7905, opened 2015-07-21
check value of $wgSharedDB when selecting the database

nmonterroso:
@Wikia/services-team 
update the database selection in `PreferenceModule` to be consistent with what's in User.php [`loadOptions`](https://github.com/Wikia/app/blob/dev/includes/User.php#L4777) and [`saveOptions`](https://github.com/Wikia/app/blob/dev/includes/User.php#L4814)


ArturKlajnerok:
Lgtm :+1:

PR #7908, opened 2015-07-22
CONCF-806 | Additional data for getData method

vforge:
/cc @hakubo @bognix @rogatty

### DIFF
--- a/extensions/wikia/CuratedContent/CuratedContentController.class.php
+++ b/extensions/wikia/CuratedContent/CuratedContentController.class.php
@@ -440,6 +440,9 @@ class CuratedContentController extends WikiaController {
 
 		if ( !empty( $wgWikiaCuratedContent ) && is_array( $wgWikiaCuratedContent )  ) {
 			foreach ( $wgWikiaCuratedContent as $section ) {
+				// update information about node type
+				$section['node_type'] = 'section';
+
 				// rename $section['title'] to $section['label']
 				$section['label'] = $section['title'];
 				unset( $section['title'] );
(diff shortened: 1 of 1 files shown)


rogatty:
LGTM

rogatty:
LGTM

extensions/wikia/CuratedContent/CuratedContentController.class.php
```diff
@@ -440,14 +440,24 @@ public function getData( ) {
...
-				// sections
-				if ( !empty( $section['title'] ) && empty( $section['featured'] ) ) {
+				// update information about node type
+				$section['node_type'] = 'section';
+
+				// rename $section['title'] to $section['label']
+				$section['label'] = $section['title'];
+				unset( $section['title'] );
+
+				if ( !empty( $section['label'] ) && empty( $section['featured'] ) ) {
```
bognix, line 450: I think it doesn't solve our problem.
We wanted to have distinction for sections inside "curated" block so we don't need to do multiple ifs in Mercury.
Here you add node type to all sections (featured and optional) so we still need to add this if.

vforge, line 450: After (offline) talk we disagreed with that comment.

bognix:
LGTM

extensions/wikia/CuratedContent/CuratedContentController.class.php
```diff
@@ -440,14 +440,24 @@ public function getData( ) {
 
 		if ( !empty( $wgWikiaCuratedContent ) && is_array( $wgWikiaCuratedContent )  ) {
 			foreach ( $wgWikiaCuratedContent as $section ) {
-				// sections
-				if ( !empty( $section['title'] ) && empty( $section['featured'] ) ) {
+				// update information about node type
+				$section['node_type'] = 'section';
```
hakubo, line 444: hmm should we use camelCase?
since that is what we use on FE

rogatty, line 444: All the rest of properties use underscore as they come from `$wgWikiaCuratedContent` like that. Should we convert all of them? We will have to revert the process when saving.

PR #7910, opened 2015-07-22
CE-2366 Make the Go back to the list message not look like a call to …

adamkarminski:
…action.

I've also change the order of links to a more sane one.

CC: @lgarczewski


kamilkoterba:
LGTM

PR #7912, opened 2015-07-22
CE-2210 Display injected flags as a flex-box

adamkarminski:
Hey all!

Portable flags are usually displayed at the top of an article. However, you can modify this behavior with the _**FLAGS**_ magic word. If you put it in a content of an article, flags are injected in this specific place.

The problem was that sometimes the flags container was pushing other elements down on the page because of a `clear: both;` rule.

This PR differentiates flags displayed at the top from the ones injected in the content. The _inline_ ones are going to be displayed with the `display: flex;` rule which allows them to fit to the rest of a page nicely.

CC: @bkoval

### DIFF
--- a/extensions/wikia/Flags/scripts/FlagsModal.js
+++ b/extensions/wikia/Flags/scripts/FlagsModal.js
@@ -243,7 +243,11 @@ require(
 			$div = $(document.createElement('div'))
 				.addClass('flags-edit')
 				.html($a);
-		$('.portable-flags').prepend($div).append($div.clone(true));
+		if ($('.portable-flags').length !== 0) {
+			$('.portable-flags').prepend($div).append($div.clone(true));
+		} else if ($('.portable-flags-inline').length !== 0) {
+			$('.portable-flags-inline').prepend($div).append($div.clone(true));
+		}
 	}
 
 	// Run initialization method on DOM ready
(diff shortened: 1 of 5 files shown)


extensions/wikia/Flags/scripts/FlagsModal.js
```diff
@@ -243,7 +243,11 @@ require(
 			$div = $(document.createElement('div'))
 				.addClass('flags-edit')
 				.html($a);
-		$('.portable-flags').prepend($div).append($div.clone(true));
+		if ($('.portable-flags').length !== 0) {
```
bkoval: You could cache jQuery selector output to avoid querying DOM twice :)

extensions/wikia/Flags/styles/Flags.scss
```diff
@@ -15,6 +15,16 @@
 	margin: 5px 0;
 }
 
+.portable-flags-inline {
+	display: -ms-flexbox;
+	display: -webkit-flex;
+	display: flex;
+	-webkit-flex-direction: column;
+	-ms-flex-direction: column;
```
bkoval, line 22: We could swap lines 22 and 23 to be alphabetically correct ;)

bkoval:
Two details, except that LGTM :dart: :palm_tree: :santa:

PR #7914, opened 2015-07-22
CE-2158 Turn on preview for infobox templates with ACE editor enabled

adamkarminski:

This fixes the problem with previewing infobox templates. It turns out that the `showPagePreview` variable was set to `false` for them even though they were whitelisted in `EditPageLayoutHelper::isCodePageWithPreview`.

CC: @lgarczewski


lgarczewski:
LGTM.

PR #7917, opened 2015-07-22
don't run delete statement if array is empty

nmonterroso:
@Wikia/services-team @macbre 
https://wikia-inc.atlassian.net/browse/SERVICES-535

I suspect this db query was throwing an error when the preferences list was empty

### DIFF
--- a/includes/User.php
+++ b/includes/User.php
@@ -4851,7 +4851,7 @@ class User {
 		if ($wgPreferencesUseService) {
 			$preferenceNames = array_keys($this->userPreferences()->getPreferences($this->getId()));
 
-			(new WikiaSQL())
+			(new WikiaSQL())->skipIf(empty($preferenceNames))
 				->DELETE('user_properties')
 				->WHERE('up_user')->EQUAL_TO($this->getId())
 					->AND_('up_property')->NOT_IN($preferenceNames)
--- a/includes/User.php
+++ b/includes/User.php
@@ -4851,11 +4851,15 @@ class User {
 		if ($wgPreferencesUseService) {
 			$preferenceNames = array_keys($this->userPreferences()->getPreferences($this->getId()));
 
-			(new WikiaSQL())->skipIf(empty($preferenceNames))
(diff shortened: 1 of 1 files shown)


includes/User.php
```diff
@@ -4851,11 +4851,15 @@ protected function saveOptions() {
...
-				->WHERE('up_user')->EQUAL_TO($this->getId())
-					->AND_('up_property')->NOT_IN($preferenceNames)
-				->run($dbw);
+				->WHERE('up_user')->EQUAL_TO($this->getId());
+
+			if (!empty($preferenceNames)) {
+				$sql->AND_('up_property')->NOT_IN($preferenceNames);
+			}
+
+			$sql->run($dbw);
```
drsnyder, line 4862: Seems like this is also a bug with `WikiaSQL`. Should an `IN` operation be a noop if the scalar passed is empty?

drsnyder:
LGTM

includes/User.php
nmonterroso, line 4862: Maybe, but I'd rather it throw an error and alert (like it did), instead of silently throwing queries on the ground.

drsnyder, line 4862: Let the queries hit the floor!
let the queries hit the floor

macbre, line 4862: Making a DELETE query in `saveOptions` method is a bit counter-intuitive :) Can we add a comment here describing why we're doing that?

includes/User.php
```diff
@@ -4851,11 +4851,15 @@ protected function saveOptions() {
 		if ($wgPreferencesUseService) {
```
macbre, line 4851: Please consider using `UserSaveOptions` hook instead (or introducing a new one).

includes/User.php
nmonterroso, line 4862: I agree, it's definitely not something you'd expect to happen. Though, we only added it to maintain compatibility with what MediaWiki was doing (see the `else` statement)

PR #7918, opened 2015-07-22
SOC-788 Use cachebuster to bypass external cache

mklucsarits:
https://wikia-inc.atlassian.net/browse/SOC-788

@garthwebb @jsutterfield

### DIFF
--- a/extensions/wikia/ArticleComments/js/ArticleComments.js
+++ b/extensions/wikia/ArticleComments/js/ArticleComments.js
@@ -123,8 +123,7 @@
 					id: commentId,
 					method: 'axEdit',
 					rs: 'ArticleCommentsAjax',
-					useskin: window.skin,
-					cb: Date.now()
+					useskin: window.skin
 
 				}, function (json) {
 					if (!json.error) {
--- a/extensions/wikia/ArticleComments/js/ArticleComments.js
+++ b/extensions/wikia/ArticleComments/js/ArticleComments.js
@@ -123,7 +123,8 @@
 					id: commentId,
 					method: 'axEdit',
 					rs: 'ArticleCommentsAjax',
-					useskin: window.skin
+					useskin: window.skin,
+					cb: Date.now()
 
 				}, function (json) {
 					if (!json.error) {
--- a/extensions/wikia/ArticleComments/ArticleComments_setup.php
+++ b/extensions/wikia/ArticleComments/ArticleComments_setup.php
@@ -145,6 +145,11 @@ function ArticleCommentsAjax() {
 			$response->setContentType('text/html; charset=utf-8');
 		}
 
+		// Don't cache requests made to edit comment, see SOC-788
+		if ( $method == "axEdit" ) {
+			$response->setCacheDuration(0);
+		}
+
 		wfProfileOut(__METHOD__);
 		return $response;
 	}


extensions/wikia/ArticleComments/js/ArticleComments.js
```diff
@@ -123,7 +123,8 @@
 					id: commentId,
 					method: 'axEdit',
 					rs: 'ArticleCommentsAjax',
-					useskin: window.skin
+					useskin: window.skin,
+					cb: Date.now()
```
macbre: I'd rather find the root cause of cache invalidation problems then simply ignoring the cache.

hakubo: +1

as far as I remember we have enabled caching of any GET for 60 sec, could this be it?

macbre: Git commit please :) Or even better: HTTP response headers

hakubo: or maybe imwrong, if I'll find it I'll share it with you.

jsutterfield: Hmm, yea, it looks like in production there's an `Age` header which is getting added. It seems to expire after 2 minutes. I was able to recreate the bug in production for 2 minutes getting the following headers until it cleared:

``` HTTP
Age:14
Cache-Control:no-cache, must-revalidate

Age:43
Cache-Control:no-cache, must-revalidate

Age:93
Cache-Control:no-cache, must-revalidate

Age:0
Cache-Control:private, s-maxage=0, max-age=0, must-revalidate
(cache cleared here)
```

Those headers differ in dev. They always return

``` HTTP
Cache-Control:private, s-maxage=0, max-age=0, must-revalidate
```

Could be worth looking into more. Looks like we may need to ensure `s-maxage=0, max-age=0` gets added? We may just be able to add those [here](https://github.com/Wikia/app/blob/dev/includes/AjaxResponse.php#L130).

What do you think @mklucsarits?

macbre: @jsutterfield, try sending a request directly to one of the Apache nodes to get the response headers that the CDN layer processes. The `Cache-Control` you see on the client-side is actually `X-Pass-Cache-Control` generated on the backend.

jsutterfield: Ah, good to know @macbre. Yea, that does return different headers. I don't see the `X-Pass-Cache-Control` header, these are the results, but they are different.

``` HTTP
HTTP/1.1 200 OK
Date: Thu, 30 Jul 2015 18:25:51 GMT
Server: Apache
X-Content-Type-Options: nosniff
Status: 200 OK
Last-Modified: Thu, 30 Jul 2015 18:25:51 GMT
Expires: Mon, 26 Jul 1997 05:00:00 GMT
Cache-Control: no-cache, must-revalidate
Pragma: no-cache
Content-Encoding: gzip
Vary: Accept-Encoding
Content-Length: 808
Keep-Alive: timeout=15, max=10000
Connection: Keep-Alive
Content-Type: application/json; charset=utf-8
```

That was after Curling the url and adding the flag `-x ap-s38:80`.

That being said, I'm starting to worry that this PR is starting to languish as it's been open for a week already. Fixing the underlying issue would be preferred, but it's also a relatively edge-casey bug (as I understand it) and we have plenty of more business-critical tickets to work on. To ration our time effectively, I think this cb as a work around is sufficient for now. We can add a TODO as a reminder to address this later if we'd like.

macbre: @jsutterfield: I still think it's better to understand the code before applying the fix. That's the only way we can pay off the technical debt and avoid such issues in the future :)

`AjaxResponse` class is used to generate the [ArticleComments response](https://github.com/Wikia/app/blob/808a769df6cf8524aa6defcab4f971367e3e3fd8/extensions/wikia/ArticleComments/ArticleComments_setup.php#L136). Try setting the cache TTL to zero (as in [WikiFactory code](https://github.com/Wikia/app/blob/808a769df6cf8524aa6defcab4f971367e3e3fd8/extensions/wikia/WikiFactory/SpecialWikiFactory_ajax.php#L120)):

``` php
$response = new AjaxResponse($payload);
$response->setCacheDuration(0);
```

jsutterfield: Well, geez, that was super easy. This ticket had been bandied back and forth for a while before this PR and was starting to drag. Thanks for the help @macbre, that looks like exactly the way to do it :)

garthwebb: @macbre Can you double check the varnish config?  I can confirm what @jsutterfield has said.  Here's the AJAX request URL that I'm using:

``` bash
curl -v -x ap-s23:80 "http://garth.wikia.com/index.php?action=ajax&article=2460&convertToFormat=&id=2479&method=axEdit&rs=ArticleCommentsAjax&useskin=oasis"
```

1) Via debugging on a sandbox I can confirm that this request hits this line of code: https://github.com/Wikia/app/blob/dev/includes/AjaxResponse.php#L130
which sets the correct headers.
2) Via this curl command to the backend on `ap-s23`, I can confirm that this produces a `Cache-Control` with `no-cache`, an `Expires` header in the past and no `X-Pass-Cache-Control` to be found:

```
< HTTP/1.1 200 OK
< Date: Thu, 30 Jul 2015 21:49:23 GMT
* Server Apache is not blacklisted
< Server: Apache
< X-Content-Type-Options: nosniff
< Status: 200 OK
< Last-Modified: Thu, 30 Jul 2015 21:49:23 GMT
< Expires: Mon, 26 Jul 1997 05:00:00 GMT
< Cache-Control: no-cache, must-revalidate
< Pragma: no-cache
< Vary: Accept-Encoding
< Content-Length: 46
< Content-Type: application/json; charset=utf-8
```

3) When I remove the `-x` option to allow it to hit varnish, the `Cache-Control` header no longer has `no-cache` and there is no longer an `Expires` header:

```
< HTTP/1.1 200 OK
* Server Apache is not blacklisted
< Server: Apache
< X-Content-Type-Options: nosniff
< Status: 200 OK
< Last-Modified: Thu, 30 Jul 2015 21:57:40 GMT
< Content-Type: application/json; charset=utf-8
< X-Cacheable: YES
< Content-Length: 46
< Accept-Ranges: bytes
< Date: Thu, 30 Jul 2015 21:57:40 GMT
< Age: 0
< Connection: keep-alive
< X-Served-By: cache-wk-sjc3160-WIKIA, cache-sjc3134-SJC
< X-Cache: MISS, MISS
< X-Cache-Hits: 0, 0
< X-Timer: S1438293460.059463,VS0,VE113
< Vary: Accept-Encoding
< Set-Cookie: wikia_beacon_id=sMNTnjZbJC; domain=.wikia.com; path=/; expires=Tue, 26 Jan 2016 21:57:40 GMT;
< Set-Cookie: Geo={%22city%22:%22FIXME%22%2C%22country%22:%22US%22%2C%22continent%22:%22NA%22}; path=/
< Cache-Control: private, s-maxage=0, max-age=0, must-revalidate
```

I've retried the curl request several times and I always get new content every 2 minutes

garthwebb:
Ok, the setDuration with an explicit zero looks like it does the trick.

macbre:
:+1: 

I think we can change the title of this pull request ;)

extensions/wikia/ArticleComments/js/ArticleComments.js
macbre: @garthwebb: `AjaxResponse` is the legacy way of handling AJAX requests. It should emit response caching headers in a similar fashion as `WikiaResponse` - in this case `Cache-Control: private, s-maxage=0`. Please file a ticket for the Platform Team - we will fix it.

PR #7921, opened 2015-07-23
when setting preferences in memory, make sure to add defaults

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-536

Previously we were initializing a user's preferences to an empty array in `load()`. We changed that to default to `$this->defaultPreferences`, but some users have their preferences array saved in memcache already. This change makes it so that a user's preferences are merged onto the defaults when their preferences are set from memcache


nmonterroso:
Spoke to @drsnyder offline, got a verbal LGTM

PR #7922, opened 2015-07-23
Update Lookup User/Contribs and show it to the user on their User: page

garthwebb:
https://wikia-inc.atlassian.net/browse/SOC-1641


extensions/wikia/LookupContribs/SpecialLookupContribs_ajax.php
```diff
@@ -12,159 +13,144 @@
...
 	 * @param array $activityData data retrieved from LookupContribsCore::checkUserActivity()
 	 *
 	 * @return array
-	 *
-	 * @author Andrzej 'nAndy' Łukaszewski
 	 */
-	public static function prepareLookupUserData($activityData, $username) {
-		global $wgLang, $wgContLang;
+	public static function prepareLookupUserData( $activityData, $username ) {
+		$wg = F::app()->wg;
```
hakubo, line 141: I thought that using 'global' is the way to go
as using F::app()->wg only adds overhead without clear benefits.

@Wikia/platform-team amiright?

garthwebb, line 141: AFAIK use of `F::app()->wg` is still the standard for global variables as nothing different was ever communicated.

ghost, line 141: I do not think we have a strict standard about it. I adjust my style (or lack thereof) to be consistent within the scope of a particular file or piece of code. Generally, `global` is the way to go with the core code, whereas `F::app()` is for everything Wikia. Also, local scope is always faster than the same operation in global scope. And just declaring a variable as global with `global` (without even using it later) slows things down as PHP needs to check whether it exists and create a reference. Note, that passing/returning by reference in PHP is slower than passing/returning by value. Generally speaking, references are not the way to increase performance. That implies the wide topic of global scope in PHP should be approached in a way that does not rely on PHP global variables but uses local scope within instances of objects that can be easily accessed via Something::getInstance(). I am fairly sure `F::app()` addresses the problem. Although at the end of the day, any change via F::app() needs to be visible via `global`, too...

extensions/wikia/LookupContribs/SpecialLookupContribs_body.php
```diff
@@ -40,63 +40,63 @@ public function execute( $subpage ) {
...
+		if ( !empty( $this->mMode ) && !empty( $this->mWiki ) ) {
+			if ( !in_array( $this->mMode, array_keys( $this->mModes ) ) ) {
 				return;
 			}
 			$this->showWikiForm();
 		} else {
 			$this->showMainForm();
 		}
-		#$this->showUserList();
+		# $this->showUserList();
```
jsutterfield: Minor, but can we delete this?

extensions/wikia/LookupContribs/SpecialLookupContribs_helper.php
```diff
@@ -610,163 +625,174 @@ private function getArchive( $dbr ) {
...
+		$memc = F::app()->wg->Memc;
 
-		$fetched_data = array(
+		$fetched_data = [
 			'cnt' => 0,
-			'data' => array()
-		);
+			'data' => []
+		];
 		$dbr = wfGetDB( DB_SLAVE, 'stats', $this->getDBname() );
```
jsutterfield: What about waiting to get this db handler until after we're sure the value doesn't exist in memcache?

extensions/wikia/UserActivity/SpecialUserActivityController.class.php
```diff
@@ -0,0 +1,97 @@
...
+			'order' => $order,
+		] );
+		$data = $resp->getData();
+
+		$this->getResponse()->setData( [
+			'pageDescription' => wfMessage( 'user-activity-page-description' )->text(),
+			'tableTitle' => wfMessage( 'user-activity-table-title' )->text(),
+			'tableEdits' => wfMessage( 'user-activity-table-edits' )->text(),
+			'tableLastEdit' => wfMessage( 'user-activity-table-lastedit' )->text(),
+			'tableRights' => wfMessage( 'user-activity-table-rights' )->text(),
```
jsutterfield: Should all of these be `escaped()`? In case some admin decides to override those messages.

extensions/wikia/UserActivity/templates/wordmark.mustache
```diff
@@ -0,0 +1,6 @@
+{{#isText}}
+<!--<span class="wordmark text {{textSize}} {{textFont}}"><a href=""{{wikiUrl}}>{{text}}</a></span> -->
```
jsutterfield: Should this be commented out?

jsutterfield:
LGTM

extensions/wikia/WikiFactory/WikiFactory.php
```diff
@@ -497,6 +497,23 @@ static public function DomainToID( $domain ) {
 		return $city_id;
 	}
 
+	public static function getHostByDbName( $dbName ) {
```
armonr, line 507: Will this handle sandbox environments also?

armonr, line 507: Consider adding a description for this method.

extensions/wikia/UserActivity/styles/UserActivity.scss
```diff
@@ -0,0 +1,118 @@
...
+  background-color: #f0f0f0;
+  background-image: url(/extensions/wikia/VideoHandlers/images/Brackets.png);	/* inline */
+  background-position: 50px 25px;
+  background-repeat: no-repeat;
+  height: 90px;
+  width: 160px;
+}
+
+/* The Grid ---------------------- */
+.row { width: 940px; max-width: 100%; min-width: 768px; margin: 0 auto; }
```
armonr, line 15: Inline rules are against ["Only one rule per line" in the coding conventions](https://github.com/Wikia/guidelines/blob/master/SCSS/CodingConventions.md#whitespace-and-structure).

extensions/wikia/UserActivity/UserActivityController.class.php
```diff
@@ -0,0 +1,176 @@
...
+
+		if ( empty( $user ) ) {
+			return;
+		}
+
+		if ( $user->isBlocked() ) {
+			return;
+		}
+
+		if ( !$user->isLoggedIn() ) {
```
armonr, line 29: Minor: Lots of returns are normally discouraged. Combining these three conditionals could be an option :)

extensions/wikia/UserActivity/UserActivityController.class.php
```diff
@@ -0,0 +1,176 @@
...
+			'list' => 'users',
+			'ususers' => urlencode( $userName ),
+			'usprop' => 'groups',
+		];
+
+		$resp = \ApiService::foreignCall( $dbName, $params, \ApiService::API );
+
+		if ( ( $resp === false ) || ( empty( $resp['query']['users'][0]['groups'] ) ) ) {
+			return [];
+		} else {
```
armonr: Trivial: `else` is superfluous after the `return` statement

extensions/wikia/UserActivity/UserActivityController.class.php
```diff
@@ -0,0 +1,176 @@
...
+		if ( !$contribs->checkUser() ) {
+			$this->response->setData( $result );
+			return;
+		}
+
+		$contribs->setLimit( $limit );
+		$contribs->setOffset( $offset );
+
+		$activity = $contribs->checkUserActivity(
+			$addEditCount = true,
```
armonr, line 43: What's the purpose of the variable?

extensions/wikia/UserActivity/UserActivityController.class.php
```diff
@@ -0,0 +1,176 @@
...
+		}
+
+		$this->response->setData( $result );
+	}
+
+	protected function formatItems( $items ) {
+		$flattened = [];
+		foreach ( $items as $sortKey => $contribItem ) {
+			$lastEditTS = wfTimestamp( TS_MW, $contribItem[ 'last_edit' ] );
+			$localizedDate = \F::app()->wg->Lang->timeanddate( $lastEditTS, $localTZ = true );
```
armonr, line 62: Same here, this is a python-style method invocation :)

extensions/wikia/UserActivity/SpecialUserActivityController.class.php
```diff
@@ -0,0 +1,97 @@
...
+		$this->getResponse()->setData( [
+			'pageDescription' => wfMessage( 'user-activity-page-description' )->escaped(),
+			'tableTitle' => wfMessage( 'user-activity-table-title' )->escaped(),
+			'tableEdits' => wfMessage( 'user-activity-table-edits' )->escaped(),
+			'tableLastEdit' => wfMessage( 'user-activity-table-lastedit' )->escaped(),
+			'tableRights' => wfMessage( 'user-activity-table-rights' )->escaped(),
+			'items' => $data['items'],
+			'username' => $this->app->wg->User->getName(),
+			'total' => $data['total'],
+    		'totalReturned' => $data['totalReturned'],
```
armonr: whitespace

extensions/wikia/UserActivity/UserActivityController.class.php
garthwebb, line 29: I generally like to keep if statements as sparse as possible.  Bundling these together just invites more bundling and we get big multiline statements that are hard to read.

extensions/wikia/UserActivity/UserActivityController.class.php
garthwebb, line 43: Its to give context to what is otherwise a meaningless floating `true` and shorthand for setting this variable else where to pass here.

extensions/wikia/UserActivity/UserActivityController.class.php
garthwebb, line 62: This tells the reader that I want this date in the local timezone, rather than "the second argument to `timeanddate` is true for some reason"

extensions/wikia/UserActivity/styles/UserActivity.scss
garthwebb, line 15: This whole thing needs to be reworked by someone with better CSS skills.  This was just lifted from the foundation grid CSS.  This applies to the design as a whole as well since this hack had no designer participating.  As such don't want to spend much time fixing this CSS that we'll likely toss if this gets a real design.

extensions/wikia/WikiFactory/WikiFactory.php
garthwebb, line 507: The `\WikiFactory::getLocalEnvURL` call handles sandboxes.

armonr:
Thanks! :shipit:

extensions/wikia/UserActivity/SpecialUserActivityController.class.php
garthwebb: Gah, nope.  These are already escaped by handlebars.  ... make that doubly escaped by handlebars.

PR #7926, opened 2015-07-24
album_name_fc and genres were not being entered into Solr correctly d…

SeanColombo:
…ue to a difference in the expected case of some array values.

To verify the data is getting where you want now, you can go to sandbox-s1 and modify some code in /usr/wikia/slot1/current/src/maintenance/wikia/lyrics to output info, then run "perl /usr/wikia/backend/bin/run_maintenance --id=43339 --script='wikia/lyrics/LyricsWikiCrawler.php' >> /home/sean/lyricsSolrScrape.log" and inspect the log that it's appending to.

Hopefully once this fix is live, the indexing will cause album results to start showing up in the mobile app again. We will probably have to re-index the whole site for all of our current albums to show up though.

### DIFF
--- a/maintenance/wikia/lyrics/scrapers/AlbumScraper.class.php
+++ b/maintenance/wikia/lyrics/scrapers/AlbumScraper.class.php
@@ -18,11 +18,11 @@ class AlbumScraper extends BaseScraper {
 			'article_id' => $article->getId(),
 		];
 		$albumData = array_merge( $albumData, $this->getHeader( $article ) );
-		$albumData['album_lowercase'] = LyricsUtils::lowercase( $albumData['Album'] );
+		$albumData['album_lowercase'] = LyricsUtils::lowercase( $albumData['album'] );
 
 		$albumData['genres'] = $this->getGenres( $article );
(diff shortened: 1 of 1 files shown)


nandy-andy:
LGTM :+1: Thanks, Sean! :smile_cat:

This whole issue is quite a shocking - it seems it never worked as expected and we didn't catch it :cry:

SeanColombo:
It happens ^_^ thanks for the quick review @nandy-andy :D

nandy-andy:
Despite the fact it's merged, if you're bored on holidays @aquilax :palm_tree: have a look, please :smile_cat: (I saw your comments in other our PRs, so I assume you like it :smiling_imp:)

It's just because this whole issue smells for me and I'm embarrassed @SeanColombo did have to spend time on it ;)

aquilax:
10x @nandy-andy. The whole upper-lower case data soup is error prone and maybe it is a good idea to switch all key names to mb_lowercase to prevent such situations. :see_no_evil: 

@SeanColombo 10x for catching this. We should have tested it better.

PR #7930, opened 2015-07-27
Remove Uncyclopedia specifc checks

macbre:
Thanks to [PLATFORM-1055](https://wikia-inc.atlassian.net/browse/PLATFORM-1055) Uncyclopedia uses shared users database. We can remove these checks / comments from our code.

@wladekb / @michalroszka


extensions/wikia/UserRenameTool/RenameUserHelper.class.php
```diff
@@ -10,9 +10,7 @@ class RenameUserHelper {
 
 	const CLUSTER_DEFAULT = '';
 
-	public static $excludedWikis = array(
-		425, /* uncyclopedia */
-	);
+	public static $excludedWikis = [];
```
hakubo: is it edited dynamically?
maybe we could remove the var all together?

macbre: Good point, let's leave the code a bit better than it was :)

71b709c

ghost:
:+1:

extensions/wikia/UserRenameTool/RenameUserHelper.class.php
hakubo: :heart:

PR #7933, opened 2015-07-27
ADEN-2247 Fixed interstitial scaling on Safari and Internet Explorer

fraszczakszymon:
Also added condition to prevent multiple execution of onload function


resources/wikia/modules/iframeWriter.js
```diff
@@ -35,6 +36,10 @@ define('wikia.iframeWriter', [
 		}
 
 		iframe.onload = function () {
+			if (loaded) {
```
nandy-andy, line 40: What do you think about leaving a nice comment in which we blame IE and describe the issue?

nandy-andy:
LGTM :+1:

PR #7934, opened 2015-07-27
DAT-3008 make tabular default layout 

RafLeszczynski:


extensions/wikia/PortableInfobox/controllers/PortableInfoboxParserTagController.class.php
```diff
@@ -13,7 +13,7 @@ class PortableInfoboxParserTagController extends WikiaController {
 	private $markers = [ ];
 	private $supportedLayouts = [
 		'default',
-		'tabular'
+		'stacked'
```
sebastianmarzjan, line 16: That got me worried for a second about as to what will happen to the current layout="tabular" boxes, however I see that this is covered in the layout selection logic assigning the default val in such a case.

sebastianmarzjan:
Please fix the tests and then LGTM

PR #7935, opened 2015-07-27
Some other attributes had case-sensitivity issues. Debugged these cha…

SeanColombo:
…nges on sandbox to make sure it works correctly. Now album, length, cover, wikipedia all get through correctly.

For MAIN-4950


sebastianmarzjan:
lgtm

PR #7937, opened 2015-07-27
ADEN-2255 Added highimpactslot url param to enable INVISIBLE_HIGHIMPACT slot

fraszczakszymon:


nandy-andy:
LGTM :+1:

extensions/wikia/AdEngine/js/AdContext.js
```diff
@@ -76,7 +76,7 @@ define('ext.wikia.adEngine.adContext', [
 			context.providers.openX = true;
 		}
 
-		context.slots.invisibleHighImpact = (
+		context.slots.invisibleHighImpact = qs.getVal('highimpactslot', false) || (
```
nandy-andy: Will it work with `0` and `1`? Probably `Querystring` will cast it to `false` and `true` but it's better to ask ;) And I'm asking since most of our URL params are described in the way they accept `0` or `1`.

jacekjursza:
LG

PR #7938, opened 2015-07-27
SOC-1098 We don't let anons vote, so don't waste time checking if they did

garthwebb:
https://wikia-inc.atlassian.net/browse/SOC-1098

### DIFF
--- a/extensions/wikia/Wall/VoteHelper.class.php
+++ b/extensions/wikia/Wall/VoteHelper.class.php
@@ -1,45 +1,42 @@
 <?php
 
-// TODO: move this to services
-
+/**
+ * Class VoteHelper
+ */
 class VoteHelper {
+	const PAGE_CACHE_TTL = 60 * 60;
+
 	protected $userId;
 	protected $pageId;
 	protected $userIP = '';
 
-	function __construct( $user = null, $pageId = 0 ) {
+	function __construct( User $user = null, $pageId = 0 ) {
 		global $wgRequest;
-		wfProfileIn( __METHOD__ );
-		$this->userIP = $wgRequest->getIP();
-		if ( $user->isLoggedIn() ) {
-			$this->userId = $user->getId();
-		} else {
-			$this->userId = 0;
-		}
 
+		$this->userIP = $wgRequest->getIP();
+		$this->userId = $user->getId();
 		$this->pageId = $pageId;
-		wfProfileOut( __METHOD__ );
 	}
 
 	// TODO: caching ?
 	function getVotersList( $from = 0, $count = 50 ) {
-		wfProfileIn( __METHOD__ );
 		$dbr = wfGetDB( DB_SLAVE );
 		// TODO: abstract the query
-		$res = $dbr->select( array( 'page_vote', 'ipblocks' ),
-			array( 'user_id' ),
-			array(
+		$res = $dbr->select(
+			[ 'page_vote', 'ipblocks' ],
+			[ 'user_id' ],
+			[
 				"ifnull(ipb_expiry, 0) != 'infinity'",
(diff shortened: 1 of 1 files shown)


extensions/wikia/Wall/VoteHelper.class.php
```diff
@@ -52,162 +49,151 @@ function getVotersList( $from = 0, $count = 50 ) {
...
+		$dbr->insert(
+			'page_vote',
 			$values,
-		__METHOD__ );
+			__METHOD__
+		);
 
 		$cache = $this->getCache();
-		$cache->set( $this->getUserCacheKey(), 'Y', 60 * 60 );
+		$cache->set( $this->getUserCacheKey(), 'Y' );
```
garthwebb, line 75: There are memcached updates for updates and deletes and we should keep this data as long as it can stay before getting pushed out of cache.

extensions/wikia/Wall/VoteHelper.class.php
```diff
@@ -52,162 +47,151 @@ function getVotersList( $from = 0, $count = 50 ) {
...
+		$val = $row->cnt > 0;
 
-		// Y:N solve the problem with defult val
+		// Y:N solve the problem with default val
 		$cache->set( $this->getUserCacheKey(), $val ? 'Y': 'N' );
-		wfProfileOut( __METHOD__ );
 		return $val;
 	}
 
 	protected function getUserWhere() {
```
garthwebb, line 186: We don't need to care about anyone but logged in users.

extensions/wikia/Wall/VoteHelper.class.php
```diff
@@ -52,162 +47,151 @@ function getVotersList( $from = 0, $count = 50 ) {
...
-		}
-		wfProfileOut( __METHOD__ );
-		return $where;
+		return [
+			'article_id' => $this->pageId,
+			'user_id' => $this->userId,
+		];
 	}
 
 	protected function getUserCacheKey() {
```
garthwebb, line 193: Don't care about anon users.

extensions/wikia/Wall/VoteHelper.class.php
```diff
@@ -1,45 +1,40 @@
...
-	function __construct( $user = null, $pageId = 0 ) {
+	function __construct( User $user = null, $pageId = 0 ) {
 		global $wgRequest;
-		wfProfileIn( __METHOD__ );
-		$this->userIP = $wgRequest->getIP();
-		if ( $user->isLoggedIn() ) {
-			$this->userId = $user->getId();
-		} else {
-			$this->userId = 0;
-		}
```
garthwebb, line 18: A user who is not logged in will have an ID of zero (the `isLoggedIn` method just checks if `$this->id` != 0) so we don't need this conditional.

mklucsarits:
LGTM

extensions/wikia/Wall/VoteHelper.class.php
```diff
@@ -52,162 +47,151 @@ function getVotersList( $from = 0, $count = 50 ) {
...
 
 	protected function getUserWhere() {
-		wfProfileIn( __METHOD__ );
-		$where = array(
-				'article_id' => $this->pageId,
-		);
-
-		if ( $this->userId == 0 ) {
-			$where['user_id'] = 0;
-			$where['ip'] = $this->userIP;
```
kvas-damian, line 194: As I understand this code, it limits number kudos for anon users from one IP address. Was it intended to remove this check?

garthwebb, line 194: Anon users can't give kudos, so this check isn't necessary.  It likely allowing this was part of the original feature, but its not used that way now.

kvas-damian:
LGTM

PR #7939, opened 2015-07-27
Attributes on Special:Preferences

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-536

There are some attributes on Special:Preferences, so when `$user->setGlobalPreference()` was called, the save would fail (because the preference is not on the whitelist). This PR adds a list of attributes to `Preferences.php` to account for that possibility.

Also, because all a user's entries in `user_properties` are deleted in `User->saveOptions`, it's possible that:

``` php
$user->setGlobalPreference('someAttribute', 'foo'); // this won't work, 'foo' isn't on the whitelist
$user->saveOptions();

// in saveOptions
$dbw->delete( 'user_properties', array( 'up_user' => $this->getId() ), __METHOD__ );
// attribute 'foo' is now deleted, but wasn't saved by `setGlobalPreference`
$dbw->insert(...); // 'foo' isn't here either, since it's managed by the UserPreference service
```

This also changes the semantics of `$user->saveOptions` to only delete options when they are equal to the default value for that option. This way, a bug with the above scenario will manifest as "this preference isn't being updated" instead of "this preference is getting deleted"

Requires the change in Wikia/config#1203 for the whitelist

### DIFF
--- a/includes/Preferences.php
+++ b/includes/Preferences.php
@@ -36,11 +36,6 @@ class Preferences {
 		'searchlimit' => array( 'Preferences', 'filterIntval' ),
 	);
 
-	private static $attributes = [
-		'nickname',
-		'fancysig'
-	];
-
 	/**
 	 * @throws MWException
 	 * @param $user User
@@ -1590,7 +1585,8 @@ class Preferences {
 
 	// attributes that show up on the preferences page. TODO: separate somehow?
 	private static function getAttributes() {
-		return self::$attributes;
+		global $wgUserAttributeWhitelist;
+		return $wgUserAttributeWhitelist;
 	}
 }
 
--- a/includes/Preferences.php
+++ b/includes/Preferences.php
@@ -36,6 +36,11 @@ class Preferences {
 		'searchlimit' => array( 'Preferences', 'filterIntval' ),
 	);
 
+	private static $attributes = [
+		'nickname',
+		'fancysig'
+	];
+
 	/**
 	 * @throws MWException
 	 * @param $user User
@@ -109,7 +114,15 @@ class Preferences {
 	 * @return array|String
 	 */
 	static function getUserPreference( $name, $info, $user ) {
-		$val = $user->getGlobalPreference( $name );
+		$getCallback = function($property) use ($user) {
+			if (in_array($property, self::getAttributes())) {
+				return $user->getGlobalAttribute($property);
+			} else {
+				return $user->getGlobalPreference($property);
+			}
+		};
+
+		$val = $getCallback($name);
 
 		// Handling for array-type preferences
 		if ( ( isset( $info['type'] ) && $info['type'] == 'multiselect' ) ||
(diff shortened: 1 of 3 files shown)


drsnyder:
To make sure I understand this, Special:Preferences was iterating over all the settings and calling `setGlobalPreference`?

includes/Preferences.php
```diff
@@ -109,7 +109,15 @@ static function getPreferences( $user, IContextSource $context ) {
...
 	 */
 	static function getUserPreference( $name, $info, $user ) {
-		$val = $user->getGlobalPreference( $name );
+		$getCallback = function($property) use ($user) {
+			if (in_array($property, self::getAttributes())) {
+				return $user->getGlobalAttribute($property);
+			} else {
+				return $user->getGlobalPreference($property);
+			}
+		};
```
drsnyder: For clarity, It seems like this should be local function and not an anonymous one.

nmonterroso:
Yeah, it assumed that everything on `Special:Preferences` was a preference

includes/User.php
```diff
@@ -35,7 +35,7 @@
  * Int Serialized record version.
  * @ingroup Constants
  */
-define( 'MW_USER_VERSION', 8 );
+define( 'MW_USER_VERSION', 9 );
```
drsnyder, line 38: What is this used for?

nmonterroso, line 38: Basically a cache buster. Since some people now have their memcache data completely incorrect without an expiration, this will force a load from the database for all users.

includes/User.php
```diff
@@ -4898,6 +4897,10 @@ private function shouldOptionBeStored( $key, $value ) {
 		return false;
 	}
 
+	private function shouldOptionBeDeleted($key, $value) {
```
drsnyder: Would a better name be 'isDefaultOption`?

includes/User.php
drsnyder, line 38: I was afraid of that.

includes/User.php
nmonterroso: sure. 499fb49

drsnyder:
I'm beginning to wonder if we aren't going about this the wrong way. I tried to think of a pattern for handling this, but nothing came to mind. Filter?

The basic flow we want is `read-from-db -> fill-in-defaults` or when writing `remove-defaults -> upsert`. I wonder if the logically equivalent change could be made that conveys what we are trying to do?

includes/Preferences.php
nmonterroso: 9f7795a

nmonterroso:
The problem is that multiple places are reading the same data from the same tables, but we want one part of it to not write some parts of the table, but have it managed by the same class (`User`)

drsnyder:
Ugh. Is this the cost of integration-at-the-db?

nmonterroso:
@drsnyder I think so. Best way to deal with it is the double face-palm you saw earlier today, or with scotch-o-clock

drsnyder:
> @drsnyder I think so. Best way to deal with it is the double face-palm you saw earlier today, or with scotch-o-clock

Yeah, it seems like at the data layer we have no way to distinguish a preference from a property from a .... In hindsight, once we knew what each was maybe we should have added a column that identified it then update the column with what we now have in the white list.

nmonterroso:
We could still do that, but I think we could potentially run into the same issues

drsnyder:
Yeah- if you write preferences that are attributes it doesn't matter how you try and separate it :(.

PR #7940, opened 2015-07-28
SOC-1027 Don't redirect if we're constructing an article that isn't the current request's article

garthwebb:
https://wikia-inc.atlassian.net/browse/SOC-1027

### DIFF
--- a/extensions/wikia/ArticleComments/classes/ArticleCommentList.class.php
+++ b/extensions/wikia/ArticleComments/classes/ArticleCommentList.class.php
@@ -6,10 +6,13 @@
 class ArticleCommentList {
 	const CACHE_VERSION = 'v2';
 
+	/** @var Title */
 	private $mTitle;
 	private $mText;
 	private $mCommentId = null;
 	private $mComments = false;
+
+	/** @var bool|array */
 	private $mCommentsAll = false;
 	private $mCount = false;       // Count of comments actually loaded after paging rules are applied
 	private $mCountNested = false;
@@ -44,7 +47,7 @@ class ArticleCommentList {
 
 	function __construct() {
 		global $wgArticleCommentsMaxPerPage;
-		$this->setMaxPerPage($wgArticleCommentsMaxPerPage);
+		$this->setMaxPerPage( $wgArticleCommentsMaxPerPage );
 	}
 
 	public function setMaxPerPage( $val ) {
@@ -57,14 +60,17 @@ class ArticleCommentList {
 
 	/**
 	 * setId -- set mCommentId it will limit select to only this comment
+	 *
+	 * @param int $id
 	 */
-
 	public function setId( $id ) {
 		$this->mCommentId = $id;
 	}
 
 	/**
 	 * setTitle -- standard accessor/setter
+	 *
+	 * @param Title $title
 	 */
 	public function setTitle( Title $title ) {
 		$this->mTitle = $title;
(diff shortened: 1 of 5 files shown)


extensions/wikia/ArticleComments/classes/ArticleCommentList.class.php
```diff
@@ -916,63 +924,98 @@ static function formatList($comments) {
...
-	 * static entry point for hook
+	 * Static entry point for hook
 	 *
 	 * @param Title $title
 	 * @param Article $article
 	 *
-	 * @static
-	 * @access public
+	 * @return bool
 	 */
```
garthwebb: This is where the actual fix for this ticket takes place.

extensions/wikia/ArticleComments/classes/ArticleCommentList.class.php
```diff
@@ -916,63 +924,98 @@ static function formatList($comments) {
...
-					$query = array();
-
-					if ( $permalink || $commentId !== 0 ) {
-						if( $commentId !== 0 ) {
-							/** bugId:11179 @author: nAndy */
-							$permalink = $commentId;
-						}
+	static public function ArticleFromTitle( Title &$title, &$article ) {
+		// Don't bother checking for redirects if we're not loading this article for the current request
+		if ( !self::isTitleForCurrentRequest( $title ) ) {
```
garthwebb, line 946: This is the actual fix for this ticket.  Some code earlier the request creates one or more article objects via `Article::newFromTitle` and this code was intercepting it.  If the title in question was a comment it was issuing a redirect to the wall or forum page, with a jump to the comment there.  This makes sense if the user actually wants to load the comment, but if the comment is being loaded to get its info, we shouldn't be redirecting the user.

extensions/wikia/ArticleComments/classes/ArticleCommentList.class.php
```diff
@@ -916,63 +924,98 @@ static function formatList($comments) {
...
+			return false;
+		}
+
+		if ( $title->getPrefixedDBkey() != F::app()->wg->Title->getPrefixedDBkey() ) {
+			return false;
+		}
+	}
+
+	static private function canSetRedirect() {
+		$wg = F::app()->wg;
```
kvas-damian: you could store `Request` in variable while you are not touching `wg` in this method

kvas-damian:
One minor comment. Besides that it looks good to me.

extensions/wikia/ArticleComments/classes/ArticleCommentList.class.php
```diff
@@ -916,63 +924,98 @@ static function formatList($comments) {
...
+		$redirectTitle = Title::newFromText( $parts['title'], MWNamespace::getSubject( $title->getNamespace() ) );
+		if ( empty( $redirectTitle ) ) {
+			return true;
+		}
+
+		$query = [];
+		$commentId = $title->getArticleID();
+		$permalink = F::app()->wg->Request->getInt( 'permalink', 0 );
+
+		// Use comment ID if available - https://wikia.fogbugz.com/f/cases/11179
```
jsutterfield, line 970: Are we still linking to fogbugz tickets?

garthwebb, line 970: We still have a reduced access FogBugz account with all our history, so I'm just making a link out of the possibly confusing number given in the [previous comment](https://github.com/Wikia/app/pull/7940/files#diff-e4fe86c4b1039ccf72823b9e6b3135b9L959).

extensions/wikia/ArticleComments/classes/ArticleCommentList.class.php
garthwebb: Sounds like a good idea!

jsutterfield:
LGTM

PR #7943, opened 2015-07-28
SOC-786 Create MediaWiki Client For User Attribute Service

jsutterfield:
This ticket used swagger-codegen to create a php client for communicating with the User Attribute Service (see [here](https://github.com/Wikia/app/blob/dev/lib/Swagger/README.md) for more info on the swagger-codegen). That client was then configured to use the dependency injection library the services team set-up.

Ticket: https://wikia-inc.atlassian.net/browse/SOC-786


includes/User.php
```diff
@@ -23,6 +23,7 @@
 use Wikia\DependencyInjection\Injector;
 use Wikia\Logger\Loggable;
 use Wikia\Service\User\Preferences\UserPreferences;
+use Wikia\Service\User\Attributes\UserAttributes;
```
jsutterfield, line 26: Whoa! This is a huge pull request!

Fear not, fair code reviewer. The bulk of this is the auto-generated client code created by swagger-autogen which looks at the Swagger annotations in the User Attribute Service. The code which I added to wire that client into MW starts [here](https://github.com/Wikia/app/pull/7943/files#diff-5f12f40345b8d749abbb52d78fda227eR1) which is much more manageable.

lib/Wikia/src/DependencyInjection/InjectorInitializer.php
```diff
@@ -12,6 +13,7 @@ public static function init(CacheProvider $cacheProvider = null) {
 			(new InjectorBuilder())
 				->withCache($cacheProvider)
 				->addModule(new PreferenceModule())
+				->addModule(new AttributesModule())
```
jsutterfield, line 16: If you're not familiar with the dependency injection library we just added (which I assume you're not b/c it was just added, I wasn't), this is the entry point.

When MW lurches into existence, this Attributes Module is called which binds a set of interfaces to concrete classes which implement those interfaces. It also binds the class constant `UserAttributes::DEFAULT_ATTRIBUTES` to `\User::getDefaultOptions();`. All this allows the various dependencies to be mocked easily for testing (which you can see in the unit tests below).

Instantiating the client is done through the DI library [here](https://github.com/Wikia/app/pull/7943/files#diff-897fba1dd6b8edc3238762e35d0b7c71R275) which takes care of resolving all the dependencies, and injecting instances of those dependencies.

lib/Wikia/src/Service/User/Attributes/AttributesModule.php
```diff
@@ -0,0 +1,19 @@
...
+use Wikia\DependencyInjection\Module;
+use Wikia\Persistence\User\Attributes\AttributePersistence;
+use Wikia\Persistence\User\Attributes\AttributePersistenceSwagger;
+
+class AttributesModule implements Module {
+	public function configure( InjectorBuilder $builder ) {
+		$builder
+			->bind( AttributeService::class )->toClass( AttributeKeyValueService::class )
+			->bind( AttributePersistence::class )->toClass( AttributePersistenceSwagger::class )
+			->bind( UserAttributes::DEFAULT_ATTRIBUTES )->to( function() {
```
garthwebb, line 15: When does this become relevant?  Does the `@Inject({` comment below get treated as an annotation somehow?

garthwebb:
LGTM

lib/Wikia/src/Service/ForbiddenException.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+namespace Wikia\Service;
+
+class ForbiddenException extends \Exception {
+	const CODE = 403;
+}
```
kvas-damian: Missing new line at the end of file

lib/Wikia/src/Service/User/Attributes/AttributeKeyValueService.php
```diff
@@ -0,0 +1,74 @@
...
+		$this->persistenceAdapter = $persistenceAdapter;
+	}
+
+	public function setAttribute( $userId, $attribute ) {
+		if ( empty( $attribute ) || $userId == 0 ) {
+			return false;
+		}
+
+		try {
+			$profiler_start = $this->startProfile();
```
kvas-damian: our PHP guidelines suggest lowerCamelCase for variable names.

lib/Wikia/src/Service/User/Attributes/AttributeKeyValueService.php
```diff
@@ -0,0 +1,74 @@
...
+	 *    Wikia\Persistence\User\Attributes\AttributePersistence::class
+	 * })
+	 * @param AttributePersistence $persistenceAdapter
+	 */
+	function __construct( AttributePersistence $persistenceAdapter ) {
+		$this->persistenceAdapter = $persistenceAdapter;
+	}
+
+	public function setAttribute( $userId, $attribute ) {
+		if ( empty( $attribute ) || $userId == 0 ) {
```
kvas-damian: What do you think about strict comparison `===`?

kvas-damian: ~~Why it's not possible to set empty value for attribute? 
I can imagine setting attribute flags - (`true/false` value). In this case I cannot set false value. Additionally there is no way to delete attribute right now.~~

kvas-damian: NVM I misinterpreted code ;)

lib/Wikia/src/Service/User/Attributes/AttributeKeyValueService.php
```diff
@@ -0,0 +1,74 @@
...
+			return false;
```
kvas-damian: Shouldn't we throw an exception in this case?

lib/Wikia/src/Service/User/Attributes/AttributeKeyValueService.php
```diff
@@ -0,0 +1,74 @@
+<?php
+
+namespace Wikia\Service\User\Attributes;
+
+use Wikia\Logger\Loggable;
+use Wikia\Persistence\User\Attributes\AttributePersistence;
+use Wikia\Util\WikiaProfiler;
+use Wikia\Service\User\Attributes;
+
+class AttributeKeyValueService implements AttributeService {
```
kvas-damian, line 10: Don't we need `delete` method?

lib/Wikia/src/Service/User/Attributes/AttributeKeyValueService.php
```diff
@@ -0,0 +1,74 @@
...
+	 * @Inject({
+	 *    Wikia\Persistence\User\Attributes\AttributePersistence::class
+	 * })
+	 * @param AttributePersistence $persistenceAdapter
+	 */
+	function __construct( AttributePersistence $persistenceAdapter ) {
+		$this->persistenceAdapter = $persistenceAdapter;
+	}
+
+	public function setAttribute( $userId, $attribute ) {
```
kvas-damian: While we're in `AttributeKeyValueService` class what do you think about making method names more simple, eg. `set`, `get`?

lib/Wikia/src/Service/User/Attributes/AttributeKeyValueService.php
jsutterfield, line 10: I added it in the API for the service, but I don't think it's a very common operation. From what I can tell the User class doesn't even expose a way to do it, you can get and update attributes, but not delete.

That being said, it is a functionality we expose in the API so we should probably support it in the client as well. I'll update.

lib/Wikia/src/Service/User/Attributes/AttributeKeyValueService.php
jsutterfield: Sure, that sounds fine to me.

lib/Wikia/src/Service/User/Attributes/AttributesModule.php
jsutterfield, line 15: Yea, exactly. The DI library looks at that annotation to determine what to inject. It sees that we want to inject `UserAttributes::DEFAULT_ATTRIBUTES` which is bound to this function, so it injects the result of this function.

lib/Wikia/src/Service/ForbiddenException.php
jsutterfield: Added new line.

lib/Wikia/src/Service/User/Attributes/AttributeKeyValueService.php
```diff
@@ -0,0 +1,91 @@
...
+	 * })
+	 * @param AttributePersistence $persistenceAdapter
+	 */
+	function __construct( AttributePersistence $persistenceAdapter ) {
+		$this->persistenceAdapter = $persistenceAdapter;
+	}
+
+	public function set( $userId, $attribute ) {
+		if ( empty( $attribute ) || $userId === 0 ) {
+			return false;
```
kvas-damian: There was a question to this line which most likely got missed after code changes ;)
Shouldn't we throw an exception in this case?

jsutterfield: Yep, that definitely got missed :) Sure, that works too. A lot of this code was heavily inspired by the User Preferences Service (read, borderline plagiarized), so this was a hold over from that. An exception sounds good there. I'll add it.

kvas-damian:
LGTM

PR #7956, opened 2015-07-29
HG-785 - replace data://uri with //:0 for images in HG article API response

katanka:
Because base 64 image processing is slower on mobile devices, it is faster to use a `src='//:0'`.


lizlux:
This looks good, I'm just wondering if we want to updated the many places in the mercury app that reference `emptyGif:'data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAQAIBRAA7'`.

PR #7962, opened 2015-07-30
DataMartService - check wgStatsDBEnabled before making a query

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-1378

Prevent making queries to `statsdb_mart` when `$wgStatsDBEnabled` is set to false. Wikia One doesn't need to get any data from there.

Plus a small code cleanup.

@michalroszka / @wladekb


wladekb:
lgtm

PR #7964, opened 2015-07-30
(CE-2382) Escape user input on Flags HQ

Grunny:
Fixes stored XSS vulnerability in Flags HQ. All user input must be
properly escaped.


lukaszkonieczny:
lgtm

PR #7965, opened 2015-07-30
(MAIN-5060) Fix Blog comment redirects

Grunny:
The fix for SOC-1027 (#7940) added ArticleCommentList::isTitleForCurrentRequest
to check if the comment was the current title, but it failed to return anything
when this was true, so the self::isTitleForCurrentRequest( $title ) check would
always evaluate to false (empty) and no redirect to the Blog page would occur
when viewing comments.

This explicitly returns true when the checks in isTitleForCurrentRequest pass
and adds a return value for the ArticleFromTitle when the redirect occurs, as
all hook methods require a return value.

/cc @garthwebb


ghost:
Makes total sense. LGTM.

garthwebb:
D'oh!  Makes sense

PR #7978, opened 2015-07-31
CONCF-741 change media API so it works well with Curated Main Page Editor

bognix:
/cc @inez @mari-mcmurtrie @hakubo 

We would like to re-use existing API for image upload inside Curated Main Page Editor.
In order to do this I needed to add one wg variable check which allows for main page editing without logging in. This wg variable is only going to be enabled on our test wikis until we have authorization in place.

related PR: https://github.com/Wikia/mercury/pull/1150

### DIFF
--- a/extensions/VisualEditor/wikia/ApiAddMediaPermanent.php
+++ b/extensions/VisualEditor/wikia/ApiAddMediaPermanent.php
@@ -40,11 +40,11 @@ class ApiAddMediaPermanent extends ApiAddMedia {
 			$file->upload( $tempFile->getPath(), '', $pageText ? $pageText : '' );
 		}
 
-		return array(
+		return [
 			'title' => $file->getTitle()->getText(),
 			'url' => $file->getUrl(),
 			'article_id' => $file->getTitle()->getArticleID()
-		);
+		];
 	}
 
 	private function executeVideo() {
--- a/extensions/VisualEditor/wikia/ApiAddMediaPermanent.php
+++ b/extensions/VisualEditor/wikia/ApiAddMediaPermanent.php
@@ -29,7 +29,9 @@ class ApiAddMediaPermanent extends ApiAddMedia {
 		$duplicate = $this->getFileDuplicate( $tempFile->getLocalRefPath() );
 		if ( $duplicate ) {
 			return array(
-				'title' => $duplicate->getTitle()->getText()
+				'title' => $duplicate->getTitle()->getText(),
+				'url' => $duplicate->getUrl(),
+				'image_id' => $duplicate->getTitle()->getArticleID()
 			);
 		} else {
 			$title = $this->getUniqueTitle(
@@ -42,7 +44,8 @@ class ApiAddMediaPermanent extends ApiAddMedia {
 			$file->upload( $tempFile->getPath(), '', $pageText ? $pageText : '' );
 			return array(
 				'title' => $file->getTitle()->getText(),
-				'url' => $file->getUrl()
+				'url' => $file->getUrl(),
+				'image_id' => $file->getTitle()->getArticleID()
 			);
 		}
 	}
--- a/extensions/VisualEditor/wikia/ApiAddMediaPermanent.php
+++ b/extensions/VisualEditor/wikia/ApiAddMediaPermanent.php
@@ -41,7 +41,8 @@ class ApiAddMediaPermanent extends ApiAddMedia {
 			$file = new LocalFile( $title, RepoGroup::singleton()->getLocalRepo() );
 			$file->upload( $tempFile->getPath(), '', $pageText ? $pageText : '' );
 			return array(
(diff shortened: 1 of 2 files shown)


extensions/VisualEditor/wikia/ApiAddMediaPermanent.php
```diff
@@ -41,7 +41,8 @@ private function executeImage() {
 			$file = new LocalFile( $title, RepoGroup::singleton()->getLocalRepo() );
 			$file->upload( $tempFile->getPath(), '', $pageText ? $pageText : '' );
 			return array(
-				'title' => $file->getTitle()->getText()
+				'title' => $file->getTitle()->getText(),
+				'url' => $file->getUrl()
```
inez: API response should be consistent regardless if uploaded image is duplicate or not - so perhaps you should apply similar modification around line 32 of this file. However: this API originally didn't return URL because client should be able to construct URL to thumbnail based on title (and requested size) - if such functionality (of constructing URL) is currently missing in Mercury then I suggest building it for Mercury - it would be more useful and generic than just modifying this one API.

bognix: It would involve firing another request to MW to generate Title object based on file name and then generating URL for it. Or maybe you know a better way of doing this without firing another request? I was thinking about adding separate API method for mercury in MW so I don't have to change this one but there would be too much duplicated code and I decided not to proceed this way.

extensions/VisualEditor/wikia/ApiAddMediaPermanent.php
```diff
@@ -41,7 +43,9 @@ private function executeImage() {
 			$file = new LocalFile( $title, RepoGroup::singleton()->getLocalRepo() );
 			$file->upload( $tempFile->getPath(), '', $pageText ? $pageText : '' );
 			return array(
-				'title' => $file->getTitle()->getText()
+				'title' => $file->getTitle()->getText(),
```
inez: Given that response becomes little but more complex - I suggest that there is only one place which constructs it - because on file object. Something like:

```
if ( $duplicate ) {
    $file = $duplicate;
} else {
    $file = someOtherWayOfGettingFile();
}
return array(
    'title' => $file->getTitle()->getText(),
    'url' => $file->getUrl(),
    'image_id' => $file->getTitle()->getArticleID()
)

```

extensions/VisualEditor/wikia/ApiAddMediaPermanent.php
```diff
@@ -41,7 +43,9 @@ private function executeImage() {
...
+				'url' => $file->getUrl(),
+				'image_id' => $file->getTitle()->getArticleID()
```
inez: Well in MW images do not really have ids - article have ids and there are images associated with those articles based on titles. I'm encouraging you to not create new term - image_id - unless there is a really good reason for it. Other than that this change looks good to me.

extensions/VisualEditor/wikia/ApiAddMediaPermanent.php
```diff
@@ -28,9 +28,7 @@ private function executeImage() {
 		);
 		$duplicate = $this->getFileDuplicate( $tempFile->getLocalRefPath() );
 		if ( $duplicate ) {
```
tomasznapieralski, line 30: What do you think about converting if..else to single if, where condition will be `!$file = $duplicate`?

extensions/VisualEditor/wikia/ApiAddMediaPermanent.php
```diff
@@ -40,10 +38,13 @@ private function executeImage() {
 			}
 			$file = new LocalFile( $title, RepoGroup::singleton()->getLocalRepo() );
 			$file->upload( $tempFile->getPath(), '', $pageText ? $pageText : '' );
-			return array(
-				'title' => $file->getTitle()->getText()
-			);
 		}
+
+		return array(
```
hakubo: minor: use shorthand [] for array

extensions/VisualEditor/wikia/ApiAddMediaPermanent.php
bognix, line 30: let's leave it as it is so we don't introduce any accidental regression

hakubo:
LGTM

rogatty:
LGTM

extensions/VisualEditor/wikia/ApiAddMediaPermanent.php
inez, line 30: Or:

```
$file = $this->getFileDuplicate( $tempFile->getLocalRefPath() );
if(!$file) {
    $title = $this->getUniqueTitle(
        wfStripIllegalFilenameChars( $this->mParams['title'] )
    }
    ...
}
```

Anyways this is pretty minor and not a merge blocker for me.

vforge:
LGTM

PR #7979, opened 2015-07-31
SOC-1085 Use text rather than parse

garthwebb:
https://wikia-inc.atlassian.net/browse/SOC-1085


jsutterfield:
:shipit: :shipit: :shipit: :shipit: :shipit:

kvas-damian:
LGTM

PR #7983, opened 2015-08-03
SOC-678 Update (set|get)LocalAttribute call to (get|set)LocalPreference

jsutterfield:
Updated a `(set|get)LocalAttribute` call to `(get|set)LocalPreference`. The user data, "EditorMainPageNotificationHidden", is used to determine whether to display notification to the user, which affects site behavior and therefore falls under preferences rather than attributes.


rafalkalinski:
LGTM

mklucsarits:
LGTM

PR #7987, opened 2015-08-03
SOC-1112 Make sure we catch and log any errors while looking up the user mapping

garthwebb:
It doesn't actually matter why this fails for the user.  Its just for showing a connection success message.

https://wikia-inc.atlassian.net/browse/SOC-1112


mklucsarits:
LGTM

jsutterfield:
LGTM

PR #7989, opened 2015-08-03
don't save a preference value with what's already saved for the user

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-572

We're currently always saving a preference when `setGlobalPreference` is called, regardless of whether or not the value is already stored for that user. This change makes it so only preferences that differ from what's stored in the database are set.


ArturKlajnerok:
Lgtm :+1:

PR #7993, opened 2015-08-04
Log strict standard errors in /var/log/php

bognix:
Error Reporter creates P3s with Strict Standards violations.
I figured it makes sense to log them to /var/log/php so we can detect them during development.

This solution is the one which worked for me but I'm not sure if it's the best one.

/cc @macbre @michalroszka @hakubo


hakubo:
Nice, thanks!

ghost:
:+1:

vforge:
LGTM, cheers

macbre:
:+1:

PR #7995, opened 2015-08-04
Change @internal from classes in includes

vforge:
Remove obsolete "@" from docblocks allows us to see classnames without strikeout in phpStorm :)

/cc @hakubo @lgarczewski @bognix @Warkot


Warkot:
LGTM as phpDoc denotes a different use of `@internal` tag then we have in our code.
http://www.phpdoc.org/docs/latest/references/phpdoc/tags/internal.html

hakubo:
LGTM

PR #7999, opened 2015-08-04
CE-2399 Fix for fetching default pages list without filtering

kamilkoterba:
https://wikia-inc.atlassian.net/browse/CE-2399

It wasn't available for mcache key creation so it was returning empty result.

@Wikia/community-engineering


lukaszkonieczny:
lgtm

PR #8001, opened 2015-08-04
Lazily Inject PreferencePersistenceMySQL

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-575

We were always instantiating `PreferencePersistenceMySQL` and creating database connections even if they weren't used. This change makes it so that `PreferencePersistenceMySQL` is only created if it's used, using [lazy injection](http://php-di.org/doc/lazy-injection.html) in the dependency injection framework. 

The diff here is pretty huge, but it's mostly new dependency code being added. The actual changes for lazy injection is in 33ab104


drsnyder:
Are the large number of files coming from the `proxy-manager`? Github isn't able to display the diff.

nmonterroso:
Yeah, `proxy-manager` and its dependencies. The actual changes on our side are in 33ab104

drsnyder:
Haha, all of that for 3 lines.

Is there an additional build step required or does the DI framework do that when the files are autoloaded?

nmonterroso:
No build step required, the DI framework takes care of the magic

drsnyder:
lgtm

ArturKlajnerok:
Lgtm :+1:

PR #8004, opened 2015-08-05
CE-2405 Create a maintenance script for reruning the local rename job

adamkarminski:

This is a maintenance script that makes it easier to rerun the local rename job for all wikias that a user has contributed at. It saves logs from the procedure to a subpage in a User's namespace. The goal here is to enable a ComSup member and a user to be able to take a look at it and ensure that every wikia has been covered.

The next step will be providing a visual interface for running the script.

CC: @michalroszka @Grunny @lukaszkonieczny

### DIFF
--- a/includes/User.php
+++ b/includes/User.php
@@ -70,6 +70,7 @@ class User {
 	 * Traits extending the class
 	 */
 	use PowerUserTrait;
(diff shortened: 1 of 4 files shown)


adamkarminski:
LGTM from @michalroszka :)

PR #8008, opened 2015-08-05
Ce 2407

MariuszCz:
Fixed: Flag Insights fail if there is no flags on wiki.
Added: filtering only if there are some types of flags.


extensions/wikia/Insights/InsightsController.class.php
```diff
@@ -85,9 +84,11 @@ private function renderSubpage() {
 	private function renderFlagsFiltering() {
 		if ( $this->model instanceof InsightsFlagsModel ) {
 			/* Enable flags filter in layout */
```
kamilkoterba: As we're not using flagsFiltering anymore this comment is misleading. Can be removed.

kamilkoterba:
LGTM

PR #8014, opened 2015-08-05
default preferences fix

nmonterroso:
@Wikia/services-team 

I noticed a bug introduced in [earlier](https://github.com/Wikia/app/commit/18ab54b8bac76bd9ae2d6c944613697fa686ee6c#diff-897fba1dd6b8edc3238762e35d0b7c71R4845) where, because we are now preventing the `User` object from deleting preferences contained by the `UserPreferences` module, but we're setting defaults which contain non-preferences into that `UserPreferences` module, we wouldn't properly delete non-preferences in the `user_properties` table when the user reset those properties to their default.

This hasn't been brought up (at least that I know about) in any bug reports, but we should get ahead of that and push this out since it will definitely come up at some point.

### DIFF
--- a/lib/Wikia/src/Service/User/Preferences/PreferenceModule.php
+++ b/lib/Wikia/src/Service/User/Preferences/PreferenceModule.php
@@ -18,7 +18,19 @@ class PreferenceModule implements Module {
 				return $wgHiddenPrefs;
 			})
 			->bind(UserPreferences::DEFAULT_PREFERENCES)->to(function() {
-				return User::getDefaultOptions();
+				global $wgUserPreferenceWhiteList;
+				$defaultOptions = User::getDefaultOptions();
+				$defaultOptionNames = array_keys($defaultOptions);
+
+				return array_reduce(
+					$defaultOptionNames,
+					function($preferences, $option) use ($wgUserPreferenceWhiteList, $defaultOptions) {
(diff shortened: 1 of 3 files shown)


lib/Wikia/src/Service/User/Preferences/PreferenceModule.php
```diff
@@ -18,7 +18,19 @@ public function configure(InjectorBuilder $builder) {
...
+
+				return array_reduce(
+					$defaultOptionNames,
+					function($preferences, $option) use ($wgUserPreferenceWhiteList, $defaultOptions) {
+						if (in_array($option, $wgUserPreferenceWhiteList['literals'])) {
+							$preferences[$option] = $defaultOptions[$option];
+						}
+
+						return $preferences;
+					}, []);
```
drsnyder: Should we separate this inner block into `User::getDefaultPreferences`? If I understand correctly what's happening we are replacing "get default options" with "get default preferences".

drsnyder:
Minor comment, otw, lgtm.

lib/Wikia/src/Service/User/Preferences/PreferenceModule.php
nmonterroso: That would make this easier, for suresies. I'll update.

nmonterroso: a12830c

PR #8015, opened 2015-08-05
preserve query string on 404Handler redirect to a known Special or Ti…

owend:
…tle page

Redirecting from a short url to a /wiki/ url caused the query string to drop off.  Can't think of any reason to do that, and there are some cases where it is the preferred behavior.  

@macbre 
Details in ticket PLATFORM-1358


hakubo:
LGTM

macbre:
:+1: 

Sorry for the delayed response, I'm back from PTO

PR #8016, opened 2015-08-06
(CE-2408) Updating Infobox translations from Crowdin

Grunny:
Update infobox related translations from Crowdin.


kamilkoterba:
lgtm. 

New translations from Insights are not used yet because CE-flags-insights is not yet merged to dev, but I'd say it's save to push them.

PR #8018, opened 2015-08-06
CE-2388 Fix for fatal error on non existent parent page. 

kamilkoterba:
If parent page doesn't exist it should create it

https://wikia-inc.atlassian.net/browse/CE-2388

@Wikia/community-engineering


Grunny:
LGTM.

PR #8020, opened 2015-08-06
CE-2411 Allow to rerun RenameUser_local

adamkarminski:

This Pull Request provides a simple interface for running the `RerunRenameUserLocal` maintenance script. It also brings a few minor fixes to the script itself.


extensions/wikia/BrokenRenameFix/BrokenRenameFixTask.class.php
```diff
@@ -0,0 +1,39 @@
...
+
+		if ( !file_exists( $rerunRenameScript ) ) {
+			$this->error( "The maintenance script {$rerunRenameScript} does not exist." );
+			return false;
+		} else {
+			$this->info( "Running the {$rerunRenameScript}. Please wait for the results.
+			It may take a long time, depending on the amount of wikias the user edited at." );
+
+			$cmd = sprintf( 'SERVER_ID=177 php %s --userId=%d --oldName="%s" --newName="%s"',
+				$rerunRenameScript, $userId, $oldName, $newName );
```
Grunny: Never (ever! ;)) construct a shell command from user input without escaping them (even if the user is theoretically trusted as a staff member). Always make use of something that sanitizes it to avoid RCE. In MW, use `wfEscapeShellArg` (which will also add quotes around the variables), you can grep for examples of it.

adamkarminski: OMG! :scream: Taking care of it immidiately!

adamkarminski:
ping @Grunny :)

extensions/wikia/BrokenRenameFix/SpecialBrokenRenameFixController.class.php
```diff
@@ -0,0 +1,162 @@
...
+		$newName;
+
+	function __construct() {
+		parent::__construct( 'BrokenRenameFix', 'renameuser', true );
+	}
+
+	public function index() {
+		Wikia::addAssetsToOutput( 'special_broken_rename_fix' );
+
+		$this->wg->Out->setPageTitle( wfMessage( 'brf-title' )->escaped() );
```
Grunny: There's no permissions check here right now, so an anon or any non-staff member could access this form.

extensions/wikia/BrokenRenameFix/templates/SpecialBrokenRenameFixController_index.php
```diff
@@ -0,0 +1,20 @@
...
+		<label class="brf-form-label" for="brf-form-user-id"><?= $userIdLabel ?></label>
+		<input class="brf-form-input" type="text" name="brf-form-user-id" id="brf-form-user-id" value="<?= Sanitizer::encodeAttribute( $userIdValue ) ?>">
+
+		<label class="brf-form-label" for="brf-form-old-name"><?= $oldNameLabel ?></label>
+		<input class="brf-form-input" type="text" name="brf-form-old-name" id="brf-form-old-name" value="<?= Sanitizer::encodeAttribute( $oldNameValue ) ?>">
+
+		<label class="brf-form-label" for="brf-form-new-name"><?= $newNameLabel ?></label>
+		<input class="brf-form-input" type="text" name="brf-form-new-name" id="brf-form-new-name" value="<?= Sanitizer::encodeAttribute( $newNameValue ) ?>">
+
+		<input type="hidden" name="brf-form-token" value="<?= $editToken ?>">
```
Grunny: As a best practice, edit tokens should also be wrapped in `Sanitizer::encodeAttribute`.

adamkarminski:
Thank you @Grunny! Should be better now. :)

extensions/wikia/BrokenRenameFix/SpecialBrokenRenameFixController.class.php
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	/**
+	 * Creates a link to logs that will be available after the rename rerun finishes.
+	 *
+	 * @return string An HTML <a> element
+	 * @throws MWException
+	 */
+	private function getLogsLink() {
+		$logTitleText = sprintf( '%s/%s to %s rename rerun log',
```
Grunny: This isn't used any more, right? i.e. we're not saving the log on the wikia, so this link can be removed.

extensions/wikia/BrokenRenameFix/SpecialBrokenRenameFixController.class.php
```diff
@@ -0,0 +1,166 @@
...
+		$this->oldName = $this->request->getVal( 'brf-form-old-name' );
+		$this->newName = $this->request->getVal( 'brf-form-new-name' );
+
+		if ( empty( $this->userId ) || empty( $this->oldName ) || empty( $this->newName ) ) {
+			$this->addNotice( wfMessage( 'brf-error-empty-fields' )->escaped() );
+			$status = false;
+		}
+
+		$usernameFromId = User::newFromId( $this->userId )->getName();
+		if ( $usernameFromId !== $this->oldName && $usernameFromId !== $this->newName ) {
```
Grunny: For rerunning local renames, the user ID must match the new name, not the old name. If the user ID is still associated with the old user name, then the global rename needs to be rerun, not just the local.

extensions/wikia/BrokenRenameFix/BrokenRenameFix.setup.php
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Messages
+ */
+$wgExtensionMessagesFiles['BrokenRenameFix'] = __DIR__ . '/BrokenRenameFix.i18n.php';
+
+/**
+ * Special pages
+ */
+$wgAutoloadClasses['SpecialBrokenRenameFixController'] = __DIR__ . '/SpecialBrokenRenameFixController.class.php';
```
Grunny: The standard naming convention for our special page controllers is `<feature>SpecialController.class.php`.

lukaszkonieczny:
LGTM

PR #8021, opened 2015-08-06
CONCF-855 | ensure that article is in content namespace and add logging when $art…

hakubo:
…icle is not Article

I will handle removing it after 2 weeks if any one asks


extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -337,6 +337,22 @@ public function getArticle() {
 				$data['redirected'] = true;
 			}
 
+			//CONCF-855: $article is null sometimes, fix added
+			//I add logging as well to be sure that this not happen anymore
+			//TODO: Remove after 2 weeks
```
bognix: could you be more specific than 2 weeks? Ticket to follow up?

rogatty:
LGTM

vforge:
edit: LGTM

PR #8026, opened 2015-08-06
adding mechanism for switching between mysql and swagger based services

nmonterroso:
@Wikia/services-team 
adding some simple logic to switch between the mysql service and swagger service

### DIFF
--- a/lib/Wikia/src/Service/User/Preferences/PreferenceModule.php
+++ b/lib/Wikia/src/Service/User/Preferences/PreferenceModule.php
@@ -10,6 +10,8 @@ use Wikia\Persistence\User\Preferences\PreferencePersistenceModuleMySQL;
 use Wikia\Persistence\User\Preferences\PreferencePersistenceSwaggerService;
 
 class PreferenceModule implements Module {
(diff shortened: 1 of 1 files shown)


lib/Wikia/src/Service/User/Preferences/PreferenceModule.php
```diff
@@ -10,6 +10,8 @@
 use Wikia\Persistence\User\Preferences\PreferencePersistenceSwaggerService;
 
 class PreferenceModule implements Module {
+	const SWAGGER_SERVICE_RAMP_USAGE = 0;
```
ArturKlajnerok, line 13: Should we put that in the config ?

nmonterroso, line 13: I thought about that, but I think it'd be nice to have a config in MediaWiki as close to the thing that uses it since there are a million configs in the config repo. Also, this config should not be used by anything other than this class, so i don't think we need to introduce yet-another `$wg...` variable. It'd be confusing to have both `$wgPreferencesUseService` and (potentially) `$wgPreferencesUseSwaggerService` in the config repo.

PR #8027, opened 2015-08-07
SOC-1096 Send digest emails from the correct wikia

garthwebb:
https://wikia-inc.atlassian.net/browse/SOC-1096

### DIFF
--- a/extensions/wikia/FounderEmails/events/FounderEmailsCompleteDigestEvent.class.php
+++ b/extensions/wikia/FounderEmails/events/FounderEmailsCompleteDigestEvent.class.php
@@ -18,7 +18,10 @@ class FounderEmailsCompleteDigestEvent extends FounderEmailsEvent {
 
 	/**
 	 * Called from maintenance script only.  Send Digest emails for any founders with that preference enabled
-	 * @param array $events
+	 *
+	 * @param array $events This array is empty most of the time.  If the --wikiId flag is given to
+	 *                      the maintenance script however, that single wiki ID will be given as the only
+	 *                      array element.
 	 */
 	public function process ( Array $events ) {
 		global $wgTitle;
--- a/extensions/wikia/FounderEmails/events/FounderEmailsCompleteDigestEvent.class.php
+++ b/extensions/wikia/FounderEmails/events/FounderEmailsCompleteDigestEvent.class.php
@@ -1,7 +1,7 @@
 <?php
 
 class FounderEmailsCompleteDigestEvent extends FounderEmailsEvent {
-	const EMAIL_CONTROLLER = 'Email\Controller\FounderActivityDigest';
+	const EMAIL_CONTROLLER = Email\Controller\FounderActivityDigestController::class;
 
 	public function __construct( Array $data = array() ) {
 		parent::__construct( 'completeDigest' );
@@ -31,6 +31,7 @@ class FounderEmailsCompleteDigestEvent extends FounderEmailsEvent {
 		$wikiService = new WikiService();
 
 		foreach ( $cityList as $cityID ) {
+			Wikia::initAsyncRequest( $cityID );
 			$userIds = $wikiService->getWikiAdminIds( $cityID );
 			$emailParams = [
 				'wikiId' => $cityID,
--- a/extensions/wikia/FounderEmails/events/FounderEmailsCompleteDigestEvent.class.php
+++ b/extensions/wikia/FounderEmails/events/FounderEmailsCompleteDigestEvent.class.php
@@ -27,7 +27,11 @@ class FounderEmailsCompleteDigestEvent extends FounderEmailsEvent {
 		$wgTitle = Title::newMainPage();
 		$founderEmailObj = FounderEmails::getInstance();
 		// Get list of founders with digest mode turned on (defined in FounderEmailsEvent)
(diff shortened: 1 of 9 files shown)


extensions/wikia/FounderEmails/events/FounderEmailsCompleteDigestEvent.class.php
```diff
@@ -27,10 +27,15 @@ public function process ( Array $events ) {
...
-		$cityList = $founderEmailObj->getFoundersWithPreference( 'founderemails-complete-digest' );
+		if ( empty( $events ) ) {
+			$cityList = $founderEmailObj->getFoundersWithPreference( 'founderemails-complete-digest' );
+		} else {
+			$cityList = $events;
+		}
 		$wikiService = new WikiService();
 
 		foreach ( $cityList as $cityID ) {
+			Wikia::initAsyncRequest( $cityID );
```
jsutterfield, line 41: Is it necessary to call `Wikia::initAsyncRequest` here? It looks like the email controller [takes care of pulling the correct wikiname form the wiki id](https://github.com/Wikia/app/blob/dev/extensions/wikia/Email/Controller/FounderDigestController.class.php#L25-L26)

The fix is that we weren't passing the wikiId [here](https://github.com/Wikia/app/pull/8027/files#diff-77c4dcf675364821f0c6836036116805R101), right?

extensions/wikia/FounderEmails/events/FounderEmailsDaysPassedEvent.class.php
```diff
@@ -32,14 +32,14 @@ public function process( Array $events ) {
 			if ( empty( $emailController ) ) {
 				continue;
 			}
+			Wikia::initAsyncRequest( $wikiId );
```
jsutterfield, line 35: Same question about the call to `Wikia::initAsyncRequest`, is it doing anything necessary? I think the logic in these email controllers doesn't rely on globals.

extensions/wikia/FounderEmails/events/FounderEmailsViewsDigestEvent.class.php
```diff
@@ -27,11 +27,16 @@ public function process ( Array $events ) {
 
 		$wgTitle = Title::newMainPage();
 		// Get list of founders with digest mode turned on
-		$cityList = $founderEmailObj->getFoundersWithPreference( 'founderemails-views-digest' );
+		if ( empty( $events ) ) {
+			$cityList = $founderEmailObj->getFoundersWithPreference( 'founderemails-complete-digest' );
+		} else {
+			$cityList = $events;
```
jsutterfield, line 35: The comment for this method says that `$events` should always be empty when this is called, why are we setting cityList to it?

extensions/wikia/FounderEmails/events/FounderEmailsCompleteDigestEvent.class.php
garthwebb, line 41: This fix isn't about explicit requests for a wiki ID, its for code that implicitly assumes it is running in the context of a particular wiki. For example, the two lines you identified won't help wfMessage correctly translate `{{SITENAME}}` when it appears in a string (the error happening in this ticket).  For things like this we need to set global variables.

extensions/wikia/FounderEmails/events/FounderEmailsDaysPassedEvent.class.php
garthwebb, line 35: I think we should be using this in _every_ task regardless of whether we know a place where its necessary.  Given how much our code relies on global state, its very likely that one of the following will be true:
- We miss/forgot a place where we've added code relying on this
- There is code that relies on this within some seldom used (or not used in our testing) conditional
- Someone adds code later on, not realizing it executes within a task and assuming implicit use of globals will be ok

In these situations we allow the introduction of possibly difficult to debug issues that can bite us long after we've forgotten about the code we've added.

extensions/wikia/FounderEmails/events/FounderEmailsViewsDigestEvent.class.php
garthwebb, line 35: @jsutterfield, oh I didn't notice that, I'll have to change the comment.  The issue is with the `maintenance.php` script in the root directory of this extension.  This script takes a `--wikiId` argument which is intended to make it run _only_ against that wiki.  For the events that do take an `$events` array, that array is populated by `maintenance.php` and the restriction to act on that wiki ID alone is enforced.  However the events that are not given an `$events` array gather their own data and run on every single wiki.

This means that if you run the maintenance script for these types of events with only one wiki ID given, you actually run it for all wikis (as I did when I was testing this).

extensions/wikia/FounderEmails/events/FounderEmailsCompleteDigestEvent.class.php
jsutterfield, line 41: Ah, okay. Makes sense :+1:

extensions/wikia/FounderEmails/events/FounderEmailsDaysPassedEvent.class.php
jsutterfield, line 35: Fair enough, I like making things easier for our future selves. We spend enough time inadvertently making things more difficult, so being cautious for the sake of future us sounds good.

jsutterfield:
LGTM!

PR #8028, opened 2015-08-07
ADEN-2265 Refresh ads on scroll event

drets:


extensions/wikia/AdEngine/js/AdContext.js
```diff
@@ -81,6 +81,11 @@ define('ext.wikia.adEngine.adContext', [
 			isProperCountry(instantGlobals.wgAdDriverHighImpactSlotCountries)
 		) ||  parseInt(qs.getVal('highimpactslot', '0'));
 
+		context.opts.enableScrollHandler = !!(
+			isProperCountry(instantGlobals.wgAdDriverScrollHandlerCountries) ||
+			parseInt(qs.getVal('scrollhandler', '0'))
```
nandy-andy: Once we merge https://github.com/Wikia/app/pull/8022 we most likely want to re-use `isUrlParamSet()`.

extensions/wikia/AdEngine/js/slot/scrollHandler.js
```diff
@@ -0,0 +1,49 @@
...
+            PREFOOTER_LEFT_BOXAD: false,
+            PREFOOTER_RIGHT_BOXAD: false
+        };
+
+    function init() {
+        if (adContext.getContext().opts.enableScrollHandler) {
+            win.addEventListener('scroll', function () {
+                log('Scroll event listener has been added', 'debug', logGroup);
+                for (var slotName in isRefreshed) {
+                    refreshSlot(slotName);
```
nandy-andy: If you're using `for` loop to iterate through an object it's a good practice to use:

``` js
if (isRefreshed.hasOwnProperty(slotName))
```

so it doesn't loop through primitive object's prototype.

drets: Thanks, I'll remember :wink:

nandy-andy: You're welcome :) It's a pain and in modern browsers you can just use `Object.keys()` to iterate through all object's properties. But it's not supported in IE8, though :(

nandy-andy:
LGTM :+1:

extensions/wikia/AdEngine/js/slot/scrollHandler.js
```diff
@@ -0,0 +1,51 @@
...
+
+    var logGroup = 'ext.wikia.adEngine.slot.scrollHandler',
+        isRefreshed = {
+            PREFOOTER_LEFT_BOXAD: false,
+            PREFOOTER_RIGHT_BOXAD: false
+        };
+
+    function init() {
+        if (adContext.getContext().opts.enableScrollHandler) {
+            win.addEventListener('scroll', function () {
```
nandy-andy: We could also add some throttling here, so it fires the function every `200ms`. It would make performance of our page better. What do you think?

extensions/wikia/AdEngine/js/slot/scrollHandler.js
```diff
@@ -0,0 +1,51 @@
...
+                for (var slotName in isRefreshed) {
+                    if (isRefreshed.hasOwnProperty(slotName)) {
+                        refreshSlot(slotName);
+                    }
+                }
+            });
+        }
+    }
+
+    function refreshSlot(slotName) {
```
nandy-andy, line 29: We can count here the time a slot was refreshed and pass it to key-value which @nefretete mentioned. What do you think about this idea?

extensions/wikia/AdEngine/js/slot/scrollHandler.js
drets: :+1:

nandy-andy:
LGTM :+1:

PR #8036, opened 2015-08-07
CE-2419 Fix fatal error

kamilkoterba:
Fix fatal error Argument 1 passed to InsightsHelper::getTitleLink() must be an instance of Title, null given

https://wikia-inc.atlassian.net/browse/CE-2419

@Wikia/community-engineering


Grunny:
LGTM.

PR #8040, opened 2015-08-07
ADEN-2297 Get mappedVerticalName in a safer way

nandy-andy:
The `adsContext` may be cached on varnishes whereas changed JS code may come "fresh". Therefore, we need this safety fallback to the old version.


fraszczakszymon:
LGTM :+1:

PR #8053, opened 2015-08-10
don't let core-mw save options that conflict with preferences

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-588

Core functionality is currently overwriting preferences in the case that a user's `mOptions` contains a preference key. This change filters those out so only preferences should be saving them now.

### DIFF
--- a/includes/User.php
+++ b/includes/User.php
@@ -36,7 +36,7 @@ define( 'USER_TOKEN_LENGTH', 32 );
  * Int Serialized record version.
  * @ingroup Constants
  */
-define( 'MW_USER_VERSION', 10 );
+define( 'MW_USER_VERSION', 11 );
 
 /**
  * String Some punctuation to prevent editing from broken text-mangling proxies.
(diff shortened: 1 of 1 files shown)


includes/User.php
```diff
@@ -36,7 +36,7 @@
  * Int Serialized record version.
  * @ingroup Constants
  */
-define( 'MW_USER_VERSION', 10 );
+define( 'MW_USER_VERSION', 11 );
```
harnash, line 39: This can/will force reload from DB all user data. Do we need this since there is no change in user internal representation?

harnash:
Except one comment - LGTM

includes/User.php
nmonterroso, line 39: We've done this a few times with the preferences ramp (the value when we started was `9`), so loading the data from the DB isn't really a problem. I'm bumping the version here to force a sync between what's in the DB and what's in memcache, so that any issues with this change are immediately visible instead of delayed to when the memcache key expires in the cache.

harnash, line 39: OK. Sounds good.

PR #8055, opened 2015-08-10
Concf 847 - send article ID in API response

bognix:
/cc @rogatty @hakubo @vforge 

related PR: https://github.com/Wikia/mercury/pull/1195


tomasznapieralski:
LGTM

rogatty:
LGTM but can it break some API tests?
/cc @inez

hakubo:
LGTM

inez:
LGTM - however in this case I would add "id" to both videos and images (instead of only images) to keep the response more consistent.

@rogatty Currently we don't have tests for this API. We used to have them in VisualEditor-old but they were not migrated during upstream sync. It would be good to do it within next upstream sync - coming later this year.

PR #8059, opened 2015-08-10
SERVICES-587 Enable "internal" password reset emails with The Schwartz Token

drsnyder:
Enable "internal" password reset emails with The Schwartz Token. Currently, if you are not logged in or routing this request within the current request you can't send the password reset email.

We need to be able to call this from helios for [SERVICES-276](https://wikia-inc.atlassian.net/browse/SERVICES-276) so I'm enabling the the external invocation of this endpoint using the token convention.

Hopefully this will be a short lived solution. Once we move to password reset links this will no longer be necessary.

https://wikia-inc.atlassian.net/browse/SERVICES-587

/cc @Wikia/services-team @michalroszka


nmonterroso:
How weird that `return`, which returns `null` if you assign the value, acts as "yes you can access this controller" rather than true/false O.o

But I guess LGTM

drsnyder:
> How weird that return, which returns null if you assign the value, acts as "yes you can access this controller" rather than true/false O.o

Yeah, it's like an `assert`.

PR #8060, opened 2015-08-10
allow mOptions to reload properly

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-592

fixes an issue that was surfaced with the transition to the user preferences service where `clearInstanceCache()` clears a user's `mOptions`, subsequent `loadOptions()` would not properly fill out the user's `mOptions` since `mOptionsLoaded` was set to `true`.

### DIFF
--- a/includes/User.php
+++ b/includes/User.php
@@ -1357,6 +1357,8 @@ class User {
 		$this->mEffectiveGroups = null;
 		$this->mImplicitGroups = null;
 		$this->mOptions = null;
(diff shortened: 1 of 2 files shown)


drsnyder:
lftm

nmonterroso:
what does the "f" stand for in lftm? :)

drsnyder:
f as in fine

PR #8063, opened 2015-08-10
MAIN-5118: restrict import to NS_MEDIAWIKI

owend:
We may want a feature/global flag for this behavior at some point...
@Grunny @lgarczewski @michalroszka


lgarczewski:
LGTM.

Grunny:
:+1:

PR #8065, opened 2015-08-11
INT-124: Adding message string for mobile wikia-in-your-lang

mari-mcmurtrie:
We have to display different text for wikia-in-your-lang in mobile. Hence, adding text for mobile to the existing architecture.


mari-mcmurtrie:
@lizlux When you have time, please take a look. Thank you.

lizlux:
I was envisioning a separate API resource for mobile vs desktop, but this is probably fine. Pinging @gbenson-wikia for philosophical input.

xaroth8088:
For the time being, this feels like the right approach to me.  Closer to the end of the year we hope to tackle i18n management more comprehensively, so as long as new strings are added to existing translation methods (i.e. no new methods for getting translated string bundles are added), we should be in good shape from that perspective.

gbenson-wikia:
For the time being, this feels like the right approach to me.  Closer to the end of the year we hope to tackle i18n management more comprehensively, so as long as new strings are added to existing translation methods (i.e. no new methods for getting translated string bundles are added), we should be in good shape from that perspective.

(Sorry - was logged into my personal account for my previous comment)

lizlux:
Cool, thanks @gbenson-wikia!

PR #8066, opened 2015-08-11
disallow core CSS pages in verbatim tag

owend:


extensions/3rdparty/Verbatim/Verbatim.php
```diff
@@ -18,5 +18,14 @@ function wfVerbatimExtension( $parser ) {
 
 // The callback function for converting the input text to HTML output
 function renderVerbatim( $input ) {
+	global $wgEditInterfaceWhitelist;
+
+	// Begin wikia change
+	if (is_array($wgEditInterfaceWhitelist) and in_array(trim($input), $wgEditInterfaceWhitelist)) {
```
wladekb: just in case I'd add !empty(...) so if the variable is not defined warning won't be emitted

Grunny: Probably need to check `ucfirst` on the input as MW messages are case insensitive on the first letter.

wladekb:
lgtm apart from a small comment that might be addressed later and shouldn't be treated as a blocker

Grunny:
LGTM once the `ucfirst` check is in place (otherwise the check can be bypassed).

wladekb:
@Grunny good catch :+1:

PR #8068, opened 2015-08-11
MAIN-5121: check permissions and whitelist for NS_MEDIAWIKI

owend:
Tested on devbox, but it can definitely use more eyeballs and testing.

### DIFF
--- a/includes/wikia/Wikia.php
+++ b/includes/wikia/Wikia.php
@@ -30,6 +30,7 @@ $wgHooks['ContributionsToolLinks']   [] = 'Wikia::onContributionsToolLinks';
 $wgHooks['AjaxAddScript']            [] = 'Wikia::onAjaxAddScript';
 $wgHooks['TitleGetSquidURLs']        [] = 'Wikia::onTitleGetSquidURLs';
 $wgHooks['ImportHandlePageXMLTag']   [] = 'Wikia::onImportHandlePageXMLTagFilter';
+$wgHooks['userCan']                  [] = 'Wikia::canEditInterfaceWhitelist';
 
 # changes in recentchanges (MultiLookup)
 $wgHooks['RecentChange_save']        [] = "Wikia::recentChangesSave";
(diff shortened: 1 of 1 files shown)


includes/wikia/Wikia.php
```diff
@@ -2248,6 +2249,30 @@ static function onImportHandlePageXMLTagFilter ( WikiImporter $importer, &$pageI
...
+	 * Restrict editinterface right to whitelist
+	 * set $result true to allow, false to deny, leave alone means don't care
+	 * usually return true to allow processing other hooks
+	 * return false stops permissions processing and we are totally decided (nothing later can override)
+	 */
+	static function canEditInterfaceWhitelist (&$title, &$wgUser, $action, &$result) {
+		global $wgEditInterfaceWhitelist;
+
+		// List the conditions we don't care about for early exit
+		if ( $action == "read" || $title->getNamespace() != NS_MEDIAWIKI || empty( $wgEditInterfaceWhitelist )) {
```
nandy-andy, line 2261: How about `$title === null`? Should we add it also to this condition?

includes/wikia/Wikia.php
```diff
@@ -2248,6 +2249,30 @@ static function onImportHandlePageXMLTagFilter ( WikiImporter $importer, &$pageI
...
+			return true;
+		}
+
+		// In this NS, editinterface applies only to white listed pages and staff users
+		if (in_array($title->getDBKey(), $wgEditInterfaceWhitelist) || in_array('staff', $wgUser->getGroups())) {
```
nandy-andy, line 2266: What do you think about making `$wgEditInterfaceWhitelist` case-insensitive and using `mb_lowercase()` in the first part of the condition?

includes/wikia/Wikia.php
lgarczewski, line 2261: I believe if title is null we wouldn't get this far. :)

includes/wikia/Wikia.php
lgarczewski, line 2266: The whitelist is only going to have a couple entries, and we want it to be tight. Case-insensitive and other transformations make this more complicated than it needs to be.

lgarczewski:
LGTM. Merging and will test on verify.

PR #8039, opened 2015-08-07
PLATFORM-1385: remove postback token validation and legacy mail library

owend:
- remove old lib/vendor/Mail (we use Mail2 now) 
- remove wikiadb mail implementation
- remove old sendgrid postback controller

@garthwebb

PR #8070, opened 2015-08-11
Fix fatal errors in unit tests

Grunny:
We switched to using Mail2, but it still had references to the old
code path. Follow up to https://github.com/Wikia/app/pull/8039

/cc @nandy-andy @owend


nandy-andy:
LGTM :+1: Thank you!

PR #8071, opened 2015-08-11
Changes from release-319

ghost:
@lgarczewski @owend @Grunny @ludwikkazmierczak

These are changes made during the P1 fight, I am porting them to release-320 and dev so keep the changes in the following releases until the crisis is gone.

@garthwebb - you merge some changes to release-319 and dev, but they did not make its way to release-320 - I am hoping I am doing you a friendly favor now and nothing will break :)


lgarczewski:
LGTM

PR #8073, opened 2015-08-11
CE-2367 infobox migration message in editor only for the insights list

MariuszCz:
CE-2367 infobox migration message in editor only for the insights list templates

Please merge with config: https://github.com/Wikia/config/pull/1237


lukaszkonieczny:
lgtm

extensions/wikia/PortableInfobox/services/Helpers/PortableInfoboxClassification.php
```diff
@@ -0,0 +1,42 @@
...
+			return false;
+		}
+
+		$titleNeedle = 'infobox';
+		if ( strripos( $titleText, $titleNeedle ) !== false ) {
+			$portableInfoboxNeedle = '<infobox';
+
+			// If a portable infobox markup was found
+			// it means that the template doesn't have a non-portable infobox
+			return strpos( $contentText, $portableInfoboxNeedle ) === false;
```
dianafa, line 31: It's not very important at all, but I would rephrase this comment to:
`if a portable infobox was not found it means that the template has a non-portable one` because this is what we check here.

extensions/wikia/PortableInfobox/services/Helpers/PortableInfoboxClassification.php
```diff
@@ -0,0 +1,42 @@
...
+	 *
+	 * @param string $titleText
+	 * @param string $contentText
+	 * @return bool
+	 */
+	public static function isTitleWithNonportableInfobox( $titleText, $contentText ) {
+		// ignore docs pages
+		if ( stripos( $titleText, '/doc' ) ) {
+			return false;
+		}
```
dianafa, line 23: What about user pages, sandboxes etc ?

dianafa:
LGTM

extensions/wikia/PortableInfobox/services/Helpers/PortableInfoboxClassification.php
lukaszkonieczny, line 23: I think we should leave it on sandboxes for testing purposes.

lukaszkonieczny:
Thanks @dianafa !

PR #8075, opened 2015-08-11
MAIN-5121: Allow sysops to edit whitelisted NS_MEDIAWIKI pages.

ghost:
The code works fine. I have just made it simpler. The real change is in the config: https://github.com/Wikia/config/pull/1230 :-)

https://wikia-inc.atlassian.net/browse/MAIN-5121

@Grunny @lgarczewski @owend

### DIFF
--- a/includes/wikia/Wikia.php
+++ b/includes/wikia/Wikia.php
@@ -2264,12 +2264,10 @@ class Wikia {
 
 		// In this NS, editinterface applies only to white listed pages and staff users
 		if (in_array($title->getDBKey(), $wgEditInterfaceWhitelist) || in_array('staff', $wgUser->getGroups())) {
-			$result = $wgUser->isAllowed('editinterface');
-			return $result;
-		} else {
-			$result = false;
-			return false;
+			return $wgUser->isAllowed('editinterface');
 		}
+		
+		return false;
 	}
 
 	/**


includes/wikia/Wikia.php
```diff
@@ -2264,12 +2264,10 @@ static function canEditInterfaceWhitelist (&$title, &$wgUser, $action, &$result)
 
 		// In this NS, editinterface applies only to white listed pages and staff users
 		if (in_array($title->getDBKey(), $wgEditInterfaceWhitelist) || in_array('staff', $wgUser->getGroups())) {
```
adamkarminski, line 2266: Maybe I've missed something, but is the second check actually necessary (given that it is an OR condition)?

lgarczewski:
LGTM

includes/wikia/Wikia.php
ghost, line 2266: Yes, it makes staff users going through the whitelist and edit all pages if they have rights (which they have).

At least I believe this was the author's intention. At the first glance I wanted to remove it, too. :)

adamkarminski, line 2266: Got it, makes sense if you treat `||` as `UNLESS` not `OR` :)

owend:
According to the docs, you're supposed to set both $result and return value in this hook.  However, since we are returning false, the $result doesn't actually matter, because MW.  But that's why the code seemed redundant...

PR #8081, opened 2015-08-12
CE-2429: Migrate templates after adding new flag via Special:Flags

lukaszkonieczny:


extensions/wikia/Flags/tasks/FlagsExtractTemplatesTask.php
```diff
@@ -118,5 +118,13 @@ public function extractTemplatesFromPage( $pageId, Array $flagTypesToExtract ) {
 		}
 	}
 
+	public function migrateTemplatesToFlags( $template ) {
+		global $IP;
 
+		$wikiId = $this->getWikiId();
+
+		$cmd = "SERVER_ID={$wikiId} /usr/bin/php {$IP}/extensions/wikia/Flags/maintenance/MoveNotice.php --template='{$template}' --replaceTop --add --remove";
```
Grunny: Is `$template` potentially user controlled/defined? If so, this needs to be escaped with `wfEscapeShellArg` (same in MoveNotices.php).

Grunny:
:+1:

PR #8083, opened 2015-08-12
Fix window.ga being undefined when wgNoExternals is set

onei:
@Grunny For repro'ing, see http://runescape.wikia.com/wiki/RuneScape:Sandbox?noexternals=1 (check the js console). Let me know if I need to created/link to a ticket in jira :)


harnash:
LGTM

PR #8085, opened 2015-08-12
MAIN-5141 Tasks don't often set $wgTitle.  Return early if this isn't set

garthwebb:
This was causing an uncatchable fatal for tasks.

https://wikia-inc.atlassian.net/browse/MAIN-5141


jsutterfield:
LGTM!

PR #8086, opened 2015-08-12
MAIN-5131: load global groups early without breaking add/remove groups

owend:
FIxed 3 bugs
- possibility of leaving stale data in memcache if the db query fails
- master/slave race condition during add/remove process
- and the original bug: not able to edit local mediawiki namespace because global groups were null

@macbre @michalroszka @Grunny 

The fix is basically the same one I made before, I just added isCentralWiki() checks to add/remove group hook handlers so that you can't accidentally remove global groups from a local wiki.  During testing, I found the other bugs and fixed them also.


Grunny:
:+1:

PR #8087, opened 2015-08-13
MAIN-5141 Tasks don't often set $wgTitle.  Return early if this isn't…

harnash:
… set.
@garthwebb 
(cherrypicking to fix problems on prod)


wladekb:
lgtm

PR #8091, opened 2015-08-13
CE-2452

adamkarminski:
Description to be written, sorry :(


extensions/wikia/ContentReview/ContentReview.hooks.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace Wikia\ContentReview;
+
+class Hooks {
+
+	public static function onGetRailModuleList( Array &$railModuleList ) {
+		global $wgTitle, $wgUser;
+
+		if ( $wgTitle->getNamespace() === NS_MEDIAWIKI && $wgTitle->userCan( 'edit', $wgUser ) ) {
```
Grunny: It's not that important for our code, but at least as a MW standard, it's [preferable](https://github.com/Wikia/app/blob/fae57ff57bdc4b4f8e/includes/Title.php#L947-L949) to use `$wgTitle->inNamespace( NS_MEDIAWIKI );`.

lukaszkonieczny:
lgtm

PR #8095, opened 2015-08-13
return 404 if the user page does not exist

owend:
MAIN-4970 was opened to investigate an increase in "soft 404" responses.  These are pages that do not actually have any content, but we return 200 anyway.  A few years ago, this hook was added to UPPV3 to return a 200 if the global profile masthead exists.  This was fine at the time, but it's bad for SEO now. 

There's no visible change for the user, just the status code and the caching behavior.


tedgill:
:shipit:

PR #8097, opened 2015-08-13
INT-158: For Japanese Wikia on Mercury, no edit/add photo when anonymous edit is disabled.

mari-mcmurtrie:
If a Japanese Wikia disabled anonymous edit, then we do not want to even show add photo/edit icons. For Mercury to know if a wikia disabled anonymous edit, it has to get $wgDisableAnonymousEditing from back-end. 

Anonymous edit can be disabled while someone is in the middle of adding a photo; hence, we still have to check if anonymous upload is allowed for JP wikia. checkPermissionsForJA() checks the condition.


mari-mcmurtrie:
@inez will you please take a look when you have time? Thanks.

extensions/VisualEditor/wikia/ApiAddMediaTemporary.php
```diff
@@ -40,11 +40,13 @@ private function executeImage() {
 				$this->mRequest->getUpload( 'file' )
 			);
 
-			// If wiki is Japanese content, then we do not check permissions. INT-102
+			// If wiki is Japanese content, then we check if anonyous edit is allowed. INT-158
 			// Enable unauthorized save for Curated Main Page Editor
 			// if $wgEnableCuratedContentUnauthorizedSave not empty (CONCF-741)
 			// Ticket for removal wg check: CONCF-978
-			if ( $wgContLanguageCode !== 'ja' && empty( $wgEnableCuratedContentUnauthorizedSave ) ) {
+			if ( $wgContLanguageCode === 'ja') {
```
inez: Missing space after 'ja'

extensions/VisualEditor/wikia/ApiAddMediaTemporary.php
```diff
@@ -40,11 +40,13 @@ private function executeImage() {
...
+				$this->checkPermissionsForJA();
+			} else if (empty( $wgEnableCuratedContentUnauthorizedSave ) ) {
```
inez: Missing space before empty

extensions/VisualEditor/wikia/ApiAddMediaTemporary.php
```diff
@@ -154,6 +156,14 @@ protected function checkPermissions() {
 		}
 	}
 
+	protected function checkPermissionsForJA() {
```
inez: In case of wikis with anonymous editing disabled wouldn't it be enough to call checkPermissions method? This code in checkPermissions `$this->mUpload->verifyTitlePermissions( $this->mUser ) !== true` perhaps should be already taking `$wgDisableAnonymousEditing` into consideration - if it's not then maybe it should be fixed.

mari-mcmurtrie: Upload is not allowed at all if anonymous. Existing checkPermissions() checks that. For Japan, if $wgDisableAnonymousEditing = true, then we can call checkPermissions()  or simply return error. I thought just returning error is a lot cheaper than going down into checkPermissions(). In addition, we are going to add login status to Mercury soon that we have to rewrite the logic here anyways. Having a completely separated function for 'JP' (checkPermissionsForJA()) makes the near-future code change easier, don't you think? We just delete this checkPermissionsForJA() function. 
I keep this as-is.

mari-mcmurtrie:
Thanks @inez for CRing it!!

extensions/VisualEditor/wikia/ApiAddMediaTemporary.php
inez: Well, let's put performance aside for a second.
If you can write logic based on generic variables such as wgdisableanonymousediting then I think you should - it's cleaner and cheaper to maintain than adding more of a lang specific logic.

Method checkPermissionsForJA might be faster than checkPermissions but given the context - how often it is going to be called and in what situation - I don't see how this possibly could matter and be worth introducing more complex code.

When it comes to the future change to this code - I don't know what those changes exactly gonna be.

mari-mcmurtrie: Again, Japan is special case that we allow anonymous file upload. So even if I follow your suggestion, then I still have to do: if (local is japan) then allow anonymous upload if not wgdisableanonymousediting (pseudo code). 

Rather than if.. statement woven into the existing function ( checkPermissions ), it is a lot clear to have a separate function, in my opinion. But you think opposite, that is interesting.

inez: You could just have `if(local is japan && !wgdisableanonymouseditibg) { ... } else { checkPermissions(); }`

Btw. In general I like idea of small single responsibility functions/methods, however having value of having value of config setting (ja) in function name is a big red flag to me.

mari-mcmurtrie: @owend how do you suggest about checking wgdisableanonymousediting for Japan specific behavior?

For all local except Japan, we do not allow anonymous file upload no matter what. For Japan, we allow it unless wgdisableanonymousediting = true.

owend: I would suggest moving the business logic into checkPermissions (even though it's called in 2 places) rather than create a new checkPermissionsForJA() function.  the test for if ( JA and Upload and Mobile) can all be in checkPermissions().  And then maybe a unit test to ensure that the function works in those environments.   It looks like checkPermissions() is called in two places, and this logic only applies to one of them, but it's a protected function so it can have a parameter true/false for the different behavior, which makes it more explicit.  

Otherwise, having 2 tests before calling check permissions (as Inez suggested) is also fine.  We have a lot more complicated tests elsewhere for various languages + application states.

mari-mcmurtrie:
@inez @owend Thanks for reviewing it. I changed code. 
... I just heard Inez is on vacation. @Grunny, will you please code review this? Thank you!

Grunny:
I'd consider moving the logic inside `checkPermissions` (even though it's called twice, so you may need an extra parameter), but in general LGTM.

mari-mcmurtrie:
Thanks for reviewing @Grunny. if ( $wgContLanguageCode === 'ja' ) condition will be removed in near future as soon as Mercury has login for all. So I keep it as is.  I add note for it.

PR #8098, opened 2015-08-13
Provide a hook for Helios to determine if a user is blocked

drsnyder:
This PR provides a hook (via Nirvana) that will allow Helios to determine if a user is blocked without us having to port groups/rights, and the IP blocking logic to Helios (now). The controller method is being added under the Helios extension in MW and will be protected by the `secret` token.

I'm also introducing an "auth" service that will provide the core functionality given a username.

https://wikia-inc.atlassian.net/browse/SERVICES-595

TODO:
- [x] Complete the implementation of `isUsernameBlocked`
- [x] Add more controller and service testing

/cc @Wikia/services-team


extensions/wikia/Helios/HelperController.class.php
```diff
@@ -127,4 +130,49 @@ public function sendConfirmationEmail()
...
+		$username = $this->getVal( 'username', null );
+		if ( !isset( $username ) ) {
+			$this->response->setVal( 'message', $failureMessage );
+			$this->response->setCode( \WikiaResponse::RESPONSE_CODE_BAD_REQUEST );
+		}
+
+		return $username;
+	}
+
+	protected function authenticateViaTheSchwartz() {
```
harnash, line 165: Would be nice to use this function in other functions in this class where we check The Schwartz Secret Token

drsnyder, line 165: I updated the other controllers in 3ead1d1.

drsnyder:
Any more comments?

extensions/wikia/Helios/HelperController.class.php
```diff
@@ -127,4 +131,57 @@ public function sendConfirmationEmail()
...
+		if ( $this->getVal( 'secret' ) != $this->wg->TheSchwartzSecretToken ) {
+			$this->response->setVal( 'message', 'invalid secret' );
+			$this->response->setCode( \WikiaResponse::RESPONSE_CODE_FORBIDDEN );
+			return false;
+		}
+
+		return true;
+	}
+
+	public function setAuthService( \Wikia\Service\User\Auth\AuthService $authService ) {
```
nmonterroso: Why do we use the fully qualified class name here but not below? seems like we could just have this be 

``` php
public ...(AuthService $authService) {
```

nmonterroso:
minor comment otw lgtm

extensions/wikia/Helios/HelperController.class.php
drsnyder: Good point. I must have been in a typing frenzy.

drsnyder: d0c6850

lib/Wikia/src/Service/User/Auth/AuthService.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace Wikia\Service\User\Auth;
+
+interface AuthService {
```
nmonterroso, line 5: Should this be `AuthorizationService`? `AuthService` to me says Authentication (but this class looks like authorization?), it's probably better to be explicit

drsnyder, line 5: `Auth` was chosen so we can safely do both authentication and authorization without offending the name police.

I'm ok either splitting or combining. Since there is only one method should we wait to split until we add more functionality?

nmonterroso, line 5: Ah ok, makes sense. In that case LGTM

ArturKlajnerok:
Lgtm :+1:

PR #8106, opened 2015-08-14
SOC-959 OpenGraph attributes for discussion posts

mklucsarits:
This change makes the main wiki image available to the wiki variables collection in Mercury. The image path is used for the OpenGraph 'og:image' tag.

https://wikia-inc.atlassian.net/browse/SOC-959

Related wikia/mercury changes: https://github.com/Wikia/mercury/pull/1218


extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -308,6 +308,11 @@ public function getWikiVariables() {
 			$wikiVariables['facebookAppId'] = $egFacebookAppId;
 		}
 
+		$wikiImages = (new WikiService())->getWikiImages( [$this->wg->CityId], 500 );
+		if ( is_array( $wikiImages ) && is_string( $wikiImages[$this->wg->CityId] ) ) {
```
garthwebb: Is it important that result be a string (or rather, is it possible it would be something else)?  I would use `!empty` here which will take care of checking if this is an array at whether there is anything at that array position at the same time, e.g.:

``` php
if ( !empty( $wikiImages[$this->wg->CityId] ) ) {
```

extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -308,6 +308,11 @@ public function getWikiVariables() {
 			$wikiVariables['facebookAppId'] = $egFacebookAppId;
 		}
 
+		$wikiImages = (new WikiService())->getWikiImages( [$this->wg->CityId], 500 );
```
garthwebb: Array definitions should have spaces, where indexing syntax does not (https://github.com/Wikia/guidelines/blob/master/PHP/CodingConventions.md#array-syntax)

extensions/wikia/MercuryApi/MercuryApiController.class.php
mklucsarits: I think PHP raises a Notice if you try to access `$wikiImages` as an array and it's not an array.

garthwebb: @mklucsarits -- PHP will definitely complain if you try to access a non-array as an array or a key of an array that doesn't exist, but not within `empty`.  I think its worth using empty here; applying this level of scrutiny may cause confusion for future coders about what is so special about this return value.  Check out: 

https://github.com/Wikia/app/blob/dev/extensions/wikia/WAM/controllers/api/WAMApiController.class.php#L92-L96

PR #8107, opened 2015-08-14
SOC-1070 Bail if there's no user title object and log an error

garthwebb:
https://wikia-inc.atlassian.net/browse/SOC-1070


jsutterfield:
LGTM!

PR #8108, opened 2015-08-16
MAIN-5118: Backporting https://phabricator.wikimedia.org/T66657

mszabo-wikia:
@owend @jsutterfield please review per Slack discussion


owend:
From what I remember this is a pre-requisite for another related fix to import/export code.  This patch itself looks fine to me.

PR #7977, opened 2015-07-31
ADEN-2166 SourcePoint integration

fraszczakszymon:

PR #8110, opened 2015-08-17
Aden 2165

jacekjursza:


extensions/wikia/AdEngine/ResourceLoaders/ResourceLoaderAdEngineBase.php
```diff
@@ -0,0 +1,166 @@
+<?php
+
+abstract class ResourceLoaderAdEngineBase extends ResourceLoaderModule {
+	const TTL_SCRIPTS = 1800; // half an hour -- cache scripts from ad.71i.de for this time
```
fraszczakszymon: I did the same mistake ;D
https://github.com/Wikia/app/pull/7977#discussion-diff-36091167

extensions/wikia/AdEngine/ResourceLoaders/ResourceLoaderAdEngineBase.php
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	/**
+	 * Get data from cache / generate new data if TTL is less than time()
+	 * @return array|bool|Object
+	 */
+	protected function getData() {
+		global $wgMemc;
+
+		if (static::$localCache) {
```
fraszczakszymon: Probably it will be broken when there is more than one resource needed in request? Let's say that we want to use SourcePoint and SevenOneMedia. After the first one is fetched then the second returns the same value due to static value of $localCache, am I wrong?

extensions/wikia/AdEngine/ResourceLoaders/ResourceLoaderAdEngineSourcePointModule.php
```diff
@@ -0,0 +1,59 @@
+<?php
+
+class ResourceLoaderAdEngineSourcePointModule extends ResourceLoaderAdEngineBase {
+	const TTL_SCRIPTS = 86400;   // one day for fresh scripts from SourcePoint
+	const TTL_GRACE = 3600; // one hour for old scripts (served if we fail to fetch fresh scripts)
+	const CACHE_BUSTER = 12;     // increase this any time the local files change
+	const REQUEST_TIMEOUT = 30;
```
fraszczakszymon, line 7: :gem: What do you think about decreasing this value? It could be also used in SevenOneMedia resources.

extensions/wikia/AdEngine/ResourceLoaders/ResourceLoaderAdEngineBase.php
jacekjursza: Should be safe (as it is late static binding - per "current" class):

php > class A { public static $v = "5"; public function pr(){ echo static::$v; } }
php > class B extends A { public static $v = "6"; }
php > $aa = new A();
php > $bb = new B();
php > $aa->pr();
5
php > $bb->pr();
6
php > $aa::$v = "7";
php > $aa->pr();
7
php > $bb->pr();
6
php > $bb::$v = "100";
php > $aa->pr();
7
php > $bb->pr();
100

fraszczakszymon: But you don't override $v value in child classes, try with:

``` php
<?php

abstract class A {
    public static $value = 'foo';

    public static function getValue() {
        return static::$value;
    }
}

class B extends A {
}

class C extends A {
}

echo A::getValue(); // foo
echo B::getValue(); // foo
echo C::getValue(); // foo

B::$value = 'bar';

echo A::getValue(); // bar
echo B::getValue(); // bar
echo C::getValue(); // bar

C::$value = 'baz';

echo A::getValue(); // baz
echo B::getValue(); // baz
echo C::getValue(); // baz
```

extensions/wikia/AdEngine/AdEngine2.setup.php
```diff
@@ -16,8 +16,11 @@
 $wgAutoloadClasses['AdEngine2Hooks'] =  __DIR__ . '/AdEngine2Hooks.class.php';
 $wgAutoloadClasses['AdEngine2PageTypeService'] = __DIR__ . '/AdEngine2PageTypeService.class.php';
 $wgAutoloadClasses['AdEngine2Service'] =  __DIR__ . '/AdEngine2Service.class.php';
-$wgAutoloadClasses['ResourceLoaderAdEngineSevenOneMediaModule'] = __DIR__ . '/ResourceLoaderAdEngineSevenOneMediaModule.php';
-$wgAutoloadClasses['ResourceLoaderAdEngineSourcePointModule'] = __DIR__ . '/ResourceLoaderAdEngineSourcePointModule.php';
+$wgAutoloadClasses['ResourceLoaderAdEngineBase'] = __DIR__ . '/ResourceLoaders/ResourceLoaderAdEngineBase.php';
+$wgAutoloadClasses['ResourceLoaderScript'] = __DIR__ . '/ResourceLoaders/ResourceLoaderScript.php';
+$wgAutoloadClasses['ResourceLoaderAdEngineSevenOneMediaModule'] = __DIR__ . '/ResourceLoaders/ResourceLoaderAdEngineSevenOneMediaModule.php';
+$wgAutoloadClasses['ResourceLoaderAdEngineSourcePointModule'] = __DIR__ . '/ResourceLoaders/ResourceLoaderAdEngineSourcePointModule.php';
```
nandy-andy, line 22: What do you think about renaming it to `ResourceLoaderAdEngineSourcePointRecoveryModule`? It's more descriptive and consistent with `ResourceLoaderAdEngineSourcePointDetectionModule`, isn't it?

extensions/wikia/AdEngine/ResourceLoaders/ResourceLoaderAdEngineBase.php
```diff
@@ -0,0 +1,180 @@
...
+			return $data;
+		}
+
+		$data = [
+			'script' => $generated,
+			'modTime' => $now,
+			'ttl' => $now + static::TTL_SCRIPTS,
+		];
+
+		if ( md5($data['script']) != md5($cached['script']) ) {
```
nandy-andy: It's a good practice in PHP to use `===` and `!==` operators but those methods generates always a string, so we're fine here IMHO.

extensions/wikia/AdEngine/tests/ResourceLoaderAdEngineTest.php
```diff
@@ -0,0 +1,120 @@
...
+			->setMethods( [ 'fetchRemoteScript', 'fetchLocalScript', 'getInlineScript' ] )
+			->getMock();
+		ResourceLoaderAdEngineSourcePointModule::$localCache[get_class($mock)] = null;
+		$mock->method( 'fetchRemoteScript' )->willReturn( self::REMOTE_SCRIPT_MOCK_CONTENT );
+		$mock->method( 'fetchLocalScript' )->willReturn( self::LOCAL_SCRIPT_MOCK_CONTENT );
+		$mock->method( 'getInlineScript' )->willReturn( self::INLINE_SCRIPT_MOCK_CONTENT );
+
+		$script = $mock->getScript( $this->resourceLoaderContext );
+
+		$sevenOneMediaScripts = [
```
nandy-andy: `$sevenOneMediaScripts` => `$sourcePointScripts`

nandy-andy:
Seems like minor changes left: re-naming of things and re-setting some values if you agree with the proposals. After addressing these comments: LGTM :+1:

extensions/wikia/AdEngine/AdEngine2.setup.php
nandy-andy, line 22: Discussed offline and it make sense to keep `ResourceLoaderAdEngineSourcePointModule` name since the module loads more than recovery scripts.

nandy-andy:
The notes addressed. Unit tests pass:

```
Time: 14.68 seconds, Memory: 120.00Mb
OK (59 tests, 210 assertions)
```

LGTM :+1:

PR #8111, opened 2015-08-17
ADEN-2335 SOI_SUB3SITE fix

nandy-andy:
After implementing our mapping of verticals the `undefined` issue happened for 71M integration. The code below fixes it and also we remove not-needed-anymore fallback to `wikiCategory`.


fraszczakszymon:
LGTM

PR #8114, opened 2015-08-17
ADEN-2337 Added Wikia vertical to mapping

fraszczakszymon:


nandy-andy:
LGTM :+1: Will it be a huge pain if we add a unit test for it? (as documentation :P)

PR #8102, opened 2015-08-14
SERVICES-463: Added invalidateToken() to Helios/Client.

harnash:
This is the implementation of token invalidation request to Helios. I've tried to follow the conventions but there are some parameters in existing calls that are not documented (i.e. `grant_type` etc.).

@Wikia/services-team

PR #8115, opened 2015-08-17
Services-464 add helios/client invalidate token call in UserLogout hook

sqreek:
Added call from UserLogout hook to invalidate token in helios.
this is next step after https://github.com/Wikia/app/pull/8102
//cc @Wikia/services-team


extensions/wikia/Helios/User.class.php
```diff
@@ -450,4 +460,9 @@ private static function clearCookie( $cookieName ) {
 			\WebResponse::NO_COOKIE_PREFIX
 		);
 	}
+
+	private static function getHeliosClient() {
+		global $wgHeliosBaseUri, $wgHeliosClientId, $wgHeliosClientSecret;
+		return new Client( $wgHeliosBaseUri, $wgHeliosClientId, $wgHeliosClientSecret );
+	}
```
drsnyder, line 473: Can we use [dependency injection for this](https://github.com/Wikia/app/tree/dev/lib/Wikia/src/DependencyInjection)? Then this could be something like `Injector::getInjector()->get(HeliosClient::class);`

extensions/wikia/Helios/User.class.php
```diff
@@ -226,6 +224,21 @@ public static function setAccessTokenCookie( $accessToken ) {
...
+
+	/**
+	 * Call helios invalidate token.
+	 */
+	private static function invalidateAccessTokenInHelios() {
+		$request = \RequestContext::getMain()->getRequest();
+		$heliosClient = self::getHeliosClient();
+		$heliosClient->invalidateToken( self::getAccessToken( $request ) );
+	}
```
drsnyder, line 243: I think we should put the login/logout and session management in the [auth service](https://github.com/Wikia/app/blob/dev/lib/Wikia/src/Service/User/Auth/AuthService.php). As it is here it's difficult to test. See also the comment below.

sqreek:
ping @drsnyder @nmonterroso @harnash 
Dependency Injection documentation was incorrect (had missing build()), but In test's this aproach was override default Injector (it's static), so I'm using only container to do this (added this in this PR) and updated documentation. 

We can consider using dependency injection in this way (using container instead helper class) or refactor DI. Also we don't need to use DI in this place (Helios\UserTest) refactor it and leave over.

lib/Wikia/src/Service/Helios/HeliosClient.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+namespace Wikia\Service\Helios;
+
+interface HeliosClient {
```
harnash, line 5: How about AuthClient? So we can get rid of Helios connotation?

lib/Wikia/src/Service/Helios/HeliosClientImpl.php
```diff
@@ -8,15 +7,26 @@
  *
  * This is a naive implementation.
  */
-class Client
+class HeliosClientImpl implements HeliosClient
```
harnash, line 10: If we change interface name we could just name this class as `HeliosClient` and it will be in consistency with `AuthService` etc.

harnash:
No more remarks.

sqreek:
We can use HeliosClientImpl as implementation of HeliosClient used in AuthClient, or we can use HeliosClient as implementation of AuthClient.

First aproach seems a little more confusing (naming) but, in my opinion, is more valid becuse we can have different implementation of HeliosClient (using HeliosClient via HttpRequest, or Swagger for example)

PR #8116, opened 2015-08-17
CE-2454: Add posibility to test js changes by admins

lukaszkonieczny:


extensions/wikia/ContentReview/controllers/ContentReviewApiController.class.php
```diff
@@ -87,6 +128,18 @@ public function getLatestReviewedRevision() {
 		}
 	}
 
+	public function showTestModeNotificaion() {
+		$notification = wfMessage( 'content-review-test-mode-enabled' )->escaped();
+		$notification.= Xml::element(
+			'a',
+			[ 'id' => 'content-review-test-mode-disable', 'href' => '#' ],
+			wfMessage( 'content-review-test-mode-disable' )->escaped(),
+			false
```
Grunny: You shouldn't need to set the `$allowShortTag` option as you have content in your tag.

extensions/wikia/ContentReview/controllers/ContentReviewApiController.class.php
```diff
@@ -87,6 +128,18 @@ public function getLatestReviewedRevision() {
 		}
 	}
 
+	public function showTestModeNotificaion() {
+		$notification = wfMessage( 'content-review-test-mode-enabled' )->escaped();
+		$notification.= Xml::element(
+			'a',
+			[ 'id' => 'content-review-test-mode-disable', 'href' => '#' ],
+			wfMessage( 'content-review-test-mode-disable' )->escaped(),
```
Grunny: The contents of `Xml::element` are [escaped](https://github.com/Wikia/app/blob/9c3dabe4a3dd/includes/Xml.php#L31) so this can be `->plain()` so it isn't double escaped.

PR #8117, opened 2015-08-17
ADEN-2339 Fix floor adhesion ad unit on Mercury

nandy-andy:
`$.exists()` does not exist in regular jQuery - let's not use it in features we want to re-use in both skins.


vforge:
LGTM

jacekjursza:
$('LGTM')

PR #8119, opened 2015-08-17
SOC-1099 Fixed problem not getting views digest plus some cleanup

garthwebb:
https://wikia-inc.atlassian.net/browse/SOC-1099


extensions/wikia/Email/Controller/FounderDigestController.class.php
```diff
@@ -53,8 +53,8 @@ protected function assertValidWikiId() {
 	 * @throws \Email\Check
 	 */
 	protected function assertPageViewsSet() {
```
garthwebb, line 55: This logic was a little off; checking for an empty value but not a stringified zero value.  My assumption was that this was supposed to let zero pages views generate an email, but this won't work for integer zero values.  Cristina and I agree we should just skip all zero page view emails.

extensions/wikia/Email/Controller/FounderDigestController.class.php
```diff
@@ -126,7 +126,12 @@ public function assertCanEmail() {
 	 * @throws \Email\Check
 	 */
 	protected function assertSubscribedToCompleteDigest() {
```
garthwebb, line 128: Updated call here and two other places to the new get*Preference calls

extensions/wikia/FounderEmails/FounderEmails.class.php
```diff
@@ -102,7 +106,7 @@ public function processEvents( $eventType, $useMasterDb = false, $wikiId = null
 			}
 
 			$oEvent = FounderEmailsEvent::newFromType( $eventType );
-			$result = $oEvent->process( $aEventsData );
+			$oEvent->process( $aEventsData );
```
garthwebb, line 109: Nothing gets returned by `process`

extensions/wikia/FounderEmails/FounderEmails.class.php
```diff
@@ -4,8 +4,8 @@ class FounderEmails {
 	static private $instance = null;
```
garthwebb, line 4: Almost all clean up here, one exception marked (still pretty much clean up though).

extensions/wikia/FounderEmails/generate.php
```diff
@@ -1,164 +0,0 @@
-<?php
```
garthwebb, line 1: This is replaced by the Special:Email page.

includes/wikia/services/DataMartService.class.php
```diff
@@ -26,46 +26,44 @@ class DataMartService extends Service {
 	const TTL = 43200; // WikiaSQL results caching time (12 hours)
```
garthwebb, line 26: Ran stylize.php on this file.  Lots of spacing changes.  Only other changes here are removing wfProfile\* calls and a few variable renames to meet code standards.  I'd recommend adding `?w=1` to look at those.

mklucsarits:
LGTM

extensions/wikia/FounderEmails/events/FounderEmailsViewsDigestEvent.class.php
```diff
@@ -40,14 +38,22 @@ public function process ( Array $events ) {
...
+
+			// Don't bother sending this email if there are no page views for this period
+			if ( empty( $views ) ) {
+				continue;
+			}
+
 			$emailParams = [
-				'pageViews' => $founderEmailObj->getPageViews( $cityID )
+				'pageViews' => $views,
+				'wikiId' => $cityID,
```
garthwebb, line 50: This is the real bug of this PR; `wikiId` is required field but it wasn't being passed in.

jsutterfield:
LGTM

PR #8121, opened 2015-08-18
SOC-677 Start writing attributes to Attribute Service

jsutterfield:
During a call to `User->setGlobalAttribute()` we now write the attributes to both the `wikicities.user_attributes` table through MW as well as the `userattributes.user_attribute` table through the user attribute service. These writes will allow us to achieve data parity between the 2 databases and eventually flip the switch to just read/write from the user attribute service exclusively.

Ticket: https://wikia-inc.atlassian.net/browse/SOC-677


extensions/wikia/UserProfilePageV3/UserIdentityBox.class.php
```diff
@@ -358,7 +358,7 @@ public function saveUserData($data) {
 		}
 
 		if (true === $changed) {
-			$this->user->setGlobalAttribute(self::USER_EVER_EDITED_MASTHEAD, true);
+			$this->user->setGlobalFlag(self::USER_EVER_EDITED_MASTHEAD, true);
```
jsutterfield, line 361: `self::USER_EVER_EDITED_MASTHEAD` is actually a flag, not an attribute. The [call to get this value](https://github.com/Wikia/app/blob/dev/extensions/wikia/UserProfilePageV3/UserIdentityBox.class.php#L264-L266) uses $user->getGlobalFlag. This is just updating it to use the correct setter.

includes/User.php
```diff
@@ -4866,7 +4891,7 @@ protected function saveOptions() {
 
 		$this->loadOptions();
 		// wikia change
-		global $wgExternalSharedDB, $wgSharedDB, $wgGlobalUserProperties;
```
jsutterfield, line 4886: `$wgGlobalUserProperties` wasn't being used.

lib/Wikia/src/Persistence/User/Attributes/AttributePersistenceSwagger.php
```diff
@@ -120,16 +120,16 @@ private function getApi( $userId ) {
 	private function handleApiException( ApiException $e ) {
 		switch ( $e->getCode() ) {
 			case UnauthorizedException::CODE:
-				throw new UnauthorizedException();
+				throw new UnauthorizedException( $e->getResponseBody() );
```
jsutterfield, line 123: `$e->getResponseBody()` has the data from the response we're actually interested in. It includes the HTTP response code, as well as the message returned by the attribute service.

lib/Wikia/tests/Persistence/User/AttributePersistenceSwaggerTest.php
```diff
@@ -32,12 +32,10 @@ public function setUp() {
 		$this->apiProvider = $this->getMockBuilder( ApiProvider::class )
 			->setMethods( [ 'getAuthenticatedApi' ] )
 			->disableOriginalConstructor()
-			->disableAutoload()
```
jsutterfield, line 35: These `disableAutoload()` calls were causing problems with the unit tests. Classes which were mocked here couldn't be found by later tests. It seems to somehow relate to the DI library we started using. Turns out we don't actually need these calls though, so we're just removing them which fixes it.

includes/User.php
```diff
@@ -2702,6 +2708,25 @@ public function setLocalAttribute($attribute, $value, $cityId = null, $sep = '-'
...
+		$this->setAttributeInService($attribute, $value);
+	}
+
+	/**
+	 * Sets an attribute into the User Attribute Service
+	 *
+	 * @param $attributeName
+	 * @param $attributeValue
+	 */
+	private function setAttributeInService($attributeName, $attributeValue) {
```
mklucsarits, line 2720: Don't forget whitespace formatting in PHP Land (spaces inside parentheses). :)

lib/Wikia/src/Service/User/Attributes/AttributeKeyValueService.php
```diff
@@ -59,14 +71,18 @@ public function get( $userId ) {
...
 		}
 
 		return $attributeArray;
 	}
 
-	public function delete( $userId, $attribute ) {
+	/**
+	 * @param int $userId
+	 * @param Attribute $attribute
+	 * @return bool|true
```
mklucsarits: I don't think you need a pipe for two things that are basically the same, and also, this can return `false`, so you may as well just have `bool` here.

lib/Wikia/src/Service/User/Attributes/UserAttributes.php
```diff
@@ -64,38 +64,77 @@ private function loadAttributes( $userId ) {
...
+		$this->setAttributeInCache( $userId, $attribute );
+		$this->setAttributeInService( $userId, $attribute );
 	}
 
 	private function isAnonUser( $userId ) {
 		return $userId === 0;
 	}
 
+	private function attributeValueAlreadySet( $userId, Attribute $attribute ) {
+		return $this->getAttribute( $userId, $attribute->getName() ) == $attribute->getValue();
```
mklucsarits: Are there any "gotchas" here where an attribute would need to be set to `0` and this would return true if compared to `null`? In other words, should you use `===` here or are you sure it'll never be needed?

lib/Wikia/src/Service/User/Attributes/UserAttributes.php
```diff
@@ -64,38 +64,77 @@ private function loadAttributes( $userId ) {
...
+	/**
+	 * @param $userId
+	 * @param Attribute $attribute
+	 */
+	private function setAttributeInService( $userId, $attribute ) {
+		$this->attributeService->set( $userId, $attribute );
+	}
+
+	public function deleteAttribute( $userId, Attribute $attribute ) {
+		if ( $this->isAnonUser( $userId ) ) {
```
mklucsarits, line 115: What do you think about moving the anon user logic into `attributeNotSetForUser`, and return false in that method if user is anon? That way these two cases don't need to be handled separately.

lib/Wikia/src/Service/User/Attributes/AttributeKeyValueService.php
```diff
@@ -27,7 +28,13 @@ function __construct( AttributePersistence $persistenceAdapter ) {
 		$this->persistenceAdapter = $persistenceAdapter;
 	}
 
-	public function set( $userId, $attribute ) {
+	/**
+	 * @param int $userId
+	 * @param Attribute $attribute
+	 * @return true
```
garthwebb: Is it always `true` or is it `bool`?

lib/Wikia/src/Service/User/Attributes/UserAttributes.php
```diff
@@ -64,38 +64,77 @@ private function loadAttributes( $userId ) {
...
 
-		$this->loadAttributes( $userId );
-		$this->setAttributeInService( $userId, $attribute );
-		$this->setAttributeInCache( $userId, $attribute );
+		if ( $this->attributeValueAlreadySet( $userId, $attribute ) ) {
+			return;
+		}
 
+		$this->setAttributeInCache( $userId, $attribute );
+		$this->setAttributeInService( $userId, $attribute );
```
garthwebb, line 76: Any reason the order of the two lines above were switched?  It seems like you'd want to set it in the service first.  If the service call fails now, the cache will be invalid.  When flipped, if the cache call fails, the service is still the holder of truth.

lib/Wikia/src/Service/User/Attributes/UserAttributes.php
garthwebb: I'd agree `===` might be appropriate here.

includes/User.php
jsutterfield, line 2720: Normally I try and be a zealot when it comes to whitespace in PHP Land, but since this is core code which doesn't follow that convention (and I didn't want to run the stylize script on this 5k line file), I figured it would be better to be consistent here :)

lib/Wikia/src/Service/User/Attributes/UserAttributes.php
jsutterfield, line 115: Personally I like the explicitness of keeping the 2 separate. Having `attributeNotSetForUser` check if `isAnonUser` works, but they do feel distinct to me. Either you're anon, or you're a logged in user and we're checking if you have the attribute set. It feels more clear this way.

lib/Wikia/src/Service/User/Attributes/AttributeKeyValueService.php
jsutterfield: Before it was just true or exception, but I think there's a case where it should return false. I'll update.

lib/Wikia/src/Service/User/Attributes/UserAttributes.php
jsutterfield, line 76: Not that I can recall specifically. Setting the service first makes sense to me. I'll unswitch.

mklucsarits:
LGTM

garthwebb:
LGTM

PR #8122, opened 2015-08-18
SOC-1075 Make sure cache is cleared for the old user ID

garthwebb:
Previously, user data was loaded from cache to delete user data from cache.  This ridiculous loop meant that the very thing we were trying to clear was overwriting the new data we were trying to write.

Also quite a bit of code clean up was done while trying to understand what was going on in the rename code.

https://wikia-inc.atlassian.net/browse/SOC-1075

### DIFF
--- a/extensions/wikia/UserRenameTool/RenameUserProcess.class.php
+++ b/extensions/wikia/UserRenameTool/RenameUserProcess.class.php
@@ -1,7 +1,5 @@
 <?php
 
-use Wikia\Tasks\AsyncTaskList;
-
 class RenameUserProcess {
 
 	const RENAME_TAG = 'renamed_to';
@@ -18,7 +16,7 @@ class RenameUserProcess {
 	/*
 	 * Task definition format:
 	 *	 'table' => (string) table name
-	 *   'userid_column' => (string) column name with user ID (or null if none
+	 *   'userid_column' => (string) column name with user ID or null if none
 	 *   'username_column' => (string) column name with user name
 	 *   'conds' => (array) additional conditions for the query
 	 */
@@ -84,11 +82,9 @@ class RenameUserProcess {
 	private $mUserId = 0;
 	private $mFakeUserId = 0;
 	private $mPhalanxBlockId = 0;
-	private $mGlobalTask = null;
 	private $mRequestorId = 0;
 	private $mRequestorName = '';
 	private $mReason = null;
-	private $mRenameIP = false;
 	private $mNotifyUser;
 
 	private $mErrors = array();
@@ -113,13 +109,14 @@ class RenameUserProcess {
 	/**
 	 * Creates new rename user process
 	 *
-	 * @param $oldUsername string Old username
-	 * @param $newUsername string New username
-	 * @param $confirmed bool Has the user confirmed all the warnings he got?
-	 * @param $notifyUser bool Whether to notify the renamed user
+	 * @param string $oldUsername Old username
+	 * @param string $newUsername New username
+	 * @param bool $confirmed Has the user confirmed all the warnings he got?
+	 * @param string $reason
+	 * @param bool $notifyUser Whether to notify the renamed user
 	 */
 	public function __construct( $oldUsername, $newUsername, $confirmed = false, $reason = null, $notifyUser = true ) {
-		global $wgContLang, $wgUser;
+		global $wgUser;
 
 		// Save original request data
 		$this->mRequestData = new stdClass();
@@ -216,7 +213,7 @@ class RenameUserProcess {
 
 	protected function renameAccount() {
 		global $wgExternalSharedDB;
-		wfProfileIn( __METHOD__ );
+
 		$dbw = wfGetDb( DB_MASTER, array(), $wgExternalSharedDB );
 
 		$table = '`user`';
@@ -235,7 +232,9 @@ class RenameUserProcess {
 			if ( $affectedRows ) {
 				$dbw->commit();
 				$this->addLog( "Changed user {$this->mOldUsername} to {$this->mNewUsername} in {$wgExternalSharedDB}" );
-				wfProfileOut( __METHOD__ );
+
+				User::clearUserCache( $this->mUserId );
+
 				return true;
 			} else {
 				$this->addLog( "No changes in {$wgExternalSharedDB} for user {$this->mOldUsername}" );
@@ -244,7 +243,6 @@ class RenameUserProcess {
 			$this->addLog( "Table \"{$table}\" not found in {$wgExternalSharedDB}" );
 		}
 
-		wfProfileOut( __METHOD__ );
 		return false;
 	}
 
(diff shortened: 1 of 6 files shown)


extensions/wikia/UserRenameTool/RenameUserProcess.class.php
```diff
@@ -84,11 +82,9 @@ class RenameUserProcess {
 	private $mUserId = 0;
 	private $mFakeUserId = 0;
 	private $mPhalanxBlockId = 0;
-	private $mGlobalTask = null;
```
garthwebb, line 87: This was never set anywhere but here to `null`

extensions/wikia/UserRenameTool/RenameUserProcess.class.php
```diff
@@ -84,11 +82,9 @@ class RenameUserProcess {
...
 	private $mRequestorId = 0;
 	private $mRequestorName = '';
 	private $mReason = null;
-	private $mRenameIP = false;
```
garthwebb, line 91: Unused.

extensions/wikia/UserRenameTool/RenameUserProcess.class.php
```diff
@@ -235,7 +232,9 @@ protected function renameAccount() {
 			if ( $affectedRows ) {
 				$dbw->commit();
 				$this->addLog( "Changed user {$this->mOldUsername} to {$this->mNewUsername} in {$wgExternalSharedDB}" );
-				wfProfileOut( __METHOD__ );
+
+				User::clearUserCache( $this->mUserId );
```
garthwebb, line 236: This is the real bug of this PR; there is code a little later on in execution that calls `User::invalidateCache()` on a user object that has been constructed with the new username.  Every single function in the `User` class seems to call `User::load`, including this one, which will try to load the rest of the user data.  In this case it makes a direct DB query to get the user ID from the (new) user name.  Since we keep the user ID the same for the user, this new name now gets this old ID (where as the old name is set to a new row in the user table with a new ID).  As load continues, it tries to get the user data from memcached, by ID.  If its a hit, memcached then returns OLD DATA INCLUDING THE OLD USERNAME LIKE A COMPLETE DOPE.

Now the old username is mixing about with the new user name and eventually we get a `duplicate key` error when saving settings to the DB later on, which inexplicably saves the user object again, because its trying to write a row with the old user name and the old user ID, when there already is a row with that old username (the new record created at the beginning of the process).

This new method we're calling clears memcache without ANY SIDE EFFECTS.

extensions/wikia/UserRenameTool/RenameUserProcess.class.php
```diff
@@ -296,42 +292,37 @@ protected function setup() {
...
 			$this->addError( wfMessage( 'userrenametool-error-same-user' )->inContentLanguage()->text() );
-			wfProfileOut( __METHOD__ );
 			return false;
 		}
 
 		// validate new username and disable validation for old username
-		$olduser = User::newFromName( $oldTitle->getText(), false );
-		$newuser = User::newFromName( $newTitle->getText(), 'creatable' );
+		$oldUser = User::newFromName( $oldTitle->getText(), false );
+		$newUser = User::newFromName( $newTitle->getText(), 'creatable' );
```
garthwebb, line 312: Just renaming these variables to comply with our code style rules.

extensions/wikia/UserRenameTool/RenameUserProcess.class.php
```diff
@@ -408,17 +392,7 @@ protected function setup() {
 						$repeating = (
 							count( $nameTokens ) == 2 &&
 							$nameTokens[0] === self::RENAME_TAG &&
-							$nameTokens[1] === $newuser->getName()
-						);
-					}
-
```
garthwebb, line 414: The diff is a little funny here;  I really just removed the whole `if` block below.  Maybe the newuser -> newUser rename messed it up.  First I got rid of the `$processing = ...` section because that variable isn't used anywhere. Once I did that I realized `$statusTokens` was no longer used anywhere.  Thus the whole block became irrelevant.

extensions/wikia/UserRenameTool/RenameUserProcess.class.php
```diff
@@ -505,13 +472,17 @@ public function run() {
 
 		// Analyze status
 		if ( !$status ) {
-
-			$problems = array_merge( $this->mErrors, $this->mWarnings );
```
garthwebb, line 509: This variable is unused.

extensions/wikia/UserRenameTool/RenameUserProcess.class.php
```diff
@@ -597,7 +561,6 @@ private function doRun() {
 			$this->mFakeUserId = $fakeUser->getId();
 			$this->addLog( "Created fake user account for {$fakeUser->getName()} with ID {$this->mFakeUserId} and renameData '{$fakeUser->getGlobalAttribute( 'renameData', '')}'" );
 		} else {
-			$fakeUser = User::newFromId( $this->mFakeUserId );
```
garthwebb, line 600: The `$fakeUser` var is constructed and never used.

extensions/wikia/UserRenameTool/RenameUserProcess.class.php
```diff
@@ -699,14 +653,23 @@ public function updateLocal() {
 			// Determine all namespaces which need processing
 			$allowedNamespaces = array( NS_USER, NS_USER_TALK );
 			// Blogs extension
-			if ( defined( 'NS_BLOG_ARTICLE' ) )
+			if ( defined( 'NS_BLOG_ARTICLE' ) ) {
 				$allowedNamespaces = array_merge( $allowedNamespaces, array( NS_BLOG_ARTICLE, NS_BLOG_ARTICLE_TALK ) );
+			}
 			// NY User profile
```
garthwebb, line 659: Below I did three things.  First, I added brackets per coding conventions.  Second I got rid of `NS_USER_PROFILE` from the array because I don't see it defined anywhere.  Lastly I reformatted the long line to several lines.

extensions/wikia/UserRenameTool/RenameUserProcess.class.php
```diff
@@ -982,28 +943,19 @@ public function addLog( $text, $arg1 = null ) {
 	/**
 	 * Logs the message to main user-visible log
 	 *
+	 * @param string $action
 	 * @param $text string Log message
-	 * @param $arg1 mixed Multiple format parameters
 	 */
-	public function addMainLog( $action, $text, $arg1 = null ) {
```
garthwebb, line 988: I don't think we need we need 12 lines to explain why we don't need 5 lines

includes/User.php
```diff
@@ -429,7 +429,11 @@ public function saveToCache() {
 	}
 
 	private function getCacheKey() {
-		return wfMemcKey('user', 'id', $this->mId);
+		return self::getCacheKeyById( $this->mId );
+	}
+
```
garthwebb, line 434: Needed a static method to return the cache key so there could be a static method to clear the user cache.  Updated the instance method `getCacheKey` to use this static method to keep everything aligned.

extensions/wikia/UserRenameTool/SpecialRenameuser_body.php
```diff
@@ -37,86 +38,93 @@ public function execute( $par ) {
...
 		if ( !$wgUser->isAllowed( 'renameuser' ) ) {
-			wfProfileOut( __METHOD__ );
 			throw new PermissionsError( 'renameuser' );
 		}
 
 		// Get the request data
-		$oldusername = $wgRequest->getText( 'oldusername', $par );
-		$newusername = $wgRequest->getText( 'newusername' );
+		$oldUsername = $wgRequest->getText( 'oldusername', $par );
+		$newUsername = $wgRequest->getText( 'newusername' );
```
garthwebb, line 50: More naming convention updates.

kvas-damian:
LGTM

rafalkalinski:
LGTM

PR #8125, opened 2015-08-18
PLATFORM-1317: set cURL options key by key

macbre:
Use `curl_setopt` instead of `curl_setopt_array` to catch the option that has null in its value.

@michalroszka / @WikiaSparrow


ghost:
At last! :+1:

PR #8128, opened 2015-08-18
DAT-3082 show all params used in infoboxes

dianafa:
https://wikia-inc.atlassian.net/browse/DAT-3082


extensions/wikia/PortableInfobox/controllers/ApiQueryPortableInfobox.class.php
```diff
@@ -23,16 +23,27 @@ protected function runOnPageSet( ApiPageSet $pageSet ) {
...
 		foreach ( $articles as $id => $article ) {
-			$d = $article->getParserOutput()->getProperty( PortableInfoboxDataService::INFOBOXES_PROPERTY_NAME );
-			if ( is_array( $d ) ) {
+			$parsedInfoboxes = $article->getParserOutput()->getProperty( PortableInfoboxDataService::INFOBOXES_PROPERTY_NAME );
 
+			//if in template there are no infoboxes, skip the <includeonly> tags and parse again
+			if (!$parsedInfoboxes) {
+				$templateText = $article->fetchContent();
+				$parser = new Parser();
+				$templateText = $parser->getPreloadText($templateText, $article->getTitle(), new ParserOptions());
```
idradm: What if includeonly consists of more than just infobox, or no infobox at all?

extensions/wikia/PortableInfobox/controllers/ApiQueryPortableInfobox.class.php
```diff
@@ -23,16 +23,27 @@ protected function runOnPageSet( ApiPageSet $pageSet ) {
...
 		 */
 		foreach ( $articles as $id => $article ) {
-			$d = $article->getParserOutput()->getProperty( PortableInfoboxDataService::INFOBOXES_PROPERTY_NAME );
-			if ( is_array( $d ) ) {
+			$parsedInfoboxes = $article->getParserOutput()->getProperty( PortableInfoboxDataService::INFOBOXES_PROPERTY_NAME );
 
+			//if in template there are no infoboxes, skip the <includeonly> tags and parse again
+			if (!$parsedInfoboxes) {
+				$templateText = $article->fetchContent();
+				$parser = new Parser();
```
idradm: Do you need a new instance of parser on each article?

extensions/wikia/PortableInfobox/controllers/ApiQueryPortableInfobox.class.php
```diff
@@ -23,16 +23,27 @@ protected function runOnPageSet( ApiPageSet $pageSet ) {
...
+				$templateText = $parser->getPreloadText($templateText, $article->getTitle(), new ParserOptions());
+
+				PortableInfoboxParserTagController::getInstance()->render($templateText, $parser, $parser->getPreprocessor()->newFrame() );
+
+				$parsedInfoboxes = $parser->getOutput()->getProperty(PortableInfoboxDataService::INFOBOXES_PROPERTY_NAME);
+			}
+
+			if ( is_array( $parsedInfoboxes ) ) {
```
idradm: $d is not defined at all, and is used in further code

extensions/wikia/PortableInfobox/controllers/ApiQueryPortableInfobox.class.php
```diff
@@ -19,20 +19,38 @@ protected function runOnPageSet( ApiPageSet $pageSet ) {
...
+		$parser = new Parser();
+		$parserOptions = new ParserOptions();
+		$frame = $parser->getPreprocessor()->newFrame();
 		/**
 		 * @var Article $article
 		 */
 		foreach ( $articles as $id => $article ) {
-			$d = $article->getParserOutput()->getProperty( PortableInfoboxDataService::INFOBOXES_PROPERTY_NAME );
-			if ( is_array( $d ) ) {
+			$parsedInfoboxes = $article->getParserOutput()->getProperty( PortableInfoboxDataService::INFOBOXES_PROPERTY_NAME );
```
idradm: Could you extract infobox validation and reparsing part into separate method?

extensions/wikia/PortableInfobox/controllers/ApiQueryPortableInfobox.class.php
```diff
@@ -41,4 +59,36 @@ protected function runOnPageSet( ApiPageSet $pageSet ) {
...
 	}
+
+	/**
+	 * @desc From the template string with removed <includeonly> tags, creates an array of
+	 * strings containing only infoboxes. All template content which is not an infobox is removed.
+	 *
+	 * @param $text string Content of template which uses the <includeonly> tags
+	 * @return array of striped infoboxes ready to parse
+	 */
+	protected function processTemplate( $text ) {
```
idradm: Are you sure this preg_match_all("/<infobox.*>.*<\/infobox>/sU", $input_lines, $output_array); will be slower? If yes by how much? Maybe it's worth to use it in favor of readability?

idradm:
LG

extensions/wikia/PortableInfobox/services/PortableInfoboxDataService.class.php
```diff
@@ -63,4 +59,50 @@ public function getImages() {
...
+	 * @return mixed
+	 */
+	protected function getParsedInfoboxes( $article ) {
+		//on empty parser cache this should be regenerated, see WikiPage.php:2996
+		$parserOutput = $article->getParserOutput();
+		$parsedInfoboxes = $parserOutput ?
+			$parserOutput->getProperty( self::INFOBOXES_PROPERTY_NAME )
+			: false;
+
+		if ( !$parsedInfoboxes ) {
```
idradm: add check for template namespace, please

extensions/wikia/PortableInfobox/services/PortableInfoboxDataService.class.php
```diff
@@ -63,4 +58,51 @@ public function getImages() {
...
+	 */
+	protected function getParsedInfoboxes( $title ) {
+		$article = Article::newFromTitle( $title, RequestContext::getMain() );
+		//on empty parser cache this should be regenerated, see WikiPage.php:2996
+		$parserOutput = $article->getParserOutput();
+		$parsedInfoboxes = $parserOutput ?
+			$parserOutput->getProperty( self::INFOBOXES_PROPERTY_NAME )
+			: false;
+
+		if ( !$parsedInfoboxes && $title->getNamespace() === 10 ) {
```
idradm: Use const here NS_TEMPLATE

idradm:
LG

dianafa:
One bug found here, connected with template name normalisation (underscores to spaces ) performed by API, fixed in DAT-ve-infobox-edit: https://github.com/Wikia/app/blob/DAT-ve-infobox-edit/extensions/VisualEditor/wikia/modules/ve/dm/ve.dm.WikiaTransclusionModel.js#L79

extensions/wikia/PortableInfobox/services/PortableInfoboxDataService.class.php
```diff
@@ -63,4 +58,68 @@ public function getImages() {
...
+
+	/**
+	 * @desc From the template without <includeonly> tags, creates an array of
+	 * strings containing only infoboxes. All template content which is not an infobox is removed.
+	 *
+	 * @param $text string Content of template which uses the <includeonly> tags
+	 * @return array of striped infoboxes ready to parse
+	 */
+	protected function getInfoboxes( $text ) {
+		preg_match_all( "/<infobox.+<\/infobox>/sU", $text, $result );
```
inez, line 110: Fact that a inofobox tag might be wrapped in a nowiki tag makes this a very risky approach.

extensions/wikia/PortableInfobox/services/PortableInfoboxDataService.class.php
```diff
@@ -63,4 +58,68 @@ public function getImages() {
...
+
+		return $result[0];
+	}
+
+	/**
+	 * @desc returns the text from inside of the first <includeonly> tag
+	 * @param $text string template text
+	 * @return mixed
```
inez, line 118: If you change this method to return string or null then you could change documentation here to "@return string" which is more specific than "@return mixed".

extensions/wikia/PortableInfobox/services/PortableInfoboxDataService.class.php
dianafa, line 110: It will appear only in case when inside the `<includeonly>` tags will contain the `<nowiki>` tag which will include the `<infobox>` `</infobox>` tags and between them strange, invalid wikitext which actually is not a real infobox.

While checking how will it really behave in this case, I found an interesting issue in out Infobox Parser which I've created the ticket for: https://wikia-inc.atlassian.net/browse/DAT-3141
so basically for now, templates like this will not parse and shouldn't appear. I'll however add a comment to check this out when the DAT-3141 will be fixed and cases like this will be possible. Thanks!

dianafa, line 110: but it have to be done here, true!

extensions/wikia/PortableInfobox/services/PortableInfoboxDataService.class.php
```diff
@@ -63,4 +58,68 @@ public function getImages() {
...
+			}
+		}
+
+		//on empty parser cache this should be regenerated, see WikiPage.php:2996
+		$parserOutput = $article->getParserOutput();
+		$parsedInfoboxes = $parserOutput ?
+			$parserOutput->getProperty( self::INFOBOXES_PROPERTY_NAME )
+			: false;
+
+		return $parsedInfoboxes;
```
macbre, line 99: Why `$parserOutput->getProperty( self::INFOBOXES_PROPERTY_NAME )` is not called at the top of this method, so we can avoid heavy wikitext parsing and use the parser cache property?

Please see [PLATFORM-1430](https://wikia-inc.atlassian.net/browse/PLATFORM-1430) for details.

dianafa, line 99: we need to cover the case when article is a template, contains `includeonly` tags and inside them we have one or more infoboxes we want to get the params from. The includeonly inside templates are not getting parsed in the Template page (they are parsed only when parsing invocations of them inside articles) thus, we need to manually run parser for the separated infoboxes found in the `includeonly` tags.

Discussed offline.

PR #8129, opened 2015-08-18
ADEN-2341 Ported floor ad unit fix from dev

nandy-andy:
Without this simple code change Mercury version of Floor Adhesion unit will not work. More: [ADEN-2339](https://wikia-inc.atlassian.net/browse/ADEN-2339), connected P2: [ADEN-2341](https://wikia-inc.atlassian.net/browse/ADEN-2341).


fraszczakszymon:
LGTM :+1:

PR #8130, opened 2015-08-18
Remove ExternalRubyCompiler for SASS

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-1313

We use libsass PHP extension now, no need to spawn a Ruby process

@wladekb / @michalroszka / @WikiaSparrow


macbre:
@WikiaSparrow

PR #8137, opened 2015-08-18
Don't Filter Preferences When Saving

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-639

Now that we don't wipe `user_properties` rows for a user when saving their preferences, it's now impossible for a user to delete a preference that has no default defined. This is because since getting a non-existent default returns `null`, it would never pass the `isSaveable` check, so the changed value would never appear in the database. 

At first my approach was to modify the `isSaveable` check to account for missing default values, but I realized that would introduce a different bug since preferences still are not deleted. Instead, I've removed the filtering of preferences to save altogether, so preferences will always be saved, regardless of how they compare to the default value.

I added a simple `Wikia\Util\Optional\Optional` class as part of my initial implementation. It's not strictly needed, but it's probably a good addition to have. I can remove it if the team feels it's unnecessary.

### DIFF
--- a/lib/Wikia/src/Service/User/Preferences/UserPreferences.php
+++ b/lib/Wikia/src/Service/User/Preferences/UserPreferences.php
@@ -3,6 +3,7 @@
 namespace Wikia\Service\User\Preferences;
 
 use Wikia\Domain\User\Preference;
+use Wikia\Util\Optional\Optional;
 
 class UserPreferences {
 	const HIDDEN_PREFS = "user_preferences_hidden_prefs";
@@ -59,7 +60,7 @@ class UserPreferences {
 		$preferences = $this->load($userId);
 
 		if (in_array($pref, $this->hiddenPrefs) && !$ignoreHidden) {
-			return $this->getFromDefault($pref);
+			return $this->getFromDefault($pref)->orElse($default);
 		} elseif (!array_key_exists($pref, $preferences)) {
 			return $default;
 		}
@@ -90,8 +91,8 @@ class UserPreferences {
 			}
 
 			$default = $this->getFromDefault( $pref );
-			if ( $val === null && isset( $default ) ) {
-				$val = $default;
+			if ( $val === null && $default->isPresent() ) {
+				$val = $default->get();
 			}
 			$this->preferences[ $userId ][ $pref ] = $val;
 			$prefToSave[ ] = new Preference( $pref, $val );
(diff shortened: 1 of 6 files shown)


lib/Wikia/src/Util/Optional/Optional.php
```diff
@@ -0,0 +1,67 @@
+<?php
+
+namespace Wikia\Util\Optional;
+
+use Wikia\Util\Assert;
+
+class Optional {
+
+	/** @var bool */
+	private $isPresent;
```
sqreek: If we consider that null is no present, we don't need this field.

sqreek:
one minor comment, else LGTM

lib/Wikia/src/Util/Optional/Optional.php
```diff
@@ -0,0 +1,67 @@
...
+		return self::of($value);
+	}
+
+	public static function emptyOptional() {
+		$optional = new Optional();
+		$optional->isPresent = false;
+
+		return $optional;
+	}
+}
```
drsnyder, line 61: Should we add tests for this class?

lib/Wikia/src/Service/User/Preferences/UserPreferences.php
```diff
@@ -100,12 +95,16 @@ public function setMultiple( $userId, $prefs ) {
 		$this->save( $userId, $prefToSave );
 	}
 
+	/**
+	 * @param $pref
+	 * @return Optional
+	 */
 	public function getFromDefault($pref) {
```
drsnyder, line 102: Is this function being used anywhere else? If so the caller may be broken since the return value changed.

drsnyder:
Minor comment, otw LGTM

lib/Wikia/src/Service/User/Preferences/UserPreferences.php
nmonterroso, line 102: https://github.com/Wikia/app/pull/8137/files#diff-897fba1dd6b8edc3238762e35d0b7c71R2649

lib/Wikia/src/Util/Optional/Optional.php
nmonterroso: Good point. I'll make that change

lib/Wikia/src/Util/Optional/Optional.php
nmonterroso, line 61: Oops. Yeah I'll add that.

lib/Wikia/src/Util/Optional/Optional.php
nmonterroso: 5a5d655

lib/Wikia/src/Util/Optional/Optional.php
nmonterroso, line 61: tests added in ad5c7db

PR #8139, opened 2015-08-19
ADEN-2352 Initialize providers before first fillInSlot call

gabrys:


extensions/wikia/AdEngine/js/AdEngine2.js
```diff
@@ -60,6 +60,32 @@ define('ext.wikia.adEngine.adEngine', [
...
+	function initializeProviderOnce(provider) {
+		if (!provider.fillInSlotQueue) {
+			provider.fillInSlotQueue = [];
+			lazyQueue.makeQueue(provider.fillInSlotQueue, function (args) {
+				provider.fillInSlot.apply(provider, args);
+			});
+
+			if (provider.initialize) {
+				// Only start flushing the fillInSlot queue once the provider is initialized
+				provider.initialize(provider.fillInSlotQueue.start);
```
nandy-andy, line 81: Since you mentioned flushing here I wonder how it affects our ATF implementation. It doesn't, does it? Currently we don't have a provider with `initialize()` method. Do you think we should re-implement ATF/BTF logic with this function?

gabrys, line 81: Well, maybe :-)

nandy-andy:
LGTM :+1: Do you want us to test it?

gabrys:
If possible, yes please :-)

PR #8140, opened 2015-08-19
CurlHttpRequest: set a proxy to an empty string instead of null

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-1317

Try to avoid `PHP Warning: curl_setopt_array(): Curl option contains invalid characters (\0) in /includes/HttpFunctions.php on line 824`

@michalroszka


macbre:
@WikiaSparrow

ghost:
Mea culpa. :+1:

PR #8146, opened 2015-08-19
CONCF-1008 | Do not save empty sections (without items)

vforge:


extensions/wikia/CuratedContent/CuratedContentHelper.class.php
```diff
@@ -11,27 +11,32 @@ class CuratedContentHelper {
...
 		if ( !empty( $sections ) && is_array( $sections ) ) {
 			foreach ( $sections as $section ) {
-				$processedSections[ ] = $this->processLogicForSection( $section );
+				$processedSections[] = $this->processLogicForSection( $section );
 			}
 		}
-		return $processedSections;
+
+		// remove null elements from array
+		return array_filter($processedSections, function($var) { return !is_null($var); } );
```
bognix: white spaces and how about more descriptive name than $var?

bognix:
minor comment, except that LGTM

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
vforge: `$x` ? :P :imp:

hakubo:
yup, some whitespace issues
otherwise LGTM

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
```diff
@@ -11,27 +11,32 @@ class CuratedContentHelper {
...
 				$this->fillItemInfo( $item );
 				$this->processCrop( $item );
 			}
-		}
 
-		return $section;
+			return $section;
+		} else {
+			// return null if we don't have any items inside section
+			return null;
```
tomasznapieralski, line 32: What do you think about removing `else` and moving `return` after `if`?

extensions/wikia/CuratedContent/CuratedContentHelper.class.php
bognix: $section?

bognix:
how about adding unit tests to processSection method?

vforge:
@bognix I was thinking what logic would I test, exactly and only thing I could test in `processSection` is proper use of `array_filter`. Do you think it's necessary?

bognix:
Example tests:
- $sections not empty and is an array
- $sections is empty
- $sections not empty and isn't an array
- $sections not empty and is an array
- $processedSections is [null, null]
- $processedSections is [null, 'foo']
- $processedSections is ['foo', 'foo']

bognix:
LGTM

tomasznapieralski:
LGTM

PR #8148, opened 2015-08-19
Add User::isCurrent() method

ghost:
This method is loosely inspired by User::equals() from MW 1.25
and is meant to replace all the Id / Name comparisons the context
of which would be to determine whether the user in question is
the current performer.

This is a self-contained part of User::isBlocked() refactor (scope of https://wikia-inc.atlassian.net/browse/PLATFORM-1387)

@macbre


macbre:
:+1:

PR #8150, opened 2015-08-19
Use standard PHP function instead of custom one

hakubo:
This method got removed in CONCF-908
as it could be changed to standard PHP function
I have missed to change it here as well


bognix:
LGTM

rogatty:
LGTM

PR #8153, opened 2015-08-19
CONCF-989 Open Mercury's Curated Content Tool in modal

rogatty:
https://wikia-inc.atlassian.net/browse/CONCF-989

### DIFF
--- a/extensions/wikia/CuratedContent/CuratedContentHooks.class.php
+++ b/extensions/wikia/CuratedContent/CuratedContentHooks.class.php
@@ -1,0 +1,66 @@
+<?php
+class CuratedContentHooks {
+	/**
+	 * @brief Whenever data is saved in Curated Content Management Tool
+	 * purge Varnish cache for it and Game Guides
+	 *
+	 * @return bool
+	 */
+	static function onCuratedContentSave() {
+		global $wgServer, $wgWikiaCuratedContent;
+
+		( new SquidUpdate( array_unique( array_reduce(
+			$wgWikiaCuratedContent,
+			function ( $urls, $item ) use ( $wgServer ) {
+				if ( $item['title'] !== '' && empty( $item['featured'] ) ) {
+					// Purge section URLs using urlencode() (standard for MediaWiki), which uses implements RFC 1738
+					// https://tools.ietf.org/html/rfc1738#section-2.2 - spaces encoded as `+`.
+					// iOS apps use this variant.
+					$urls[] = self::getUrl( 'getList' ) . '&section=' . urlencode( $item['title'] );
+					// Purge section URLs using rawurlencode(), which uses implements RFC 3986
+					// https://tools.ietf.org/html/rfc3986#section-2.1 - spaces encoded as `%20`.
+					// Android apps use this variant.
+					$urls[] = self::getUrl( 'getList' ) . '&section=' . rawurlencode( $item['title'] );
+				}
+
+				return $urls;
+			} ,
+			// Purge all sections list getter URL - no additional params
+			[ self::getUrl( 'getList' ), self::getUrl( 'getData' ) ]
+		) ) ) )->doUpdate();
+
+		// Purge cache for obsolete (not updated) apps.
+		if ( class_exists( 'GameGuidesController' ) ) {
+			GameGuidesController::purgeMethod( 'getList' );
+		}
+
+		return true;
+	}
+
+	/**
+	 * Adds assets on the bottom of the body tag
+	 *
+	 * @param {String} $skin
+	 * @param {String} $text
+	 *
+	 * @return bool
+	 */
+	public static function onSkinAfterBottomScripts( $skin, &$text ) {
+		if ( self::shouldDisplayCuratedContentToolButton() ) {
+			$scripts = AssetsManager::getInstance()->getURL( 'curated_content_tool_button_js' );
+
+			foreach ( $scripts as $script ) {
+				$text .= Html::linkedScript( $script );
+			}
+		}
+		return true;
+	}
+
+	private static function shouldDisplayCuratedContentToolButton() {
+		global $wgEnableCuratedContentExt, $wgUser;
+
+		return WikiaPageType::isMainPage() &&
+		!empty( $wgEnableCuratedContentExt ) &&
+		$wgUser->isAllowed( 'curatedcontent' );
+	}
+}
--- a/extensions/wikia/CuratedContent/CuratedContentHooks.class.php
+++ b/extensions/wikia/CuratedContent/CuratedContentHooks.class.php
@@ -60,7 +60,7 @@ class CuratedContentHooks {
 		global $wgEnableCuratedContentExt, $wgUser;
 
 		return WikiaPageType::isMainPage() &&
-		!empty( $wgEnableCuratedContentExt ) &&
-		$wgUser->isAllowed( 'curatedcontent' );
+			!empty( $wgEnableCuratedContentExt ) &&
+			$wgUser->isAllowed( 'curatedcontent' );
 	}
 }
--- a/extensions/wikia/CuratedContent/CuratedContentHooks.class.php
+++ b/extensions/wikia/CuratedContent/CuratedContentHooks.class.php
@@ -47,10 +47,17 @@ class CuratedContentHooks {
 	 */
 	public static function onSkinAfterBottomScripts( $skin, &$text ) {
 		if ( self::shouldDisplayCuratedContentToolButton() ) {
-			$scripts = AssetsManager::getInstance()->getURL( 'curated_content_tool_button_js' );
+			$assetsManager = AssetsManager::getInstance();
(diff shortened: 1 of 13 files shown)


extensions/wikia/CuratedContent/CuratedContentHooks.class.php
```diff
@@ -0,0 +1,66 @@
...
+			}
+		}
+		return true;
+	}
+
+	private static function shouldDisplayCuratedContentToolButton() {
+		global $wgEnableCuratedContentExt, $wgUser;
+
+		return WikiaPageType::isMainPage() &&
+		!empty( $wgEnableCuratedContentExt ) &&
```
vforge: missing `\t`

extensions/wikia/CuratedContent/js/CuratedContentToolButton.js
```diff
@@ -0,0 +1,15 @@
+require(
+	[
```
vforge: i
think
you
shouldn't
declare
like
that
when
it's
not
needed ;)

skins/oasis/modules/templates/PageHeader_Index.php
```diff
@@ -7,12 +7,20 @@
...
+}
+
 $runNjord = ( !empty( $wg->EnableNjordExt ) && WikiaPageType::isMainPage() );
 if ( $runNjord ) {
 	// edit button with actions dropdown
 	if ( !empty( $action ) ) {
 		echo F::app()->renderView( 'MenuButton', 'Index', array( 'action' => $action, 'image' => $actionImage, 'dropdown' => $dropdown, 'name' => $actionName ) );
 	}
+
+	echo $curatedContentToolButton;
```
vforge, line 17: Why you don't have check for `null` here?

skins/oasis/modules/templates/PageHeader_Index.php
```diff
@@ -26,6 +34,8 @@
 				echo F::app()->renderView( 'MenuButton', 'Index', array( 'action' => $action, 'image' => $actionImage, 'dropdown' => $dropdown, 'name' => $actionName ) );
 			}
 
+			echo $curatedContentToolButton;
```
vforge, line 31: Why you don't have check for `null` here?

skins/oasis/modules/PageHeaderController.class.php
```diff
@@ -165,6 +167,11 @@ public function executeIndex( $params ) {
 		// check for video add button permissions
 		$this->showAddVideoBtn = $wgUser->isAllowed( 'videoupload' );
 
+		$this->showCuratedContentToolBtn =
+			WikiaPageType::isMainPage() &&
```
vforge: This should go in previous line.

extensions/wikia/CuratedContent/CuratedContentHooks.class.php
rogatty: Autoformatting said it's fine like that ¯_(ツ)_/¯

skins/oasis/modules/templates/PageHeader_Index.php
rogatty, line 17: What's wrong with echoing `null`?

skins/oasis/modules/PageHeaderController.class.php
rogatty: But now it looks so good :(

extensions/wikia/CuratedContent/CuratedContentHooks.class.php
vforge: `¯\_(ツ)_/¯`

skins/oasis/modules/templates/PageHeader_Index.php
```diff
@@ -7,12 +7,20 @@
 	}
 }
 
+if ( $showCuratedContentToolBtn ) {
+	$curatedContentToolButton = '<a id="CuratedContentTool" class="wikia-button">' . wfMsg( 'wikiacuratedcontent-edit-mobile-main-page' ) . '</a>';
```
Warkot: `wfMsg()` is deprecated — please use `wfMessage()->text()` (or `->escaped()`, there are two different views on this). https://www.mediawiki.org/wiki/Manual:Messages_API#Help_with_replacing_deprecated_wfMsg.2A_functions

rogatty: I used it because it's already being used in this file. Do you think I should change all occurences of it? That would probably mean regression testing.

extensions/wikia/CuratedContent/CuratedContentHooks.class.php
bognix: Have you used this https://github.com/Wikia/guidelines/blob/master/PHP/bin/php-mediawiki-stylize?

extensions/wikia/CuratedContent/js/CuratedContentToolButton.js
```diff
@@ -0,0 +1,15 @@
...
+		'jquery'
+	],
+	function ($) {
+		$('#CuratedContentTool').click(function () {
+			var iframe = '<iframe data-url="/main/edit?useskin=mercury" id="CuratedContentToolIframe" name="curated-content-tool" width="100%" height="500"></iframe>',
+				title = 'Mobile Main Page';
+
+			require(['curatedContentTool.modal'], function (curatedContentToolModal) {
```
bognix: how about moving this require to the top of the file?

rogatty: That would mean loading additional file on every page load instead of when user clicks the button. This is a solution copied from WikiaMaps and IMO it makes sense.

skins/oasis/modules/templates/PageHeader_Index.php
Warkot: Hard to tell. It's good to remove deprecations and vulnerabilities, but on the other hand those possible regressions… I would still do it, but first ask someone from platform (ping @macbre, @michalroszka) should we do it — and if yes, then should we use here the `->escaped()` method instead of `->text()`.

extensions/wikia/CuratedContent/js/CuratedContentToolButton.js
Warkot: +1 for @rogatty if the script actually sends the request for the code after the click.

skins/oasis/modules/PageHeaderController.class.php
```diff
@@ -165,6 +167,11 @@ public function executeIndex( $params ) {
...
+			!empty( $wgEnableCuratedContentExt ) &&
```
bognix: maybe we should consider checking this wg variable in mercury too because we are asking controller directly there and if it's set to false it would throw an error

skins/oasis/modules/templates/PageHeader_Index.php
vforge: On the other note - Creating snippets like that seems wrong - have you considered creating new function in controller and separate template for this?

extensions/wikia/CuratedContent/js/CuratedContentToolButton.js
bognix: are you sure about not loading the file everywhere? I thought curated_content_tool_button_js  asset's manager group is loaded when this is called https://github.com/Wikia/app/pull/8153/files#diff-f2c81fb514ec8b3f2c72e17d647ac290R50

skins/oasis/modules/templates/PageHeader_Index.php
vforge, line 17: You're checking for that later in the template

rogatty, line 17: So we should probably remove the other `if` that I forgot about :)

skins/oasis/modules/templates/PageHeader_Index.php
rogatty: This whole file seems wrong and is a candidate for refactoring. CONCF-989 was estimated at 3 story points. Do you think we can do the refactor in that time?

extensions/wikia/CuratedContent/js/CuratedContentToolButton.js
```diff
@@ -0,0 +1,15 @@
+require(
+	[
+		'jquery'
+	],
+	function ($) {
+		$('#CuratedContentTool').click(function () {
+			var iframe = '<iframe data-url="/main/edit?useskin=mercury" id="CuratedContentToolIframe" name="curated-content-tool" width="100%" height="500"></iframe>',
```
dianafa: will this string be always the same?

hakubo:
LGTM

skins/oasis/modules/templates/PageHeader_Index.php
hakubo, line 31: what does echo null; returns?

bognix, line 31: nothing

vforge:
LGTM

PR #8155, opened 2015-08-19
SOC-789 log attribute mismatches between attirbute service and MW

jsutterfield:
Ticket: https://wikia-inc.atlassian.net/browse/SOC-789

### DIFF
--- a/includes/User.php
+++ b/includes/User.php
@@ -2672,7 +2672,7 @@ class User {
 	 * @param mixed $default
 	 * @return string
 	 */
-	public function getGlobalAttribute( $attribute, $default=null ) {
+	public function getGlobalAttribute( $attribute, $default = null ) {
 		global $wgEnableReadsFromAttributeService;
 
 		$valueFromMW = $this->getOptionHelper( $attribute, $default );
--- a/includes/User.php
+++ b/includes/User.php
@@ -2672,32 +2672,32 @@ class User {
 	 * @param mixed $default
 	 * @return string
 	 */
-	public function getGlobalAttribute($attribute, $default=null) {
+	public function getGlobalAttribute( $attribute, $default=null ) {
 		global $wgEnableReadsFromAttributeService;
 
-		$valueFromMW = $this->getOptionHelper($attribute, $default);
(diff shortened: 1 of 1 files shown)


includes/User.php
```diff
@@ -2673,7 +2673,31 @@ public function getLocalAttribute($attr, $cityId = null, $sep = "-") {
...
+	}
+
+	private function logAttributeMismatch($valueFromMW, $valueFromService, $attribute, $default) {
+		$this->error("USER_ATTRIBUTES attribute_mismatch", [
+				"valueFromMW" => $valueFromMW,
+				"valueFromService" => $valueFromService,
+				"attribute" => $attribute,
+				"default" => $default,
+				"userId" => $this->getId()
+			]);
```
kvas-damian: Coding conventions ;)
spaces everywhere + single quotes instead of doublequotes

Besides that LGTM

includes/User.php
```diff
@@ -2673,7 +2673,31 @@ public function getLocalAttribute($attr, $cityId = null, $sep = "-") {
 	 * @return string
 	 */
 	public function getGlobalAttribute($attribute, $default=null) {
-		return $this->getOptionHelper($attribute, $default);
+		global $wgEnableReadsFromAttributeService;
+
+		$valueFromMW = $this->getOptionHelper($attribute, $default);
```
rafalkalinski: spaces, spaces everywhere :)

includes/User.php
jsutterfield: Originally I thought to do this with no spaces since it's core code and I wanted to be consistent, but it looks like there's code riddled throughout this file which follows our conventions. So, I'll update :)

includes/User.php
```diff
@@ -2672,8 +2672,32 @@ public function getLocalAttribute($attr, $cityId = null, $sep = "-") {
 	 * @param mixed $default
 	 * @return string
 	 */
-	public function getGlobalAttribute($attribute, $default=null) {
-		return $this->getOptionHelper($attribute, $default);
+	public function getGlobalAttribute( $attribute, $default=null ) {
```
kvas-damian: Missing spaces around equals sign ;)

Besides that LGTM

PR #8156, opened 2015-08-19
SOC-1186 Make sure we don't set wgDBname

garthwebb:
Doing this has weird effects on DB handle loading from wgGetDB()

Also did some clean up of WikiService::getWikiAdminIds

https://wikia-inc.atlassian.net/browse/SOC-1186

### DIFF
--- a/includes/wikia/Wikia.php
+++ b/includes/wikia/Wikia.php
@@ -105,8 +105,12 @@ class Wikia {
 	public static function initAsyncRequest( $wikiId, $user = null ) {
 		$wg = F::app()->wg;
 		$wg->CityID = $wikiId;
-		$wg->DBname = WikiFactory::IDtoDB( $wikiId );
-		$wg->Server = trim( WikiFactory::DBtoUrl( $wg->DBname ), '/' );
+
+		// Do NOT set $wgDbname here.  The wfGetDB method will no longer do
+		// what you think it should since it pulls from a LoadBalance cache
+		// that likely already has cached DB handles for the previous value
+		$dbName = WikiFactory::IDtoDB( $wikiId );
+		$wg->Server = trim( WikiFactory::DBtoUrl( $dbName ), '/' );
 
 		if ( !empty( $wg->DevelEnvironment ) ) {
 			$wg->Server = WikiFactory::getLocalEnvURL( $wg->Server );
@@ -220,8 +224,7 @@ class Wikia {
      *
      * @return string composed HTML/XML code
      */
-    static public function errormsg($what)
-    {
+    static public function errormsg($what) {
         return Xml::element("span", array( "style"=> "color: #fe0000; font-weight: bold;"), $what);
     }
 
@@ -242,8 +245,7 @@ class Wikia {
      *
      * @return string composed HTML/XML code
      */
-    static public function linkTag($url, $title, $attribs = null )
-    {
+    static public function linkTag($url, $title, $attribs = null ) {
         return Xml::element("a", array( "href"=> $url), $title);
     }
 
@@ -260,8 +262,7 @@ class Wikia {
      *
      * @return string composed HTML/XML code
      */
-    static public function successmsg($what)
-    {
+    static public function successmsg($what) {
         return Xml::element("span", array( "style"=> "color: darkgreen; font-weight: bold;"), $what);
     }
 
--- a/includes/wikia/services/WikiService.class.php
+++ b/includes/wikia/services/WikiService.class.php
@@ -92,7 +92,7 @@ class WikiService extends WikiaModel {
 		return $userIds;
 	}
 
-	private function getAdminIdsFromDB( DatabaseBase $db, $excludeBots = false, $limit = null ) {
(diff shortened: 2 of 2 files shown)


includes/wikia/Wikia.php
```diff
@@ -105,8 +105,12 @@ public static function unsetVar($key) {
...
 		$wg = F::app()->wg;
 		$wg->CityID = $wikiId;
-		$wg->DBname = WikiFactory::IDtoDB( $wikiId );
-		$wg->Server = trim( WikiFactory::DBtoUrl( $wg->DBname ), '/' );
+
+		// Do NOT set $wgDbname here.  The wfGetDB method will no longer do
+		// what you think it should since it pulls from a LoadBalance cache
+		// that likely already has cached DB handles for the previous value
+		$dbName = WikiFactory::IDtoDB( $wikiId );
+		$wg->Server = trim( WikiFactory::DBtoUrl( $dbName ), '/' );
```
garthwebb, line 113: This was the fix.  There was a cached DB handle to the 'wikia' db since this is the db that was set when the script starts up that runs this (e.g. we run with SERVER_ID=177).  When calling:

``` php
$wgDBname = 'somewikiname';

...

$db = wfGetDB( DB_SLAVE, null, 'somewikiname' )
```

Code within `wfGetDB` notices that the current wiki as given by `$wgDBname` is the same as the name passed into the function.  It then clears out the wiki name for the rest of the call, assuming it can always get the name from `$wgDBname` if its needs it, and allowing default actions to take place from there on.  This eventually leads it to the LoadBalancer cache where it gets the first connection it finds in the default pool, a pool populated by connections to the 'wikia' DB.

includes/wikia/services/WikiService.class.php
```diff
@@ -47,66 +52,74 @@ class WikiService extends WikiaModel {
 	 * @return array of $userIds
 	 */
 	public function getWikiAdminIds( $wikiId = 0, $useMaster = false, $excludeBots = false, $limit = null, $includeFounder = true ) {
```
garthwebb, line 54: Did some clean up here.  There were three levels of nested "Is this condition true?  If not, stop doing anything"  For those we might as well bail immediately and not create so much nesting and indenting.

includes/wikia/services/WikiService.class.php
```diff
@@ -47,66 +52,73 @@ class WikiService extends WikiaModel {
...
-		if ( empty($this->wg->FounderEmailsDebugUserId) ) {
-			// get founder
-			$wikiId = ( empty($wikiId) ) ? $this->wg->CityId : $wikiId ;
-			$wiki = WikiFactory::getWikiById($wikiId);
-			if ( !empty($wiki) && $wiki->city_public == 1 ) {
-				if ($includeFounder) {
-					$userIds[] = $wiki->city_founding_user;
-				}
+		if ( !empty( $this->wg->EnableAnswers ) ) {
+			return [];
```
jsutterfield: Based on the old code it looks like it would be possible to return `$wiki->city_founding_user;` even when `!empty( $this->wg->EnableAnswers )`, is that a case we should support still?

garthwebb: Good point, I missed that.

includes/wikia/services/WikiService.class.php
```diff
@@ -47,66 +52,73 @@ class WikiService extends WikiaModel {
...
+				return self::getAdminIdsFromDB( $db, $excludeBots, $limit );
+			}
+		);
+
+		$userIds = array_unique( array_merge( $userIds, $adminIds ) );
+
 		return $userIds;
 	}
 
+	private function getAdminIdsFromDB( DatabaseBase $db, $excludeBots = false, $limit = null ) {
```
jsutterfield: Should this be static? Or should the call to it use `$this->` instead of `self::`?

garthwebb: @jsutterfield, yah, should be static

includes/wikia/services/WikiService.class.php
```diff
@@ -47,66 +52,73 @@ class WikiService extends WikiaModel {
...
+			}
+		);
+
+		$userIds = array_unique( array_merge( $userIds, $adminIds ) );
+
 		return $userIds;
 	}
 
+	private static function getAdminIdsFromDB( DatabaseBase $db, $excludeBots = false, $limit = null ) {
+		$conditions = [ "ug_group in ('sysop','bureaucrat')" ];
```
Grunny: Minor: Preferred way to do IN conditions when using MW's database class would be `[ 'ug_group' => [ 'sysop', 'bureaucrat' ] ];`. Could also be done in the subquery, using `DatabaseBase::selectSQLText` to build the subquery, but this is simple enough.

Grunny:
LGTM.

includes/wikia/services/WikiService.class.php
garthwebb: Yah, this was a copy paste rearrangement of the old code but we should update this.  If I was a little more motivated I'd move this to the Fluent interface :)

PR #8157, opened 2015-08-20
VOLDEV-135 - Unset fixed width on article comment replies textarea

onei:
@Grunny https://wikia-inc.atlassian.net/browse/VOLDEV-135


Grunny:
LGTM.

PR #8159, opened 2015-08-20
Improve the recognition of the current performer

ghost:
- `mIsCurrent` is a boolean property of `User` instance; `true` indicates
  that the instance represents the current performer, `false` otherwise.
  The property _must not be cached_ as it only makes sense within the scope
  of a particular execution / request.
- `User::isCurrent()` method returns the value of the property. In the
  first implementation is double-checks the value with the original
  comparison used in the past in order to detect errors and edge
  cases.
- `User::setCurrent()` method marks that an instance of the `User` class
  reflects the current performer.

@macbre @owend @Grunny @wladekb @drsnyder 

Also, @Nikerabbit ― this is pretty much what I meant in https://gerrit.wikimedia.org/r/#/c/232683/ and your input here is more than welcome! :-)

### DIFF
--- a/includes/User.php
+++ b/includes/User.php
@@ -243,6 +243,8 @@ class User {
 	 */
 	private $mBlockedFromCreateAccount = false;
 
+	var $mIsCurrent = false; # Wikia - indicates that the instance represents the current performer
+
 	static $idCacheByName = array();
 
 	/**
@@ -1886,6 +1888,13 @@ class User {
 	}
 
 	/**
+	 * Mark that the instance represents the current performer
+	 */
+	public function setCurrent() {
+		$this->mIsCurrent = true;
+	}
+
+	/**
 	 * Get the user name, or the IP of an anonymous user
 	 * @return String User's name or IP address
 	 */
@@ -3124,15 +3133,25 @@ class User {
 	}
 
 	/**
-	 * Get whether the user is the current performer
+	 * Get whether the instance represents the current performer
 	 *
-	 * Inspired by User::equals() and its usage in MW 1.25
 	 * @return Bool
 	 * @author Michał Roszka <michal@wikia-inc.com>
 	 */
 	public function isCurrent() {
+		// This is the first implementation. I still rely on $wgUser here
+		// and want to log all cases when mIsCurrent property is false
+		// for the current performer and fix it for future checks.
 		global $wgUser;
-		return $this->getName() === $wgUser->getName();
+		if ( !$this->mIsCurrent && $this->getName() === $wgUser->getName() ) {
+			Wikia\Logger\WikiaLogger::instance()->error(
+				'wrong-mIsCurrent',
+				[ 'exception' => new Exception ]
+			);
+			// fix the wrong mIsCurrent
+			$this->setCurrent();
+		}
+		return $this->mIsCurrent;
 	}
 
 	/**
--- a/includes/context/RequestContext.php
+++ b/includes/context/RequestContext.php
@@ -218,6 +218,11 @@ class RequestContext implements IContextSource {
 			$this->user = User::newFromSession( $this->getRequest() );
 		}
 
+		// Wikia change - begin - $author: Michał Roszka
+		// Indicate that this particular User instance represents the current performer.
+		$this->user->setCurrent();
+		// Wikia change - end
+
 		// Replace the user object according to the context, e.g. Piggyback.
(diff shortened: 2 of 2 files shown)


macbre:
:+1:

Grunny:
:+1:

includes/User.php
```diff
@@ -3124,15 +3133,25 @@ public function isAnon() {
...
 	 * @return Bool
 	 * @author Michał Roszka <michal@wikia-inc.com>
 	 */
 	public function isCurrent() {
+		// This is the first implementation. I still rely on $wgUser here
+		// and want to log all cases when mIsCurrent property is false
+		// for the current performer and fix it for future checks.
 		global $wgUser;
-		return $this->getName() === $wgUser->getName();
+		if ( !$this->mIsCurrent && $this->getName() === $wgUser->getName() ) {
```
drsnyder, line 3146: Isn't the user's id the only invariant that we have for the user? Should we be comparing user id instead of the name?

drsnyder:
Did something specific prompt this change?

ghost:
@drsnyder ― Yes. My intention is to distinguish situations when `User::isBlocked()` is called _in the first person_ (on the very instance representing the current performer, as in "can I do this") or _in the third person_ (on some other instance of `User`, as in "can she do that"). In the first case it pretty much always means preventing from some action whereas in the latter this mere information. I turns out we need that distinction since we decoupled Phalanx from MediaWiki. The other reason is making stuff cleaner and simpler. Imagine that three hundred fifty six layers of code deeper in some method hooked to some event nobody knows when it is actually happening ― you want to know whether some object represents the current performer, because you want to modify it. Today you would need to

``` php
global $wgUser; 
if ( $u->getName() === $wgUser->getName() ) {
// blah blah blah
}
```

... and you only know that `$u` is an object representing the same entity and there is still a fair chance it is not _the instance_ representing the current performer, which could result with a very obscure bug. With `User::isCurrent()` based on the property set at the right moment it is going to be dead simple and conclusive.

drsnyder:
Ok, makes sense. The only other concern is whether or not it makes sense to compare the user id's or the usernames. As far as I know the user can never change but the username can.

includes/User.php
ghost, line 3146: No, the name is a unique property, too. All the `User:Name` pages, for instance, need the name to be a conclusive identifier. MediaWiki 1.25 uses names for the same type of comparisons. I suspect the reason behind it is as follows:
- it is generally better to use string identifiers to identify a resource; means: integers should only be used for quantity and arithmetics;
- type-sensitive comparisons of numeric identifiers require the operands to be of the same type while they are integers when retrieved from the database by many ORMs and strings when provided as GET parameters. The name is a string and strings are strings. :-)
- type-insensitive comparison could have serious consequences for values like `false`, `0`, `'0'`, `''`, `null`, `0.1e2`, because PHP.

ghost:
Good catch with the `user_id` that can never change. Although, the `user_name` does not change within the scope of a single request. Therefore in a situation when you are compare a user with id X and some 'OldName' with a user with id X and some 'NewName', only one of them can represent the actual _current performer_, because you can never act under your old name :-)

ghost:
Also, note that using names for that detection will be gone soon and replaced with the boolean `mIsCurrent` property set at the right moment during the execution. The comparison of the names is just for debugging purposes to help me detect where and when should I set the property to `true`. As a matter of fact, if you want to compare two objects, it is up to you, depending on the context, to decide which properties to compare or whether to compare the entire objects.

And there is another thing - replacing `$wgUser` during the execution which I am guilty of having done (HAWelcome) and we both didn't like it. `mIsCurrent` kind of sets some limits and requires re-thinking implementation rather than using a simple and quick hacks.

Nikerabbit:
Today you would not write

```
global $wgUser; 
if ( $u->getName() === $wgUser->getName() ) {
// blah blah blah
}
```

because global objects are banned in new MediaWiki code. You would do
`$context->getUser()->equals( $u );`

ghost:
@Nikerabbit ― good point and I appreciate that you involved into this discussion. The thing is https://github.com/Wikia/app is pretty much 1.16-based legacy code ported to 1.19 which makes it even more legacy (because of how it was ported) :-) And in our case de-`global`-isation of our Frankenstein must be approached in a slightly different manner. Tones of code to refactor. Although backporting `User::equals()` anyways from the current MW back to our 1.19 makes perfect sense. And it is a slightly different story, too, as inspired by that, I am proposing the `mIsCurrent` for the current performer distinction specifically...

PR #8164, opened 2015-08-20
ADEN-2322 Handle hashed SourcePoint div elements

fraszczakszymon:


extensions/wikia/AdEngine/js/provider/gpt/googleTag.js
```diff
@@ -140,5 +142,12 @@ define('ext.wikia.adEngine.provider.gpt.googleTag', [
 		registeredCallbacks[id] = callback;
 	};
 
+	GoogleTag.prototype.onAdLoad = function (slotName, element, gptEvent, onAdLoadCallback) {
```
nandy-andy, line 145: This is a new functionality in our `GoogleTag` wrapper I wonder how much it affects ads' load time. What do you think about checking it with some benchmarking?

nandy-andy, line 145: OK, nevermind. We did it before in `GptHelper`. So, it doesn't change anything ;) Correct me if I'm wrong, please :P

extensions/wikia/AdEngine/js/provider/gpt/sourcePointTag.js
```diff
@@ -36,16 +73,70 @@ define('ext.wikia.adEngine.provider.gpt.sourcePointTag', [
...
+			slotElementId;
+
+		if (blocking && gptEvent.slot && gptEvent.slot.getSlotElementId) {
+			slotElementId = gptEvent.slot.getSlotElementId() || element.getId();
+			newSlotName = slotElementId.replace(/(.*)([\/])([^\/]*$)/, '$3');
+
+			if (slotName !== newSlotName) {
+				copyStyles(slotName, newSlotName);
+
+				iframe = doc.getElementById(slotElementId).querySelector('div[id*="_container_"] iframe');
```
nandy-andy, line 97: How bad idea it is to extract `'div[id*="_container_"] iframe'` to a variable and re-use this variable here and in `GoogleTag`?

skins/oasis/css/core/ads.scss
```diff
@@ -14,6 +14,10 @@
 	display: none !important;
 }
 
+.source-point .wikia-ad.default-height {
```
nandy-andy, line 17: What are the reasons for this new rule? Is it really needed with the whole JavaScript logic of copying styles between elements?

extensions/wikia/AdEngine/js/provider/gpt/googleTag.js
fraszczakszymon, line 145: Yup, I think so. Unfortunately it's more nested callback

extensions/wikia/AdEngine/js/provider/gpt/sourcePointTag.js
```diff
@@ -36,16 +73,70 @@ define('ext.wikia.adEngine.provider.gpt.sourcePointTag', [
...
+
+	SourcePointTag.prototype.onAdLoad = function (slotName, element, gptEvent, onAdLoadCallback) {
+		log(['onAdLoad', slotName], 'debug', logGroup);
+		var iframe,
+			newSlotName,
+			slotElementId;
+
+		if (blocking && gptEvent.slot && gptEvent.slot.getSlotElementId) {
+			slotElementId = gptEvent.slot.getSlotElementId() || element.getId();
+			newSlotName = slotElementId.replace(/(.*)([\/])([^\/]*$)/, '$3');
```
nandy-andy: What do you think about extracting it to new unit-tested function? Don't we have similar functionality somewhere?

extensions/wikia/AdEngine/js/provider/gpt/sourcePointTag.js
```diff
@@ -5,17 +5,54 @@ define('ext.wikia.adEngine.provider.gpt.sourcePointTag', [
...
+
+		cssText = '';
+		for (var i = 0; i < style.length; i++) {
+			cssText += style[i] + ':' + style.getPropertyValue(style[i]) + ';';
+		}
+
+		return cssText;
+	}
+
+	function copyStyles(from, to) {
```
nandy-andy: Those two functions look like things which don't really belong to `SourcePointTag`. Shall we create `ext.wikia.adEngine.cssTweaker`?

extensions/wikia/AdEngine/js/provider/gpt/sourcePointTag.js
```diff
@@ -36,16 +73,70 @@ define('ext.wikia.adEngine.provider.gpt.sourcePointTag', [
...
+		var iframe,
+			newSlotName,
+			slotElementId;
+
+		if (blocking && gptEvent.slot && gptEvent.slot.getSlotElementId) {
+			slotElementId = gptEvent.slot.getSlotElementId() || element.getId();
+			newSlotName = slotElementId.replace(/(.*)([\/])([^\/]*$)/, '$3');
+
+			if (slotName !== newSlotName) {
+				copyStyles(slotName, newSlotName);
```
nandy-andy: I'm thinking of gathering slots in an array and copy those styles in one loop. But it's impossible since we do it on an ad-load event and we don't know here if there are more slots to checked or not :/ Any other ideas how to make it better performance-wise? If not, then we'll have to live with it, I guess.

extensions/wikia/AdEngine/js/provider/gpt/sourcePointTag.js
fraszczakszymon, line 97: In this case **slotElementId** is different

skins/oasis/css/core/ads.scss
fraszczakszymon, line 17: It's used only in "old" slots divs, but I guess we can do the same thing by removing those classes in JS code instead of adding .source-point to body, what do you think?

extensions/wikia/AdEngine/js/provider/gpt/sourcePointTag.js
fraszczakszymon: :+1:

extensions/wikia/AdEngine/js/provider/gpt/sourcePointTag.js
fraszczakszymon: :+1: I've found it in adDetect

skins/oasis/css/core/ads.scss
nandy-andy, line 17: As we discussed offline having it in JS only doesn't work nice since the end user can still see the jumping. Let's leave it as it is now.

nandy-andy:
LGTM :+1:

PR #8166, opened 2015-08-20
Pref service ramp

nmonterroso:
@Wikia/services-team 

bumping usage of the user preference service to 15% and fixing some `getOption` callsites


drsnyder:
lgtm

PR #8167, opened 2015-08-20
update User::setGlobalAttribute calls to User::setGlobalFlag calls

jsutterfield:
These `User::setGlobalAttribute` calls should be `User::setGlobalFlag` calls.

When pulling this data, we're using `User::getGlobalFlag` which is inconsistent. You can see [Here](https://github.com/Wikia/app/blob/1741bf19b3205c85d3c3d9d3dbfbe3c9388048e5/extensions/wikia/UserProfilePageV3/UserIdentityBox.class.php#L273) and [here](https://github.com/Wikia/app/blob/1741bf19b3205c85d3c3d9d3dbfbe3c9388048e5/extensions/wikia/UserProfilePageV3/UserIdentityBox.class.php#L137)


mklucsarits:
LGTM

PR #8168, opened 2015-08-20
INT-155: Adding new wiki global vairable available for Mercury

mari-mcmurtrie:
For JP wikia on Mercury, we currently allow anonymous file upload. We would like to add a new control for enable/disable anonymous file upload. Please also see https://github.com/Wikia/config/pull/1249 for the new global variable to control this. This change makes this variable available for Mercury front-end so that we can control UI.


mari-mcmurtrie:
@lizlux will you please code review? We talked about this yesterday. Thanks!!

lizlux:
LGTM

PR #8171, opened 2015-08-21
SERVICES-644 Read-only mode for preferences

drsnyder:
This is for emergency use only to assist in the resolution of https://wikia-inc.atlassian.net/browse/SERVICES-644. While the new table is being moved into place we don't want to write to preferences.

https://wikia-inc.atlassian.net/browse/SERVICES-644

Failed tests are expected since this is unexpected behavior. Once the table is in place on the slaves this can be reverted.


nmonterroso:
LGTM

PR #8176, opened 2015-08-21
SERVICES-627_Consolidate the logout functionality

sqreek:
Add AuthService and HeliosAuthService as implementation
Move logout call from Helios extension into User.php
remove mercury access token

ping @drsnyder @nmonterroso @harnash @ArturKlajnerok

### DIFF
--- a/lib/Wikia/src/Service/User/Auth/AuthServiceBase.php
+++ b/lib/Wikia/src/Service/User/Auth/AuthServiceBase.php
@@ -1,0 +1,30 @@
+<?php
+
+namespace Wikia\Service\User\Auth;
+
+
+abstract class AuthServiceBase implements AuthService {
+
+	/**
+	 * @see AuthService
+	 */
+	public abstract function logout();
+
+	/**
+	 * @see AuthService
+	 */
+	public function isUsernameBlocked( $username ) {
+		$user = \User::newFromName( $username );
+
+		if ( !$user ) {
+			return null;
+		}
+
+		$user->load();
+		if ( $user->getId() == 0 ) {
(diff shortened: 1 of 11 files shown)


sqreek:
Tests was fallen becuse of jenkins (http://qa-s2.wikia-prod:8080/job/unitTests_js_branches/16696/console)

Now they passes but no sending check (http://qa-s2.wikia-prod:8080/job/unitTests_js_branches/16711/)

harnash:
Seems fine to me.

lib/Wikia/src/Service/User/Auth/HeliosAuthService.php
```diff
@@ -0,0 +1,38 @@
...
+	}
+
+	/**
+	 * Call helios invalidate token.
+	 */
+	private function invalidateAccessTokenInHelios() {
+		$request = \RequestContext::getMain()->getRequest();
+		$this->heliosClient->invalidateToken( HeliosHelper::getAccessToken( $request ) );
+	}
+}
```
drsnyder, line 38: Isn't there also a `logout` required on the user object?

lib/Wikia/src/Service/User/Auth/AuthServiceBase.php
```diff
@@ -0,0 +1,30 @@
...
+			return null;
+		}
+
+		$user->load();
+		if ( $user->getId() == 0 ) {
+			return null;
+		}
+
+		return $user->isBlocked();
+	}
```
drsnyder, line 29: I don't think this is necessary here since it's on `User`.

PR #8181, opened 2015-08-21
PLATFORM-1334: Masthead will now use the user avatar service

macbre:
- https://wikia-inc.atlassian.net/browse/PLATFORM-1334 (`Masthead` class to use the avatar service)
- https://wikia-inc.atlassian.net/browse/PLATFORM-1419 (the avatars migration script)

The avatar functionality (which is a part of `Masthead` class) will now use a shiny new avatars service instead of direct uploads to DFS.
- introduce `$wgAvatarsUseService` global that will turn on avatar uploads via service
- clean-up `Masthead` class code and remove NFS support
- introduce `UserAvatarsService` class generated by Swagger (used to communicate with a service)
- `extensions/wikia/UserProfilePageV3/maintenance/migrateAvatarsViaService.php` maintenance script that will upload all existing avatars via avatars service (tested on devboxes)
- extend the Vignette URL generator by introducing `StaticAssetsUrlGenerator` that will generate URLs to assets uploaded via Static Assets Service (used by avatars service)
## The plan

**Phase 1** set `$wgAvatarsUseService` to true and handle avatar uploads via service (make sure it works stable under production traffic)
**Phase 2** run the migration script and upload all custom avatars to StaticAssetsService (via avatars service)

@jcellary / @wladekb / @michalroszka


extensions/wikia/UserProfilePageV3/maintenance/migrateAvatarsViaService.php
```diff
@@ -0,0 +1,142 @@
...
+		// no avatar set, skip this account
+		if ( is_null( $avatar ) ) {
+			$this->output( 'no avatar set - skipping' );
+			return;
+		}
+
+		// no need to store default avatars in user properties - remove the entry
+		else if ( self::isDefaultAvatar( $avatar ) ) {
+			$this->output( 'default avatar set - removing an attribute' );
+			$user->removeGlobalPreference( AVATAR_USER_OPTION_NAME );
```
macbre: @jsutterfield - can I call `User::removeGlobalPreference` method to remove the global attribute?

It makes no sense to keep the default avatar in the user preferences database (142k rows currently).

extensions/wikia/UserProfilePageV3/maintenance/migrateAvatarsViaService.php
```diff
@@ -0,0 +1,169 @@
...
+		// get all accounts
+		global $wgExternalSharedDB;
+		$db = $this->getDB( DB_SLAVE, [], $wgExternalSharedDB );
+
+		$res = $db->select(
+			'`user`',
+			'user_id AS id',
+			[],
+			__METHOD__
+		);
```
macbre, line 55: @jsutterfield, Is the user preferences service going to keep storing preferences in `user_preferences` table or do we need to get the list of all accounts and access preferences via `User` object?

In other words: can we perform the following query on `wikicities` to get the list of all accounts that require avatars migration:

``` sql
select * from user_properties where up_property = 'avatar';
```

jsutterfield, line 55: @macbre as far as I know, yes, the preferences service will keep storing preferences in the `user_preferences`. I haven't heard any plans to migrate that data to another table yet. Especially not in the short term. So yes, performing that query will get a list of all accounts that requires an avatar migration.

macbre, line 55: Great, thanks! f7716d3

extensions/wikia/UserProfilePageV3/maintenance/migrateAvatarsViaService.php
macbre: @jsutterfield: ping :)

jsutterfield: Oof, sorry for not responding to this earlier...I remember looking at this and thought I had, but clearly didn't.

I don't think we should actually call `$user->removeGlobalPreference( AVATAR_USER_OPTION_NAME );` here. The current behavior when a user deletes their attribute from their profile page is to set the value as an empty string. This is the behavior that I know the discussion service expects from the attributes service (ie, when a user doesn't have an avatar to receive an empty string), and until we update the behavior of deleting avatars from the profile page, we'll end up with a mixture of no values in the db for user's avatars, and empty strings.

So for now I'd argue we should leave them as empty string. They should probably be removed at some point (since I agree, it is silly to store 142k rows of nothing) but we should make sure to coordinate the change with updating the user profile page, and what clients of the attribute service can expect when a user doesn't have an avatar set.

macbre: Thanks. Changed the approach in 47a70c5

extensions/wikia/UserProfilePageV3/maintenance/migrateAvatarsViaService.php
macbre, line 55: @jsutterfield, I've reverted f7716d3d8e183be85e63b309ab3303719bd7fe29

The plan is to first enable avatar uploads via the new service and migrate the avatars when we verify that the avatars service is stable. Hence, we can not query `user_preferences` table as it can have settings that are out of date (the preferences service uses a separate database to store settings) - get the list of **all** users and query the preferences service (via `User::getGlobalAttribute` method). This will make the process take longer, but "better safe than sorry" :)

jsutterfield, line 55: Yea, that makes sense to me. :+1:

Sorry for the confusion earlier.

pawelchcki:
Looks good

extensions/wikia/UserProfilePageV3/Masthead.class.php
```diff
@@ -427,7 +429,7 @@ public function getSwiftStorage() {
 	/**
 	 * @return string temporary file to be used for upload from URL
 	 */
-	public function getTempFile() {
+	private function getTempFile() {
```
ghost, line 432: `protect` your `private` parts :-)

extensions/wikia/UserProfilePageV3/Masthead.class.php
```diff
@@ -454,13 +456,13 @@ public function isDefault() {
 	 * @return bool result
 	 */
 	private function doRemoveFile() {
```
ghost, line 458: `protect` your `private` parts :-)

extensions/wikia/UserProfilePageV3/Masthead.class.php
```diff
@@ -488,18 +491,20 @@ private function doRemoveFile() {
 	 *
 	 * @return bool
 	 */
-	public function fileExists() {
+	private function fileExists() {
```
ghost, line 494: `protect` your `private` parts :-)

extensions/wikia/UserProfilePageV3/Masthead.class.php
```diff
@@ -488,18 +491,20 @@ private function doRemoveFile() {
...
 		wfProfileIn( __METHOD__ );
 
-		global $wgAvatarsUseSwiftStorage;
+		global $wgAvatarsUseSwiftStorage, $wgAvatarsUseService;
 		if ( !empty( $wgAvatarsUseSwiftStorage ) ) {
 			$swift = $this->getSwiftStorage();
 
 			$avatarRemotePath = $this->getLocalPath();
 			$res = $swift->exists( $avatarRemotePath );
-		} else {
```
ghost, line 500: as there is no `else` now, therefore no default case, please initialise to silent `Notice: Undefined variable: res in extensions/wikia/UserProfilePageV3/Masthead.class.php on line 512` or assert that at least one of `$wgAvatarsUseSwiftStorage` and `$wgAvatarsUseService` is not empty.

extensions/wikia/UserProfilePageV3/Masthead.class.php
```diff
@@ -726,67 +739,65 @@ private function postProcessImageInternal( $sTmpFile, &$errorNo = UPLOAD_ERR_OK,
...
-			wfProfileOut( __METHOD__ );
-
-			return UPLOAD_ERR_CANT_WRITE;
-		}
-
+		// render the pre-processed avatar to a PNG file
+		$sFilePath = $this->getTempFile();
 		if ( !imagepng( $oImg, $sFilePath ) ) {
-			wfDebugLog( "avatar", __METHOD__ . ": Cannot save png Avatar: $sFilePath", true );
+			wfDebugLog("avatar", __METHOD__ . ": Cannot save png Avatar: $sFilePath", true);
```
ghost: apply MW coding convention, please

extensions/wikia/UserProfilePageV3/UserProfilePageController.class.php
```diff
@@ -498,20 +499,31 @@ public function saveUsersAvatar( $userId = null, $data = null ) {
...
 					break;
 			}
 
-			// TODO: $user->getTouched() get be used to invalidate avatar URLs instead
-			$user->setGlobalAttribute( 'avatar_rev', date( 'U' ) );
-			$user->saveSettings();
+			// user avatars service updates user preferences on its own
+			if ( empty( $wgAvatarsUseService ) ) {
+				// TODO: $user->getTouched() get be used to invalidate avatar URLs instead
+				$user->setGlobalAttribute('avatar_rev', date('U'));
```
ghost: apply MW coding conventions, please

extensions/wikia/UserProfilePageV3/UserProfilePageController.class.php
```diff
@@ -766,13 +778,18 @@ public function uploadByUrl( $url, $userData, &$errorMsg = '' ) {
...
-			$userData['avatar'] = $localPath;
-			$userIdentityBox->saveUserData( $userData );
+
+			// user avatars service updates user preferences on its own
+			global $wgAvatarsUseService;
+			if ( empty( $wgAvatarsUseService ) ) {
+				/**
+				 * @var $userIdentityBox UserIdentityBox
+				 */
+				$userIdentityBox = new UserIdentityBox($user);
```
ghost: apply MW coding conventions, please

extensions/wikia/UserProfilePageV3/UserProfilePageController.class.php
```diff
@@ -766,13 +778,18 @@ public function uploadByUrl( $url, $userData, &$errorMsg = '' ) {
...
+				$userData = $userIdentityBox->getFullData();
+				$userData['avatar'] = $localPath;
+				$userIdentityBox->saveUserData($userData);
```
ghost: apply MW coding conventions, please

extensions/wikia/UserProfilePageV3/api/UserAvatarsService.class.php
```diff
@@ -0,0 +1,120 @@
+<?php
```
ghost, line 1: Nice, clean class, short methods, love it! :heart:

extensions/wikia/UserProfilePageV3/maintenance/migrateAvatarsViaService.php
```diff
@@ -0,0 +1,224 @@
...
+				\Wikia\Logger\WikiaLogger::instance()->error( __CLASS__, [
+					'exception' => $e,
+					'user_id'   => $user->getId(),
+				] );
+
+				$this->output( sprintf( "\n\t%s", $e->getMessage() ) );
+			}
+		}
+
+		$this->output( "Dome!\n" );
```
ghost: You mean http://dictionary.reference.com/browse/dome ? :-)

Kopuła!

extensions/wikia/UserProfilePageV3/maintenance/migrateAvatarsViaService.php
```diff
@@ -0,0 +1,224 @@
...
+
+			if ( $this->isDryRun ) return;
+
+			$avatarContent = Http::get( $avatarUrl, 'default', [ 'noProxy' => true ] );
+			if ( empty( $avatarContent ) ) {
+				throw new AvatarsMigratorException( 'Avatar fetch failed' );
+			}
+
+			// store it in temporary file
+			$tmpFile = tempnam( wfTempDir(), 'avatar' );
```
ghost, line 133: Is it worth doing some randomization here to prevent race conditions (two migrations running at the same time or something)?

extensions/wikia/UserProfilePageV3/maintenance/migrateAvatarsViaService.php
```diff
@@ -0,0 +1,224 @@
...
+		else if ( !self::isNewAvatar( $avatar ) ) {
+			$service = new UserAvatarsService( $user->getId() );
+
+			// fetch the user-uploaded avatar
+			$masthead = Masthead::newFromUser( $user );
+			$avatarUrl = $masthead->getPurgeUrl();
+
+			$this->output( sprintf( 'uploading <%s>', $avatarUrl ) );
+
+			if ( $this->isDryRun ) return;
```
ghost, line 125: I don't quite get the dry-run check here. What about the above cases and actions?

lib/Wikia/src/Vignette/UrlGenerator.php
```diff
@@ -33,34 +33,34 @@ class UrlGenerator {
 	const REVISION_LATEST = 'latest';
 
 	/** @var UrlConfig */
-	private $config;
+	protected $config;
```
ghost, line 36: You see? This is exactly why `private` parts should have been `protect`-ed in the first place. :-)

extensions/wikia/UserProfilePageV3/Masthead.class.php
macbre, line 500: Modified the check in d0aa9b6

extensions/wikia/UserProfilePageV3/maintenance/migrateAvatarsViaService.php
macbre: Dome removed in 678c45e

extensions/wikia/UserProfilePageV3/maintenance/migrateAvatarsViaService.php
macbre, line 133: No needed as we're going to run a single migration process that will process accounts sequentially.

extensions/wikia/UserProfilePageV3/maintenance/migrateAvatarsViaService.php
macbre, line 125: `setAvatarUrl` method checks dry-run internally.

extensions/wikia/UserProfilePageV3/Masthead.class.php
macbre: `stylize.php` applied and commited in d97c069

PR #8183, opened 2015-08-21
CE-2512: Move update status function to API controller and add permis…

lukaszkonieczny:
…sions check

### DIFF
--- a/extensions/wikia/ContentReview/controllers/ContentReviewApiController.class.php
+++ b/extensions/wikia/ContentReview/controllers/ContentReviewApiController.class.php
@@ -108,13 +108,13 @@ class ContentReviewApiController extends WikiaApiController {
 			throw new BadRequestApiException();
 		}
 
-		$pageId = $this->request->getInt( 'pageId' );
-		$wikiId = $this->request->getInt( 'wikiId' );
-		$status = $this->request->getInt( 'status' );
-
 		if ( !$this->wg->User->isAllowed( 'content-review' ) ) {
 			throw new PermissionsException( 'content-review' );
 		}
+
+		$pageId = $this->request->getInt( 'pageId' );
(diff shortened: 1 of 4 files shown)


Grunny:
:+1:

extensions/wikia/ContentReview/controllers/ContentReviewApiController.class.php
```diff
@@ -94,6 +95,31 @@ public function disableTestMode() {
...
+			|| !$this->wg->User->matchEditToken( $this->request->getVal( 'editToken' ) )
+		) {
+			throw new BadRequestApiException();
+		}
+
+		$pageId = $this->request->getInt( 'pageId' );
+		$wikiId = $this->request->getInt( 'wikiId' );
+		$status = $this->request->getInt( 'status' );
+
+		if ( !$this->wg->User->isAllowed( 'content-review' ) ) {
```
Grunny, line 111: I'd move this above getting the parameters from the request though, since we don't have to get them when the request will fail and they won't be used.

extensions/wikia/ContentReview/scripts/contentReviewSpecialPage.js
```diff
@@ -23,14 +23,16 @@ define(
         }
 
         function updateReviewStatus() {
-            var self = $(this);
-            var data = {
-                wikiId: self.attr("data-wiki-id"),
-                pageId: self.attr("data-page-id"),
-                status: self.attr("data-status")
-            };
+            var self = $(this),
```
Grunny: I also think the variable could be something other than `self` which is usually used as an alias for `this` when you need to deal with scope. In this case something like `$button` or something related to what the jQuery element you're storing in it is would probably be better.

PR #8184, opened 2015-08-21
(CE-2260) Display a warning on JS pages when site JS is disabled

Grunny:
Display a warning on JS pages when site JS is disabled for that wikia.


lukaszkonieczny:
LGTM

PR #8185, opened 2015-08-21
(CE-2516) Make site and user JS parameters respect default config

Grunny:
Make the URL parameters for loading site and user JS respect the default
config so that setting the default config for these variables to false
cannot be overridden per request.


lukaszkonieczny:
LGTM

PR #8186, opened 2015-08-21
ADEN-2340 Report AdBlock status

fraszczakszymon:
cc: @harnash


extensions/wikia/AnalyticsEngine/js/universal_analytics.js
```diff
@@ -347,6 +353,15 @@
     // Unleash
     _gaWikiaPush(['send', 'pageview']);
 
+    if (window.ads.context.opts.showAds) {
```
nandy-andy: What if `window.ads` isn't set at this point? Is it possible?

nandy-andy:
There is only one question I had. Nef also replied in the ticket when you can address both things we can do another quick CR and move it to QA.

harnash:
LGTM

extensions/wikia/AnalyticsEngine/js/universal_analytics.js
fraszczakszymon: I've added extra condition, but I'm not sure if it's possible. It's set at the php level - https://github.com/Wikia/app/blob/dev/extensions/wikia/AdEngine/AdEngine2ContextService.class.php#L45

nandy-andy: Cool, thanks :) It's safer that way, I think. Since both AdEngine and AnalyticsEngine use the same hook to add scripts to head:
- https://github.com/Wikia/app/blob/dev/extensions/wikia/AdEngine/AdEngine2Hooks.class.php#L81,
- https://github.com/Wikia/app/blob/dev/extensions/wikia/AnalyticsEngine/AnalyticsProviderGoogleUA.php#L20.

So, there was a chance of a race condition and it was possible that some users would have JS error in their browsers. 

Now, I only wonder if we shouldn't log to Kibana how many such race condition happens. Do we have a Kibana JS client? :smiling_imp:

nandy-andy:
LGTM :+1:

nandy-andy:
The changes touch only JS. The failing PHP unit tests (`PreferenceKeyValueTest::testSetPreferenceSuccess` and `PreferenceKeyValueTest::testSetWithError`) aren't failing because the changes.

PR #8189, opened 2015-08-23
VOLDEV-148 Verbatim Replacement: Twitter Widget

ty-a:
@Grunny

### DIFF
--- a/extensions/wikia/TwitterTag/TwitterTagHooks.class.php
+++ b/extensions/wikia/TwitterTag/TwitterTagHooks.class.php
@@ -3,7 +3,7 @@
 class TwitterTagHooks {
 	
 	static public function onParserFirstCallInit( Parser &$parser ) {
-		$parser->setHook( 'twitter', array( __CLASS__, 'parseTwitterTag' ));
+		$parser->setHook( 'twitter', [ __CLASS__, 'parseTwitterTag' ] );
 		return true;
 	}
 	
@@ -45,10 +45,11 @@ class TwitterTagHooks {
 		if( !empty( $args['alt-text'] ) ) {
 			$html = Html::element( 'a', $attributes, $args['alt-text'] );
 		} else {
-			$html = Html::element( 'a', 
-									$attributes, 
-									wfMessage( 'twittertag-alt-text' )->params( $args['user'] )->parse() 
-					);
+			$html = Html::element( 
+				'a', 
+				$attributes, 
+				wfMessage( 'twittertag-alt-text' )->params( $args['user'] )->parse() 
+			);
 		}
 
 		$parser->getOutput()->addModules( 'ext.TwitterTag' );
--- a/extensions/wikia/TwitterTag/TwitterTagHooks.class.php
+++ b/extensions/wikia/TwitterTag/TwitterTagHooks.class.php
@@ -10,7 +10,7 @@ class TwitterTagHooks {
 	 * @param $parser - The parser
 	 * @return true
 	 */
-	static public function onParserFirstCallInit( Parser &$parser ) {
+	static public function onParserFirstCallInit( Parser $parser ) {
 		$parser->setHook( 'twitter', [ __CLASS__, 'parseTwitterTag' ] );
 		return true;
 	}
--- a/extensions/wikia/TwitterTag/TwitterTagHooks.class.php
+++ b/extensions/wikia/TwitterTag/TwitterTagHooks.class.php
@@ -2,37 +2,52 @@
 
 class TwitterTagHooks {
 	
+	/*
+	 * onParserFirstCallInit
+	 *
+	 * Registers the <twitter> tag with the parser and sets its callback
+	 *
+	 * @param $parser - The parser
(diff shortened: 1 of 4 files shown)


extensions/wikia/TwitterTag/TwitterTagHooks.class.php
```diff
@@ -0,0 +1,74 @@
...
+	 * @param $input - not used
+	 * @param $args - The attributes to the tag in an assoc array
+	 * @param $parser - not used
+	 * @param $frame - not used
+	 * @return $html - The HTML for the twitter tag
+	 */
+	static public function parseTwitterTag( $input, array $args, Parser $parser, PPFrame $frame ) {
+		if( empty( $args['user'] ) ) {
+			$html = wfMessage( "twittertag-nouser" )->parse();
+			return "<strong class='error'>" . $html . "</strong>";
```
Grunny: The coding convention is to [prefer single quotes unless you're doing string interpolation](https://www.mediawiki.org/wiki/Manual:Coding_conventions/PHP#String_literals). So, this (and the one below) should either be single quotes, or double quotes but use interpolation.

extensions/wikia/TwitterTag/TwitterTagHooks.class.php
```diff
@@ -0,0 +1,74 @@
...
+			$attributes['data-aria-polite'] = $args['aria-polite'];
+		}
+		
+		if( !empty( $args['alt-text'] ) ) {
+			$html = Html::element( 'a', $attributes, $args['alt-text'] );
+		} else {
+			$html = Html::element( 
+				'a', 
+				$attributes, 
+				wfMessage( 'twittertag-alt-text' )->params( $args['user'] )->parse()
```
Grunny: The contents parameter of `Html::element` is escaped so parsing would result in double escaping depending on the text. Is there a reason this is parsed rather than just `->plain` or `->text` (if you need GENDER and so forth)?

extensions/wikia/TwitterTag/TwitterTagHooks.class.php
```diff
@@ -0,0 +1,74 @@
...
+		$quickChecks = [
+			'height' => '',
+			'width' => '',
+			'data-chrome' => '',
+		];
+		
+		$attributes = array_intersect_key( $args, $quickChecks );
+		
+		$attributes['class'] = "twitter-timeline";
+		$attributes['href'] = "https://twitter.come/" . $args['user'];
```
Grunny: `.come` -> `.com`.

And `$args['user']` should probably be URL encoded, and single quotes per above.

extensions/wikia/TwitterTag/TwitterTagHooks.class.php
```diff
@@ -0,0 +1,74 @@
...
+		
+		$quickChecks = [
+			'height' => '',
+			'width' => '',
+			'data-chrome' => '',
+		];
+		
+		$attributes = array_intersect_key( $args, $quickChecks );
+		
+		$attributes['class'] = "twitter-timeline";
```
Grunny: Should be single quotes per conventions.

extensions/wikia/TwitterTag/TwitterTagHooks.class.php
```diff
@@ -0,0 +1,74 @@
...
+	 *   Then constructs the HTML after seeing which attributes are in use. 
+	 *
+	 * @param $input - not used
+	 * @param $args - The attributes to the tag in an assoc array
+	 * @param $parser - not used
+	 * @param $frame - not used
+	 * @return $html - The HTML for the twitter tag
+	 */
+	static public function parseTwitterTag( $input, array $args, Parser $parser, PPFrame $frame ) {
+		if( empty( $args['user'] ) ) {
```
Grunny: Conventions require a space between the `if` and the left parenthesis (`if(` -> `if (`).

extensions/wikia/TwitterTag/TwitterTagHooks.class.php
```diff
@@ -0,0 +1,74 @@
...
+			$html = wfMessage( "twittertag-nouser" )->parse();
```
Grunny: Should be single quotes.

extensions/wikia/TwitterTag/TwitterTagHooks.class.php
```diff
@@ -0,0 +1,74 @@
...
+	
+	/*
+	 * onParserFirstCallInit
+	 *
+	 * Registers the <twitter> tag with the parser and sets its callback
+	 *
+	 * @param $parser - The parser
+	 * @return true
+	 */
+	static public function onParserFirstCallInit( Parser &$parser ) {
```
Grunny: Objects are effectively passed by reference in PHP anyway, so you don't need the `&`.

ty-a: Okay.

Grunny:
One minor comment, otherwise LGTM.

PR #8203, opened 2015-08-25
CE-2524 change reviewer id to name

MariuszCz:


extensions/wikia/ContentReview/controllers/ContentReviewSpecialController.class.php
```diff
@@ -59,6 +59,9 @@ private function prepareReviewData( $reviewsRaw ) {
 					? wfMessage( 'content-review-special-start-review' )->escaped()
 					: wfMessage( 'content-review-special-continue-review' )->escaped();
 
+				if ( !empty( $review['review_user_id'] ) ) {
+					$review['review_user_id'] = User::newFromId( $review['review_user_id'] )->getName();
```
adamkarminski: Since it's not a `user_id` anymore we should rename the key e.g. to `review_user_name`

PR #8205, opened 2015-08-25
 CE-2532: Serve empty string if there is no approved revision id

lukaszkonieczny:


extensions/wikia/ContentReview/ContentReviewHelper.php
```diff
@@ -16,7 +16,14 @@ public function getSiteJsScriptsHash() {
 			$maxTimestamp = max( $maxTimestamp, $revision['touched'] );
 		}
 
-		return $maxTimestamp;
+		if ( empty( $maxTimestamp ) ) {
+			return 0;
+		}
+
+		$datetime = new \DateTime( $maxTimestamp );
+		$timestamp = $datetime->getTimestamp();
```
adamkarminski: I think we should use `$datetime->format( 'Y-m-d H:i:s' )` so that the returned format is always the same as the one in the database under the `touched` field.

extensions/wikia/ContentReview/ContentReviewHelper.php
```diff
@@ -28,6 +35,10 @@ public function getReviewedRevisionIdFromText( $pageName ) {
 		$currentRevisionModel = new Models\CurrentRevisionModel();
```
adamkarminski, line 35: I know it's not a part of this PR but I generally believe that we should use the controller to get this data (check the `ContentReviewApiController::getLatestReviewedRevision()` method).

adamkarminski:
Beside the side-note it LGTM

kamilkoterba:
tested

PR #8206, opened 2015-08-25
Dat 3139 fix infobox params API

dianafa:
https://wikia-inc.atlassian.net/browse/DAT-3139


extensions/wikia/PortableInfobox/services/PortableInfoboxDataService.class.php
```diff
@@ -113,13 +113,32 @@ protected function getInfoboxes( $text ) {
...
 	 * @param $text string template text
-	 * @return mixed
+	 * @return string
 	 */
 	protected function getIncludeonlyText( $text ) {
 		preg_match_all( "/<includeonly>(.+)<\/includeonly>/sU", $text, $result );
 
-		return isset( $result[1][0] ) ? $result[1][0] : false;
+		if ( !isset( $result[1][0] ) ) {
+			return null;
```
idradm, line 125: This is not consistent with previous approach, why the change?

extensions/wikia/PortableInfobox/services/PortableInfoboxDataService.class.php
```diff
@@ -113,13 +113,32 @@ protected function getInfoboxes( $text ) {
...
+		return $result;
+
+	}
+
+	/**
+	 * @desc for given template text returns it without text in <nowiki> and <pre> tags
+	 * @param $text string
+	 * @return string
+	 */
+	protected function removeNowikiPre( $text ) {
```
idradm, line 138: Are you sure those are the only cases?

extensions/wikia/PortableInfobox/services/PortableInfoboxDataService.class.php
dianafa, line 125: we don't want to proceed if there are no includeonly tags

idradm, line 125: Previously you were returning false here, right now it's null. Is there a reason for this change? Will the code behave the same way?

idradm:
:ship: it

dianafa:
tested, merging.

Failing rss test not related.

PR #8207, opened 2015-08-25
CE-2528 hide buttons after approve/reject and show only if status is …

MariuszCz:
…in review
Throw exception when we click twice on approve/reject buttons.

### DIFF
--- a/extensions/wikia/ContentReview/ContentReview.hooks.php
+++ b/extensions/wikia/ContentReview/ContentReview.hooks.php
@@ -76,40 +76,46 @@ class Hooks {
 	}
 
 	public static function onArticleContentOnDiff( $diffEngine, \OutputPage $output ) {
-		global $wgTitle, $wgCityId;
+		global $wgTitle, $wgCityId, $wgRequest;
 
 		if ( $wgTitle->inNamespace( NS_MEDIAWIKI )
 			&& $wgTitle->isJsPage()
 			&& $wgTitle->userCan( 'content-review' )
 		) {
-			\Wikia::addAssetsToOutput( 'content_review_diff_page_js' );
-			\JSMessages::enqueuePackage( 'ContentReviewDiffPage', \JSMessages::EXTERNAL );
+			$reviewModel = new ReviewModel();
+			$reviewData = $reviewModel->getReviewedContent( $wgCityId,
+				\Title::newFromText( $wgTitle->getArticleID() ), ReviewModel::CONTENT_REVIEW_STATUS_IN_REVIEW );
+			$diff = $wgRequest->getInt( 'diff' );
+			if ( !empty( $reviewData ) && (int)$reviewData['revision_id'] === $diff ) {
 
-			$output->prependHTML(
-				\Xml::element( 'button',
-					[
-						'class' => 'content-review-diff-button',
-						'data-wiki-id' => ( $wgCityId ),
-						'data-page-id' => \Title::newFromText( $wgTitle->getArticleID() ),
-						'data-status' => ReviewModel::CONTENT_REVIEW_STATUS_REJECTED
-					],
-					wfMessage( 'content-review-diff-reject' )->plain()
-				)
-			);
+				\Wikia::addAssetsToOutput( 'content_review_diff_page_js' );
+				\JSMessages::enqueuePackage( 'ContentReviewDiffPage', \JSMessages::EXTERNAL );
 
-			$output->prependHTML(
-				\Xml::element( 'button',
-					[
-						'class' => 'content-review-diff-button',
-						'data-wiki-id' => ( $wgCityId ),
-						'data-page-id' => \Title::newFromText( $wgTitle->getArticleID() ),
-						'data-status' => ReviewModel::CONTENT_REVIEW_STATUS_APPROVED
-					],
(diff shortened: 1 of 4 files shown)


extensions/wikia/ContentReview/ContentReview.hooks.php
```diff
@@ -76,40 +76,47 @@ public static function onBeforePageDisplay( \OutputPage $out, \Skin $skin ) {
...
-						'data-status' => ReviewModel::CONTENT_REVIEW_STATUS_APPROVED
-					],
-					wfMessage( 'content-review-diff-approve' )->plain()
-				)
-			);
+			$reviewModel = new ReviewModel();
+			$pageId = \Title::newFromText( $wgTitle->getArticleID() );
+			$reviewData = $reviewModel->getReviewedContent( $wgCityId, $pageId, ReviewModel::CONTENT_REVIEW_STATUS_IN_REVIEW );
+			$diff = $wgRequest->getInt( 'diff' );
+			if ( !empty( $reviewData ) && (int)$reviewData['revision_id'] === $diff ) {
```
adamkarminski: You do not need the `(int)` type casting here, it is covered by the `->getInt()` method.

kamilkoterba:
Button sends API request. If we send it twice we end up with error
INSERT\n\t INTO reviewed_content_logs SET wiki_id = NULL, page_id = NULL, revision_id = NULL, status = '3', submit_user_id = NULL, submit_time = NULL, review_user_id = '5338185', review_start = NULL\nFunction: Wikia\ContentReview\Models\ReviewModel:backupCompletedReview\nError: 1048 Column 'wiki_id' cannot be null (10.14.30.100)\n","code":0,"details":""}}
Because we hae empty record from content_review_status table
$review var empty here:
https://github.com/Wikia/app/blob/CE-content-review/extensions/wikia/ContentReview/controllers/ContentReviewApiController.class.php#L150

Can we handle that as a part of this ticket?

extensions/wikia/ContentReview/ContentReview.hooks.php
```diff
@@ -76,40 +76,47 @@ public static function onBeforePageDisplay( \OutputPage $out, \Skin $skin ) {
...
-						'class' => 'content-review-diff-button',
-						'data-wiki-id' => ( $wgCityId ),
-						'data-page-id' => \Title::newFromText( $wgTitle->getArticleID() ),
-						'data-status' => ReviewModel::CONTENT_REVIEW_STATUS_APPROVED
-					],
-					wfMessage( 'content-review-diff-approve' )->plain()
-				)
-			);
+			$reviewModel = new ReviewModel();
+			$pageId = \Title::newFromText( $wgTitle->getArticleID() );
```
adamkarminski: `\Title::newFromText` will return a `Title` object, not a page's ID. What about using just `$wgTitle->getArticleID`?

extensions/wikia/ContentReview/ContentReview.hooks.php
```diff
@@ -76,40 +76,47 @@ public static function onBeforePageDisplay( \OutputPage $out, \Skin $skin ) {
...
+			$reviewData = $reviewModel->getReviewedContent( $wgCityId, $pageId, ReviewModel::CONTENT_REVIEW_STATUS_IN_REVIEW );
```
adamkarminski: It seems that we will need a method to retrieve an ID of a revision that is being reviewed in other places as well. Can you extract this piece to a helper method that can be reused? Also, if you had one method e.g. 'isRevisionInReview()` you could just make it a part of the condition, which would be much cleaner.

extensions/wikia/ContentReview/scripts/contentReviewDiffPage.js
```diff
@@ -40,6 +40,8 @@ define(
 							response.notification,
 							'confirm'
 						);
+						$('.content-review-diff-button').hide();
```
adamkarminski: Regarding the double request that @kamilkoterba has pointed out - I think that the buttons should be hidden **immediately** after one of them is clicked. Then, in case of an error, they could reappear. The similar behavior is used in Special:Flags to prevent such behavior.

extensions/wikia/ContentReview/scripts/contentReviewDiffPage.js
```diff
@@ -29,7 +29,7 @@ define(
 					status: $button.attr('data-status'),
 					editToken: mw.user.tokens.get('editToken')
 				};
-
+			$('.content-review-diff-button').prop("disabled",true);
```
lukaszkonieczny: According our code conventions we should use single quotes in js files

extensions/wikia/ContentReview/ContentReviewHelper.php
```diff
@@ -90,4 +92,17 @@ public static function isContentReviewTestModeEnabled() {
 
 		return $contentReviewTestModeEnabled;
 	}
+
+	public function isPageInReviewProcess() {
```
lukaszkonieczny: I think this method name is a little bit misleading, because it checks is a current page in review process only on diff pages and you cannot use it somewhere else.

kamilkoterba:
Moved here https://github.com/Wikia/app/pull/8222

PR #8214, opened 2015-08-25
Prefs code regen

nmonterroso:
@Wikia/services-team 
The core problem of SERVICES-644 was because the generated code for user-preferences was out of date. This PR regenerates and adds a check to make sure no null preference names come in. A task to create an integration test to make sure the round-trip works is [here](https://wikia-inc.atlassian.net/browse/SERVICES-662).

It's easier to look at this diff using `?w=1`


drsnyder:
LGTM

PR #8216, opened 2015-08-26
SOC-1201 Pull logic out of MediaWiki's attribute service client

jsutterfield:
After turning on shadow mode for the user attribute service, it looks like attributes aren't being saved/deleted from the service properly.

The approach we were taking before was to save attributes to the service as soon as a call to `User::saveGlobalAttribute()` was made. This required some additional logic and assumptions to be made since MW doesn't actually persist attributes until `User::saveOptions()` is called. It seems that by saving those attributes in the service early we were missing some additional logic later on which changes the value or bails out of saving that attribute. That would explain why we we're see discrepancies between the `wikicities.user_properties` table used by MW and the `userattributes.user_attribute` table used by the service.

This ticket then updates the approach so that we wait until `User::saveOptions()` to save the attributes in the service, allowing MW to do all it's additional magic before finally persisting the attribute.

Ticket: https://wikia-inc.atlassian.net/browse/SOC-1201


lib/Wikia/src/Service/User/Attributes/UserAttributes.php
```diff
@@ -126,7 +117,7 @@ public function deleteAttribute( $userId, Attribute $attribute ) {
 
 	private function attributeNotSetForUser( $userId, Attribute $attribute ) {
 		$this->loadAttributes( $userId );
-		return empty( $this->attributes[$userId][$attribute->getName()] );
```
jsutterfield, line 129: This was actually incorrect since it would return true if the _value_ of `$this->attributes[$userId][$attribute->getName()];` was falsey, we only want to know if that key exists or not.

includes/User.php
```diff
@@ -395,9 +397,6 @@ public function loadFromId() {
 			if (isset($data[self::CACHE_PREFERENCES_KEY])) {
 				$this->userPreferences()->setPreferencesInCache($this->mId, $data[self::CACHE_PREFERENCES_KEY]);
 			}
-			if (isset($data[self::CACHE_ATTRIBUTES_KEY])) {
-				$this->userAttributes()->setAttributesInCache($this->mId, $data[self::CACHE_ATTRIBUTES_KEY]);
```
jsutterfield, line 399: So before we were having the `$this->userAttributes()` manage attributes much more actively. Any call to `User::getGetAttribute()` would have `$this->userAttributes()` check it's own cache to see if an attribute was available and return null if not. Because of that, if we loaded a user object from cache, we needed to make sure it's corresponding `$this->userAttributes()` was loaded as well.

Going forward `$this->userAttributes()` is taking a much more background role and is only in charge of loading the attributes initially. Those are then stored in the member variable `mOptions` [here](https://github.com/Wikia/app/pull/8216/files#diff-897fba1dd6b8edc3238762e35d0b7c71R4927) which is where attributes are traditionally stored so the user object can continue to take care of all of the caching it wants. So this line isn't necessary any more.

includes/User.php
```diff
@@ -2741,25 +2715,11 @@ public function setLocalAttribute($attribute, $value, $cityId = null, $sep = '-'
 	 */
 	public function setGlobalAttribute($attribute, $value) {
 		$this->setOptionHelper($attribute, $value);
-		$this->setAttributeInService($attribute, $value);
+		$this->markAttributesToSetInService( $attribute );
```
jsutterfield: The `$this->setOptionHelper` call above will set the attribute in the general `mOptions` array which also contains preferences and flags. Later on when it comes time to actually save these options, we want to know which options were set with this `User::setGlobalAttribute`method. Those options are attributes which we'll then save to the service. See [here](https://github.com/Wikia/app/pull/8216/files#diff-897fba1dd6b8edc3238762e35d0b7c71R5074)

lib/Wikia/src/Service/User/Attributes/UserAttributes.php
```diff
@@ -13,20 +13,15 @@ class UserAttributes {
...
 	/**
 	 * @Inject({
 	 *    Wikia\Service\User\Attributes\AttributeService::class,
-	 *    Wikia\Service\User\Attributes\UserAttributes::DEFAULT_ATTRIBUTES
 	 * })
 	 * @param AttributeService $attributeService
 	 * @param string[string] $defaultAttributes
 	 */
-	public function __construct( AttributeService $attributeService, $defaultAttributes ) {
+	public function __construct( AttributeService $attributeService ) {
```
jsutterfield, line 23: This class used to handle additional logic when it came to default options. That logic is now all handled by the User class so we can remove default options from here.

includes/User.php
```diff
@@ -5055,6 +5059,22 @@ private function isDefaultOption($key, $value) {
...
+		}
+	}
+
+	private function deleteAttributeInService( $attributeName ) {
+		if ( $this->isAttribute( $attributeName ) ) {
+			$this->userAttributes()->setAttribute( $this->getId(), new Attribute( $attributeName ) );
+		}
+	}
+
+	private function isAttribute( $attributeName ) {
```
kvas-damian: this method name doesn't cover logic behind it.

kvas-damian:
One minor comment. Besides that LGTM

includes/User.php
```diff
@@ -5055,6 +5059,22 @@ private function isDefaultOption($key, $value) {
...
 
+	private function setAttributeInService( $attributeName, $attributeValue ) {
+		if ( $this->isAttribute( $attributeName ) ) {
+			$this->userAttributes()->setAttribute( $this->getId(), new Attribute( $attributeName, $attributeValue ) );
+		}
+	}
+
+	private function deleteAttributeInService( $attributeName ) {
+		if ( $this->isAttribute( $attributeName ) ) {
+			$this->userAttributes()->setAttribute( $this->getId(), new Attribute( $attributeName ) );
```
rafalkalinski: shouldn't it be deleteAttribute?

includes/User.php
```diff
@@ -4953,6 +4915,46 @@ protected function loadOptions() {
 		wfRunHooks( 'UserLoadOptions', array( $this, &$this->mOptions ) );
 	}
 
+	private function loadAttributes() {
+		global $wgEnableReadsFromAttributeService;
+
+		if ( !empty( $wgEnableReadsFromAttributeService ) ) {
+			$attributes = $this->userAttributes()->getAttributes($this->getId());
```
rafalkalinski, line 4924: There is a couple of $this->userAttributes()->[something] calls.
Is there a way or would it be possible to cache the userAttributes object without calling the Injector each time?

includes/User.php
jsutterfield: Oops! Yes, it should certainly be deleteAttibute

includes/User.php
jsutterfield, line 4924: Updated :) https://github.com/Wikia/app/commit/73d366ad55770eccce58a27cb5ab6860c208bd9b

rafalkalinski:
LGTM

kvas-damian:
LGTM

includes/User.php
kvas-damian: When docblock is in place method name makes more sense.

rafalkalinski:
LGTM

kvas-damian:
LGTM

PR #8217, opened 2015-08-26
CE-2458: Serve approved revision of JS pages in mediawiki namespace

lukaszkonieczny:
called via import functions

### DIFF
--- a/extensions/wikia/ContentReview/ContentReview.hooks.php
+++ b/extensions/wikia/ContentReview/ContentReview.hooks.php
@@ -114,12 +114,12 @@ class Hooks {
 		return true;
 	}
 
-	public static function onRawPageViewBeforeOutput( \RawAction &$rawAction, &$text ) {
+	public static function onRawPageViewBeforeOutput( \RawAction $rawAction, &$text ) {
 		global $wgCityId;
 
 		$title = $rawAction->getTitle();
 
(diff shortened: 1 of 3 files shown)


extensions/wikia/ContentReview/ContentReview.hooks.php
```diff
@@ -112,4 +113,33 @@ public static function onArticleContentOnDiff( $diffEngine, \OutputPage $output
 
 		return true;
 	}
+
+	public static function onRawPageViewBeforeOutput( \RawAction &$rawAction, &$text ) {
```
Grunny: Objects are effectively passed by reference in PHP anyway, so no need for the `&`.

extensions/wikia/ContentReview/ContentReview.hooks.php
```diff
@@ -112,4 +113,33 @@ public static function onArticleContentOnDiff( $diffEngine, \OutputPage $output
...
+		global $wgCityId;
+
+		$title = $rawAction->getTitle();
+
+		if ( $title->getNamespace() === NS_MEDIAWIKI && $title->isJsPage() ) {
```
Grunny: Preferred method of checking namespaces in MW is `$title->inNamespace( NS_MEDIAWIKI )`.

Grunny:
Two minor comments otherwise LGTM.

PR #8219, opened 2015-08-26
(CE-2559) Add a Verbatim blacklist

Grunny:
Add a Verbatim blacklist of MW pages we don't want admins to either edit
or be able to use with Verbatim.


kamilkoterba:
lgtm

PR #8220, opened 2015-08-26
XW-87 | Display button only when it's possiblie to run modal

vforge:


extensions/wikia/CuratedContent/CuratedContentHooks.class.php
```diff
@@ -74,7 +74,7 @@ private static function isAllowedWikia() {
 		);
 	}
 
-	private static function shouldDisplayCuratedContentToolButton() {
+	public static function shouldDisplayCuratedContentToolButton() {
```
bognix: is hook a good place for this method? given we have CuratadContentHelper which is a bag of stuff maybe creating another one here is not the best idea

skins/oasis/modules/PageHeaderController.class.php
```diff
@@ -153,8 +153,7 @@ protected function getDropdownActions() {
 	private function getCuratedContentButton() {
 		global $wgUser, $wgEnableCuratedContentExt;
 
-		if ( !empty( $wgEnableCuratedContentExt ) && WikiaPageType::isMainPage() &&
-			$wgUser->isAllowed( 'curatedcontent' ) ) {
+		if ( !empty( $wgEnableCuratedContentExt ) && CuratedContentHooks::shouldDisplayCuratedContentToolButton() ) {
```
bognix: maybe this check should happen inside CuratedContent::editButton?

bognix:
LGTM

vforge:
tested

PR #8225, opened 2015-08-26
DAT-3151 P2 fix handle errors in includeonly tag

dianafa:
https://wikia-inc.atlassian.net/browse/DAT-3151


idradm:
LGTM :shipit:

PR #8230, opened 2015-08-26
bumping user version to force db read

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-670

Users are reporting that their preferences are not showing up as expected. I did a db check on some of the users and their data seems fine. I think this has to do with the memcache object being saved from when we were still using the preference object


drsnyder:
lgtm

PR #8231, opened 2015-08-26
Stop writing to user attribute service

jsutterfield:
Turn off writes to the attribute service as it's been disabled. Once we resolve the security concerns, we'll turn on the service again and can start writing to it.


garthwebb:
LGTM

PR #8232, opened 2015-08-26
VE-1601: Bugfix for a scenario when source editor is loaded first (as opposed to loading it on demand when user switch from the visual mode)

inez:


macbre:
:+1:

PR #8233, opened 2015-08-26
Make sure to not call attribute service when saving cache either

jsutterfield:
Was a little bit hasty when putting together https://github.com/Wikia/app/pull/8231, this should have been done as part of that PR.


mklucsarits:
LGTM

PR #8240, opened 2015-08-27
CE-2550 Adding link to provide feedback. 

kamilkoterba:
Prefilled with standard explanation of rejection.

https://wikia-inc.atlassian.net/browse/CE-2550

@Wikia/community-engineering

### DIFF
--- a/extensions/wikia/ContentReview/ContentReview.i18n.php
+++ b/extensions/wikia/ContentReview/ContentReview.i18n.php
@@ -50,8 +50,8 @@ $messages['en'] = [
 
 	'content-review-diff-approve' => 'Approve',
 	'content-review-diff-reject' => 'Reject',
-	'content-review-diff-approve-confirmation' => 'Reviewed code has been approved.',
-	'content-review-diff-reject-confirmation' => 'Reviewed code has been rejected.',
+	'content-review-diff-approve-confirmation' => 'Reviewed code has been approved. Go back to [[w:c:dev:Special:ContentReview|Special:ContentReview]]',
+	'content-review-diff-reject-confirmation' => 'Reviewed code has been rejected. [$1 Provide feedback] or go back to [[w:c:dev:Special:ContentReview|Special:ContentReview]]',
 	'content-review-diff-page-error' => 'Something went wrong. Please try again later.',
 	'content-review-diff-already-done' => 'You are trying to add an changes to the revision that isn\'t "In review" stage.',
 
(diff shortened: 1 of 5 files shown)


extensions/wikia/ContentReview/ContentReview.i18n.php
```diff
@@ -59,6 +59,10 @@
 	'content-review-status-in-review' => 'In review',
 	'content-review-status-approved' => 'Approved',
 	'content-review-status-rejected' => 'Rejected,',
+
+	'content-review-feedback-link-text' => 'Provide feedback',
+	'content-review-rejection-explanation' => '==Submitted script change was rejected==
```
adamkarminski: Do you think we could add an `<a>` anchor with a revision's ID so that we could link directly to it from a right rail module? That would be great!

adamkarminski:
LGTM in general, but I'd great if you could consider my question before merging. Thanks! :)

extensions/wikia/ContentReview/ContentReview.i18n.php
kamilkoterba: Unfortunately functionality for prefilling edit doesn't allow to pass parameters...
https://github.com/Wikia/app/blob/CE-2550-talk-page-link/extensions/wikia/CreatePage/CreatePage.php#L126-126
Spoke to George, and he accepted general message.

adamkarminski: We could make it a parameter of the `content-review-rejection-explanation` message. But if it's too hacky, we can think of another way to link it correctly.

extensions/wikia/ContentReview/ContentReviewHelper.php
```diff
@@ -134,4 +134,19 @@ public function hasPageApprovedId( $wikiId, $pageId, $oldid ) {
...
+	 * @param \Title $title Title object of JS page
+	 * @return string full link to edit page
+	 */
+	public function prepareProvideFeedbackLink( \Title $title ) {
+		$params = [
+			'action' => 'edit',
+			'section' => 'new',
+			'useMessage' => 'content-review-rejection-explanation'
+		];
+		$talkPage = $title->getTalkPage();
```
adamkarminski: You could skip setting the Talk Page object to a variable and just return the URL, what do you think?

adamkarminski:
Minor comment and it LGTM beside it

PR #8244, opened 2015-08-27
Get rid of wgDBtransactions

macbre:
https://www.mediawiki.org/wiki/Manual:$wgDBtransactions - was removed completely in MW 1.20.0.

This was set to false in our case (despite us using transactions-aware InnoDB) potentially causing data inconsistency issues like [PLATFORM-1311](https://wikia-inc.atlassian.net/browse/PLATFORM-1311)

@michalroszka / @wladekb / @owend / @drozdo


Grunny:
:+1:

PR #8246, opened 2015-08-27
XW-75 add set focus method to CuratedContentToolPontoBridge.js

bognix:
**DON'T MERGE**


vforge:
LGTM, if you add some comment.

bognix:
Let's check if it works first and if it does I promise I'll add comment

PR #8247, opened 2015-08-27
CE-2529 Content Review right rail module 2.0 upgrade

adamkarminski:

This PR upgrades the existing right rail module into an informative and visually appealing tool. To achieve this I had to do the following:
1. Change the data retrieved by `ReviewModel::getPageStatus()` so that it returns data on the Latest, Last and Live revisions.
2. Adjust the `ContentReviewModuleController.class.php` to interpret the data to a set of status messages.
3. Write unit tests to check the logic behind the interpretation process.
4. Modify what we store in the database - **the last** approved and rejected revisions are stored in the `content_review_status` table so it is easy to retrieve all necessary info in 1 query.
5. Add styling and modify JS to interact well with the new module.

Besides that I made some changes in naming which I found confusing and cleaned up i18n documentation.

Have a good review!

Yours,
Adam

### DIFF
--- a/extensions/wikia/ContentReview/controllers/ContentReviewSpecialController.class.php
+++ b/extensions/wikia/ContentReview/controllers/ContentReviewSpecialController.class.php
@@ -3,7 +3,7 @@ use  Wikia\ContentReview\Models\ReviewModel;
 
 class ContentReviewSpecialController extends WikiaSpecialPageController {
 
-	public static $status = [
+	public static $statusMessageKeys = [
 		ReviewModel::CONTENT_REVIEW_STATUS_UNREVIEWED => 'content-review-status-unreviewed',
 		ReviewModel::CONTENT_REVIEW_STATUS_IN_REVIEW => 'content-review-status-in-review',
 		ReviewModel::CONTENT_REVIEW_STATUS_APPROVED => 'content-review-status-approved',
@@ -35,6 +35,8 @@ class ContentReviewSpecialController extends WikiaSpecialPageController {
 			$this->displayRestrictionError();
 			return false;
 		}
+
+		$this->wg->Out->setPageTitle( wfMessage( 'content-review-special-title' )->escaped() );
 
 		$model = new ReviewModel();
 		$reviews = $model->getContentToReviewFromDatabase();
@@ -74,5 +76,4 @@ class ContentReviewSpecialController extends WikiaSpecialPageController {
 
 		return $reviews;
 	}
-
 }
--- a/extensions/wikia/ContentReview/controllers/ContentReviewSpecialController.class.php
+++ b/extensions/wikia/ContentReview/controllers/ContentReviewSpecialController.class.php
@@ -36,7 +36,7 @@ class ContentReviewSpecialController extends WikiaSpecialPageController {
 			return false;
 		}
 
-		$this->wg->Out->setPageTitle( wfMessage( 'content-review-special-title' )->escaped() );
+		$this->getOutput()->setPageTitle( wfMessage( 'content-review-special-title' )->plain() );
 
 		$model = new ReviewModel();
 		$reviews = $model->getContentToReviewFromDatabase();
--- a/extensions/wikia/ContentReview/controllers/templates/ContentReviewSpecialController_index.php
+++ b/extensions/wikia/ContentReview/controllers/templates/ContentReviewSpecialController_index.php
@@ -1,7 +1,7 @@
 <div class="content-review-special-header">
 	<div class="content-review-special-header-content">
 		<h1 class="content-review-special-header-content-title">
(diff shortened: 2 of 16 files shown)


extensions/wikia/ContentReview/controllers/ContentReviewSpecialController.class.php
```diff
@@ -36,6 +36,8 @@ public function index() {
 			return false;
 		}
 
+		$this->wg->Out->setPageTitle( wfMessage( 'content-review-special-title' )->escaped() );
```
Grunny: `$this->getOutput()` is preferred over accessing the global when it's available, which it is in special pages. Also, `OutputPage::setPageTitle` sanitizes and escapes its input, so no need to use `->escaped()`.

adamkarminski: :+1:

extensions/wikia/ContentReview/scripts/contentReviewModule.js
```diff
@@ -58,7 +45,7 @@ define(
 						} else {
 							errorMsg = e.exception.message;
 						}
-						notification = new BannerNotification(
+						var notification = new BannerNotification(
```
Grunny: `var` declaration should be at the start of the function scope per conventions.

adamkarminski: :+1: (it actually is :) )

skins/oasis/modules/ContentReviewModuleController.class.php
```diff
@@ -12,10 +20,118 @@ class ContentReviewModuleController extends WikiaController {
...
+			],
+			[
+				'diff' => $revisionId,
+			]
+		);
+	}
 
-		$this->moduleType = $params['moduleType'];
-		$this->isTestModeEnabled = Wikia\ContentReview\Helper::isContentReviewTestModeEnabled();
+	public function createRevisionTalkpageLink( $revisionId ) {
```
Grunny: These two methods should probably be private as they're an implementation detail.

extensions/wikia/ContentReview/controllers/templates/ContentReviewSpecialController_index.php
```diff
@@ -26,7 +26,7 @@
 				<td><?= htmlspecialchars( $review['wiki'] ) ?></td>
 				<td><a href="<?= Sanitizer::encodeAttribute( $review['url'] ) ?>" target="_blank"><?= htmlspecialchars( $review['title'] ) ?></a></td>
```
kamilkoterba, line 27: No reason to decide for user to open link in new tab. Especially that opening in same window doesn't break anything.

extensions/wikia/ContentReview/controllers/templates/ContentReviewSpecialController_index.php
```diff
@@ -36,7 +36,7 @@
 						<?= wfMessage( 'content-review-special-review-open' )->escaped() ?>
 					<? else: ?>
 						<a href="<?= Sanitizer::encodeAttribute( $review['diff'] ) ?>" target="_blank"
```
kamilkoterba, line 38: No reason to decide for user to open link in new tab. Especially that opening in same window doesn't break anything.

skins/oasis/modules/ContentReviewModuleController.class.php
```diff
@@ -12,10 +20,118 @@ class ContentReviewModuleController extends WikiaController {
...
+		return Linker::linkKnown(
+			$this->getContext()->getTitle()->getTalkPage(),
+			wfMessage( 'content-review-rejection-reason-link' )->escaped(),
+			[
+				'target' => '_blank'
```
kamilkoterba: No reason to decide for user to open link in new tab. Especially that opening in same window doesn't break anything.

skins/oasis/modules/ContentReviewModuleController.class.php
```diff
@@ -12,10 +20,118 @@ class ContentReviewModuleController extends WikiaController {
...
+
+		return $templateData;
+	}
+
+	public function createRevisionLink( $revisionId ) {
+		return Linker::linkKnown(
+			$this->getContext()->getTitle(),
+			"#{$revisionId}",
+			[
+				'target' => '_blank'
```
kamilkoterba: No reason to decide for user to open link in new tab. Especially that opening in same window doesn't break anything.

extensions/wikia/ContentReview/controllers/templates/ContentReviewSpecialController_index.php
kamilkoterba, line 27: Actually, I already removed target blank from Special page in separate PR.

adamkarminski:
LGTMd on Slack :)

PR #8248, opened 2015-08-27
INT-35: Wikia In Your Language for Article

mari-mcmurtrie:
If corresponding article page is available, link to the article page instead of link to a main page always.


mari-mcmurtrie:
@nandy-andy, will you CR? Thanks!

extensions/wikia/WikiaInYourLang/WikiaInYourLangController.class.php
```diff
@@ -217,4 +227,16 @@ private function prepareMessage( $sTargetLanguage, $aMessageParams ) {
 
 		return ['desktop' => $sMsg, 'mobile' => $sMsgMobile];
 	}
+
+	private function getArticleURL($sArticleTitle, $city_id) {
```
adamkarminski: Per conventions we stick to upperCamelCase for variables names in PHP :)

adamkarminski:
LGTM :+1:

mari-mcmurtrie:
@adamkarminski Thanks for CR! I have 1 more change (sorry) pushed. If you have time, will you CR again? I have to change cache key so that value is stored per article, not per wiki. @nandy-andy if you can review, also great. Thanks!

extensions/wikia/WikiaInYourLang/modules/ext.wikiaInYourLang.js
```diff
@@ -138,9 +140,9 @@ require(
 			};
 			tracker.track(trackingParams);
 
-			cache.set(targetLanguage + 'WikiaInYourLangMessage' + cacheVersion, null);
+			cache.set(targetLanguage + 'WikiaInYourLangMessage' + articleTitle + cacheVersion, null);
 			// Cache for a month
-			cache.set('wikiaInYourLangNotificationShown' + cacheVersion, true, cache.CACHE_LONG);
+			cache.set('wikiaInYourLangNotificationShown' + articleTitle + cacheVersion, true, cache.CACHE_LONG);
```
Grunny: With this they'll have to dismiss a notification on every new article they view. Should they instead still just be able to dismiss once and not see a notification on any other article on that wikia for a month?

c933103:
Can this feature be turned off completely via user setting?

extensions/wikia/WikiaInYourLang/modules/ext.wikiaInYourLang.js
mari-mcmurtrie: @Grunny: Good point. I will ask Yoko about it.

mari-mcmurtrie:
Currently, no, users do not have a way to configure the feature and it is not required by spec. I think it is nice to have feature so I discuss with Yoko. Thanks!

extensions/wikia/WikiaInYourLang/modules/ext.wikiaInYourLang.js
mari-mcmurtrie: We store cached info (a user closes alert?, etc) for each link address, not each article. There are multiple articles to share the same link address (e.g. main) so those share the same cache value.

mari-mcmurtrie:
@Grunny I implemented the way we talked about. Basically, there are 3 existing caches
- ’wikiaInYourLangRequestSent' (boolean, 1 day) … check if we request to get alert or read from cache
- ’wikiaInYourLangNotificationShown'  (boolean, 1 month) … check if a user closes alert or not
- targetLanguage + 'WikiaInYourLangMessage' (string, null) … Backend response from wikiaInYourLang controller. Basically cached for 1 day

These 3 caches are scoped for "link title" base. "link title" is obtained from back-end response "linkAddress". I added 1 more cache to keep track of article&link title map. This map is per wikia scope. I tested lightly and it behaves expected.

mari-mcmurtrie:
@Grunny will you please CR? Thanks!

mari-mcmurtrie:
@Grunny Please CR, thank you.

extensions/wikia/WikiaInYourLang/modules/ext.wikiaInYourLang.js
```diff
@@ -155,6 +160,43 @@ require(
...
+			var linkAddressAry = linkAddress.split('/'),
+				listOfCachedTitles = '',
+				linkTitle = linkAddressAry[linkAddressAry.length - 1].length === 0 ? 'main' : linkAddressAry[linkAddressAry.length - 1];
+
+			listOfCachedTitles = cache.get(getWIYLLinkTitlesKey());
+			if ( listOfCachedTitles ) {
+				listOfCachedTitles[articleTitle] = linkTitle;
+			} else {
+				listOfCachedTitles = {};
+				listOfCachedTitles[articleTitle] = linkTitle;
```
Grunny: Rather than doing `listOfCachedTitles[articleTitle] = linkTitle;` twice, this could be simplified to something like:

``` javascript
if (!listOfCachedTitles) {
    listOfCachedTitles = {};
}

listOfCachedTitles[articleTitle] = linkTitle;
```

extensions/wikia/WikiaInYourLang/WikiaInYourLangController.class.php
```diff
@@ -218,4 +238,15 @@ private function prepareMessage( $sTargetLanguage, $aMessageParams ) {
 
 		return ['desktop' => $sMsg, 'mobile' => $sMsgMobile];
 	}
+
+	private function getArticleURL( $sArticleTitle, $cityId ) {
+		$articleURL = null;
+		if ( $sArticleTitle !== false ) {
+			$title = GlobalTitle::newFromText( $sArticleTitle, NS_MAIN, $cityId );
+			if ( !is_null( $title ) && $title->exists()) {
```
Grunny: Whitespace per conventions `$title->exists()) {` -> `$title->exists() ) {`

extensions/wikia/WikiaInYourLang/WikiaInYourLangController.class.php
```diff
@@ -65,8 +69,17 @@ public function getNativeWikiaInfo() {
...
+					if ( $wgWikiaEnvironment !== WIKIA_ENV_DEV ) {
+						$articleURL = $this->getArticleURL( $sArticleTitle, $oNativeWiki->city_id );
+						if ( $articleURL ) {
+							$aMessageParams[1] = $articleURL;
+							$this->response->setVal( 'linkAddress', $articleURL );
+							$isMainPageLink = false;
+						}
+					}
+
+					$sMessagesAry = $this->prepareMessage( $sTargetLanguage, $aMessageParams, $isMainPageLink );
```
Grunny: This code is using Hungarian notation (though we don't have any convention of doing (or not doing) so), but if it is going to use Hungarian notation, this variable is an array not a string, so should be something like `$aMessages`.

extensions/wikia/WikiaInYourLang/WikiaInYourLangController.class.php
```diff
@@ -65,8 +69,17 @@ public function getNativeWikiaInfo() {
 						$oNativeWiki->city_url,
 						$oNativeWiki->city_title,
 					];
-
-					$sMessagesAry = $this->prepareMessage( $sTargetLanguage, $aMessageParams );
+					$isMainPageLink = true;
+					if ( $wgWikiaEnvironment !== WIKIA_ENV_DEV ) {
```
Grunny: Is there a reason not to run this code on devboxes?

extensions/wikia/WikiaInYourLang/modules/ext.wikiaInYourLang.js
```diff
@@ -155,6 +160,43 @@ require(
...
+			return targetLanguage + 'WikiaInYourLangMessage' + retrieveLinkTitle() + cacheVersion;
+		}
+
+		function getWIYLLinkTitlesKey() {
+			return targetLanguage + 'WikiaInYourLangLinkTitles' + cacheVersion;
+		}
+
+		function saveLinkTitle(linkAddress) {
+			var linkAddressAry = linkAddress.split('/'),
+				listOfCachedTitles = '',
```
Grunny: Why is this being initialised as a string?

extensions/wikia/WikiaInYourLang/modules/ext.wikiaInYourLang.js
```diff
@@ -155,6 +160,43 @@ require(
...
+				linkTitle = linkAddressAry[linkAddressAry.length - 1].length === 0 ? 'main' : linkAddressAry[linkAddressAry.length - 1];
```
Grunny: What happens if the page name is something like `Foo/Bar` (subpages and titles like that are common)?

extensions/wikia/WikiaInYourLang/modules/ext.wikiaInYourLang.js
```diff
@@ -155,6 +160,43 @@ require(
...
+				listOfCachedTitles[articleTitle] = linkTitle;
+			} else {
+				listOfCachedTitles = {};
+				listOfCachedTitles[articleTitle] = linkTitle;
+			}
+
+			cache.set(getWIYLLinkTitlesKey(),listOfCachedTitles, cache.CACHE_LONG);
+		}
+
+		function retrieveLinkTitle() {
```
hakubo, line 212: can this function accept articleTitle as a parameter instead of using globals?
It took me a while to understand what this function do

extensions/wikia/WikiaInYourLang/WikiaInYourLangController.class.php
mari-mcmurtrie: @nandy-andy helped me around here. I vaguely remember that we did this because in dev box, there is not much article there so it is almost always does nothing. That is why we skip this for dev box. Correct me if I am wrong, @nandy-andy.

extensions/wikia/WikiaInYourLang/modules/ext.wikiaInYourLang.js
mari-mcmurtrie: I thought "/" should be URL-encoded if it is not a path separator.
So basically, what I am trying to do here is:
If linkAddress = "'http://ja.destiny.wikia.com/'" then, linkAddressAry[linkAddressAry.length - 1] = '' (empty string) so linkTitle = 'main'.
If linkAddress = "http://ja.destiny.wikia.com/wiki/Weapon" then, linkAddressAry[linkAddressAry.length - 1] = 'Weapon', so linkTitle = 'Weapon'. 

Do you mean if there is an URL like "http://ja.destiny.wikia.com/wiki/Foo/Bar", not "http://ja.destiny.wikia.com/wiki/Foo%2FBar"?

extensions/wikia/WikiaInYourLang/modules/ext.wikiaInYourLang.js
mari-mcmurtrie: You are right indeed. I just quickly wrote it though and did not improve redundant code.

extensions/wikia/WikiaInYourLang/modules/ext.wikiaInYourLang.js
mari-mcmurtrie, line 212: I should have added comments. 
Article title is a current article title a user is looking at. LinkTitle is wikia-in-your-lang link's article title, which may or may not be the same as the current article title. Cache is used to store the mapping of article title => link title

Example) 
Suppose there are starwars.wikia.com, and ja.starwars.wikia.com, and a user's language is Japanese, and the user visits English starwars wikia. (sounds already complicated!)
1) User is in main page -> we show 'wikia-in-your-lang' with link 'ja.starwars.com', article title is 'main', link title is 'main'.
2) User is in Luke Skywalker article page ->  we show 'wikia-in-your-lang' with link 'ja.starwars.com/wiki/Luke_Skywalker' since there is an corresponding article in Japanese, article title is 'Luke_Skywalker', link title is 'Luke_Skywalker'.
3) User is in Crossroads article page -> We show 'wikia-in-your-lang' with link 'ja.starwars.com' because there is no corresponding article 'Crossroads' in Japanese. Article is 'Crossroadsd', link title is 'main'.

listOfCachedTitles is a mapping of article title => link title. It should look like {'main':'main', 'Luke_skywalker':'Luke_skywalker', 'Crossroad':'main'}.

mari-mcmurtrie, line 212: articleTitle is class variable so I do not have to pass it into function, correct?

extensions/wikia/WikiaInYourLang/modules/ext.wikiaInYourLang.js
Grunny: Yeah, because above you are getting the URL with `->getFullURL()` which does not URL encode slashes in MediaWiki. i.e. for http://starwars.wikia.com/wiki/Slicer/Canon:

```
> $title = GlobalTitle::newFromText( 'Slicer/Canon', NS_MAIN, 147 );

> echo $title->getFullURL();
http://starwars.wikia.com/wiki/Slicer/Canon
```

In that case, `linkAddressAry[linkAddressAry.length - 1]` would be "Canon" which would be incorrect. What you want to strip is native domain + /wiki/ from the start. Maybe since `nativeDomain` is being set as the naive value [here](https://github.com/Wikia/app/blob/36652bbe14d67/extensions/wikia/WikiaInYourLang/WikiaInYourLangController.class.php#L52), it should instead be set as `$oNativeWiki->city_url` which would then be correct URL, and this could simplified just to strip `nativeDomain + 'wiki/'`?

extensions/wikia/WikiaInYourLang/modules/ext.wikiaInYourLang.js
hakubo, line 212: 1. some excerpt of what you wrote could end up in JSDoc
2. I know that you don't have to, but the smaller the scope of a variable the easier it is to reason when it gets changed and harder to introduce bugs. There are 172 lines of code between declaration and use. It's just easy to forget where this variable is coming from.

extensions/wikia/WikiaInYourLang/WikiaInYourLangController.class.php
Grunny: Unless it breaks on devboxes for some reason, I'd still call it, even if it usually won't return a matching page. I think it's better to minimise environmental differences unless absolutely necessary.

extensions/wikia/WikiaInYourLang/modules/ext.wikiaInYourLang.js
mari-mcmurtrie, line 212: Aw I got it about second point you said. You are talking about readability. Let me improve it.

mari-mcmurtrie:
Hi, I addressed issues. @Grunny @hakubo if you guys can take a look again, it will be great. Thanks!

extensions/wikia/WikiaInYourLang/modules/ext.wikiaInYourLang.js
```diff
@@ -155,6 +163,52 @@ require(
...
+
+			cache.set(getWIYLLinkTitlesKey(),listOfCachedTitles, cache.CACHE_LONG);
+		}
+
+		/**
+		 * Retrieve linkTitle for this articleTitle from cache.
+		 * articleTitle is a current article title of this wiki.
+		 * linkTitle is wikia-in-your-lang link's article title, which may or may not be the same as the articleTitle.
+		 * listOfCachedTitles is a map of articleTitle => linkTitle
+		 * @returns {
```
hakubo: `@return {string} wikia-in-your-lang link's article title`

mari-mcmurtrie:
Hi, I addressed issues. @Grunny @hakubo if you guys can take a look again, it will be great. Thanks!

extensions/wikia/WikiaInYourLang/modules/ext.wikiaInYourLang.js
```diff
@@ -155,6 +163,50 @@ require(
...
+
+		function getWIYLLinkTitlesKey() {
+			return targetLanguage + 'WikiaInYourLangLinkTitles' + cacheVersion;
+		}
+
+		function saveLinkTitle(linkAddress) {
+			var articleTitle = w.wgTitle,
+				linkAddressAry = linkAddress.split('/'),
+				listOfCachedTitles = {},
+				linkTitle = linkAddressAry[linkAddressAry.length - 1].length === 0 ? 'main' : linkAddressAry[linkAddressAry.length - 1];
```
Grunny: This part about [handling articles with a forward slash in the title](https://github.com/Wikia/app/pull/8248#discussion_r39253797) isn't addressed yet.

mari-mcmurtrie:
@Grunny I addressed the way to retrieve article title from URL... I did not want to change 'nativeDomain' part since the code had already existed. Let me know if this is ok.

Grunny:
Cool. I can't find anywhere `nativeDomain` is being used, but fixing it this way is OK too. :)

PR #8252, opened 2015-08-27
CE-2506: Serve approved script revisions

lukaszkonieczny:
1. For Gadget extension
2. For import functions
3. Improve serving scripts from ResourceLoader and with action=raw


extensions/wikia/ContentReview/ContentReview.hooks.php
```diff
@@ -30,8 +30,12 @@ public static function onGetRailModuleList( Array &$railModuleList ) {
 		return true;
 	}
 
-	public static function onMakeGlobalVariablesScript( &$vars ) {
-		$vars['contentReviewTestModeEnabled'] = Helper::isContentReviewTestModeEnabled();
+	public static function onMakeGlobalVariablesScript(&$vars) {
```
Grunny: Problem with IDE settings? ;) Since it went from spacey to non-spacey.

Grunny:
Minor comment, otherwise LGTM, so feel free to merge if tested. :)

PR #8253, opened 2015-08-27
ADEN-1957 Taboola recirculation

nandy-andy:
In order to enabled Taboola Recirculation on selected wikis I've added recirculation wiki list in code and new `getTaboolaMode()` function.

### DIFF
--- a/extensions/wikia/AdEngine/js/provider/taboola.js
+++ b/extensions/wikia/AdEngine/js/provider/taboola.js
@@ -16,7 +16,20 @@ define('ext.wikia.adEngine.provider.taboola', [
 		readMoreDiv = document.getElementById('RelatedPagesModuleWrapper'),
 		context = adContext.getContext(),
 		isMobile = context.targeting.skin === 'wikiamobile',
-		pageType = context.targeting.pageType;
+		pageType = context.targeting.pageType,
+		recirculationWikis = {
+			'ageofempires': true,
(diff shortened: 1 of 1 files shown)


extensions/wikia/AdEngine/js/provider/taboola.js
```diff
@@ -16,7 +16,20 @@ define('ext.wikia.adEngine.provider.taboola', [
 		readMoreDiv = document.getElementById('RelatedPagesModuleWrapper'),
 		context = adContext.getContext(),
 		isMobile = context.targeting.skin === 'wikiamobile',
-		pageType = context.targeting.pageType;
+		pageType = context.targeting.pageType,
+		recirculationWikis = {
```
fraszczakszymon, line 21: I checked this list and unfortunately there are some differences between domains and dbNames :-(
`batman` -> `Batman`
`shadowhunters` -> `mortalinstruments`
`dcanimateduniverse` -> `dcanimated`

So maybe we should use here a map with expected Taboola domains instead of `true` values

nandy-andy, line 21: Thanks for checking! I didn't push it to CR because I wanted to check it and happy-path test it ;)

Good idea, will do! :)

fraszczakszymon:
LGTM :+1:

PR #8254, opened 2015-08-27
[SERVICES-672] Helios helper api for sending forgot password emails

ArturKlajnerok:
Dedicated api method is needed for sending temporary password email, as using email controller directly with `Email\\Controller\\ForgotPassword` don't return any http status information and using it with output set to json for additional success/error data in json object also changes the email body to json.   

/cc @Wikia/services-team


nmonterroso:
LGTM, though I wonder if we should start breaking apart `HelperController`, since it's quickly growing into the "stick anything that Helios needs MediaWiki to do", which is a lot of stuff :P

PR #8255, opened 2015-08-27
SOC-1177 Escape a few select message key translations

garthwebb:
https://wikia-inc.atlassian.net/browse/SOC-1177


extensions/wikia/JSMessages/js/JSMessages.js
```diff
@@ -59,6 +59,14 @@
 			return ret;
 		};
 
+		JSMessages.html = function(){
+			var expandedText = this.msg.apply(this, arguments),
+				div = document.createElement('div');
+
+			div.appendChild(document.createTextNode(expandedText));
+			return div.innerHTML;
```
garthwebb: There were a few approaches to escaping I found, most of which either listed out all the things to escape, or used jQuery or something similar to stuff content into the DOM and then remove it in a way that forces an escape of HTML (as does the above).  I chose this way because apparently it forces the browser do the escaping rather than JS so its much faster.

kvas-damian:
@garthwebb what do you think about using existing MW solution? I'd be great if we could avoid creating one more solution for the same problem.

https://www.mediawiki.org/wiki/Manual:Messages_API#Format_options
We already use it in https://github.com/Wikia/app/blob/dev/extensions/wikia/BannerNotifications/js/BannerNotifications.js

garthwebb:
@kvas-damian The MW solution looks good but we already have JSMessages and are using it in a _lot_ of places.  I'm ok adding a 6 line function that duplicates some functionality rather than switch over all of chat to a new system to avoid it.

kvas-damian:
In that case let's leave as it is.
LGTM

Grunny:
This escaping approach has the problem of not escaping quotes, so if a dev uses this in an attribute they may have a false sense of security. A HTML escaping method should also escape both single and double quotes. Take a look at [`mw.html.escape`](https://github.com/Wikia/app/blob/49aeec24f267/resources/mediawiki/mediawiki.js#L1471-L1473) and possibly just make use of that.

kvas-damian:
@Grunny is right again ;)
I changed my mind it does NOT LGTM now.

garthwebb:
@Grunny what is the security risk of quotes when everything else is escaped?  Moving to mw.html.escape will involve changing over all the message registration, and updating all chat js to use these calls, and all the QA that will require.  I'd prefer not to spearhead a move to change a system used by 90% of our code:

```
lemongrab:wikia (SOC-1176) garth$ ack JSMessages | egrep -o '^[a-zA-Z][^/]+' | sort | uniq -c
   2 AdEngine
   2 AdminDashboard
   3 ArticleComments
  12 AssetsManager
   2 BlogEditCategoryPrompter
   2 CategorySelect
   8 Chat2
   1 CreateNewWiki
   5 CuratedContent
   2 EditPageLayout
   3 EditPreview
   3 FacebookClient
   1 FilePage
   2 Flags
   3 Forum
   7 GameGuides
   1 GlobalFooter
   2 ImagePlaceholder
   1 InWikiGame
  49 JSMessages
   2 LicensedVideoSwap
   1 Lightbox
   2 MediaGallery
   3 MiniEditor
   2 Oasis
   7 Places
   2 RTE
   2 Recommendations
   5 RelatedPages
   2 SpecialCss
   2 SpecialEditHub
   2 SpecialManageWikiaHome
   2 SpecialPromote
   2 SpecialStyleguide
   2 SpecialVideos
   1 ThemeDesigner
   5 UserLogin
   1 UserProfilePageV3
   1 Venus
   1 VenusTest
   1 VideoEmbedTool
   3 VideoPageTool
   2 VideosModule
   4 Wall
   2 WikiFeatures
   2 WikiaHomePage
   6 WikiaMaps
   9 WikiaMobile
   4 WikiaMobileEditor
   1 WikiaPhotoGallery
   2 WikiaPoll
   3 WikiaStyleGuide
lemongrab:wikia (SOC-1176) garth$ ack "mw\.(msg|html)" | egrep -o '^[a-zA-Z][^/]+' | sort | uniq -c
   1 AbTesting
  21 InfoboxBuilder
   2 PhalanxII
   4 QuickTools
   6 SemanticMediaWiki
   1 Venus
```

Grunny:
@garthwebb the current method only escapes `<>&`, if a dev puts a message into an attribute in this way, then quotes allow an attacker to break out of the attribute and load JS through other attributes (i.e. onmouseover, onerror, etc.). I'm not suggesting to not use JSMessages, but to use `mw.html.escape` or the same implementation in JSMessages instead of using the `document.createElement` approach, which isn't sufficient. So, all I'm suggesting is to switch to:

``` javascript
        JSMessages.html = function(){
            var expandedText = this.msg.apply(this, arguments);
            return mw.html.escape(expandedText);
        };
```

Or at least recreate the same logic as `mw.html.escape` in here. `mw.html.escape` has no dependencies and isn't tied to `mw.message`. We also have a not as clean method to do the same thing basically called [`$.htmlentities`](https://github.com/Wikia/app/blob/49aeec24f267/resources/wikia/jquery.wikia.js#L335-L338).

kvas-damian:
LGTM

Grunny:
:+1:

bkoval:
Please make sure unit tests don't fail because of our changes (seems like Vignette is to blame) and LGTM.

rafalkalinski:
Please check the uTests and LGTM

rafalkalinski:
LGTM

PR #8256, opened 2015-08-28
SOC-1176 Don't match anything with a double quote to use as a glyph

garthwebb:
https://wikia-inc.atlassian.net/browse/SOC-1176

### DIFF
--- a/extensions/wikia/Chat2/js/emoticons.js
+++ b/extensions/wikia/Chat2/js/emoticons.js
@@ -11,99 +11,138 @@
  *
  */
 
+'use strict';
 
-/** CONSTANTS & VARIABLES **/
+var WikiaEmoticons,
+	EmoticonMapping;
+
 // By explicitly setting the dimensions, this will make sure the feature stays as emoticons instead of getting
 // spammy or inviting disruptive vandalism (19px vandalism probably won't be AS offensive).
-if(typeof WikiaEmoticons == 'undefined'){
+if (typeof WikiaEmoticons === 'undefined'){
 	WikiaEmoticons = {};
 }
 WikiaEmoticons.EMOTICON_WIDTH = 19;
 WikiaEmoticons.EMOTICON_HEIGHT = 19;
 
-/** FUNCTIONS **/
-
 /**
- * Takes in some chat text and an emoticonMapping (which strings should be replaced with images at which URL) and returns
- * the text modified so that the replacable text (eg ":)") is replaced with the HTML for the image of that emoticon.
+ * Takes in some chat text and an emoticonMapping (which strings should be replaced with images at which URL) and
+ * returns the text modified so that the replacable text (eg ":)") is replaced with the HTML for the image of
+ * that emoticon.
  */
-WikiaEmoticons.doReplacements = function(text, emoticonMapping){
-	// This debug is probably noisy, but I added it here just in case all of these regexes turn out to be slow (so we could
-	// see that in the log & know that we need to make this function more efficient).
-	$().log("Processing any emoticons... ");
+WikiaEmoticons.doReplacements = function (text, emoticonMapping) {
+	// This debug is probably noisy, but I added it here just in case all of these regexes turn out
+	// to be slow (so we could see that in the log & know that we need to make this function more efficient).
+	$().log('Processing any emoticons... ');
 
-	var imgUrlsByRegexString = emoticonMapping.getImgUrlsByRegexString();
-	for(var regexString in imgUrlsByRegexString){
-		/*
-		 * empty string for emote icons crash Chat
-		 * so ignore them
-		 */
-		if(regexString == '') continue;
+	var imgUrlsByRegexString = emoticonMapping.getImgUrlsByRegexString(),
+		regexString,
+		imgSrc,
+		numIters,
+		origText,
+		regex,
+		glyphUsed,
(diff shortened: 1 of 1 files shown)


extensions/wikia/Chat2/js/emoticons.js
```diff
@@ -101,7 +101,7 @@ if(typeof EmoticonMapping === 'undefined'){
 					currentKey = url;
 					//$().log("  " + url + "...");
 				} else if (self._settings[currentKey]) {
-					var glyphMatch = emoticonArray[i].match(/^\*\*[ ]*([^*].*)/); // line starting with 2 "**"'s then optional spaces, then some non-empty content.
+					var glyphMatch = emoticonArray[i].match(/^\*\* *([^*"][^"]*)/); // line starting with 2 "**"'s then optional spaces, then some non-empty content.
```
garthwebb: This basically stops matching content from the special emoticon page as soon as a `"` is encountered.  So for the example in the ticket:

```
(angry" onmouseover="alert(0);" "
```

it would match up to `(angry` as the glyph name and ignore everything else on the rest of the line.  If someone enters this in chat, the `(angry` part will trigger an emoticon to be drawn and the `" onmouseover="alert(0);" "` will be treated as chat text and escaped and printed.

Grunny:
This is part of the fix as an input validation step, but we still need to escape at the output, i.e. the [`alt` and `title` attributes](https://github.com/Wikia/app/blob/6e0edd84478/extensions/wikia/Chat2/js/emoticons.js#L55) that the emoticon text is put into should still be escaped properly (either with an appropriate escaping method, or a safe way to escape attributes is to build the output HTML with jQuery and use `.attr`). When preventing XSS, input validation is optional depending on the context, but you should always escape the output regardless. In this case as an example, imagine if someone changes or breaks the input validation/sanitization step here in the future without knowing the consequences.

extensions/wikia/Chat2/js/emoticons.js
kvas-damian: I'm not sure if preventing only against `"` is ideal solution.
I remember really old IE treating ``` as a delimeter so old IE is still vulnerable here.
https://groups.google.com/d/msg/null-co-in/iqxoB7PXT5A/fKIjAsijn0kJ

Escaping output which @Grunny suggested sounds like more bulletproof solution.

rafalkalinski: Agree with Damian

extensions/wikia/Chat2/js/emoticons.js
```diff
@@ -11,99 +11,138 @@
...
-} // end doReplacements()
+}; // end doReplacements()
+
+WikiaEmoticons.buildTagGenerator = function (imgSrc) {
+	return function (match, leading, tag, trailing) {
+		tag = mw.html.escape(tag);
+		return (
+			leading +
+			' <img ' +
+				'src="' + imgSrc + '" ' +
```
kvas-damian, line 92: @garthwebb what do you think about escaping `imgSrc` additionally?
I know it's not a part of this ticket but goes under the same pattern - we escape only input.

rafalkalinski:
except what Damian said, LGTM

extensions/wikia/Chat2/js/emoticons.js
garthwebb, line 92: Actually, I forgot that part of the acceptance criteria was making sure the image source was from Wikia.  I'll do both of those ...

garthwebb:
Updated PR with feedback about imgSrc

extensions/wikia/Chat2/js/emoticons.js
```diff
@@ -11,99 +11,145 @@
...
-	$().log("Done processing emoticons.");
+	$().log('Done processing emoticons.');
 	return text;
-} // end doReplacements()
+};
+
+WikiaEmoticons.buildTagGenerator = function (imgSrc) {
+	return function (match, leading, tag, trailing) {
+		// Don't return any img tag if this is an external image
+		if (!imgSrc.match(/^http:\/\/[^\/]+wikia(?:-dev)?(.com|.nocookie.net)\//)) {
```
kvas-damian: this check does not protect from `http://hackingwikia.com/`.
Additionally I think we should accept HTTPS

extensions/wikia/Chat2/js/emoticons.js
```diff
@@ -11,99 +11,145 @@
...
-	$().log("Done processing emoticons.");
+	$().log('Done processing emoticons.');
 	return text;
-} // end doReplacements()
+};
+
+WikiaEmoticons.buildTagGenerator = function (imgSrc) {
+	return function (match, leading, tag, trailing) {
+		// Don't return any img tag if this is an external image
+		if (!imgSrc.match(/^https?:\/\/(?:[^\/]+\.)*?wikia(?:-dev)?(?:\.com|\.nocookie\.net)\//)) {
```
kvas-damian: @rafalkalinski what about protocol relative URLs?

rafalkalinski: Is there really a chance for that there?

kvas-damian: Yes, but we can leave that for now ;)

kvas-damian:
LGTM

PR #8257, opened 2015-08-28
replace User::isCurrent() with MW's User::equals()

ghost:
Per conversation with @wladekb - the refactor of the `User` class
in the way I initially had in mind would be either too complex
or not enough reliable (due to nature of PHP, objects passed
by reference, object cloning, etc).

@macbre


macbre:
LGTM

PR #8258, opened 2015-08-28
ArticleComment::doAfterPost - log failed comments saves

macbre:
ArticleComments seems to be the most affected by [PLATFORM-1311](https://wikia-inc.atlassian.net/browse/PLATFORM-1311). Log failed saves to help us find the root cause of this Very Interesting P3™.

@michalroszka


ghost:
LG™

PR #8263, opened 2015-08-28
(CE-2509,CE-2510) Tweaks to personal JS permissions

Grunny:
Add a preference requiring users to explicitly opt-in to personal JS. Allow
users to edit their personal JS, but only the main skin JS files.

### DIFF
--- a/extensions/wikia/ProtectSiteJS/ProtectSiteJS.class.php
+++ b/extensions/wikia/ProtectSiteJS/ProtectSiteJS.class.php
@@ -37,8 +37,9 @@ class ProtectSiteJS {
 		global $wgCityId;
 		$allowedJsSubpages = [
 			'common',
-			'wikia',
 			'monobook',
+			'wikia',
+			'uncyclopedia',
 		];
 
 		if ( $wgCityId == Wikia::COMMUNITY_WIKI_ID ) {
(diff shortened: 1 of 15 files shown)


extensions/wikia/ProtectSiteJS/ProtectSiteJS.class.php
```diff
@@ -0,0 +1,60 @@
...
+			'wikia',
+			'monobook',
+		];
+
+		if ( $wgCityId == Wikia::COMMUNITY_WIKI_ID ) {
+			$allowedJsSubpages[] = 'global';
+		}
+
+		$allowedJsSubpages = implode( '|', $allowedJsSubpages );
+		$regex = '/^' . preg_quote( $user->getName(), '/' ) . '\/(' . $allowedJsSubpages . ')\.js$/';
```
lukaszkonieczny, line 50: not sure, but should this regex has case-insensitive option?

Grunny, line 50: Nope, since it only loads lower case `/common.js`, `/wikia.js`, `/monobook.js`, and `/global.js`. Upper case won't work for any of them, and we only want to allow editing to these specific pages. :)

lukaszkonieczny, line 50: :+1:

lukaszkonieczny:
LGTM

PR #8268, opened 2015-08-28
INT-169: Bug fix - login user can upload photo when anonymou…

mari-mcmurtrie:
…s edit is disabled.

The condition had a bug: it was not allow upload at all when anonymous edit was disabled, regardless of login status.

### DIFF
--- a/extensions/VisualEditor/wikia/ApiAddMediaTemporary.php
+++ b/extensions/VisualEditor/wikia/ApiAddMediaTemporary.php
@@ -42,11 +42,7 @@ class ApiAddMediaTemporary extends ApiAddMedia {
 
 			// If wiki is Japanese content, then we check if anonymous edit is allowed. INT-158
 			// This condition will be changed as soon as Mercury has login for all wikis.
-			if ( $wgContLanguageCode === 'ja' ) {
-				if ( $wgDisableAnonymousEditing && !$this->mUser->isLoggedIn()) {
-					$this->dieUsageMsg( 'noedit-anon' );
-				}
-			} else {
+			if ( $wgContLanguageCode != 'ja' || $wgDisableAnonymousEditing ) {
 				$this->checkPermissions();
 			}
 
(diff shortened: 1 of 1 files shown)


mari-mcmurtrie:
@inez will you CR? Thank you!

extensions/VisualEditor/wikia/ApiAddMediaTemporary.php
```diff
@@ -43,7 +43,7 @@ private function executeImage() {
 			// If wiki is Japanese content, then we check if anonymous edit is allowed. INT-158
 			// This condition will be changed as soon as Mercury has login for all wikis.
 			if ( $wgContLanguageCode === 'ja' ) {
-				if ( $wgDisableAnonymousEditing ) {
+				if ( $wgDisableAnonymousEditing && !$this->mUser->isLoggedIn()) {
```
inez: This will work for this particular issue - however I'm not a fond of this solution - let me explain why.
Originally this code was just using checkPermissions() method to figure out if it should proceed forward or not. Later on we decided that we want to make an exception for JA and we did - by just not calling checkPermissions in that particular case. But now it seems like you are not just making an exception (as possibly you could) but instead implementing another path that is checking permissions.
Wouldn't this just work:

```
if ( ! ( $wgContLanguageCode === 'ja' && !$wgDisableAnonymousEditing  ) ) {
    $this->checkPermissions();
}
```

jkuner:
One other thing is that we are enabling logged-in state on mercury on all wikis on Monday

extensions/VisualEditor/wikia/ApiAddMediaTemporary.php
mari-mcmurtrie: Thanks, you are right. I checked all cases and confirmed that we want to skip checkPremission() when it is JA and anonymous editing is allowed. So negating that is when we want to do checkPermission(), as you wrote above.

mari-mcmurtrie:
@Inez CR again? Thanks!

inez:
LGTM
(However identity comparisons in our code are preferred over equality comparisons.)

PR #8270, opened 2015-08-31
SOC-1203 update (set|get)GlobalAttribute calls to (set|get)Global(Flag|Preference) calls

jsutterfield:
Following the audit of which attributes we want to store in the attribute service, we came up with [a list](https://wikia-inc.atlassian.net/secure/attachment/66632/suggested_attributes.txt) for the time being. These attributes have been added to a `$wgPublicUserAttributes` array in the config which we'll use to filter which attributes we set in the service.

As part of this audit, I also went through and checked all `(set|get)Attribute` calls in MW. The changes below reflect `(set|get)Attribute` which should be actually be `(set|get)Flag` or `(set|get)Preference` calls. I also removed a couple unused methods.

Ticket: https://wikia-inc.atlassian.net/browse/SOC-1203

### DIFF
--- a/extensions/wikia/UserProfilePageV3/UserIdentityBox.class.php
+++ b/extensions/wikia/UserProfilePageV3/UserIdentityBox.class.php
@@ -210,7 +210,10 @@ class UserIdentityBox {
 
 		if (empty($memcData)) {
 			foreach (array('location', 'occupation', 'gender', 'birthday', 'website', 'twitter', 'fbPage', 'hideEditsWikis') as $key) {
-				if (!in_array($key, array('gender', 'birthday'))) {
+				if ( $key === 'hideEditsWikis' ) {
+					// hideEditsWikis is a preference, everything else is an attribute
+					$data[$key] = $this->user->getGlobalPreference($key);
+				} elseif (!in_array($key, array('gender', 'birthday'))) {
 					$data[$key] = $this->user->getGlobalAttribute($key);
 				} else {
 					$data[$key] = $this->user->getGlobalAttribute(self::USER_PROPERTIES_PREFIX . $key);
@@ -314,7 +317,10 @@ class UserIdentityBox {
 						}
 					}
 
-					if ($option === 'gender') {
+					if ($option === 'hideEditsWikis') {
+						// hideEditsWikis is a preference, everything else is an attribute
+						$this->user->setGlobalPreference($option, $data->$option);
+					} elseif ($option === 'gender') {
 						$this->user->setGlobalAttribute(self::USER_PROPERTIES_PREFIX . $option, $data->$option);
 					} else {
 						$this->user->setGlobalAttribute($option, $data->$option);
--- a/extensions/wikia/UserProfilePageV3/UserIdentityBox.class.php
+++ b/extensions/wikia/UserProfilePageV3/UserIdentityBox.class.php
@@ -213,10 +213,10 @@ class UserIdentityBox {
 				if ( $key === 'hideEditsWikis' ) {
 					// hideEditsWikis is a preference, everything else is an attribute
 					$data[$key] = $this->user->getGlobalPreference($key);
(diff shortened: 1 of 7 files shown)


includes/User.php
```diff
@@ -2664,19 +2664,6 @@ public function getDefaultGlobalPreference($preference) {
...
 	/**
-	 * Get a user attribute local to this wikia.
-	 *
-	 * @param string $attr the attribute name
-	 * @param int $cityId the city id
-	 * @param string $sep the separator between the name and the city id
-	 * @return string
-	 * @see getGlobalAttribute for more documentation about attributes
-	 */
-	public function getLocalAttribute($attr, $cityId = null, $sep = "-") {
```
jsutterfield, line 2675: This wasn't being used. All attributes we currently store are global, if we need to start supporting wiki-specific attributes down the line we can re-add this.

includes/User.php
```diff
@@ -2715,23 +2702,6 @@ private function logAttributeMismatch( $valueFromMW, $valueFromService, $attribu
...
-	 * A PLAN TO NORMALIZE IT TO "-".
-	 *
-	 * @param string $attribute
-	 * @param string $value
-	 * @param int $cityId
-	 * @param string $sep
-	 * @return bool
-	 * @see getGlobalAttribute for more documentation about attributes
-	 */
-	public function setLocalAttribute($attribute, $value, $cityId = null, $sep = '-') {
```
jsutterfield, line 2730: This wasn't being used. All attributes we currently store are global, if we need to start supporting wiki-specific attributes down the line we can re-add this.

includes/wikia/models/WikiaUserPropertiesHandlerBase.class.php
```diff
@@ -7,12 +7,12 @@ class WikiaUserPropertiesHandlerBase extends WikiaModel {
 	protected function savePropertyValue($property, $value) {
 		$this->throwExceptionForAnons();
 
-		$this->wg->User->setGlobalAttribute($property, $value);
+		$this->wg->User->setGlobalPreference($property, $value);
```
jsutterfield, line 10: `$property` here is the preference `WikiaBarDisplayState`

kvas-damian:
LGTM

extensions/wikia/UserProfilePageV3/UserIdentityBox.class.php
```diff
@@ -314,7 +317,10 @@ public function saveUserData($data) {
 						}
 					}
 
-					if ($option === 'gender') {
+					if ($option === 'hideEditsWikis') {
+						// hideEditsWikis is a preference, everything else is an attribute
+						$this->user->setGlobalPreference($option, $data->$option);
+					} elseif ($option === 'gender') {
 						$this->user->setGlobalAttribute(self::USER_PROPERTIES_PREFIX . $option, $data->$option);
```
rafalkalinski, line 324: I'm not sure how it works, but here is the line where you add USER_PROPERTIES_PREFIX to the 'gender' key, and in the line 216 it is filtered out, and the prefix is not added. Is this correct?

jsutterfield, line 324: Not quite. It's a little confusing, and initially I actually updated line 216 to clear up the logic a bit, but undid it b/c I felt like I was changing too much :)

L216 is actually checking if the option is not "gender" or "birthday" and if that's true (eg the option is say, "location" or "occupation") then it gets it without the prefix. If it is gender or birthday, then it gets it with the prefix. I actually think it would be a lot clearer written like:

``` php
if ( $key === 'gender' || $key === 'birthday' ) {
    $data[$key] = $this->user->getGlobalAttribute( self::USER_PROPERTIES_PREFIX . $key );
} else {
    $data[$key] = $this->user->getGlobalAttribute( $key );
}
```

I think I'll go ahead and update that.

PR #8274, opened 2015-08-31
LocalFile: do not call $dbw->begin when we are in the middle of transaction

macbre:
> Beginning a transaction causes any pending transaction to be committed

Inspired by https://github.com/wikimedia/mediawiki/commit/b9ac85cbf304a65d900cda00fafe53bf82d7a227

And log commits from `WikiPage::doPurge` method

@michalroszka


ghost:
:+1:

PR #8276, opened 2015-08-31
DAT-3105 Fix saving invalid titles to imagelinks table in PortableInfoboxes

sebastianmarzjan:
- Use Title::makeTitleSafe instead of Title::newFromTitle
- Add image only if title is valid


idradm:
seems obviously all right

PR #8277, opened 2015-08-31
CE-2527 Content reviewers toolbar

adamkarminski:
Content reviewers toolbar


extensions/wikia/ContentReview/ContentReviewHelper.php
```diff
@@ -112,25 +117,48 @@ public static function isStatusAwaiting( $status ) {
...
+				'toolbarTitle' => wfMessage( 'content-review-diff-toolbar-title' )->plain(),
+				'wikiId' => $wgCityId,
+				'pageId' => $wgTitle->getArticleID(),
+				'approveStatus' => ReviewModel::CONTENT_REVIEW_STATUS_APPROVED,
+				'buttonApproveText' => wfMessage( 'content-review-diff-approve' )->plain(),
+				'rejectStatus' => ReviewModel::CONTENT_REVIEW_STATUS_REJECTED,
+				'buttonRejectText' => wfMessage( 'content-review-diff-reject' )->plain(),
+				'talkpageUrl' => $this->prepareProvideFeedbackLink( $wgTitle ),
+				'talkpageLinkText' => wfMessage( 'content-review-diff-toolbar-talkpage' )->plain(),
+				'guidelinesUrl' => wfMessage( 'content-review-diff-toolbar-guidelines-url' )->plain(),
```
Grunny: In general, if it's a URL able to be set by a MW message, I think it's preferable to use `->useDatabase( false )`. While admins can only edit whitelisted MW pages right now, if they were able to edit this again, it could be set to any URL.

adamkarminski: A very good point, noted! Thanks!

extensions/wikia/ContentReview/ContentReviewHelper.php
```diff
@@ -112,25 +115,48 @@ public static function isStatusAwaiting( $status ) {
 	}
 
 	public function isDiffPageInReviewProcess( $wikiId, $pageId, $diff ) {
+		global $wgRequest;
```
Grunny, line 118: Quite a few methods in the helper now access `$wgRequest` and now the one below access `$wgTitle`. It might be worth either injecting the quest context on object creation and making use of that here (i.e. pass `$this->getContext()` from the API controllers to the constructor here, or `RequestContext::getMain()` if you're outside a WikiaController), or make it extend `ContextSource` (see [here](https://github.com/Wikia/app/blob/22bc3d9bf738/extensions/wikia/RestrictSessions/RestrictSessionsHooks.class.php#L12) and [here](https://github.com/Wikia/app/blob/22bc3d9bf738/extensions/wikia/RestrictSessions/RestrictSessionsHooks.class.php#L50) for an example). This would let you avoid the globals and use `$this->getRequest()`, `$this->getTitle()`, etc. and is the best practice where possibly in MW 1.18+.

Grunny:
Small suggestion, otherwise LGTM.

adamkarminski:
@Grunny Since this PR focuses mainly on providing UI elements, I'm merging it and continue the refactoring of the Helper class on a separate branch. Thanks!

PR #8278, opened 2015-08-31
Add an assert in WikiPage::updateRevisionOn() that page_latest is not a zero

macbre:
https://www.mediawiki.org/wiki/Manual:Page_table#page_latest says the following:

> `WikiPage::updateRevisionOn()` should set `page_latest` to a non-zero value.
> It needs to link to a revision with a valid `revision.rev_page`

@michalroszka


ghost:
LGTM

PR #8280, opened 2015-08-31
DAT-3171 Adding a maintenance script to handle invalid titles

sebastianmarzjan:
Invalid titles were added to imagelinks table from PortableInfoboxes because of too liberal check. This script removes those titles from the table when ran.


idradm:
LG

dianafa:
LGTM

PR #8282, opened 2015-08-31
CE-2598 Refactor Helper to extend ContextSource and change static met…

adamkarminski:
…hods to dynamic ones

@Grunny :) & @lukaszkonieczny


Grunny:
LGTM.

PR #8287, opened 2015-08-31
changing the auth header name to X-Wikia-AccessToken

nmonterroso:
@Wikia/services-team @ludwikkazmierczak @justnpT 

updating the auth header in MediaWiki from the oauth `AUTHORIZATION` header to `X-Wikia-AccessToken`, to be consistent with what's expected in the API gateway.

From what I understand, social and mobile do not use the `AUTHORIZATION` header, so there are no updates or backwards compatibility needed there, but the Selenium tests need to be updated with the new header.


ArturKlajnerok:
Lgtm :+1:

PR #8292, opened 2015-09-01
WikiPage: log revision and page id when page.page_latest update fails

macbre:
This will allow us to verify that the transaction is rolled back on failure.

@michalroszka


ghost:
:+1:

PR #8294, opened 2015-09-01
add invalid response from helios to log

sqreek:
Add invalid response and status code to get more data while investigating HeliosException's Tickets

ping @macbre @ArturKlajnerok

### DIFF
--- a/lib/Wikia/src/Service/Helios/HeliosClientImpl.php
+++ b/lib/Wikia/src/Service/Helios/HeliosClientImpl.php
@@ -54,10 +54,10 @@ class HeliosClientImpl implements HeliosClient
 		// Add client_id and client_secret to the GET data.
 		$getParams['client_id'] = $this->clientId;
 		$getParams['client_secret'] = $this->clientSecret;
-		
+
 		// Request URI pre-processing.
 		$uri = "{$this->baseUri}{$resourceName}?" . http_build_query($getParams);
-		
+
 		// Request options pre-processing.
 		$options = [
 			'method'          => 'GET',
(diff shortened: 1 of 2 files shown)


lib/Wikia/src/Service/Helios/HeliosClientImpl.php
```diff
@@ -91,10 +91,12 @@ public function request( $resourceName, $getParams = [], $postData = [], $extraR
 
 		$this->status = $request->status;
 
-		$output = json_decode( $request->getContent() );
+		$response = $request->getContent();
+		$output = json_decode( $response );
 
 		if ( !$output ) {
-			throw new ClientException ( 'Invalid response.' );
+			$data[ "response" ] = $response;
```
macbre, line 99: Is `$data` variable initialized as an array above?

lib/Wikia/src/Service/Helios/HeliosClientImpl.php
```diff
@@ -91,10 +91,12 @@ public function request( $resourceName, $getParams = [], $postData = [], $extraR
...
+			throw new ClientException ( 'Invalid Helios response.', 0, null, $data );
```
macbre, line 101: Let's log not only the response content, but the response code and `X-Served-By` response header (if Helios sets it) as well

lib/Wikia/src/Service/Helios/HeliosClientImpl.php
```diff
@@ -91,10 +91,14 @@ public function request( $resourceName, $getParams = [], $postData = [], $extraR
...
-		$output = json_decode( $request->getContent() );
+		$response = $request->getContent();
+		$output = json_decode( $response );
 
 		if ( !$output ) {
-			throw new ClientException ( 'Invalid response.' );
+			$data = [];
+			$data[ "response" ] = $response;
+			$data["status_code"] = $request->getStatus();
+			throw new ClientException ( 'Invalid Helios response.', 0, null, $data );
```
sqreek, line 101: Message was changed for Jira reporter filter

ArturKlajnerok:
Lgtm :+1:

ArturKlajnerok:
We don't have `X-Served-By` in helios.

PR #8295, opened 2015-09-01
CE-2603 Lazy load the right rail module to always get the freshest info

adamkarminski:
Since the lazy loaded modules are not cached - making our a lazy loaded one makes more sense. All it took was giving it an index lower than `const LAZY_LOADING_BEAKPOINT = 1440;`.


MariuszCz:
LGTM :)

PR #8296, opened 2015-09-01
ADEN-2298 New wiki vertical and wiki cats in page params

nandy-andy:
Recently we mapped all verticals to one of "three musketeers" (gaming, life and ent). But we still want to use new verticals (a wikia belongs to one from seven verticals) and possibly categories (a wikia belongs to none or many categories). In this pull request we're passing this data to page params, so they can be sent to DFP when requesting for an ad.

### DIFF
--- a/extensions/wikia/AdEngine/tests/AdEngine2ContextServiceTest.php
+++ b/extensions/wikia/AdEngine/tests/AdEngine2ContextServiceTest.php
@@ -47,7 +47,7 @@ class AdEngine2ContextServiceTest extends WikiaBaseTest {
 			'titleMockType' => 'article',
 			'flags' => [],
 			'expectedOpts' => [],
-			'expectedTargeting' => [ 'newWikiCategories' => [ 'test', 'test' ] ],
+			'expectedTargeting' => [ 'newWikiCategories' => [ 'test' ] ],
 			'expectedProviders' => [],
 			'expectedForceProviders' => null,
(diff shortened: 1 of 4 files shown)


extensions/wikia/AdEngine/tests/AdEngine2ContextServiceTest.php
```diff
@@ -130,74 +131,102 @@ public function adContextDataProvider() {
...
 				'expectedMappedVertical' => 'wikia']
 			],
+
+			array_merge($defaultParameters, [
+				'expectedTargeting' => [ 'newWikiCategories' => [ 'test', 'test1' ] ],
+				'categories' => [ 'old' => ['test'], 'new' => ['test1'] ]
+			]),
+
+			array_merge($defaultParameters, [
+				'expectedTargeting' => [ 'newWikiCategories' => [ 'test', 'test', 'test1' ] ],
```
fraszczakszymon: Do we really want here duplicated values? Why don't we use array_unique()?

nandy-andy: Good question! I don't think we've got the same categories names but it won't hurt us if we use `array_unique()` to make sure if we send only unique values to DFP :)

fraszczakszymon:
LGMT :+1:

fraszczakszymon:
LGTM :+1:

PR #8297, opened 2015-09-01
CE-2615 Do not display reviewer toolbar after completed reviews

adamkarminski:
@kamilkoterba

### DIFF
--- a/extensions/wikia/ContentReview/ContentReviewHelper.php
+++ b/extensions/wikia/ContentReview/ContentReviewHelper.php
@@ -103,6 +103,14 @@ class Helper extends \ContextSource {
 		);
 	}
 
+	public static function isStatusCompleted( $status ) {
+		return in_array( (int) $status, [
+				ReviewModel::CONTENT_REVIEW_STATUS_APPROVED,
+				ReviewModel::CONTENT_REVIEW_STATUS_REJECTED,
+			]
+		);
+	}
+
 	public function isDiffPageInReviewProcess( $wikiId, $pageId, $diff ) {
 		/**
 		 * Do not hit database if there is a URL parameter that indicates that a user
@@ -123,6 +131,29 @@ class Helper extends \ContextSource {
 		$currentData = $currentModel->getLatestReviewedRevision( $wikiId, $pageId );
 
 		return ( !empty( $currentData ) && (int)$currentData['revision_id'] === $oldid );
+	}
+
+	public function shouldDisplayReviewerToolbar() {
+		global $wgCityId;
+
+		$title = $this->getTitle();
+
+		if ( $title->inNamespace( NS_MEDIAWIKI )
+			&& $title->isJsPage()
+			&& $title->userCan( 'content-review' )
+		) {
+			$diffRevisionId = $this->getRequest()->getInt( 'diff' );
+			$diffRevisionInfo = ( new ReviewModel() )->getRevisionInfo(
+				$wgCityId,
+				$title->getArticleID(),
+				$diffRevisionId
+			);
+
+			return ( !self::isStatusCompleted( $diffRevisionInfo['status'] )
+				&& ( $this->getRequest()->getInt( self::CONTENT_REVIEW_URL_PARAM ) === 1
+					|| self::isStatusAwaiting( $diffRevisionInfo['status'] ) )
+			);
+		}
 	}
 
 	public function getToolbarTemplate() {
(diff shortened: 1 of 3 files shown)


extensions/wikia/ContentReview/ContentReviewHelper.php
```diff
@@ -125,6 +133,29 @@ public function hasPageApprovedId( $wikiId, $pageId, $oldid ) {
...
+				$wgCityId,
+				$title->getArticleID(),
+				$diffRevisionId
+			);
+
+			return ( !self::isStatusCompleted( $diffRevisionInfo['status'] )
+				&& ( $this->getRequest()->getInt( self::CONTENT_REVIEW_URL_PARAM ) === 1
+					|| self::isStatusAwaiting( $diffRevisionInfo['status'] ) )
+			);
+		}
```
Grunny, line 156: Should return false here so the method always returns a boolean.

extensions/wikia/ContentReview/models/ReviewModel.php
```diff
@@ -233,11 +233,26 @@ public function getReviewedContent( $wiki_id, $page_id, $status ) {
...
+		$db = $this->getDatabaseForRead();
+
+		$revisionInfo = ( new \WikiaSQL() )
+			->SELECT_ALL()
+			->FROM( self::CONTENT_REVIEW_STATUS_TABLE )
+			->WHERE( 'wiki_id' )->EQUAL_TO( $wikiId )
+			->AND_( 'page_id' )->EQUAL_TO( $pageId )
+			->AND_( 'revision_id' )->EQUAL_TO( $revisionId )
+			->runLoop( $db, function ( &$revisionInfo, $row ) {
+				$revisionInfo = get_object_vars( $row );
```
kamilkoterba, line 253: There's no point in using runLoop if you're returning just one record. Just run should be fine.
Perhaps query needs limit?
I would advice documenting output type once you're here as it may be overhead to check again in the end of development what it returns.

extensions/wikia/ContentReview/ContentReviewHelper.php
```diff
@@ -103,6 +103,14 @@ public static function isStatusAwaiting( $status ) {
 		);
 	}
 
+	public static function isStatusCompleted( $status ) {
+		return in_array( (int) $status, [
```
kamilkoterba: no space: `(int)$status` per [MW coding conventions](https://www.mediawiki.org/wiki/Manual:Coding_conventions/PHP#Spaces)

extensions/wikia/ContentReview/models/ReviewModel.php
```diff
@@ -233,11 +233,26 @@ public function getReviewedContent( $wiki_id, $page_id, $status ) {
 			->WHERE( 'wiki_id' )->EQUAL_TO( $wiki_id )
 			->AND_( 'page_id' )->EQUAL_TO( $page_id )
 			->AND_( 'status' )->EQUAL_TO( $status )
-			->LIMIT(1)
```
kamilkoterba, line 236: Why limit is removed?

extensions/wikia/ContentReview/models/ReviewModel.php
```diff
@@ -233,11 +233,26 @@ public function getReviewedContent( $wiki_id, $page_id, $status ) {
...
 			->runLoop( $db, function ( &$content, $row ) {
```
kamilkoterba, line 236: as below. loop not needed?

extensions/wikia/ContentReview/ContentReviewHelper.php
adamkarminski, line 156: Right!

extensions/wikia/ContentReview/models/ReviewModel.php
adamkarminski, line 236: It's not necessary since `wiki_id`, `page_id` and `status` form a unique key in our DB.

extensions/wikia/ContentReview/models/ReviewModel.php
adamkarminski, line 236: `run()` is a wrapper for `runLoop()` anyway

extensions/wikia/ContentReview/models/ReviewModel.php
adamkarminski, line 253: `run()` is a wrapper for `runLoop()` anyway. Query does not need a limit since `wiki_id`, `page_id` and `revision_id` form another unique key in our DB.

Would you mind if the documentation was done with the overall documentation task?

adamkarminski, line 253: And `run()` still returns a `ResultWrapper` object so you still need a callback to handle it.

extensions/wikia/ContentReview/ContentReviewHelper.php
adamkarminski, line 156: Done :+1:

kamilkoterba:
LGTM!

PR #8300, opened 2015-09-01
Lyr 239

SeanColombo:
Added tracking immediately before the call to our pixel (which then hits the LyricFind API).

Set the category & action as recommended in AN-861

### DIFF
--- a/extensions/3rdparty/LyricWiki/LyricFind/js/tracking.js
+++ b/extensions/3rdparty/LyricWiki/LyricFind/js/tracking.js
@@ -8,6 +8,16 @@ require(['jquery','LyricFindTracker', 'wikia.log'], function($, tracker, log) {
         return;
     }
 
+	// Track the view in GA/internal data warehouse.
+	Wikia.Tracker.track({
+		action: Wikia.Tracker.ACTIONS.VIEW,
+		category: 'lyricView',
+		//label: 'lyricFindView',
+		trackingMethod: 'analytics'
+	});
+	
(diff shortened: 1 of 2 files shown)


extensions/3rdparty/LyricWiki/LyricFind/js/tracking.js
```diff
@@ -8,6 +8,16 @@ require(['jquery','LyricFindTracker', 'wikia.log'], function($, tracker, log) {
         return;
     }
 
+	// Track the view in GA/internal data warehouse.
+	Wikia.Tracker.track({
+		action: Wikia.Tracker.ACTIONS.VIEW_LYRIC,
+		category: 'lyricView',
+		//label: 'lyricFindView',
```
bkoval: Commented code, I'm guessing it can be removed

resources/wikia/modules/tracker.stub.js
```diff
@@ -88,7 +88,10 @@
 				TAKE_SURVEY: 'take-survey',
 
 				// View
-				VIEW: 'view'
+				VIEW: 'view',
+				
+				// View song lyrics (for cross-referencing against LyricFind reports) AN-861
+				VIEW_LYRIC: 'view-lyric'
```
bkoval: Maybe you could move this declaration to LyricWIki's tracking.js to keep the stub as clean as possible, it's used only in the tracking.js file anyway :)

SeanColombo:
@bkoval thanks for the suggestions!  I agree w/both and committed those changes.

bkoval:
LGTM :+1:

PR #8302, opened 2015-09-02
CE-2615 Compose the valid check for toolbar appearance

adamkarminski:
Now, the toolbar will only appear for diffs of revisions "In review" or for the ones with a status "Unreviewed" and a `contentreview=1` URL param in case of slow master-slave replication.

@lukaszkonieczny


lukaszkonieczny:
LGTM

PR #8311, opened 2015-09-02
DAT-3184 escape wikiamobile menu i18n

idradm:


sebastianmarzjan:
After looking at the wikiamobile-menu defs - LGTM
:ship:

PR #8312, opened 2015-09-02
DAT-3183 escape messages from user identity box group

idradm:


sebastianmarzjan:
The change itself looks ok. Please take a look at UserProfilePage_renderUserIdentityBox.php though, especially the hgroup in div[class="masthead-info"]

idradm:
I've check it at it should be escaped when this data is saved.

PR #8313, opened 2015-09-02
Updated HeliosClientImpl::invalidateToken() to match Helios endpoint.

harnash:
After changes in Helios, initial endpoint for invalidating user access token changed so this change updates HeliosClient to use the current one.

@Wikia/services-team


ArturKlajnerok:
Lgtm :+1:

PR #8321, opened 2015-09-03
Animate the latest revision status if it needs to be submitted

adamkarminski:
Per a suggestion from @PiotrGackowski :)

### DIFF
--- a/skins/oasis/css/modules/ContentReview.scss
+++ b/skins/oasis/css/modules/ContentReview.scss
@@ -48,8 +48,28 @@ $content-review-color-unsubmitted: #ffad00;
 
 .content-review-status-unsubmitted {
 	border-color: $content-review-color-unsubmitted;
+	animation: submitReminder 3s infinite;
 }
 
 .content-review-status-awaiting {
 	border-color: $content-review-color-awaiting;
+}
+
+/* Animation */
+@keyframes submitReminder {
+	0% {
+		margin-left: 0;
+	}
+
+	25% {
+		margin-left: 10px;
+	}
+
+	50% {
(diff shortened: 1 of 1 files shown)


skins/oasis/css/modules/ContentReview.scss
```diff
@@ -48,8 +48,28 @@ $content-review-color-unsubmitted: #ffad00;
...
 	border-color: $content-review-color-unsubmitted;
+	animation: submitReminder 3s infinite;
 }
 
 .content-review-status-awaiting {
 	border-color: $content-review-color-awaiting;
 }
+
+/* Animation */
+@keyframes submitReminder {
```
hakubo, line 59: 0% margin-left: 0;
100% margin-left: 10px;

and when invoking
animation: submitReminder .75s ease 1.5s infinite;

that should be the same right?

adamkarminski: Yup, checked it, thanks! :)

adamkarminski, line 59: I want it to both start and end at 0 and wait for a while before the next iteration. Your solution makes it start from 0 and end at 10px everytime which looks like this: http://jsfiddle.net/kge01467/

P.S. But thanks to you I've found out, that I can skip the 100% declaration :)

kamilkoterba, line 59: @adamkarminski @hakubo per js fiddle visualization, I think it could sound well if you start a band together ;) very rhythmic visualization :)

kamilkoterba:
LGTM

PR #8327, opened 2015-09-03
(CE-2509,CE-2510) Tweaks to personal JS permissions

Grunny:
Add a preference requiring users to explicitly opt-in to personal JS. Allow
users to edit their personal JS, but only the main skin JS files.

Cherry picked from https://github.com/Wikia/app/pull/8263


lgarczewski:
LGTM.

PR #8329, opened 2015-09-04
SOC-1216 Purge user cache in MW after an update to the attribute service

jsutterfield:
This PR creates a MW API for the attribute service to clear a user's cache after an update to one of their attributes. See https://github.com/Wikia/pandora/pull/440 for more information.

### DIFF
--- a/extensions/wikia/AttributeServiceMWApi/AttributeServiceMWApiController.class.php
+++ b/extensions/wikia/AttributeServiceMWApi/AttributeServiceMWApiController.class.php
@@ -1,6 +1,6 @@
 <?php
 
-class AttributeServiceHelperController extends WikiaController {
+class AttributeServiceMWApiController extends WikiaController {
 
     /**
      * Helper function for the attribute service to clear user MediaWiki.
@@ -9,17 +9,21 @@ class AttributeServiceHelperController extends WikiaController {
      */
     public function purgeUserCache() {
 
+        jmark();
         $userId = $this->request->getVal( 'userId' );
         $signature = $this->request->getVal( 'signature' );
 
        if ( !$this->assertValidParameters( $userId, $signature ) ) {
+           jmark();
            return;
        }
 
         if ( !$this->assertValidSignature( $userId, $signature ) ) {
+            jmark();
             return;
         }
 
+        jmark();
         User::clearUserCache( $userId );
         $userIdentityBox = new UserIdentityBox( User::newFromId( $userId ) );
         $userIdentityBox->clearCache();
--- a/extensions/wikia/AttributeServiceMWApi/AttributeServiceMWApiController.class.php
+++ b/extensions/wikia/AttributeServiceMWApi/AttributeServiceMWApiController.class.php
@@ -2,80 +2,80 @@
 
 class AttributeServiceMWApiController extends WikiaController {
 
-    /**
-     * Helper function for the attribute service to clear user MediaWiki.
-     * This is called when a client other than MW updates an attribute
-     * using the attribute service to clear the User cache.
-     */
-    public function purgeUserCache() {
+	/**
+	 * Helper function for the attribute service to clear user MediaWiki.
+	 * This is called when a client other than MW updates an attribute
+	 * using the attribute service to clear the User cache.
+	 */
+	public function purgeUserCache() {
 
-        $userId = $this->request->getVal( 'userId' );
-        $signature = $this->request->getVal( 'signature' );
+		$userId = $this->request->getVal( 'userId' );
+		$signature = $this->request->getVal( 'signature' );
 
-       if ( !$this->assertValidParameters( $userId, $signature ) ) {
-           $this->setErrorResponse( 'invalid parameters: (userId) and (signature) must not be empty',
-               \WikiaResponse::RESPONSE_CODE_BAD_REQUEST );
-           return;
-       }
+	   if ( !$this->assertValidParameters( $userId, $signature ) ) {
+		   $this->setErrorResponse( 'invalid parameters: (userId) and (signature) must not be empty',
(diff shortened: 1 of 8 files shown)


jsutterfield:
I know, failed unit tests again....I feel like this is becoming a reputation of mine :/ I'll be better about fixing these going forward, but I think the rest of the ticket is ready to be CRed and we need to move this along if we can. I'll have a look at the broken tests first thing tomorrow.

lib/Wikia/src/Persistence/User/Attributes/AttributePersistenceSwagger.php
```diff
@@ -107,7 +107,11 @@ public function deleteAttribute( $userId, Attribute $attribute ) {
 	 * @return UsersAttributesApi
 	 */
 	private function getApi( $userId ) {
-		return $this->apiProvider->getAuthenticatedApi( self::SERVICE_NAME, $userId, UsersAttributesApi::class );
+		/** @var UsersAttributesApi $apiClient */
+		$apiClient = $this->apiProvider->getAuthenticatedApi( self::SERVICE_NAME, $userId, UsersAttributesApi::class );
+		$apiClient->getApiClient()->getConfig()->addDefaultHeader("X-From-MW", "true");
```
jsutterfield: Adding a header so the User Attribute service can identify if a request came from MW or not.

extensions/wikia/AttributeServiceMWApi/AttributeServiceMWApiController.class.php
```diff
@@ -0,0 +1,81 @@
+<?php
+
+class AttributeServiceMWApiController extends WikiaController {
+
+    /**
```
rafalkalinski: tabs instead of spaces :)

extensions/wikia/AttributeServiceMWApi/AttributeServiceMWApiController.class.php
```diff
@@ -0,0 +1,81 @@
...
+    public function purgeUserCache() {
+
+        $userId = $this->request->getVal( 'userId' );
+        $signature = $this->request->getVal( 'signature' );
+
+       if ( !$this->assertValidParameters( $userId, $signature ) ) {
+           return;
+       }
+
+        if ( !$this->assertValidSignature( $userId, $signature ) ) {
```
rafalkalinski: I would say it is to much to put this functionality into two separate functions (assertValidSignature and assertValidParameters).
But that's just me :)

extensions/wikia/AttributeServiceMWApi/AttributeServiceMWApiController.class.php
```diff
@@ -0,0 +1,81 @@
...
+            return;
+        }
+
+        User::clearUserCache( $userId );
```
macbre: This method will only clear the "local" user cache (i.e. per-wiki one).

Call `User::invalidateCache` instead as it clears shared cache from external user as well (which invalidates all local user caches) and updates `user_touched` in the database.

extensions/wikia/AttributeServiceMWApi/AttributeServiceMWApiController.class.php
```diff
@@ -0,0 +1,81 @@
...
+     * This is called when a client other than MW updates an attribute
+     * using the attribute service to clear the User cache.
+     */
+    public function purgeUserCache() {
+
+        $userId = $this->request->getVal( 'userId' );
+        $signature = $this->request->getVal( 'signature' );
+
+       if ( !$this->assertValidParameters( $userId, $signature ) ) {
+           return;
```
macbre: This will make the request return HTTP 200 response code. Let's throw an exception here to clearly indicate that something went wrong.

rafalkalinski:
Two minor comments, some failing test, but beside that LGTM ;)

extensions/wikia/AttributeServiceMWApi/AttributeServiceMWApiController.class.php
jsutterfield: It actually sets the error response code [here](https://github.com/Wikia/app/pull/8329/files#diff-341ac26a93d9bf763b1539072665fa4cR39). But, it's not super obviously that asserting here has the side effect of altering the response. I'll move the `setErrorResponse` call higher up to make that more clear.

extensions/wikia/AttributeServiceMWApi/AttributeServiceMWApiController.class.php
jsutterfield: Bah! After spilling water on my old laptop I'm still setting up this new one...guess I missed that part. I'll update.

extensions/wikia/AttributeServiceMWApi/AttributeServiceMWApiController.class.php
jsutterfield: Hmm, I was trying `User::invalidateCache` at first but it wasn't working properly. It looks like it loads the previous values form cache, deletes the cache, and then sets that old cache again...I do need to cleared the shared cache though, and update, the `user_touched`. Calling `User::clearUserCache( $userId );` first seems to fix it.

extensions/wikia/AttributeServiceMWApi/AttributeServiceMWApiController.class.php
jsutterfield: Can we call this one a stylistic difference? I like having it split in 2 methods, I think they're distinct enough, and makes it clear what's going on :)

armonr:
LGTM

extensions/wikia/AttributeServiceMWApi/AttributeServiceMWApiController.class.php
rafalkalinski: sure :)

PR #8335, opened 2015-09-04
ADEN-2407 Added SP status to all ads GA events

fraszczakszymon:


extensions/wikia/AdEngine/js/AdTracker.js
```diff
@@ -7,6 +7,11 @@ define('ext.wikia.adEngine.adTracker', ['wikia.tracker', 'wikia.window'], functi
 
 	function encodeAsQueryString(extraParams) {
 		var out = [], key, keys = [], i, len;
+
+		if (window.ads.context.opts.sp) {
```
nandy-andy: How different would it be to use `ext.wikia.adEngine.adContext` dependency instead of reaching for global `ads.context.opts.sp`?

nandy-andy:
I know I gave the idea of using our own namespace and setting global variable but given the 2nd thought on that after reviewing the code I wouldn't use `window.ads.context` for this. Since, we want to make it representation of what we create in PHP and then modify only with our configs.

If we can't make it cleaner let's go with `ads.sourcepoint` object for example? `window.ads.sourcepoint.adBlockDetected = 'yes';` and `window.ads.sourcepoint.adBlockDetected = 'no';`. What do you think?

fraszczakszymon:
:+1: 
We have `window.ads.runtime` object. What do you think about `window.ads.runtime.sp.blocking`?

nandy-andy:
I didn't know about `window.ads.runtime` :) Sounds and looks better, I guess :smiley_cat: 

LGTM :+1:

PR #8340, opened 2015-09-04
SOC-1231 Clicking on an avatar in global nav shows a page you are viewing instead of a login form in the iframe

rafalkalinski:
https://wikia-inc.atlassian.net/browse/SOC-1231


bkoval:
LGTM

PR #8342, opened 2015-09-04
SOC-1231 | proper event target handler used while getting href

rafalkalinski:
https://wikia-inc.atlassian.net/browse/SOC-1231


bkoval:
LGTM :hammer:

PR #8344, opened 2015-09-04
Send the auth header when deleting the users access token in helios

drsnyder:
Send the auth header when deleting the users access token in helios.

https://wikia-inc.atlassian.net/browse/SERVICES-707

/cc @Wikia/services-team


lib/Wikia/src/Service/Helios/HeliosClientImpl.php
```diff
@@ -13,6 +13,8 @@ class HeliosClientImpl implements HeliosClient
 	const CLIENT_ID = "client_id";
 	const CLIENT_SECRET = "client_secret";
 
+	const HELIOS_AUTH_HEADER = "X-Wikia-UserId";
```
nmonterroso: Can we change usage of https://github.com/Wikia/app/blob/dev/lib/Wikia/src/Service/Swagger/ApiProvider.php#L10 to use this instead?

nmonterroso:
minor comment otw lgtm

ArturKlajnerok:
Lgtm :+1:

lib/Wikia/src/Service/Helios/HeliosClientImpl.php
drsnyder: I saw that reference... Just so I'm clear, that's not auto generated?

drsnyder: Doesn't look like it... I think I'm going to move this into `Wikia\Service\Constants`.

harnash: @drsnyder, @nmonterroso: Added `Wikia\Service\Constants` and moved the `X-Wkia-UserId` there (a8cd822).

harnash:
Implemented all the changes requested in comments.

@Wikia/services-team

sqreek:
Lgtm

nmonterroso:
@harnash the constant was moved out of `ApiProvider`, but looks like there are references to the constant in the `UserAvatarsApi`

lib/Wikia/src/Service/Helios/HeliosClientImpl.php
```diff
@@ -155,16 +157,18 @@ public function refreshToken( $token )
 	 * A shortcut method for token invalidation requests.
 	 *
 	 * @param $token string - a token to be invalidated
+	 * @param $userId integer - the current user id
 	 *
 	 * @return string - json encoded response
 	 */
-	public function invalidateToken( $token )
+	public function invalidateToken( $token, $userId )
```
nmonterroso: This is incompatible with the `HeliosClient` interface

PR #8346, opened 2015-09-07
(CE-2623) Allow editing of MW JS pages if ContentReview is enabled

Grunny:
Allow editing of MediaWiki JS pages if ContentReview and site JS are
enabled. Checking site JS as well allows us to enable ContentReview
everywhere in the future without opening up editing of site JS where
it's not in use.

Also blacklist *.js pages from Verbatim now that they will be editable.


lukaszkonieczny:
LGTM

PR #8347, opened 2015-09-07
SOC-1247 take attribute service out of shadow mode

jsutterfield:
Ticket: https://wikia-inc.atlassian.net/browse/SOC-1247

### DIFF
--- a/includes/User.php
+++ b/includes/User.php
@@ -4894,9 +4894,8 @@ class User {
 			foreach ( $attributes as $attributeName => $attributeValue ) {
 				$this->compareAttributeValueFromService( $attributeName, $attributeValue );
 
-				// Once shadow mode verifies these are correct, we can uncomment this
-				// $this->mOptionOverrides[$attributeName] = $attributeValue;
-				// $this->mOptions[$attributeName] = $attributeValue;
+				 $this->mOptionOverrides[$attributeName] = $attributeValue;
+				 $this->mOptions[$attributeName] = $attributeValue;
 			}
 		}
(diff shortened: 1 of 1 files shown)


macbre:
LGTM (as far as shadow mode proved that the attributes service works correctly)

includes/User.php
```diff
@@ -4894,9 +4894,8 @@ private function loadAttributes() {
 			foreach ( $attributes as $attributeName => $attributeValue ) {
 				$this->compareAttributeValueFromService( $attributeName, $attributeValue );
```
macbre, line 4895: Shouldn't we raise an exception here (when there's a mismatch) instead of writing wrong data?

includes/User.php
```diff
@@ -4894,9 +4894,8 @@ private function loadAttributes() {
...
 
-				// Once shadow mode verifies these are correct, we can uncomment this
-				// $this->mOptionOverrides[$attributeName] = $attributeValue;
-				// $this->mOptions[$attributeName] = $attributeValue;
+				 $this->mOptionOverrides[$attributeName] = $attributeValue;
+				 $this->mOptions[$attributeName] = $attributeValue;
```
armonr, line 4898: @jsutterfield: Should this be controlled via a wg variable as discussed in the previous ticket [SOC-1216](https://wikia-inc.atlassian.net/browse/SOC-1216?focusedCommentId=193377&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-193377)?

PR #8350, opened 2015-09-07
Dat 3150 skip rendering labels row in infobox horizontal group if all labels are empty

RafLeszczynski:


dianafa:
LGTM

sebastianmarzjan:
What about the coverage through api tests, do they pass too?

RafLeszczynski:
@SebastianMarzjan good catch - will check it

PR #8353, opened 2015-09-07
CE-2610 Adding a method for importing local wikia scripts ...

kamilkoterba:
in user custom JS scripts

https://wikia-inc.atlassian.net/browse/CE-2610

@Wikia/community-engineering


extensions/wikia/ContentReview/scripts/contentReviewAddImportWikiaScriptPagesScript.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * Adds a method to window object for importing wikia local scripts
+ */
+require(['wikia.window'], function (window) {
+	'use strict';
+	/**
+	 * Imports script from provided JS page in MediaWiki namespace
+	 * @param {string} scriptName Name of page without namespace prefix
```
lukaszkonieczny: I think should be string|array

extensions/wikia/ContentReview/scripts/contentReviewAddImportWikiaScriptPagesScript.js
```diff
@@ -0,0 +1,30 @@
...
+	 */
+	window.importWikiaScriptPages = function (articles, server) {
+
+		for (var i = 0; i < articles.length; i++) {
```
lukaszkonieczny: This loop will return unexpected result if articles will be a string

extensions/wikia/ContentReview/scripts/contentReviewAddImportWikiaScriptPagesScript.js
```diff
@@ -0,0 +1,30 @@
...
+
+		window.importArticles({
+			type: 'script',
+			articles: articles,
+			server: server
+		});
+	};
+
+	function isJsPage (scriptName) {
+		return scriptName.substr(scriptName.length - 3) === '.js';
```
lukaszkonieczny: Please check is there possibility to add page with `.JS` or `.Js`

adamkarminski: Maybe the next step after checking that a name does not end with .js should be adding this part? It could be convenient to just give a name of an article (e.g. just MyScript instead of MyScript.js).

resources/wikia/wikia.wikibits.js
```diff
@@ -155,8 +161,40 @@ var importArticleMissing = (function() {
...
+
+		window.importArticles({
+			type: 'script',
+			articles: articlesToImport,
+			server: server
+		});
+	}
+
+	function isJsPage (scriptName) {
+		return scriptName.substr(scriptName.length - 3).toLowerCase() === '.js';
```
Grunny: We shouldn't be converting to lowercase, we should only allow lowercase .js pages, i.e. only pages that match `Title::isJsPage`. i.e. http://starwars.wikia.com/wiki/MediaWiki:Common.js is a JS page according to MediaWiki, but http://starwars.wikia.com/wiki/MediaWiki:Common.Js?action=edit is not and shouldn't be allowed as an import.

extensions/wikia/ContentReview/scripts/contentReviewAddImportWikiaScriptPagesScript.js
kamilkoterba: MediaWiki: prefix is namespace when .js extension is part of the title. I prefer to tell a user to explicitly provide extension as it's a part of title. Added banner notification for user when detected missing .js

resources/wikia/wikia.wikibits.js
```diff
@@ -109,54 +110,145 @@ var importArticle = (function() {
...
+var importWikiaScriptPages = (function () {
+	var namespacePrefix = 'MediaWiki:',
+		externalPrefix = 'external:',
+		externalPrefixLength = externalPrefix.length;
+
+	function importWikiaScriptPages(articles) {
+		var articlesToImport = [],
+			articlesFailed = [];
+
+		for (var i = 0; i < articles.length; i++) {
```
lukaszkonieczny, line 200: I think we should check if `articles` is an array. What do you think about allowing importing one script with parameter as a string?

resources/wikia/wikia.wikibits.js
```diff
@@ -109,54 +110,145 @@ var importArticle = (function() {
...
+			articles: articlesToImport
+		});
+	}
+
+	function isJsPage(scriptName) {
+		return scriptName.substr(scriptName.length - 3) === '.js';
+	}
+
+	function isExternal(scriptName) {
+		return scriptName.substr(0, externalPrefixLength) === externalPrefix;
```
lukaszkonieczny, line 230: I'm not sure about it. In previous version we had `server` variable which was much simpler for me to use

resources/wikia/wikia.wikibits.js
```diff
@@ -109,54 +110,145 @@ var importArticle = (function() {
...
+			if (isExternal(articles[i])) {
+				var articlePreparedName = prepareExternalArticleName(articles[i]);
+				if (articlePreparedName.length === 0) {
+					articlesFailed.push(articles[i]);
+					continue;
+				}
+				articlesToImport.push(articlePreparedName);
+				continue;
+			}
+			articlesToImport.push(namespacePrefix + articles[i]);
```
lukaszkonieczny, line 214: Maybe we can use one use `push` method once?

resources/wikia/wikia.wikibits.js
lukaszkonieczny, line 230: Ahhh but it doesnt allow to add multiple scripts :/

resources/wikia/wikia.wikibits.js
```diff
@@ -109,54 +110,145 @@ var importArticle = (function() {
...
+
+	function isExternal(scriptName) {
+		return scriptName.substr(0, externalPrefixLength) === externalPrefix;
+	}
+
+	function prepareExternalArticleName(articleName) {
+		var indexOfSubdomain = articleName.indexOf(':', externalPrefixLength) + 1,
+			articlePreparedName = '';
+
+		if (indexOfSubdomain === 0) {
```
lukaszkonieczny, line 237: You should also check `-1` value

kamilkoterba, line 237: I'm adding +1 line above, hence 0 not -1.
any number lower than externalPrefixLength and different than 0 is not possible here.

resources/wikia/wikia.wikibits.js
kamilkoterba, line 230: It's similar to importArticles format and I think we can add help page link to notification once one is ready.

resources/wikia/wikia.wikibits.js
lukaszkonieczny, line 237: :+1:

lukaszkonieczny:
LGTM

PR #8358, opened 2015-09-07
CE-2614: Ignore expanding script templates via templates=expand

lukaszkonieczny:


Grunny:
This looks OK, but do we only want this when ContentReview is enabled, or just always? If so, it might be better to make use of the `RawPageViewBeforeOutput` hook, to make sure all ContentReview-specific stuff stays in the ContentReview extension, otherwise, just remove the `!empty( $wgEnableContentReviewExt )` part of the check.

PR #8360, opened 2015-09-07
(CE-2628) Allow editing of *.js talk pages

Grunny:
These are normal talk pages, so were always editable by all users.
This allows editing of MediaWiki talk:*.js pages in order to raise
issues with the scripts.


lukaszkonieczny:
LGTM

PR #8363, opened 2015-09-07
ADEN-2392 Run recovery only on blocking

fraszczakszymon:


extensions/wikia/AdEngine/js/provider/gpt/googleTag.js
```diff
@@ -96,7 +99,14 @@ define('ext.wikia.adEngine.provider.gpt.googleTag', [
 	};
 
 	GoogleTag.prototype.push = function (callback) {
-		googleTag.cmd.push(callback);
+		if (this.storeCommands) {
+			cmd.push(callback);
+		}
+		window.googletag.cmd.push(callback);
```
nandy-andy, line 109: Do we want to add callback to `window.googletag.cmd` even if `this.storeCommands` is set and we add the same callback to `cmd` array?

extensions/wikia/AnalyticsEngine/js/universal_analytics.js
```diff
@@ -244,6 +244,10 @@
     }
 
     function trackBlocking(value) {
+        if (blockingTracked) {
```
nandy-andy, line 247: Can it be removed if SourcePoint fixes issue with sending too many events?

extensions/wikia/AdEngine/js/provider/gpt/sourcePointTag.js
```diff
@@ -38,44 +38,27 @@ define('ext.wikia.adEngine.provider.gpt.sourcePointTag', [
...
-			log(['sp.blocking'], 'debug', logGroup);
-			doc.body.classList.add('source-point');
-			blocking = true;
-		}).bind(this));
+		doc.body.classList.add('source-point');
 
-		doc.addEventListener('sp.not_blocking', (function () {
-			log(['sp.not_blocking'], 'debug', logGroup);
-		}).bind(this));
+		delete window.googletag;
```
nandy-andy: Add a comment here explaining why we have to do that. I believe it's connected to the note in the ticket about AdBlock Plus.

extensions/wikia/AdEngine/js/provider/gpt/googleTag.js
fraszczakszymon, line 109: Yup, callback has to go to googletag because we don't want to affect non blocking users. On the other hand we need those callbacks when we want to run recovery (googletag.cmd is cleared for AdBlock Plus users). Now I see it's better to rename this temporary cmd array - maybe `recoveryCmd`?

extensions/wikia/AnalyticsEngine/js/universal_analytics.js
fraszczakszymon, line 247: Not really, first `sp.blocking` is dispatched by detection script and when user is blocking then the second call is from recovery. Actually SP already fixed duplicated events in recovery :)

extensions/wikia/AdEngine/js/provider/gpt/sourcePointTag.js
fraszczakszymon: Nice catch! :gem: Since `window.googletag` is overridden this line is not needed any more, but I added a comment here :)

extensions/wikia/AdEngine/js/provider/gpt/googleTag.js
nandy-andy, line 109: Makes sense :)

extensions/wikia/AdEngine/js/provider/gpt/googleTag.js
```diff
@@ -8,15 +8,18 @@ define('ext.wikia.adEngine.provider.gpt.googleTag', [
...
 		pageLevelParams,
-		googleTag,
 		pubAds;
 
-	function GoogleTag() {
+	function GoogleTag(storeCommands) {
 		this.initialized = false;
+		this.storeCommands = storeCommands;
+		slots = {};
+		slotQueue = [];
```
nandy-andy: Do we have to reset `slots` and `slotQueue` in the constructor? They aren't "properties" of an instance and we define them with assignment above. I wonder if we can get rid of one or another assignment :japanese_ogre:

nandy-andy:
LG :) OK, let's test it.

PR #8366, opened 2015-09-07
Dat 3194 - for videos in infobox use link to file page instead of link to original image

RafLeszczynski:
https://wikia-inc.atlassian.net/browse/DAT-3194

### DIFF
--- a/extensions/wikia/PortableInfobox/services/Parser/Nodes/NodeImage.php
+++ b/extensions/wikia/PortableInfobox/services/Parser/Nodes/NodeImage.php
@@ -39,7 +39,7 @@ class NodeImage extends Node {
 			];
 
 			if ( $this->isVideo( $file ) ) {
-				$this->data = $this->videoDataDecorator( $this->data, $file );
+				$this->data = $this->videoDataDecorator( $this->data, $file, $title );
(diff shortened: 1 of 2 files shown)


extensions/wikia/PortableInfobox/services/Parser/Nodes/NodeImage.php
```diff
@@ -107,11 +107,13 @@ private function isVideo( $file ) {
...
+	 * @param Title $title
 	 * @return array
 	 */
-	private function videoDataDecorator( $data, $file ) {
-		$data['isVideo'] = true;
-		$data['duration'] = WikiaFileHelper::formatDuration( $file->getMetadataDuration());
+	private function videoDataDecorator( $data, $file, $title ) {
+		$data[ 'isVideo' ] = true;
+		$data[ 'duration' ] = WikiaFileHelper::formatDuration( $file->getMetadataDuration());
+		$data[ 'url' ] = $title->getFullURL();
```
idradm: You can get title object from $file->getTitle(), however you should check if title is not empty. You can get rid of title param that way.

idradm:
LG

PR #8368, opened 2015-09-08
DAT-3102 avoid double HTML escaping and do it only in mustache template

RafLeszczynski:
https://wikia-inc.atlassian.net/browse/DAT-3102


dianafa:
LGTM

PR #8369, opened 2015-09-08
MON-416: disable SEOTweaks heavy database query

macbre:
https://wikia-inc.atlassian.net/browse/MON-416

Disable `SEOTweaksHooksHelper::onArticleViewHeader` hook.

``` sql
SELECT page_title FROM page WHERE page_title LIKE '255961%' LIMIT 1
```

These queries take a really long time to complete (median value is around a minute) generating a long tail of MediaWiki response times. Additionally, [in only 5% of cases it actually returns matching article](https://kibana.wikia-inc.com/index.html#/dashboard/elasticsearch/MON-416).

@wladekb / @michalroszka / @jacekjursza

### DIFF
--- a/extensions/wikia/SEOTweaks/SEOTweaks.setup.php
+++ b/extensions/wikia/SEOTweaks/SEOTweaks.setup.php
@@ -17,7 +17,7 @@ $wgHooks['BeforePageDisplay'][] = 'SEOTweaksHooksHelper::onBeforePageDisplay';
 $wgHooks['AfterInitialize'][] = 'SEOTweaksHooksHelper::onAfterInitialize';
 $wgHooks['ImagePageAfterImageLinks'][] = 'SEOTweaksHooksHelper::onImagePageAfterImageLinks';
 $wgHooks['BeforeParserMakeImageLinkObjOptions'][] = 'SEOTweaksHooksHelper::onBeforeParserMakeImageLinkObjOptions';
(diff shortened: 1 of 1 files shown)


extensions/wikia/SEOTweaks/SEOTweaks.setup.php
```diff
@@ -17,7 +17,7 @@
 $wgHooks['AfterInitialize'][] = 'SEOTweaksHooksHelper::onAfterInitialize';
 $wgHooks['ImagePageAfterImageLinks'][] = 'SEOTweaksHooksHelper::onImagePageAfterImageLinks';
 $wgHooks['BeforeParserMakeImageLinkObjOptions'][] = 'SEOTweaksHooksHelper::onBeforeParserMakeImageLinkObjOptions';
-$wgHooks['ArticleViewHeader'][] = 'SEOTweaksHooksHelper::onArticleViewHeader';
+#$wgHooks['ArticleViewHeader'][] = 'SEOTweaksHooksHelper::onArticleViewHeader';
```
pawelchcki: Why its commented, wouldn't it be better to remove it?

macbre: Let's leave the trace that it was commented out on purpose (with a link to the ticket - added in b17e7f1).

MON ticket will not be closed when this one is merged. It's just a temporary fix to remove the heavy DB query. This still needs to be refactored.

wladekb:
lgtm

PR #8370, opened 2015-09-08
DAT-3127 treat 0 as valid value inside <format> tag

RafLeszczynski:
https://wikia-inc.atlassian.net/browse/DAT-3127


idradm:
LG

extensions/wikia/PortableInfobox/tests/nodes/NodeDataTest.php
```diff
@@ -92,6 +92,8 @@ public function dataProvider() {
 				[ 'value' => '{{{not_defined_var}}}%', 'label' => 'l' ] ],
 			[ '<data source="test"><label>l</label><format>{{{test}}}%</format><default>def</default></data>', [ ],
 				[ 'value' => 'def', 'label' => 'l' ] ],
+			[ '<data source="test"><format>{{{test}}}%</format></data>', [ 'test' => 0 ],
```
sebastianmarzjan, line 95: It looks like we're missing a test for an analogous structure with a default value. As far as I remember, '0' in param value should be rendered as-is when there is only default tag inside of data tag and looking at condition in https://github.com/Wikia/app/pull/8370/files#diff-b09ea7c9a303b75f3ed56e8a326e0e0cL148 I think it does not work like that now and processes the default.

PR #8373, opened 2015-09-08
Clean up wfSuppressWarnings()

wladekb:
wfSuppressWarnings() is called so many times and these checks are completely unnecessary. Let's remove them!

/cc @macbre


macbre:
Mohr red! :devil:

LGTM

PR #8377, opened 2015-09-08
Experimental unrolling of selected functions

wladekb:
There are some hotspots that are executed thousands or millions times per minute. Some of them can be easily optimized for performance without a big cost for code readability or maintainability.

Let's try a few of them to collect more data how much unrolling of those functions helps us achieve better backend performance.

List of functions that are unrolled:
- WikiaGlobalRegistry::__get()
- WikiaView::normalizeControllerClass()

/cc @macbre


ghost:
:+1:

PR #8379, opened 2015-09-08
VE-1914: Implement VirtualRESTService for VisualEditor

pauloslund:
The main idea behind VRS is that it provides an abstraction layer over protocols and backends; which hopefully makes dealing with REST based interfaces easier and more convenient. More reading around the ideas behind VRS can be found here: https://www.mediawiki.org/wiki/Requests_for_comment/PHP_Virtual_REST_Service

As a note, some configuration options have been removed. This is because they were written in a way specific to a newer version of MediaWiki.

This update brings in two libraries from MediaWiki.
- VirtualRESTService (https://github.com/wikimedia/mediawiki/commit/bff922c)
- MultiHttpClient (https://github.com/wikimedia/mediawiki/commit/ff0816c)


extensions/VisualEditor/ApiVisualEditor.php
```diff
@@ -47,6 +46,17 @@ public function getVersion() {
...
 	}
 
+	// Wikia change
+	/**
+	 * @protected
+	 * @description Simple helper to retrieve relevant host
+	 * @return String
+	 */
+	protected function getApiSource() {
+		$url = parse_url( wfExpandUrl( wfScript( 'api' ) ) );
```
inez: It seems like a unnecessary overhead to construct URL to api.php just so later you can extract from it the host part.

inez: Btw. syntax like `return parse_url( wfExpandUrl( wfScript( 'api' ) ) )['host'];` should work too

inez: Perhaps you could use global $wgCanonicalServer

extensions/VisualEditor/ApiVisualEditor.php
```diff
@@ -63,31 +73,16 @@ private function getVRSObject() {
...
+
+		$params = array(
+			'URL' => $config->get( 'VisualEditorParsoidURL' ),
+			'domain' => $this->getApiSource(),
+			'prefix' => $config->get( 'VisualEditorParsoidPrefix' ),
+			'timeout' => $config->get( 'VisualEditorParsoidTimeout' ),
+			'HTTPProxy' => $config->get( 'VisualEditorParsoidHTTPProxy' ),
+			'forwardCookies' => $config->get( 'VisualEditorParsoidForwardCookies' )
+		);
+
```
inez, line 92: @pauloslund Would you be able to "minimize" this change - as basically from what I see you want to keep what's in the "else" of original code, and just add "domain" to it. Also - we are most likely to use RESTBase later on (but soon) so that code might be useful again.

extensions/VisualEditor/ApiVisualEditor.php
```diff
@@ -63,8 +73,12 @@ private function getVRSObject() {
 		// the VRS class to use, defaults to Parsoid
 		$class = 'ParsoidVirtualRESTService';
 		$config = $this->veConfig;
+
 		// the global virtual rest service config object, if any
-		$vrs = $this->getConfig()->get( 'VirtualRestConfig' );
```
inez, line 67: This `$this->getConfig()->get( 'VirtualRestConfig' )` is just a proxy for accessing global variables - so in this particular case it accesses $wgVirtualRestConfig which is defined in newer version of MW in DefaultSettings.php - https://github.com/wikimedia/mediawiki/blob/master/includes/DefaultSettings.php#L7671-L7699
It seems that you could actually totally use it , and then you wouldn't even have to have that getApiSource method at all.
Btw. you can pass domain together with "http://" and it will work ok - https://github.com/wikimedia/mediawiki/blob/master/includes/libs/virtualrest/ParsoidVirtualRESTService.php#L69-L76

pauloslund, line 67: Oh awesome, I'll make that change

includes/Setup.php
```diff
@@ -394,6 +394,9 @@
 	$wgCanonicalServer = wfExpandUrl( $wgServer, PROTO_HTTP );
 }
 
+// Setup default domain for VirtualRest
+$wgVirtualRestConfig['global']['domain'] = $wgCanonicalServer;
```
inez, line 398: This perhaps should go into includes/wikia/DefaultSettings.php

PR #8381, opened 2015-09-08
SERVICES-707: Changed Avatar API to use new Wikia\Service\Constants.

harnash:
Missing refactor: changed Avatar API to use proper constant from Wikia\Services\Constants.

@Wikia/services-team


nmonterroso:
LGTM

PR #8382, opened 2015-09-08
fixing undefined constant error

nmonterroso:
@Wikia/services-team 
fixing undefined constant error introduced in 3e8ec2b57c7936e469e3a57d38c072601bb17f66


nmonterroso:
Note: there are now a lot of preferences saved as `MAIN_PAGE_NOTIFICATIONS_HIDDEN_PROP_NAME-<int>` :(

nmonterroso:
hm actually i wonder if we should leave this as-is? Since we have to do some migration it'll be easier to keep track of the latest preference if we leave it broken.

PR #8383, opened 2015-09-08
DAT-3192 Invalidate memcache on allinfoboxes query update

sebastianmarzjan:


idradm:
I'm not really fond of introducing new hooks, but it seems like a solid approach so let it go :snowflake:

PR #8384, opened 2015-09-08
DAT-3200 Adding support for urlencoded image names in infoboxes

sebastianmarzjan:


idradm:
LGTM :rocket: it to production

PR #8388, opened 2015-09-09
XW-259 Parser Tag Abstract

vforge:
Abstract PR for Verbatim Parser Tags.

/cc @nandy-andy @bognix @Warkot @rogatty

### DIFF
--- a/includes/wikia/nirvana/WikiaParserTagController.class.php
+++ b/includes/wikia/nirvana/WikiaParserTagController.class.php
@@ -14,7 +14,7 @@ abstract class WikiaParserTagController extends WikiaController {
 		$isValid = false;
 
 		$validator = $this->buildParamValidator( $paramName );
-		if( !empty( $validator ) && $validator instanceof WikiaValidator ) {
+		if( $validator instanceof WikiaValidator ) {
 			$isValid = $validator->isValid( $paramValue );
 
 			if( !$isValid ) {
@@ -26,10 +26,11 @@ abstract class WikiaParserTagController extends WikiaController {
 	}
 
 	/**
-	 * @desc Factory method to create validators for params; if should returns fasle if validator can't be created
+	 * @desc Factory method to create validators for params
+	 * if should return WikiaValidatorAlwaysTrue if validator can't be created or won't be used
 	 *
 	 * @param String $paramName
-	 * @return false|WikiaValidator
+	 * @return WikiaValidator
 	 */
 	abstract protected function buildParamValidator( $paramName );
 }
--- a/includes/wikia/nirvana/WikiaParserTagController.class.php
+++ b/includes/wikia/nirvana/WikiaParserTagController.class.php
@@ -14,7 +14,7 @@ abstract class WikiaParserTagController extends WikiaController {
 		$isValid = false;
 
 		$validator = $this->buildParamValidator( $paramName );
-		if( !is_null($validator) ) {
+		if( !is_null($validator) && $validator instanceof WikiaValidator ) {
 			$isValid = $validator->isValid( $paramValue );
 
 			if( !$isValid ) {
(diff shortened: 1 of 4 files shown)


includes/wikia/nirvana/WikiaParserTagController.class.php
```diff
@@ -0,0 +1,35 @@
...
+		}
+
+		return $isValid;
+	}
+
+	/**
+	 * @desc Factory method to create validators for params; returns null if validator can't be created
+	 *
+	 * @param String $paramName
+	 * @return null|WikiaValidator
```
bognix: update the desc.
Consider returning false instead of null

includes/wikia/nirvana/WikiaParserTagController.class.php
```diff
@@ -0,0 +1,35 @@
...
+	 * @param String|Mixed $paramValue value of the parameter
+	 * @param String $errorMessage reference to a string variable which will get error message if one occurs
+	 *
+	 * @return bool
+	 */
+	protected function isTagParamValid( $paramName, $paramValue, &$errorMessage ) {
+		$isValid = false;
+
+		$validator = $this->buildParamValidator( $paramName );
+		if( !is_null($validator) && $validator instanceof WikiaValidator ) {
```
bognix: whitespaces

includes/wikia/nirvana/WikiaParserTagController.class.php
rogatty: How is `false` better than `null`? It unnecessarily mixes return type.

vforge: `null` is non-descriptive - any type can be `null` and it doesn't have value per se. see https://www.lucidchart.com/techblog/2015/08/31/the-worst-mistake-of-computer-science/

hakubo: to me false is more descriptive
it's like program do tell you that something
false is a value

null on the other hand is also falsy
but null is not a value

null is nothing basically, it's just happens to be falsy

to be honest it would make sense if this function throws if it can't create a validator
as that means that someone is trying to create something that cannot exist

nandy-andy: Yep, from what I remember from wise books there are two "correct" ways:
a) thrown an exception
b) return `NullValidator`

a) has the downside that you'll then have to handle an exception everywhere you use the factory, b) is easier to work with basically because it's still an instance of validator which does nothing.

bognix:
LGTM

includes/wikia/nirvana/WikiaParserTagController.class.php
```diff
@@ -0,0 +1,36 @@
...
+	}
+
+	/**
+	 * @desc Factory method to create validators for params
+	 * if should return WikiaValidatorAlwaysTrue if validator can't be created or won't be used
+	 *
+	 * @param String $paramName
+	 * @return WikiaValidator
+	 */
+	abstract protected function buildParamValidator( $paramName );
```
rogatty, line 35: How about removing `abstract` and returning `WikiaValidatorAlwaysTrue` by default? Does it make sense?

vforge, line 35: I would rather keep it - it enforces you to implement `buildParamValidator` when you implement class.

bognix, line 35: :+1: to what V said. It makes you to declare WikiaValidatorAlwaysTrue as return type explicitly if you want to use it

hakubo:
LGTM

nandy-andy:
LGTM :+1:

rogatty:
LGTM

PR #8398, opened 2015-09-09
DAT-3073 set max height for infobox images on desktop

dianafa:
https://wikia-inc.atlassian.net/browse/DAT-3073

### DIFF
--- a/extensions/wikia/PortableInfobox/services/Helpers/PortableInfoboxRenderServiceHelper.php
+++ b/extensions/wikia/PortableInfobox/services/Helpers/PortableInfoboxRenderServiceHelper.php
@@ -39,7 +39,7 @@ class PortableInfoboxRenderServiceHelper {
 					$horizontalGroupData[ 'renderLabels' ] = true;
 				}
 			} else if ( $item[ 'type' ] === 'header' ) {
--				$horizontalGroupData[ 'header' ] = $data[ 'value' ];
+				$horizontalGroupData[ 'header' ] = $data[ 'value' ];
 			}
 		}
 
--- a/extensions/wikia/PortableInfobox/services/Helpers/PortableInfoboxRenderServiceHelper.php
+++ b/extensions/wikia/PortableInfobox/services/Helpers/PortableInfoboxRenderServiceHelper.php
@@ -11,7 +11,7 @@ class PortableInfoboxRenderServiceHelper {
 	const DESKTOP_THUMBNAIL_WIDTH = 270;
 	const MOBILE_THUMBNAIL_WIDTH = 360;
 	const MINIMAL_HERO_IMG_WIDTH = 300;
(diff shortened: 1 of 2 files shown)


extensions/wikia/PortableInfobox/services/Helpers/PortableInfoboxRenderServiceHelper.php
```diff
@@ -37,8 +38,11 @@ public function createHorizontalGroupData( $groupData ) {
 				if ( !empty( $data[ 'label' ] ) ) {
 					$horizontalGroupData[ 'renderLabels' ] = true;
 				}
-			} else if ( $item[ 'type' ] === 'header' ) {
-				$horizontalGroupData[ 'header' ] = $data[ 'value' ];
+			}
+			else {
+				if ( $item[ 'type' ] === 'header' ) {
```
sebastianmarzjan: What's the rationale behind change in this condition?

dianafa: sorry, my IDE got crazy :see_no_evil:

idradm:
LGTM

extensions/wikia/PortableInfobox/tests/PortableInfoboxRenderServiceHelperTest.php
```diff
@@ -296,4 +301,127 @@ public function testIsTypeSupportedInTemplatesDataProvider() {
...
+		return [
+			[
+				'mockParams' => [
+					'fileHeight' => 2000,
+					'fileWidth' => 3000
+				],
+				'isWikiaMobile' => false,
+				'result' => [
+					'height' => 500,
+					'width' => 270
```
sebastianmarzjan, line 341: Shouldn't this case generate:

```
[
                    'height' => 180,
                    'width' => 270
]
```

As far as I recall we were supposed to avoid cropping images / changing aspect ratio, unless I understand the test case wrongly

dianafa, line 341: this are the sizes we pass to `transform` function, not the sizes we want image to have. It automatically adjusts the correct sizes, that is creates thumbnail with sizes not bigger than passed, keeping the aspect ratio. That's why, the real sizes of the image (which we pass to html attrs) we take from:
`$data[ 'height' ] = $thumbnail->getHeight();
$data[ 'width' ] = $thumbnail->getWidth();`

sebastianmarzjan, line 341: It hit me after I left the comment, thanks for the explanation too. 
What do you think about leaving a comment here so it's explicit to future readers? :)

dianafa, line 341: :dancer:

PR #8400, opened 2015-09-09
PLATFORM-1417 - DatabaseBase::isWriteQuery - ignore leading whitespaces

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-1417

This is actually a "SELECT" query despite all leading whitespaces

``` sql

            SELECT comment_id FROM comments_index
                WHERE parent_page_id = 12648  and deleted = 0 and removed = 0 and parent_comment_id = 0 
                ORDER BY last_child_comment_id desc
                LIMIT 0, 25
```

This will allow Message Wall to work correctly on Reston instead of throwing 20k of `DB readonly mode` exceptions each hour.


wladekb:
lgtm

PR #8404, opened 2015-09-09
DAT-3204 filter out subpage from infobox template list

idradm:
https://wikia-inc.atlassian.net/browse/DAT-3204


extensions/wikia/PortableInfobox/querypage/AllinfoboxesQueryPage.php
```diff
@@ -104,9 +104,14 @@ public function reallyDoQuery( $limit = false, $offset = false ) {
 			} );
 
 		return array_filter( $result, function ( $tmpl ) {
-			$data = PortableInfoboxDataService::newFromPageID( $tmpl[ 'pageid' ] )->getData();
+			$title = Title::newFromID( $tmpl[ 'pageid' ] );
+			if ( $title && $title->exists() && !$title->isSubpage() ) {
+				$data = PortableInfoboxDataService::newFromTitle( $title )->getData();
```
sebastianmarzjan: Based on our offline conversation about use cases (doc / Draft / real hierarchy) what do you think: should we eliminate all subpages, or for now just the doc and draft ones?

sebastianmarzjan:
One question to consider on the actual way we address the problem; code LGTM

extensions/wikia/PortableInfobox/querypage/AllinfoboxesQueryPage.php
idradm: This question should be directed to Jakub. However this approach shields us from changes in names of subpages (especially translation connected). Is there any major opposing point for that?

sebastianmarzjan: From my perspective it depends on the scope of such usage among the existing infoboxes. 
I don't think that it's a blocker; However, I suggest to validate this either as a follow-up task or part of testing (preferred) should we go with the approached proposed with this PR.

extensions/wikia/PortableInfobox/querypage/AllinfoboxesQueryPage.php
```diff
@@ -106,9 +107,20 @@ public function reallyDoQuery( $limit = false, $offset = false ) {
 			} );
 
 		return array_filter( $result, function ( $tmpl ) {
-			$data = PortableInfoboxDataService::newFromPageID( $tmpl[ 'pageid' ] )->getData();
+			$title = Title::newFromID( $tmpl[ 'pageid' ] );
 
-			return !empty( $data );
+			if ( $title && $title->exists() &&
```
sebastianmarzjan: In order to make this logic flow easier to read, please consider making this conditional a method that returns boolean ( i.e. isValidInfoboxPageTitle($title) )

sebastianmarzjan:
One thing to consider, in general LGTM

PR #8405, opened 2015-09-09
DatabaseMysqli: log the last query when assert fails

macbre:
As we can pass additional context to `Assert::true` since bd3583bf4e112ae88aeab76cba1cc4af835ed7b4

@michalroszka


ghost:
Excellent.

PR #8406, opened 2015-09-09
XW-216 | Create SoundCloudTag

vforge:
/cc @bognix @Warkot


extensions/wikia/SoundCloudTag/SoundCloudTagController.class.php
```diff
@@ -0,0 +1,44 @@
...
+		}
+
+		foreach ( $allowedArgs as $name ) {
+			if ( isset( $args[$name] ) ) {
+				$attributes[$name] = $args[$name];
+			}
+		}
+
+		$attributes['data-tag'] = 'soundcloud';
+		$attributes['sandbox'] = 'allow-scripts allow-same-origin';
```
rogatty: Why would we allow users to define things like that? What's the business value? I found options that are far more useful and IMO they should be here instead of technicalities:
- `auto_play`
- `color`
- `hide_related`
- `show_comments`
- `show_user`
- `show_reposts`
- `visual`

rogatty: https://developers.soundcloud.com/docs/api/html5-widget#params

rogatty:
Please do it like VolDev did in terms of parameters that we accept:
https://github.com/Wikia/app/pull/8198/files#diff-f9128714a37a935b74cae60b94783da9R24

User doesn't even have to click "Embed", URL to track/playlist/etc is enough.

extensions/wikia/SoundCloudTag/SoundCloudTagController.class.php
```diff
@@ -0,0 +1,44 @@
...
+	public function renderTag( $input, array $args, Parser $parser, PPFrame $frame ) {
+		return Html::element(
+			'iframe',
+			$this->prepareAttributes( $args ),
+			wfMessage( 'soundcloud-tag-could-not-render' )->text()
+		);
+	}
+
+	private function prepareAttributes( array $args ) {
+		$allowedArgs = ['width', 'height', 'scrolling', 'frameborder', 'src'];
```
rogatty: Please use the same list as in https://github.com/Wikia/app/pull/8198/files#diff-f9128714a37a935b74cae60b94783da9R24

Also user doesn't have to click "Embed" and copy the iframe, `url` parameter (to track/playlist/etc) is enough.

includes/wikia/helpers/tests/WikiaIframeTagBuilderHelperTests.class.php
```diff
@@ -0,0 +1,20 @@
...
+	public function setUp() {
+		parent::setUp();
+	}
+
+	public function testWrapForMobile() {
+		$helper = new WikiaIframeTagBuilderHelper();
+		$this->getMethodMock( $helper, 'isMobileSkin' )
+			->expects( $this->any() )
+			->method( 'isMobileSkin' )
+			->will( $this->returnValue( true ) );
```
bognix: how about unit test when isMobileSkin returns false?

includes/wikia/helpers/WikiaIframeTagBuilderHelper.class.php
```diff
@@ -0,0 +1,13 @@
+<?php
+class WikiaIframeTagBuilderHelper extends WikiaTagBuilderHelper {
```
hakubo: I believe this should be `WikiaIFrameTagBuilderHelper`

tomasznapieralski:
lgtm

bognix:
LGTM

includes/wikia/helpers/WikiaIFrameTagBuilderHelper.class.php
```diff
@@ -0,0 +1,13 @@
+<?php
+class WikiaIFrameTagBuilderHelper extends WikiaTagBuilderHelper {
+
+	public function isMobileSkin( ) {
+		return F::app()->checkSkin( [ 'wikiamobile', 'mercury' ] );
```
rogatty: Is it ever `mercury`?

tomasznapieralski:
LGTM

includes/wikia/DefaultSettings.php
```diff
@@ -1797,3 +1798,8 @@
  * flag for using (possibly external) service for managing a user's preferences
  */
 $wgPreferencesUseService = false;
+
+/**
+ * Parser Tag
+ */
+require_once "$IP/extensions/wikia/SoundCloudTag/SoundCloudTag.setup.php" ;
```
gabrys, line 1805: I don't believe this is a good place for requires

@vforge @macbre @owend

vforge, line 1805: As far as I remember, we can put site-wide enabled extension in this file. See https://github.com/Wikia/app/pull/8406#discussion-diff-39293694R606

Can you please share an idea about better place to put those?

gabrys, line 1805: I would set a $wgEnableXXXExt = true in here and then include the setup file in CommonExtensions as usual. I'm not the expert though, since @owend and @macbre ;-)

owend, line 1805: It seems there are a very small number of other extension inclusions in wikia/DefaultSettings... and one of them even has a TODO that says "move this to commonextensions".  I agree that since most extensions are enabled globally with a variable in Default settings, this is a better practice, even though it's an otherwise useless/unused var.

macbre, line 1805: @gabrys - we do have includes / requires in `wikia/DefaultSettings.php`, but they are used for core features - e.g. WikFactory, AssetsManager.

`SoundCloudTag` looks like a Yet Another Extension :tm:  that can be enabled / disabled without changing the functionality of the whole site. I'd move the inclusion to the config repo. It's actually a matter of personal preferences :)

vforge, line 1805: @macbre thanks for explanation!

PR #8408, opened 2015-09-10
[SERVICES-723] Fixing helios client status check

ArturKlajnerok:
Fix the comparison of the request status to the integer value and not the object.

/cc @Wikia/services-team


drsnyder:
LGTM

nmonterroso:
does this work? the default is 

``` php
protected $respStatus = "200 Ok";
```

nmonterroso:
spoke offline with @ArturKlajnerok, LGTM

ArturKlajnerok:
Logged out values from `$request->getStatus()` and there are http status codes in proper integer form, like `204`

PR #8423, opened 2015-09-10
PLATFORM-1460: increase wgRCMaxRows to 20k

macbre:
Keep more entries in `recentchanges` table on the most active wikis

https://wikia-inc.atlassian.net/browse/PLATFORM-1460

@michalroszka / @wladekb / @lgarczewski


lgarczewski:
LGTM, thanks Macbre.

PR #8424, opened 2015-09-10
VE 1918: Fix visual editor font sizing

pauloslund:


inez:
Icons in the dropdowns (and I think in the toolbar itself) don't look very sharp.

inez:
I just confirmed if you change that font-size from 12px to 13px the toolbar itself look way more sharp (still problem with content of dropdowns).

PR #8425, opened 2015-09-10
SEO-51 New robots.txt as an extension

gabrys:
Robots.txt is currently generated by redirect-robots.php file in the main directory. It has a few problems:
- A separate section for Googlebot
- Using a NoIndex keyword, which according to the Google docs is not recognized by the bot
- It's quite hard to mainain
- It uses wildcards when they are not necessary (at the ends of lines). Remember some robots may not understand wildcards
- It allows crawling special pages by their English and local wiki language names, but not mixed ones (Spezial:Random)
- You cannot test robots on devboxes or sandboxes because the code always serves a "Disallow everything" rule on those.

This PR basically fixes all the above by supplying a RobotsTxt class with a clean API (and a unit test) and a new version of the entry PHP script: wikia-robots-txt.php, which can be used to add more special pages to white list or to block more bots.

I want wikia-robots-txt.php to eventually replace redirect-robots.php (once we prove there's no negative impact of the changes proposed).

### DIFF
--- a/extensions/wikia/CreateNewWiki/CreateNewWiki.alias.php
+++ b/extensions/wikia/CreateNewWiki/CreateNewWiki.alias.php
@@ -2,10 +2,10 @@
 /**
  * Aliases for special pages of CreateNewWiki extension.
  */
-$specialPageAliases = array();
+$specialPageAliases = [];
 
 /** English
  */
-$specialPageAliases['en'] = array(
-	'CreateNewWiki' => array( 'CreateNewWiki', 'CreateWiki' ),
-);
+$specialPageAliases['en'] = [
+	'CreateNewWiki' => [ 'CreateNewWiki', 'CreateWiki' ],
+];
--- a/extensions/wikia/CreateNewWiki/CreateNewWiki.alias.php
+++ b/extensions/wikia/CreateNewWiki/CreateNewWiki.alias.php
@@ -1,0 +1,11 @@
+<?php
+/**
+ * Aliases for special pages of CreateNewWiki extension.
+ */
+$specialPageAliases = array();
+
+/** English
+ */
+$specialPageAliases['en'] = array(
+	'CreateNewWiki' => array( 'CreateNewWiki', 'CreateWiki' ),
+);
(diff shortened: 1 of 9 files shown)


wikia-robots-txt.php
```diff
@@ -0,0 +1,80 @@
...
+require_once( __DIR__ . '/includes/WebStart.php' );
+require_once( __DIR__ . '/includes/Setup.php' );
+
+if ( empty( $wgEnableRobotsTxtExt ) ) {
+	exit;
+}
+
+$robots = new RobotsTxt();
+
+if ( !empty( $_SERVER['HTTP_X_STAGING'] ) ) {
```
owend: should probably block devboxes here too (just to be safe, i know that's currently done another way now)

extensions/wikia/Sitemap/SpecialSitemap.php
```diff
@@ -39,7 +39,7 @@
  */
 $dir = dirname( __FILE__ ) . '/';
 $wgExtensionMessagesFiles['Sitemap'] = $dir . 'Sitemap.i18n.php';
-$wgExtensionMessagesFiles['Sitemap'] = $dir . 'Sitemap.alias.php';
+$wgExtensionMessagesFiles['SitemapAlias'] = $dir . 'Sitemap.alias.php';
```
owend, line 42: this change is unrelated to robots.txt, but definitely a bug fix.

owend:
Overall this LGTM

extensions/wikia/CreateNewWiki/CreateNewWiki.alias.php
```diff
@@ -0,0 +1,11 @@
+<?php
+/**
+ * Aliases for special pages of CreateNewWiki extension.
+ */
+$specialPageAliases = array();
```
lizlux: let's use php 5.4 syntax for arrays `[]` here and below.

wikia-robots-txt.php
```diff
@@ -0,0 +1,79 @@
...
+
+	// Params
+	$robots->disallowParam( 'action' );
+	$robots->disallowParam( 'feed' );
+	$robots->disallowParam( 'printable' );
+	$robots->disallowParam( 'useskin' );
+	$robots->disallowParam( 'uselang' );
+
+	// Nasty robots
+	$robots->blockRobot( 'IsraBot' );
```
lizlux, line 38: Might be nice to have these values as an array and loop through them to allow, disallow, block, etc. Would make this file more DRY.

lizlux:
Just one minor comment but this LGTM.

PR #8428, opened 2015-09-10
(CE-2670) Properly escape user name in remove avatar link attribute

Grunny:
Properly escape user name in remove avatar link attribute in the user profile
boxes.

/cc @jsutterfield

### DIFF
--- a/extensions/wikia/UserProfilePageV3/templates/UserProfilePage_renderUserIdentityBox.php
+++ b/extensions/wikia/UserProfilePageV3/templates/UserProfilePage_renderUserIdentityBox.php
@@ -13,7 +13,7 @@
 			<?php if ( $canRemoveAvatar ): ?>
 				<span>
 					<img src="<?= $wgBlankImgUrl ?>" class="sprite trash">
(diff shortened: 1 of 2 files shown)


extensions/wikia/UserProfilePageV3/templates/UserProfilePage_renderUserIdentityBox.php
```diff
@@ -13,7 +13,7 @@
 			<?php if ( $canRemoveAvatar ): ?>
 				<span>
 					<img src="<?= $wgBlankImgUrl ?>" class="sprite trash">
-					<a id="UserAvatarRemove" data-name="<?= $user['name']; ?>" href="#" data-confirm="<?= wfMessage( 'user-identity-remove-confirmation' )->escaped(); ?>">
+					<a id="UserAvatarRemove" data-name="<?= Sanitizer::encodeAttribute( $user['name'] ); ?>" href="#" data-confirm="<?= wfMessage( 'user-identity-remove-confirmation' )->escaped(); ?>">
```
jsutterfield: What do you think about adding 'name' to the list we're already escaping [here](https://github.com/Wikia/app/blob/dev/extensions/wikia/UserProfilePageV3/UserIdentityBox.class.php#L124)?

If we did that, we could also remove the other escaping we're doing [here](https://github.com/Wikia/app/blob/dev/extensions/wikia/UserProfilePageV3/UserIdentityBox.class.php#L124)

jsutterfield:
LGTM!

PR #8429, opened 2015-09-10
[SERVICES-739] Global and local preferences 

ArturKlajnerok:
This is introducing new domain for `Preferences` that will contain global and local preferences for user. We need to support that for further integration and testing of `user-preferences` service.

It's work in progress so feel free to take it over.

/cc @Wikia/services-team


lib/Wikia/src/Domain/User/Preferences.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+namespace Wikia\Domain\User;
+
+use Wikia\Util\Assert;
+
+class Preferences {
+
+    private $globalPreferences;
```
nmonterroso, line 10: we should annotate these so IDEs can figure out usages. Like:

``` php
/** @var GlobalPreference[] */
private $globalPreferences;

...
/**
 * @return GlobalPreference[]
 */
public function getGlobalPreferences()
```

ArturKlajnerok, line 10: fixed in : fd1e0b2

nmonterroso:
LGTM, we can build off of this

PR #8432, opened 2015-09-11
CE-2676: Fix loading scripts with the same name

lukaszkonieczny:
If we want to import pages with the same name (via `importArticles` method), but from different subdomains ie.

from local (Mediawiki:Script.js) 
and external (external:subdomain:Mediawiki:Script.js) 

the last overwrites the first.


lukaszkonieczny:
@wladekb @macbre

wladekb:
lgtm

PR #8435, opened 2015-09-11
remove skins/Venus.php

ghost:
`Skin::getSkinNames` scans the `skins/` dir and `Venus.php` is there
which makes `venus` a valid skin and if the class does not exist
(`SkinVenus` in this case), MW will make it exist by including
`skins/Venus.php`. That, in turn will cause fatal errors because
the skin has been disabled in config and `Venus.setup.php`
is no longer included.

https://wikia-inc.atlassian.net/browse/XW-117

@hakubo


hakubo:
LGTM

vforge:
LGTM

macbre:
:+1:

PR #8437, opened 2015-09-11
CE-2677 Make the destroying session data method testable and write the tests

adamkarminski:
/cc @kamilkoterba @Grunny

### DIFF
--- a/extensions/wikia/ContentReview/tests/ContentReviewHooksTest.php
+++ b/extensions/wikia/ContentReview/tests/ContentReviewHooksTest.php
@@ -12,19 +12,21 @@ class ContentReviewHooksTest extends WikiaBaseTest {
 	 * because there are no IDs of wikias stored in a session under the test mode key.
 	 */
 	public function testDisableTestModeNoSessionData() {
-		$requestMock = $this->getMockBuilder( 'WebRequest' )
-			->disableOriginalConstructor()
-			->setMethods( [ 'getSessionData', 'setSessionData' ] )
-			->getMock();
-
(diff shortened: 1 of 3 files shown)


kamilkoterba:
LGTM

extensions/wikia/ContentReview/tests/ContentReviewHooksTest.php
```diff
@@ -0,0 +1,49 @@
...
+
+	/**
+	 * In this test case the setSessionData() method should not be called
+	 * because there are no IDs of wikias stored in a session under the test mode key.
+	 */
+	public function testDisableTestModeNoSessionData() {
+		$requestMock = $this->getMockBuilder( 'WebRequest' )
+			->disableOriginalConstructor()
+			->setMethods( [ 'getSessionData', 'setSessionData' ] )
+			->getMock();
```
Grunny: It's simpler to just use `getMock` unless you need to disable the constructor, which you don't here, i.e. just:
`$requestMock = $this->getMock( 'WebRequest', [ 'getSessionData', 'setSessionData' ] );`

Grunny:
I think it would be better to just fully test the hook, i.e. test `onUserLogoutComplete` by simply injecting a mocked `User` object and just mock its `getRequest` method. I'm not sure of the value of splitting out a method and only testing that here. Then `disableTestMode` should be made private as it's an implementation detail (and either static, or switch the entire class to have the hooks be non-static using one static method with the `Hooks::register` calls), or it could even end up in a helper class if there is enough need.

kamilkoterba:
Great!

PR #8442, opened 2015-09-11
CE-2678 Check for BadRequestApiException for submitPageForReview API …

kamilkoterba:
…method

https://wikia-inc.atlassian.net/browse/CE-2678

@adamkarminski, @Grunny


adamkarminski:
If the checks pass - it LGTM :)

PR #8450, opened 2015-09-11
[SERVICES-748] Update user to support global and local preferences methods

ArturKlajnerok:
This should provide proper support in MW for changes to global and local preferences around `PreferenceService`.

/cc @Wikia/services-team


includes/User.php
```diff
@@ -2569,9 +2568,26 @@ public function getIntOption( $oname, $defaultOverride=0 ) {
...
-		$globalPref = self::localToGlobalPropertyName($pref, $cityId, $sep);
-		return $this->getGlobalPreference($globalPref, $default, $ignoreHidden);
+	public function getLocalPreference($preference, $cityId = null, $sep = "-", $default = null, $ignoreHidden = false) {
+		global $wgPreferencesUseService;
+
+		if ($wgPreferencesUseService) {
+			$this->load();
+			$value = $this->userPreferences()->getLocalPreference($this->mId, $preference, $default, $ignoreHidden);
+			wfRunHooks(
+				'UserGetPreference',
```
nmonterroso: does the implementation of this hook need to change? `getPreferences` used to return a map, but now it returns a strongly typed object

nmonterroso:
one comment otw lgtm

nmonterroso:
I left a comment and github says it's outdated, but it's not:

> does the implementation of this hook need to change? getPreferences used to return a map, but now it returns a strongly typed object

It looks like there's only one implementation of the hook, and it relies on being able to access preferences as an array, so that definitely needs to be updated

includes/User.php
```diff
@@ -2569,9 +2568,26 @@ public function getIntOption( $oname, $defaultOverride=0 ) {
...
+			$this->load();
+			$value = $this->userPreferences()->getLocalPreference($this->mId, $cityId, $preference, $default, $ignoreHidden);
+			wfRunHooks(
+				'UserGetPreference',
+				[
+					$this->userPreferences()->getPreferences($this->mId),
+					$preference,
+					&$value
+				]
+			);
```
drsnyder: Why is this hook run here but not below?

drsnyder:
One clarification, OTW, LGTM.

includes/User.php
ArturKlajnerok: I've updated usage of `UserGetPreference` in : 0e37b43 , it seems that the only usage is based on global preferences array.

includes/User.php
ArturKlajnerok: @drsnyder 
It came out that `UserGetPreference` hook need to be only used with global preferences, this is the only usage : https://github.com/Wikia/app/blob/48f8caf925b02009e9b20033eab437a7f972faf5/extensions/wikia/RTE/RTE.class.php#L680

But going back to your question, this hook is also called for `getOptionHelper` but inside the method : https://github.com/Wikia/app/blob/68ca84ba1e509d998893c6c883fbf3eb9b929bcc/includes/User.php#L2496

ArturKlajnerok:
I've answered to feedback directly under specific comments :+1:

nmonterroso:
LGTM :shipit:

PR #8452, opened 2015-09-12
INT-185: Adding homepage URL for a user language

mari-mcmurtrie:
In Mercury, wikia logo in header is linked to "www.wikia.com" (hard-coded), and we want to internationalize it so that header in Japan (or any other locale) wikia goes to its locale's homepage (e.g. ja.wikia.com). After discussion, we decided to generate such url in backend and return it as wiki variable to Mercury. Please also refer to https://github.com/Wikia/mercury/pull/1351.


extensions/wikia/MercuryApi/models/MercuryApi.class.php
```diff
@@ -126,7 +126,8 @@ public function getWikiVariables() {
 			'siteMessage' => $this->getSiteMessage(),
 			'siteName' => $wgSitename,
 			'theme' => SassUtil::getOasisSettings(),
-			'wikiCategories' => WikiFactoryHub::getInstance()->getWikiCategoryNames( $wgCityId )
+			'wikiCategories' => WikiFactoryHub::getInstance()->getWikiCategoryNames( $wgCityId ),
+			'homepage' => ( new WikiaLogoHelper() )->getCentralUrlForLang( $wgLang->getCode() )
```
rogatty: Keep the alphabetic order, please.

Maybe we should also add a check for `WikiaLogoHelper` to not break Mercury when this class is missing somehow. How about a new method `$this->getHomepageUrl()`?

lizlux: Or a test could be a good way to ensure this value is there.

lizlux:
LGTM

PR #8455, opened 2015-09-14
ADEN-2417 Recoverable slots and new queue

fraszczakszymon:


extensions/wikia/AdEngine/js/SlotTweaker.js
```diff
@@ -33,6 +33,9 @@ define('ext.wikia.adEngine.slotTweaker', [
 		if (slot) {
```
nandy-andy: What do you think about merging the conditions in terms of readability?

```
if (a && b) {
   // do something for a
   // do something for b
} else if (a && !b) {
   //do something for a
}
```

instead of:

```
if (a) {
   // do something for a
   if (b) {
      // do something for b
   }
}
```

extensions/wikia/AdEngine/js/run/desktop.run.js
```diff
@@ -77,6 +80,14 @@ require([
 		win.adslots2 = win.adslots2 || [];
 		adEngine.run(adConfigDesktop, win.adslots2, 'queue.desktop');
 
+		if (context.opts.sourcePoint && win.ads) {
```
nandy-andy, line 85: What do you think about not executing `adEngine` for regular slots if we detect ad blockers?

Nevermind. I don't think it's a best idea to risk executing our regular queue if SP detection from unknown reasons stops working ;)

nandy-andy:
One minor comment about indentation and conditions. Except it, LGTM :+1:

extensions/wikia/AdEngine/js/provider/gpt/helper.js
```diff
@@ -113,10 +147,21 @@ define('ext.wikia.adEngine.provider.gpt.helper', [
 		}
 
 		if (!googleApi.isInitialized()) {
-			googleApi.init();
+			googleApi.init(function () {
```
nandy-andy: Let's change it as we discussed offline :+1:

nandy-andy:
LGTM :+1:

PR #8457, opened 2015-09-14
SOC-1177 | Escape Chat frontend messages

kvas-damian:
https://wikia-inc.atlassian.net/browse/SOC-1177

Added new method for `mw.message` - `rawParams`. This way we can pass params that should not be escaped when using `escaped` format.

ping @Grunny @garthwebb @rafalkalinski 
PS. It's easier to use https://github.com/Wikia/app/compare/dev...4bbc4a54259d0ace94332835e017e7a2b3b33b9d for reviewing changes

### DIFF
--- a/extensions/wikia/Chat2/Chat_setup.php
+++ b/extensions/wikia/Chat2/Chat_setup.php
@@ -158,6 +158,31 @@ JSMessages::registerPackage('ChatEntryPoint', array(
 	'chat-member-since'
 ));
 
+/**
+ * ResourceLoader module
+ */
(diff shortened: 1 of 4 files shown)


rafalkalinski:
LGTM

Grunny:
LGTM. :) Might be cool to contribute this back up to the latest MW core as well if it applies easily enough, since they could make use of it as well.

kvas-damian:
Will try when it passes QA.

garthwebb:
LGTM

PR #8460, opened 2015-09-14
CE-677 Remove debuging logs after fix

kamilkoterba:
Since it's fixed we don't need debuging logs

https://wikia-inc.atlassian.net/browse/CE-677


macbre:
Thanks for keeping the code clean :+1:

PR #8461, opened 2015-09-14
Fix unit tests by changing parent

vforge:
@nandy-andy @bognix @wladekb


nandy-andy:
LGTM but it's sad :crying_cat_face:

PR #8463, opened 2015-09-14
ADEN-2438 fix hub verticals

nandy-andy:
After refactoring AdEng code and forcing our slot paths to have only one out of three verticals: ent, gaming and life. We found an issues with hub pages and slot paths there. The changes below fixes it and cleans the code.

Also, I removed one line in a unit test which was failing on my local machine.


fraszczakszymon:
LGTM :+1:

PR #8466, opened 2015-09-14
Add Global/Local preference concept to MW service layer

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-749

Adding the notion of local and global preferences to the preference service layer in MW.


lib/Wikia/src/Domain/User/Preferences/UserPreferences.php
```diff
@@ -0,0 +1,83 @@
...
+		$this->localPreferences = [];
+	}
+
+	public function setGlobalPreference($name, $value) {
+		$this->globalPreferences[$name] = new GlobalPreference($name, $value);
+		return $this;
+	}
+
+	public function setLocalPreference($name, $wikiId, $value) {
+		$this->localPreferences[$wikiId][$name] = new LocalPreference($name, $value, $wikiId);
```
ArturKlajnerok: Do we need to assign it to key based on `wikiId`, is there any use-case when there will be loaded local preferences for multiple wikis ??

nmonterroso: I'm unsure of what you're asking. This definitely supports multiple wikis having the same preference. The full path is `$this->localPreferences[$wikiId][$name]`, so maybe you missed it? Or do you mean something else?

ArturKlajnerok: I was thinking more to put a constraints on that implementation to have global preferences and local preferences for one wiki, to limit the possible size of local ones, as request to MW use local preferences only for specific wiki.

nmonterroso: I don't know if we should impose that restriction at the service layer. It's up to whatever is using the service to make the decision to load data for multiple wikis, and if it's loading data for 500k wikis, that's a problem they should fix on their end, but we shouldn't restrict it here imo

lib/Wikia/src/Domain/User/Preferences/UserPreferences.php
```diff
@@ -0,0 +1,83 @@
...
+		return $this->globalPreferences;
+	}
+
+	/**
+	 * @return array[wikiId => [name => LocalPreference]]
+	 */
+	public function getLocalPreferences() {
+		return $this->localPreferences;
+	}
+}
```
drsnyder, line 83: Please apply the PHP code formatting to the new files.

lib/Wikia/src/Persistence/User/Preferences/PreferencePersistence.php
```diff
@@ -12,21 +12,20 @@
...
 	 *
 	 * @param int $userId
-	 * @param GlobalPreference[] $preferences
+	 * @param UserPreferences $preferences
 	 * @return true success, false or exception otherwise
 	 * @throws PersistenceException
 	 * @throws UnauthorizedException
 	 */
-	public function save( $userId, array $preferences );
+	public function save( $userId, UserPreferences $preferences );
```
drsnyder, line 20: :+1: it's nice to be able to use the type here

lib/Wikia/src/Service/User/Preferences/PreferenceModule.php
```diff
@@ -10,68 +11,28 @@
 use Wikia\Persistence\User\Preferences\PreferencePersistenceSwaggerService;
 
 class PreferenceModule implements Module {
-	const SWAGGER_SERVICE_RAMP_USAGE = 0;
+	const PREFERENCE_CACHE_VERSION = 1;
```
drsnyder, line 14: Could this (cache version) be config driven?

lib/Wikia/src/Service/User/Preferences/PreferenceService.php
```diff
@@ -2,26 +2,243 @@
...
-	 * @param int $userId
-	 * @return GlobalPreference[]
+	 * @param $userId
+	 * @return UserPreferences
+	 * @throws \Exception
 	 */
-	public function getPreferences( $userId );
+	private function load($userId) {
+		if ($userId == 0) {
+			return [];
```
ArturKlajnerok: Should we log that, or it's common in MW to load preferences for anon users ?

lib/Wikia/src/Service/User/Preferences/PreferenceService.php
```diff
@@ -2,26 +2,243 @@
...
-use Wikia\Domain\User\GlobalPreference;
+use Doctrine\Common\Cache\CacheProvider;
+use Wikia\Domain\User\Preferences\LocalPreference;
+use Wikia\Domain\User\Preferences\UserPreferences;
+use Wikia\Logger\Loggable;
+use Wikia\Persistence\User\Preferences\PreferencePersistence;
+use Wikia\Util\WikiaProfiler;
 
-interface PreferenceService {
+class PreferenceService {
```
drsnyder: Should this still satisfy an interface? It would make the consumers of this more testable.

lib/Wikia/src/Service/User/Preferences/PreferenceService.php
nmonterroso: I wouldn't be surprised. In either case I don't think we need to log it

lib/Wikia/src/Service/User/Preferences/PreferenceService.php
```diff
@@ -2,26 +2,243 @@
...
+
+		$this->load($userId)->setGlobalPreference($name, $value);
+	}
+
+	public function deleteGlobalPreference($userId, $name) {
+		$this->load($userId)->deleteGlobalPreference($name);
+	}
+
+	public function getLocalPreference($userId, $wikiId, $name, $default = null, $ignoreHidden = false) {
+		if (in_array($name, $this->hiddenPrefs) && !$ignoreHidden) {
```
drsnyder: Do you mind documenting the meaning of hidden preferences. Is a hidden preference always the default?

lib/Wikia/src/Service/User/Preferences/PreferenceService.php
nmonterroso: The class that `PreferenceService` is replacing was not behind an interface, so I removed the interface here along with its only implementation (`PreferenceKeyValueService`) since they seemed redundant. 

I guess I can add an interface here, though realistically the only consumer of this is the `User` object, which is untestable for much bigger reasons.

lib/Wikia/src/Service/User/Preferences/PreferenceService.php
```diff
@@ -2,26 +2,243 @@
...
+
+		return $default;
+	}
+
+	public function setLocalPreference($userId, $wikiId, $name, $value) {
+		if ($value == null) {
+			$value = $this->getFromDefault($name);
+		}
+
+		$this->load($userId)->setLocalPreference($name, $wikiId, $value);
```
drsnyder: Will this fail when [`$userId`](https://github.com/Wikia/app/blob/SERVICES-749/lib/Wikia/src/Service/User/Preferences/PreferenceService.php#L184) is 0?

drsnyder: Should load always return an object even if it's one that's a noop for the anonymous user?

lib/Wikia/src/Service/User/Preferences/PreferenceModule.php
nmonterroso, line 14: From a `$wg...` variable? Yes, but I don't think we should add to that noise, since nothing else will use that value.

lib/Wikia/src/Service/User/Preferences/PreferenceService.php
nmonterroso: I think that's the intention, though I'm unsure. This is duplicated logic from what MW was doing.

lib/Wikia/src/Service/User/Preferences/PreferenceService.php
nmonterroso: Hm good point. the `[]` return from `load` is leftover from when this was maintained as a map rather than an object. I'll change it to return an empty `UserPreferences` object

ArturKlajnerok:
Lgtm :+1:

lib/Wikia/src/Service/User/Preferences/PreferenceService.php
nmonterroso: 5aeabd1

lib/Wikia/src/Service/User/Preferences/PreferenceService.php
nmonterroso: 5aeabd1

lib/Wikia/src/Service/User/Preferences/PreferenceService.php
```diff
@@ -2,26 +2,243 @@
...
+	}
+
+	public function deleteLocalPreference($userId, $name, $wikiId) {
+		$this->load($userId)->deleteLocalPreference($name, $wikiId);
```
drsnyder: It's a little bit confusing seeing `deleteLocalPreference` dispatch to a method of the same name.

lib/Wikia/src/Domain/User/Preferences/UserPreferences.php
nmonterroso, line 83: 1aefd250

lib/Wikia/src/Service/User/Preferences/PreferenceService.php
nmonterroso: all the `get`/`set`/`delete` functions do that eventually, and it's because externally everything goes through the service to talk to a user's `UserPreferences` object

lib/Wikia/src/Service/User/Preferences/PreferenceModule.php
drsnyder, line 14: Ok, is this purely defensive-- do you expect that we might have to invalidate it?

nmonterroso, line 14: Hmm.. I hope not, but I think the version of [`MW_USER_VERSION`](https://github.com/Wikia/app/blob/dev/includes/User.php#L40) was `1` when we started, so probably

drsnyder:
LGTM. :sheep: it

PR #8467, opened 2015-09-14
VE-1923: Put Wikia's stuff in wikia subdirectory and reduce amount of changes to core code

inez:


extensions/VisualEditor/wikia/extension.json
```diff
@@ -0,0 +1,38 @@
+{
+	"ResourceModules": {
+		"ext.visualEditor.wikia.desktopArticleTarget.init": {
+			"class": "VisualEditorDesktopArticleTargetInitModule",
+			"scripts": "ve/init/targets/ve.init.wikia.DesktopArticleTarget.init.js",
+			"styles": "ve/init/styles/ve.init.wikia.DesktopArticleTarget.init.css",
```
pauloslund: Is it an issue that this css file doesn't exist? I imagine not since `addModuleStyles` is never called on this module

PR #8468, opened 2015-09-14
Preference scope

nmonterroso:
@Wikia/services-team 
small mw service to test if a preference is local or global. I'll add some tests later, but since @ArturKlajnerok and I both need this functionality for our current tasks, it's probably best we just merge this now


nmonterroso:
Bleh. Opened to the wrong branch

PR #8469, opened 2015-09-14
Preference Scope

nmonterroso:
@Wikia/services-team 
small mw service to test if a preference is local or global. I'll add some tests later, but since @ArturKlajnerok and I both need this functionality for our current tasks, it's probably best we just merge this now

Here is the output from a sample run with the current whitelist in MW:

``` php
        $scope = Injector::getInjector()->get(PreferenceScopeService::class);
        $a = $scope->isGlobalPreference('foo'); // false
        $b = $scope->isGlobalPreference('language'); // true
        $c = $scope->isGlobalPreference('oasis-toolbar-contents'); // true
        $d = $scope->isLocalPreference('bar'); // false
        $e = $scope->isLocalPreference('adoptionmails-12345'); // true
        list($name, $wikiId) = $scope->splitLocalPreference('adoptionmails-12345'); // ['adoptionmails', 12345]
```


ArturKlajnerok:
Lgtm :+1:

PR #8475, opened 2015-09-15
ADEN-2435 Ads on event on mercury

fraszczakszymon:


extensions/wikia/AdEngine/js/slot/scrollHandler.js
```diff
@@ -1,67 +1,110 @@
...
+					reloadedViewMax: 3
+				}
+			},
+			mercury: {
+				MOBILE_PREFOOTER: {
+					reloadedViewMax: 1
+				}
+			}
+		},
+		skin = context.targeting.skin === 'oasis' ? 'oasis' : 'mercury';
```
nandy-andy, line 30: Do we really want to set `skin` to `'mercury'` each time `context.targeting.skin` isn't `'oasis'`? What does `context.targeting.skin` return for Mercury? Can't we just go with `skin = context.targeting.skin`? :japanese_ogre:

extensions/wikia/AdEngine/js/slot/scrollHandler.js
```diff
@@ -1,67 +1,110 @@
...
+		}));
+	}
 
-    function isReached(el) {
-        return win.innerHeight + win.scrollY >= getTopPos(el);
-    }
+	function refreshSlot(slotName) {
+		if (skin === 'oasis') {
+			win.adslots2.push([slotName]);
+		} else {
```
nandy-andy, line 64: What do you think about making it more restricted? 

```
if (skin === 'oasis') {
   // ...
} else if (skin === 'mercury') {
   // ...
} else {
   // log here an exception or even throw it?
}
```

extensions/wikia/AssetsManager/config.php
```diff
@@ -773,6 +773,7 @@
 		'//extensions/wikia/AdEngine/js/template/modal.js',
 		'//extensions/wikia/AdEngine/js/template/modalHandlerFactory.js',
 		'//extensions/wikia/AdEngine/js/template/modalMercuryHandler.js',
+		'//extensions/wikia/AdEngine/js/slot/scrollHandler.js',
```
nandy-andy: Shouldn't it be above `//extensions/wikia/AdEngine/js/template/`s?

extensions/wikia/AdEngine/js/slot/scrollHandler.js
fraszczakszymon, line 30: At this point `context.targeting.skin` gives me `undefined` for mercury :(

extensions/wikia/AdEngine/js/slot/scrollHandler.js
fraszczakszymon, line 64: Since skin is set to `oasis` or `mercury` and it's initialised only in `desktop.run.js` and `mercury.run.js` I think we don't need this else :P

extensions/wikia/AssetsManager/config.php
fraszczakszymon: ABCDEF...PRST...
sure :D

extensions/wikia/AdEngine/js/slot/scrollHandler.js
nandy-andy, line 30: Damn, let's just then leave a comment describing it. At one point we'll re-think and refactor ads in Mercury hopefully. And then we'll resolve this issue as well.

nandy-andy:
LGTM :+1:

PR #8478, opened 2015-09-15
OPS-6830: increase memcache client connection and read/write timeouts

macbre:
https://wikia-inc.atlassian.net/browse/OPS-6830 - give enough time to connect to McRouter on Poznań devbox

Backporting  a small piece of https://github.com/wikimedia/mediawiki/commit/76c3ab462ffa3bf7fd57a0e39413bdb12051a7db:
- Increase the default stream and connect timeouts from 100ms to 500ms

And improve error logging on connection problems

@wladekb / @adamw-wikia


ghost:
Can we define the timeout in one place as a class constant and do the conversion in some other place (perhaps in relevant `__construct`)? Currently it is prone to errors - Some People™ might not spot that the value must be changed in two places. You know, in the future.

macbre:
@michalroszka, these are actually two different timeouts: one is for the connection and the second one for reads & writes. The first one is in seconds and the second one in microseconds - yay for consistency :)

I simply wanted to keep it consistent with how it's defined in the MediaWiki upstream.

macbre:
Introduced `wgMemCachedConnectionTimeout` in c20e0c6 and improved the documentation of memcache timeout settings.

ghost:
:+1:

PR #8479, opened 2015-09-15
CE-2531 Restore certain revisions of MediaWiki JS files using auto-approving mechanism

adamkarminski:
Hey @Wikia/community-engineering !

This PR is focused on two major changes:
1. Providing a way for reviewers to auto-approve their revisions.
2. Creating a Special:ContentReview subpages mechanism with archives of reviews for wikias. A single archive can be found at `http://dev.wikia.com/Special:ContentReview/{wikiId}`.

It also introduces a new class - `ContentReviewService` which goal is to be the main controlling level of the ContentReview tool soon making the `ContentApiController` just a wrapper callable via http and available publicly.

Have fun!

/cc @kamilkoterba @Grunny

PR #8485, opened 2015-09-15
CE-2693 Prefill summary with a default message in a revision restoring process

adamkarminski:
**DO NOT MERGE BEFORE https://github.com/Wikia/app/pull/8479**

This PR is about prefilling the summary field when restoring a revision. The default message will include the ID number of a revision that is being restored.


kamilkoterba:
LGTM

PR #8487, opened 2015-09-15
log/correct preference mismatches

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-660
https://wikia-inc.atlassian.net/browse/SERVICES-736
https://wikia-inc.atlassian.net/browse/SERVICES-737

when calling `loadOptions` in the `User` class, compare `User->mOptions` to the output of `PreferenceService`, log any differences, and then correct the data in the `PreferenceService`. Also piggybacking in tests for `PreferenceScopeService`


lib/Wikia/src/Service/User/Preferences/Migration/PreferenceCorrectionService.php
```diff
@@ -0,0 +1,136 @@
...
+		$differences = 0;
+
+		foreach ( $options as $name => $value ) {
+			if ( $this->scopeService->isGlobalPreference( $name ) ) {
+				$expectedPreferences->setGlobalPreference( $name, $value );
+
+				if ( !$actualPreferences->hasGlobalPreference( $name ) ||
+					$actualPreferences->getGlobalPreference( $name ) !== $value ) {
+
+					$this->logMissingPreference( $userId, $name, $value, $actualPreferences->getGlobalPreference( $name ) );
```
ArturKlajnerok: This seems to log mismatch value or missing preference, should we split it to have more precise logs ?

ArturKlajnerok:
Other then one small clarification, LGTM :+1:

lib/Wikia/src/Service/User/Preferences/Migration/PreferenceCorrectionService.php
nmonterroso: That's a good idea. I'll do that.

nmonterroso: d681ea1

PR #8488, opened 2015-09-15
preference read/writes

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-761
Separate preference read/writing into two globals, since we'll always want to write using the new setup, but only read when we think the service is returning correct values.


sqreek:
If I understand it correctly, this introduce shadow mode now.
consider use $wgPreferenceServiceShadow(Read/Write) for more explicit name for that, 
except that LGTM

nmonterroso:
@sqreek the writes are shadow, but the reads are not (they're either using the service or mw). I like the idea of changing the write name to be more explicit. That was done in 35b6d6f

ArturKlajnerok:
Lgtm :+1:

PR #8494, opened 2015-09-16
Fix number of wikis with curated content

rogatty:
We were using incorrect query to get list of wikis with `wgWikiaCuratedContent` filled. Now we'll get wikis which have value of this variable other than an empty array.

/cc @michalroszka @bognix


bognix:
LGTM

ghost:
i LIKE it

PR #8495, opened 2015-09-16
SERVICES-766 remove warnings source in Include/User.php

sqreek:
remove source of warnings in Include/User.php
https://wikia-inc.atlassian.net/browse/SERVICES-766

/cc @harnash @macbre

### DIFF
--- a/includes/User.php
+++ b/includes/User.php
@@ -1419,7 +1419,8 @@ class User {
 		return array_reduce(
 			$defaultOptionNames,
 			function($preferences, $option) use ($wgUserPreferenceWhiteList, $defaultOptions) {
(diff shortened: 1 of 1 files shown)


includes/User.php
```diff
@@ -1419,7 +1419,8 @@ public static function getDefaultPreferences() {
 		return array_reduce(
 			$defaultOptionNames,
 			function($preferences, $option) use ($wgUserPreferenceWhiteList, $defaultOptions) {
-				if (in_array($option, $wgUserPreferenceWhiteList['literals'])) {
+				$literals = $wgUserPreferenceWhiteList[ 'literals' ];
```
macbre, line 1422: Where is `wgUserPreferenceWhiteList` defined? I can't find it in neither app nor config.

PR #8498, opened 2015-09-16
CE-2704 Hide the diff view for new JS pages

adamkarminski:
If you review a page that does not have an approved revision - no oldid param will be provided for a diff view.
It makes it really easy to miss any changes not introduced only by the latest revision since the diff is constructed
from it and the previous one. In this case it makes sense to hide the diff view.

/cc @Wikia/community-engineering @Grunny


Grunny:
Other than the comment about wrapping the message text mentioned in Slack, LGTM! :)

Grunny:
:+1:

PR #8501, opened 2015-09-16
VignetteUrlToUrlGenerator::reportError - add a backtrace to track the caller

macbre:
This will allow us to track the caller - https://wikia-inc.atlassian.net/browse/PLATFORM-1486

@wladekb / @michalroszka

### DIFF
--- a/includes/wikia/vignette/VignetteUrlToUrlGenerator.php
+++ b/includes/wikia/vignette/VignetteUrlToUrlGenerator.php
@@ -123,14 +123,17 @@ class VignetteUrlToUrlGenerator {
 		}
 	}
 
-	private function reportError($message, $context=[]) {
+	private function reportError($message, Array $context=[]) {
 		$exception = new InvalidArgumentException($message);
+		$context['exception'] = $exception;
+
 		$this->error($message, $context);
 		throw $exception;
 	}
 
 	protected function getLoggerContext() {
 		return [
+			'class' => __CLASS__,
 			'url' => $this->url,
 		];
 	}


includes/wikia/vignette/VignetteUrlToUrlGenerator.php
```diff
@@ -123,14 +123,17 @@ private function addThumbDefinition(UrlGenerator $generator) {
 		}
 	}
 
-	private function reportError($message, $context=[]) {
+	private function reportError($message, Array $context=[]) {
```
ghost, line 126: `protect` your `private` parts? :-)

In PHP, `private` is not really `private`, after all. Private properties may be accessed and private methods can be called from another instance of the same class. Despite of what many developers might think, private methods and properties are not private to a particular instance. Combined with mutable state (meh) and passing objects by reference by default (rather than cloning them) (meh) it is an insanely rich source of nasty bugs. So `private` is simply misleading. Using `protected` does not solve the problem at all, but for some reason just feels "more right", not that misleading.

``` php
<?php

class User {

    private $name;

    private function setName( $name ) {
        $this->name = $name;
    }

    public function quirk( User $user, $name ) {
        $user->setName( $name );
    }

    public function quirk2( User $user ) {
        return $user->name;
    }
}


$a = new User;
$b = new User;

$b->quirk( $a, 'Macbre' );
$a->quirk( $b, 'Mix' );

var_dump( $b->quirk2( $a ), $a->quirk2( $b ) );
/*
Will produce:
string(6) "Macbre"
string(3) "Mix"
*/

var_dump( $a->name, $b->name );
/*
Will produce:
PHP Fatal error:  Cannot access private property User::$name
*/

$a->setName('Władek');
$b->setName('Jarek');
/*
Will produce:
PHP Fatal error:  Call to private method User::setName() from context ''
*/

exit;
```

PR #8502, opened 2015-09-16
CE-2680 Create a unit test for updateReviewsStatus

adamkarminski:
/cc @Grunny

### DIFF
--- a/extensions/wikia/ContentReview/tests/ContentReviewApiControllerTest.php
+++ b/extensions/wikia/ContentReview/tests/ContentReviewApiControllerTest.php
@@ -15,9 +15,10 @@ class ContentReviewApiControllerTest extends WikiaBaseTest {
 	}
 
 	/**
+	 * @param array $params An array as defined in the dataProvider
 	 * @dataProvider submitPageForReviewProvider
 	 */
-	public function testSubmitPageForReview( $params, $textExpected, $message ) {
+	public function testSubmitPageForReview( $params ) {
 
 		$this->prepareControllerPropertiesMocks( $params );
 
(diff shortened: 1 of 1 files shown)


extensions/wikia/ContentReview/tests/ContentReviewApiControllerTest.php
```diff
@@ -33,6 +34,45 @@ public function testSubmitPageForReview( $params, $textExpected, $message ) {
...
+	 * @param bool $isAllowed
+	 * @param string $exception A name of a class of an expected exception
+	 * @dataProvider updateReviewsStatusDataProvider
+	 */
+	public function testUpdateReviewsStatus( $wasPosted, $matchEditToken, $isAllowed, $exception ) {
+		/* @var \WikiaRequest $requestMock */
+		$requestMock = $this->getMockBuilder( '\WikiaRequest' )
+			->disableOriginalConstructor()
+			->setMethods( [ 'wasPosted' ] )
+			->getMock();
```
Grunny: Unless you need to disable the constructor, this can be simplified to just: `$requestMock = $this->getMock( '\WikiaRequest', [ 'wasPosted' ] );`.

Grunny:

extensions/wikia/ContentReview/tests/ContentReviewApiControllerTest.php
adamkarminski: Yup, I could simplify it, but because the original `__constructor` requires the first parameter to be an `array` so the actual line is: `$requestMock = $this->getMock( '\WikiaRequest', [ 'wasPosted' ], [ [] ] );`

Thanks!

PR #8504, opened 2015-09-16
SERVICES-662 user preferences e2e [DO NOT MERGE]

harnash:
This integration tests check if User Preference Service is reachable from the Media Wiki side. It tries to get/set some user preferences on a test user account. Tests fails and the problem which addresses this is in separate issue (SERVICES-757).

Since we don't have this fixed now we should merge it after the fix is in place on dev at least. Please add comments and suggestion so they can be resolved ASAP.
@Wikia/services-team


nmonterroso:
should this PR be to the `user-preferneces-service-integration` branch? the changes from that branch aren't in `dev` yet, which include global/local preferences

nmonterroso:
I wonder if we should add a way to ignore the cache on the user preferences side, so that we know the integration with the service (like deserializing the responses) works properly, since that's what bit us last time.

harnash:
I will try to merge `user-preferneces-service-integration` into this branch and see what I can improve on. I think the tests should pass at least it will verify out theory for SERVICES-757.

harnash:
Closing since we need to diff against different branch.

PR #8508, opened 2015-09-16
[SERVICES-661] Save preferences when options are saved.

ArturKlajnerok:
Saving preferences each time `User::saveOptions()` is called, also removed unused `UserSetPreferences` hook that duplicated `UserSaveOptions` hook.

/cc @Wikia/services-team

### DIFF
--- a/includes/User.php
+++ b/includes/User.php
@@ -5007,6 +5007,8 @@ class User {
 
 		$dbw->upsert('user_properties', $insertRows, [], self::$PROPERTY_UPSERT_SET_BLOCK);
 
+		$this->userPreferences()->save($this->getId());
+
 		if ( $extuser ) {
 			$extuser->updateUser();
 		}
--- a/includes/User.php
+++ b/includes/User.php
@@ -3547,6 +3547,7 @@ class User {
 		);
 
 		$this->saveOptions();
+		$this->savePreferences();
 
 		wfRunHooks( 'UserSaveSettings', array( $this ) );
 		$this->clearSharedCache();
@@ -3670,6 +3671,7 @@ class User {
 		$this->clearInstanceCache();
 
 		$this->saveOptions();
+		$this->savePreferences();
 	}
 
 	/**
@@ -4945,6 +4947,11 @@ class User {
 		] );
 	}
 
+	protected function savePreferences() {
+		$this->userPreferences()->save($this->getId());
+		wfRunHooks( "UserSetPreferences", array( $this ) );
+	}
+
 	/**
 	 * @todo document
 	 */
@@ -5006,8 +5013,6 @@ class User {
 		}
 
 		$dbw->upsert('user_properties', $insertRows, [], self::$PROPERTY_UPSERT_SET_BLOCK);
-
-		$this->userPreferences()->save($this->getId());
 
 		if ( $extuser ) {
 			$extuser->updateUser();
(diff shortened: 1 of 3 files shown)


includes/User.php
```diff
@@ -5007,6 +5007,8 @@ protected function saveOptions() {
 
 		$dbw->upsert('user_properties', $insertRows, [], self::$PROPERTY_UPSERT_SET_BLOCK);
 
+		$this->userPreferences()->save($this->getId());
```
nmonterroso: Hmm I wonder if this should be done in `User->saveSettings()` right after the call to `$this->saveOptions()`? since it's really outside of options now

nmonterroso:
we don't need the hook?

ArturKlajnerok:
@nmonterroso The `UserSaveOptions` hook is called each time `User::saveOptions()` is invoked  and now when we save the preferences along with `saveOptions()` I don't see any reason to have duplicated `UserSetPreferences` hook.

The hook is used in two places:
https://github.com/Wikia/app/blob/48f8caf925b02009e9b20033eab437a7f972faf5/extensions/SemanticWatchlist/SemanticWatchlist.hooks.php#L189
https://github.com/Wikia/app/blob/4ba318893fee37e72112bc80ac689bd9eb70bb0a/extensions/wikia/Helios/User.class.php#L445

includes/User.php
ArturKlajnerok: We can move it there, just note that there is also `saveOptions()` in `addToDatabase()` so we would need to save settings in both places.

nmonterroso:
I think that goes back to "we shouldn't do this in `saveOptions`" since we should think of `options` as containing non-preferences. We should save our preferences separately from options in `saveSettings`, which functions as the save function for a user. 

If we do that then we probably need to implement the Helios hook, but not the other since it loops over the options and looks for 

``` php
if ( strpos( $name, 'swl_watchgroup_' ) === 0 && $value )
```

options. I just looked in the database and there are no hits in production:

```
mysql@10.8.64.63[wikicities]>select * from user_properties where up_property like 'swl_watchgroup_%' limit 10;
Empty set (0.00 sec)
```

So we probably don't have to keep that implementation, but we should keep the other one.

includes/User.php
nmonterroso: Yeah, I think we should do that to properly separate preferences from options. Maybe it should be a function `savePreferences` that saves the preferences in the db and also invokes the hook?

ArturKlajnerok: Changed that in : 87bb94d

ArturKlajnerok:
@nmonterroso I've applied the change but still wonder if we shouldn't clean up / simplify the hooks around saving settings, now we will have `UserSaveSettings`, `UserSaveOptions` and `UserSetPreferences` that basically do the same thing and are all invoked when `saveSettings` is called.

nmonterroso:
Yeah, that does suck. Actually since the user's preferences are no longer part of the user cache, and that's what Helios clears, does the hook need to be called anymore? I might have overlooked that part.

ArturKlajnerok:
@nmonterroso  It seems that `Helios\\User::onUserSave` don't need to be called for user preferences. So this should be the final change : 2fe6c01

nmonterroso:
LGTM will merge when tests pass

PR #8509, opened 2015-09-16
Preference Service Fixes

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-770

Miscellaneous things I had to do to get the preference service to work. Had to move the default preferences outside of the extension functions, since those functions could be executed after a call to `loadOptions`, which would create the default preferences and so we'd be missing some defaults. Also had to move default preferences in `PreferenceServiceImpl` to be a `UserPreferences` object rather than a map, since it contains locals as well.

### DIFF
--- a/extensions/wikia/AutomaticWikiAdoption/AutomaticWikiAdoption_setup.php
+++ b/extensions/wikia/AutomaticWikiAdoption/AutomaticWikiAdoption_setup.php
@@ -39,23 +39,22 @@ $wgAutoloadClasses['AutomaticWikiAdoptionController'] = "$dir/AutomaticWikiAdopt
 $wgAutoloadClasses['SpecialWikiAdoption'] = "$dir/SpecialWikiAdoption.class.php";
 $wgSpecialPages['WikiAdoption'] = 'SpecialWikiAdoption';
 
+# 194785 = ID of wiki created on 2010-12-14 so it will work for wikis created after this project has been deployed
+if ( $wgCityId > 194785 ) {
+	$wgDefaultUserOptions["adoptionmails-$wgCityId"] = 1;
+}
+
 /**
  * Initialize hooks
  *
  * @author Maciej Błaszkowski <marooned at wikia-inc.com>
  */
 function AutomaticWikiAdoptionInit() {
-	global $wgHooks, $wgDefaultUserOptions, $wgCityId;
-
-	# 194785 = ID of wiki created on 2010-12-14 so it will work for wikis created after this project has been deployed
-	if ( $wgCityId > 194785 ) {
-		$wgDefaultUserOptions["adoptionmails-$wgCityId"] = 1;
-	}
+	global $wgHooks;
 
 	$wgHooks['SkinTemplateOutputPageBeforeExec'][] = 'AutomaticWikiAdoptionHelper::onSkinTemplateOutputPageBeforeExec';
 	$wgHooks['GetPreferences'][] = 'AutomaticWikiAdoptionHelper::onGetPreferences';
 	$wgHooks['ArticleSaveComplete'][] = 'AutomaticWikiAdoptionHelper::onArticleSaveComplete';
-
 }
 
 // Ajax dispatcher
--- a/lib/Wikia/src/Domain/User/Preferences/LocalPreference.php
+++ b/lib/Wikia/src/Domain/User/Preferences/LocalPreference.php
@@ -14,6 +14,12 @@ class LocalPreference {
 		Assert::true( !empty( $name ), "invalid preference name" );
 		Assert::true( !empty( $wikiId ), "invalid wiki id" );
 
+		if ( $value === "true" ) {
+			$value = true;
+		} elseif ( $value === "false" ) {
+			$value = false;
+		}
+
 		$this->name = $name;
 		$this->value = $value;
 		$this->wikiId = $wikiId;
--- a/lib/Wikia/src/Service/User/Preferences/PreferenceModule.php
+++ b/lib/Wikia/src/Service/User/Preferences/PreferenceModule.php
@@ -2,12 +2,15 @@
 
 namespace Wikia\Service\User\Preferences;
 
+use Interop\Container\ContainerInterface;
 use User;
 use Wikia\Cache\BagOStuffCacheProvider;
 use Wikia\DependencyInjection\InjectorBuilder;
 use Wikia\DependencyInjection\Module;
+use Wikia\Domain\User\Preferences\UserPreferences;
 use Wikia\Persistence\User\Preferences\PreferencePersistence;
 use Wikia\Persistence\User\Preferences\PreferencePersistenceSwaggerService;
+use Wikia\Service\User\Preferences\Migration\PreferenceScopeService;
 
 class PreferenceModule implements Module {
 	const PREFERENCE_CACHE_VERSION = 1;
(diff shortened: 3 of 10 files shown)


extensions/wikia/AutomaticWikiAdoption/AutomaticWikiAdoption_setup.php
```diff
@@ -39,23 +39,22 @@
 $wgAutoloadClasses['SpecialWikiAdoption'] = "$dir/SpecialWikiAdoption.class.php";
 $wgSpecialPages['WikiAdoption'] = 'SpecialWikiAdoption';
 
+# 194785 = ID of wiki created on 2010-12-14 so it will work for wikis created after this project has been deployed
+if ( $wgCityId > 194785 ) {
+	$wgDefaultUserOptions["adoptionmails-$wgCityId"] = 1;
+}
```
drsnyder, line 45: Why 2010-12-14?

drsnyder, line 45: Oh, I see you moved this.

lib/Wikia/src/Domain/User/Preferences/LocalPreference.php
```diff
@@ -14,6 +14,12 @@ function __construct( $name, $value, $wikiId ) {
 		Assert::true( !empty( $name ), "invalid preference name" );
 		Assert::true( !empty( $wikiId ), "invalid wiki id" );
 
+		if ( $value === "true" ) {
+			$value = true;
+		} elseif ( $value === "false" ) {
+			$value = false;
+		}
```
drsnyder, line 21: I think `$value = filter_var($value, FILTER_VALIDATE_BOOLEAN);` might be better.

lib/Wikia/src/Service/User/Preferences/PreferenceModule.php
```diff
@@ -27,8 +30,22 @@ public function configure( InjectorBuilder $builder ) {
 				global $wgHiddenPrefs;
 				return $wgHiddenPrefs;
 			} )
-			->bind( PreferenceServiceImpl::DEFAULT_PREFERENCES )->to( function() {
-				return User::getDefaultPreferences();
+			->bind( PreferenceServiceImpl::DEFAULT_PREFERENCES )->to( function( ContainerInterface $c ) {
+				/** @var PreferenceScopeService $scopeService */
+				$scopeService = $c->get( PreferenceScopeService::class );
```
drsnyder, line 35: Will there be a problem with the lookup of `PreferenceScopeService::class`? Does this create a load order that, if not correct, will cause a failure?

lib/Wikia/src/Service/User/Preferences/PreferenceModule.php
```diff
@@ -27,8 +30,22 @@ public function configure( InjectorBuilder $builder ) {
...
+				foreach ( $defaultOptions as $name => $val ) {
+					if ( $scopeService->isGlobalPreference( $name ) ) {
+						$defaultPreferences->setGlobalPreference( $name, $val );
+					} elseif ( $scopeService->isLocalPreference( $name ) ) {
+						list( $prefName, $wikiId ) = $scopeService->splitLocalPreference( $name );
+						$defaultPreferences->setLocalPreference( $prefName, $wikiId, $val );
+					}
+				}
+
+				return $defaultPreferences;
```
drsnyder, line 48: It doesn't seem like this should be happening here. This (the building of the list) isn't directly related to DI.

lib/Wikia/src/Service/User/Preferences/PreferenceServiceImpl.php
```diff
@@ -235,23 +235,30 @@ private function deleteFromCache( $userId ) {
...
+				$preferences->setGlobalPreference( $globalPreference->getName(), $globalPreference->getValue() );
+			}
+		}
+
+		foreach ( $this->defaultPreferences->getLocalPreferences() as $wikiId => $localPreferences ) {
+			foreach ( $localPreferences as $localPreference ) {
+				/** @var LocalPreference $localPreference */
+				if ( !$preferences->hasLocalPreference( $localPreference->getName(), $wikiId ) ) {
+					$preferences->setLocalPreference( $localPreference->getName(), $wikiId, $localPreference->getValue() );
+				}
```
drsnyder, line 249: Is this block also duplicated in the `PreferenceModule`?

lib/Wikia/src/Service/User/Preferences/PreferenceModule.php
nmonterroso, line 35: No. I think it would only create a problem if there's a cyclic dependency detected when trying to resolve `PreferenceScopeService`, but the order of stuff doesn't matter.

lib/Wikia/src/Service/User/Preferences/PreferenceModule.php
nmonterroso, line 48: I'd say it is, since nothing else in the codebase needs this `UserPreferences` object. It's only needed to instantiate the `PreferenceService`. I'd say this is pretty much equivalent to something like this using Guice:

``` java
class MyClass {
  @Inject
  public MyClass(OtherClass otherClass) {
    ..
  }
}
...
class MyModule extends AbstractModule {
  @Provides
  public OtherClass providesOtherClass(MyConfiguration config) {
    // build OtherClass from stuff inside MyConfiguration
  }
}
```

lib/Wikia/src/Service/User/Preferences/PreferenceServiceImpl.php
nmonterroso, line 249: No. The `PreferenceModule` is iterating over a map of options (`name => val`) and creating a new `UserPreferences` object, this is iterating over a `UserPreferences` object and filling in gaps in the provided `UserPreferences` object

drsnyder:
Minor comments, otw LGTM.

lib/Wikia/src/Domain/User/Preferences/LocalPreference.php
nmonterroso, line 21: I think that would turn everything into a boolean, right? But that's not what we want.

```
[14:20] ~/Downloads> cat test.php
<?php
var_dump(filter_var("true", FILTER_VALIDATE_BOOLEAN));
var_dump(filter_var("false", FILTER_VALIDATE_BOOLEAN));
var_dump(filter_var("1", FILTER_VALIDATE_BOOLEAN));
var_dump(filter_var("0", FILTER_VALIDATE_BOOLEAN));
var_dump(filter_var(1, FILTER_VALIDATE_BOOLEAN));
var_dump(filter_var(0, FILTER_VALIDATE_BOOLEAN));
var_dump(filter_var("something-else", FILTER_VALIDATE_BOOLEAN));
var_dump(filter_var("42", FILTER_VALIDATE_BOOLEAN));
var_dump(filter_var(42, FILTER_VALIDATE_BOOLEAN));
[14:20] ~/Downloads> php test.php
bool(true)
bool(false)
bool(true)
bool(false)
bool(true)
bool(false)
bool(false)
bool(false)
bool(false)
```

drsnyder, line 21: Sorry, I don't mean for everything, but perhaps in the case where it's a string that looks bool-like. Just a suggestion...

PR #8511, opened 2015-09-17
SOC-1315 When clearing user cache for attribute service, clear staging envs as well

jsutterfield:
When a client other than MediaWiki writes an attribute to the attribute service, the attribute service sends a request to MediaWiki to clear the user cache for that user.

The problem was clearing this user cache only cleared for one of 2 environments, production or dev. This meant if an attribute was updated in preview or verify, the cache wouldn't be cleared and the old value would continue to show. This ticket makes sure we clear the cache in all staging environments when an attribute is updated outside of dev.

### DIFF
--- a/extensions/wikia/AttributeServiceMWApi/AttributeServiceMWApiController.class.php
+++ b/extensions/wikia/AttributeServiceMWApi/AttributeServiceMWApiController.class.php
@@ -105,8 +105,8 @@ class AttributeServiceMWApiController extends WikiaController {
     private function clearCacheInStagingEnvs( $userId ) {
         global $wgStagingList, $wgPreviewHostname, $wgVerifyHostname;
 
-        // $wgStagingList uses the domain name "preview" and "verify" rather than the hostnames "staging-s2"
-        // and "staging-s3" for those 2 envs, so we have to manually add the hostnames for those here.
+        // $wgStagingList uses the domain names "preview" and "verify" rather than the hostnames "staging-s2" and
+        // "staging-s3", so we have to manually add the those hostnames here.
         $stagingEnvs = array_merge( $wgStagingList, [ $wgPreviewHostname, $wgVerifyHostname ] );
         foreach ( $stagingEnvs as $stagingEnv ) {
             $this->clearCacheInEnv( $userId, $stagingEnv );
--- a/extensions/wikia/AttributeServiceMWApi/AttributeServiceMWApiController.class.php
+++ b/extensions/wikia/AttributeServiceMWApi/AttributeServiceMWApiController.class.php
@@ -5,9 +5,9 @@ use \Wikia\Util\GlobalStateWrapper;
 class AttributeServiceMWApiController extends WikiaController {
 
 	/**
-	 * Helper function for the attribute service to clear user MediaWiki.
-	 * This is called when a client other than MW updates an attribute
-	 * using the attribute service to clear the User cache.
+	 * Helper function for the attribute service to clear user cache in MediaWiki.
+	 * This is called by the attribute service when a client other than MW updates
+     * an attribute in the service.
 	 */
 	public function purgeUserCache() {
 
--- a/extensions/wikia/AttributeServiceMWApi/AttributeServiceMWApiController.class.php
+++ b/extensions/wikia/AttributeServiceMWApi/AttributeServiceMWApiController.class.php
@@ -7,7 +7,7 @@ class AttributeServiceMWApiController extends WikiaController {
 	/**
 	 * Helper function for the attribute service to clear user cache in MediaWiki.
 	 * This is called by the attribute service when a client other than MW updates
-     * an attribute in the service.
+	 * an attribute in the service.
 	 */
 	public function purgeUserCache() {
 
@@ -25,10 +25,10 @@ class AttributeServiceMWApiController extends WikiaController {
 			return;
 		}
 
-        $this->clearUserCache( $userId );
-        if ( !$this->inDevEnvironment() ) {
-            $this->clearCacheInStagingEnvs( $userId );
-        }
+		$this->clearUserCache( $userId );
+		if ( !$this->inDevEnvironment() ) {
+			$this->clearCacheInStagingEnvs( $userId );
+		}
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


includes/User.php
```diff
@@ -2123,8 +2123,6 @@ private function clearSharedCache() {
 		if( $this->mId ) {
 			global $wgMemc, $wgSharedDB; # Wikia
 			$wgMemc->delete( $this->getCacheKey() );
-			// Wikia: and save updated user data in the cache to avoid memcache miss and DB query
-			$this->saveToCache();
```
jsutterfield, line 2127: This is a weird and unnecessary side effect. The function is called "clearSharedCache", not "clearSharedCache-thenResaveSharedCache".

It also has the annoying consequence of sometimes saving stale data when called from `User::invalidateCache`.  The way this happens is first we create a user object so we can clear its cache:

``` php
$myUser = User::newFromId( 1 );
```

As that user object is being created, it tries to pull data for itself out of cache [here](https://github.com/Wikia/app/blob/dev/includes/User.php#L356). Chances are it succeeds and now we have a user object initialized from cache.

We then call `$myUser->invalidateCache()`. That method creates a new value for the member variable `mTouched`  [here](https://github.com/Wikia/app/blob/dev/includes/User.php#L2153), which is used to indicate the last time this user was updated, and then saves that value to the db. So far so good.

The problem is `User::invalidateCache()` would call `User::clearSharedCache` which would call `User::saveToCache` which ends up saving the data we pulled from cache originally, but with the new value for mTouched! So now we've saved old data, with the latest mTouched value and stale cache is served.

Let's just get rid this.

includes/User.php
```diff
@@ -2169,6 +2167,8 @@ public function invalidateCache() {
 					__METHOD__ );
 			}
 
+			$dbw->commit();
```
jsutterfield: The `selectField` statement above ends up starting a `BEGIN` statement on this connection. We need to explicitly `COMMIT` it here or the update above won't persist.

extensions/wikia/Staging/Staging.setup.php
```diff
@@ -29,5 +29,6 @@
  * @author macbre
  * @see PLATFORM-664
  */
+$wgBaseShareKeyPrefix = $wgSharedKeyPrefix;
 $wgCachePrefix = gethostname() . '-' . wfWikiID(); // e.g. staging-s3-muppet / sandbox-qa02-glee / ...
-$wgSharedKeyPrefix = gethostname() . '-' . $wgSharedKeyPrefix; // e.g. staging-s3-wikicities
+$wgSharedKeyPrefix = Wikia::getCurrentServerSharedKeyPrefix();
```
jsutterfield: `wgSharedKeyPrefix` used to overwrite itself here using it's original value as part of the key.

This made updating it for different environments difficult so I ended up saving it's original value in `$wgBaseShareKeyPrefix` and using a couple helper methods in `Wikia.php` to construct the key.

extensions/wikia/AttributeServiceMWApi/AttributeServiceMWApiController.class.php
```diff
@@ -82,5 +80,44 @@ private function setErrorResponse( $message, $responseCode ) {
...
+            $this->clearCacheInEnv( $userId, $stagingEnv );
+        }
+    }
+
+    private function clearCacheInEnv( $userId, $stagingEnv ) {
+        $wrapper = new GlobalStateWrapper( [ 'wgSharedKeyPrefix' => Wikia::getSharedKeyPrefix( $stagingEnv ) ] );
+        $wrapper->wrap( function() use ( $userId ) {
+            $this->clearUserCache( $userId );
+        } );
+    }
```
macbre: Please use tabs instead of spaces

includes/User.php
kvas-damian: Shouldn't this `commit` be put in the `if` statement?

macbre: I assume that we want to make user preferences / attributes / whatever service be the **only** source of users data. Shouldn't we instead of this fix implement a method in a service that will return "user touched" from services point of view - this will be used to invalidate MW cache (I believe we still need it).

kvas-damian:
LGTM as a short term fix.

For future we need to find better way of invalidating cache on Service clients side. Current solution implements circular dependency (User Atrribute service <--> MW). In future it could happen that other services would need to be notified when invalidate internal cache. It's quite bit architectural change which is out of scope of this ticket.

@cmiller-wikia I think it's task for you ;)

Quickly I can find couple possible solutions for this problem:
- don't cache anything on client side - if required implement cache at service level (varnish could be enough here)
- create journal that lists all changes, client subscribes journal and clears cache when followed user was changed.
  But definitely we should consider other valid options.

rafalkalinski:
after addressing others comments, LGTM

extensions/wikia/AttributeServiceMWApi/AttributeServiceMWApiController.class.php
jsutterfield: Oops, just got a new laptop and thought I had it configured properly. Apparently not...

includes/User.php
jsutterfield: @macbre that's a good point. The problem for now is that attributes are being stored in the same `mOptions` array as preferences which uses a different service. So even if we checked the attribute service to see when the last "user touched" value was, we'd have to do the same for the preference service, and even then we still have user "flags" which are still stored in MW which have yet to be split out into their own service.

An approach to address this would probably be to split this user data into it's components in the user object (eg, have `attributes`, `preferences`, and `flags` in their own arrays). That way we could reference the corresponding service when checking if cache is stale.

Part of the ground work for that has been done with the creation of the `User::get(Attribute|Preference|Flag)` and `User::set(Attribute|Preference|Flag` methods, but all user data still goes through the `mOptions` variable. Once we split that apart more, we can (and probably should) use the service as the source of truth for last updated.

jsutterfield:
@kvas-damian, I agree, this should be viewed as a temporary measure. Something more long term and robust should be figured out going forward. I like the sound of your first suggestion. Keeping the cache in one place on the service side makes it simple, and if we could get away with just using varnish then all the better. Open to other options as well though :)

drsnyder:
> When a client other than MediaWiki writes an attribute to the attribute service, the attribute service sends a request to MediaWiki to clear the user cache for that user.

Say more about this. Is this a work around until we are able to properly determine if the values in the MW cache are stale?

jsutterfield:
@drsnyder see [this comment](https://wikia-inc.atlassian.net/browse/SOC-1315?focusedCommentId=196926&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-196926). It was decided this was a pretty brittle approach, and we were seeing problems after we deployed it, so I reverted this code and we're going to approach this differently. The plan going forward is to decouple the attribute service and MW more, ie, the attribute service shouldn't be aware of, or care about MW's cache.

There's currently only 2 attributes that are going to be modified by clients other than MW, "avatar" and "location". The short term plan is to have MW talk to the attribute service every time it needs one of these values in case another client updated that value, rather than checking its own user cache. We're going to add some degree of caching for these values (maybe a minute or so), but for the most part MW will be checking the attribute service when it needs them.

We still need to work out a more long term solution, one that encapsulates all attributes rather than just these that can be modified outside of MW. It may involve caching on the service side, implementing e-tags/conditional gets, etc. Whatever the solution is, I imagine it'll involve continuing to extract these attributes out of the `User::mOptions` array and the caching it's been doing.

PR #8513, opened 2015-09-17
XW-199 | Polldaddy tag

vforge:
https://wikia-inc.atlassian.net/browse/XW-199
https://github.com/Wikia/config/pull/1321


extensions/wikia/PolldaddyTag/PolldaddyTagValidator.class.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+class PolldaddyTagValidator {
+
+	public static function validateAttributes( $params ) {
+		return array_key_exists( 'id', $params ) && !empty( $params['id'] );
```
Warkot: I could agree that outside the `empty()` function, checking `$params['id']` while `id` key doesn't exist, would throw `E_NOTICE : type 8 -- Undefined index: id -- at line 6`. However, it is safe to do such check using `empty()`. You can drop ~~the base~~ `array_key_exists` check.

extensions/wikia/PolldaddyTag/PolldaddyTagController.class.php
```diff
@@ -0,0 +1,50 @@
...
+		}
+
+		return $this->helper->isMobileSkin() ?
+			Html::rawElement(
+				'a',
+				array_merge( self::DATA_WIKI_WIDGET_ATTRIBUTE, [ 'data-id' => $args['id'] ] )
+			) :
+			Html::rawElement(
+				'span',
+				self::DATA_WIKI_WIDGET_ATTRIBUTE,
```
Warkot, line 46: We need to have a conversation (including @justnpT) about this `data-wikia-widget` attribute and where we're adding it and do some code cleanup. Why? It differs along the widgets and we like consistency — especially when writing selenium-tests. :coffee: 
- Simple inline framed widgets have the `data-` attribute added to the `<iframe>` — simple indeed.
- VK has it added to the `<div>` inside the `<span>` wrapper; This wrapper has a `widget-vk` class.
- Twitter has it added the `<a>` inside the `<span>` wrapper; This wrapper has a `widget-twitter` class, but the aforementioned `<a>` tag is swapped by the Twitter script for an `<iframe>`. We don't control this `<iframe>`.
- Here you add it to the wrapper… and IMHO it's the best way. :smiley: 

Also I'm not sure about adding the wrapper to the VK widget, as we control the `<div>` which the VK script looks for and adds the `<iframe>` into it — we naturally have a wrapper.

vforge, line 46: @Warkot I've added it to the wrapper, because I had no choice ;)

bognix:
LGTM

extensions/wikia/PolldaddyTag/PolldaddyTagValidator.class.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+class PolldaddyTagValidator {
+
+	public static function validateAttributes( $params ) {
+		return array_key_exists( 'id', $params ) && ctype_digit( $params['id'] );
```
Warkot, line 6: :+1:

extensions/wikia/PolldaddyTag/PolldaddyTagController.class.php
Warkot, line 46: That's cool, but do we have a ticket for refactoring/polishing/unifying all of the widgets? I would add this comment there.

Warkot:
LGTM

PR #8514, opened 2015-09-17
CE-2531 Purge assets cache after auto-approval of a revision

adamkarminski:
/cc: @Grunny

### DIFF
--- a/extensions/wikia/ContentReview/services/ContentReviewService.class.php
+++ b/extensions/wikia/ContentReview/services/ContentReviewService.class.php
@@ -10,7 +10,8 @@ namespace Wikia\ContentReview;
 
 use Wikia\ContentReview\Models\ReviewModel,
 	Wikia\ContentReview\Models\ReviewLogModel,
-	Wikia\ContentReview\Models\CurrentRevisionModel;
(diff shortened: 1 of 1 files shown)


extensions/wikia/ContentReview/services/ContentReviewService.class.php
```diff
@@ -10,7 +10,8 @@
 
 use Wikia\ContentReview\Models\ReviewModel,
 	Wikia\ContentReview\Models\ReviewLogModel,
-	Wikia\ContentReview\Models\CurrentRevisionModel;
+	Wikia\ContentReview\Models\CurrentRevisionModel,
+	Wikia\ContentReview\Helper;
```
Grunny: This shouldn't be needed for `Helper` as this service is also in the `Wikia\ContentReview` namespace.

Grunny:
Minor comment, otherwise LGTM. :)

extensions/wikia/ContentReview/services/ContentReviewService.class.php
adamkarminski: D'oh! True, a habit :) Thanks!

PR #8520, opened 2015-09-17
CE-2718 Use oldid to preview specific revisions of pages on Special:ContentReview

adamkarminski:
/cc @lgarczewski


lgarczewski:
LGTM.

PR #8522, opened 2015-09-17
fix equality checks

nmonterroso:
@Wikia/services-team 
changing the equality check in the preference correction service to be non-strict, since in mediawiki `false`-like values get translated to empty strings when saving to the db, but from the preference service `false` is returned. Also changing the `null` check when setting preferences to be strict, because otherwise we couldn't set preferences to `false`-like values.


drsnyder:
:hankey:

:shipit:

PR #8523, opened 2015-09-17
SEO-66 Fix SEO tweak for broken links and re-enable it

gabrys:
Reverts #8369 which might be the cause for the traffic drop.

The code is now enabled only for the content namespaces, which should eliminate the database calls for Thread pages.

Single queries are rather fast (roughly 0.1-0.3 seconds) so it seems it's the number of queries that made the database slow.

With this fix the number of queries should be reasonably small and traffic regained.


owend:
LGTM :game_die:

PR #8524, opened 2015-09-17
SEO-66 Fix SEO tweak for broken links and re-enable it

gabrys:
Reverts #8369 which might be the cause for the traffic drop.

The code is now enabled only for the content namespaces, which should eliminate the database calls for Thread pages.

Single queries are rather fast (roughly 0.1-0.3 seconds) so it seems it's the number of queries that made the database slow.

With this fix the number of queries should be reasonably small and traffic regained.

### DIFF
--- a/extensions/wikia/SEOTweaks/SEOTweaks.setup.php
+++ b/extensions/wikia/SEOTweaks/SEOTweaks.setup.php
@@ -17,7 +17,7 @@ $wgHooks['BeforePageDisplay'][] = 'SEOTweaksHooksHelper::onBeforePageDisplay';
 $wgHooks['AfterInitialize'][] = 'SEOTweaksHooksHelper::onAfterInitialize';
 $wgHooks['ImagePageAfterImageLinks'][] = 'SEOTweaksHooksHelper::onImagePageAfterImageLinks';
 $wgHooks['BeforeParserMakeImageLinkObjOptions'][] = 'SEOTweaksHooksHelper::onBeforeParserMakeImageLinkObjOptions';
-#$wgHooks['ArticleViewHeader'][] = 'SEOTweaksHooksHelper::onArticleViewHeader'; // macbre: disabled because of MON-416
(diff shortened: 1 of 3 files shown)


macbre:
@gabrys, unit tests for this feature are broken after your changes.

macbre:
```
1) SEOTweaksTest::testOnArticleViewHeaderTitleNotExistsNoResults
Expectation failed for method name is equal to <string:query> when invoked at sequence index 2
Parameter 0 for invocation DatabaseBase::query('SELECT page_title FROM page W...IMIT 1', 'SEOTweaksHooksHelper::onArtic...Header', false) does not match expected value.
--- Expected
+++ Actual
@@ @@
-'SELECT page_title FROM page WHERE page_title LIKE 'Wanted%' LIMIT 1'
+'SELECT page_title FROM page WHERE page_title LIKE 'Wanted%' AND page_namespace = 0 LIMIT 1'

/var/lib/jenkins/workspace/unitTests_php_branches/extensions/wikia/SEOTweaks/SEOTweaksHooksHelper.class.php:145
/var/lib/jenkins/workspace/unitTests_php_branches/extensions/wikia/SEOTweaks/tests/SEOTweaksTest.php:713
/var/lib/jenkins/workspace/unitTests_php_branches/tests/run-test.php:27

2) SEOTweaksTest::testOnArticleViewHeaderTitleNotExistsWithResults
Expectation failed for method name is equal to <string:query> when invoked at sequence index 2
Parameter 0 for invocation DatabaseBase::query('SELECT page_title FROM page W...IMIT 1', 'SEOTweaksHooksHelper::onArtic...Header', false) does not match expected value.
--- Expected
+++ Actual
@@ @@
-'SELECT page_title FROM page WHERE page_title LIKE 'Wanted%' LIMIT 1'
+'SELECT page_title FROM page WHERE page_title LIKE 'Wanted%' AND page_namespace = 0 LIMIT 1'
```

macbre:
Scratch that, tests do pass

PR #8525, opened 2015-09-18
SO-1315-2 Set $wgBaseShareKeyPrefix in CommonSettings and fix Attribute Service Logging

jsutterfield:

### DIFF
--- a/extensions/wikia/Staging/Staging.setup.php
+++ b/extensions/wikia/Staging/Staging.setup.php
@@ -29,6 +29,5 @@ $wgHooks['BeforePageRedirect'][] = 'StagingHooks::onBeforePageRedirect';
  * @author macbre
  * @see PLATFORM-664
  */
-$wgBaseShareKeyPrefix = $wgSharedKeyPrefix;
 $wgCachePrefix = gethostname() . '-' . wfWikiID(); // e.g. staging-s3-muppet / sandbox-qa02-glee / ...
 $wgSharedKeyPrefix = Wikia::getCurrentServerSharedKeyPrefix(); // e.g. staging-s3-wikicities
--- a/includes/User.php
+++ b/includes/User.php
@@ -4903,19 +4903,15 @@ class User {
 	}
 
 	private function compareAttributeValueFromService( $attributeName, $attributeValue ) {
-		if ( !array_key_exists( $attributeName, $this->mOptions ) ) {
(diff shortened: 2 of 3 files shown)


extensions/wikia/Staging/Staging.setup.php
```diff
@@ -29,6 +29,5 @@
  * @author macbre
  * @see PLATFORM-664
  */
-$wgBaseShareKeyPrefix = $wgSharedKeyPrefix;
```
jsutterfield, line 32: Because this extension is [only loaded on staging environments](https://github.com/Wikia/config/blob/dev/CommonExtensions.php#L3019-L3021), `$wgBaseShareKeyPrefix` wasn't getting set in production which is where we need it. Moving this assignment to CommonSettings.php

includes/User.php
```diff
@@ -4903,19 +4903,15 @@ private function loadAttributes() {
 	}
 
 	private function compareAttributeValueFromService( $attributeName, $attributeValue ) {
-		if ( !array_key_exists( $attributeName, $this->mOptions ) ) {
```
jsutterfield, line 4906: Now that we're iterating over `$wgPublicAttributes`, we don't need to check if any attributes are missing since we'll be comparing all possible values the attribute service could return. See: https://github.com/Wikia/app/commit/99c6426e8aa35640270752283d4a76d78ccafa2a.

includes/User.php
```diff
@@ -4903,19 +4903,15 @@ private function loadAttributes() {
...
-			$this->logAttributeMissing( $attributeName, $attributeValue );
-		} elseif ( $this->mOptions[$attributeName] !== $attributeValue  ) {
-			$this->logAttributeMismatch( $attributeName, $attributeValue );
+
+		// bio and coverPhoto are not used in MW
+		if ( $attributeName == "bio" || $attributeName == "coverPhoto" ) {
```
rafalkalinski, line 4908: If we expect more attributes like that, I would be good it they were placed in one structure in one place.

jsutterfield, line 4908: Good point. For now we don't expect more attributes like this, and this logging will probably go away in the near future. If we add more attributes to this `if` statement though, we should split them into one place :)

sebastianmarzjan, line 4908: Please consider moving string consts outside of the runtime logic, i.e. to the const definitions.

rafalkalinski:
LGTM

PR #8527, opened 2015-09-18
PLATFORM-1494: store the full URL of the predefined avatar and skip an upload via service

macbre:
When the avatar service is enabled **always** store the full avatar URL - even if it's one of the predefined ones

``` sql
mysql> select * from user_properties where up_user = 119245 and up_property = 'avatar';
+---------+-------------+-----------------------------------------------------------+
| up_user | up_property | up_value                                                  |
+---------+-------------+-----------------------------------------------------------+
|  119245 | avatar      | http://images.wikia.com/messaging/images/c/cf/Avatar6.jpg |
+---------+-------------+-----------------------------------------------------------+
```

See #8181 for more context

@jcellary / @jsutterfield

### DIFF
--- a/extensions/wikia/UserProfilePageV3/UserProfilePageController.class.php
+++ b/extensions/wikia/UserProfilePageV3/UserProfilePageController.class.php
@@ -526,9 +526,8 @@ class UserProfilePageController extends WikiaController {
 			if ( empty( $wgAvatarsUseService ) ) {
 				// TODO: $user->getTouched() get be used to invalidate avatar URLs instead
 				$user->setGlobalAttribute( 'avatar_rev', date( 'U' ) );
+				$user->saveSettings();
 			}
-
-			$user->saveSettings();
 		}
 
 		wfProfileOut( __METHOD__ );
--- a/extensions/wikia/UserProfilePageV3/UserProfilePageController.class.php
+++ b/extensions/wikia/UserProfilePageV3/UserProfilePageController.class.php
@@ -507,6 +507,7 @@ class UserProfilePageController extends WikiaController {
 					else {
 						// store the full URL of the predefined avatar and skip an upload via service (PLATFORM-1494)
 						$user->setGlobalAttribute( AVATAR_USER_OPTION_NAME, Masthead::getDefaultAvatarUrl( $data->file ) );
+						$user->saveSettings();
 					}
 					break;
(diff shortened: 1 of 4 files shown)


extensions/wikia/UserProfilePageV3/UserProfilePageController.class.php
```diff
@@ -504,6 +504,10 @@ private function saveUsersAvatar( $userId = null, $data = null ) {
 					if ( empty( $wgAvatarsUseService ) ) {
 						$user->setGlobalAttribute( AVATAR_USER_OPTION_NAME, $data->file );
 					}
+					else {
```
jsutterfield, line 507: Since Wikia's coding conventions are a superset of MediaWiki's coding conventions, we should follow the rule that [an else is placed on the same line as the previous closing brace](https://www.mediawiki.org/wiki/Manual:Coding_conventions#Indenting_and_alignment).

extensions/wikia/UserProfilePageV3/UserProfilePageController.class.php
```diff
@@ -522,8 +526,9 @@ private function saveUsersAvatar( $userId = null, $data = null ) {
 			if ( empty( $wgAvatarsUseService ) ) {
 				// TODO: $user->getTouched() get be used to invalidate avatar URLs instead
 				$user->setGlobalAttribute( 'avatar_rev', date( 'U' ) );
-				$user->saveSettings();
 			}
+
+			$user->saveSettings();
```
jsutterfield: This actually causes a problem when saving a custom avatar.

On L514 we save an avatar via the avatar service, which writes the new avatar value directly to the attribute service, which then makes a call to MW to clear the user cache. But at no point is that new value for the avatar added to this user object and its instance variables.

So when we call  `saveSettings()` here we end up overwriting that new avatar value with whatever old value this user object had.

macbre: Fixed in 4b9b338 (added the call to `User::saveSettings` when setting the full URL of the predefined avata in a23ad91)

macbre:
```
1) SwiftStorageTest::testStoreAndRemove
InvalidResponseException: Invalid response (401): Unexpected HTTP return code: 401
```

Failing tests are unrelated

PR #8528, opened 2015-09-18
ADEN-2447 OpenX Bidder ad unit

fraszczakszymon:
We need to implement a proper way of defining ad unit for OpenX bidder, due to this fact we decided to refactor AdLogicPageParams module - by extracting shared features to new module.


jacekjursza:
LGTM

PR #8529, opened 2015-09-18
ADEN-2456 Fix lookup services extending parameters method

fraszczakszymon:
Instead of checking if method exists, let's execute it :)


nandy-andy:
Yep, this is what we want to do: execute instead of check :japanese_ogre: 

LGTM :+1:

PR #8533, opened 2015-09-18
ADEN-2452 The simplest recoverable module

nandy-andy:
In order to have different variances of recovery we'd like a simple module which injects our product if we detect an ad blocker. Here is an experiment with "Hello world!" injected above Wikia right rail.

I also created a new recovery helper into which I moved methods from our gpt helper.

### DIFF
--- a/extensions/wikia/AdEngine/js/provider/gpt/helper.js
+++ b/extensions/wikia/AdEngine/js/provider/gpt/helper.js
@@ -44,7 +44,7 @@ define('ext.wikia.adEngine.provider.gpt.helper', [
 	}
 
 	function loadSourcePoint() {
-		if (recoveryHelper.isBlocking()) {
+		if (recoveryHelper.isBlocking() && recoveryHelper.isRecoveryEnabled()) {
 			loadRecovery();
 		} else {
 			doc.addEventListener('sp.blocking', function () {
--- a/extensions/wikia/AdEngine/js/provider/gpt/helper.js
+++ b/extensions/wikia/AdEngine/js/provider/gpt/helper.js
@@ -47,7 +47,7 @@ define('ext.wikia.adEngine.provider.gpt.helper', [
 		if (recoveryHelper.isBlocking()) {
 			loadRecovery();
 		} else {
-			recoveryHelper.addOnBlockingEvent(function () {
+			recoveryHelper.addOnBlockingCallback(function () {
 				loadRecovery();
 			});
 		}
--- a/extensions/wikia/AdEngine/js/provider/gpt/helper.js
+++ b/extensions/wikia/AdEngine/js/provider/gpt/helper.js
@@ -9,8 +9,8 @@ define('ext.wikia.adEngine.provider.gpt.helper', [
 	'ext.wikia.adEngine.provider.gpt.adDetect',
 	'ext.wikia.adEngine.provider.gpt.adElement',
 	'ext.wikia.adEngine.provider.gpt.googleTag',
+	'ext.wikia.adEngine.recovery.helper',
 	'ext.wikia.adEngine.slotTweaker',
-	require.optional('ext.wikia.adEngine.provider.gpt.sourcePointTag'),
 	require.optional('ext.wikia.adEngine.provider.gpt.sraHelper'),
 	require.optional('ext.wikia.adEngine.slot.scrollHandler')
 ], function (
@@ -22,40 +22,16 @@ define('ext.wikia.adEngine.provider.gpt.helper', [
 	adDetect,
 	AdElement,
 	GoogleTag,
+	recoveryHelper,
 	slotTweaker,
-	SourcePointTag,
 	sraHelper,
 	scrollHandler
 ) {
 	'use strict';
 
 	var logGroup = 'ext.wikia.adEngine.provider.gpt.helper',
-		context = adContext.getContext(),
 		googleApi = new GoogleTag(),
-		slotsToRecover = [],
 		sourcePointInitialized = false;
-
-	function isRecoveryEnabled() {
-		return !!(context.opts.sourcePoint && SourcePointTag);
-	}
-
-	function isBlocking() {
-		return !!(isRecoveryEnabled() && window.ads && window.ads.runtime.sp.blocking);
-	}
-
-	function isRecoverable(slotName, recoverableSlots) {
-		return isRecoveryEnabled() && recoverableSlots.indexOf(slotName) !== -1;
-	}
-
-	function recoverSlots() {
-		if (!isBlocking()) {
-			return;
-		}
-		log(['Starting recovery', slotsToRecover], 'debug', logGroup);
-		while (slotsToRecover.length){
-			window.ads.runtime.sp.slots.push([slotsToRecover.shift()]);
-		}
-	}
 
 	function loadRecovery() {
 		if (sourcePointInitialized) {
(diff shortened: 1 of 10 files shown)


extensions/wikia/AdEngine/js/provider/gpt/helper.js
```diff
@@ -110,7 +86,7 @@ define('ext.wikia.adEngine.provider.gpt.helper', [
 
 		function callSuccess(adInfo) {
 			if (adInfo && adInfo.adType === 'collapse') {
-				slotTweaker.hide(element.getSlotName(), isBlocking());
+				slotTweaker.hide(element.getSlotName(), recoveryHelper.isBlocking());
```
fraszczakszymon: `recoveryHelper.isBlocking() && recoveryHelper. isRecoveryEnabled()`, otherwise it will have an impact to users from `wgAdDriverSorucePointDetection` countries

extensions/wikia/AdEngine/js/provider/gpt/helper.js
```diff
@@ -162,8 +138,8 @@ define('ext.wikia.adEngine.provider.gpt.helper', [
 			return;
 		}
 
-		if (!isBlocking()) {
-			slotsToRecover.push(slotName);
+		if (!recoveryHelper.isBlocking()) {
```
fraszczakszymon: `!recoveryHelper.isBlocking() && recoveryHelper. isRecoveryEnabled()`, otherwise it will have an impact to users from `wgAdDriverSorucePointDetection` countries. In this case it's not so dangerous but still we don't need to add this slots when recovery is disabled.

extensions/wikia/AdEngine/js/recovery/helper.js
```diff
@@ -0,0 +1,56 @@
...
+	function isBlocking() {
+		return !!(window.ads && window.ads.runtime.sp.blocking);
+	}
+
+	function isRecoverable(slotName, recoverableSlots) {
+		return isRecoveryEnabled() && recoverableSlots.indexOf(slotName) !== -1;
+	}
+
+	function recoverSlots() {
+		if (!isBlocking()) {
```
fraszczakszymon: `!isBlocking() || !isRecoveryEnabled()`

extensions/wikia/AdEngine/js/provider/gpt/helper.js
```diff
@@ -94,7 +70,7 @@ define('ext.wikia.adEngine.provider.gpt.helper', [
 		var count,
 			element,
 			recoverableSlots = extra.recoverableSlots || [],
-			shouldPush = !isBlocking() || isRecoverable(slotName, recoverableSlots);
+			shouldPush = !recoveryHelper.isBlocking() || recoveryHelper.isRecoverable(slotName, recoverableSlots);
```
fraszczakszymon: `!recoveryHelper.isBlocking() || !recoveryHelper.isRecoveryEnabled() || recoveryHelper.isRecoverable(slotName, recoverableSlots);`

extensions/wikia/AdEngine/js/provider/gpt/helper.js
```diff
@@ -22,53 +22,29 @@ define('ext.wikia.adEngine.provider.gpt.helper', [
...
 		sourcePointInitialized = true;
-		googleApi = new SourcePointTag();
-		recoverSlots();
+		googleApi = recoveryHelper.createSourcePointTag();
+		recoveryHelper.recoverSlots();
 	}
 
 	function loadSourcePoint() {
-		if (isBlocking()) {
+		if (recoveryHelper.isBlocking() && recoveryHelper.isRecoveryEnabled()) {
```
fraszczakszymon: Maybe it's worth to add:

```
if (!recoveryHelper.isRecoveryEnabled()) {
  return;
}
```

above this if statement - to prevent adding listener on `sp.blocking` when recovery is disabled or we can just remove `isRecoveryEnabled()` since loadSourcePoint is executed only then (https://github.com/Wikia/app/pull/8533/files#diff-d0456a56fa8978685b777a3c10aaf5abR129).

extensions/wikia/AdEngine/js/run/desktop.run.js
```diff
@@ -89,7 +91,20 @@ require([
...
+
+		if (context.opts.recoveredAdsMessage) {
+			loader({
+				type: loader.AM_GROUPS,
+				resources: ['adengine2_ads_recovery_message_js']
+			}).done(function () {
+				require(['ext.wikia.adEngine.recovery.message'], function (recoveredAdMessage) {
+					recoveredAdMessage.init();
+				});
+			});
```
fraszczakszymon, line 108: If I'm not wrong, here will be made a new, separate request for `ext.wikia.adEngine.recovery.message` and probably it will be much easier to block this single request (with adblock rules) and break module.

nandy-andy, line 108: I don't think you're wrong. However, I'd go with this solution until we launch it globally then it'd make sense to load it for every page view, I guess.

extensions/wikia/AdEngine/js/provider/gpt/helper.js
nandy-andy: Funny thing when I did it the unit tests started to fail :japanese_ogre:

extensions/wikia/AdEngine/js/run/desktop.run.js
fraszczakszymon, line 108: :+1:

fraszczakszymon:
I'm not able to comment untouched code lines? Here are not changed tests:
[#1 Should push/flush when sourcepoint is enabled but pageview is not blocked](https://github.com/Wikia/app/blob/aden-2452-the-simplest-recoverable-module/extensions/wikia/AdEngine/js/spec/provider/gpt/helper.spec.js#L165)
[#2 Should push/flush when pageview is blocked but sourcepoint is disabled](https://github.com/Wikia/app/blob/aden-2452-the-simplest-recoverable-module/extensions/wikia/AdEngine/js/spec/provider/gpt/helper.spec.js#L188)
Maybe we should consider removing them if they didn't fail with those changes?

extensions/wikia/AdEngine/js/recovery/helper.js
```diff
@@ -0,0 +1,75 @@
...
+			});
+		});
+		onBlockingEventsQueue.start();
+	}
+
+	function addSlotToRecover(slotName) {
+		slotsToRecover.push(slotName);
+	}
+
+	function addOnBlockingEvent(eventHandler) {
```
fraszczakszymon: What do you think about `addOnBlockingListener`/`addOnBlockingCallback`?

extensions/wikia/AdEngine/js/recovery/helper.js
```diff
@@ -0,0 +1,75 @@
...
+		slotsToRecover = [],
+		onBlockingEventsQueue = [];
+
+	function initEventQueue() {
+		lazyQueue.makeQueue(onBlockingEventsQueue, function (eventHandler) {
+			doc.addEventListener('sp.blocking', function () {
+				eventHandler();
+			});
+		});
+		onBlockingEventsQueue.start();
```
fraszczakszymon: What do you think about making lazy queue of callbacks instead of new EventListeners? We could add only callbacks to queue and start it (execute callbacks) once `sp.blocking` event is dispatched. In this case later added callbacks would be executed immediately after pushing to queue. Other benefit is that we would have only one listener for `sp.blocking` event.

**Other note if you agree:**
~~After creating this kind of queue we should check whether `sp.blocking` is already dispatched - if it is we should start() queue immediately, otherwise we can set up start() as a callback for `sp.blocking` listener.~~
I see that queue is initialised before detection, so we can just start queue on `sp.blocking` event.

extensions/wikia/AdEngine/js/recovery/helper.js
nandy-andy: Sure makes sense. Naming is always a tough task :japanese_ogre: I'll go with `addOnBlockingCallback` :)

fraszczakszymon:
LGTM :+1:

jacekjursza:
LGTM

PR #8535, opened 2015-09-18
this is really a flag

nmonterroso:
@Wikia/services-team 

This is really a flag which controls a small one-time notif a user sees. Note that this was previously broken to not use the constant, so this changeset also fixes that.

See Wikia/config#1325 for the config change


ArturKlajnerok:
Lgtm :+1:

PR #8538, opened 2015-09-18
INT-196: Apply different label if wikia-in-your-lang link is to artic…

mari-mcmurtrie:
…le page.

Branched out from INT-35. Missed a requirement to track impression/link-click/close of wikia-in-your-lang link to article separated from link to main page.

@Grunny will you CR? You know INT-35 well, and this is just about applying different "label" string for tracking.


Grunny:
LGTM.

PR #8540, opened 2015-09-21
(CE-2727) Escape query in WantedPages

Grunny:
Properly escape LIKE clause in our custom WantedPages query page.

/cc @macbre @Wikia/community-engineering


macbre:
LGTM, thanks @Grunny !

PR #8541, opened 2015-09-21
PLATORM-1489: Refactor wfWaitForSlaves to work with consul DB config

macbre:
> Please skip f9a3cce and review on a per-commit basis.

https://wikia-inc.atlassian.net/browse/PLATFORM-1489

After the introduction of Consul-powered DB config in our app, we noticed an increase of CreateNewWiki error rate. It was caused by `wfWaitForSlaves` function not being able to check the lag of all slaves in a given DB cluster (as it was checking only one node that was returned by consul).

This PR introduces `Wikia\Consul\Client` helper class that allows us to query Consul for all service nodes via [it's HTTP API](https://www.consul.io/docs/agent/http/health.html#health_service) (via `sensiolabs/consul-php-sdk` composer package that queries `http://127.0.0.1:8500`).

``` php
> $consul = new Wikia\Consul\Client();
> var_dump( $consul->getNodesFromHostname( 'slave.db-a.service.consul' ) );
Wikia\Consul\Client::getNodes: got nodes for 'db-a' service ('slave' tag): 10.8.32.37, 10.8.38.33, 10.8.32.36, 10.8.36.29
array(4) {
  [0]=>
  string(10) "10.8.32.37"
  [1]=>
  string(10) "10.8.38.33"
  [2]=>
  string(10) "10.8.32.36"
  [3]=>
  string(10) "10.8.36.29"
}

> var_dump( $consul->getNodes( 'db-c', 'slave' ) );
Wikia\Consul\Client::getNodes: got nodes for 'db-c' service ('slave' tag): 10.8.62.59, 10.8.66.72, 10.8.62.56
array(3) {
  [0]=>
  string(10) "10.8.62.59"
  [1]=>
  string(10) "10.8.66.72"
  [2]=>
  string(10) "10.8.62.56"
}
```

It's the same as making the following cURL request:

```
$ curl "http://127.0.0.1:8500/v1/health/service/db-g?tag=slave&passing"
```

```
> wfWaitForSlaves( 'wikicities_c7' )
LBFactory_Wikia::getSectionForWiki: section c7, wiki wikicities_c7
Wikia\Consul\Client::getNodes: got nodes for 'db-g' service ('slave' tag): 10.8.36.19, 10.8.62.64, 10.8.62.58
Connecting to master.db-g.service.consul wikicities_c7...
LoadBalancer::openForeignConnection: opened new connection for 0/wikicities_c7
Connecting to 10.8.36.19 wikicities_c7...
LoadBalancer::openForeignConnection: opened new connection for 1/wikicities_c7
LoadBalancer::doWait: Waiting for slave #1 (10.8.36.19) to catch up...
LoadBalancer::doWait: Done
Connecting to 10.8.62.64 wikicities_c7...
LoadBalancer::openForeignConnection: opened new connection for 2/wikicities_c7
LoadBalancer::doWait: Waiting for slave #2 (10.8.62.64) to catch up...
LoadBalancer::doWait: Done
Connecting to 10.8.62.58 wikicities_c7...
LoadBalancer::openForeignConnection: opened new connection for 3/wikicities_c7
LoadBalancer::doWait: Waiting for slave #3 (10.8.62.58) to catch up...
LoadBalancer::doWait: Done
```

Additional changes:
- do not check slave lags on MW app level when consul is used (cdc0507) - [ER-7747](https://wikia-inc.atlassian.net/browse/ER-7747) and others
- make `Wikia\Logger\WikiaLogger` compatible with `Psr\Log\LoggerInterface` (so that it can be passed to the Consul client for logging to logstash)

@pchojnacki / @drozdo / @gbenson-wikia / @wladekb / @michalroszka / @Wikia/services-team


lib/Wikia/src/Consul/Client.php
```diff
@@ -0,0 +1,90 @@
...
+	 *
+	 * $ curl "http://127.0.0.1:8500/v1/health/service/db-g?tag=slave&passing"
+	 *
+	 * @param string $service
+	 * @param string $tag
+	 * @return array list of IP addresses
+	 */
+	function getNodes( $service, $tag ) {
+		global $wgWikiaDatacenter;
+		$resp = $this->api->service( $service, [ 'tag' => $tag, 'dc' => $wgWikiaDatacenter, 'passing' => true ] )->json();
```
pawelchcki: :+1: 
although I think the DC  should be optional if not specified.
Consul by default will only fetch data from current datacenter the Agent is configured to be in.

But sometimes you might want to address resources from other DC for testing etc.

macbre: I've removed the explicit `dc` param in 7fd0b04

nmonterroso:
it looks like this serves an almost identical purpose as [ConsulUrlProvider](https://github.com/Wikia/app/blob/3d321a9552f2cb6357b7a509eec99e0c74434c7f/lib/Wikia/src/Service/Gateway/ConsulUrlProvider.php). Can we use that instead?

macbre:
ConsulUrlProvider returns one, random node from Consul. In this case we need to get all nodes.

Additionally,  the code from this PR uses the recommended composer package with Consul client and it queries the local Consul Agent instead of the one from `$wgConsulUrl`.

nmonterroso:
We could extend `ConsulUrlProvider` to have an additional function to return the list of all the nodes. I just feel like we're adding a whole lot of changes and a different way to query consul for the same information that we're already able to get using another class.

Your point about the local Consul agent is correct - we should change that in the [config](https://github.com/Wikia/config/search?utf8=%E2%9C%93&q=wgConsulUrl) to be `localhost`

macbre:
Ok, I'll revert f9a3cce (that added the recommended consul client) and modify the existing one + use the localhost to query consul.

pawelchcki:
@nmonterroso I think we could consider using pre existing library to contact Consul. Especially since it already has support for other consul features like KV and Sessions. That will become useful sooner or later.
And incidentally @macbre  did all the required legwork by including it through composer etc.

Additionally I think wgConsulUrl setting should be removed altogether because we should never query on production non local agent.

nmonterroso:
@pchojnacki sure, that sounds like a good idea. But we should consolidate on one way of doing it and not have two separate libraries in `lib/Wikia` providing the essentially same functionality.

pawelchcki:
@nmonterroso I agree 100% that we should only have one library.

includes/GlobalFunctions.php
```diff
@@ -3808,28 +3808,32 @@ function wfWaitForSlaves( $wiki = false ) {
...
-		$masterHostName = $lb->getServerName(0);
-		if ( strpos( $masterHostName, '.service.consul' ) !== false ) {
-			# I am terribly sorry, but we do really need to wait for all slaves
-			# not just the one returned by consul
-			sleep( 1 );
+		// PLATFORM-1489: check if we're using consul configuration for DB slave
+		if ( $lb->hasConsulConfig() ) {
+			// get the list of IP addresses of all slave nodes from consul
+			// so that we can check all of them explicitly
+			$consul = new Wikia\Consul\Client();
```
drsnyder, line 3816: I think I would prefer to use the DI framework that @nmonterroso added to inject the client.

includes/GlobalFunctions.php
```diff
@@ -3808,28 +3808,32 @@ function wfWaitForSlaves( $wiki = false ) {
...
+
+			$slaveInfo = $lb->getServerInfo( 1 ); // e.g. slave.db-g.service.consul
+			$slaves = $consul->getDatabaseNodes( $slaveInfo['hostName'] );
```
drsnyder: This seems like something more general than `getDatabaseNodes`. Should we rename it to `getNodesFromHostname`?

lib/Wikia/src/Consul/Client.php
```diff
@@ -0,0 +1,89 @@
...
+	 * Make it easier to "grep" through the logs
+	 *
+	 * @return array
+	 */
+	protected function getLoggerContext() {
+		return [
+			'class' => __CLASS__
+		];
+	}
+}
```
drsnyder, line 92: Is this going away in favor of fixing `ConsulUrlProvider`?

drsnyder, line 92: Actually, I might favor replacing `ConsulUrlProvider` with this one since it's already using a 3rd party client.

nmonterroso, line 92: Yeah, I agree that this is probably better than interacting directly with the API. But we should consolidate on one. If we use this, it should provide a module to be inject-able within other `lib/Wikia` stuff

includes/GlobalFunctions.php
macbre, line 3816: To be honest, I do not see a benefit of using dependency injection here. There's no reason to test this specific, MW core function nor allow a different provider of IP addresses of nodes behind consul.

includes/GlobalFunctions.php
macbre: This method assumes the following format of the provided address - `slave.db-smw.service.consul` and extracts tag and service name out of it. Can we assume that not only database addresses use this addressing schema?

lib/Wikia/src/Consul/Client.php
macbre, line 92: > it should provide a module to be inject-able within other lib/Wikia stuff

@nmonterroso, can you provide more context / examples here?

macbre:
@nmonterroso / @Danon9111 - thanks for the feedback.

This change is rather big at the moment and addresses the initial requirements. I'll create a separate task to use `Wikia\Consul\Client` in favor of `ConsulUrlProvider` - [PLATFORM-1502](https://wikia-inc.atlassian.net/browse/PLATFORM-1502) Sounds ok?

includes/GlobalFunctions.php
macbre: Renamed in fdd4a0a

Grunny:
LGTM.

macbre:
Thanks, @Grunny!

includes/GlobalFunctions.php
nmonterroso, line 3816: Just because we haven't tested something in the past doesn't mean we can't or shouldn't in the future. Using the DI layer allows us to begin to think about how we can test some of the core MW stuff. Is there any reason _not_ to use it?

PR #8543, opened 2015-09-21
SERVICES-782: User preference get local default fix

harnash:
Added missing parameter to getLocalDefault() call.
@macbre


macbre:
:+1: 

Thanks for the fix!

PR #8545, opened 2015-09-21
LocalPreference usage fixes

nmonterroso:
@Wikia/services-team @garthwebb 
https://wikia-inc.atlassian.net/browse/SERVICES-781
The branch name here is wrong - this is actually for SERVICES-781.

Making cityId non-optional for getting/setting local preferences, and fixing usages. The initial cause for investigation here was dcabb300a1771363dcc050c9ba5607b34cc65238

### DIFF
--- a/extensions/wikia/AutomaticWikiAdoption/AutomaticWikiAdoptionHelper.class.php
+++ b/extensions/wikia/AutomaticWikiAdoption/AutomaticWikiAdoptionHelper.class.php
@@ -190,7 +190,7 @@ class AutomaticWikiAdoptionHelper {
 					//log
 					self::addLogEntry($admin, $oldGroups, $newGroups);
 					//Unset preference for receiving future adoption emails
-					$admin->setLocalPreference( "adoptionmails", $wikiId, 0 );
+					$admin->setLocalPreference( "adoptionmails", 0, $wikiId );
 					$admin->saveSettings();
 				}
 			}
--- a/extensions/wikia/FounderEmails/events/FounderEmailsDaysPassedEvent.class.php
+++ b/extensions/wikia/FounderEmails/events/FounderEmailsDaysPassedEvent.class.php
@@ -79,13 +79,13 @@ class FounderEmailsDaysPassedEvent extends FounderEmailsEvent {
 	}
 
 	public static function register( $wikiParams, $debugMode = false ) {
-		global $wgFounderEmailsEvents;
+		global $wgFounderEmailsEvents, $wgCityId;
 
 		$founderEmailObj = FounderEmails::getInstance();
 
 		$wikiFounder = $founderEmailObj->getWikiFounder();
-		$wikiFounder->setLocalPreference( "founderemails-joins", true );
-		$wikiFounder->setLocalPreference( "founderemails-edits", true );
+		$wikiFounder->setLocalPreference( "founderemails-joins", true, $wgCityId );
+		$wikiFounder->setLocalPreference( "founderemails-edits", true, $wgCityId );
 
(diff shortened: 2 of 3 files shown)


extensions/wikia/AutomaticWikiAdoption/AutomaticWikiAdoptionHelper.class.php
```diff
@@ -190,7 +190,7 @@ static function adoptWiki($wikiId, $user) {
 					//log
 					self::addLogEntry($admin, $oldGroups, $newGroups);
 					//Unset preference for receiving future adoption emails
-					$admin->setLocalPreference( "adoptionmails", $wikiId, 0 );
+					$admin->setLocalPreference( "adoptionmails", 0, $wikiId );
```
drsnyder, line 193: Just to confirm-- was this an error in the position of the `0` and `$wikiId`?

extensions/wikia/FounderEmails/events/FounderEmailsDaysPassedEvent.class.php
```diff
@@ -79,13 +79,13 @@ private function getEmailController( $activateDay ) {
...
-		global $wgFounderEmailsEvents;
+		global $wgFounderEmailsEvents, $wgCityId;
 
 		$founderEmailObj = FounderEmails::getInstance();
 
 		$wikiFounder = $founderEmailObj->getWikiFounder();
-		$wikiFounder->setLocalPreference( "founderemails-joins", true );
-		$wikiFounder->setLocalPreference( "founderemails-edits", true );
+		$wikiFounder->setLocalPreference( "founderemails-joins", true, $wgCityId );
+		$wikiFounder->setLocalPreference( "founderemails-edits", true, $wgCityId );
```
drsnyder, line 88: Isn't this case handled [here](https://github.com/Wikia/app/blob/SERVICES-783/includes/User.php#L2763) as long as `getLocalPreference` sets `$cityId` to `null` by default below?

extensions/wikia/AutomaticWikiAdoption/AutomaticWikiAdoptionHelper.class.php
nmonterroso, line 193: yeah, i found it when i was looking for usages to see if any other place was omitting the wikiId

extensions/wikia/FounderEmails/events/FounderEmailsDaysPassedEvent.class.php
nmonterroso, line 88: it's handled for the mediawiki side of set/get, but the service (which has wikiId as a separate field) does not fall back to the current wiki, it expects it to be explicitly set

drsnyder, line 88: In that case, should it be an error if you try and set a local preference and it's `null` at the Service layer in PHP?

drsnyder:
Minor comments, otw, LGTM

extensions/wikia/FounderEmails/events/FounderEmailsDaysPassedEvent.class.php
nmonterroso, line 88: [It is](https://github.com/Wikia/app/blob/SERVICES-783/lib/Wikia/src/Domain/User/Preferences/LocalPreference.php#L15), that's how we found this :P

drsnyder, line 88: :shipit:

PR #8547, opened 2015-09-21
CE-2707 Add a new method to retrieve just a revision raw from any DB

adamkarminski:
This PR creates a new way to retrieve data on a revision. Instead of trying to supply you with a full dataset with `page` and `user` JOINS it reflects the data that is stored only in the revision table.

That makes it possible to retrieve the data from a GlobalTitle which calls an external DB. As @lukaszkonieczny found out, using `Revision::loadFromId` results in an error because the logic beneath tries to find a User on a given cluster instead of the shared db. Also, if you do not need all the additional data - why bother with retrieving it?

/cc @Grunny @macbre @lukaszkonieczny

### DIFF
--- a/includes/Revision.php
+++ b/includes/Revision.php
@@ -152,6 +152,35 @@ class Revision {
 	}
 
 	/**
+	 * Wikia change begin
+	 *
+	 * Loads Revision with raw data from its ID which is an equivalent of a row in the `revision` table.
+	 *
+	 * @param $revisionId
+	 * @param DatabaseMysqli $db Optional parameter to overwrite usage of the local db.
+	 * @return bool|Revision
+	 */
+	public static function loadRawRevision( $revisionId, DatabaseMysqli $db = null ) {
+		if ( $db === null ) {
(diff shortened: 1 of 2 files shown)


adamkarminski:
Unit tests failing are just the SwiftStorage ones again...

includes/Revision.php
```diff
@@ -152,6 +152,35 @@ public static function newFromRow( $row ) {
...
+		}
+
+		$revision = ( new WikiaSQL() )
+			->SELECT_ALL()
+			->FROM( 'revision' )
+			->WHERE( 'rev_id' )->EQUAL_TO( $revisionId )
+			->LIMIT( 1 )
+			->runLoop( $db, function( &$revision, $row ) {
+				$revision = self::newFromRow( $row );
+			} );
```
macbre: Can't we simply call `$db->selectRow` here? :)

macbre:
Minor comment left, LGTM

macbre:
Looks Awesome To Me :)



Ship it!

PR #8548, opened 2015-09-21
SOC-1331 Some email should always be sent

garthwebb:
Make sure users who choose not to receive any wikia email or are blocked still get email change and confirm emails.

https://wikia-inc.atlassian.net/browse/SOC-1331


jsutterfield:
LGTM!

PR #8550, opened 2015-09-21
Add tests for preferences around the MW User object

drsnyder:
This PR sets up the `Injector` with a mock `PreferenceService` to test the preference related functionality in the `User` object. I'm also filling in tests around getting and setting preferences as I go. 

https://wikia-inc.atlassian.net/browse/SERVICES-648

/cc @Wikia/services-team


includes/wikia/tests/UserTest.php
```diff
@@ -1,24 +1,144 @@
...
+		parent::setUp();
+		$this->setupAndInjectUserPreferenceServiceMock();
+
+		$this->testUser = User::newFromId( self::TEST_USER_ID );
+	}
+
+	private function setupAndInjectUserPreferenceServiceMock() {
+		global $wgGlobalUserPreferenceWhiteList, $wgLocalUserPreferenceWhiteList;
+
+		$this->userPreferenceServiceMock = $this->getMock( 'Wikia\Service\User\Preferences\PreferenceService',
```
nmonterroso: It'd probably be better to use `PreferenceService::class` here, that way if we ever rename the interface IDEs could find usages of it here

includes/wikia/tests/UserTest.php
```diff
@@ -1,24 +1,144 @@
...
+			['getGlobalPreference', 'getPreferences', 'setPreferences', 'setGlobalPreference', 'deleteGlobalPreference',
+			'getLocalPreference', 'setLocalPreference', 'deleteLocalPreference', 'save', 'getGlobalDefault'] );
+
+		$this->userAttributeServiceMock = $this->getMock( 'Wikia\Service\User\Attributes\AttributeService' );
```
nmonterroso: same here, `AttributeService::class`

includes/wikia/tests/UserTest.php
```diff
@@ -1,24 +1,144 @@
...
+		$this->userPreferenceServiceMock->expects( $this->once() )
+			->method( 'getLocalPreference' )
+			->with( $this->testUser->getId(), $cityId, $preference, null, false )
+			->willReturn( $value );
+
+		$this->assertEquals( $value, $this->testUser->getLocalPreference( $preference, $cityId ) );
+	}
+
+	public function testSetLocalPreferenceWithMockedUserPreferenceService() {
+		// this has side effects in the database as long as we are in migration mode
```
nmonterroso: does it? since we're mocking the `PreferenceService` it's probably ok?

nmonterroso:
minor comments otw lgtm

includes/wikia/tests/UserTest.php
drsnyder: Because we are still using `setOptionHelper` in this mode to write to both it should be writing to `user_properties`.

nmonterroso: Ah, ok, the comment is on the function then. I thought it was on 

```
$this->mockGlobalVariable( 'wgPreferenceServiceShadowWrite', true );
```

drsnyder: Sorry, moved the comment up in 34d8004.

PR #8554, opened 2015-09-22
(MAIN-5420) Properly sanitize and validate timestamp in Lightbox query

Grunny:
Make sure the passed value is a timestamp and make sure to still escape
it in the queries.

/cc @macbre @Wikia/community-engineering


lgarczewski:
LGTM.

macbre:
:+1:  Grunny to the rescue :)

PR #8473, opened 2015-09-15
PLATFORM-1476: WikiFactory AJAX entry point for changing variables is vulnerable to CSRF

macbre:
Introduce `axWFactoryValidateRequest` helper that asserts:
- that the request was posted
- that the request has a valid edit token set

If not, throw an exception and return HTTP 500 response code.

@michalroszka / @Grunny

PR #8556, opened 2015-09-22
Filtering variable list does not work in Special:WhereIsExtension

macbre:
`axWFactoryFilterVariables` AJAX method: no need to force POST and check the token here + send a proper `Content-Type` response header.

Fix a regression introduced by #8473

@michalroszka / @wladekb / @Grunny / @lukaszkonieczny


Grunny:
:+1:

PR #8558, opened 2015-09-22
SOC-1296 Helios and Wikia email validation does not match

rafalkalinski:
mail validation regex in mediawiki have to match what is happening in helios

### DIFF
--- a/extensions/wikia/UserLogin/UserLoginHooksHelper.class.php
+++ b/extensions/wikia/UserLogin/UserLoginHooksHelper.class.php
@@ -119,7 +119,7 @@ class UserLoginHooksHelper {
 		// this validation match Helios email validation
 		// [see: https://github.com/Wikia/helios/blob/ce490394ae10eb248afd2a063dc465d8eb4efb01/domain/user_validator_funs.go#L29]
 		return preg_match(
-			'^[\w!#$%&\'*+\/=?^_`{|}~-]+(?:\.[\w!#$%&\'*+\/=?^_`{|}~-]+)*@(?:[\w](?:[\w-]*[\w])?\.)+[a-zA-Z0-9](?:[\w-]*[\w])?$/i',
+			'/^[\w!#$%&\'*+\/=?^_`{|}~-]+(?:\.[\w!#$%&\'*+\/=?^_`{|}~-]+)*@(?:[\w](?:[\w-]*[\w])?\.)+[a-zA-Z0-9](?:[\w-]*[\w])?$/i',
 			$addr
 		) !== 0;
 	}
--- a/extensions/wikia/UserLogin/UserLoginHooksHelper.class.php
+++ b/extensions/wikia/UserLogin/UserLoginHooksHelper.class.php
@@ -116,7 +116,12 @@ class UserLoginHooksHelper {
 	}
 
 	public static function isValidEmailAddr( $addr ) {
-		return preg_match( '/^[a-z0-9._%+-]+@(?:[a-z0-9\-]+\.)+[a-z]{2,4}$/i', $addr ) !== 0;
+		// this validation match Helios email validation
+		// [see: https://github.com/Wikia/helios/blob/ce490394ae10eb248afd2a063dc465d8eb4efb01/domain/user_validator_funs.go#L29]
+		return preg_match(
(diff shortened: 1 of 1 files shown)


extensions/wikia/UserLogin/UserLoginHooksHelper.class.php
```diff
@@ -116,7 +116,12 @@ public static function onSetUserEmail( User $user, $newEmail, &$result, &$info )
 	}
 
 	public static function isValidEmailAddr( $addr ) {
-		return preg_match( '/^[a-z0-9._%+-]+@(?:[a-z0-9\-]+\.)+[a-z]{2,4}$/i', $addr ) !== 0;
+		// this validation match Helios email validation
+		// [see: https://github.com/Wikia/helios/blob/ce490394ae10eb248afd2a063dc465d8eb4efb01/domain/user_validator_funs.go#L29]
+		return preg_match(
+			'^[\w!#$%&\'*+\/=?^_`{|}~-]+(?:\.[\w!#$%&\'*+\/=?^_`{|}~-]+)*@(?:[\w](?:[\w-]*[\w])?\.)+[a-zA-Z0-9](?:[\w-]*[\w])?$/i',
```
kvas-damian: aren't we missing `/` as a pattern delimiter at the beginning of pattern?

kvas-damian:
LGTM as a fix for this bug.
Will create separate ticket for Helios to allow domains like `.பரிட்சை`
SERVICES-796

extensions/wikia/UserLogin/UserLoginHooksHelper.class.php
```diff
@@ -116,7 +116,12 @@ public static function onSetUserEmail( User $user, $newEmail, &$result, &$info )
 	}
 
 	public static function isValidEmailAddr( $addr ) {
-		return preg_match( '/^[a-z0-9._%+-]+@(?:[a-z0-9\-]+\.)+[a-z]{2,4}$/i', $addr ) !== 0;
+		// this validation match Helios email validation
+		// [see: https://github.com/Wikia/helios/blob/ce490394ae10eb248afd2a063dc465d8eb4efb01/domain/user_validator_funs.go#L29]
+		return preg_match(
+			'/^[\w!#$%&\'*+\/=?^_`{|}~-]+(?:\.[\w!#$%&\'*+\/=?^_`{|}~-]+)*@(?:[\w](?:[\w-]*[\w])?\.)+[a-zA-Z0-9](?:[\w-]*[\w])?$/i',
+			$addr
+		) !== 0;
```
garthwebb, line 124: We shouldn't replace one hand made email regex with another.  The actual email validation regex is lengthy and we're guaranteed to miss something if we try our own.  PHP has a validation function that can be used:

```
http://php.net/manual/en/filter.examples.validation.php
```

So, looks like:

```
if (filter_var($email, FILTER_VALIDATE_EMAIL)) { ... }
```

I've asked the Helios team to update their code to use the Go library rather than this lib as well.

armonr, line 124: :+1: :+1:

garthwebb, line 124: Though, it seems that filter_var doesn't handle multibyte emails either.  Given neither supports this, I'd rather go with the built in `filter_var` for now and then address this when the Helios team expands to support multibyte.

kvas-damian, line 124: @garthwebb `filter_var` is not compatible with regex used in Helios, see example: https://ideone.com/8tD9TZ so it won't fix initial problem. MW will still failing for those email addresses (See ticket description).

As you can see in SERVICES-796 it won't be fixed in near future, so for now I'd proceed with existing Helios regex to avoid problems in MW. When it gets fixed in Helios we you update our way of validating email addresses.

rafalkalinski, line 124: @garthwebb if the goal was to have the same behavior in helios and in mediawiki, so it looks like for now we have to use the same regex as in helios.

garthwebb, line 124: Fine with that.  We can change it out later when we work this out.

garthwebb:
LGTM

PR #8560, opened 2015-09-22
ADEN-2393  extract robot policy to separate fn 

jacekjursza:
Extract robot policy to separate fn in order to reuse the  logic in MercuryApiController.


includes/wikia/Wikia.php
```diff
@@ -1604,18 +1625,9 @@ static public function onSkinTemplateOutputPageBeforeExec(SkinTemplate $skinTemp
...
-		if( !empty($stagingHeader) ) {
-		// we've got special cases like externaltest.* and showcase.* aliases:
-		// https://github.com/Wikia/wikia-vcl/blob/master/wikia.com/control-stage.vcl#L15
-		// those cases for backend look like production,
-		// therefore we don't want to base only on environment variables
-		// but on HTML headers as well, see:
-		// https://github.com/Wikia/app/blob/dev/redirect-robots.php#L285
-			$out->setRobotPolicy( 'noindex,nofollow' );
+		$robotPolicy = Wikia::getEnvironmentRobotPolicy( $skinTemplate->getRequest() );
+		if ( $robotPolicy ) {
```
nandy-andy: How about being more strict and use `!is_null()` here? On the other hand there is a nice read: https://www.lucidchart.com/techblog/2015/08/31/the-worst-mistake-of-computer-science/

Therefore, another idea: `Wikia::getEnvironmentRobotPolicy()` returns **always** a string if it's an empty string we won't call `Output::setRobotPolicy()` :) What do you think?

nandy-andy:
Except this small comment looks fine. SEO Team, take a look! :smiley_cat: 

//cc @gabrys @saipetch

gabrys:
My only concern is, the robots policy seems to be per-environment only. What about per-page robots policy? Like articles: yes please crawl, special pages: keep away, selected special pages: please follow, but don't index?

jacekjursza:
@gabrys Good question, but fortunately SpecialPages are covered, as they fallback to wikiamobile skin. For example take a look at: http://muppet.wikia.com/wiki/Special:AllPages?useskin=mercury.

PR #8563, opened 2015-09-22
CE-2672 Refactoring of the replaceWithLastApprovedRevision unit test

adamkarminski:
I recommend viewing this one in the Split mode. :)

The refactoring included:
1. Separation of the `isPageReviewed()` method that initially qualifies pages to undergo the further process and writing a unit test for it.
2. Rewriting of the main unit test and adding cases that cover all possible scenarios.

/cc @Grunny @lukaszkonieczny

### DIFF
--- a/extensions/wikia/ContentReview/tests/ContentReviewHelperTest.php
+++ b/extensions/wikia/ContentReview/tests/ContentReviewHelperTest.php
@@ -234,10 +234,10 @@ class ContentReviewHelperTest extends WikiaBaseTest {
 			[
 				[
 					'isPageReviewed' => false,
-					'latestRevID' => 0,
+					'latestRevID' => 100,
 					'latestReviewedRevision' => [],
 					'isTestModeEnabled' => false,
-					'revisionExists' => false,
+					'revisionExists' => true,
 					'originalText' => $originalText,
 					'lastReviewedText' => $lastReviewedText,
 				],
--- a/extensions/wikia/ContentReview/tests/ContentReviewHelperTest.php
+++ b/extensions/wikia/ContentReview/tests/ContentReviewHelperTest.php
@@ -8,24 +8,16 @@ class ContentReviewHelperTest extends WikiaBaseTest {
 	}
 
 	/**
-	 * @param bool $isJsPage
-	 * @param bool $inNamespace
+	 * @param $titleData
 	 * @param string $contentType
 	 * @param bool $expected
+	 * @param $message
 	 * @dataProvider isPageForReviewTestData
 	 */
-	public function testIsPageForReview( $isJsPage, $inNamespace, $contentType, $expected ) {
-		$titleMock = $this->getMock( '\Title', [ 'isJsPage', 'inNamespace' ] );
-		$titleMock->expects( $this->once() )
-			->method( 'isJsPage' )
-			->willReturn( $isJsPage );
-
-		$titleMock->expects( $this->any() )
-			->method( 'inNamespace' )
-			->willReturn( $inNamespace );
-
-		$value = ( new Wikia\ContentReview\Helper() )->isPageReviewed( $titleMock, $contentType );
-		$this->assertEquals( $expected, $value );
+	public function testIsPageForReview( $titleData, $contentType, $expected, $message ) {
+		$title = Title::makeTitle( $titleData['titleNamespace'], $titleData['titleText'] );
+		$value = ( new Wikia\ContentReview\Helper() )->isPageReviewed( $title, $contentType );
+		$this->assertEquals( $expected, $value, $message );
 	}
 
 	/**
@@ -106,7 +98,7 @@ class ContentReviewHelperTest extends WikiaBaseTest {
 					->method( 'getRawText' )
 					->willReturn( $lastReviewedText );
 			} else {
-				$revisionMock = false;
+				$revisionMock = null;
 			}
 
 			$helperMock->expects( $this->any() )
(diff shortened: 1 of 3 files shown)


extensions/wikia/ContentReview/tests/ContentReviewHelperTest.php
```diff
@@ -8,61 +8,119 @@ public function setUp() {
...
 
 	/**
-	 * @dataProvider replaceWithLastApprovedRevisionProvider
+	 * @param bool $isJsPage
+	 * @param bool $inNamespace
+	 * @param string $contentType
+	 * @param bool $expected
+	 * @dataProvider shouldPageContentBeReplacedData
+	 */
+	public function testShouldPageContentBeReplaced( $isJsPage, $inNamespace, $contentType, $expected ) {
```
Grunny: I'd suggest using the standard naming convention of `test<method name>` so `testIsPageReviewed`, as that helper method may be used for other purposes.

adamkarminski: Awww, sorry, a leftover :)

extensions/wikia/ContentReview/tests/ContentReviewHelperTest.php
```diff
@@ -98,96 +156,113 @@ public function testHasPageApprovedId( $latestReviewedId, $oldId, $expected, $me
...
-		$text1 = '';
-		$text2 = '';
-		$jsType = 'jstype';
+		$originalText = 'This is the original text.';
+		$lastReviewedText = 'This is the last reviewed text.';
+		$emptyText = '';
+
 		return [
+			/**
+			 * Case 1 - a page does not qualify to be reviewed (isPageReviewed === false).
```
Grunny: I'd pass these comments to the test method and using them with the assert as they provide context when the test case fails when running the tests (either manually or in Jenkins).

adamkarminski: :+1:

extensions/wikia/ContentReview/tests/ContentReviewHelperTest.php
```diff
@@ -8,61 +8,119 @@ public function setUp() {
...
+	 */
+	public function testShouldPageContentBeReplaced( $isJsPage, $inNamespace, $contentType, $expected ) {
+		$titleMock = $this->getMock( '\Title', [ 'isJsPage', 'inNamespace' ] );
+		$titleMock->expects( $this->once() )
+			->method( 'isJsPage' )
+			->willReturn( $isJsPage );
+
+		$titleMock->expects( $this->any() )
+			->method( 'inNamespace' )
+			->willReturn( $inNamespace );
```
Grunny: This doesn't check much, as it doesn't tell us anything about what the title is and what namespace is being checked. You could just add a `with` here, but as a small suggestion to abstract it a little more and allow explicit tests for things like https://wikia-inc.atlassian.net/browse/CE-2708, you could avoid mocking altogether given the methods used and just use `Title::makeTitle` so you're instead just controlling what the title is, as in this test:
https://github.com/Wikia/app/blob/3efa2cfe7144f/extensions/wikia/ProtectSiteJS/tests/ProtectSiteJSTest.php#L16

Up to you though. The `with` is needed at the very least to verify you're just checking the MW namespace.

extensions/wikia/ContentReview/tests/ContentReviewHelperTest.php
```diff
@@ -8,61 +8,119 @@ public function setUp() {
...
-		$helperMock->method( 'isContentReviewTestModeEnabled' )
-			->will( $this->returnValue( $params['isContentReviewTestModeEnabled'] ) );
-		$helperMock->method( 'getRevisionById' )
-			->will( $this->returnValue( $revisionMock ) );
-
-		$helperMock->replaceWithLastApproved( $titleMock, $params['contentType'], $params['text'] );
-
-		$this->assertEquals( $textExpected, $params['text'], $message );
+	public function testReplaceWithLastApprovedRevision( $isPageReviewed, $latestRevID,
+		array $latestReviewedRevision, $isTestModeEnabled, $revisionExists, $originalText, $lastReviewedText, $expectedText
```
Grunny: This method signature is getting a bit big. I'd suggest switching to passing an array of test data, which will also make your data provider clearer (without the need to comment array entries) since the keys will be descriptive (i.e. https://github.com/Wikia/app/blob/3efa2cfe7144f/extensions/wikia/ProtectSiteJS/tests/ProtectSiteJSTest.php#L49-L56).

extensions/wikia/ContentReview/tests/ContentReviewHelperTest.php
```diff
@@ -8,61 +8,119 @@ public function setUp() {
...
-		$helperMock->method( 'isContentReviewTestModeEnabled' )
-			->will( $this->returnValue( $params['isContentReviewTestModeEnabled'] ) );
-		$helperMock->method( 'getRevisionById' )
-			->will( $this->returnValue( $revisionMock ) );
-
-		$helperMock->replaceWithLastApproved( $titleMock, $params['contentType'], $params['text'] );
-
-		$this->assertEquals( $textExpected, $params['text'], $message );
+	public function testReplaceWithLastApprovedRevision( $isPageReviewed, $latestRevID, array $latestReviewedRevision,
+		$isTestModeEnabled, $revisionExists, $originalText, $lastReviewedText, $expectedText, $message
```
Grunny: This method signature is getting a bit big. I'd suggest switching to passing an array of test data, which will also make your data provider clearer (without the need to comment array entries) since the keys will be descriptive (i.e. https://github.com/Wikia/app/blob/3efa2cfe7144f/extensions/wikia/ProtectSiteJS/tests/ProtectSiteJSTest.php#L49-L56).

extensions/wikia/ContentReview/tests/ContentReviewHelperTest.php
```diff
@@ -8,61 +8,119 @@ public function setUp() {
...
+				/* @var \Revision $revisionMock */
+				$revisionMock = $this->getMockBuilder( '\Revision' )
+					->disableOriginalConstructor()
+					->setMethods( [ 'getRawText' ] )
+					->getMock();
+				$revisionMock->expects( $this->any() )
+					->method( 'getRawText' )
+					->willReturn( $lastReviewedText );
+			} else {
+				$revisionMock = false;
```
Grunny: Minor, but to be explicit `Revision::newFromId` returns `null` on a non-existent rev ID, rather than `false`.

extensions/wikia/ContentReview/tests/ContentReviewHelperTest.php
```diff
@@ -98,96 +156,103 @@ public function testHasPageApprovedId( $latestReviewedId, $oldId, $expected, $me
...
-					'isContentReviewTestModeEnabled' => false,
-					'text' => $text2,
-				],
-				$text1,
-				'Current text replaced with last approved revision',
+				true, // isPageReviewed()
+				100, // getLatestRevID()
+				[ 'revision_id' => 100 ], // getLatestReviewedRevision()
+				false, // isContentReviewTestModeEnabled()
+				false, // $revisionExists
```
Grunny: The revision does exist in this case, right?

extensions/wikia/ContentReview/tests/ContentReviewHelperTest.php
```diff
@@ -98,96 +156,103 @@ public function testHasPageApprovedId( $latestReviewedId, $oldId, $expected, $me
...
-					'isContentReviewTestModeEnabled' => false,
-					'text' => $text2,
-				],
-				$textEmpty,
-				'No approved revision. Text empty.',
+				true, // isPageReviewed()
+				100, // getLatestRevID()
+				[ 'revision_id' => 99 ], // getLatestReviewedRevision()
+				true, // isContentReviewTestModeEnabled()
+				false, // $revisionExists
```
Grunny: Same here.

extensions/wikia/ContentReview/tests/ContentReviewHelperTest.php
adamkarminski: You're right, I've confused it with `ResultWrapper|false` returns. Thanks! :)

extensions/wikia/ContentReview/tests/ContentReviewHelperTest.php
adamkarminski: The value does not really matter, since the script should not get to it, but it makes more sense in terms of crafting a scenario. Thanks! :)

Grunny:
:+1:

PR #8564, opened 2015-09-22
ramping preference correction to 5%

nmonterroso:
@Wikia/services-team 

bumping preference comparison/correction to 5%


drsnyder:
lgtm

PR #8565, opened 2015-09-22
SEO-69 Fix PHP Warning

gabrys:
PHP Warning: array_values() expects parameter 1 to be array, null given in /extensions/wikia/SeoLinkHreflang/SeoLinkHreflang.class.php on line 89


saipetch:
LGTM

PR #8566, opened 2015-09-22
Pref comparison fixes

nmonterroso:
@Wikia/services-team 

The PreferenceCorrectionService has been trying to correct some preferences, but has been unsuccessful because our `prefIsSaveable` check is not the same as [`shouldOptionBeStored`](https://github.com/Wikia/app/blob/dev/includes/User.php#L5037).

### DIFF
--- a/lib/Wikia/src/Service/User/Preferences/Migration/PreferenceCorrectionService.php
+++ b/lib/Wikia/src/Service/User/Preferences/Migration/PreferenceCorrectionService.php
@@ -50,7 +50,7 @@ class PreferenceCorrectionService {
 				'correcting preferences',
 				[
 					'num_differences' => $differences,
-					'user_id' => $userId] );
+					'userId' => $userId] );
 			$this->preferenceService->setPreferences( $userId, $expected );
 			$this->preferenceService->save( $userId );
 		}
@@ -86,7 +86,7 @@ class PreferenceCorrectionService {
 				if ( !$actualPreferences->hasLocalPreference( $prefName, $wikiId ) ) {
 					$this->logMissingPreference( $userId, $name );
 					++$differences;
-				} elseif ( $actualPreferences->getLocalPreference( $prefName, $wikiId ) !== $value ) {
+				} elseif ( $actualPreferences->getLocalPreference( $prefName, $wikiId ) != $value ) {
 					$this->logPreferenceValueDifference( $userId, $name, $value, $actualPreferences->getLocalPreference( $prefName, $wikiId ) );
 					++$differences;
 				}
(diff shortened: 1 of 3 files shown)


lib/Wikia/src/Service/User/Preferences/Migration/PreferenceCorrectionService.php
```diff
@@ -50,7 +50,7 @@ public function compareAndCorrect( $userId, $userOptions ) {
 				'correcting preferences',
 				[
 					'num_differences' => $differences,
-					'user_id' => $userId] );
+					'userId' => $userId] );
```
nmonterroso, line 53: Just changing this to be consistent with the other logs

lib/Wikia/src/Service/User/Preferences/PreferenceServiceImpl.php
```diff
@@ -258,11 +258,7 @@ private function applyDefaults( UserPreferences $preferences ) {
 	}
 
 	private function prefIsSaveable( $pref, $value, $valueFromDefaults ) {
```
nmonterroso, line 260: The problem here was that we're doing a comparison against the defaults and returning early based on that, but we already do a check against the default below, but the following condition should cause the pref to be saved.

lib/Wikia/src/Service/User/Preferences/PreferenceServiceImpl.php
```diff
@@ -258,11 +258,7 @@ private function applyDefaults( UserPreferences $preferences ) {
...
-		if ( $value == $valueFromDefaults ) {
-			return false;
-		}
-
 		return in_array( $pref, $this->forceSavePrefs ) || $value != $valueFromDefaults ||
-			( $valueFromDefaults != null && $value !== false && $value !== null );
+			( $valueFromDefaults === null && $value !== false && $value !== null );
```
nmonterroso, line 262: I think this is leftover from the previous migration

lib/Wikia/src/Service/User/Preferences/Migration/PreferenceCorrectionService.php
```diff
@@ -86,7 +86,7 @@ private function compare( UserPreferences $actualPreferences, $userId, $options
 				if ( !$actualPreferences->hasLocalPreference( $prefName, $wikiId ) ) {
 					$this->logMissingPreference( $userId, $name );
 					++$differences;
-				} elseif ( $actualPreferences->getLocalPreference( $prefName, $wikiId ) !== $value ) {
+				} elseif ( $actualPreferences->getLocalPreference( $prefName, $wikiId ) != $value ) {
```
nmonterroso, line 89: This change was also to be consistent with what the globals comparison does

drsnyder:
lgtm

PR #8567, opened 2015-09-22
only do comparisons when loading a user from the database

nmonterroso:
@Wikia/services-team 

Only do comparisons when loading a user's options from the database. We need this because there are multiple caches inside mediawiki for a user (ex: `UserService`, `User`), which are not kept in sync, so we can't be certain that the options we're comparing against are up-to-date with the preferences (which only maintain a single cache, so should always be up-to-date)

### DIFF
--- a/includes/User.php
+++ b/includes/User.php
@@ -4905,12 +4905,12 @@ class User {
 			}
 
 			$this->loadAttributes();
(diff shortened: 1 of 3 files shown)


drsnyder:
Makes sense. LGTM

PR #8568, opened 2015-09-23
Add a one minute cache to attribute logging

jsutterfield:
Update the logging we're doing to include a one minute cache. This is to test how many requests per minutes we could expect the attribute service to receive if we had MW call out to it whenever it needed a value for "avatar" or "location", with a one minute cache after each request. 

See https://wikia-inc.atlassian.net/browse/SOC-1315 for more details.


rafalkalinski:
LGTM

PR #8569, opened 2015-09-23
SOC-1315 Skip MW User cache for "avatar" and "location" attributes

jsutterfield:
Update MW so that any time it needs a value for "avatar" or "location" it skips its user cache and talks directly to the service. Those are the only 2 attributes which are updated by clients other than MW, so everything else can still be accessed in the user cache.


extensions/wikia/AttributeServiceMWApi/AttributeServiceMWApi.setup.php
```diff
@@ -1,9 +0,0 @@
-<?php
```
jsutterfield: I reverted all of this `AttributeServiceMWApiController.class.php` nonsense.

extensions/wikia/UserProfilePageV3/UserIdentityBox.class.php
```diff
@@ -20,6 +20,8 @@ class UserIdentityBox {
 	const USER_OCCUPATION_CHAR_LIMIT = 200;
 	const USER_GENDER_CHAR_LIMIT = 200;
 
+	const CACHE_TTL = 60 * 60 * 24; // 24 hours
```
jsutterfield: User data on profile pages used to be cached indefinitely until the next update. With the discussion apps, "location" can now be updated which is part of that cached data. I added a TTL so that we would have eventual consistency in the case of an update by the discussion app.

I'm being kind of conservative here and picking 24 hours to reduce db hits and requests to the attribute service. That may be too long though.

includes/User.php
```diff
@@ -2683,32 +2683,40 @@ public function getDefaultGlobalPreference( $preference ) {
...
-	private function logAttribute( $attributeName ) {
-		$this->info( 'USER_ATTRIBUTES get_attribute_call', [
-			'attributeName' => $attributeName,
-			'userId' => $this->getId()
-		] );
+		if ( empty( $wgEnableReadsFromAttributeService ) ) {
+			return false;
+		}
+
+		if ( !$this->isLoggedIn() ) {
```
kvas-damian: @jsutterfield do we really want if user is logged in or is it only check if user exists?

I can imagine our code calling this method not for current user object to fetch info about past contributors or to display avatars in the forum thread like this: http://muppet.wikia.com/wiki/Thread:203563

kvas-damian:
In https://github.com/Wikia/app/pull/8568 we dropped number of requests by 80%. We might try to use shared memcache for caching  UserAttributes instead of making local copies for every wiki. I don't think we support local attributes in our service.

Counter based on shared memcache implemented in: https://github.com/Wikia/app/pull/8576 - looks it doesn't help us much :(

lib/Wikia/src/Service/User/Attributes/UserAttributes.php
```diff
@@ -67,55 +90,53 @@ public function setAttribute( $userId, Attribute $attribute ) {
...
+		$this->attributeService->delete( $userId, $attribute );
+	}
+
+	private function deleteFromInstanceCache( $userId, Attribute $attribute ) {
 		unset( $this->attributes[$userId][$attribute->getName()] );
 	}
 
-	private function deleteAttributeFromService( $userId, Attribute $attribute ) {
-		$this->attributeService->delete( $userId, $attribute );
+	private function isAnonUser( $userId ) {
```
kvas-damian: I'm not sure if this is responsibility of `UserAttributes` class.

lib/Wikia/src/Service/User/Attributes/UserAttributes.php
```diff
@@ -47,15 +53,32 @@ public function getAttribute( $userId, $attributeName, $default = null ) {
...
 
 	private function loadAttributes( $userId ) {
-		if ( !isset( $this->attributes[$userId] ) ) {
-			$this->attributes[$userId] = [];
+
+		if ( isset( $this->attributes[$userId] ) ) {
+			return $this->attributes[$userId];
+		}
+
+		$attributes = $this->loadFromMemcache( $userId );
```
kvas-damian, line 65: Did you consider using https://github.com/Wikia/app/blob/dev/includes/wikia/WikiaDataAccess.class.php#L52 ?

lib/Wikia/src/Service/User/Attributes/UserAttributes.php
```diff
@@ -47,15 +53,32 @@ public function getAttribute( $userId, $attributeName, $default = null ) {
...
+		if ( empty( $attributes ) ) {
```
kvas-damian: What do you think about allowing empty array here?

lib/Wikia/src/Service/User/Attributes/UserAttributes.php
jsutterfield, line 65: I didn't think about that. I borrowed this approach from what the [preferences guys are doing](https://github.com/Wikia/app/blob/dev/lib/Wikia/src/Service/User/Preferences/PreferenceServiceImpl.php). For a second pass we could consider updating though.

lib/Wikia/src/Service/User/Attributes/UserAttributes.php
jsutterfield: Sure, that sounds good.

lib/Wikia/src/Service/User/Attributes/UserAttributes.php
jsutterfield: I think you're right. I was being a little paranoid about making sure anon users don't use the service, but it looks like `User.php` [checks for that](https://github.com/Wikia/app/blob/dev/includes/User.php#L3525) when it calls `saveSettings()`

includes/User.php
jsutterfield: Good point @kvas-damian. It looks like this method name is actually misleading. It just calls `$this->getId()` and returns if it's 0. `$this->getId()` says it returns 0 if [the user is anonymous or nonexistent](https://github.com/Wikia/app/blob/dev/includes/User.php#L1860). So it does what we want, but the method name is confusing. I'll update to just use `$this->getId()` to make it more clear.

extensions/wikia/UserProfilePageV3/UserIdentityBox.class.php
```diff
@@ -20,6 +20,8 @@ class UserIdentityBox {
 	const USER_OCCUPATION_CHAR_LIMIT = 200;
 	const USER_GENDER_CHAR_LIMIT = 200;
 
+	const CACHE_TTL = 60 * 60; // 1 hour
```
jsutterfield, line 23: Cache profile pages for an hour.

This means if a user updates their location in the discussion app it won't be reflected on their profile page for an hour, which isn't super ideal, but far less noticeable than avatars since location is only used on profile pages. This used to be cached indefinitely, so I'm concerned about dropping this TTL too much until we see if there's much of a performance cost.

extensions/wikia/AttributeServiceMWApi/AttributeServiceMWApi.setup.php
jsutterfield, line 1: Reverted this AttributeServiceMWApiController nonsense since we're taking a different approach.

lib/Wikia/src/Service/User/Attributes/UserAttributes.php
```diff
@@ -47,78 +57,84 @@ public function getAttribute( $userId, $attributeName, $default = null ) {
...
 	private function loadAttributes( $userId ) {
-		if ( !isset( $this->attributes[$userId] ) ) {
-			$this->attributes[$userId] = [];
+
+		if ( isset( $this->attributes[$userId] ) ) {
+			return $this->attributes[$userId];
+		}
+
+		$attributes = $this->loadFromMemcache( $userId );
+		if ( ( $attributes == false ) ) {
```
garthwebb: Extra layer of parenthesis here

garthwebb:
One minor comment but LGTM after addressing unit tests

PR #8573, opened 2015-09-23
VignetteRequest::isVignetteUrl - improve detection of b0rken URLs

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-1486

`` tags with a proper Vignette URL were treated by `VignetteRequest::isVignetteUrl` as valid URLs.

And log `InvalidArgumentException` as warnings.

@pchojnacki

### DIFF
--- a/includes/wikia/vignette/VignetteUrlToUrlGenerator.php
+++ b/includes/wikia/vignette/VignetteUrlToUrlGenerator.php
@@ -23,13 +23,13 @@ class VignetteUrlToUrlGenerator {
 		$this->asOriginal = $asOriginal;
 
 		if (!VignetteRequest::isVignetteUrl($url)) {
-			$this->reportError('invalid url');
+			$this->reportWarning('invalid url');
 		}
 	}
 
 	public function build() {
 		if (!$this->parseUrl()) {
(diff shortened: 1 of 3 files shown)


includes/wikia/vignette/VignetteUrlToUrlGenerator.php
```diff
@@ -123,11 +123,11 @@ private function addThumbDefinition(UrlGenerator $generator) {
...
 	}
 
-	private function reportError($message, Array $context=[]) {
+	private function reportWarning($message, Array $context=[]) {
 		$exception = new InvalidArgumentException($message);
 		$context['exception'] = $exception;
 
-		$this->error($message, $context);
+		$this->warning($message, $context);
 		throw $exception;
```
ghost, line 131: While triggering a warning, we expect the execution to continue. Will it continue after we throw an exception here?

macbre, line 131: It will as this method calls (from Vignette class) are wrapped in `try / catch` blocks.

ghost:
:+1:

PR #8576, opened 2015-09-23
SOC-1315 | Shared memcache for user attributes counter

kvas-damian:
User attributes are global, so there is no need to count queries to service when shared instance of cache has this data.

ping @bkoval

### DIFF
--- a/lib/Wikia/src/Service/User/Attributes/UserAttributes.php
+++ b/lib/Wikia/src/Service/User/Attributes/UserAttributes.php
@@ -119,6 +119,6 @@ class UserAttributes {
 	}
 
 	public static function getCacheKey( $userId ) {
-		return wfMemcKey( $userId, __CLASS__ );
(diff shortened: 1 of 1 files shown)


nandy-andy:
LGTM :+1:

macbre:
:+1:

lib/Wikia/src/Service/User/Attributes/UserAttributes.php
```diff
@@ -119,6 +119,6 @@ private function deleteAttributeFromService( $userId, Attribute $attribute ) {
 	}
 
 	public static function getCacheKey( $userId ) {
-		return wfMemcKey( $userId, __CLASS__ );
+		return wfSharedMemcKey($userId, __CLASS__);
```
macbre, line 122: A minor one, we usually pass a `__CLASS__` as the first argument.

PR #8578, opened 2015-09-23
CE-2721: Check more permissions while approving template draft

lukaszkonieczny:
Do not allow to approve template drafts if template draft or draft parent page is protected.

### DIFF
--- a/extensions/wikia/TemplateDraft/ApprovedraftAction.php
+++ b/extensions/wikia/TemplateDraft/ApprovedraftAction.php
@@ -77,10 +77,6 @@ class ApprovedraftAction extends FormlessAction {
 			throw new ErrorPageError( 'badaccess', 'templatedraft-protect-edit' );
 		}
 
-		if ( !$parentTitle->userCan( 'move' ) ) {
-			throw new ErrorPageError( 'badaccess', 'templatedraft-protect-move' );
-		}
-
 		// Get contents of draft page
 		$article = Article::newFromId( $draftTitle->getArticleID() );
 		$draftContent = $article->getContent();
--- a/extensions/wikia/TemplateDraft/ApprovedraftAction.php
+++ b/extensions/wikia/TemplateDraft/ApprovedraftAction.php
@@ -73,8 +73,12 @@ class ApprovedraftAction extends FormlessAction {
 		$parentTitle = TemplateDraftHelper::getParentTitle( $draftTitle );
 
 		// Check edit rights
-		if ( !$parentTitle->userCan( 'templatedraft' ) ) {
-			throw new PermissionsException( 'edit' );
+		if ( !$parentTitle->userCan( 'templatedraft' ) || !$parentTitle->userCan( 'edit' ) ) {
+			throw new ErrorPageError( 'badaccess', 'templatedraft-protect-edit' );
+		}
+
+		if ( !$parentTitle->userCan( 'move' ) ) {
+			throw new ErrorPageError( 'badaccess', 'templatedraft-protect-move' );
 		}
 
 		// Get contents of draft page
--- a/extensions/wikia/TemplateDraft/ApprovedraftAction.php
+++ b/extensions/wikia/TemplateDraft/ApprovedraftAction.php
@@ -70,8 +70,7 @@ class ApprovedraftAction extends FormlessAction {
 		global $wgEnableInsightsInfoboxes;
 
 		// Get Title object of parent page
-		$helper = new TemplateDraftHelper();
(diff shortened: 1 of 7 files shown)


skins/oasis/modules/TemplateDraftModuleController.class.php
```diff
@@ -28,8 +28,17 @@ public function executeApprove( $params ) {
...
-		] );
+		$this->allowApprove = false;
+		$draftTitle = $this->wg->Title;
+		$parentTitle = TemplateDraftHelper::getParentTitle( $draftTitle );
+
+		if ( $draftTitle->userCan( 'edit' ) && $draftTitle->userCan( 'move' )
+			&& $parentTitle->userCan( 'edit' ) && $parentTitle->userCan( 'move' )
+		) {
+			$this->allowApprove = true;
+			$this->draftUrl = $this->wg->Title->getFullUrl( [
```
Grunny: Super minor, but I'd be consistent in using `$draftTitle` here.

extensions/wikia/TemplateDraft/ApprovedraftAction.php
```diff
@@ -70,12 +70,15 @@ private function approveDraft( Title $draftTitle ) {
...
+		$parentTitle = TemplateDraftHelper::getParentTitle( $draftTitle );
 
 		// Check edit rights
-		if ( !$parentTitle->userCan( 'templatedraft' ) ) {
-			throw new PermissionsException( 'edit' );
+		if ( !$parentTitle->userCan( 'templatedraft' ) || !$parentTitle->userCan( 'edit' ) ) {
+			throw new ErrorPageError( 'badaccess', 'templatedraft-protect-edit' );
+		}
+
+		if ( !$parentTitle->userCan( 'move' ) ) {
```
Grunny: Do we actually need to check this? We don't move either page, only edit it.

skins/oasis/modules/TemplateDraftModuleController.class.php
```diff
@@ -28,8 +28,17 @@ public function executeApprove( $params ) {
...
 
-		$this->draftUrl = $this->app->wg->Title->getFullUrl( [
-			'action' => 'approvedraft'
-		] );
+		$this->allowApprove = false;
+		$draftTitle = $this->wg->Title;
+		$parentTitle = TemplateDraftHelper::getParentTitle( $draftTitle );
+
+		if ( $draftTitle->userCan( 'edit' ) && $draftTitle->userCan( 'move' )
+			&& $parentTitle->userCan( 'edit' ) && $parentTitle->userCan( 'move' )
```
Grunny: Same here, I'm not sure why we're checking the `move` permission.

skins/oasis/modules/templates/TemplateDraftModule_Approve.php
```diff
@@ -1,9 +1,13 @@
 <section class="templatedraft-module module">
 	<h2><?= wfMessage( 'templatedraft-module-title-approve' )->escaped() ?></h2>
 	<p><?= wfMessage( 'templatedraft-module-content-approve' )->escaped() ?></p>
-	<a href="<?= $draftUrl ?>">
-		<button class="templatedraft-module-button" type="button" data-id="templatedraft-module-button-approve">
-			<?= wfMessage( 'templatedraft-module-button-approve' )->escaped() ?>
-		</button>
-	</a>
+	<?php if ( $allowApprove ): ?>
+		<a href="<?= $draftUrl ?>">
```
Grunny: As a best practice, `Sanitizer::encodeAttribute` should be used here, even though it's URL encoded. Additionally HTML escaping provides correctness to the output. And though it's very unlikely in this case, what if someone changed how `$draftUrl` was constructed and suddenly a parameter or another part of the URL that a user controls isn't URL encoded? Better to always escape unless you need it to not be, which is why templating languages like Mustache with `{{}}` provide advantages due to most people defaulting to the escaped version without extra work.

Grunny:
LGTM in general once the permissions issue is addressed.

extensions/wikia/TemplateDraft/ApprovedraftAction.php
Grunny: And actually, this would likely break functionality if a template was `move` protected, but not `edit` protected (which is somewhat common), as then a user wouldn't be able to approve a draft for template they can edit, just because they can't move it.

trevorbolliger: yes, if a template is only move protected, a non-admin should be able to approve.

However, if this is very complicated to solve for, we should open another ticket and solve it later (if it ever becomes a real-world problem for users.)

PR #8579, opened 2015-09-23
(CE-2642) Move the creation of achievements tables out of CreateWiki

Grunny:
Move the creation of the achievements tables out of the CreateWiki flow
and instead create them if needed when the extension is enabled. This
follows the same approach [we use for AbuseFilter](https://github.com/Wikia/app/blob/106d8bcf1da2b/extensions/wikia/WikiFactoryChangedHooks/WikiFactoryChangedHooks.class.php#L154-L194), and is less prone to
errors.

/cc @Wikia/community-engineering @owend


owend:
Yep, LGTM :crocodile:

PR #8586, opened 2015-09-24
QueryPage UnwatchedpagesPage - fix class autoloader key

ghost:
`QueryPage` used wrong class autoloader key and the `UnwatchedpagesPage` could not have been loaded. The thing is, class names in PHP are case-insensitive whereas autoloader keys are not.

This is a quick fix for https://wikia-inc.atlassian.net/browse/PLATFORM-1514, but we really should fix the autoloader in https://wikia-inc.atlassian.net/browse/PLATFORM-1516.

@macbre @wladekb @owend - you are going to love it :-)


macbre:
Wunderbar!

PR #8587, opened 2015-09-24
ADEN-2462 Limit OX bidder slots

drets:


extensions/wikia/AdEngine/AdEngine2ContextService.class.php
```diff
@@ -66,7 +66,8 @@ public function getContext( Title $title, $skinName ) {
 					'pageName' => $title->getPrefixedDBKey(),
 					'pageType' => $wikiaPageType->getPageType(),
 					'sevenOneMediaSub2Site' => $wg->AdDriverSevenOneMediaOverrideSub2Site,
-					'skin' => $skinName,
+					// TODO: determine mercury skin in the right way
+					'skin' => $skinName === 'oasis' ? 'oasis' : 'mercury',
```
fraszczakszymon: This change will break wikiamobile. Are we ok with that?

drets: ping @nandy-andy 
2 times duplicate this logic in 2 different JS modules VS broken ads on `Wikiamobile`
do you have another idea?

nandy-andy: Other idea: silly module reused in those two different JS modules which has this silly logic ;) But before we implement it: do we know what will get broken in `wikiamobile` skin?

fraszczakszymon:
One minor note:
Please change this line:
https://github.com/Wikia/app/pull/8587/files#diff-82abeb1fa36bc9a730c90a9c4f4476e1R53
to
`} else if (skin === 'mercury') {`
because now we can pass there whatever and it may not work

LGTM :+1:

PR #8589, opened 2015-09-24
ADEN-2425 Fixed hops with SP recovery

fraszczakszymon:
After launching recovery slots have new DOM id and our slotTweaker was working on old ids. Here is the fix.


nandy-andy:
LGTM :+1:

jacekjursza:
LGTM

PR #8592, opened 2015-09-24
bumping preference comparison to 100%

nmonterroso:
@Wikia/services-team


drsnyder:
Let's go all the way.

PR #8593, opened 2015-09-24
Serve Optimizely script from original CDN for devbox and sandbox

Warkot:
On prod env Optimizely script is loaded by default from our CDN (http://community.wikia.com/wikia.php?controller=Optimizely&method=getCode). It's periodically downloaded by cron running `/maintenance/wikia/cronjobs/fetchOptimizelyScript.php` script (every 10 minutes as I recall and according to https://one.wikia-inc.com/wiki/Engineering/ABTesting/Optimizely#Technical_implementation). There's no such job for devboxes — developer has to know to set `$wgOptimizelyLoadFromOurCDN` to `false` in `DevboxSettings.php` or run the script manually using a command like this:

```
/usr/wikia/backend/bin/run_maintenance --script=fetchOptimizelyScript.php --ip=/usr/wikia/source/app/maintenance/wikia/cronjobs/ --id=80433
```

This is counter intuitive and adds redundant complexity.

Nevertheless, if you make changes to an experiment in Optimizely, you would probably want to see the changes ASAP on the sandbox where you're testing it. ASAP in this case is after clicking save in Optimizely app and after propagation in their CDN, without the need to wait for coping the script to Wikia's CDN.

By the way, sandboxes are serving Optimizely code meant for production — testing the experiments on sandboxes then could pollute the production Optimizely script, which should be kept as thin as possible and not change to often to encourage caching. But that could be a subject of a different PR preceded by a long discussion. :wink:

### DIFF
--- a/extensions/wikia/Optimizely/Optimizely.class.php
+++ b/extensions/wikia/Optimizely/Optimizely.class.php
@@ -64,17 +64,6 @@ class Optimizely {
 		return true;
 	}
 
-	/**
-	 * Is current environment considered a developemnt/testing environment for Optimizely experiments?
-	 *
-	 * @return bool
-	 */
-	public static function isOptimizelyDevEnv() {
-		global $wgWikiaEnvironment;
-
-		return in_array( $wgWikiaEnvironment, static::OPTIMIZELY_DEV_ENVIRONMENTS );
-	}
-
 	protected static function loadFromOurCDN() {
 		$scriptDomain = WikiFactory::getLocalEnvURL(
 			WikiFactory::getVarValueByName( 'wgServer', Wikia::COMMUNITY_WIKI_ID )
@@ -87,5 +76,16 @@ class Optimizely {
 		global $wgOptimizelyUrl, $wgOptimizelyDevUrl;
 		// do not async - we need it for UA tracking
 		return '<script src="' . ( static::isOptimizelyDevEnv() ? $wgOptimizelyDevUrl : $wgOptimizelyUrl ) . '"></script>';
+	}
+
+	/**
+	 * Is current environment considered a developemnt/testing environment for Optimizely experiments?
+	 *
+	 * @return bool
+	 */
+	protected static function isOptimizelyDevEnv() {
+		global $wgWikiaEnvironment;
+
+		return in_array( $wgWikiaEnvironment, static::OPTIMIZELY_DEV_ENVIRONMENTS );
 	}
 }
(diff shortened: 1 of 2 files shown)


maintenance/wikia/cronjobs/fetchOptimizelyScript.php
```diff
@@ -9,9 +9,9 @@
 
 require_once( dirname( __FILE__ ) . '/../../commandLine.inc' );
 
-global $wgDevelEnvironment, $wgOptimizelyUrl, $wgOptimizelyDevUrl;
+global $wgOptimizelyUrl, $wgOptimizelyDevUrl;
 
-$url = $wgDevelEnvironment ? $wgOptimizelyDevUrl : $wgOptimizelyUrl;
+$url = Optimizely::isOptimizelyDevEnv() ? $wgOptimizelyDevUrl : $wgOptimizelyUrl;
```
mklucsarits: Since you're already making a call to the Optimizely class, and since this ternary statement is in two places now, maybe the ternary could be part of an `Optimizely::getOptimizelyUrl` method?

mklucsarits:
LGTM :ship:

vforge:
LGTM

extensions/wikia/Optimizely/Optimizely.class.php
```diff
@@ -6,6 +6,14 @@
  *
  */
 class Optimizely {
+
+	/**
+	 * Which environments are considered a developemnt/testing environments for Optimizely experiments?
+	 * Optimizely Wikia Dev project script should be served also on sandboxes, to prevent poluting Wikia Prod project
+	 * script with experiments that are still under developement.
+	 */
+	const OPTIMIZELY_DEV_ENVIRONMENTS = [ WIKIA_ENV_DEV, WIKIA_ENV_SANDBOX ];
```
kvas-damian: Sandboxes are using production accounts, DBs, etc. When we introduce exception like this we make Sandbox environment inconsistent.
Finally mercury uses production Optimizely account on sandbox which makes things even more complicated. https://github.com/Wikia/mercury/blob/dev/config/localSettings.base.ts#L50

Warkot: I didn't see that recent change in Mercury. :open_mouth:  Then it would be consistent to serve "Wikia [production]" Optimizely project on sandboxes on Oasis, too. I'll just cut the PR to just serve the code on devbox and sandbox from original CDN. Thanks for the insights! :bow:

Warkot:
@kvas-damian I cropped this PR to just serve the code from original CDN for devboxes (dev experiments) and sandboxes (prod experiments – as it used to be). What do you think about it now?

kvas-damian:
LGTM

vforge:
LGTM

PR #8595, opened 2015-09-25
Added support for proper handling of <gallery> and <tabber> tags

inez:
Support for `<tabber>`, `<gallery>` and `<group layout="slider">` tags inside portable infobox.


inez:
Hey @idradm It's not a finished code yet, but I would like you to take a look at it - I think it expresses the approach @dianafa and I took pretty well.

extensions/wikia/PortableInfobox/services/PortableInfoboxRenderService.class.php
```diff
@@ -154,6 +154,15 @@ private function renderInfoboxHero( $data ) {
 	private function renderItem( $type, array $data ) {
 		$helper = new PortableInfoboxRenderServiceHelper();
 
+		if($type === 'image' && count($data) === 1) {
+			$data = $data[0];
```
inez: Dude, really?

extensions/wikia/PortableInfobox/services/PortableInfoboxRenderService.class.php
```diff
@@ -154,6 +154,15 @@ private function renderInfoboxHero( $data ) {
...
+		} else if($type === 'image' && count($data) > 1) {
+			return $this->renderItem(
```
inez, line 163: More than one return statement in a function is a redflag - rethink this approach please.

extensions/wikia/PortableInfobox/services/Parser/Nodes/NodeImage.php
```diff
@@ -10,46 +10,79 @@ class NodeImage extends Node {
...
+					for( $i = 0; $i < count( $tabberOut[0] ); $i++ ) {
+						$tabberImages[] = array('key' => $tabberOut[2][$i], 'caption' => $tabberOut[1][$i]);
+					}
+				}
+				$this->data = [];
+				$images = $galleryImages + $tabberImages;
+				for( $i = 0; $i < count( $images ); $i++ ) {
+					$this->data[] = $this->getImage(
+						$images[$i]['key'],
+						$images[$i]['caption'],
```
inez, line 51: Passing same value twice (caption), think about it.

extensions/wikia/PortableInfobox/services/Parser/Nodes/NodeImage.php
```diff
@@ -10,46 +10,79 @@ class NodeImage extends Node {
...
+				$galleryImages = array();
+				if ( preg_match('#\sdata-model="([^"]+)"#', $parsed, $galleryOut) ) {
+					$galleryImages = array_map(function($gallery) {
+						return array('key' => $gallery['dbKey'], 'caption' => $gallery['caption']);
+					}, json_decode(htmlspecialchars_decode($galleryOut[1]), true));
+				}
+
+				// TODO: Move to separated function + error handling
+				$tabberImages = array();
+				if ( preg_match_all('/class="tabbertab" title="([^"]+)".*?\sdata-image-key="([^"]+)"/is', $parsed, $tabberOut) ) {
```
inez: I hope you gonna have interns write unit tests for it ;)

extensions/wikia/PortableInfobox/services/Parser/Nodes/NodeImage.php
```diff
@@ -10,46 +10,79 @@ class NodeImage extends Node {
...
+		$title = $this->getImageAsTitleObject( $key );
+		$file = $this->getFilefromTitle( $title );
+		if ( $title instanceof \Title ) {
+			$this->getExternalParser()->addImage( $title->getDBkey() );
+		}
+		$ref = null;
+		wfRunHooks( 'PortableInfoboxNodeImage::getData', [ $title, &$ref, $caption ] );
+		$item = [
+			'url' => $this->resolveImageUrl( $file ),
+			'name' => ( $title ) ? $title->getText() : '',
```
inez, line 23: I don't think those brackets around $title are making things more readable. Thoughts?

extensions/wikia/PortableInfobox/services/Parser/Nodes/NodeImage.php
```diff
@@ -10,46 +10,79 @@ class NodeImage extends Node {
...
-				'alt' => $alt,
-				'caption' => $caption,
-				'ref' => $ref
-			];
-
-			if ( $this->isVideo( $file ) ) {
-				$this->data = $this->videoDataDecorator( $this->data, $file );
+			$imageDataLower = strtolower($imageData);
+
+			if ( strpos( $imageDataLower, '-tabber-' ) !== false || strpos( $imageDataLower, '-gallery-' ) !== false ) {
```
idradm, line 40: Parsing HTML output carries a lot of issues with it, for example gallery produces different output depending on params used (slider, slideshow, etc.). However by using PortableInfoboxDataBag we can get wikitext, additionally we could refactor galleries to extract parsing logic and reuse it. That way we are immune to gallery parsing changes and we don't need to make so many regexes and additional parsing. The same could be applied to tabber extension. For inside template usage it could go the same flow, as raw wikitext is provided in template. What do you think about that?

inez, line 40: Thank you for your feedback.

I think code could be way cleaner if we had parsers for galleries and tabbers that parse them to some sort of intermidate data structures instead of HTML right away. Unfortunetly those parsers currently do not exists and creating them would have extra, possibly high, cost - I think it's important to point out here that such parser would be way easier to create for gallery [which is just list of images with captions] but in case of tabber situation is more complicated - cause tabber can have any wikitext inside it.

Given the situation: nor galleries nor tabber is under active development, and feature we are developing is "experimental" I suggest we stick to taken approach and review it in the future depending on the future shape of this feature. In order to become more "immune" to gallery parsing changes we should implement solid unit tests.

Thoughts?

Btw. approach with PortableInfoboxDataBag is interesting, but note that it doesn't work in case when `<gallery>` (or `<tabber>`) is wrapped in `<center>` for instance, e.g.: `image=<center><tabber>...</tabber></center>` because then in Node you get `<center>UNIQUE...</center>` as a parameter value (instead of just the `UNIQUE...`) so you can't retrieve it from the data bag (at least not without extra processing).

extensions/wikia/PortableInfobox/services/Parser/Nodes/NodeImage.php
dianafa, line 51: I don't think is even needed, AFAIK we do not support alts in galleries nor tabbers, do we? So we can care about caption only

extensions/wikia/PortableInfobox/services/Parser/Nodes/NodeImage.php
```diff
@@ -76,6 +97,40 @@ private function getImageAsTitleObject( $imageName ) {
...
 
+	private function getGalleryItems( $imageData ) {
+		preg_match( '/.(UNIQ.*QINU)./U', $imageData, $galleryMarkers );
+		$galleryImages = array();
+
+		foreach ( $galleryMarkers as $marker ) {
+			$galleryWikitext = PortableInfoboxDataBag::getInstance()->getGallery( $marker );
+			$galleryLines = preg_split('/\r\n|\n|\r/', $galleryWikitext);
+
+			foreach ( $galleryLines as $galleryImage) {
```
inez, line 108: @dianafa I sent you couple of messages on slack regarding this change (based on conversations we had before) - but high level summary is: for now I'm sticking to regexes instead of creating another yet parser for galleries. Ultimately it would be good to have one solid parser for galleries and let's talk about it during some sort of retrospective - after we release this project for our beta communities.

extensions/wikia/PortableInfobox/services/Parser/Nodes/NodeImage.php
sebastianmarzjan, line 40: Nice find! Could you file it as a separate task for potential improvement of the class in question?

extensions/wikia/PortableInfobox/services/Parser/Nodes/NodeImage.php
dianafa, line 108: @idradm @inez Good proposition is to reuse the existing gallery parser which will save us lot of regexing and reduce costs of future maintenance in case of any HTML changes in any of multiple gallery types we support. Fast implementation here: https://github.com/Wikia/app/commit/63effc2c54019bf006c3310f93e13d8552ff5700

inez, line 108: For now let's not spend more time on it - as we discussed - I'm moving forward with regex based implementation and creating a follow up ticket to come up later with "better" approach (we will discuss it more during retrospective).

extensions/wikia/PortableInfobox/services/Parser/Nodes/NodeImage.php
```diff
@@ -10,43 +10,67 @@ class NodeImage extends Node {
...
+				$images = $galleryImages + $tabberImages;
+				for( $i = 0; $i < count( $images ); $i++ ) {
+					$this->data[] = $this->getImage(
+						$images[$i]['key'],
+						$images[$i]['caption'],
+						$images[$i]['alt']
+					);
+				}
+			} else {
+				$this->data = array( $this->getImage(
```
inez, line 56: There was no feedback about changing NodeImage to return array of images instead of just one image so I'm assuming all interested people are ok with it.

extensions/wikia/PortableInfobox/services/Parser/Nodes/NodeImage.php
sebastianmarzjan, line 108: I recommend incorporating feedback at this stage. The more code we can keep on the gallery extension parsing logic the better. Still, follow-up will be needed, but is there a rationale not to use the code that's already there?

inez:
Closing this PR (it was just a prototype) and moving proper development to DAT-3320 branch.

extensions/wikia/PortableInfobox/services/Parser/Nodes/NodeImage.php
inez, line 108: @SebastianMarzjan Yes. Feedback been incorporated. Btw. This is not extending gallery parsing logic - it's another yet parser for gallery.

PR #8596, opened 2015-09-25
CF_Http: PLATFORM-1521 - report an error only when there are no retries left

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-1521

Errors reported so far shows that in 99% of cases one retry is enough.

@michalroszka


ghost:
:+1:

PR #8597, opened 2015-09-25
ADEN-2475 Stabilize adTracker unit tests

fraszczakszymon:


extensions/wikia/AdEngine/js/spec/AdTracker.spec.js
```diff
@@ -253,7 +253,7 @@ describe('ext.wikia.adEngine.adTracker', function () {
 
 		timer = adTracker.measureTime('test/event', {abc: 'def', xyz: 123});
 
-		windowMock.wgNow = new Date(new Date().getTime() - 999);
+		windowMock.wgNow = new Date(windowMock.wgNow.getTime() - 110);
```
nandy-andy, line 256: I know it's not your code but can we somehow get rid off magic numbers? It took me a while to understand why `110` ;) And I'm still not sure if I got it correctly >.< Maybe we can name it somehow and define it at the beginning of the file together with `888`. Or maybe we don't need `888`... then we could leave here a number with a comment maybe... What do you think?

nandy-andy, line 256: Oh, OK. We discussed it offline. And in the assertions below we can find `110` and `888` as well, so it seems the author came up with the numbers. Let's leave it as it is.

nandy-andy:
LGTM :+1:

PR #8599, opened 2015-09-25
CE-2703: Add possibility to import external pages using subdomain or host

lukaszkonieczny:
Added possibility to import external scripts by passing wiki subdomain or host using "url" prefix, ie.

`url:muppet:MyScript.js` or `url:muppet.wikia.com:MyScript.js`

The import method is moved to separate amd module.


resources/wikia/modules/importScriptHelper.js
```diff
@@ -0,0 +1,64 @@
...
+		return prefix === externals.db;
+	};
+
+	isExternalDomain = function(prefix, domain) {
+		return prefix === externals.domain && hasWikiaDomain(domain);
+	};
+
+	hasWikiaDomain = function(scriptName) {
+		if (scriptName.indexOf('.') !== -1) {
+			return scriptName.substr(scriptName.length - wikiaDomain.length) === wikiaDomain;
```
Grunny: I'd include the period in the `wikiaDomain` as this would return true for something like `notreallywikia.com` (which you could add as a test case).

lukaszkonieczny: awesome catch :)

Grunny:
LGTM.

resources/wikia/modules/importScriptHelper.js
```diff
@@ -0,0 +1,60 @@
...
+	function isExternalDb(prefix) {
+		return prefix === externals.db;
+	}
+
+	function isExternalDomain(prefix, domain) {
+		return prefix === externals.domain && hasWikiaDomain(domain);
+	}
+
+	function hasWikiaDomain(scriptName) {
+		if (scriptName.indexOf('.') !== -1) {
```
Grunny: What about non-English subdomains? i.e. `url:ja.starwars:Foo.js`? Right now, only single subdomains are possible without being the full domain.

Grunny:
:+1:

resources/wikia/modules/importScript.js
```diff
@@ -0,0 +1,52 @@
...
+			articlesFailed = [],
+			scriptName;
+
+		if (!$.isArray(articles)) {
+			articles = [articles];
+		}
+
+		for (var i = 0; i < articles.length; i++) {
+			if (!importScriptHelper.isJsPage(articles[i])) {
+				articlesFailed.push(namespacePrefix + articles[i]);
```
adamkarminski: Shouldn't it be `importScriptHelper.getNamespacePrefix()`?

resources/wikia/modules/importScript.js
```diff
@@ -0,0 +1,52 @@
...
+				continue;
+			}
+			if (importScriptHelper.isLocal(articles[i])) {
+				articlesToImport.push(importScriptHelper.getNamespacePrefix() + articles[i]);
```
adamkarminski, line 32: To consider - what about accepting not only scripts in a form of `Common.js` but also with the namespace prefix: `MediaWiki:Common.js`? After all, it is still a valid article name.

adamkarminski:
Tested, works like a charm. However, please consider the 2 inline comments.

resources/wikia/modules/importScript.js
lukaszkonieczny, line 32: I will leave it as it is now. We have more control right now. If we would like to move scripts to different place or namespace while working on scripts library in the future, we just need to change one variable in the code and we don't need to make any change in each script which will use `importWikiaScriptPages`. Also this will add more complexity to the logic here and will require testing it one more time.

adamkarminski, line 32: In this case, I'd consider changing the user-facing message that currently is: `{script name} was not loaded as it's not a JS page (requested by user-supplied javascript). Please make sure the article title has .js extension.` which is misleading.

PR #8600, opened 2015-09-25
ADEN-2468 OX bidder on mobile

fraszczakszymon:
OX bidder enabled on mercury + new queue with callbacks which starts on mercury load (after setting ad context).


jacekjursza:
lgtm

PR #8604, opened 2015-09-25
account for local preferences in Preferences::getUserPropertyHelper

nmonterroso:
@Wikia/services-team 

properly read from locals when filling out existing values on `Special:Preferences`

### DIFF
--- a/includes/Preferences.php
+++ b/includes/Preferences.php
@@ -41,10 +41,10 @@ class Preferences {
 	);
 
 	/**
-	 * @return UserPreferences
+	 * @return PreferenceScopeService
 	 */
 	static function preferenceScope() {
-		return Injector::getInjector()->get(PreferenceScopeService::class);
+		return Injector::getInjector()->get( PreferenceScopeService::class );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


includes/Preferences.php
```diff
@@ -139,12 +139,19 @@ static function getUserPreference( $name, $info, $user ) {
...
+		if ( in_array( $property, self::getAttributes() ) ) {
+			return $user->getGlobalAttribute( $property );
+		} elseif ( $scopeService->isGlobalPreference( $property ) ) {
+			return $user->getGlobalPreference( $property );
+		} elseif ( $scopeService->isLocalPreference( $property ) ) {
+			list( $prefName, $wikiId ) = $scopeService->splitLocalPreference( $property );
+			return $user->getLocalPreference( $prefName, $wikiId );
 		}
+
+		return null;
```
nmonterroso, line 154: When doing this I noticed some non-attributes and non-preferences got here, specifically:

```
emailauthentication
username
usergroups
emailaddress
password
oldsig
fbconnect-disconnect
adoptionmails-label-<some_number>
```

but I did a search on the database and none of them has any entries, so they're always returning `null` for any user, except `username` which is something that I spoke to @jsutterfield about, and it seems ok for this to return `null` in that case as well.

drsnyder:
lgtm

PR #8605, opened 2015-09-25
instead of creating a new user object, create a mock

nmonterroso:
@Wikia/services-team 

When running tests with `$wgPreferenceServiceRead = true`, these tests failed because we don't allow an anonymous user to have preferences in the preference service (but mediawiki allows it for the `User` object), so this changeset changes this test to mock the user object rather than use it directly, which is probably better anyway


drsnyder:
lgtm

PR #8609, opened 2015-09-28
CE-2681 Add a unit test for removeCompletedAndUpdateLogs method

adamkarminski:

Ok, so this one is a really nasty one. It is complex, long and REALLY convinced me that we should extract the actual logic from the API controller to a service/class/whatever.

IMO it is pointless to waste time right now on trying to retrieve what is actually the outcome of the method (which is saving a notification message to `$this->notification`) in a test. I think that an API test should focus on request validation, permissions etc. They are being checked right now. The rest should go into the service test.

Please, review my point and code.

/cc @Grunny @lukaszkonieczny

### DIFF
--- a/extensions/wikia/ContentReview/ContentReviewHelper.php
+++ b/extensions/wikia/ContentReview/ContentReviewHelper.php
@@ -227,7 +227,7 @@ class Helper extends \ContextSource {
 			return true;
 		}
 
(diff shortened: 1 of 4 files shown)


adamkarminski:
reping @Grunny @lukaszkonieczny :)

Grunny:
LGTM. Agreed on APIs, they should generally be tested as more of a kind of integration test (i.e. if I pass these parameters, I get this response), the actual work of individual units should be elsewhere (and so tested elsewhere).

PR #8617, opened 2015-09-29
SOC-926 Load localized Facebook JS SDK

armonr:
/CC @lizlux @bkoval @jsutterfield 

Previously, all Facebook plugins on Wikia pages have been rendered in English because the SDK URL had been hardcoded to point to the English version. This changeset aims respect the language of the Wikia the SDK loads for; as a result, all FB plugins on a page are properly localized to the same language.
Facebook documentation on supported languages can be found [here](https://developers.facebook.com/docs/internationalization#locales). Also both country code and language code are required to get the locale right.

Update: Geo cookie is being used now to try and match the user's location with Wikia's language if possible.

For more background info, please see [SOC-926](https://wikia-inc.atlassian.net/browse/SOC-926)

### DIFF
--- a/resources/wikia/modules/facebookLocale.js
+++ b/resources/wikia/modules/facebookLocale.js
@@ -132,7 +132,7 @@
 		function getSdkUrl(langCode) {
 			var lowerLangCode = langCode.toLowerCase(),
 				countryCode = codes[langCode];
-			if (langCode.length !== 2 || countryCode === undefined) {
+			if (langCode.length !== 2 || !countryCode) {
 				lowerLangCode = 'en';
 				countryCode = 'US';
 			}
--- a/resources/wikia/modules/facebookLocale.js
+++ b/resources/wikia/modules/facebookLocale.js
@@ -2,10 +2,15 @@
 (function (context) {
 	'use strict';
 
+	/**
+	 * Determines the locale for a Facebook user based on Wikia language and user's location
+	 *
+	 * Facebook languages/countries supported: https://www.facebook.com/translations/FacebookLocales.xml
+	 **/
 	function facebookLocale() {
-		/** https://www.facebook.com/translations/FacebookLocales.xml
-		 *  Note: only one code per language is used. **/
-		var languageCodes = {
+		/** Note: only one default country per language is used. */
+		var defaultCountryCodes = {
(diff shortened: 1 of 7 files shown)


bkoval:
Seems like unit tests failure is connected to this pull request, please take a look at the Jenkins build :santa:

resources/wikia/modules/facebookLocale.js
```diff
@@ -0,0 +1,159 @@
...
+			'yo': 'NG',
+			'zh': 'CN',
+			'zu': 'ZA',
+			'zz': 'TR'
+		};
+
+		function getSdkUrl(langCode) {
+			var lowerLangCode = langCode.toLowerCase(),
+				countryCode = codes[langCode];
+			if (langCode.length !== 2 || countryCode === undefined) {
```
bkoval: I think !countryCode would be a sufficient check here, since all the keys in the lookup object are 2 chars long.

resources/wikia/modules/facebookLocale.js
```diff
@@ -0,0 +1,159 @@
...
+	//UMD inclusive
+	if (!context.Wikia) {
+		context.Wikia = {};
+	}
+
+	//namespace
+	context.Wikia.fbLocale = facebookLocale(context);
+
+	if (context.define && context.define.amd) {
+		context.define('wikia.fbLocale', ['wikia.window'], facebookLocale);
```
bkoval: Are both Wikia.fbLocale and AMD module needed at the same pageview? Is there a chance both forms will be used on one pageview?

bkoval:
Minor stuff. Overall it looks nice!

mklucsarits:
Aside from failing tests, LGTM

resources/wikia/modules/facebookLocale.js
armonr: I'm not quite sure if it is needed, but I tried to follow existing patterns such as https://github.com/Wikia/app/blob/dev/resources/wikia/modules/cookies.js#L146-148. Is there a better way to go about this?

bkoval: I get it. Future-proof JS for the win! Let's leave it as it is :)

lizlux:
:+1:

PR #8624, opened 2015-09-29
SEO-58 Block Template pages from being crawled

gabrys:
Template pages should be tagged as noindex; these pages are not a major problem, but Google is seeing these as soft 404s.


saipetch:
LGTM

PR #8626, opened 2015-09-29
better preference profiling

nmonterroso:
@Wikia/services-team 

Separate service and consul service discovery profiling so we get a better idea of where time is being spent when getting/setting preferences


drsnyder:
lgtm

PR #8628, opened 2015-09-29
PLATFORM-1513: user service does not need separate user object cache

owend:
- removed user object cache from user service (these objects are already cached)
- cleaned up some space/tab issues in User.php from wikia code additions (use &w=1 to ignore those)
- added profiling to Helios::newFromToken function
- backported performance fix from MW1.25 for User::getDefaultOptions (cache using static var)

@drsnyder @ArturKlajnerok


drsnyder:
LGTM /cc @nmonterroso

nmonterroso:
LGTM

PR #8634, opened 2015-09-30
Automate detection of cases where revision is inserted, but the edit token is not checked

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-1540
- introduce `UserMatchEditToken` hook
- set a flag when `User::matchEditToken` is called
- on revision insert (bind to `RevisionInsertComplete` hook) assert that the edit token has been checked, if not - log a report

Feel free to suggest more hooks to bind this check to :)

@Grunny / @michalroszka

### DIFF
--- a/extensions/wikia/Security/Security.setup.php
+++ b/extensions/wikia/Security/Security.setup.php
@@ -1,0 +1,16 @@
+<?php
+
+/**
+ * This is a set of tools that automates the reporting of potential security problems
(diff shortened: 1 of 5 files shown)


Grunny:
:+1:

macbre:
Backported to preview in 8d96cb3

ghost:
:+1:

PR #8301, opened 2015-09-01
VOLDEV-145 - Implement article header changes and new rail module

onei:
**Awaiting review by VolDevs**

Ticket: https://wikia-inc.atlassian.net/browse/VOLDEV-145

**Changes:**
- Move edit buttons (and other buttons) back above the page header border. This also fixes CONCF-501.
- Move the page count tally to it's own sidebar module, and move the contribute button to said module as well. Special:Videos keeps it's own tally (and no longer has the page share buttons).

**Cleanup includes:**
- Fixed various code issues, e.g. `array()` -> `[]`, `wfMsg*` -> `wfMessage`, whitespace, etc. to conform with the style guides.
- Altered the `PageHeaderIndexAfterActionButtonPrepared` and `PageHeaderIndexExtraButtons` hooks to only modify the intended variables to reduce the potential for subtle, undocumented behaviour/bugs.
- Moved extension specific buttons to hook callbacks in `PageHeader_Index.php` (per a TODO comment), thus adding `SpecialVideosHooks.php` and `WikiaNewFilesHooks.php`.
- Added icons to the user talk archive edit button (seen when Wall is enabled).
- Fixed inconsistencies in `DefaultSettings.php` (simply picked what seemed to be the more predominant method).
- Removed unused sass, e.g. removing styling related to search in the page header.
- Removed use of globals in controller methods in favour of accessing them via `F::app()` (or `$this` if non-static.
- Overhauled logic for `PageHeaderContoller` and it's templates.
- Prevented modification of URLs for the share buttons in the page header for hub wikis (always use database defaults).
- Deleted `PageHeader_EditBox.php` as it was unused (and it's associated method).

**To be fixed before merging:**
- Remove temp code in `PageHeaderController::executeHubs` (sass loading) and `ContributeController::index` (commented out code for testing wikis with high page counts)
- Move the new sidebar module to the correct place in `BodyController::getRailModuleList` (awaiting confirmation of how it's meant to behave in ticket).
- Rebase to fix conflicts (largely related to curatedcontent).

PR #8635, opened 2015-09-30
XW-396 | VOLDEV-145 - Implement article header changes and new rail module

Warkot:
:hand: :cop: :warning: **WIP — do not merge (yet)!**

Original PR: https://github.com/Wikia/app/pull/8301

XW ticket: https://wikia-inc.atlassian.net/browse/XW-396
Original VOLDEV ticket: https://wikia-inc.atlassian.net/browse/VOLDEV-145


onei:
@Warkot I've fixed the merge conflicts with #8301. I don't know which is easier for you to test with, but if it happens again feel free to give me a nudge and I'll sort it out asap :)

Warkot:
@onei Awesome! :+1: Thank you, I'll grab it then. :)

Warkot:
@onei I've checked out the original branch after your merge (commit 81408cfbe4daba83f6645ed8c6d56fe1b1d8132c) on my devbox, but I'm afraid it doesn't work the same way as it didn't work before my hotfix (commit 954650e1307da25b8e0bc40b231f32a677a877f7). → `/usr/wikia/source/app/includes/wikia/nirvana/WikiaView.class.php:131 Template file not found: /usr/wikia/source/app/skins/oasis/modules/templates/ContributeController_index.php`, so it's not the problem with merge. :disappointed:

onei:
The merge conflict fix was to allow it to be merged automatically rather than fix a bug, however that's not an error I've come across during testing. I'm limited to my vagrant devbox, but it looks like you're using a different setup judging by the path in the error which might possibly explain it? Having said that I'm running it on windows which has a habit of adding interesting quirks so I'm not too sure.

onei:
@Warkot I think I've figured out the issue - Windows considers file names to be case insensitive which somehow spilled over to my devbox. I've renamed the template, so hopefully it should be working correctly now.

Warkot:
@onei Your fixes helped to resolve the issue — I ported them here and reverted my hotfix. :+1: But now I'm trying to track the problem why the content of the page (i.e. of the `<div id="WikiaArticle" …>`) is not loading — do you think it may share the root of the cause with the previous one, the case (in)sensitivity?

onei:
@Warkot That one I can reproduce, but it seems to be fixed by a null edit - have you tried that? It's a bug I've noticed intermittently while working on something else, but I haven't managed to nail down the cause of it unfortunately.

P.S. If it's easier, we could chat on [Slack](https://wikia.slack.com/team/cam)?

PR #8639, opened 2015-09-30
XW-287 | Removing user language from wiki variables call

bkoval:
https://wikia-inc.atlassian.net/browse/XW-287 

User lang will be fetched through UserInfo call https://github.com/Wikia/mercury/pull/1465

There is already a ticket (https://wikia-inc.atlassian.net/browse/XW-423) to implement user language properly in Mercury. WikiVariables won't be involved.

### DIFF
--- a/extensions/wikia/MercuryApi/models/MercuryApi.class.php
+++ b/extensions/wikia/MercuryApi/models/MercuryApi.class.php
@@ -247,9 +247,9 @@ class MercuryApi {
 	 * @return string homepage URL. Default is US homepage.
 	 */
 	private function getHomepageUrl() {
-		global $wgLang;
+		global $wgContLang;
 		if ( class_exists('WikiaLogoHelper') ) {
-			return ( new WikiaLogoHelper() )->getCentralUrlForLang( $wgLang->getCode() );
+			return ( new WikiaLogoHelper() )->getCentralUrlForLang( $wgContLang->getCode() );
 		}
 		return 'http://www.wikia.com'; //default homepage url
 	}
--- a/extensions/wikia/MercuryApi/models/MercuryApi.class.php
+++ b/extensions/wikia/MercuryApi/models/MercuryApi.class.php
@@ -246,9 +246,9 @@ class MercuryApi {
 	 * @return string homepage URL. Default is US homepage.
 	 */
 	private function getHomepageUrl() {
-		global $wgContLang;
+		global $wgLanguageCode;
 		if ( class_exists('WikiaLogoHelper') ) {
-			return ( new WikiaLogoHelper() )->getCentralUrlForLang( $wgContLang->getCode() );
+			return ( new WikiaLogoHelper() )->getCentralUrlForLang( $wgLanguageCode );
 		}
 		return 'http://www.wikia.com'; //default homepage url
 	}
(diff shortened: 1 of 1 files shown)


extensions/wikia/MercuryApi/models/MercuryApi.class.php
```diff
@@ -117,7 +117,6 @@ public function getWikiVariables() {
 			'homepage' => $this->getHomepageUrl(),
 			'id' => (int) $wgCityId,
 			'language' => [
-				'user' => $wgLang->getCode(),
 				'userDir' => SassUtil::isRTL() ? 'rtl' : 'ltr',
```
rogatty: `userDir` also doesn't make sense - response for this method is cached per wiki.

rogatty:
Should we also update `getHomepageUrl` to use wiki language instead of user language? Right now if user with `pl` language set in preferences opens english wiki as the first visitor (wiki not cached yet) then all following users will have link to polish homepage. I didn't confirm it by test but it probably works like that right now.

https://github.com/Wikia/app/blob/XW-287/extensions/wikia/MercuryApi/models/MercuryApi.class.php#L252

extensions/wikia/MercuryApi/models/MercuryApi.class.php
bkoval: I considered that, but it's not trivial - Mercury's backend is making SCSS ltr / rtl decision on backend side. I can create a ticket for investigating if we can use a different way to act there.

extensions/wikia/MercuryApi/models/MercuryApi.class.php
```diff
@@ -248,9 +247,9 @@ private function clearUsers() {
 	 * @return string homepage URL. Default is US homepage.
 	 */
 	private function getHomepageUrl() {
-		global $wgLang;
+		global $wgContLang;
 		if ( class_exists('WikiaLogoHelper') ) {
-			return ( new WikiaLogoHelper() )->getCentralUrlForLang( $wgLang->getCode() );
+			return ( new WikiaLogoHelper() )->getCentralUrlForLang( $wgContLang->getCode() );
```
bognix: you sure about this change? It changes behavior a little bit.
Before: If user was logged in he would be redirected wikia in his lang
Now: It doesn't matter if user is logged in or not he always will be redirected to wikia in content lang.

I'm not sure if it works that way on desktop and I'm not sure if we want to be consistent with desktop but I would like to be sure we are aware of this change and it has some purpose

rogatty: It doesn't work anyway as it's cached for the first user visiting the wiki (for 60 seconds).

bkoval: @bognix We spoke with @rogatty and there's a ticket in the X-Wing backlog for making things straight with the user language. This is a fix for malfunctioning behavior cause by caching the WikiVariables response in iCache.

tomasznapieralski:
LGTM

extensions/wikia/MercuryApi/models/MercuryApi.class.php
rogatty: Here you use `$wgContLang->getCode()` but in `language.content` it's `$wgLanguageCode`. What's the difference? Is it the right thing to do? 

/cc @bognix @vforge

bognix: https://www.mediawiki.org/wiki/Manual:$wgLanguageCode
https://www.mediawiki.org/wiki/Manual:$wgContLang
everything is there

bkoval: Yeah, they both return the same language code, but it makes sense to unify this. I'll use $wgLanguageCode here as well.

vforge:
LGTM

PR #8640, opened 2015-09-30
XW-422 throw exception when method called from non POST request

bognix:
https://wikia-inc.atlassian.net/browse/XW-422

PR in Mercury: https://github.com/Wikia/mercury/pull/1466

### DIFF
--- a/extensions/wikia/CuratedContent/CuratedContentValidatorController.class.php
+++ b/extensions/wikia/CuratedContent/CuratedContentValidatorController.class.php
@@ -17,7 +17,7 @@ class CuratedContentValidatorController extends WikiaController {
 
 	public function validateSection() {
 		global $wgRequest;
-		if (!$wgRequest->wasPosted()) {
+		if ( !$wgRequest->wasPosted() ) {
 			throw new CuratedContentValidatorMethodNotAllowedException();
 		}
 
@@ -36,7 +36,7 @@ class CuratedContentValidatorController extends WikiaController {
 
 	public function validateSectionWithItems() {
 		global $wgRequest;
-		if (!$wgRequest->wasPosted()) {
+		if ( !$wgRequest->wasPosted() ) {
 			throw new CuratedContentValidatorMethodNotAllowedException();
 		}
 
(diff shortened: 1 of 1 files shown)


vforge:
LGTM

extensions/wikia/CuratedContent/CuratedContentValidatorController.class.php
```diff
@@ -15,6 +16,11 @@ public function __construct() {
 	}
 
 	public function validateSection() {
+		global $wgRequest;
+		if (!$wgRequest->wasPosted()) {
```
rogatty: Missing spaces ;)

extensions/wikia/CuratedContent/CuratedContentValidatorController.class.php
```diff
@@ -15,6 +16,11 @@ public function __construct() {
 	}
 
 	public function validateSection() {
+		global $wgRequest;
```
rogatty, line 19: Quoting https://github.com/Wikia/app/blob/dev/includes/wikia/nirvana/WikiaController.class.php#L143:

> if you want to access request, use: RequestContext::getMain()->getRequest()

extensions/wikia/CuratedContent/CuratedContentValidatorController.class.php
vforge: @Warkot

extensions/wikia/CuratedContent/CuratedContentValidatorController.class.php
bognix, line 19: above comment makes sense in context of method it applies to. In this particular case I don't see difference

rogatty, line 19: Well, the difference is that we don't use a global. I don't know if it's better ;) PHP magician needed.

rogatty:
LGTM

extensions/wikia/CuratedContent/CuratedContentValidatorController.class.php
Warkot, line 19: Lately I bumped into this https://github.com/Wikia/app/blob/XW-396/skins/oasis/modules/BodyController.class.php#L121, where `global $wgRequest` was exchanged for `F::app()->wg->Request`. Not using globals sounds right, but I'm just a low level PHP warlock. ;)

PR #8644, opened 2015-09-30
CE-2733 Provide a script to set any wg var to any value based on the …

adamkarminski:
…WAM rank


includes/wikia/services/DataMartService.class.php
```diff
@@ -853,6 +854,45 @@ public static function getWAM200Wikis() {
...
 	}
 
+	/**
+	 * Returns an array of IDs of wikias ordered by WAM rank.
+	 * The default limit is 200. If 0 is provided - no limit is used.
+	 * @param int $limit Default is set as the DEFAULT_TOP_WIKIAS_LIMIT const
+	 * @param array $wikisIds Limit the range to certain wikis
+	 * @return bool|array An array of IDs or `false` on no results
+	 */
+	public function getWikisOrderByWam( $limit = self::DEFAULT_TOP_WIKIAS_LIMIT, array $wikisIds = [] ) {
```
lukaszkonieczny, line 864: Please consider moving this method to `WAMService`

maintenance/wikia/WikiFactoryVariables/setGlobalBasedOnWam.class.php
```diff
@@ -0,0 +1,113 @@
+<?php
+/**
+ * Use this script to enable any global variable on wikias based on their WAM rank.
+ */
+require_once( __DIR__ . '/../../Maintenance.php' );
+
+class setGlobalBasedOnWam extends Maintenance {
```
Grunny: Class names should be UpperCamelCase.

includes/wikia/services/DataMartService.class.php
```diff
@@ -853,6 +854,45 @@ public static function getWAM200Wikis() {
...
+			->cacheGlobal( self::TTL )
+			->SELECT( 'wiki_id' )
+			->FROM( 'dimension_top_wikis' )
+			->ORDER_BY( 'rank' );
+
+		/**
+		 * Limit the results to certain wikias
+		 */
+		if ( !empty( $wikisIds ) ) {
+			$sql->WHERE( 'wiki_id' )->IN( $wikisIds );
```
lukaszkonieczny, line 879: What do you think about adding some max limit for `$wikisIds` elements?

maintenance/wikia/WikiFactoryVariables/setGlobalBasedOnWam.class.php
```diff
@@ -0,0 +1,113 @@
...
+
+	public function __construct() {
+		parent::__construct();
+		$this->addOption( 'limit', 'How many wikis should I change the variable for?', true, true );
+		$this->addOption( 'var', 'What variable should I set?', true, true );
+		$this->addOption( 'varValue', 'What value should I set the variable to?', true, true );
+		$this->addOption( 'dependency', 'Should I consider any other variable\'s value?', false, true );
```
lukaszkonieczny: I'm not sure about naming here. Maybe we can at least add some example of usage.

maintenance/wikia/WikiFactoryVariables/setGlobalBasedOnWam.class.php
```diff
@@ -0,0 +1,113 @@
...
+		if ( !$dry ) {
+			$this->output( "All done! Let's validate the result...\n\n" );
+
+			$success = $fail = [ ];
+
+			foreach ( $wikiIdsToConvert as $wikiId ) {
+				if ( WikiFactory::getVarValueByName( $var, $wikiId ) === $varValue ) {
+					$success[] = $wikiId;
+				} else {
+					$fail[] = $wikiId;
```
lukaszkonieczny, line 94: Maybe we can check, which wikis failed by checking value returned by `WikiFactory::setVarById( $varId, $wikiId, $varValue )` ?

maintenance/wikia/WikiFactoryVariables/setGlobalBasedOnWam.class.php
adamkarminski: Yeah, sorry...

includes/wikia/services/DataMartService.class.php
adamkarminski, line 879: I think that if it is not in our public API, we should rely on engineers sanity when using it. :) You know, we could set a stiff limit to SQL queries to 1000 records as well. :)

maintenance/wikia/WikiFactoryVariables/setGlobalBasedOnWam.class.php
adamkarminski, line 94: I want to verify if all values are actually returned properly, not just the transaction status... I figured out that it could pass in a script that you only run from time to time. :)

maintenance/wikia/WikiFactoryVariables/setGlobalBasedOnWam.class.php
adamkarminski: :+1:

includes/wikia/services/DataMartService.class.php
lukaszkonieczny, line 879: But in our case we don't have control on this array because it's filled by `WikiFactory::getCityIDsFromVarValue` so we can get thousands of ids here.

lukaszkonieczny, line 879: Maybe we can limit it in our maintenance script?

lukaszkonieczny:
LGTM

includes/wikia/services/DataMartService.class.php
adamkarminski, line 879: A comment added to the `dependency` parameter in the maintenance script.

PR #8645, opened 2015-09-30
PLATFORM-1540 assert that User::saveSettings checks edit token

macbre:
Extends #8634

Bind to `UserSaveSettings` to protect us against cases like #5963

Check if `WebRequest::wasPosted` or `WikiaRequest::wasPosted` were called in addition to `User::matchEditToken` (and report the state of these checks when logging CSRF candidates).

@Grunny / @michalroszka


Grunny:
:+1:

macbre:
Backported to preview in 47412dd

PR #8649, opened 2015-09-30
PLATFORM-1457: getDatabase should import city_variables_pool from pro…

owend:
…duction

This fixes the foreign key constraint issue on the city_variables and city_variables_pool.
The down side is that if a variable is created in dev, it might be clobbered by a production variable with the same id.  But in my opinion we should have the production variables migrated to dev on import for consistency.

@macbre @gabrys

### DIFF
--- a/maintenance/wikia/getDatabase.php
+++ b/maintenance/wikia/getDatabase.php
@@ -17,9 +17,6 @@ $dirName = dirname(__FILE__);
 
 require_once( $dirName . "/../commandLine.inc" );
 
-$wgDBdevboxUser = 'devbox';
-$wgDBdevboxPass = 'devbox';
-
 $USAGE =
 	"Usage:\tphp getDatabase.php [[-h [hostname] | [-f [hostname] | -i [filename] | [ -p [env] ]  | -?]\n" .
 	"\toptions:\n" .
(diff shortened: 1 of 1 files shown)


maintenance/wikia/getDatabase.php
```diff
@@ -193,26 +193,36 @@ function getFile($databaseDirectory, $dbname, $filedir) {
...
 		print $response;
 		unlink ("/tmp/city_domains.csv");
-		// dump city_vars rows to local CVS file and import into local database
+
+		// dump city_variables_pool rows to local CVS file and import into local database
+		$response = `mysql -u $wgDBbackenduser -p$wgDBbackendpassword --database wikicities -h $dbhost -ss -e "select * from city_variables_pool where cv_id in (select cv_variable_id from city_variables where cv_city_id = $city_id)" | $prepareCsv > /tmp/city_variables_pool.csv`;
+		print "city_variables_pool dump ok\n";
+		$response = `mysqlimport -u $wgDBdevboxUser -p$wgDBdevboxPass -h $wgDBdevboxCentral --replace --fields-enclosed-by=\\" --fields-terminated-by=, --local wikicities /tmp/city_variables_pool.csv`;
+		print $response;
+		unlink ("/tmp/city_variables_pool.csv");
```
macbre: I think it's a good moment to remove this repeatable dump and import sections and use a function to perform them.

owend: yeah, i almost did that... got lazy.  i'll fix it... :)

nandy-andy:
LGTM :+1:

I agree that we should have production variables migrated to dev on import. Thank you! :smile_cat:

PR #8650, opened 2015-09-30
Platform-1355

owend:
Users are reporting increasing numbers of blank/empty pages in production that are fixed with action=purge.  Just logging errors for now, will merge this tomorrow after the release.

@macbre @michalroszka

### DIFF
--- a/includes/Article.php
+++ b/includes/Article.php
@@ -220,9 +220,7 @@ class Article extends Page {
 			$this->fetchContent();
 			// Wikia: Temporary Investigation for PLATFORM-1355
 			if(empty($this->mContent)) {
-					global $wgCityId;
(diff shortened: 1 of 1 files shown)


includes/Article.php
```diff
@@ -218,6 +218,14 @@ public function getContent() {
 			return $text;
 		} else {
 			$this->fetchContent();
+			// Wikia: Temporary Investigation for PLATFORM-1355
+			if(empty($this->mContent)) {
+					global $wgCityId;
+					Wikia\Logger\WikiaLogger::instance()->error( __METHOD__ . ' empty content PLAT1355', [
+						'wiki_id' => $wgCityId,
```
macbre: city id is always logged as `@fields.city_id` by `WikiaLogger`. No need to add it here.

owend: Thanks! I've seen it logged in other places, so that may be a legacy.  I'll remove it.

PR #8651, opened 2015-10-01
(CE-2809) Remove Flash MP3 player extensions

Grunny:
See ticket for reasons.

/cc @owend


owend:
LGTM, thanks G :koala:

macbre:
:+1:

PR #8654, opened 2015-10-01
(CE-2810) Add a reusable method to check valid write requests

Grunny:
Add a reusable method to check valid write requests to verify that the
request was POSTed and provided a valid edit token. This can be used
in any Nirvana controller to protect against CSRF.

/cc @macbre @adamkarminski


adamkarminski:
Looks splendid to me! :gb:

macbre:
Jolly good :+1:

PR #8660, opened 2015-10-01
CE-2724: Prepare Special:JsPages code

lukaszkonieczny:
We create new Special:JsPages. This page consist list of all js pages on given wiki with their review statuses (latest revision, latest reviewed and live revision).

While this data is now used in two different places, we introduce `ContentReviewStatusesService` which now is also used in content review module in right rail. This required some code refactoring.

To simplify and do not duplicate work part of (or even all) CE-2621 was also done. So permissions to enable test mode changed. From now all logged in users are able to run wiki in test mode. They also will see content review module. To submit revision for review still appropriate rights are required.


extensions/wikia/ContentReview/ContentReview.i18n.php
```diff
@@ -36,6 +36,8 @@
 	'content-review-rejection-explanation' => '==Submitted script change rejected==
 The recently submitted change to this JS script was rejected by the Wikia review process. Please make sure you meet the [[Help:JavaScript review process|Custom JavaScript guidelines]].',
 	'content-review-status-link-text' => 'Review status',
+
+	'content-review-special-js-pages-title' => 'Javascript pages',
```
Grunny: "Javascript" -> "JavaScript". Also should be added to the `qqq` docs.

Grunny:
This seems to be broken when checked out on my devbox. The Special:JsPages doesn't display the pages properly and the right rail module on JS pages stops displaying the revision data when this branch is checked out.

skins/oasis/modules/ContentReviewModuleController.class.php
```diff
@@ -38,31 +32,34 @@ public function executeRender( $params ) {
...
-			self::STATUS_TEMPLATE_PATH, $latestStatus
+			self::STATUS_TEMPLATE_PATH, $pageStatus['latestRevision']
 		) );
-
-		/* Set displaySubmit */
-		$this->setVal( 'displaySubmit', $latestStatus['statusKey'] === self::STATUS_UNSUBMITTED );
+		$this->setVal(
+			'displaySubmit',
+			$pageStatus['latestRevision']['statusKey'] === ContentReviewStatusesService::STATUS_UNSUBMITTED
+			&& $helper->userCanEditJsPage( $this->wg->Title, $this->wg->User )
```
Grunny: Indent this one more level so it doesn't look like another parameter.

skins/oasis/css/modules/ContentReview.scss
```diff
@@ -39,13 +39,16 @@
...
 .content-review-status-unsubmitted {
-	border-color: $content-review-color-unsubmitted;
-	animation: submitReminder 3s infinite;
+  border-color: $content-review-color-unsubmitted;
 }
 
-.content-review-status-awaiting {
-	border-color: $content-review-color-awaiting;
+.WikiaRail .content-review-status-unsubmitted {
+  animation: submitReminder 3s infinite;
```
Grunny: Indentation here and above.

extensions/wikia/ContentReview/controllers/templates/JsPagesSpecialController_index.php
```diff
@@ -0,0 +1,39 @@
...
+	<tbody>
+		<?php foreach( $jsPages as $page ): ?>
+			<tr>
+				<td><?= $page['pageLink'] ?></td>
+				<td><?= $page['latestRevision']['statusMessage'] ?></td>
+				<td><?= $page['latestReviewed']['statusMessage'] ?></td>
+				<td><?= $page['liveRevision']['statusMessage'] ?></td>
+				<td>
+					<?php if ( $page['submit'] ): ?>
+						<button class="content-review-module-submit primary" data-page-name="<?= $page['pageName'] ?>">
```
Grunny: This is vulnerable to stored XSS, escape everything! :)

extensions/wikia/ContentReview/controllers/templates/JsPagesSpecialController_index.php
```diff
@@ -0,0 +1,39 @@
...
+		<th><?= wfMessage( 'content-review-special-list-header-actions' )->escaped() ?></th>
+	</tr>
+	</thead>
+	<tbody>
+		<?php foreach( $jsPages as $page ): ?>
+			<tr>
+				<td><?= $page['pageLink'] ?></td>
+				<td><?= $page['latestRevision']['statusMessage'] ?></td>
+				<td><?= $page['latestReviewed']['statusMessage'] ?></td>
+				<td><?= $page['liveRevision']['statusMessage'] ?></td>
```
Grunny: Is everything here already escaped?

extensions/wikia/ContentReview/controllers/JsPagesSpecialController.class.php
```diff
@@ -0,0 +1,89 @@
...
+
+		return $jsPages;
+	}
+
+	private function formatMessage( $jsPage ) {
+		$statusMessage = '';
+
+		if ( !empty( $jsPage['diffLink'] ) ) {
+			$statusMessage .= '<span class="content-review-status content-review-status-' . $jsPage['statusKey'] . '">'
+				. $jsPage['diffLink'] . '</span> ' . $jsPage['message'];
```
Grunny: Is everything here escaped? If not, escaped it (and consider building using `Html::element` or `Xml::element` which handles escaping for you.

extensions/wikia/ContentReview/controllers/ContentReviewApiController.class.php
```diff
@@ -62,15 +65,7 @@ public function submitPageForReview() {
...
-
-		$title = Title::newFromID( $pageId );
-		if ( $title === null || !$title->isJsPage() ) {
-			throw new NotFoundApiException( "JS page with ID {$pageId} does not exist");
-		}
-
-		$submitUserId = $this->wg->User->getId();
-		if ( !$submitUserId > 0 || !$this->canUserSubmit( $title )	) {
+		if ( !$this->wg->User->isLoggedIn() ) {
 			throw new PermissionsException( 'edit' );
```
Grunny: This now only checking logged in, so the permission specified in the exception shouldn't be `edit`.

extensions/wikia/ContentReview/services/ContentReviewStatusesService.class.php
```diff
@@ -0,0 +1,324 @@
...
+			}
+		} elseif ( !empty( $liveRevisionId ) ) {
+			$data['diffLink'] = $this->createRevisionLink( $title, $liveRevisionId );
+		}
+
+		return $data;
+	}
+
+	public function createPageLink( \Title $title ) {
+		return \HTML::element( 'a', [ 'href' => $title->getFullURL() ], $title->getText() );
```
Grunny: This will cause a fatal error:
`PHP Fatal error:  Class 'HTML' not found in extensions/wikia/ContentReview/services/ContentReviewStatusesService.class.php on line 297`

The class name is `Html` and our autoloader is case-sensitive, so same issue could happen in the other uses of the `Html` class below

extensions/wikia/ContentReview/controllers/templates/ContentReviewModule.mustache
```diff
@@ -12,11 +12,9 @@
 
 	{{#displaySubmit}}
 	<div class="content-review-module-submit-container">
-		<a href="#" id="content-review-module-submit" title="{{submit}}">
-		<button class="content-review-module-button primary" type="button">
+		<button class="content-review-module-submit primary" data-page-name="{{{pageName}}}">
```
Grunny: This one is still vulnerable to XSS. Always default to two braces in Mustache unless you know it's already escaped.

extensions/wikia/ContentReview/controllers/JsPagesSpecialController.class.php
```diff
@@ -0,0 +1,91 @@
...
+use Wikia\ContentReview\Helper;
+
+class JsPagesSpecialController extends WikiaSpecialPageController {
+
+	private static $topJsPages = [
+		'Wikia.js', 'Common.js'
+	];
+
+	function __construct() {
+		parent::__construct( 'JsPages', 'content-review-js-pages', true );
```
Grunny: I wonder if the special page name mmight be better as "Special:JSPages".

extensions/wikia/ContentReview/ContentReview.hooks.php
```diff
@@ -23,20 +23,10 @@ public static function register() {
...
-				'ContentReviewApiController',
-				'getPageStatus',
-				[
-					'wikiId' => $wgCityId,
-					'pageId' => $wgTitle->getArticleID(),
-				],
-				true
-			)->getData();
-
+		if ( $wgUser->isLoggedIn() ) {
```
Grunny: As this is not checking `->isJsPage()` any more (the check was in `userCanEditJsPage`) the rail module will now show up for all pages, not just JS pages in the MediaWiki namespace.

extensions/wikia/ContentReview/services/ContentReviewStatusesService.class.php
```diff
@@ -0,0 +1,324 @@
...
+	/**
+	 * Returns status key based on status
+	 *
+	 * @param int $status
+	 * @return string
+	 */
+	public function getStatusKey( $status = 0 ) {
+		switch ( $status ) {
+			case 1:
+			case 2: $statusKey = self::STATUS_AWAITING; break;
```
Grunny: Coding conventions, these should be:

``` javascript
case 2:
    $statusKey = self::STATUS_AWAITING;
    break;
```

adamkarminski:
LGTM after the fixes!

PR #8661, opened 2015-10-01
CE-2622 Prefill the feedback message with a link to a revision

adamkarminski:
Important stuff!
1. Ability to add `messageParams` as an array for prefilling an edit message
2. Feedback's section title becomes an anchor added to the Why? link in the ContentReview module.
3. **Restore updating of a review's status to ContentReviewApiController**

/cc @lgarczewski @Grunny


extensions/wikia/ContentReview/ContentReviewHelper.php
```diff
@@ -288,17 +294,40 @@ public function getToolbarTemplate() {
...
 			'action' => 'edit',
 			'section' => 'new',
-			'useMessage' => 'content-review-rejection-explanation'
+			'useMessage' => 'content-review-rejection-explanation',
 		];
+
+		if ( (int)$revisionId !== 0 ) {
+			$params['messageParams'] = [
+				1 => wfMessage( 'content-review-rejection-explanation-title' )->params( $revisionId )->escaped(),
+				2 => $title->getFullURL( "oldid={$revisionId}" ) . " {$revisionId}",
```
Grunny: I'd suggest making the URL and revision ID number separate parameters so the link in the message would become `[$2 $3]`. This gives more control to the translator as to what the link text should be and how to use the revision ID.

adamkarminski: Was thinking about it but did not want to "overengineer" it. However, if you too find it reasonable, I am going to do it. Thanks! :)

extensions/wikia/CreatePage/CreatePage.php
```diff
@@ -122,8 +122,13 @@ function wfCreatePageLoadPreformattedContent( $editpage ) {
...
 			}
-		} else if ( $msg = $wgRequest->getVal( 'useMessage' ) ) {
-			$editpage->textbox1 = wfMsgForContentNoTrans( $msg );
+		} else if ( $msgKey = $wgRequest->getVal( 'useMessage' ) ) {
+			$msg = wfMessage( $msgKey );
+			$msgParams = $wgRequest->getArray( 'messageParams' );
+			if ( $msgParams !== null ) {
+				$msg = $msg->params( $msgParams );
+			}
+			$editpage->textbox1 = $msg->escaped();
```
Grunny: `textbox1` should be escaped on output (obviously double check ;)) as it is run through `Html::textarea` so this would result in double escaping. Also, since it should replicate `wfMsgForContentNoTrans`, this is currently both missing the `->inContentLanguage()` (so it will be in the user's language rather than the wikia's) and it's doing the Trans step as part of escaped. So, the output format should be `->plain()` so it is "NoTrans" (unless we want the transformations to happen in the wikitext now?) with any extra escaping being done with `htmlspecialchars` or whatnot.

extensions/wikia/CreatePage/CreatePage.php
```diff
@@ -122,8 +122,13 @@ function wfCreatePageLoadPreformattedContent( $editpage ) {
 			} else {
 				$editpage->textbox1 = wfMsgForContentNoTrans( 'newpagelayout' );
 			}
-		} else if ( $msg = $wgRequest->getVal( 'useMessage' ) ) {
-			$editpage->textbox1 = wfMsgForContentNoTrans( $msg );
+		} else if ( $msgKey = $wgRequest->getVal( 'useMessage' ) ) {
```
Grunny: Minor, but while you're in there, this should be `elseif`.

adamkarminski: Haha, sure :D

Grunny:
LGTM. Merge if all is tested and good to go. :)

PR #8662, opened 2015-10-01
account for reston for user preference

nmonterroso:
@Wikia/services-team 

only write to preference service and compare preferences when not in Reston since there are currently no services in Reston. Related config PR Wikia/config#1362


ArturKlajnerok:
Lgtm :+1:

drsnyder:
Will all of this go away once the migration is complete? I suppose by that time we'll be on our way to having the UP service setup in RES.

LGTM

nmonterroso:
@drsnyder yah, this is definitely temporary.

PR #8664, opened 2015-10-01
SEO-84 Every page to link to global sitemap

gabrys:
Also removing the local sitemap link.

Exploring the possibility of sitemap links update (SEO-6) breaking the SEO


saipetch:
LGTM

PR #8665, opened 2015-10-01
adding noblockcheck query param to /info call

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-853

adding the `noblockcheck` query parameter to `/info` calls in mediawiki


drsnyder:
lgtm

PR #8667, opened 2015-10-02
Enable OX Bidder in default settings

fraszczakszymon:
It doesn't mean that OX bidder is enabled at all - it's needed to set up `wgAdDriverOpenXBidderCountries`, but it's possible in WikiFactory.


jacekjursza:
lgtm

PR #8668, opened 2015-10-02
WebStart.php: modify the order of read-only and HTTP POST check

macbre:
The previous order was setting the flag in [`CSRFDetector`](https://github.com/Wikia/app/blob/dev/extensions/wikia/Security/classes/CSRFDetector.class.php#L28-L36) that the HTTP method was checked for all requests. This change should improve the detection of "real" HTTP method checks.

@Grunny


Grunny:
:+1:

PR #8671, opened 2015-10-02
Common Request interface for WebRequest and WikiaRequest

adamkarminski:
I had issues trying to force a type of a method's argument but getting `WebRequest` || `WikiaRequest` which did not share any common parent class or interface. After consulting it with @macbre I decided to change it and provide an interface with the shared methods for interactions with a request.

/cc @macbre @Grunny


macbre:
`ILooksGoodToMe` :)

Grunny:
:+1:

PR #8672, opened 2015-10-02
WikiaMiniUpload: protect against CSRF

macbre:
Return `500 Internal Error` if the `chooseImage` request was not POSTed or had no token passed.

@Grunny


Grunny:
LGTM. :lock:

PR #8673, opened 2015-10-02
WikiaMiniUpload: protect against CSRF (take two)

macbre:
Protect `insertImage` method in a similar fashion as in #8672

@Grunny


Grunny:
LGTM.

PR #8676, opened 2015-10-02
CE-2813: Script to approve revision before given date

lukaszkonieczny:
Script to approve last JS page revision made by user with `editinterfacetrusted` right before given date passed as script argument `beforeDate`


adamkarminski:
LGTM

PR #8678, opened 2015-10-02
INT-221: Redlink dialog isn't firing under certain [currently unknown] circumstances

mari-mcmurtrie:
The cause was due to URL-encoded link title. So if English, it worked but if Japanese, then not. Fixed to decode URL-encoded title at the right spot. Ticket: https://wikia-inc.atlassian.net/browse/INT-221

### DIFF
--- a/extensions/wikia/CreatePage/js/CreatePage.js
+++ b/extensions/wikia/CreatePage/js/CreatePage.js
@@ -40,15 +40,18 @@ var CreatePage = {
 		'use strict';
 
 		var rs, dialogCallback;
+
 		// BugId:4941
 		if ( Boolean( window.WikiaEnableNewCreatepage ) === false ) {
 			// create page popouts are disabled - follow the link
 			return;
 		}
+
 		// Ignore middle-click. BugId:12544
 		if ( e && e.which === 2 ) {
 			return;
 		}
+
 		// don't follow the link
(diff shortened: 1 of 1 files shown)


mari-mcmurtrie:
@inez will you CR? The reason I ask you is simply because you are the one last touched this file. Thanks!

extensions/wikia/CreatePage/js/CreatePage.js
```diff
@@ -263,7 +263,7 @@ var CreatePage = {
 
 	redLinkClick: function( e, titleText ) {
 		'use strict';
-		var title = new mw.Title.newFromText( titleText ),
+		var title = new mw.Title.newFromText( decodeURIComponent( titleText ) ),
```
RafLeszczynski, line 266: @inez, @mari-mcmurtrie: is the a case where `titleText` could be `undefined`? If yes, are we OK with `decodeURIComponent` throwing `Uncaught ReferenceError` or should it be handled in the code?

inez, line 266: Such case should not happen - string is always passed to redLinkClick.

mari-mcmurtrie: Inez said this is ok.

PR #8681, opened 2015-10-05
ADEN-2515 ADEN-2514 check if is_array

jacekjursza:
fix for php warning


nandy-andy:
We don't want to do it. More information in the ticket :japanese_ogre:

PR #8684, opened 2015-10-05
[PLATFORM-1539] Pool counter error log improvement

jcellary:
Pool counter log error method used getWikiText method which returns error message which is localized - we don't want that, so I log the error array instead.

Additionally the error that was logged is very generic and doesn't provide any context. I added logging of the original status, which will potentially allow us to get more information about the root cause.

Pinging @macbre and @wladekb 

Warning! I'm a Media Wiki newbie so please make a good code review, to make sure I'm not breaking something :)


includes/PoolCounter.php
```diff
@@ -212,6 +214,7 @@ function execute( $skipcache = false ) {
 			default:
 				$errors = array( PoolCounter::QUEUE_FULL => 'pool-queuefull', PoolCounter::TIMEOUT => 'pool-timeout' );
 
+				$this->logError( $status );
```
macbre, line 220: Are you sure that you want this here? :)

includes/PoolCounter.php
```diff
@@ -161,7 +161,9 @@ function error( $status ) {
 	 * @param $status Status
 	 */
 	function logError( $status ) {
-		wfDebugLog( 'poolcounter', $status->getWikiText() );
+		$msg = json_encode( $status->getErrorsArray() );
+
+		wfDebugLog( 'poolcounter', $msg);
```
macbre: I'd keep the existing call to `wfDebugLog` and use `WikiaLogger` to log a message with more context - ideally with a context field `errors` with `$status->getErrorsArray()` (without JSON encoding).

includes/PoolCounter.php
```diff
@@ -161,7 +161,9 @@ function error( $status ) {
 	 * @param $status Status
 	 */
 	function logError( $status ) {
```
macbre: And as we're here - let's add a type hinting for this method.

includes/PoolCounter.php
jcellary, line 220: Yes, that's the main point of the change - I want to log the original status, hoping that I'll get information about the root cause. Does that make sense?

includes/PoolCounter.php
jcellary: Ok, I can change the log message, however part of the original problem was that the error message was being translated - this is part of the getWikiText routine - if we use it we'll get a localized message, that is why I only used the getErrorsArray method

includes/PoolCounter.php
jcellary: :+1:

includes/PoolCounter.php
macbre, line 220: Taking a look at more lines in diff does help - makes sense :)

includes/PoolCounter.php
macbre:
``` php
use Wikia\Logger\WikiaLogger;

WikiaLogger::instance()->error( 'poolcounter', [
  'message' => $msg,
  'errors' => $status->getErrorsArray(),
  'error' => $status->value
]);
```

macbre: Or even better, normalize `Status` objects in `WikiaLogger`. We do that for exceptions - https://github.com/Wikia/app/blob/dev/lib/composer/monolog/monolog/src/Monolog/Formatter/NormalizerFormatter.php#L81-L83

And then:

``` php
use Wikia\Logger\WikiaLogger;

WikiaLogger::instance()->error( 'poolcounter', [
  'status' => $status
]);
```

jcellary:
Addressed all PR issues, @macbre please take a second look :)

lib/composer/monolog/monolog/src/Monolog/Formatter/NormalizerFormatter.php
```diff
@@ -116,6 +119,18 @@ protected function normalizeException(Exception $e)
...
+        $data = array(
+            'is_ok' => $status->isOK(),
+            'errors' => $status->getErrorsArray(),
+            'message' => $status->getMessage(),
+            'value' => $status->value,
+        );
+
+        return $data;
+    }
+
```
macbre, line 138: Wrap this block in `Wikia change - begin` / `Wikia change - end` comments and we're good

macbre:
LGTM

lib/composer/monolog/monolog/src/Monolog/Formatter/NormalizerFormatter.php
```diff
@@ -116,6 +119,18 @@ protected function normalizeException(Exception $e)
         return $data;
     }
 
+    protected function normalizeStatus(Status $status)
+    {
+        $data = array(
+            'is_ok' => $status->isOK(),
+            'errors' => $status->getErrorsArray(),
+            'message' => $status->getMessage(),
```
macbre, line 131: Consider adding `'warnings' => $status->getWarningsArray()`

jcellary, line 131: Added

macbre:
:+1:

PR #8685, opened 2015-10-05
Read/Write only to preference service

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-664

Read and write only to the preference service. Before we can really call this "done" we need to figure out what to do in Reston, since Reston is reading from the MW datastore. However I don't know if that should be done as part of this ticket, since I suspect that preferences in Reston are only loaded as part of user loads, to get at other parts of the user.


ArturKlajnerok:
Lgtm :+1:

PR #8686, opened 2015-10-05
INT-219: Update new translation string for Achievement

mari-mcmurtrie:
Task: https://wikia-inc.atlassian.net/browse/INT-219
Also, some newly translated strings were not showing up and found that is due to existing messaging wikia. We want to remove mesaging wikia once the new translations go out to production. Follow up task for removing messaging wikia is INT-220.


lizlux:
LGTM. Please use the crowdin tool for branching and committing next time.

mari-mcmurtrie:
Follow up task to correct crowdin configuration: https://wikia-inc.atlassian.net/browse/INT-223

PR #8694, opened 2015-10-06
removed conditional part of TLS usage

natewatson999:
This change gets rid of the regular expression which determines the use of TLS, and changes the feed source to always use HTTPS instead of varying it. This is simpler and more secure.


natewatson999:
This addresses and fixes issue #235575.

jplatte:
Really surprising and disappointing to see that a big site like wikia still doesn't use https for everything. And even redirects you back if you try to use it.

PR #8697, opened 2015-10-06
(CE-2840) Properly escape parameters to contact form

Grunny:
Properly escape parameters to contact form.


lgarczewski:
LGTM, good find.

PR #8698, opened 2015-10-06
(CE-1564) Add contact form for reporting security issues

Grunny:
Add simple contact form for reporting security issues with an initial
responsible disclosure policy.

This just uses the standard Special:Contact form and adds some hacks
to handle selecting a security issue type. It's not pretty, but I don't think
it's worth rewriting parts of S:C right now to clean it up. ;)


lukaszkonieczny:
LGTM

PR #8699, opened 2015-10-06
Removed some functionality that was used to deliver full lyrics to th…

SeanColombo:
…e ORIGINAL LyricWiki app. Once this change is live (but NOT before) we can delete wgFullLyricsWikiApiToken from config.


wladekb:
lgtm

PR #8700, opened 2015-10-06
SEO-12 Fix HTML titles for most of the pages

gabrys:
Construct the full HTML title in OutputPage::setHTMLTitle. It used to be
distributed to a few files including Article.php and BlogArticle.php.
Now you just call the setHTMLTitle method and it adds the
" - Community Name - Wikia" suffix to the end. This makes nice titles for
all pages including Error and Special pages.

Revert an old hack removing the "Category:" prefix in titles. We want the
category pages to have titles distinct from articles with the same names.

Updated the wikia-pagetitle message to use the em dash for a separator
rather than a regular dash in the languages which use the em dash for
separating article title from community title in HTML title.


saipetch:
LGTM

PR #8701, opened 2015-10-06
DAT-3339: Mock HTML for Media Collection for Oasis

pauloslund:
This mock relies on the `visualeditor` wiki on wikia-dev.com

### DIFF
--- a/extensions/wikia/PortableInfobox/services/PortableInfoboxRenderService.class.php
+++ b/extensions/wikia/PortableInfobox/services/PortableInfoboxRenderService.class.php
@@ -17,7 +17,8 @@ class PortableInfoboxRenderService extends WikiaService {
 		'group' => 'PortableInfoboxItemGroup.mustache',
 		'horizontal-group-content' => 'PortableInfoboxHorizontalGroupContent.mustache',
 		'navigation' => 'PortableInfoboxItemNavigation.mustache',
-		'hero-mobile' => 'PortableInfoboxItemHeroMobile.mustache'
+		'hero-mobile' => 'PortableInfoboxItemHeroMobile.mustache',
+		'gallery' => 'PortableInfoboxGallery.mustache'
 	];
 	private $templateEngine;
 
@@ -74,6 +75,15 @@ class PortableInfoboxRenderService extends WikiaService {
 		if ( !empty( $heroData ) ) {
 			$infoboxHtmlContent = $this->renderInfoboxHero( $heroData ) . $infoboxHtmlContent;
 		}
+
+		$galleryData = Array (
+			'url' => 'http://vignette.wikia-dev.com/visualeditor/images/b/b8/Challenger.jpg/revision/latest?cb=20140626002212',
+			'name' => 'Challenger.jpg',
+			'key' => 'Challenger.jpg',
+			'alt' => '',
+			'caption' => 'This is a caption'
+		);
+		$infoboxHtmlContent = $this->renderItem('gallery', $galleryData) . $infoboxHtmlContent;
 
 		if ( !empty( $infoboxHtmlContent ) ) {
 			$output = $this->renderItem( 'wrapper',
@@ -151,7 +161,7 @@ class PortableInfoboxRenderService extends WikiaService {
 	private function renderItem( $type, array $data ) {
 		$helper = new PortableInfoboxRenderServiceHelper();
 
-		if ( $type === 'image' ) {
+		if ( $type === 'image' || $type === 'gallery' ) {
 			$data[ 'image' ][ 'context' ] = self::MEDIA_CONTEXT_INFOBOX;
 			$data = $helper->extendImageData( $data );
 			if ( !$data ) {
--- a/extensions/wikia/PortableInfobox/services/PortableInfoboxRenderService.class.php
+++ b/extensions/wikia/PortableInfobox/services/PortableInfoboxRenderService.class.php
@@ -76,29 +76,7 @@ class PortableInfoboxRenderService extends WikiaService {
 			$infoboxHtmlContent = $this->renderInfoboxHero( $heroData ) . $infoboxHtmlContent;
 		}
 
-		$galleryData = Array (
-			Array (
-				'url' => 'http://vignette.wikia-dev.com/visualeditor/images/b/b8/Challenger.jpg/revision/latest?cb=20140626002212',
-				'name' => 'Challenger.jpg',
-				'key' => 'Challenger.jpg',
-				'alt' => '',
-				'caption' => 'This is a caption'
-			),
-			Array (
-				'url' => 'http://vignette.wikia-dev.com/visualeditor/images/1/1d/Challenger-1.jpg/revision/latest?cb=20140626002317',
-				'name' => 'Challenger-1.jpg',
-				'key' => 'Challenger-1.jpg',
-				'alt' => '',
-				'caption' => 'This is a caption for Challenger 1'
-			),
-			Array (
-				'url' => 'http://vignette.wikia-dev.com/visualeditor/images/4/41/Challenger-0.jpg/revision/latest?cb=20140626002239',
-				'name' => 'Challenger-0.jpg',
-				'key' => 'Challenger-0.jpg',
-				'alt' => '',
-				'caption' => 'This is a caption for Challenger 0'
-			)
-		);
+		$galleryData = $this->mockGalleryData();
 		$infoboxHtmlContent = $this->renderGallery($galleryData) . $infoboxHtmlContent;
 
 		if ( !empty( $infoboxHtmlContent ) ) {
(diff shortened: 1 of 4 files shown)


extensions/wikia/PortableInfobox/templates/PortableInfoboxGallery.mustache
```diff
@@ -0,0 +1,19 @@
...
+		<span class="duration" itemprop="duration">{{duration}}</span>
+		{{/duration}}
+		<span class="play-circle"></span>
+		{{/isVideo}}
+	</a>
+  <figcaption class="pi-item-spacing pi-gallery-caption">This is a caption</figcaption>
+</figure>
+
+<div class="pi-image-gallery-extras">
+  <img src="http://vignette.wikia-dev.com/visualeditor/images/1/1d/Challenger-1.jpg/revision/latest?cb=20140626002317" class="pi-image-thumbnail" alt="A.jpg" data-{{media-type}}-key="Challenger-1.jpg" data-{{media-type}}-name="Challenger-1.jpg" />
```
inez: This is not generic - it references specific image: Challenger

extensions/wikia/PortableInfobox/services/PortableInfoboxRenderService.class.php
```diff
@@ -75,6 +76,15 @@ public function renderInfobox( array $infoboxdata, $theme, $layout ) {
 			$infoboxHtmlContent = $this->renderInfoboxHero( $heroData ) . $infoboxHtmlContent;
 		}
 
+		$galleryData = Array (
```
inez: To proof the concept this perhaps should be an array of images - not one image.

extensions/wikia/PortableInfobox/templates/PortableInfoboxGallery.mustache
```diff
@@ -0,0 +1,19 @@
+<figure class="pi-item pi-image pi-image-gallery">
+	<a href="{{url}}" class="image image-thumbnail{{#isVideo}} video video-thumbnail small{{/isVideo}}"
```
inez: Too much indentation

extensions/wikia/PortableInfobox/services/PortableInfoboxRenderService.class.php
```diff
@@ -140,6 +166,26 @@ private function renderInfoboxHero( $data ) {
...
+	 * @param array $data - array of images
+	 *
+	 * @return string
+	 */
+	private function renderGallery( $data ) {
+		$helper = new PortableInfoboxRenderServiceHelper();
+
+		$galleryData = Array();
+		$hero = array_shift($data);
+		$galleryData['hero'] = $helper->extendImageData( $hero );
```
inez: This "hero" keyword/concept in our codebase is really reserved to this particular one feature - hero image - which displays image+title. I don't think we should be using it here.

inez: Also - extendImageData - does not heed to be called for other images than first one?

inez: Wouldn't it be possible to pass entire array to template and have template render it - little bit different treatment for first item and little bit different for other items?

inez:
Btw. Please also make sure to test those things with videos.

extensions/wikia/PortableInfobox/services/PortableInfoboxRenderService.class.php
sebastianmarzjan: Please consider using the [] array notation for consistency, moving this mocked data into a separate method and marking a TODO to remove it afterwards :)

extensions/wikia/PortableInfobox/services/PortableInfoboxRenderService.class.php
```diff
@@ -75,6 +76,31 @@ public function renderInfobox( array $infoboxdata, $theme, $layout ) {
...
+			),
+			Array (
+				'url' => 'http://vignette.wikia-dev.com/visualeditor/images/1/1d/Challenger-1.jpg/revision/latest?cb=20140626002317',
+				'name' => 'Challenger-1.jpg',
+				'key' => 'Challenger-1.jpg',
+				'alt' => '',
+				'caption' => 'This is a caption for Challenger 1'
+			),
+			Array (
+				'url' => 'http://vignette.wikia-dev.com/visualeditor/images/4/41/Challenger-0.jpg/revision/latest?cb=20140626002239',
```
sebastianmarzjan: Please add a couple of images to cover use cases of:
- image in portrait orientation
- small image (narrower than 300px)
- images with aspect ratios typical for use cases on some wikis:
  - 1:2 (happens for characters, melee weapons)
  - 2:1 (i.e. firearms)

extensions/wikia/PortableInfobox/styles/PortableInfobox.scss
```diff
@@ -81,6 +81,35 @@
 	text-align: center; // center smaller image inside infobox
 }
 
+.pi-image-gallery {
+	position: relative;
+
+	&::after {
+		background-image: url("/extensions/VisualEditor/modules/ve-mw/ui/styles/images/icons/gallery-ltr-white.svg");
```
sebastianmarzjan, line 89: I know it's a mock, though please add a TODO to remove this coupling with VE styles

extensions/wikia/PortableInfobox/templates/PortableInfoboxGallery.mustache
```diff
@@ -0,0 +1,20 @@
+{{#hero}}
+  <figure class="pi-item pi-image pi-image-gallery">
```
RafLeszczynski, line 2: Please consider reusing image template for example as a mustache partial. That way you will avoid code duplication

extensions/wikia/PortableInfobox/styles/PortableInfobox.scss
```diff
@@ -81,6 +81,35 @@
...
+		background-repeat: no-repeat;
+		background-size: contain;
+		content: '';
+		display: block;
+		height: 30px;
```
RafLeszczynski: Please consider moving the magic number to a variable for example `$icon-size` and reuse this variable for both `height` and `width`

extensions/wikia/PortableInfobox/styles/PortableInfobox.scss
```diff
@@ -81,6 +81,35 @@
...
+		position: absolute;
+		right: $infobox-item-margin;
+		top: $infobox-item-margin;
+		width: 30px;
+	}
+}
+
+.pi-gallery-caption {
+	color: white;
```
RafLeszczynski: color declarations should be stored in sass variables. Please run sass-linter for this file to check for other issues

extensions/wikia/PortableInfobox/services/PortableInfoboxRenderService.class.php
pauloslund: I'm torn as to whether or not that's the best approach. Since mustache doesn't include any logic, you would have to assign a value to the first image in php, something like this:

``` php
$galleryData['images'] = [];

foreach ($data as $key => $image) {
    if ($key === 0) {
        $image['first'] = true;
    }
    $galleryData['images'][] = $helper->extendImageData($image);
}
```

and then your mustache template would look something like this:

``` mustache
{{#images}}
  {{#first}}
    <figure class="pi-item pi-image pi-image-gallery">
      <a href="{{url}}" class="image image-thumbnail{{#isVideo}} video video-thumbnail small{{/isVideo}}" title="{{alt}}">
        <img src="{{thumbnail}}" class="pi-image-thumbnail" alt="{{alt}}" width="{{{width}}}" height="{{{height}}}" data-{{media-type}}-key="{{name}}" data-{{media-type}}-name="{{name}}"/>
      </a>
      {{#caption}}<figcaption class="pi-item-spacing pi-gallery-caption">{{{caption}}}</figcaption>{{/caption}}
    </figure>
  {{/first}}

  {{^first}}
    <img src="{{thumbnail}}" class="pi-image-thumbnail" alt="{{name}}" data-image-key="{{name}}" data-image-name="{{name}}" style="display: none;" />
  {{/first}}
{{/images}}
```

I think it's more clear to have it separated into cover and extra images, however if we just pass in an array we may be able to use the same method for mobile and just use a different template

extensions/wikia/PortableInfobox/services/PortableInfoboxRenderService.class.php
```diff
@@ -140,6 +144,58 @@ private function renderInfoboxHero( $data ) {
...
+
+		return $markup;
+	}
+
+	/**
+	 * temporary function to provide mock data for a gallery
+	 *
+	 * @return array
+	 */
+	private function mockGalleryData() {
```
inez: This is a mock so doesn't matter (does not have to be fixed) but we prefer to prefix names of this kind of methods with word "get" getMockedGaleryData, etc.

inez:
@pauloslund This one should be closed, right?

PR #8702, opened 2015-10-07
DAT-3348: Modified NodeImage to return array of images instead of one image

inez:
This change is needed in order to support backward compatiblty for case where <gallery> or <tabber> is passed as a single parameter to a template.
There is no proper rendering yet for collections of images so for now only the first item is rendered.


inez:
Wrong target branch

PR #8703, opened 2015-10-07
DAT-3348: Modified NodeImage to return array of images instead of one image

inez:
This change is needed in order to support backward compatiblty for case where <gallery> or <tabber> is passed as a single parameter to a template.
There is no proper rendering yet for collections of images so for now only the first item is rendered.

### DIFF
--- a/extensions/wikia/PortableInfobox/services/Parser/Nodes/NodeImage.php
+++ b/extensions/wikia/PortableInfobox/services/Parser/Nodes/NodeImage.php
@@ -12,41 +12,118 @@ class NodeImage extends Node {
 
 	public function getData() {
 		if ( !isset( $this->data ) ) {
-			$imageData = $this->getRawValueWithDefault( $this->xmlNode );
+			$this->data = array();
 
-			if( is_string($imageData) && PortableInfoboxDataBag::getInstance()->getGallery($imageData)) {
-				$imageData = PortableInfoboxDataBag::getInstance()->getGallery($imageData);
-			}
+			$parameterValue = $this->getRawValueWithDefault( $this->xmlNode );
 
-			$title = $this->getImageAsTitleObject( $imageData );
-			$file = $this->getFilefromTitle( $title );
-			if ( $title instanceof \Title ) {
-				$this->getExternalParser()->addImage( $title->getDBkey() );
-			}
-			$ref = null;
-			$alt = $this->getValueWithDefault( $this->xmlNode->{self::ALT_TAG_NAME} );
-			$caption = $this->getValueWithDefault( $this->xmlNode->{self::CAPTION_TAG_NAME} );
+			if ( $this->containsTabberOrGallery( $parameterValue ) ) {
+				$parsed = $this->getExternalParser()->parseRecursive( $parameterValue );
 
-			$this->data = [
-				'url' => $this->resolveImageUrl( $file ),
-				'name' => ( $title ) ? $title->getText() : '',
-				'key' => ( $title ) ? $title->getDBKey() : '',
-				'alt' => $alt,
-				'caption' => $caption
-			];
-
-			if ( $this->isVideo( $file ) ) {
-				$this->data = $this->videoDataDecorator( $this->data, $file );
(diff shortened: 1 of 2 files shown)


extensions/wikia/PortableInfobox/services/Parser/Nodes/NodeImage.php
```diff
@@ -12,41 +12,118 @@ class NodeImage extends Node {
...
-			if( is_string($imageData) && PortableInfoboxDataBag::getInstance()->getGallery($imageData)) {
-				$imageData = PortableInfoboxDataBag::getInstance()->getGallery($imageData);
+			$this->data = array();
+
+			$parameterValue = $this->getRawValueWithDefault( $this->xmlNode );
+
+			if ( $this->containsTabberOrGallery( $parameterValue ) ) {
+				$parsed = $this->getExternalParser()->parseRecursive( $parameterValue );
+
+				$items = $this->getGalleryItems( $parsed ) + $this->getTabberItems( $parsed );
```
sebastianmarzjan: If the intent of this line is to merge arrays resulting from getGalleryItems with those from getTabberItems, the '+' operator will yield different results unless list indexes are unique. 
Is it the intended behavior?

extensions/wikia/PortableInfobox/services/Parser/Nodes/NodeImage.php
```diff
@@ -12,41 +12,118 @@ class NodeImage extends Node {
...
-			];
-
-			if ( $this->isVideo( $file ) ) {
-				$this->data = $this->videoDataDecorator( $this->data, $file );
+		}
+
+		return $items;
+	}
+
+	private function getTabberItems( $html ) {
```
sebastianmarzjan, line 67: Handling that differently would need more in-depth work inside of the parsing flow, similar to what the ArticleAsJson does. For now this should work, though it needs some tests both for extraction (unit) and to notify if expected format is returned by parser from tabber (one integration test)

extensions/wikia/PortableInfobox/services/Parser/Nodes/NodeImage.php
```diff
@@ -12,41 +12,118 @@ class NodeImage extends Node {
...
+	private function containsTabberOrGallery( $str ) {
+		// TODO: Consider more robust approach (UNIQ...QINU)
+		$strLower = strtolower( $str );
+		if ( strpos( $strLower, '-tabber-' ) !== false || strpos( $strLower, '-gallery-' ) !== false ) {
+			return true;
+		}
+		return false;
+	}
+
+	private function getImage( $parameterValue, $alt, $caption ) {
```
dianafa, line 96: maybe we can make this function look like: `getImage( $parameterValue, $caption, $alt = null )` and by this not to pass duplicated and redundant params here: https://github.com/Wikia/app/pull/8703/files#diff-a4400e3b90d497838c5bc1981a0b4da6R27 ?

extensions/wikia/PortableInfobox/services/Parser/Nodes/NodeImage.php
```diff
@@ -12,41 +12,118 @@ class NodeImage extends Node {
...
+					$this->getValueWithDefault( $this->xmlNode->{self::ALT_TAG_NAME} ),
+					$this->getValueWithDefault( $this->xmlNode->{self::CAPTION_TAG_NAME} )
+				);
 			}
+		}
+
+		return $this->data;
+	}
+
+	private function getGalleryItems( $html ) {
```
sebastianmarzjan, line 42: Can we plug into the gallery earlier than at the html stage, per discussion in https://github.com/Wikia/app/pull/8595 ?

idradm, line 42: :+1:

BladeBronson:
Looks good to me! @ChangeTip, give @inez an excellent Wayne!

changetip:
Hi @inez, I've delivered a tip worth an excellent Wayne ($0.50) from @BladeBronson to your ChangeTip pocket.

an excellent Wayne

**[Learn about ChangeTip](https://www.changetip.com/tip-online/github)**

PR #8704, opened 2015-10-07
INT-197: Wikia-in-your-lang: implement using the interlanguage links

mari-mcmurtrie:
Current "wikia-in-your-lang for article" shows interlanguage article links only when a title of the article is exactly the same as another language article because we do "page title" string search in database. This is a problem because an article "Darth Vader" and another article "ダースベイダー"  never matches although both are articles about Darth Vader.

I investigated some wikias and found that lots of interlanguage links in the bottom of articles are indeed coming from the same template called "ArticleInterLang"; hence, we can get interlanguage information from this template and propagate the information to the existing wikia-in-your-lang code. Code structure was discussed with @owend. Not all wikias use "ArticleInterLang" for interlanguage list in the bottom page so if the wikia is not using "ArticleInterLang", then this behavior falls into default.

Ticket: https://wikia-inc.atlassian.net/browse/INT-197


mari-mcmurtrie:
@owend @Grunny I know you guys are very very busy now but if you have a chance to CR, that will be great.

owend:
This seems fine to me.  :+1:

PR #8706, opened 2015-10-07
VOLDEV-161: Remove strict type check from LinkSuggestHooks

mszabo-wikia:
Fixes regression causing the hook to not work with SemanticForms
@Grunny  @macbre fast-tracked this one due to priority


macbre:
:+1:

Grunny:
LGTM.

PR #8709, opened 2015-10-07
 CE-2824: Add content-reviewer permission group

lukaszkonieczny:
Added new group `content-reviewer`

Users in this group are only which have access to Special:ContentReview and can approve or reject JS page revision.

Only user in this group can add/remove other user.


adamkarminski:
LGTM

PR #8710, opened 2015-10-07
Get rid of $wgWikiaBotUsers

macbre:
@michalroszka / @wladekb

### DIFF
--- a/includes/wikia/Wikia.php
+++ b/includes/wikia/Wikia.php
@@ -22,6 +22,7 @@ $wgHooks['ComposeMail']              [] = "Wikia::ComposeMail";
 $wgHooks['SoftwareInfo']             [] = "Wikia::softwareInfo";
 $wgHooks['AddNewAccount']            [] = "Wikia::ignoreUser";
 $wgHooks['ComposeMail']              [] = "Wikia::isUnsubscribed";
+$wgHooks['AllowNotifyOnPageChange']  [] = "Wikia::allowNotifyOnPageChange";
 $wgHooks['AfterInitialize']          [] = "Wikia::onAfterInitialize";
 $wgHooks['UserMailerSend']           [] = "Wikia::onUserMailerSend";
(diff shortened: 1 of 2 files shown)


lib/Wikia/tests/Service/User/UserPreferencesIntegrationTest.php
```diff
@@ -46,8 +47,7 @@ private function getUserPreferences($ignoreCache = false) {
 	}
 
 	function setUp() {
-		global $wgWikiaBotUsers;
-		$this->testUserName = $wgWikiaBotUsers["bot"];
+		$this->testUserName = self::TEST_USER_NAME;
```
macbre, line 50: @harnash

includes/wikia/Wikia.php
```diff
@@ -1341,20 +1341,22 @@ static public function isUnsubscribed( $to, $body, $subject ) {
...
+	 *
+	 * @param User $editor
+	 * @param Title $title
+	 * @return bool return false if you want to block an email
+	 */
+	static public function allowNotifyOnPageChange ( User $editor, /* Title */ $title ) {
+		global $wgWikiaBotLikeUsers;
 
-		return $allow;
+		if ( in_array( $editor->getName(), $wgWikiaBotLikeUsers)  ) {
```
ghost, line 1354: how about `return ! in_array( $editor->getName(), $wgWikiaBotLikeUsers);` ?

macbre, line 1354: It's the opposite actually

ghost, line 1354: Then with `!` :-)

macbre, line 1354: I'd keep the if / else approach as the "direct" `return` can be confusing - as the above proves ;)

macbre:
The above is `SwiftStorageTest` fail

includes/wikia/Wikia.php
ghost, line 1354: OK, then keep it :+1:

PR #8713, opened 2015-10-07
XW-432 don't fallback to empty nav if one of elements is incorrect

bognix:
https://wikia-inc.atlassian.net/browse/XW-432

Related PR: https://github.com/Wikia/mercury/pull/1503

### DIFF
--- a/extensions/wikia/MercuryApi/MercuryApiController.class.php
+++ b/extensions/wikia/MercuryApi/MercuryApiController.class.php
@@ -175,10 +175,11 @@ class MercuryApiController extends WikiaController {
 	 * @return array
 	 */
 	private function getNavigationData() {
-		//getData always return array with navigation and wiki keys.
-		//See: includes/wikia/api/NavigationApiController.class.php:28
-		//and: includes/wikia/models/NavigationModel.class.php:175
-		return $this->sendRequest( 'NavigationApi', 'getData' )->getData()['navigation']['wiki'];
+		$navData = $this->sendRequest( 'NavigationApi', 'getData' )->getData();
+		if ( isset( $navData['navigation']['wiki'] ) ) {
+			return $this->sendRequest('NavigationApi', 'getData')->getData()['navigation']['wiki'];
+		}
+		return [ ];
 	}
 
 	/**
@@ -281,7 +282,6 @@ class MercuryApiController extends WikiaController {
 			\Wikia\Logger\WikiaLogger::instance()->error(
 				'Fallback to empty navigation'
 			);
-			$navData = [ ];
 		}
 
 		$wikiVariables['navigation'] = $navData;
--- a/extensions/wikia/MercuryApi/MercuryApiController.class.php
+++ b/extensions/wikia/MercuryApi/MercuryApiController.class.php
@@ -175,7 +175,10 @@ class MercuryApiController extends WikiaController {
 	 * @return array
 	 */
 	private function getNavigationData() {
-		return $this->sendRequest( 'NavigationApi', 'getData' )->getData();
+		//getData always return array with navigation and wiki keys.
+		//See: includes/wikia/api/NavigationApiController.class.php:28
+		//and: includes/wikia/models/NavigationModel.class.php:175
+		return $this->sendRequest( 'NavigationApi', 'getData' )->getData()['navigation']['wiki'];
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


includes/wikia/api/NavigationApiController.class.php
```diff
@@ -30,11 +30,6 @@ public function getData(){
 			[ 'urlFields' => 'href' ],
 			NavigationModel::CACHE_TTL
 		);
-		$errors = $model->getErrors();
-
-		if ( !empty( $errors ) ) {
```
bognix, line 35: This method is only used by MercuryApi.
So it's safe to remove this part.

Exception is no longer thrown to unify behavior across Mercury and Oasis.
In Oasis when model has errors data attribute is added to navigation HTML node and this attr is only read in edit mode.

Without this change Mercury was displaying empty nav in case of errors inside model.
With this change Mercury should display the same thing as Oasis

extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -175,7 +175,10 @@ private function getTopContributorsDetails( Array $ids ) {
 	 * @return array
 	 */
 	private function getNavigationData() {
-		return $this->sendRequest( 'NavigationApi', 'getData' )->getData();
+		//getData always return array with navigation and wiki keys.
+		//See: includes/wikia/api/NavigationApiController.class.php:28
+		//and: includes/wikia/models/NavigationModel.class.php:175
+		return $this->sendRequest( 'NavigationApi', 'getData' )->getData()['navigation']['wiki'];
```
bognix: This change is just a proposal.
It's related with https://github.com/Wikia/mercury/pull/1503

rogatty: Will need some bullet proofing as it can easily fail.

> getData always return array with navigation and wiki keys

You're optimistic here ;)

extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -273,18 +276,15 @@ public function getWikiVariables() {
 
 		$wikiVariables = $this->mercuryApi->getWikiVariables();
 
-		try {
-			$wikiVariables['navData'] = $this->getNavigationData();
-		} catch ( Exception $e ) {
+		$navData = $this->getNavigationData();
+		if ( empty( $navData ) ) {
 			\Wikia\Logger\WikiaLogger::instance()->error(
```
rogatty, line 282: Should this be an error? We'll get tickets created by ErrorReporter. Do we know what to do with them?

extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -273,18 +276,15 @@ public function getWikiVariables() {
...
-				[
-					'exception' => $e
-				]
+				'Fallback to empty navigation'
 			);
-			$wikiVariables['navData'] = [ ];
+			$navData = [ ];
 		}
 
+		$wikiVariables['navigation'] = $navData;
```
rogatty: Can we agree on a single name?

extensions/wikia/MercuryApi/MercuryApiController.class.php
bognix: well code is straightforward and if some fatal error occurs along the way it won't work anyway.
If this change is going to be more than this one liner here and in Mercury (your comment about caching) then I'm not sure if it's worth doing at this point.
We already decided we need to rethink local nav in Mercury.
Maybe we should fix it then?

rogatty: I'd add a simple check if `$this->sendRequest( 'NavigationApi', 'getData' )->getData()` has `['navigation']['wiki']`. No need for drama :P

extensions/wikia/MercuryApi/MercuryApiController.class.php
bognix, line 282: I wanted to leave this error here because there was a lot of errors reported in Kibana and I hope this should fix all of them but if not, it's good to have some message somewhere that it's still not working

extensions/wikia/MercuryApi/MercuryApiController.class.php
bognix: the question is if we want to pursue with flattening navData array structure.
If yes, I rename variables

rogatty:
LGTM

hakubo:
LGTM

PR #8715, opened 2015-10-07
SEO-98 Roll back to the old robots.txt

gabrys:
It seems there's a lot more URLs indexed since we rolled out the new
robots.txt. We want to verify it was because of the new robots.txt
logic.


saipetch:
LGTM

PR #8716, opened 2015-10-07
add ability to overwrite the langauge of email messages for a user

nmonterroso:
@Wikia/services-team @jsutterfield 
https://wikia-inc.atlassian.net/browse/SERVICES-886

Because of the race condition with reading preferences from the preference service in MediaWiki as part of the registration request to Helios, pass the language code from Helios to MediaWiki, and honor that when sending a confirmation email.


ArturKlajnerok:
Lgtm :+1: 
/cc @garthwebb

jsutterfield:
LGTM

PR #8717, opened 2015-10-08
When loading from the session, ensure that the users helios authentication token is also valid

drsnyder:
**This is experimental.** I believe it will disable existing usage of `api.php` unless users of that end point update their scripts/requests with `X-Wikia-AccessToken` and a valid token.

This should ensure that if a user is loaded from a MediaWiki session, that their helios authentication token is also still valid. This _should_ allow us to globally logout users by deleting their helios tokens.

It should also be minimally invasive. Once all of the authentication fallbacks are completed the helios authentication token is also verified. If it is not valid then the request will proceed as the anonymous user.

https://wikia-inc.atlassian.net/browse/SERVICES-887

@Wikia/services-team @owend @macbre @michalroszka


macbre:
We should communicate this breaking change with ComSup as many wikis use bots (crafted for MediaWiki) for edits automation. And they use session cookie for keeping the logged-in session between the requests.

drsnyder:
@Grunny I would also like to heard from you on this change.

includes/User.php
```diff
@@ -246,6 +247,11 @@ class User {
 	static $idCacheByName = array();
 
 	/**
+		* @var string the service auth token (currently helios); should NEVER be cached
+	 */
+	private $globalAuthToken = null;
```
nmonterroso, line 253: I think that any class variables end up as part of the Memcache object. Will that break anything with multiple sessions?

includes/User.php
```diff
@@ -1141,6 +1155,23 @@ private function loadFromSession() {
 			return false;
 		}
 
+		if ( !$this->isUserAuthenticated() ) {
```
nmonterroso: should we remove the statements above these? seems like if we want Helios to be the only source of authentication we can't let other cases early-out

includes/User.php
drsnyder, line 253: The caching should only be for [these variables](https://github.com/Wikia/app/blob/dev/includes/User.php#L99). The construction of the cache object happens [here](https://github.com/Wikia/app/blob/dev/includes/User.php#L434). So, unless there is some other magic this variable should not be cached.

includes/User.php
drsnyder: I think we should do that only after we a) thoroughly understand how this works or b) implement a comprehensive security model and apply it to MW. In other words, I don't want to break any "security" measures that are currently in place.

includes/User.php
nmonterroso, line 253: Yah that makes sense, but I remember a case when I tried to have 

``` php
private $preferenceService

private function getPreferenceService() {
  return $this->preferenceService;
}
```

and I got an error about trying to serialize a database connection, which I think I tracked down to some caching going on somewhere.

drsnyder, line 253: I bet that was happening in the `UserService`. If I remember correctly it was caching the whole user and not invalidating it. @owend fixed that about a week or so ago.

drsnyder, line 253: Yup. It was [caching the user object](https://github.com/Wikia/app/pull/8628/files#diff-fd654af4c113963df43fbd159b1569c8L257).

nmonterroso, line 253: Ah, ok cool :+1:

nmonterroso:
LGTM

includes/User.php
```diff
@@ -1142,6 +1156,13 @@ private function loadFromSession() {
 			return false;
 		}
 
+		if ( !$this->isUserAuthenticatedViaAuthenticationService() ) {
+			$this->logFallbackToMediaWikiSessionRejection( $from );
+			$this->logout();
```
drsnyder, line 1161: I added this to clean up the users session and cookie state.

ArturKlajnerok:
Lgtm :+1:

PR #8718, opened 2015-10-08
INT-201: Exclude users who are in $wgAchExemptUsers WikiFactory varia…

mari-mcmurtrie:
…ble from achievement.

Exclude a user "Tomeito" from the Achievements board all together in ja.starwars.wikia.com. We still have to define this new wiki variable ($wgAchExemptUsers). I file a separate ticket for it (INT-226).
Ticket: https://wikia-inc.atlassian.net/browse/INT-201


mari-mcmurtrie:
@lgarczewski I do not know if you have time to take a look but here is what I did!

extensions/wikia/AchievementsII/services/AchAwardingService.class.php
```diff
@@ -695,7 +695,7 @@ public static function canEarnBadges( User $user = null ) {
 		if (
 			$user->isAnon() ||
 			$user->isBlocked() ||
-			( $user->isAllowed( 'bot' ) || in_array( $user->getName(), $wgWikiaBotLikeUsers ) ) ||
+			( $user->isAllowed( 'bot' ) || in_array( $user->getName(), $wgWikiaBotLikeUsers ) ) || (is_array( $wgAchExemptUsers ) && in_array( $user->getId(), $wgAchExemptUsers)) ||
```
RafLeszczynski: Please consider breaking condition into multiple lines for better readability

mari-mcmurtrie:
@lgarczewski @RafLeszczynski will you CR when you have a chance? thanks

mari-mcmurtrie:
@inez can you CR ??? I think Poland people are busy. I tested this in dev with the wikifactory variable set.

inez:
Overall looks good to me but I success changing variable name to use word "achievement" instead of "ach". I know that in the past we started using that weird shortcut, but it was for not a good reason.

inez:
Actually now I looked (I was on the mobile before) and the extension name is achievementS so variable probably should use that (with the S at the end).
Btw. I am POLAND people too!! ;)

extensions/wikia/AchievementsII/services/AchAwardingService.class.php
```diff
@@ -696,6 +696,7 @@ public static function canEarnBadges( User $user = null ) {
 			$user->isAnon() ||
 			$user->isBlocked() ||
 			( $user->isAllowed( 'bot' ) || in_array( $user->getName(), $wgWikiaBotLikeUsers ) ) ||
+			( is_array( $wgAchievementExemptUsers ) && in_array( $user->getId(), $wgAchievementExemptUsers ) ) ||
```
inez: If you want to avoid having to use that is_array - so code would be bit cleaner - you can just define that variable as an empty array in DefaultSettings.php

PR #8723, opened 2015-10-08
CE-2584 Updated ru message

adamkarminski:
The new message source: http://messaging.wikia.com/wiki/MediaWiki:Insights-list-description-nonportableinfoboxes/ru?action=raw

/cc @lukaszkonieczny


lukaszkonieczny:
LGTM

PR #8727, opened 2015-10-08
ADEN-2534 Insert message above WikiaTopAds

fraszczakszymon:


extensions/wikia/AdEngine/js/recovery/message.js
```diff
@@ -12,21 +13,37 @@ define('ext.wikia.adEngine.recovery.message', [
...
-	function addRecoveryCallback() {
-		recoveryHelper.addOnBlockingCallback(function () {
-			recover();
-		});
+	function createMessage(uniqueClassName) {
+		var className = 'recovered-message',
+			div = doc.createElement('div');
+
+		div.textContent = 'Hello world!';
+		div.classList.add(className, className + '-' + uniqueClassName);
```
nandy-andy, line 24: ~~Since we [support IE11](http://community.wikia.com/wiki/Help:Supported_Browsers) we [can't use the 2nd parameter](http://caniuse.com/#feat=classlist) and we need to implement toggling again :(~~ It's fine IE11 doesn't like 2nd parameter in `toggle()` not in `add()`.

skins/oasis/css/core/ads.scss
```diff
@@ -267,3 +267,19 @@ body >a >img[src$="noad.gif"] {
...
+	color: #000;
+	font-size: 2em;
+	text-align: center;
+	padding: 10px 0;
+
+	&.recovered-message-top {
+		margin: 0 auto;
+		position: relative;
+		width: 970px;
+		z-index: 2;
```
nandy-andy, line 283: We might want to use one of the [SASS variables](https://github.com/Wikia/app/blob/dev/skins/oasis/css/core/layout.scss#L2) but maybe it's a part of the other story. It's up to you.

nandy-andy:
OK, we discussed offline the issues and it looks good. The failed tests were PHP tests connected to UserProfile and they were fixed by somebody else. The `z-index` value will be moved to a variable in the other pull request. Let's move on with this story and TEST it! :smile_cat:

PR #8729, opened 2015-10-08
(CE-2851) Properly escape attributes and vars in Special:MultiLookup

Grunny:
Reviewed on dev.


macbre:
:+1:

PR #8733, opened 2015-10-09
MAIN-5540 Log changes for Junior helper group

armonr:
And add username to log

/CC @Grunny @garthwebb @jsutterfield

### DIFF
--- a/extensions/wikia/UserLogin/UserLoginSpecialController.class.php
+++ b/extensions/wikia/UserLogin/UserLoginSpecialController.class.php
@@ -615,7 +615,7 @@ class UserLoginSpecialController extends WikiaSpecialPageController {
 			return;
 		}
 
-		if ( !empty( array_intersect( F::app()->wg->AccountAdminGroups, $user->getAllGroups() ) ) ) {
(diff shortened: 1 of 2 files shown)


armonr:
Reverted my change forbidding password to be updated for admin/util through prefs.

extensions/wikia/UserLogin/UserLoginSpecialController.class.php
```diff
@@ -615,8 +615,10 @@ public function mailPassword() {
 			return;
 		}
 
-		if ( !empty( array_intersect( F::app()->wg->AccountAdminGroups, $user->getGroups() ) ) ) {
-			\Wikia\Logger\WikiaLogger::instance()->warning("Junior helper cannot change account info");
+		if ( !empty( array_intersect( F::app()->wg->AccountAdminGroups, $user->getAllGroups() ) ) ) {
```
garthwebb: Why was this changed to `getAllGroups`?  The `getAllGroups` method returns _all_ groups defined in the system, not just the groups for that user.

PR #8734, opened 2015-10-09
CE-2852 Fix XSS vulnerability(ies) in LookupUser

adamkarminski:
/cc: @Grunny @macbre

### DIFF
--- a/extensions/wikia/LookupUser/templates/contribution.table.tmpl.php
+++ b/extensions/wikia/LookupUser/templates/contribution.table.tmpl.php
@@ -138,8 +138,8 @@ $(document).ready(function() {
 								dataType: 'json',
 								type: "POST",
 								data: {
-									url: mw.html.escape( url ),
-									username: mw.html.escape( username ),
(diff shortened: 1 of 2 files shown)


extensions/wikia/LookupUser/templates/contribution.table.tmpl.php
```diff
@@ -129,15 +129,19 @@
...
+						var username = self.find('input.name').val();
 
 						try {
 							var ajaxRequest = $.ajax({
 								dataType: 'json',
 								type: "POST",
-								data: 'url=' + url + '&username=' + username + '&id=' + wikiId,
+								data: {
+									url: mw.html.escape( url ),
+									username: mw.html.escape( username ),
```
Grunny: Escaping here is unneeded as it's just POST data (and could actually lead to incorrect data if the username has a HTML entity in it.

adamkarminski:
unitTests failure unrelated.

Grunny:
LGTM.

PR #8735, opened 2015-10-09
CE-2853 Fix XSS vulnerabilities in LookupContribs

adamkarminski:
/cc @Grunny @macbre

### DIFF
--- a/extensions/wikia/LookupContribs/templates/main-form.tmpl.php
+++ b/extensions/wikia/LookupContribs/templates/main-form.tmpl.php
@@ -58,7 +58,7 @@ $(document).ready(function() {
 			{ "bVisible": true,  "aTargets": [ 2 ], "bSortable" : true, "sClass": "lc-datetime" },
 			{
 				"fnRender": function ( oObj ) {
-					var row = '<span class="lc-row"><a href="' + mw.html.escape(oObj.aData[3])+ '">' + oObj.aData[3] + '</a></span>';
+					var row = '<span class="lc-row"><a href="' + mw.html.escape(oObj.aData[3])+ '">' + mw.html.escape(oObj.aData[3]) + '</a></span>';
 					row += '&nbsp;(<a href="' + mw.html.escape(oObj.aData[3]) + 'index.php?title=Special:Contributions/' + mw.html.escape(encodeURIComponent(username)) + '">';
 					row += '<?= wfMessage( 'lookupcontribscontribs' )->escaped() ?>';
 					row += '</a>)</span>';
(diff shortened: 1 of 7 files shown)


extensions/wikia/LookupContribs/templates/main-form.tmpl.php
```diff
@@ -2,8 +2,8 @@
 <!-- DISTRIBUTION TABLE -->
 <script type="text/javascript" charset="utf-8">
 function lcShowDetails(dbname, mode) {
-	var username = '<?= urlencode( $username ) ?>';
-	var action = '<?=$action?>';
+	var username = '<?= Sanitizer::cleanUrl( $username ) ?>';
+	var action = '<?= Sanitizer::cleanUrl( $action ) ?>';
```
Grunny: Both of these should be `Xml::escapeJsString`. `Sanitizer::cleanUrl` actually unescapes HTML entities... The username should then be urlencoded when it is being used in the URL with `encodeURIComponent` (or just wrap it in `encodeURIComponent` here).

extensions/wikia/LookupContribs/templates/main-form.tmpl.php
```diff
@@ -11,12 +11,12 @@ function lcShowDetails(dbname, mode) {
...
 	}
 
-	document.location.href = action + '?target=' + username + '&wiki=' + dbname + '&mode=' + mode;
+	document.location.href = action + '?target=' + username + '&wiki=' + mw.html.escape( dbname ) + '&mode=' + mw.html.escape( mode );
 }
 
 $(document).ready(function() {
 	var baseurl = wgScript + "?action=ajax&rs=LookupContribsAjax::axData";
-	var username = '<?= urlencode( $username ) ?>';
+	var username = '<?= htmlspecialchars( $username, ENT_QUOTES ) ?>';
```
Grunny: Once again should be `Xml::escapeJsString`. You're correctly escaping the username when it is added to HTML with `mw.html.escape` so also escaping here will lead to double escaping (safe in general, but could cause issues :)).

extensions/wikia/LookupContribs/templates/main-form.tmpl.php
```diff
@@ -58,9 +58,9 @@ function lcShowDetails(dbname, mode) {
 			{ "bVisible": true,  "aTargets": [ 2 ], "bSortable" : true, "sClass": "lc-datetime" },
 			{
 				"fnRender": function ( oObj ) {
-					var row = '<span class="lc-row"><a href="' + oObj.aData[3] + '">' + oObj.aData[3] + '</a></span>';
-					row += '&nbsp;(<a href="' + oObj.aData[3] + 'index.php?title=Special:Contributions/' + username + '">';
-					row += '<?=wfMsg('lookupcontribscontribs')?>';
+					var row = '<span class="lc-row"><a href="' + mw.html.escape( oObj.aData[3] ) + '">' + mw.html.escape( oObj.aData[3] ) + '</a></span>';
+					row += '&nbsp;(<a href="' + mw.html.escape( oObj.aData[3] ) + 'index.php?title=Special:Contributions/' + mw.html.escape( username ) + '">';
```
Grunny: Since this is a URL parameter, might be best to `mw.html.escape( encodeURIComponent( username ) )` for correctness of the URL.

PR #8736, opened 2015-10-09
SOC-1430 | Fix regex for redirect after success postMessage from auth…

kvas-damian:
…Pages
https://wikia-inc.atlassian.net/browse/SOC-1430


rafalkalinski:
LGTM

PR #8737, opened 2015-10-09
CE-2854 Fix XSS vulnerabilities in WhereIsExtension

adamkarminski:
/cc @Grunny @macbre

### DIFF
--- a/extensions/wikia/WhereIsExtension/templates/list.tmpl.php
+++ b/extensions/wikia/WhereIsExtension/templates/list.tmpl.php
@@ -117,16 +117,16 @@ select {
 				?>
 					<li class="wikiList">
 						<input type="checkbox" name="wikiSelected[]" id="wikiSelected_<?= Sanitizer::encodeAttribute( $wikiID ) ?>" value="<?= Sanitizer::encodeAttribute( $wikiID ) ?>">
-						<a href="<?= rawurlencode( $editURL ) ?>" <?= $style ?>><?= wfMessage( 'whereisextension-edit' )->escaped() ?></a>
-						<a href="<?= rawurlencode( $wikiInfo['u'] ) ?>" <?= $style ?>>
+						<a href="<?= Sanitizer::encodeAttribute( $editURL ) ?>" <?= $style ?>><?= wfMessage( 'whereisextension-edit' )->escaped() ?></a>
+						<a href="<?= Sanitizer::encodeAttribute( $wikiInfo['u'] ) ?>" <?= $style ?>>
 							<?= htmlspecialchars( $wikiInfo['t'] ) ?>
 						</a> (<?= htmlspecialchars( $wikiInfo['u'] ) ?>)
 					</li>
 				<? endforeach; ?>
 			</ul>
                         <?= $sPager ?>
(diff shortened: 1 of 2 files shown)


extensions/wikia/WhereIsExtension/templates/list.tmpl.php
```diff
@@ -138,7 +149,7 @@
 
 	$.loadJQueryAutocomplete(function() {
 		$('#wikiSelectTagName').autocomplete({
-			serviceUrl: wgServer+wgScript+'?action=ajax&rs=WikiFactoryTags::axQuery',
+			serviceUrl: mw.config.get( 'wgServer' ) + mw.config.get( 'wgScript' ) + encodeURIComponent('?action=ajax&rs=WikiFactoryTags::axQuery'),
```
Grunny: `encodeURIComponent` here will break the URL as it will encode the `?` and so forth. It's not needed anyway.

extensions/wikia/WhereIsExtension/templates/list.tmpl.php
```diff
@@ -197,7 +208,7 @@ function(e) {
 		$.ajax({
 			type:"POST",
 			dataType: "json",
-			url: ajaxpath+'?action=ajax&rs=axWFactoryFilterVariables' + values,
+			url: ajaxpath + encodeURIComponent('?action=ajax&rs=axWFactoryFilterVariables' + values),
```
Grunny: `encodeURIComponent` here will break the URL as it will encode the `?` and so forth. It should instead be used on the values being set earlier and passed here, i.e.

``` javascript
        values += '&group=' + encodeURIComponent($('#groupSelect').val());
        values += '&string=' + encodeURIComponent($('#withString').val());
```

extensions/wikia/WhereIsExtension/templates/list.tmpl.php
```diff
@@ -34,88 +34,99 @@
...
-			<?php
-			$front = '&nbsp;<a href="' . Title::makeTitle( NS_SPECIAL, 'WikiFactory' )->getFullUrl() . '/';
-			$back = '/variables/' . $formData['vars'][ $formData['selectedVar'] ] . '">[edit]</a>';
-			foreach($formData['wikis'] as $wikiID => $wikiInfo) {
-				$editURL = $front . $wikiID . $back;
+				<?
+					foreach( $formData['wikis'] as $wikiID => $wikiInfo ) :
+						$style = !$wikiInfo['p'] ? 'style="color: red"' : '';
+						$editURL = Title::makeTitle( NS_SPECIAL, 'WikiFactory' )->getFullURL() . '/' . $wikiID;
+						$editURL .= '/variables/' . $formData['vars'][ $formData['selectedVar'] ]
```
Grunny: This can be simplified to: `SpecialPage::getTitleFor( 'WikiFactory', "{$wikiID}/variables/{$formData['vars'][ $formData['selectedVar'] ]} )->getFullURL();`

Grunny: This would also be a fatal error due to a missing semi colon. Was it tested?

PR #8739, opened 2015-10-09
SOC-1217 | HAWelcome task - fix user profile page creation when WidgetTips are enabled

kvas-damian:
WidgetTips throws fatal error when there is no Title object in RequestContext ( https://github.com/Wikia/app/blob/dev/extensions/wikia/WidgetFramework/Widgets/WidgetTips/WidgetTips.php#L26 ), this is the fix.

https://wikia-inc.atlassian.net/browse/SOC-1217


rafalkalinski:
LGTM

jsutterfield:
:+1:

PR #8741, opened 2015-10-09
enable preference service in reston

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-732
See Wikia/config#1413

`user-preference` is now deployed in Reston, so the mediawiki app can read from the service in Reston now.


drsnyder:
lgtm

PR #8742, opened 2015-10-09
[SERVICES-908] Increases the length of the temporary generated password from 10 characters to 20.

gbenson-wikia:
NOTE: This is an emergency fix, so I’m specifically not trying to make
the generated passwords better yet.


ArturKlajnerok:
Lgtm :+1:

PR #8744, opened 2015-10-09
Video (OPS-7272): Update api url for ustream and viddler

saipetch:


extensions/wikia/VideoHandlers/apiwrappers/ViddlerApiWrapper.class.php
```diff
@@ -2,7 +2,7 @@
 
 class ViddlerApiWrapper extends ApiWrapper {
 	protected static $RESPONSE_FORMAT = self::RESPONSE_FORMAT_PHP;
-	protected static $API_URL = 'http://api.viddler.com/api/v2/viddler.videos.getDetails.php?key=$1&add_embed_code=1&url=';
+	protected static $API_URL = 'http://api.viddler.com/api/v2/viddler.videos.getDetails.php?key=&add_embed_code=1&url=';
```
armonr: Should the key be removed? This does not look like it'd work

armonr:
Please remove the `key=` from the urls as well, then, LGTM

PR #8745, opened 2015-10-09
Don't log as error when given a null revision

drsnyder:
We short circuit the hook when this happens. This is probably not an error we need to act on.

https://wikia-inc.atlassian.net/browse/MAIN-5351

/cc @macbre


macbre:
:+1:

PR #8747, opened 2015-10-12
ADEN-2535 Implemented message design

fraszczakszymon:


extensions/wikia/AdEngine/js/recovery/message.js
```diff
@@ -14,14 +20,54 @@ define('ext.wikia.adEngine.recovery.message', [
...
+			'We can bring you fun, free, fan-oriented content until my glorious return to Earh if you ' +
+			'<strong>add us to your adblock whitelist</strong>. ' +
+			'<a class="action-accept">Click</a> my face to refresh after you\'re done!';
+
+	function accept() {
+		adTracker.track('recovery/message', 'accept');
+		location.reload();
+	}
+
+	function reject(message) {
```
nandy-andy: What do you think of changing `message` to `messageContainer` or `msgContainer`? I thinking about such semantic change because we're passing and hiding the element with the message not only the message ;)

extensions/wikia/AdEngine/js/recovery/message.js
```diff
@@ -14,14 +20,54 @@ define('ext.wikia.adEngine.recovery.message', [
...
+	}
+
+	function isRejected() {
+		return !!localStorage.getItem(localstorageKey);
+	}
 
 	function createMessage(uniqueClassName) {
 		var className = 'recovered-message',
-			div = doc.createElement('div');
+			div = doc.createElement('div'),
```
nandy-andy: I think at this point it's better to move everything to a mustache template and request it. We're starting to have more and more spaghetti here ;) Moving it to a template comes with performance bonus points, I guess: the template response can be cached :)

skins/oasis/css/core/ads.scss
```diff
@@ -270,16 +272,88 @@ body >a >img[src$="noad.gif"] {
...
-	font-size: 2em;
 	text-align: center;
 	padding: 10px 0;
+	min-height: 100px;
+
+	.message {
+		background: #ffffff;
+		color: #000000;
+		padding: 10px 12px;
+		border: 2px solid #000000;
```
nandy-andy: Please shorten all the color hexes from `#ffffff` to `#fff`, `#000000` to `#000` and so on.

Additional question is: how does it look on light and dark wikis? If it doesn't look good on dark wikis then it makes sense to define variables for the colors i.e.:

```
$message-background-color-light: #000;
$message-background-color-dark: #fff;

$message-text-color-light: #000;
$message-text-color-dark: #fff;
```

skins/oasis/css/core/ads.scss
```diff
@@ -270,16 +272,88 @@ body >a >img[src$="noad.gif"] {
...
+		.dialog-pointer {
+			width: 15px;
+			height: 15px;
+			border-top: 2px solid #000000;
+			border-left: 2px solid #000000;
+			position: absolute;
+			background: #ffffff;
+			top: -10px;
+			left: 110px;
+			-ms-transform: rotate(45deg);
```
nandy-andy: You can use [transform mixin](https://github.com/Wikia/app/blob/dev/skins/shared/mixins/transform.scss) instead of all those vendor prefixed properites :smiley_cat:

extensions/wikia/AdEngine/js/recovery/message.js
```diff
@@ -1,44 +1,96 @@
...
+		});
 	}
 
 	function injectRightRailMessage() {
-		var rail = doc.getElementById(wikiaRailId),
-			message = createMessage('right-rail');
-
 		log('recoveredAdsMessage.recover - injecting right rail message', 'debug', logGroup);
-		rail.insertBefore(message, rail.firstChild);
+		createMessage('right-rail').done(function (messageContainer) {
```
nandy-andy, line 95: We have to be careful with promises. And while reading through the changes I was wondering if we should add handling failures everywhere when we use promises or only here. What do you think?

extensions/wikia/AdEngine/templates/recovered_message.mustache
```diff
@@ -0,0 +1,9 @@
+<div class="recovered-message {{positionClass}}">
+    <img src="/skins/oasis/images/recovered_message_icon.png" class="icon">
+    <div class="message">
+        <div class="dialog-pointer"></div>
+        <h3>{{{header}}}</h3>
```
nandy-andy: I don't think we have to unescape it. The `{{{text}}}` will have HTML but here we can make it more safe and escape it as usual with `{{` instead of `{{{`.

extensions/wikia/AdEngine/templates/recovered_message.mustache
```diff
@@ -0,0 +1,9 @@
+<div class="recovered-message {{positionClass}}">
+    <img src="/skins/oasis/images/recovered_message_icon.png" class="icon">
```
nandy-andy: Hm, now I wonder why we've put it there. More common way is to put an extension's image in the extension's directory. And we used to have a way, a variable which constructs the URL... `$wgUrlPath` or something in PHP. @macbre, @vforge can you help me refresh old man's memory? Or maybe there is a newer and better way to do it?

Also, let's add there `width` and `height` parameters it's supposed to save some reflows/repaints and we're up for webperf! :smile_cat:

vforge: I would use icon as a background for `div` itself with `wgCdnStylePath`

nandy-andy: `$wgCdnStylePath` that was it! Thanks, @vforge! :smile_cat:

vforge:
:+1:

nandy-andy:
LGTM :+1:

nandy-andy:
Looks eveng more good TM ;)

PR #8748, opened 2015-10-12
ApiQueryUsers: allow user id in addition to user name

macbre:
Allow Pandora services to query for user information (block info, groups and rights) by user ID.

[The example API request](http://poznan.macbre.wikia-dev.com/api.php?action=query&list=users&usids=0|1|119245&usprop=groups|rights|blockinfo)

@michalroszka / @wladekb / @jcellary

### DIFF
--- a/includes/api/ApiQueryUsers.php
+++ b/includes/api/ApiQueryUsers.php
@@ -84,6 +84,19 @@ class ApiQueryUsers extends ApiQueryBase {
 		$users = (array)$params['users'];
 		$goodNames = $done = array();
 		$result = $this->getResult();
+
+		// Wikia change - begin
+		// @see PLATFORM-1561
+		$ids = (array)$params['ids'];
+
+		foreach( $ids as $id ) {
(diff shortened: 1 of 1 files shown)


includes/api/ApiQueryUsers.php
```diff
@@ -84,6 +84,19 @@ public function execute() {
...
 		$goodNames = $done = array();
 		$result = $this->getResult();
+
+		// Wikia change - begin
+		// @see PLATFORM-1561
+		$ids = (array)$params['ids'];
+
+		foreach( $ids as $id ) {
+			$u = User::newFromId( $id );
+			if ( $u->isLoggedIn() ) {
```
ghost: `isLoggedIn` is kind of misleading as despite what it really means it suggests something like 'is the user logged in or not at the moment'. I would use `! isAnon` here.

macbre: Good point, addressed in 3cad498 (and comment added)

ghost:
:+1:

PR #8749, opened 2015-10-12
SOC-1435 Refactor this extension to make it easier to read & disable bounce handling

garthwebb:
Note that this extension essentially does nothing now that I've disabled the bounce handling.  I'm aware of this and I cleaned it up because we'll likely add new behavior in the near future when we decide what to do with the bounce info that SendGrid gives us.

Also, you can test this endpoint by sending a POST request here:

```
http://community.garth.wikia-dev.com/wikia.php?controller=SendGridPostback&method=log
```

The minimum data needed to be a valid postback is:

```
[{"event": "bounce"}]
```

This can be achieved via the `curl` request:

```
curl -d '[{"event": "bounce"}]' -X POST 'http://community.garth.wikia-dev.com/wikia.php?controller=SendGridPostback&method=log'
```

https://wikia-inc.atlassian.net/browse/SOC-1435

### DIFF
--- a/includes/wikia/api/SendGridPostBackApiController.class.php
+++ b/includes/wikia/api/SendGridPostBackApiController.class.php
@@ -1,91 +1,115 @@
 <?php
 /**
- * API which SendGrid will post data to when an email is opened, clicked, bounced back, unsubscribed, marked as spam, etc..
+ * API which SendGrid will post data to when an email is opened, clicked, etc..
  *
- * @author Piotr Molski ( moli@wikia-inc.com )
+ * For more information see:
+ *
+ *   https://sendgrid.com/docs/API_Reference/Webhooks/event.html
  */
+
+use Wikia\Logger\WikiaLogger;
 
 class SendGridPostbackController extends WikiaApiController {
 
-	public function safe_get($arr, $key, $default = null) {
-		return isset($arr[$key]) ? $arr[$key] : $default;
-	}
-
+	/**
+	 * Requests from SendGrid are POSTed via this controller.  The POST data contains a JSON
+	 * encoded associative array with the following keys:
+	 *
+	 * - email
+	 * - wikia-email-id
+	 * - wikia-email-city-id
+	 * - wikia-db
+	 * - event
+	 * - url
+	 * - status
+	 * - reason
+	 *
+	 * NOTE: The 'response' key set before returning is for debugging purposes only.  SendGrid
+	 * only requires that we send back a 200 response code.
+	 *
+	 */
 	public function log() {
-		wfProfileIn( __METHOD__ );
-
-		// Sendgrid uses raw JSON POST format, this is how PHP ingests it
-		$data = file_get_contents("php://input");
-		$events = json_decode($data, true);  // true = return value is array
-
-		if (empty($events)) {
-			\Wikia\Logger\WikiaLogger::instance()->debug( __CLASS__ . ": No data to process");
-			wfProfileOut( __METHOD__ );
+		if ( !$this->wg->Request->wasPosted() ) {
+			// Log the token-validation problem.
+			WikiaLogger::instance()->debug( __CLASS__ . ': Request method must be POST' );
+			$this->response->setVal( 'response', 'Request method must be POST' );
 			return;
 		}
 
-		if (!is_array($events)) {
-			\Wikia\Logger\WikiaLogger::instance()->debug( __CLASS__ . ": Supplied data is not an array", [
-				'data_first_100_chars' => is_string($data) ? substr($data,0,100) : '[type: ' . gettype($data) . ']'
-			]);
-			wfProfileOut( __METHOD__ );
+		$events = $this->getEvents();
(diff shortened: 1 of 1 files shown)


includes/wikia/api/SendGridPostBackApiController.class.php
```diff
@@ -1,91 +1,115 @@
...
+ *
+ *   https://sendgrid.com/docs/API_Reference/Webhooks/event.html
  */
 
-class SendGridPostbackController extends WikiaApiController {
+use Wikia\Logger\WikiaLogger;
 
-	public function safe_get($arr, $key, $default = null) {
-		return isset($arr[$key]) ? $arr[$key] : $default;
-	}
```
garthwebb, line 12: The above function was moved to the end of the file and renamed.

includes/wikia/api/SendGridPostBackApiController.class.php
```diff
@@ -1,91 +1,115 @@
...
+	 *
+	 * NOTE: The 'response' key set before returning is for debugging purposes only.  SendGrid
+	 * only requires that we send back a 200 response code.
+	 *
+	 */
 	public function log() {
-		wfProfileIn( __METHOD__ );
-
-		// Sendgrid uses raw JSON POST format, this is how PHP ingests it
-		$data = file_get_contents("php://input");
```
garthwebb, line 18: Getting and verifying the `$events` array we get from `$data` was moved into its own method, `getEvents`, further below

includes/wikia/api/SendGridPostBackApiController.class.php
```diff
@@ -1,91 +1,115 @@
...
-				'data_first_100_chars' => is_string($data) ? substr($data,0,100) : '[type: ' . gettype($data) . ']'
-			]);
-			wfProfileOut( __METHOD__ );
+		$events = $this->getEvents();
+		if ( empty( $events ) ) {
+			$this->response->setVal( 'response', 'Unable to find any events' );
 			return;
 		}
 
-		foreach ($events as $event) {
```
garthwebb, line 35: The only thing this code did was move certain data from one array to another.  There's no need to do this when passing to WikiaLogger and we can pull off the individual values when we need them.  I gave a link to the docs above and listed the keys we use for future reference.

includes/wikia/api/SendGridPostBackApiController.class.php
```diff
@@ -1,91 +1,115 @@
...
-					"url"          => $url,
-					"status"       => $status,
-					"reason"       => $reason
-				);
-
-				// Log to logstash/kibana
-				\Wikia\Logger\WikiaLogger::instance()->info( "SendgridPostback", $insert_data);
-
-				// On bounce, invalidate the users email to force them to re-verify it
-				if ( $event_type == 'bounce' ) {
```
garthwebb, line 63: Everything in this conditional was moved to the `handleBounce` method

includes/wikia/api/SendGridPostBackApiController.class.php
```diff
@@ -1,91 +1,115 @@
...
+		foreach ( $events as $event ) {
+			$eventType = $this->safeGet( $event, 'event' );
+
+			// Log to logstash/kibana
+			WikiaLogger::instance()->info( 'SendgridPostback', $event );
+
+			// On bounce, invalidate the users email to force them to re-verify it
+			if ( $eventType == 'bounce' ) {
+				// SOC-1435 : For now don't invalidate email on bounce
+				# $this->handleBounce( $event );
```
kvas-damian, line 54: is there any reason why we used `#` for commenting out code here?

kvas-damian:
LGTM

includes/wikia/api/SendGridPostBackApiController.class.php
garthwebb, line 54: Only that its visually distinct from the comment above it, so it didn't look like it was part of the comment.  Otherwise I don't really feel strongly about it.

includes/wikia/api/SendGridPostBackApiController.class.php
```diff
@@ -1,91 +1,115 @@
...
+ * API which SendGrid will post data to when an email is opened, clicked, etc..
  *
- * @author Piotr Molski ( moli@wikia-inc.com )
+ * For more information see:
+ *
+ *   https://sendgrid.com/docs/API_Reference/Webhooks/event.html
  */
 
-class SendGridPostbackController extends WikiaApiController {
+use Wikia\Logger\WikiaLogger;
```
jsutterfield, line 10: So this isn't super important, but another approach is to use the `use Loggable` trait inside the class. When you do that, you have access to logging methods like:

``` php
$this->debug();
$this->info();
$this->warn();
```

It's pretty nifty.

jsutterfield:
LGTM

garthwebb:
@jsutterfield ah, that's a good way to do it.  I'm going to try to get this out soon but I'll use this next time!

PR #8752, opened 2015-10-12
ADEN-2533 Add translations to new extension

nandy-andy:
Message files and CrowdIn config file for new extension with message to users with ad blocking software.


extensions/wikia/AdEngine/crowdin.conf
```diff
@@ -0,0 +1,4 @@
+[crowdin.project]
+projectid = "ext-AdEngine"
```
nandy-andy, line 2: @Grunny is it easy enough to "extract" this module to a separate extension in future? Does it adds additional work for our translators or it's fine if at one point in the future we decide to extract it and create new project and change the files' structure in `Wikia/app`? What do you think?

extensions/wikia/AdEngine/i18n/AdEngine.i18n.php
```diff
@@ -0,0 +1,12 @@
+<?php
+$messages = [];
+
+$messages[ 'en' ] = [
+	'adengine-recovery-message-blocking-welcome' => 'Hey! It looks like you\'re using ad blocking software!',
+	'adengine-recovery-message-blocking-message' => 'Wikia runs ads so we can keep the lights on and so Wikia will always be free to use. We can bring you fun, free, fan-oriented content until my glorious return to Earh if you <strong>add us to your adblock whitelist</strong>. <a class="action-accept">Click</a> my face to refresh after you\'re done!',
```
nandy-andy: @Grunny @lgarczewski @adamkarminski do you have any idea how to get rid-off the HTML here? I recall it violates i18n good practices. I think we could use WikiText to bold the text within `<strong />` tag but I don't know what to do with the anchor...

extensions/wikia/AdEngine/crowdin.conf
Grunny, line 2: We can upload existing translations, so if you split it to a new project in the future, you'd just make sure any existing translations for the messages that are moved are uploaded as well and it shouldn't create any more work for translators.

PR #8753, opened 2015-10-12
Video (SEO-101): Ooyala - update maintenance script to update asset url

saipetch:


maintenance/wikia/VideoHandlers/updateOoyalaAssetUrl.php
```diff
@@ -105,16 +117,13 @@ function updateRemoteAssetUrls( $ingester, $video ) {
...
 
-$apiPageSize = 100;
+$apiPageSize = 500;
 if ( !empty( $limit ) && $limit < $apiPageSize ) {
 	$apiPageSize = $limit;
 }
 
 $ingester = FeedIngesterFactory::getIngester( $source );
 
-$extra[] = "metadata.source = '$source'";
```
gabrys, line 115: Is it faster to filter the results in PHP than on the db level?

maintenance/wikia/VideoHandlers/updateOoyalaAssetUrl.php
```diff
@@ -105,16 +117,13 @@ function updateRemoteAssetUrls( $ingester, $video ) {
...
-$extra[] = "asset_type = 'remote_asset'";
```
gabrys, line 116: Do we not need this condition anymore?

gabrys:
LGTM

PR #8755, opened 2015-10-13
VOLDEV-145 - Implement article header changes and new rail module

onei:
@grunny (Moved from #8301) see [VOLDEV-145](https://wikia-inc.atlassian.net/browse/VOLDEV-145)

@Warkot I managed to fix the blank pageload issue, caused by interacting with the `ArticleViewCustom` hook which affected how articles are purged but not how they're saved after being edited. Could you take another look?

Summary of changes:
- Move edit, and other, buttons back above page header border (also fixes CONCF-501).
- Move the page count tally to it's own sidebar module, along with the contribute button and dropdown. The tally remains for Special:Videos due to it being used for other purposes.
- Fixed various code issues inline with current style guidelines and code deprecations.
- Altered the `PageHeaderIndexAfterActionButtonPrepared` and `PageHeaderIndexExtraButtons` hooks to only allow modifications of the intended variables and reduce the potential for subtle, undocumented behaviour/bugs.
- Per a TODO in `PageHeader_Index.php`, extension specific buttons were moved to hook callbacks, specifically for SpecialVideos and WikiaNewFiles.
- Added icons to the user talk archive edit button see when Wall is enabled.
- Removed unused sass and applied general clean up.
- Removed use of globals in controller methods.
- Radically overhauled the logic for `PageHeaderController` and it's templates.
- Fixed a security weakness that allowed wikis to modify URLs used in the share buttons for hub wikis due to the URLs being stored in messages. They are now always taken from the database and cannot be customised.
- Delete `PageHeader_EditBox.php` and it's associated controller method as it was unused.
- Removed unused `ArticleSaveComplete` hook callback in `PageHeaderController`.
- Removed tally from Hubs page headers (per the brief) and hid the page title as both were hidden using local CSS anyway.

To do:
- Fix location of new contribute rail module.
- Remove temporary code for loading sass in `PageHeaderController`.
- Remove temporary code for testing the page tally in `ContributeController`.
- Fix edit button alignment - H1 font-size is current set as part of the oasis typography (and thus varies between screen widths) which causes the edit button alignment to be wrong for anything less than the max width font-sizes.


Warkot:
We need to reprioritize the ticket in our backlog, which is curated by @hakubo. However, if I have some spare time, I can take a glimpse. First thing I can see is that this branch has conflicts with `dev`. :wink:

onei:
@warkot Done :) I'd be grateful if you could take a look although there are still a couple of unresolved issues as mentioned in the "To do" above which might be worth addressing before too much work is put into reviewing this.

ghost:
I am closing this one, as the changes have been reapplied in https://github.com/Wikia/app/pull/9609

PR #8757, opened 2015-10-13
WMU_loadDetails: make a POST request with an edit token

macbre:
https://wikia-inc.atlassian.net/browse/MAIN-5576

Fix a regression caused by #8672

@michalroszka


macbre:
Backported to preview 0077d08

PR #8761, opened 2015-10-13
SERVICES-873: Added log when MW is falling back from Helios auth to legacy one

harnash:
We need to know how many cases there are when for some reason user can't authenticate using Helios service but gets logged in by MW anyway using cookie/token/whatever. We need that to safely force Helios as one source of authentication.

One caveat of this solution is that we get a lot of those logs when user is beeing logged over and over again using native/legacy MW mechanisms.
@Wikia/services-team

### DIFF
--- a/includes/context/RequestContext.php
+++ b/includes/context/RequestContext.php
@@ -203,24 +203,59 @@ class RequestContext implements IContextSource {
 	public function getUser() {
 		// Wikia change - begin - @author: Michał Roszka
 		global $wgEnableHeliosExt;
+		$authPath = [];
+
 		if ( $this->user === null && $wgEnableHeliosExt ) {
 			$this->user = \Wikia\Helios\User::newFromToken( $this->getRequest() );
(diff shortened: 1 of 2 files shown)


nmonterroso:
LGTM

includes/context/RequestContext.php
```diff
@@ -203,27 +203,64 @@ public function setUser( User $u ) {
 	public function getUser() {
 		// Wikia change - begin - @author: Michał Roszka
 		global $wgEnableHeliosExt;
+		$authPath = [];
+
 		if ( $this->user === null && $wgEnableHeliosExt ) {
 			$this->user = \Wikia\Helios\User::newFromToken( $this->getRequest() );
+			$authPath['helios'] = ($this->user !== null ? 'OK' : 'FAIL');
```
drsnyder, line 210: I wonder if this should be `$authType = "helios"`. Then [below](https://github.com/Wikia/app/pull/8761/files#diff-f3b309ab0fd7d1151c05920cae0462ddR253) we could use `auth` as a facet in ES. As it is now, it's hard to graph in kibana.

PR #8764, opened 2015-10-13
don't save language to the user_properties table

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-733

Even though the preference service is the authority for preference data, we're currently still saving the default value for language (`en`) for every user. This is because of `$wgGlobalUserProperties` usage in [`User->shouldOptionBeStored`](https://github.com/Wikia/app/blob/dev/includes/User.php#L5052), so we should remove the `language` entry from the global and maintain that configuration in the `PreferenceModule`

Since this was the only usage of `$wgGlobalUserProperties`, we can also remove that global now. :tada:

### DIFF
--- a/includes/User.php
+++ b/includes/User.php
@@ -5050,9 +5050,7 @@ class User {
 	 * @return bool
 	 */
 	private function shouldOptionBeStored( $key, $value ) {
-		global $wgGlobalUserProperties;
 		if (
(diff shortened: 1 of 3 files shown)


ArturKlajnerok:
Lgtm :+1:

PR #8766, opened 2015-10-13
SEO-108 Fix PHP Warning: sprintf(): Too few arguments in

gabrys:
/extensions/wikia/Paginator/Paginator.body.php on line 215


saipetch:
LGTM

PR #8768, opened 2015-10-14
SEO-68 Add redirect support for URLs missing a character

gabrys:
The SEO effect on redirecting from Mapping_keys_in_Vim_-_Tutorial_(Part_2 to
Mapping_keys_in_Vim_-_Tutorial_(Part_2) is minimal, it's mostly about the UX.

A good "No such article page" should do the following:
1. Inform there's no such article
2. If there's a really good match (like just a missing character in the URL), suggest the article you wanted
3. Allow you to create the page (if appropriate)
4. Let you search for other pages

SEO-wise only the second part is important. If we believe there is an article that user/crawler meant to go to,
the page should serve HTTP 200 with a link to the suggested page and canonical URL set to the suggested page.
Otherwise this should be a regular HTTP 404.

Note:

We don't do a HTTP redirect because this means users cannot create page that matches some other URL. For example
if there's already /wiki/R2D20 users won't be able to create /wiki/R2D2 .

### DIFF
--- a/extensions/wikia/Custom404Page/Custom404PageBestMatchingPageFinder.class.php
+++ b/extensions/wikia/Custom404Page/Custom404PageBestMatchingPageFinder.class.php
@@ -52,7 +52,7 @@ class Custom404PageBestMatchingPageFinder {
 
 		$originalLowerCase = mb_strtolower( $originalTitle );
 		$originalLength = mb_strlen( $originalLowerCase );
-		$originalWords = trim( preg_replace( '/\\W+/', ' ', $originalLowerCase ) );
+		$originalWords = trim( preg_replace( '/\\W+/u', ' ', $originalLowerCase ) );
 		$mismatchingCaseTitles = [];
 		$matchingPrefixTitles = [];
 		$matchingPrefixTitlesIncludingSubPages = [];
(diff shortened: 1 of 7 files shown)


extensions/wikia/Custom404Page/Custom404PageBestMatchingPageFinder.class.php
```diff
@@ -0,0 +1,140 @@
...
+	 * @param $originalTitle   string
+	 * @param $suggestedTitles array of strings
+	 *
+	 * @return string|bool the best matching string or false
+	 */
+	public function findTheBestMatchingTitleText( $originalTitle, $suggestedTitles ) {
+
+		$originalLowerCase = mb_strtolower( $originalTitle );
+		$originalLength = mb_strlen( $originalLowerCase );
+		$originalWords = trim( preg_replace( '/\\W+/', ' ', $originalLowerCase ) );
```
gabrys: /u

extensions/wikia/Custom404Page/Custom404PageBestMatchingPageFinder.class.php
```diff
@@ -0,0 +1,140 @@
...
+
+		$namespaceId = $title->getNamespace();
+		$titleText = $title->getPrefixedText();
+
+		if ( $title->isContentPage() && !$title->exists() ) {
+			$queryResults = $this->query( $titleText, $namespaceId );
+
+			if ( $queryResults['numFound'] === 0 ) {
+				// Strip last (maybe incomplete?) word form the title
+				$titleTextStripped = trim( preg_replace( '/\\W*\\w+\\W*$/', '', $titleText ) );
```
gabrys: /u?

saipetch:
LGTM

PR #8769, opened 2015-10-14
HTTPS changeovers

natewatson999:
Wikia is planning to roll out support for HTTPS, and is actively developing the feature. However, in order to properly support HTTPS without creating warnings, all resources must be loaded over HTTPS instead of HTTP as well. The dependencies must either be changed at the same time or before the formal rollout. This pull request addresses some of those dependencies, mostly API links, as another step towards HTTPS support.


natewatson999:
Also, I tested all of the changes, they do not create any linking issues.

PR #8771, opened 2015-10-14
CE-2853 Enable marking templates as not-needing classification

adamkarminski:
This PR enables marking templates as not-needing classification and storing the information in `page_wikia_props`. If a template's ID is not present in the table we assume that the template is used and needs to be classified. The same goes for a property value (`props`) being `1`. Only if the template is deliberately marked as `0` - we consider it unused.

### DIFF
--- a/extensions/wikia/TemplateClassification/UnusedTemplates/UnusedTemplatesHandler.class.php
+++ b/extensions/wikia/TemplateClassification/UnusedTemplates/UnusedTemplatesHandler.class.php
@@ -2,7 +2,7 @@
 
 namespace Wikia\TemplateClassification\UnusedTemplates;
 
-class Handler extends \WikiaService {
+class Handler {
 	/**
 	 * Reflects a `propname` in the `page_wikia_props` table.
 	 */
--- a/extensions/wikia/TemplateClassification/UnusedTemplates/UnusedTemplatesHandler.class.php
+++ b/extensions/wikia/TemplateClassification/UnusedTemplates/UnusedTemplatesHandler.class.php
@@ -1,0 +1,158 @@
+<?php
+
+namespace Wikia\TemplateClassification\UnusedTemplates;
+
+class Handler extends \WikiaService {
+	/**
+	 * Reflects a `propname` in the `page_wikia_props` table.
+	 */
+	const PROPERTY_NAME = 11000;
+	const USED = 1;
+	const UNUSED = 0;
+
+	/**
+	 * Checks if a template is used and needs to be classified because of that.
+	 * Returns true if a property is NOT FOUND for a given pageId
+	 * or if the property's value indicates that it is used.
+	 *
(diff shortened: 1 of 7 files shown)


adamkarminski:
The failing test not related to the code.

extensions/wikia/TemplateClassification/UnusedTemplates/UnusedTemplatesHandler.class.php
```diff
@@ -0,0 +1,158 @@
+<?php
+
+namespace Wikia\TemplateClassification\UnusedTemplates;
+
+class Handler {
+	/**
+	 * Reflects a `propname` in the `page_wikia_props` table.
+	 */
+	const PROPERTY_NAME = 11000;
```
kamilkoterba, line 9: Is this ID or name? Can we use more verbose name?

adamkarminski, line 9: It's a property name according to how it's called in the table (look at the
 `page_wikia_props` schema)

extensions/wikia/TemplateClassification/UnusedTemplates/UnusedTemplatesHandler.class.php
```diff
@@ -0,0 +1,158 @@
...
+		$insertRows = [];
+
+		while ( $row = $results->fetchObject() ) {
+			if ( !isset( $row->title ) ) {
+				continue;
+			}
+
+			$title = \Title::newFromText( $row->title, NS_TEMPLATE );
+			if ( $title === null ) {
+				continue;
```
kamilkoterba, line 129: I think that's the place that may need logging?
Might be hard to debug if we find some titles not reaching properties settings.
Null case here is due to title not meeting Title::secureAndSplit method requirements.

adamkarminski, line 129: Then it should be logged on the `Title::newFromText` level. This is a VERY sanity check, since it would be impossible to get invalid results from the `page` table.

kamilkoterba:
LGTM

kamilkoterba:
Feel free to merge once tested.

includes/QueryPage.php
```diff
@@ -292,6 +292,17 @@ function recache( $limit, $ignoreErrors = true ) {
 		# Do query
 		$res = $this->reallyDoQuery( $limit, false );
 		$num = false;
+
+		/**
+		 * Wikia change begin
+		 * @author <adamk@wikia-inc.com>
+		 */
+		$resClone = clone $res;
+		wfRunHooks( 'QueryPageUseResultsBeforeRecache', [ $this, $resClone ] );
```
macbre, line 301: https://wikia-inc.atlassian.net/browse/MAIN-5661 :)

PR #8777, opened 2015-10-14
SERVICE-873: Sampling InfluxDb stats at 25%

harnash:
To be safe and not to overload Scribe we will sample InfluxDb data at 25%.

@Wikia/services-team


sqreek:
LGTM

PR #8778, opened 2015-10-14
Added X-Forwared-For for all HeliosClient calls (SERVICES-913)

harnash:
We need a way for Helios client and other services to identify source of the request even it was forwarded by Media Wiki. Now all calls through HeliosClient will get additional header `X-Forwarded-For` to allow backend services to detect if user's IP is blocked etc.

@Wikia/services-team

### DIFF
--- a/lib/Wikia/src/Service/Helios/HeliosClientImpl.php
+++ b/lib/Wikia/src/Service/Helios/HeliosClientImpl.php
@@ -60,6 +60,18 @@ class HeliosClientImpl implements HeliosClient
 		// Request URI pre-processing.
 		$uri = "{$this->baseUri}{$resourceName}?" . http_build_query($getParams);
 
+		// Appending the request remote IP for client to be able to
+		// identify the source of the remote request.
+		if ( isset( $extraRequestOptions['headers'] ) ) {
+			$headers = $extraRequestOptions['headers'];
+			unset( $extraRequestOptions['headers'] );
+		} else {
+			$headers = [];
+		}
+
+		global $wgRequest;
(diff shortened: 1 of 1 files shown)


lib/Wikia/src/Service/Helios/HeliosClientImpl.php
```diff
@@ -60,6 +60,18 @@ public function request( $resourceName, $getParams = [], $postData = [], $extraR
...
+		// identify the source of the remote request.
+		if ( isset( $extraRequestOptions['headers'] ) ) {
+			$headers = $extraRequestOptions['headers'];
+			unset( $extraRequestOptions['headers'] );
+		} else {
+			$headers = [];
+		}
+
+		global $wgRequest;
+		$headers['X-Forwarded-For'] = $wgRequest->getIP();
```
nmonterroso, line 73: are we guaranteed that `$wgRequest` is always present?

nmonterroso:
minor comment otw lgtm

lib/Wikia/src/Service/Helios/HeliosClientImpl.php
harnash, line 73: Yes. That what it looks like:

`````` php
if ( $wgCommandLineMode ) {
    $wgRequest = new FauxRequest( array() );

    wfDebug( "\n\nStart command line script $self\n" );
} else {
    # Can't stub this one, it sets up $_GET and $_REQUEST in its constructor
    $wgRequest = new WebRequest;
(...)
}```
``````

PR #8782, opened 2015-10-14
SEO-87 Link from Japanese Starwars back to English one

gabrys:
Two things happen in this pull request:
- English Star Wars wiki now only point crawlers to the Japanese Star Wars wiki on canonical URLs (the big red chunk of the pull request)
- Japanese Star Wars wiki now points crawlers back to the respective articles on English Star Wars wiki

This is still considered an experiment, the prototype of more maintainable solution will be worked on in SEO-99.


saipetch:
LGTM

PR #8789, opened 2015-10-15
WikiFactoryLoader: take wgDBname and wgDBCluster from city_list instead of city_variables

macbre:
@michalroszka / @Grunny / @owend


ghost:
:+1:

PR #8790, opened 2015-10-15
ADEN-2557 Track OX Bidder lookup success/error events

fraszczakszymon:


nandy-andy:
We should think how to refactor it but as a quick fix it looks fine.

PR #8792, opened 2015-10-15
SOC-1407 Finish decoupling attributes from the User class in MW

jsutterfield:
Clean up the User class so that all public attributes are now accessed through the user attribute service client. Additionally, all logic used to determine if an attribute should be saved or deleted has been moved into the UAS client as well.

I also added some improved logging at save time to try and identify cases where what we save to MW differs from what we are saving to the service. This should help debug the attribute discrepancies we were seeing from SOC-1285 and SOC-1283.

Ticket: https://wikia-inc.atlassian.net/browse/SOC-1407


includes/User.php
```diff
@@ -2816,18 +2821,6 @@ public static function localToGlobalPropertyName($property, $cityId = null, $sep
 		return sprintf("%s%s%s", $property, $sep, $cityId);
 	}
 
-	private function sanitizePropertyArray( $array_map ) {
```
jsutterfield, line 2825: This method wasn't being used anywhere so I just removed it.

includes/User.php
```diff
@@ -4927,52 +4921,53 @@ protected function loadOptions() {
...
-				 $this->mOptionOverrides[$attributeName] = $attributeValue;
-				 $this->mOptions[$attributeName] = $attributeValue;
-			}
-		}
+	/**
+	 * Save this user's preferences into the database.
+	 *
+	 * @see getGlobalPreference for documentation about preferences
+	 */
+	protected function savePreferences() {
```
jsutterfield, line 4948: Git is getting confused here. I didn't add this `savePreferences()` method, it existed before. You can see git thinks I deleted it [here](https://github.com/Wikia/app/pull/8792/files#diff-897fba1dd6b8edc3238762e35d0b7c71L4974) and then recreated it here.

includes/User.php
```diff
@@ -2740,7 +2740,12 @@ private function shouldGetAttributeFromService( $attributeName ) {
 	 * @see getGlobalAttribute for more documentation about attributes
 	 */
 	public function setGlobalAttribute($attribute, $value) {
-		$this->setOptionHelper($attribute, $value);
+		if ( $this->isPublicAttribute( $attribute ) ) {
+			$value = $this->replaceNewlineAndCRWithSpace( $value );
+			$this->userAttributes()->setAttribute( $this->getId(), new Attribute( $attribute, $value ) );
+		}
+
+		$this->setOptionHelper( $attribute, $value );
```
jsutterfield, line 2761: Continue to set global attributes in MW for the time being. This is necessary for the logging I added to try and identify conditions where MW and the Service are having different values saved for the same attribute.

When we do SOC-1408 (remove public attributes from the MW dbs) we'll make sure to stop setting them here.

lib/Wikia/src/Service/User/Attributes/UserAttributes.php
```diff
@@ -91,9 +97,50 @@ private function setInMemcache( $userId, $attributes ) {
...
+		}
+
 		$this->setInInstanceCache( $userId, $attribute );
-		$this->setInMemcache( $userId, $this->attributes[$userId] );
+	}
+
+	public function save( $userId ) {
+		$attributes = $this->loadAttributes( $userId );
+
+		// TODO When bulk updates are complete, convert this to a single request
```
kvas-damian: @jsutterfield Can you file a ticket for that?

lib/Wikia/src/Service/User/Attributes/UserAttributes.php
```diff
@@ -91,9 +97,50 @@ private function setInMemcache( $userId, $attributes ) {
...
+	 */
+	private function attributeShouldBeSaved( $name, $value ) {
+		return (
+			( is_null( $this->defaultAttributes[$name] ) ) && !( $value === false || is_null( $value ) ) ||
+			$value != $this->defaultAttributes[$name]
+		);
+	}
+
+	private function attributeShouldBeDeleted( $name, $value ) {
+		return isset( $this->defaultAttributes[$name] ) && $value == $this->defaultAttributes[$name];
```
kvas-damian, line 144: If there is no default value, should null attributes be deleted from service?

lib/Wikia/tests/Service/User/UserAttributesTest.php
```diff
@@ -72,14 +72,49 @@ public function testGetAttributesWithDefaultParameter() {
...
 	}
 
+	public function testSaveAttribute() {
+		$defaultAttributes = [
+			'defaultAttr1' => 'defaultAttr1Val',
+			'defaultAttr2' => 'defaultAttr2Val',
+		];
+
+		$this->service->expects( $this->once() )
+			->method( "delete" )
```
kvas-damian: Please convert double quotes to single ones

lib/Wikia/src/Service/User/Attributes/UserAttributes.php
```diff
@@ -91,9 +97,50 @@ private function setInMemcache( $userId, $attributes ) {
...
+	public function save( $userId ) {
+		$attributes = $this->loadAttributes( $userId );
+
+		// TODO When bulk updates are complete, convert this to a single request
+		$savedAttributes = [];
+		foreach( $attributes as $name => $value ) {
+			if ( $this->attributeShouldBeSaved( $name, $value ) ) {
+				$this->setInService( $userId, new Attribute( $name, $value ) );
+				$savedAttributes[$name] = $value;
+			} elseif ( $this->attributeShouldBeDeleted( $name, $value ) ) {
```
kvas-damian, line 120: Is there a case when we are not saving or deleting attribute?

jsutterfield, line 120: I don't believe so. In the original `User::saveOptions` code, those are the only 2 cases which exist: https://github.com/Wikia/app/blob/dev/includes/User.php#L5009-L5015.

kvas-damian, line 120: In this case do we need 2nd `if`?

lib/Wikia/src/Service/User/Attributes/UserAttributes.php
jsutterfield, line 144: I don't think so. Only because that's not the original behavior MW had.

When an attribute was set to a null value, it would [try and set the defaultValue](https://github.com/Wikia/app/blob/dev/includes/User.php#L2882-L2884) if it exists. But if no default existed, then it would be set as null.

Then when it came time to save, it would [check if an attribute had a default value](https://github.com/Wikia/app/blob/dev/includes/User.php#L5012-L5015) and delete it if so. If it's null without a default howefver, it would still get set in the db. So there may be places in MW where it expects/uses these null values

lib/Wikia/src/Service/User/Attributes/UserAttributes.php
```diff
@@ -91,9 +97,51 @@ private function setInMemcache( $userId, $attributes ) {
...
+	 * Returns true if either is true:
+	 * 1.) No default for the attribute and the value is either not false or not null
+	 * 2.) Value is different than the default
+	 * @param $name
+	 * @param $value
+	 * @return bool
+	 */
+	private function attributeShouldBeSaved( $name, $value ) {
+		return (
+			( is_null( $this->defaultAttributes[$name] ) ) && !( $value === false || is_null( $value ) ) ||
```
garthwebb, line 138: Are `false` or `null` not valid values for an attribute?

jsutterfield, line 138: Not according to how MW has been saving attributes up to this point. I extracted the logic from [here](https://github.com/Wikia/app/blob/dev/includes/User.php#L5058) which performs the same check. I don't 100% understand why this has been the case, but wanted to keep the logic consistent.

lib/Wikia/src/Service/User/Attributes/UserAttributes.php
jsutterfield, line 120: Oops, sorry, I think I misunderstood your question. Yes, there is a case. If the attribute doesn't have a default, and has a value of `false` or `null`, then we don't save or delete it. So we do need a second `if`.

includes/User.php
armonr, line 2761: Please add this comment in the code too.

lib/Wikia/src/Service/User/Attributes/UserAttributes.php
```diff
@@ -91,9 +97,51 @@ private function setInMemcache( $userId, $attributes ) {
 	 * @param Attribute $attribute
 	 */
 	public function setAttribute( $userId, Attribute $attribute ) {
-		$this->setInService( $userId, $attribute );
+		$this->loadAttributes( $userId );
```
armonr, line 100: Are there any usages of `$this->attributes` in this method? If not, why is this call needed?

lib/Wikia/src/Service/User/Attributes/UserAttributes.php
armonr: `is_null` only checks for nullity; it returns `false` for `false`

lib/Wikia/tests/Service/User/UserAttributesTest.php
```diff
@@ -58,36 +58,71 @@ public function testGetAttributes() {
...
+		$userAttributes->setAttribute( $this->userId, new Attribute( 'newAttr', 'foo' ) );
+		$this->assertEquals( 'foo', $userAttributes->getAttribute( $this->userId, 'newAttr' ) );
+		$userAttributes->setAttribute( $this->userId, new Attribute( 'anotherNewAttr', null ) );
+		$this->assertEquals( null, $userAttributes->getAttribute( $this->userId, 'anotherNewAttr' ) );
+	}
+
+	public function testSaveAttribute() {
+		$defaultAttributes = [
+			'defaultAttr1' => 'defaultAttr1Val',
+			'defaultAttr2' => 'defaultAttr2Val',
```
armonr: Please use constants/variables for these.

lib/Wikia/src/Service/User/Attributes/UserAttributes.php
jsutterfield, line 100: It's basically to make sure we have all attributes loaded into the instance cache (`$this->attributes)` for later on.

When we call setAttribute, we set the new value for the attribute in the service, and then in `this->attributes`.

If we haven't loaded attributes for a user yet, then the only attribute they'll have in `this->attributes` is the one that was just set. Then, if we call `getAttribute`, it'll first try and load attributes and, seeing that we already have something in `this->attributes` for the user [here](https://github.com/Wikia/app/blob/dev/lib/Wikia/src/Service/User/Attributes/UserAttributes.php#L61-L63), will return those attributes. Which means they'll only have the one attribute.

armonr, line 100: :+1: Thanks for clarifying.
It is usually preferable to keep side effects to a minimum, but of course this change postdates the side-effect-y code.

armonr:
LGTM :100:

lib/Wikia/src/Service/User/Attributes/UserAttributes.php
kvas-damian, line 138: Makes sense

kvas-damian:
LGTM

PR #8793, opened 2015-10-15
CE-2784 Add basic Template Classification entry point to edit page

adamkarminski:
This PR makes use of EditPage's hook `EditPageLayoutExecute` to inject HTML with the Template Classification edit entry point. I've had to make some moves around the mocked backend as well to simplify using it from both JS and PHP.

/cc: @Wikia/community-engineering @kamilkoterba


extensions/wikia/TemplateClassification/TemplateClassificationView.php
```diff
@@ -5,44 +5,48 @@
...
+		$templateType = $this->renderTemplateType( $articleId, $user );
+		return \MustacheService::getInstance()->render(
+			__DIR__ . '/templates/TemplateClassificationEditPageEntryPoint.mustache',
 			[
-				'articleId' => $articleId,
-			],
-			true, \WikiaRequest::EXCEPTION_MODE_RETURN
+				'header' => wfMessage( 'template-classification-type-header' ),
+				'templateType' => $templateType,
+			]
```
kamilkoterba, line 48: Use a partial to render the entry point itself.

extensions/wikia/EditPageLayout/templates/EditPageLayout_EditPage.php
```diff
@@ -124,6 +124,7 @@
 									</a>
 								</div>
 							<?php endif ?>
+							<?php if ( $extraPageControlsHtml ) print $extraPageControlsHtml ?>
```
kamilkoterba: Wrap in brackets or : ... endif;

kamilkoterba:
LGTM

extensions/wikia/TemplateClassification/TemplateClassificationView.php
adamkarminski, line 48: Would have to refactor the PHP code to a mustache template - later.

PR #8799, opened 2015-10-16
MAIN-4228: limit wall replies to 500 per thread

owend:
FETCHED_REPLIES_LIMIT is 500 which seems like a reasonable number of replies in a thread.
This patch just shuts off the edit box when that limit is reached.  A better long term fix is to add pagination.


extensions/wikia/Wall/templates/Wall_message.php
```diff
@@ -122,7 +122,11 @@
 					<? $i++; ?>
 				<? endforeach; ?>
 			<? endif; ?>
-			<?= $app->renderViewCached( 'WallController', 'reply', 'Wall_message'.$showReplyForm, array('showReplyForm' => $showReplyForm )); ?>
+			<? if( $repliesNumber < $repliesLimit ) {
+				echo $app->renderViewCached( 'WallController', 'reply', 'Wall_message'.$showReplyForm, array('showReplyForm' => $showReplyForm ));
+			} else {
+				echo "<div class=message-topic-error >" . wfMsgExt('wall-message-limit-reached', array('parsemag'), array( $repliesLimit ));
```
macbre, line 128: Where is this `<div>` tag closed?

owend, line 128: Oh, it's missing.  DIdn't affect the rendering/output so I didn't notice.  Will fix.

PR #8808, opened 2015-10-16
CE-2872 Ensure piggyback logs are always secure

adamkarminski:
Even if the Piggyback ext is disabled.


extensions/wikia/Piggyback/Piggyback.php
```diff
@@ -30,7 +30,6 @@
 $wgSpecialPages['Piggyback'] = 'Piggyback'; # Let MediaWiki know about your new special page.
 
 $wgSpecialPageGroups['Piggyback'] = 'users';
-$wgLogRestrictions['piggyback'] = 'piggyback';
```
Grunny, line 33: I'd consider leaving it in the extension anyway. Can't hurt for the duplication and makes sure it's obvious this is the case when looking at the code of the extension, otherwise could be confusing/concerning as to whether this is restricted.

Grunny:

PR #8812, opened 2015-10-16
SEO-3 Don't put page views with query string to NOT_AN_ARTICLE bucket

gabrys:
This gets us in par to what Tableau can do


saipetch:
LGTM

PR #8813, opened 2015-10-17
SOC-1446 Add edit token for forum admin tools

armonr:
/CC @Grunny @garthwebb @kvas-damian

### DIFF
--- a/extensions/wikia/Forum/Forum.class.php
+++ b/extensions/wikia/Forum/Forum.class.php
@@ -238,12 +238,6 @@ class Forum extends Walls {
 			return false;
 		}
 
-		if ( !F::app()->wg->request->wasPosted() ||
-			!F::app()->wg->User->matchEditToken( F::app()->wg->request->getVal( 'editToken' ) ) ) {
-			wfProfileOut( __METHOD__ );
-			return false;
-		}
-
 		Forum::$allowToEditBoard = true;
 
 		if ( $id == null ) {
--- a/extensions/wikia/Forum/Forum.class.php
+++ b/extensions/wikia/Forum/Forum.class.php
@@ -238,6 +238,12 @@ class Forum extends Walls {
 			return false;
 		}
 
+		if ( !F::app()->wg->request->wasPosted() ||
+			!F::app()->wg->User->matchEditToken( F::app()->wg->request->getVal( 'editToken' ) ) ) {
+			wfProfileOut( __METHOD__ );
+			return false;
+		}
+
 		Forum::$allowToEditBoard = true;
 
 		if ( $id == null ) {
--- a/extensions/wikia/Forum/ForumExternalController.class.php
+++ b/extensions/wikia/Forum/ForumExternalController.class.php
@@ -59,9 +59,10 @@ class ForumExternalController extends WallExternalController {
 		}
 
 		try {
-			$this->displayAnyWriteAccessError();
+			$this->checkWriteRequest();
 		} catch ( \BadRequestException $bre ) {
-			return true;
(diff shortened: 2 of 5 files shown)


extensions/wikia/Forum/Forum.class.php
```diff
@@ -238,6 +238,12 @@ protected function createOrEditBoard( $board, $titletext, $body, $bot = false )
 			return false;
 		}
 
+		if ( !F::app()->wg->request->wasPosted() ||
+			!F::app()->wg->User->matchEditToken( F::app()->wg->request->getVal( 'editToken' ) ) ) {
```
Grunny: These can use `$this->wg` to access the globals, rather than `F::app`.

extensions/wikia/Forum/ForumSpecialController.class.php
```diff
@@ -140,6 +140,11 @@ public function createNewBoardModal() {
 						'maxlength' => '255'
 					),
 				),
+				array(
+					'type' => 'hidden',
+					'name' => 'token',
+					'value' => F::app()->wg->User->getEditToken(),
```
Grunny: `$this->getUser()->getEditToken()` would be preferred here and below.

Grunny:
I can see how this simplifies the checks, but in general, I think it would be better if the controllers were responsible for checking the validity of requests (which could be done by using [`$this->checkWriteRequest()`](https://github.com/Wikia/app/blob/3c714f0e8d3/includes/wikia/nirvana/WikiaDispatchableObject.class.php#L189-L202) to them), just as the ForumExternalController is also responsible for checking permissions right now through `checkAdminAccess`.

kvas-damian:
In addition to what @Grunny said I think we should introduce new error message for expired or not valid edit token. Current solution for this kind of errors displays `forum-board-title-validation-invalid` when trying to create new board.

armonr:
Thanks @Grunny and @kvas-damian. Comments were addressed.

extensions/wikia/Forum/ForumExternalController.class.php
```diff
@@ -58,6 +58,12 @@ public function createNewBoard() {
 			return;
 		}
 
+		try {
+			$this->displayAnyWriteAccessError();
```
garthwebb: Two things.  First, I think functions like this should be named for what they intend to do, not what they might do in an exception.  In this case, that would be something like `checkEditToken()`.  Second, given a function that does this already exists as `checkWriteRequest()` why not call that here rather than define another function that just adds another layer of `try`/`catch` blocks?  If you want to reduce repetition, I'd make a method that just sets the return values for you and use it here, e.g.:

``` php
function setMismatchError() {
    $this->status = 'error';
    $this->errormsg = wfMessage( 'forum-token-mismatch' )->escaped();
}
...
try {
    $this->checkWriteRequest();
} catch ( \BadRequestException $bre ) {
    $this->setMismatchError();
    return; # no need to return a boolean here; Oasis doesn't care what controllers return
}
```

extensions/wikia/Forum/ForumExternalController.class.php
```diff
@@ -161,6 +175,12 @@ public function removeBoard() {
 			return;
 		}
 
+		try {
+			$this->setTokenMismatchError();
```
kvas-damian: as @garthwebb already mentioned somewhere: something is wrong here.

armonr: o.O Somehow missed the commit train! Will fix!

garthwebb:
LGTM

PR #8814, opened 2015-10-19
(CE-2841) Disable transactions on SMW temp table queries

Grunny:
Disable transactions on SMW temp table queries on slave connections.
MediaWiki automatically creates transactions a lot of the time, and
creating temporary tables in a transaction is not replication-safe
and causes errors in MySQL 5.6.

/cc @Wikia/community-engineering @drozdo


drozdo:
LGTM

PR #8815, opened 2015-10-19
 CE-2769: Add TemplateClassificationStorage client

lukaszkonieczny:
Classes in `Swagger` directory are auto generated by `swagger-codegen` so shouldn't be code reviewed. These code also shouldn't be modified.


includes/wikia/api/TemplateClassificationApiController.class.php
```diff
@@ -0,0 +1,177 @@
+<?php
+
+use Wikia\Service\Gateway\ConsulUrlProvider;
+use Wikia\Service\Swagger\ApiProvider;
+use Swagger\Client\TemplateClassification\Storage\Api\TCSApi;
+use Swagger\Client\TemplateClassification\Storage\Models\TemplateTypeProvider;
+use Swagger\Client\ApiException;
+
+class TemplateClassificationApiController extends WikiaApiController
+{
```
Grunny: Curly brace should be on the same line as the class definition per conventions.

includes/wikia/api/TemplateClassificationApiController.class.php
```diff
@@ -0,0 +1,177 @@
...
+	 *
+	 * @param int $wikiId
+	 * @param int $pageId
+	 * @return string
+	 * @throws ApiException
+	 * @throws BadRequestApiException
+	 * @throws Exception
+	 */
+	public function getType( $wikiId, $pageId ) {
+		$templateType = "";
```
Grunny: Should be single quotes per conventions.

includes/wikia/api/TemplateClassificationApiController.class.php
```diff
@@ -0,0 +1,177 @@
...
+	 * @throws BadRequestApiException
+	 * @throws Exception
+	 */
+	public function getDetails( $wikiId, $pageId ) {
+		$templateDetails = [];
+
+		try {
+			$details = $this->getApiClient()->getTemplateDetails( $wikiId, $pageId );
+
+			if ( !is_null($details) ) {
```
Grunny: Should be spacey per conventions.

includes/wikia/api/TemplateClassificationApiController.class.php
```diff
@@ -0,0 +1,177 @@
...
+	 * @return TCSApi
+	 */
+	private function createApiClient() {
+		global $wgConsulUrl, $wgConsulServiceTag;
+		$urlProvider = new ConsulUrlProvider( $wgConsulUrl, $wgConsulServiceTag );
+		$apiProvider = new ApiProvider( $urlProvider );
+		return $apiProvider->getApi( self::SERVICE_NAME, TCSApi::class );
+	}
+
+}
```
Grunny: Needs a newline at the end of file.

Grunny:
Shouldn't TemplateClassificationApiController.class.php be defined in the setup file so it's accessible?

includes/wikia/api/TemplateClassificationApiController.class.php
```diff
@@ -0,0 +1,177 @@
...
+	private function validateRequest( User $user, WikiaRequest $request ) {
+		if ( !$request->wasPosted() ) {
+			throw new BadRequestApiException();
+		}
+
+		if ( !$user->isLoggedIn() || !$user->matchEditToken( $request->getVal( 'editToken' ) ) ) {
+			throw new UnauthorizedException();
+		}
+
+		if ( !$user->isAllowed( 'edit' ) ) {
```
Grunny: Should we be checking if they can edit the page they are classifying? Which would also check if the user is blocked.

Grunny:
LGTM, other than the last comment about including the class in the setup file.

lukaszkonieczny:
I assume that we include this file when we need to use it in the code.

Grunny:
Fair enough then. :)

PR #8816, opened 2015-10-19
WikiFactoryTags::getTags - optimize the query

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-1590

No need to `ORDER BY name` as it causes `Using temporary; Using filesort`

```
+----+-------------+--------------+--------+----------------+---------+---------+--------------------------------+------+----------------------------------------------+
| id | select_type | table        | type   | possible_keys  | key     | key_len | ref                            | rows | Extra                                        |
+----+-------------+--------------+--------+----------------+---------+---------+--------------------------------+------+----------------------------------------------+
|  1 | SIMPLE      | city_tag_map | ref    | PRIMARY,tag_id | PRIMARY | 4       | const                          |    1 | Using index; Using temporary; Using filesort |
|  1 | SIMPLE      | city_tag     | eq_ref | PRIMARY        | PRIMARY | 4       | wikicities.city_tag_map.tag_id |    1 |                                              |
+----+-------------+--------------+--------+----------------+---------+---------+--------------------------------+------+----------------------------------------------+
```

@drozdo / @michalroszka / @wladekb / @owend

### DIFF
--- a/extensions/wikia/WikiFactory/Tags/WikiFactoryTags.php
+++ b/extensions/wikia/WikiFactory/Tags/WikiFactoryTags.php
@@ -109,8 +109,7 @@ class WikiFactoryTags {
 				array( "city_tag", "city_tag_map" ),
 				array( "tag_id", "name" ),
 				array( "tag_id = id", "city_id" => $this->mCityId ),
-				__METHOD__,
-				array( "ORDER BY" => "name" )
+				__METHOD__
 			);
 			while( $row = $dbr->fetchObject( $sth ) ) {
 				$result[ $row->tag_id ] = $row->name;


wladekb:
I would guess this query returns only a few rows and filesort/temporary shouldn't hurt us at all. Do you have any performance data how long this query takes or whether it spikes or not?

:+1:

macbre:
It does not spike nor takes a really long time.

This one is "as we're here let's fix all of them" kind of a fix - I'm going through `EXPLAIN` for all queries being run on shared db and improving them - `ORDER BY` is not really needed here.

PR #8817, opened 2015-10-19
XW-30 don't send data about related pages and topContributors on curated main pages

bognix:
Jira ticket: https://wikia-inc.atlassian.net/browse/XW-30

Related PR: 
- mercury => https://github.com/Wikia/mercury/pull/1538
- api-tests => https://github.com/Wikia/api-tests/pull/139

APP should be released first


extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -467,7 +465,30 @@ public function getCuratedContentSection() {
...
+
+	public function getMainPageDetailsAndAdsContext() {
+		$mainPageTitle = Title::newMainPage();
+		$mainPageArticleID = $mainPageTitle->getArticleID();
+		$article = Article::newFromID( $mainPageArticleID );
+
+		$articleDetails = $this->getArticleDetails( $article );
+
+		if ( empty( $articleDetails ) ) {
+			throw new WikiaException( 'Article details empty', 500 );
```
rogatty: We don't throw those exceptions in `getArticle` so why here? User can still use the page without them, right? Maybe we should just log an error?

extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -467,7 +465,30 @@ public function getCuratedContentSection() {
...
+		$adsContext = $this->mercuryApi->getAdsContext( $mainPageTitle );
+
+		if ( empty( $adsContext ) ) {
+			throw new WikiaException( 'Ads Context empty', 500 );
+		}
+
+		$this->response->setFormat( WikiaResponse::FORMAT_JSON );
+		$this->response->setCacheValidity( WikiaResponse::CACHE_STANDARD );
+		$this->response->setVal( 'details', $articleDetails );
+		$this->response->setVal( 'adsContext',  $adsContext );
```
rogatty: Too much space :see_no_evil: I recommend auto formatting ;)

extensions/wikia/MercuryApi/MercuryApiController.class.php
bognix: you are asking API for these values.
If API can't provide value I think it should throw.
How this exception is going to be handled is another thing.

Exception is an unambiguous way of saying something is wrong.

extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -467,7 +465,30 @@ public function getCuratedContentSection() {
...
+
+		$articleDetails = $this->getArticleDetails( $article );
+
+		if ( empty( $articleDetails ) ) {
+			throw new WikiaException( 'Article details empty', 500 );
+		}
+
+		$adsContext = $this->mercuryApi->getAdsContext( $mainPageTitle );
+
+		if ( empty( $adsContext ) ) {
```
rogatty:
```
public function getAdsContext( Title $title ) {
    global $wgEnableAdEngineExt;
    if ( !empty( $wgEnableAdEngineExt ) ) {
        $adContext = new AdEngine2ContextService();
        return $adContext->getContext( $title, self::MERCURY_SKIN_NAME );
    }
    return null;
}
```

Are you sure we want to throw when ad engine is disabled?

bognix: so if I decide not to throw I would need to check return val of this function everywhere I call it in front end and reject the promise if data is empty instead of relying on automatic rejection when error is thrown?

For our use case not throwing would introduce complexity which is not needed at this point

tomasznapieralski:
LGTM

extensions/wikia/MercuryApi/MercuryApiController.class.php
rogatty: We already handle this scenario in articles. If ad engine is disabled and `adsContext` is empty we just don't display ads. This isn't an error that should prevent us from providing main page details to the SPA, it's just a flag controlling ads that's turned off on some wikis.

extensions/wikia/MercuryApi/MercuryApiController.class.php
rogatty: You already check if `details` are set in https://github.com/Wikia/mercury/pull/1538/files#diff-88b7fe5a8a84ed29d43821b2f75c6f8fR54

I don't see the need to throw exception here as it will prevent us from delivering adsContext without good reason. IMO setting them to null is good enough.

rogatty:
LGTM

vforge:
LGTM

hakubo:
LGTM

PR #8818, opened 2015-10-19
ADEN-2569 Fix Krux send event method

fraszczakszymon:
Those changes should prevent occurring error in console `Krux is not a function` - issue connected with SP detection on mobile


nandy-andy:
The changes look good as for a quick fix. We created a follow-up ticket: [ADEN-2570](https://wikia-inc.atlassian.net/browse/ADEN-2570) to investigate the root cause.

PR #8819, opened 2015-10-19
(SEC-14) Don't process message further when replacing rawParams

Grunny:
In https://github.com/Wikia/app/pull/8457 we added the ability to
supply rawParams to mw.message so that they would be passed after
the message is escaped. But if the jqueryMsg library is loaded, the
current implementation means this becomes unescaped because rather
than using the parser in mediawiki.js, it uses the parser in jqueryMsg
again. The jqueryMsg parser actually unescapes text because it
expects mediawiki.js to escape the text after the parser is run, which
we're running again after escaping.

/cc @lukaszkonieczny @macbre @kvas-damian

### DIFF
--- a/resources/mediawiki/mediawiki.js
+++ b/resources/mediawiki/mediawiki.js
@@ -266,6 +266,8 @@ var mw = ( function ( $, undefined ) {
 				text = mw.html.escape( text );
 			}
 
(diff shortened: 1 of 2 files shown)


lukaszkonieczny:
Like This Quick Fix :D LGTM

kvas-damian:
:+1:

PR #8822, opened 2015-10-19
INT-235: Manually trigger page view when search is occured

mari-mcmurtrie:
GA is not collecting data when a Google custom search occurred. After investigation, I found that we have to manually ping GA to know if we are doing search, with proper query parameter ("q", in this case). This is called "virtual page view", and we already do the same thing for Visual editor. So I use the same function that VE uses to ping GA for "virtual page view". That function is called "window.guaTrackPageview ()" in universal_analytics.js.

Ticket: https://wikia-inc.atlassian.net/browse/INT-235


mari-mcmurtrie:
@perjg will you CR? Thanks!!

extensions/wikia/GlobalNavigation/scripts/GoogleSearchAPI.js
```diff
@@ -35,6 +35,11 @@ $(function () {
 			searchQuery += ' site:' + window.location.hostname;
 		}
 
+		//Manually trigger page view
+		var currentUrlPath = window.location.pathname;
```
dianafa: please move this var declaration on the top of the function to have one var per scope :)

perjg:
@mari If my understanding is correct, this should work.

See here: https://support.google.com/analytics/answer/1012264?hl=en

In the section "Setting Up Site Search for POST-Based Search Engines":
analytics.js: ga('send', 'pageview', '/search_results.php?q=keyword');

PR #8823, opened 2015-10-19
Always sample the authentication method unless the user is null or not logged in

drsnyder:
Always sample the authentication method unless the user is null or not logged in. Without this change, we'll only log when we fall back which won't give use any information about the proportions of successful helios authentications vs MediaWiki fallbacks.

This is a follow on to #8761.

https://wikia-inc.atlassian.net/browse/SERVICES-873

@Wikia/services-team


nmonterroso:
LGTM

PR #8824, opened 2015-10-19
CONSF-51 removing video view counts

lizlux:
The view counter is buggy so we're hiding it for now. 

@saipetch @garthwebb Could one of you have a look? Thanks!


garthwebb:
LGTM

PR #8825, opened 2015-10-20
quick tool to log user email changes

owend:
- need to create this table in prod before merging/pushing
- uses old school templates because the tool already does
- added a new hook to core because the existing one does not send both email values

@Grunny @lgarczewski

### DIFF
--- a/extensions/wikia/EditAccount/SpecialEditAccount_body.php
+++ b/extensions/wikia/EditAccount/SpecialEditAccount_body.php
@@ -550,20 +550,21 @@ class EditAccount extends SpecialPage {
 	public function displayLogData() {
 		global $wgExternalSharedDB, $wgOut, $wgRequest;
 
-		$user_id = $wgRequest->getVal('user_id', null);
+		$user_id = $wgRequest->getInt('user_id', 0);
 		$rows = [];
 
 		if ( $wgExternalSharedDB && $user_id ) {
+			$user_name = User::newFromID($user_id);
+
 			$dbr = wfGetDB ( DB_SLAVE, array(), $wgExternalSharedDB );
 			$res = $dbr->select (
 				'user_email_log',   // from
 				["*"], 				// cols
-				"user_id = " . $dbr->strencode($user_id), // where
+				['user_id' => $user_id], // where
 				__METHOD__,
 				["ORDER BY" => "changed_at DESC"]  // options
 				);
 			while ( $row = $dbr->fetchObject( $res ) ) {
-				$row->user_name = User::newFromID($row->user_id);
 				$row->changed_by_name = User::newFromId($row->changed_by_id);
 				$rows[] = $row;
 			}
@@ -572,7 +573,7 @@ class EditAccount extends SpecialPage {
 		$oTmpl = new EasyTemplate( dirname( __FILE__ ) . '/templates/' );
 
 		$oTmpl->set_Vars( [
-			'userName' => 'foo',
+			'userName' => $user_name,
 			'returnURL' => $this->getTitle()->getFullURL(),
 			'rows' => $rows
 			]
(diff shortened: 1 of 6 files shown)


extensions/wikia/EditAccount/SpecialEditAccount_body.php
```diff
@@ -518,4 +529,55 @@ public static function generateRandomScrambledPassword() {
...
+
+		$user_id = $wgRequest->getVal('user_id', null);
+		$rows = [];
+
+		if ( $wgExternalSharedDB && $user_id ) {
+			$dbr = wfGetDB ( DB_SLAVE, array(), $wgExternalSharedDB );
+			$res = $dbr->select (
+				'user_email_log',   // from
+				["*"], 				// cols
+				"user_id = " . $dbr->strencode($user_id), // where
```
Grunny: `[ 'user_id' => $user_id ]` should be preferred for where clauses over manual escaping. `strencode` doesn't escape spaces and so forth, so this can still cause issues.

extensions/wikia/EditAccount/SpecialEditAccount_body.php
```diff
@@ -518,4 +529,55 @@ public static function generateRandomScrambledPassword() {
...
+				 'changed_by_ip' => $wgRequest->getIP()  // stored as string
+				]);
+		}
+		return true;
+	}
+
+	public function displayLogData() {
+		global $wgExternalSharedDB, $wgOut, $wgRequest;
+
+		$user_id = $wgRequest->getVal('user_id', null);
```
Grunny: Use `getInt` so you know the user ID is always an ID.

extensions/wikia/EditAccount/SpecialEditAccount_body.php
```diff
@@ -518,4 +529,55 @@ public static function generateRandomScrambledPassword() {
...
+			$dbr = wfGetDB ( DB_SLAVE, array(), $wgExternalSharedDB );
+			$res = $dbr->select (
+				'user_email_log',   // from
+				["*"], 				// cols
+				"user_id = " . $dbr->strencode($user_id), // where
+				__METHOD__,
+				["ORDER BY" => "changed_at DESC"]  // options
+				);
+			while ( $row = $dbr->fetchObject( $res ) ) {
+				$row->user_name = User::newFromID($row->user_id);
```
Grunny: Since this shouldn't change as we're querying on the user ID, can this just be set before the query as a single variable that can then also be used to pass the user name to the template?

extensions/wikia/EditAccount/SpecialEditAccount_body.php
```diff
@@ -518,4 +529,55 @@ public static function generateRandomScrambledPassword() {
...
+				$row->changed_by_name = User::newFromId($row->changed_by_id);
+				$rows[] = $row;
+			}
+		}
+
+		$oTmpl = new EasyTemplate( dirname( __FILE__ ) . '/templates/' );
+
+		$oTmpl->set_Vars( [
+			'userName' => 'foo',
```
Grunny: If we create a single user object above, we can pass the name.

extensions/wikia/EditAccount/templates/changelog.tmpl.php
```diff
@@ -0,0 +1,26 @@
...
+	<th> Changed By </th>
+	<th> Changed By IP </th>
+	<th> Changed At </th>
+</tr>
+<?php
+    foreach ($rows as $row) {
+	echo "<tr>";
+	echo "
+		<td> {$row->old_email} </td>
+		<td> {$row->new_email} </td>
```
Grunny: This should all be properly escaped for good measure.

owend:
Thanks for the review @Grunny !

owend:
Schema changes have been made in production, so this is good to merge any time.

PR #8826, opened 2015-10-20
SOC-1440 Check token on user attribute update or user avatar update

armonr:
/CC @garthwebb @Grunny @kvas-damian 

https://wikia-inc.atlassian.net/browse/SOC-1440

### DIFF
--- a/extensions/wikia/UserProfilePageV3/UserProfilePage.i18n.php
+++ b/extensions/wikia/UserProfilePageV3/UserProfilePage.i18n.php
@@ -115,7 +115,6 @@ $messages['en'] = array(
 	'user-identity-remove-fail' => 'There was some problem when removing the avatar!',
 	'user-identity-avatars-maintenance' => 'Upload and deletion of avatars temporarily disabled during maintenance!',
 	'userprofilepage-edit-modal-error' => 'Unexpected error occurred',
-	'user-identity-token-mismatch' => 'Oops! Token doesn\'t match',
 );
 
 /** Message documentation (Message documentation)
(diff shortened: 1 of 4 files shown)


extensions/wikia/UserProfilePageV3/js/UserProfilePage.js
```diff
@@ -388,7 +389,10 @@ var UserProfilePage = {
 			type: 'POST',
 			url: this.ajaxEntryPoint + '&method=saveUserData',
 			dataType: 'json',
-			data: {'userId': UserProfilePage.userId, 'data': JSON.stringify(userData)},
+			data: {
+				'userId': UserProfilePage.userId,
+				'data': JSON.stringify(userData),
+				'token': token},
```
macbre: You can simply call `window.mw.user.tokens.get('editToken')`here to get the edit token instead of passing it from the backend.

extensions/wikia/UserProfilePageV3/UserProfilePage.i18n.php
```diff
@@ -115,6 +115,7 @@
 	'user-identity-remove-fail' => 'There was some problem when removing the avatar!',
 	'user-identity-avatars-maintenance' => 'Upload and deletion of avatars temporarily disabled during maintenance!',
 	'userprofilepage-edit-modal-error' => 'Unexpected error occurred',
+	'user-identity-token-mismatch' => 'Oops! Token doesn\'t match',
```
kvas-damian: What do you think about having this message in common place and reuse it everywhere?

BTW current message is pretty vague and don't tell user what to do next.

armonr: @kvas-damian I would really like to; I did not find anywhere for common messages, though.

kvas-damian: What about https://github.com/Wikia/app/blob/dev/extensions/wikia/Oasis/Oasis.i18n.php or https://github.com/Wikia/app/blob/dev/languages/messages/wikia/MessagesEn.php ?

Grunny: The standard message MediaWiki shows when the provided token is incorrect is [`sessionfailure`](https://github.com/Wikia/app/blob/1dfb4dec20c0/languages/messages/MessagesEn.php#L2886-L2888), so that or a variant on it is an option.

kvas-damian:
LGTM

garthwebb:
LGTM

PR #8831, opened 2015-10-20
DAT-3377, prevent Category Select save from CSRF attack

RafLeszczynski:
https://wikia-inc.atlassian.net/browse/DAT-3377

### DIFF
--- a/extensions/wikia/CategorySelect/js/CategorySelect.view.js
+++ b/extensions/wikia/CategorySelect/js/CategorySelect.view.js
@@ -82,10 +82,11 @@
 							},
 							method: 'save'
 						}).done(function( response ) {
+							var error = response.error || response.exception;
 							$container.stopThrobbing();
 
-							if ( response.error ) {
-								throw 'Saving error: ' + response.error;
+							if ( error ) {
+								throw 'Saving error: ' + error;
 							} else {
 								$wrapper
 									.removeClass( 'editMode' )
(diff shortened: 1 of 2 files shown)


extensions/wikia/CategorySelect/js/CategorySelect.view.js
```diff
@@ -77,7 +77,8 @@
 							controller: 'CategorySelectController',
 							data: {
 								articleId: articleId,
-								categories: $wrapper.data( 'categorySelect' ).getData( '.new' )
+								categories: $wrapper.data( 'categorySelect' ).getData( '.new' ),
+								token: mw.user.tokens.get('editToken')
```
sebastianmarzjan, line 81: Just a sanity check: given this is wrapped in $.nirvana.sendRequest is .done() handling the .fail() case too? If not, probably it's worth adding that explicitly

RafLeszczynski, line 81: done

sebastianmarzjan, line 81: Not to open a can of worms, but I don't see where  the thrown exceptions (the existing one and one added in this PR) are caught - what will be the end state?

RafLeszczynski, line 81: The previous one was simply throwing JS error in console and removing throbber so the UI is not blocked. I did the same with the new one. I'm open for better UX in this cases. We could discuss it with PO and propose the right copy + translations and some UI changes to display the message. Maybe it could be done as a separate story?

sebastianmarzjan, line 81: Definitely. We are not degrading current behavior.

sebastianmarzjan:
LGTM

PR #8834, opened 2015-10-20
Improve IP logging

wladekb:
WikiaLogger intiialization happens before `$wgRequest` is intiialized and IPs that we log are our internal ones. This change fixes the issue by fetching IP from Mediawiki when it is available.

/cc @macbre @Grunny @ljagiello @owend

### DIFF
--- a/lib/Wikia/src/Logger/Hooks.php
+++ b/lib/Wikia/src/Logger/Hooks.php
@@ -108,7 +108,7 @@ class Hooks {
 	 * @return bool true
 	 */
 	public static function onWikiFactoryExecuteComplete( \WikiFactoryLoader $wikiFactoryLoader ) {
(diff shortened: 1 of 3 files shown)


ljagiello:
Sounds good.

Grunny:
:+1:

lib/Wikia/src/Logger/Hooks.php
```diff
@@ -163,6 +154,27 @@ public static function onWikiFactoryExecuteComplete( \WikiFactoryLoader $wikiFac
 		return true;
 	}
 
+	public function onWebRequestInitialized( \WebRequest $webRequest ) {
```
macbre, line 157: `static` is missing here - https://wikia-inc.atlassian.net/browse/ER-8547

PR #8835, opened 2015-10-20
DAT-3378 Top10Lists: prevent CSRF, use POST instead of GET

sebastianmarzjan:


extensions/wikia/TopLists/TopListHelper.class.php
```diff
@@ -682,6 +693,17 @@ public static function addItem() {
 		$result = array( 'result' => false );
 		$errors = array();
 
+		try {
```
RafLeszczynski, line 687: could you please move this logic to the separate function and call it in two places to avoid copy/pasting ?

extensions/wikia/TopLists/js/list.js
```diff
@@ -20,11 +20,12 @@ var TopList = {
 	voteItem: function(e) {
 		TopList.blockInput();
 
-		$.getJSON(wgScript,
+		$.post(wgScript,
 			{
 				'action': 'ajax',
-				'rs': 'TopListHelper::voteItem',
-				'title': $('#top-list-title').val() + '/' + this.id
+                'rs': 'TopListHelper::voteItem',
```
RafLeszczynski: indentation issue?

extensions/wikia/TopLists/js/list.js
```diff
@@ -46,13 +47,14 @@ var TopList = {
 		ev.preventDefault();
 		TopList.blockInput();
 
-		$.getJSON(wgScript,
+		$.post(wgScript,
 			{
 				'action': 'ajax',
+                'cb' : Math.random(),
```
RafLeszczynski: indentation issue?

includes/wikia/nirvana/WikiaRequest.class.php
```diff
@@ -267,4 +267,18 @@ public function getScriptUrl() {
 		$scriptUrl = isset( $_SERVER['SCRIPT_URL'] ) ? $_SERVER['SCRIPT_URL'] : null;
 		return $scriptUrl;
 	}
+
+	/**
```
RafLeszczynski, line 271: is there an option to move this function to some helper class and call it from multiple places instead of copy/pasting?

RafLeszczynski:
LGTM

PR #8839, opened 2015-10-20
CE-2902 added full template classification labels

lgarczewski:
@Wikia/community-engineering


extensions/wikia/TemplateClassification/TemplateClassification.i18n.php
```diff
@@ -4,12 +4,44 @@
...
+        'template-classification-description-design' => 'This template is used to design or structure article pages.',
+
+        'template-classification-type-navigation' => 'Navigation',
+        'template-classification-description-navigation' => 'This template is used for navigation.',
+
+        'template-classification-type-nonarticle' => 'Non-article',
+        'template-classification-description-nonarticle' => 'This template is not used on article pages.',
+
+        'template-classification-type-dontknow' => 'Unknown',
+        'template-classification-description-dontknow' => "I don't know.",
```
Grunny: All these new messages need QQQ docs.

extensions/wikia/TemplateClassification/TemplateClassificationController.class.php
```diff
@@ -23,6 +23,7 @@ public function getTemplateClassificationEditForm() {
 				 * template-classification-type-unclassified
 				 */
 				'name' => wfMessage( "template-classification-type-{$type}" )->escaped(),
+				'description' => wfMessage( "template-classification-description-{$type}" )->escaped(),
```
Grunny: Since the message key is being constructed with a variable, similar to the message use above, this should have a comment above it with the possible message strings that can be used here so it is possible to find where they are used with grep.

extensions/wikia/TemplateClassification/TemplateClassificationView.php
```diff
@@ -13,14 +13,17 @@ class View {
...
+		$templateType = lcfirst( ( new \TemplateClassificationMockService() )->getTemplateType( $articleId ) );
 		/**
 		 * template-classification-type-unclassified
 		 * template-classification-type-infobox
 		 * template-classification-type-navbox
 		 * template-classification-type-quote
 		 */
-		$templateName = wfMessage( "template-classification-type-{$templateType}" )->escaped();
+		$templateName = wfMessage(
+			"template-classification-indicator",
```
Grunny: Should use single quotes per conventions since it isn't using string interpolation.

extensions/wikia/TemplateClassification/TemplateClassificationView.php
```diff
@@ -13,14 +13,17 @@ class View {
...
+			wfMessage( "template-classification-type-{$templateType}" )
```
Grunny, line 25: Since this is being escaped as a parameter to the `template-classification-indicator` message, this should have an output method of `->plain()` or `->text()`. If you don't add an explicit output type it will use `->parse()` which means it would be double escaped and is an unnecessary call to the parser.

includes/wikia/TemplateClassification.class.php
```diff
@@ -30,6 +38,15 @@ class TemplateClassification {
 		self::TEMPLATE_QUOTE,
 		self::TEMPLATE_NAVBOX,
 		self::TEMPLATE_UNCLASSIFIED,
```
kamilkoterba, line 40: TEMPLATE_UNCLASSIFIED was about to be replaced with TEMPLATE_UNKNOWN

includes/wikia/TemplateClassification.class.php
```diff
@@ -30,6 +38,15 @@ class TemplateClassification {
...
+		self::TEMPLATE_NAVBOX,
+		self::TEMPLATE_FLAG,
+		self::TEMPLATE_REF,
+		self::TEMPLATE_MEDIA,
+		self::TEMPLATE_DATA,
+		self::TEMPLATE_DESIGN,
+		self::TEMPLATE_NAV,
+		self::TEMPLATE_NOT_ART,
+		self::TEMPLATE_DONTKNOW,
```
kamilkoterba: DONTKNOW is just user facing message fallback on modal for TEMPLATE_UNKNOWN type so shouldn't be visible here

skins/oasis/modules/TemplateDraftModuleController.class.php
```diff
@@ -11,7 +11,8 @@ public function executeCreate( $params ) {
 			AssetsManager::getInstance()->getSassCommonURL( "skins/oasis/css/modules/TemplateDraftModule.scss" )
 		);
 
-		$subpage = wfMessage( 'templatedraft-subpage' )->inContentLanguage()->escaped();
+		$subpage = 'foo';
+//		$subpage = wfMessage( 'templatedraft-subpage' )->inContentLanguage()->escaped();
```
Grunny: Testing code left over?

includes/wikia/TemplateClassification.class.php
```diff
@@ -52,8 +68,7 @@ public function getType() {
 
 	/**
 	 * Checks if the template is of the given type.
-	 * @param $type One of values from the $templateTypes array
-	 * @return bool
+
```
Grunny: Why was this removed?

includes/wikia/TemplateClassification.class.php
lgarczewski, line 40: Per discussion: we need to keep UNCLASSIFIED for now because it has a slightly different meaning. Once we hook up the Template Conversion flow to the service UNCLASSIFIED will be removed.

extensions/wikia/TemplateClassification/TemplateClassification.i18n.php
```diff
@@ -4,12 +4,44 @@
...
+        'template-classification-description-design' => 'This template is used to design or structure article pages.',
+
+        'template-classification-type-navigation' => 'Navigation',
+        'template-classification-description-navigation' => 'This template is used for navigation.',
+
+        'template-classification-type-nonarticle' => 'Non-article',
+        'template-classification-description-nonarticle' => 'This template is not used on article pages.',
+
+        'template-classification-type-unknown' => 'Unknown',
+        'template-classification-description-unknown' => "I don't know.",
```
Grunny: Some whitespace issues, the messages above are all indented with spaces instead of tabs.

extensions/wikia/TemplateClassification/TemplateClassification.i18n.php
```diff
@@ -17,10 +49,33 @@
...
+        'template-classification-description-quote' => 'Description (additional label) for the quote type. This is shown in dialog box under the type name to explain what each type is.',
+        'template-classification-description-data' => 'Description (additional label) for the data type. This is shown in dialog box under the type name to explain what each type is.',
+        'template-classification-description-design' => 'Description (additional label) for the design type. This is shown in dialog box under the type name to explain what each type is.',
+        'template-classification-description-navigation' => 'Description (additional label) for the navigation type. This is shown in dialog box under the type name to explain what each type is.',
+        'template-classification-description-nonarticle' => 'Description (additional label) for the non-article type. This is shown in dialog box under the type name to explain what each type is.',
+        'template-classification-description-unknown' => 'Description (additional label) for the unknown template type. This is shown in dialog box under the type name to explain what each type is.',
```
Grunny: Same here.

Grunny:
Other than last whitespace comments, LGTM. So, ready to go if all tested. :)

kamilkoterba:
LGTM if tested

PR #8841, opened 2015-10-20
VE-1982: Unit tests for <gallery> and <tabber> passed to <image> (PI)

inez:
It would be preferred to test `<tabber>` and `<gallery>` with already existing "testData" and "dataProvider" methods in NodeImageTest.php but this is not possible at this moment because the way how PortableInfoboxes extension is written - it heavily depends on MediaWiki parsing pipeline and is very "statefull".

For instance let's say we are passing `<gallery>Wiki.png</gallery>` as a value of source parameter of `<image>` tag. One could expect that you would have access to that value inside the Node - but actually you can't - it's already replaced with UNIQ marker by Parser (specifically Preprocessor) - and that UNIQ marker does not refer to the original wikitext - but to the parsed wikitext (html).

I decided to identify most important and "tricky" methods in NodeImage.php code and test just them.

Currently we are working on first version/iteration of media collections in infoboxes - we don't know yet if we are going to stick to that backward compatibility so this code in the future might be either removed or refactored.

### DIFF
--- a/extensions/wikia/PortableInfobox/services/Parser/Nodes/NodeImage.php
+++ b/extensions/wikia/PortableInfobox/services/Parser/Nodes/NodeImage.php
@@ -10,6 +10,60 @@ class NodeImage extends Node {
 	const CAPTION_TAG_NAME = 'caption';
 	const MEDIA_TYPE_VIDEO = 'VIDEO';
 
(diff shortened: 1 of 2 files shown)


pauloslund:
lgtm

PR #8843, opened 2015-10-21
INT-237: Achievements: allow badges to be uploaded without frames.

mari-mcmurtrie:
We have an admin for JA starwars request that he does not want achievements badges without frames because those frames cut off his images. So we have a new WikiFactory variable (wgBadgeCustomizationAllowed), and if this is true, then we allow an image uploaded without frame. Images are still resized into 128*128 with original ratio so that no matter what shape the image is, it is still presentable.


mari-mcmurtrie:
@inez can you CR this? Thank you!

extensions/wikia/AchievementsII/services/AchImageUploadService.class.php
```diff
@@ -167,43 +167,47 @@ public static function uploadBadge($destinationFileName, $badgeLevel) {
...
 
-		// now let's fix badge transparency
-		self::merge_alphamask($badge, $img['top'], $img['bottom']);
+			// now let's fix badge transparency
+			self::merge_alphamask($badge, $img['top'], $img['bottom']);
 
-		// save badge
-		$ret = imagepng($badge, $badgeFile, 9, PNG_ALL_FILTERS);
+			// save badge
+			$ret = imagepng($badge, $badgeFile, 9, PNG_ALL_FILTERS);
```
inez: This "$ret" does not seem to be used anywhere

extensions/wikia/AchievementsII/services/AchImageUploadService.class.php
```diff
@@ -167,43 +167,47 @@ public static function uploadBadge($destinationFileName, $badgeLevel) {
 		imagedestroy($badgeImage);
 		$badgeImage = $tmp;
 
-		// load images
-		$badgesFragments = dirname(__FILE__) . '/../images/badges/fragments';
+		if ($wgBadgeCustomizationAllowed) {
```
inez: Usually we check it this way: `if ( !empty ( $wgBadgeCustomizationAllowed ) ) {`

inez: I think this variable name is confusing. No matter if it's true or false we allow for badge customization. This change only if we apply our overlay on top of it or not.

mari-mcmurtrie:
@inez thanks for CR. Will you check this again when you have time? Thanks!

inez:
LGTM

PR #8847, opened 2015-10-21
(CE-2754) Add chat.js to personal JS editing whitelist

Grunny:
Add chat.js to personal JS editing whitelist which is loaded in Chat
for the individual user.


lukaszkonieczny:
LGTM

PR #8850, opened 2015-10-21
CONSF-51 removing video views from special videos page too

lizlux:
@saipetch pointed out that we should video views on the Special:Videos page too. This PR hides those views on both mobile and desktop, including lazy loaded videos on mobile.


saipetch:
LGTM

PR #8851, opened 2015-10-21
SOC-1444 Don't allow posting new messages via GET

garthwebb:
Require POST and an edit token

https://wikia-inc.atlassian.net/browse/SOC-1444

### DIFF
--- a/extensions/wikia/Wall/js/WallBackendBridge.js
+++ b/extensions/wikia/Wall/js/WallBackendBridge.js
@@ -50,7 +50,8 @@
 					pagetitle: page.title,
 					pagenamespace: page.namespace,
 					convertToFormat: convertToFormat,
-					relatedTopics: relatedTopics
+					relatedTopics: relatedTopics,
+					token: window.mw.user.tokens.get('editToken')
 				},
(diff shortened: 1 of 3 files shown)


armonr:
:+1:

extensions/wikia/Wall/js/WallBackendBridge.js
```diff
@@ -50,7 +50,8 @@
 					pagetitle: page.title,
 					pagenamespace: page.namespace,
 					convertToFormat: convertToFormat,
-					relatedTopics: relatedTopics
+					relatedTopics: relatedTopics,
+					token: window.mw.user.tokens.get('editToken')
 				},
 				callback: this.proxy(function (data) {
```
kvas-damian, line 56: Does this callback support error message returned when token is not valid?

garthwebb, line 56: See my note on this in #8854

kvas-damian:
It's not optimal from UI perspective, but we're going to replace Forum and Wall by new Discussion product.

LGTM

PR #8854, opened 2015-10-21
SOC-1442 Don't allow token-less GET requests for edits

garthwebb:
Note; this is based off branch SOC-1444 as it has new methods needed for this change set.

https://wikia-inc.atlassian.net/browse/SOC-1442

### DIFF
--- a/extensions/wikia/Wall/js/WallBackendBridge.js
+++ b/extensions/wikia/Wall/js/WallBackendBridge.js
@@ -155,7 +155,8 @@
 					isreply: isreply,
 					pagetitle: page.title,
 					pagenamespace: page.namespace,
-					convertToFormat: convertToFormat
+					convertToFormat: convertToFormat,
+					token: window.mw.user.tokens.get('editToken')
 				},
 				callback: this.proxy(function (data) {
 					if ($.isFunction(callback)) {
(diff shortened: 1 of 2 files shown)


armonr:
LGTM

extensions/wikia/Wall/js/WallBackendBridge.js
```diff
@@ -155,7 +155,8 @@
 					isreply: isreply,
 					pagetitle: page.title,
 					pagenamespace: page.namespace,
-					convertToFormat: convertToFormat
+					convertToFormat: convertToFormat,
+					token: window.mw.user.tokens.get('editToken')
 				},
 				callback: this.proxy(function (data) {
```
kvas-damian, line 161: Does this callback support error message returned when token is not valid?

garthwebb, line 161: @kvas-damian, this extension JS has no error handling.  As an anonymous user, try clicking 'edit' next to an article comment (or Forum after logging out in another tab) and you'll see these errors just get eaten.

I don't think we care enough about this feature to add this new functionality, but if we do its another ticket.

kvas-damian:
It's not optimal from UI perspective, but we're going to replace Forum and Wall by new Discussion product.

LGTM

PR #8855, opened 2015-10-21
SOC-1445 Don't allow token-less GETs to add comments

garthwebb:
Note this is based on the SOC-1444 branch.

https://wikia-inc.atlassian.net/browse/SOC-1445

### DIFF
--- a/extensions/wikia/ArticleComments/classes/ArticleCommentsAjax.class.php
+++ b/extensions/wikia/ArticleComments/classes/ArticleCommentsAjax.class.php
@@ -174,6 +174,7 @@ class ArticleCommentsAjax {
 		try {
 			$wgRequest->isValidWriteRequest( $wgUser );
 		} catch ( \BadRequestException $bre ) {
+			$result['msg'] = wfMessage( 'sessionfailure' )->escaped();
 			return $result;
 		}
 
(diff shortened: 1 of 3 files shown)


armonr:
LGTM

extensions/wikia/ArticleComments/classes/ArticleCommentsAjax.class.php
```diff
@@ -169,9 +169,16 @@ static public function axReply() {
 	static public function axPost() {
 		global $wgRequest, $wgUser, $wgLang;
 
+		$result = [ 'error' => 1 ];
+
+		try {
+			$wgRequest->isValidWriteRequest( $wgUser );
+		} catch ( \BadRequestException $bre ) {
+			return $result;
```
kvas-damian, line 178: For better UI there should be an error message displayed for user with info what happened and what to do now. There is already translation key that can be reused: `sessionfailure`

extensions/wikia/ArticleComments/js/ArticleComments.js
```diff
@@ -446,13 +447,11 @@
 						}
 					}
 
-					//readd events
+					//re-add events
 					ArticleComments.addHover();
-					//clear error box
-					$('#article-comm-info').html('');
```
garthwebb, line 452: This article-comm-info element has `display: none` by default, and shows up, unstyled, at the top of the page.  It doesn't seem like it was ever finished.  Using BannerNotifications instead.

kvas-damian:
LGTM

PR #8858, opened 2015-10-22
(CE-2892) Fix content review modal in Monobook

Grunny:
Fix content review modal in Monobook as it was no longer showing the
status of pages. This was because the ContentReviewModule which is
called by the API now requires the wiki ID and page ID to be passed
as parameters.


lukaszkonieczny:
LGTM :+1:

PR #8860, opened 2015-10-22
CE-2783: Use Template Classification API and service

lukaszkonieczny:
Replace usage of Template Classification Mocks by proper API and Service classes.


lgarczewski:
LGTM.

PR #8867, opened 2015-10-22
SEO-103 First attempt at Lilly service

gabrys:
The code hooks into the parser linker hook. Upon parsing wikitext
it will detect all outgoing links, check if they point to a Star
Wiki in another language and POST a HTTP request to the service
called Lilly passing the current URL as the source and the target
URL as the target for the link.

This will enable us to find the same content in different languages
and serve this information as <link rel="alternate" hreflang="..."
href="..."> tags.

Star Wars wikis don't use the standard MediaWiki mechanism for
interlinks, so we need this workaround for them.

### DIFF
--- a/extensions/wikia/Lilly/Lilly.setup.php
+++ b/extensions/wikia/Lilly/Lilly.setup.php
@@ -1,0 +1,7 @@
+<?php
+
+// Autoload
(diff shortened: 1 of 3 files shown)


saipetch:
LGTM

PR #8868, opened 2015-10-23
fix overlay on ve editor

hakubo:
with 1000vw it creates a checkboard
it looks fine in Chrome with 100vw and still covers whole content

@vforge can you take a look?


vforge:
@hakubo I've written some comment on slack.

PR #8869, opened 2015-10-23
Add hooks related to user

wladekb:
We need a bit more thorough logging around user actions. This should give us the change to hook up in the important places.

/cc @macbre @Grunny


includes/context/RequestContext.php
```diff
@@ -208,6 +208,9 @@ public function getUser() {
 		if ( $this->user === null && $wgEnableHeliosExt ) {
 			$this->user = \Wikia\Helios\User::newFromToken( $this->getRequest() );
 			$authPath['helios'] = ($this->user !== null ? 'OK' : 'FAIL');
+			if ( $this->user !== null ) {
```
macbre: Maybe use `instanceof` instead to enforce the object class (I assume you provide the class in method bound to a hook)?

Grunny:
:+1:

PR #8871, opened 2015-10-23
SOC-1447 Submit questions only when posted with valid token

armonr:
/CC @garthwebb @kvas-damian 

https://wikia-inc.atlassian.net/browse/SOC-1447

### DIFF
--- a/extensions/wikia/Answers/SpecialCreateDefaultQuestionPage.php
+++ b/extensions/wikia/Answers/SpecialCreateDefaultQuestionPage.php
@@ -1,58 +1,108 @@
 <?php
 class CreateQuestionPage extends UnlistedSpecialPage {
 
+	const QS_KEY_QUESTION = 'questiontitle';
+
 	function __construct(){
 		parent::__construct("CreateQuestionPage");
 	}
 
-	function execute( $question ){
-		global $wgRequest, $wgOut, $wgUser;
-
-                if ( wfReadOnly() ) {
-                        return false;
-                }
-
-                //don't allow blocked users to ask questions, duh
-                if( $wgUser->isBlocked() ){
-					throw new UserBlockedError( $wgUser->getBlock() );
-                }
-
-                if( !$wgUser->isAllowed( 'edit' ) ){
-                        return false;
-                }
-
-		if( empty( $question ) ) {
-			$question = $wgRequest->getVal("questiontitle");
+	function execute( $question ) {
+		if ( $this->getRequest()->wasPosted() ) {
+			return $this->createQuestion();
 		}
 
-		if( empty( $question ) ) {
-			return true;
+		return $this->questionPage( $question );
+	}
+
+	private function questionPage( $question ) {
+		if ( empty( $question ) ) {
+			$question = $this->getRequest()->getVal( self::QS_KEY_QUESTION, false );
+			if ( !$question ) {
+				return;
+			}
 		}
 
 		$q = new DefaultQuestion( $question );
 
 		if ( !is_object( $q ) ) {
-			return false;
+			return;
 		}
 
-		if ( is_object($q->title) && $q->title->exists() ) {
-			$wgOut->redirect( $q->title->getFullURL() );
-			return false;
+		if ( is_object( $q->title ) && $q->title->exists() ) {
+			$this->getOutput()->redirect( $q->title->getFullURL() );
+			return;
 		}
 
 		if ( $q->searchTest() ) {
-			$wgOut->redirect( SpecialPage::getTitleFor( 'Search' )->getFullURL("search=" . $q->question . "&fulltext=Search") );
-			return false;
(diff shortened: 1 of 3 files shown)


extensions/wikia/WikiAnswers/templates/WikiAnswers_QuestionBox.php
```diff
@@ -11,7 +11,22 @@ function askQuestion(){
...
+		type: 'POST',
+		data: {
+			'questiontitle': q.charAt(0).toUpperCase() + q.substring(1),
+			'token': window.mw.user.tokens.get('editToken')
+		}
+	}).done(function () {
+		window.location.href = path + "Special:CreateQuestionPage?questiontitle="
+			+ q.charAt(0).toUpperCase() + q.substring(1);
+	}).fail(function () {
+		console.log("Error creating question!");
```
kvas-damian, line 26: I don't think our users read `console.log`. Can you find better way to notify user that something failed?

extensions/wikia/Answers/templates/AnswersAsk.tmpl.php
```diff
@@ -15,7 +15,7 @@ function ask_question() {
 	q = encodeURIComponent( q );
 	
 	var url = "http://answer.wikia.com/api.php?action=query&titles=" + q + "&format=json";
-	var params = ''
+	var params = 'token=' + window.mw.user.tokens.get('editToken');
```
kvas-damian: Do we really need `token` for `action=query` request?

extensions/wikia/WikiAnswers/templates/WikiAnswers_QuestionBox.php
armonr, line 26: @kvas-damian: The controller code is not designed for XHR requests, so there isn't an easy way to robustly return a suitable error message in response. I did discuss this with @garthwebb, and we decided it is not worth the refactoring at the moment to display the error to the user as the ROI on this is very low given the negligible usage of this extension.
I did add a log to be able to verify user's process was in fact processed if a question was raised.

extensions/wikia/Answers/SpecialCreateDefaultQuestionPage.php
```diff
@@ -1,58 +1,108 @@
...
+
 	function __construct(){
 		parent::__construct("CreateQuestionPage");
 	}
 
-	function execute( $question ){
-		global $wgRequest, $wgOut, $wgUser;
+	function execute( $question ) {
+		if ( $this->getRequest()->wasPosted() ) {
+			return $this->createQuestion();
```
garthwebb: Neither `createQuestion` nor `questionPage` return values, so there's nothing to return here or on L15 from them.  It would be better to move the `return` statement separately to the next line here and remove it altogether for L15.

extensions/wikia/Answers/SpecialCreateDefaultQuestionPage.php
```diff
@@ -1,58 +1,108 @@
...
+		}
+
+		return $this->questionPage( $question );
+	}
+
+	private function questionPage( $question ) {
```
garthwebb: Actually, I don't think we need this method at all actually.  Previously this code essentially did data validation and took the user somewhere appropriate if the data was missing or incomplete.  Here its been duplicated (I see the same lines still exist below) and used as if it were a valid alternate branch of `execute`, e.g., "If you `GET` this endpoint you should get a help page, if you `POST` we should create a new question"

What `wasPosted()` is actually testing for is a security concern.  If it fails that, the request should fail.

garthwebb:
I think the only change needed in `SpecialCreateDefaultQuestionPage` is to add the following lines at the beginning of `execute`:

https://github.com/Wikia/app/blob/04808113e8c5c94e3c48f47b35c54dfb52b44d47/extensions/wikia/Answers/SpecialCreateDefaultQuestionPage.php#L46-L50

armonr:
@garthwebb: As discussed offline, the GET request has to be entertained (sort of!), but b9f5a67 should be simpler and a better solution.

extensions/wikia/Answers/SpecialCreateDefaultQuestionPage.php
```diff
@@ -1,58 +1,84 @@
...
+			$this->redirectToHelpPage();
+		}
+	}
 
-                //don't allow blocked users to ask questions, duh
-                if( $wgUser->isBlocked() ){
-					throw new UserBlockedError( $wgUser->getBlock() );
-                }
+	private function createQuestion() {
+		if ( !$this->getRequest()->wasPosted()
```
garthwebb: We've already checked this on L10 so this will always be `false`.  Can just remove it.

garthwebb:
Other than the extra POST check, LGTM

garthwebb:
LGTM

PR #8876, opened 2015-10-26
ApiQueryUsers: include global groups when getting list of groups and rights

macbre:
And improve the way blocks are handled (previously per-wiki IP blocks were checked on per-cluster shared users database)
- construct a "full" `User` object instead of relying on "partial" data from database query.
- make `blockinfo` prop work as well, it will call `User::getBlock` (which triggers both local blocks and Phalanx check)

http://poznan.macbre.wikia-dev.com/api.php?action=query&list=users&ususers=macbre|ABX|Fuck&usprop=groups|editcount|rights|blockinfo

@jcellary / @wladekb


jcellary:
Don't know if the code works, but I like the idea and +12 -19 balance :) If you ask me, feel free to merge

PR #8878, opened 2015-10-26
ADEN-2584 Set wf variable for krux

nandy-andy:
We need to update one variable in 130+ wikias. We decided to modify this script to do so.

//cc @gabrys @owend @saipetch

### DIFF
--- a/maintenance/wikia/setWikiFactoryVariable.php
+++ b/maintenance/wikia/setWikiFactoryVariable.php
@@ -55,8 +55,8 @@ class SetWikiFactoryVariable extends Maintenance {
 			die( "Error: Empty variable name.\n" );
 		}
 
-		if ( $set && $remove && $append ) {
-			die( "Error: Cannot set and append and remove the variable at the same time\n" );
+		if ( $set && $remove || $set && $append || $remove && $append ) {
+			die( "Error: Cannot use one of the three commands (set, append, remove) at the same time" . PHP_EOL );
 		}
 
 		if( $set ) {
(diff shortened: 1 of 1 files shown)


maintenance/wikia/setWikiFactoryVariable.php
```diff
@@ -54,22 +55,20 @@ public function execute() {
 			die( "Error: Empty variable name.\n" );
 		}
 
-		if ( $set && $remove ) {
-			die( "Error: Cannot set and remove the variable at the same time\n" );
+		if ( $set && $remove && $append ) {
```
fraszczakszymon: Shouldn't it be checked for all combinations?
`$set && $remove || $set and $append || $remove && $append`

fraszczakszymon:
One note && LGTM :+1:

owend:
LGTM

maintenance/wikia/setWikiFactoryVariable.php
```diff
@@ -104,8 +115,16 @@ public function execute() {
...
 			$status = true;
 			if ( $set ) {
-				echo "Set {$this->varName} to ".var_export( $varValue, true );
+				echo "Set {$this->varName} to " . var_export( $varValue, true );
 				$status = $this->setVariable( $id, $varValue );
+			} else if ( $append ) {
+				echo "Appending " . $varValue . " to {$this->varName}:" . PHP_EOL;
+				$varData = (array) WikiFactory::getVarByName( $this->varName, $id, true );
+				$prevValue = unserialize($varData['cv_value']);
+				$newValue = $prevValue . $varValue;
```
owend, line 124: If we're going to do this a lot, it would be safer to convert to an array, and merge new values (so it will overwrite/update values) instead of just appending possible duplicates to the string.  But I guess this is okay for now.

owend:
Merging this so I can do a bit of followup work on the ticket.

PR #8879, opened 2015-10-26
CE-2928: Check if user is allowed to add content-reviewer group

lukaszkonieczny:
Check if user is allowed to add `content-reviewer` group to other user

@kvas-damian @Wikia/community-engineering


lgarczewski:
LGTM.

extensions/wikia/ContentReview/ContentReview.hooks.php
```diff
@@ -240,7 +241,7 @@ public function onShowDiff( \DifferenceEngine $diff, &$notice ) {
 	public function onUserRightsGroupCheckboxes( $group, &$disabled, &$irreversible ) {
 		global $wgUser;
 
-		if ( $group === 'content-reviewer' && ( !$wgUser->isAllowed( 'content-review' ) || !$wgUser->isStaff() ) ) {
+		if ( $group === 'content-reviewer' && !$wgUser->isAllowed( 'content-review' ) ) {
```
Grunny: I think we wanted to check both `content-review` and staff in case we add some content reviewers who are not staff in the future, as they shouldn't be able to add the group.

kvas-damian: @Grunny exactly! We just discussed that offline.

kvas-damian:
LGTM

PR #8880, opened 2015-10-26
CE-2928: Check if user is allowed to add content-reviewer group

lukaszkonieczny:
Check if user is allowed to add content-reviewer group


lukaszkonieczny:
CR in https://github.com/Wikia/app/pull/8879

PR #8882, opened 2015-10-26
(SEC-12) Protect against CSRF on CreateNewWiki

Grunny:
Make sure the CreateNewWiki AJAX request is POSTed and has a valid
edit token to protect against CSRF.


macbre:
:+1:

PR #8805, opened 2015-10-16
CE-2834 Basic flow, Force open modal and submit type on publish

kamilkoterba:
Covered flows for editing template type in source editor.
1. Force modal on creating new tpl via URL
2. Force modal on publish if type is not defined

Synced together the rest of possible flows to use common functionality
3. Edit type on edit page via entrypoint for new page
4. Edit type on edit page via entrypoint for existing page
...

Provided different way of handling type save for editpage (on view page submit is asynch, on editpage it's stored in hidden field until publish)

Split editing on view page from editing on edit page and provied common modal control module

https://wikia-inc.atlassian.net/browse/CE-2834

@Wikia/community-engineering

PR #8883, opened 2015-10-26
CE-2930 Gracefully handle exceptions thrown by the TC service

adamkarminski:
**READY FOR MERGE**

[Compare with the parent branch](https://github.com/Wikia/app/compare/CE-2834...CE-2834-exceptions)

We call the TCS (TemplateClassification Service) directly in two places in our app. This PR deals with handling exceptions that are thrown by the service. The cases we're dealing with are:
##### 1. Displaying an entry point

Handled by `TemplateClassificationView`. That's why God created fallback messages. If the service is unreachable (witch we assume to be 99% of the exceptions) we go with not modifying the output at all so the entry point for classification is not visible.
##### 2. Creating or editing an article

Handled in `Hooks::onArticleInsertComplete`, `Hooks::onEditPageShowEditFormFields` and `TemplateClassificationEditorPlugin.js`
The classification is controlled with a hidden field in the edit form. If the field is empty (e.g. template is not pre-classified) a user is not able to submit the template. However, if the service is unreachable we should not break the main action which is template creation. The same goes for editing an existing template.

/cc @lukaszkonieczny @kamilkoterba @Grunny


extensions/wikia/TemplateClassification/TemplateClassification.hooks.php
```diff
@@ -2,7 +2,7 @@
 
 namespace Wikia\TemplateClassification;
 
-use Wikia\TemplateClassification\Permissions;
+use Wikia\TemplateClassification\Logger;
```
lukaszkonieczny: We don't need to use `Wikia\TemplateClassification\Logger` as it is in the same namespace.

lukaszkonieczny:
LGTM

Please do not merge before #8805

extensions/wikia/TemplateClassification/TemplateClassification.hooks.php
```diff
@@ -15,7 +14,96 @@ public static function register() {
...
+		try {
+			( new \TemplateClassificationService() )->classifyTemplate(
+				$wgCityId,
+				$wikiPage->getId(),
+				$type,
+				\TemplateClassificationService::USER_PROVIDER,
+				$wikiPage->getUser()
+			);
+		} catch ( \Exception $e ) {
+			( new Logger() )->exception( $e, $request );
```
Grunny: The exception method doesn't have a second parameter.

extensions/wikia/TemplateClassification/TemplateClassification.hooks.php
```diff
@@ -15,7 +14,96 @@ public static function register() {
...
+
+		try {
+			( new \TemplateClassificationService() )->classifyTemplate(
+				$wgCityId,
+				$wikiPage->getId(),
+				$type,
+				\TemplateClassificationService::USER_PROVIDER,
+				$wikiPage->getUser()
+			);
+		} catch ( \Exception $e ) {
```
Grunny: Do we have to catch a generic `Exception`? Can we be more specific or is the API client throwing generic exceptions somewhere? Ideally we want to catch specific exceptions that we're expecting to occur if the service is unavailable, and other unexpected exceptions that occur in the called method should fail loudly so we find them and either fix them or also handle them. The same goes for the other catch blocks.

adamkarminski: We had this discussion yesterday and the `insertTemplateDetails` throws either an `InvalidArgumentException` (pre-request) or a `Swagger\Client\ApiException` (post-request). Both classes have only one common parent - the generic `Exception` class. If we do not catch the `InvalidArgumentException` it will in fact fail loudly. Given how the methods are used (in hooks and when rendering a page) I'd actually consider the `InvalidArgumentException` worth it. Thank you!

Grunny: Cool. :) For reference though, if you need to, you can catch multiple Exception types, i.e. with either a catch block for each exception type, or if you really want to avoid code duplication, to hack around PHP lacking the ability to catch multiple exception types in one block, you can catch the generic exception and then use `instanceof` to check if the exception is a type you want to catch and rethrow it if it isn't (though I'd prefer just having multiple catch blocks and a helper method for the handling in that case).

adamkarminski: Noted! :) Thanks!

kamilkoterba:
LGTM. Tested.

PR #8884, opened 2015-10-26
Phalanx: handle internal requests

macbre:
Improve handling of internal HTTP requests in Phalanx:
- check `X-Wikia-Internal-Request` (added by `Http::get` and DC warmer when relying requests to Reston)
- if the request is a GET and has the above header set make Phalanx ignore it (i.e. do not perform a block check nor log missing client IP)
- log internal requests (including the source) and set `X-Wikia-Is-Internal-Request` response header to ease debugging (use `WebRequestInitialized` hook introduced in #8834 - thanks @wladekb)

### DIFF
--- a/extensions/wikia/Track/Track.php
+++ b/extensions/wikia/Track/Track.php
@@ -12,7 +12,6 @@ $wgHooks['WikiaSkinTopScripts'][] = 'Track::onWikiaSkinTopScripts';
 
 class Track {
 	const BASE_URL = 'http://a.wikia-beacon.com/__track';
-	const HEADER_MIRRORED_REQUEST = 'x-mirrored-request';
 
 	private static function getURL ($type=null, $name=null, $param=null, $for_html=true) {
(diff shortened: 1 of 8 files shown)


Grunny:
LGTM.

extensions/wikia/Track/Track.php
```diff
@@ -112,7 +111,7 @@ public static function event ($event_type, $param=null) {
 	protected static function shouldTrackEvents() {
 		// Check request headers to make sure this is not a mirrored
 		// request, which could result in duplicate tracking events
-		return F::app()->wg->request->getHeader( self::HEADER_MIRRORED_REQUEST ) === false;
+		return F::app()->wg->request->isWikiaInternalRequest() === false;
```
macbre, line 114: Pinging @garthwebb - let's use a more generic header here (which will cover all internal requests - sent by DC warmer and MediaWiki itself via `Http::get`)

PR #8886, opened 2015-10-26
changed TEMPLATE_UNKNOWN to TEMPLATE_OTHER to match WestWing's convention

lgarczewski:
Changed this based on conversation with Seba.

@Wikia/community-engineering 

CC: @SebastianMarzjan @idradm

### DIFF
--- a/extensions/wikia/TemplateClassification/TemplateClassification.i18n.php
+++ b/extensions/wikia/TemplateClassification/TemplateClassification.i18n.php
@@ -33,8 +33,8 @@ $messages['en'] = [
 	'template-classification-type-nonarticle' => 'Non-article',
 	'template-classification-description-nonarticle' => 'This template is not used on article pages.',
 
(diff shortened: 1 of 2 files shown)


sebastianmarzjan:
The more I think about this naming convention the more I'm leaning towards having both:
- **OTHER** as "We explicitly don't know what it is"
- **UNKNOWN** as "We don't know what it is as we haven't attempted to classify it"

Thoughts?

kamilkoterba:
Right now we have 
- **UNKNOWN** - means template is classified as I don't know by user or by machine
- **empty string** - means "We don't know what it is as we haven't attempted to classify it"

For user facing message we fallback empty string to UNKNOWN and also we don't indicate to user the difference between machine and user classification (There's just fallback to machine if no user classification is provided)

**So I understand that you say change current UNKNOWN to OTHER and empty string to UNKNOWN and never return empty string.**

Then for user then we can think about distinguishing these two instead of merging as it it right now.

PR #8888, opened 2015-10-26
Prevent Facebook logins for restrict-login group members

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-944
See Wikia/config#1435

Since the current MW Facebook login does not use Helios, we need to do the user rights check in MW when logging in with Facebook. This should only serve as a temporary solution until the MW login path is consolidated to use Helios (most likely through Mercury).


ArturKlajnerok:
Lgtm :+1:

PR #8890, opened 2015-10-26
[nobug] Live code coverage all reports

gbenson-wikia:
Adds shell script to generate line heat maps for every used file, and to package all available reports into a single specified directory


nmonterroso:
LGTM :shipit:

PR #8891, opened 2015-10-26
ADEN-2584: add new "merge" command to script

owend:
@nandy-andy 

I just added what I needed to complete that task.  Converting the value to an array and doing a merge+unique+remove empty values seems safer than directly appending to the string.

### DIFF
--- a/maintenance/wikia/setWikiFactoryVariable.php
+++ b/maintenance/wikia/setWikiFactoryVariable.php
@@ -37,6 +37,9 @@ class SetWikiFactoryVariable extends Maintenance {
 		$this->addOption( 'set', 'Set the variable value. Note: "true" for enable or "false" for disable the extenstion', false, true, 's' );
 		$this->addOption( 'remove', 'Remove the variable value (from the Wiki)', false, false, 'r' );
 		$this->addOption( 'append', 'Append string value to existing value', false, false, 'a' );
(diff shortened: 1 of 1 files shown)


maintenance/wikia/setWikiFactoryVariable.php
```diff
@@ -125,6 +135,34 @@ public function execute() {
...
+					// value is string so we must convert to array
+					$prevValue = unserialize($varData['cv_value']);
+					$prevValue = explode($split, $prevValue);
+				}
+
+				if (is_array($prevValue)) {
+					echo "Adding new value $varValue to array" . PHP_EOL;
+					$newValue = array_filter(array_unique(array_merge($prevValue, [$varValue])));
+				} else {
+					echo "Value is not an array" . PHP_EOL;
```
nandy-andy: What do you think about a little bit different copy: `Value is not an array. Have you forgotten about --split parameter?` :)

nandy-andy:
Looks nice :smile_cat: I've added just a small comment about an error message.

I wonder how many WikiFactory variables like our key-val we have. If more than the one connected to the ticket then it should be really useful :+1: :+1: If only ours it should be useful next time we're going to change the vars again :+1: 

I also wonder if we shouldn't extend `WikiFactory::setVariable()` and pass there a reason why a variable is going to change. I'll check later if it's easy to do.

PR #8895, opened 2015-10-27
CE-2187: Add Template Classification Stats service

lukaszkonieczny:
Added 
- Template Classification Stats Service
- Method to get all template types on given wiki in TemplateClassificationService
- Auto generated code from swagger

Please CR only `TemplateClassificationService.class.php` and  `TemplateClassificationStatsService.class.php`. Rest is auto generated.


includes/wikia/services/TemplateClassificationService.class.php
```diff
@@ -77,10 +78,9 @@ public function getType( $wikiId, $pageId ) {
 	public function getDetails( $wikiId, $pageId ) {
 		$templateDetails = [];
 
-		$details = $this->getApiClient()->getTemplateDetails( $wikiId, $pageId );
+		$providers = $this->getApiClient()->getTemplateDetails( $wikiId, $pageId );
```
adamkarminski, line 94: Shouldn't we rename the `getTemplateDetails` method along with it?

includes/wikia/services/TemplateClassificationService.class.php
```diff
@@ -110,6 +110,26 @@ public function classifyTemplate( $wikiId, $pageId, $templateType, $provider, $o
...
 
 	/**
+	 * Get all classified template types on given wiki with their page id as a key
+	 *
+	 * @param int $wikiId
+	 * @return array
+	 * @throws Exception
+	 * @throws \Swagger\Client\ApiException
+	 */
+	public function getTypesOnWiki( $wikiId ) {
```
adamkarminski: So what we actually get is an array of `page_id => type`, right? Can we call it `getClassifiedTemplatesOnWiki` or sth? `getTypes` suggests that you only get a list of types used on a given wiki. :)

lib/Swagger/src/Configuration.php
```diff
@@ -89,7 +89,7 @@ class Configuration
      *
      * @var string
      */
-    protected $host = 'https://localhost/user-preference';
+    protected $host = 'https://localhost/';
```
adamkarminski, line 92: Doesn't it break `UserPreferencesApi::getUserPreferences()` etc.? Ping @kvas-damian @sqreek @nmonterroso

lib/Swagger/src/TemplateClassification/Storage/Api/TCSApi.php
```diff
@@ -92,6 +92,87 @@ public function setApiClient(ApiClient $apiClient)
...
     /**
+     * getTemplateTypesOnWiki
+     *
+     * Provides template types on wiki
+     *
+     * @param int $wiki_id Wikia ID (required)
+     * @return \Swagger\Client\TemplateClassification\Storage\Models\TemplateTypeHolder
+     * @throws \Swagger\Client\ApiException on non-2xx response
+     */
+    public function getTemplateTypesOnWiki($wiki_id)
```
adamkarminski, line 103: Same here - do we get a list of types or a list of classified templates with their types? If the latter, I'd consider renaming the method.

includes/wikia/services/TemplateClassificationService.class.php
lukaszkonieczny, line 94: My thinking is that the class name itself `TemplateClassification` tells you on what data you operate and for what this class is responsible. Do we need to repeat it in method naming?

includes/wikia/services/TemplateClassificationService.class.php
lukaszkonieczny: Actually we cannot get not classified types from `TemplateClassification`. Maybe `getTypes` is enough?

lib/Swagger/src/Configuration.php
lukaszkonieczny, line 92: I hope not :D This was auto-generated and it's common file for all services. But it;s good to ask them :)

lib/Swagger/src/TemplateClassification/Storage/Api/TCSApi.php
lukaszkonieczny, line 103: We shouldn't change this auto generated code.

includes/wikia/services/TemplateClassificationService.class.php
adamkarminski, line 94: I only ask about `getTemplateDetails` -> `getTemplateProviders` :)

lukaszkonieczny, line 94: Ohh ok :) So it's auto generated code. We shouldn't change it.

includes/wikia/services/TemplateClassificationService.class.php
adamkarminski: I'd still go with the word `Templates` in the name since ultimately what you get is a list of templates (keys are `page_id`s) with some details and not a list of types. So - `getTemplatesOnWiki`? :)

includes/wikia/services/TemplateClassificationService.class.php
adamkarminski, line 94: Ok, case closed then! :)

includes/wikia/services/TemplateClassificationService.class.php
lukaszkonieczny: Make sense :+1:

lib/Swagger/src/Configuration.php
adamkarminski, line 92: Ping @sqreek @nmonterroso

kvas-damian, line 92: Looks that this default value is overwritten by `setHost` but would be great if User Preferences Service owners can confirm that.

lib/Swagger/src/ApiClient.php
```diff
@@ -170,6 +172,11 @@ public function callApi($resourcePath, $method, $queryParams, $postData, $header
         if ($method == self::$POST) {
             curl_setopt($curl, CURLOPT_POST, true);
             curl_setopt($curl, CURLOPT_POSTFIELDS, $postData);
+        } else if ($method == self::$HEAD) {
```
nmonterroso, line 175: did we add this or is this now part of the auto-generated code?

lib/Swagger/src/Configuration.php
nmonterroso, line 92: Correct, `setHost` overrides this. It wouldn't work as-is anyway, because it's trying to access the service on the `localhost` domain.

The reason this exists is because we share a single `ApiClient` among all consumers of auto-generated code.

lib/Swagger/src/TemplateClassification/Storage/Api/TCSApi.php
nmonterroso, line 103: this method name is generated from the Swagger spec (from `operationId` iirc)

lib/Swagger/src/ApiClient.php
lukaszkonieczny, line 175: This is part of auto-generated code

adamkarminski:
LGTM

PR #8896, opened 2015-10-27
INT-89: Added mercury variable isCoppaWiki

perjg:
For INT-89, we want to integrate login functionality with the Mercury section editor.

Coppa wikias (that is a wikia where wgWikiDirectedAtChildrenByStaff or wgWikiDirectedAtChildrenByFounder is set to true) should always require login in order to edit.

These variables are not currently available to Mercury, so they were added in getWikiVariables.


perjg:
Moving to new PRs:
app: https://github.com/Wikia/app/pull/8945
config: https://github.com/Wikia/app/pull/1441

PR #8899, opened 2015-10-27
DAT-3402 move links placeholder replacement after markers are resolved

idradm:


sebastianmarzjan:
If tests pass, LGTM

sebastianmarzjan:
Is there a corresponding PR to check if it works for the case referenced in the ticket? (apitest/selenium test)?

nikodamn:
@SebastianMarzjan that's a good one. We should definitely have selenium test for links in poem tag. I just created ticket for that.

PR #8906, opened 2015-10-27
SEO-132 Fix PHP Fatal Error

gabrys:
PHP Fatal Error: Call to a member function getFullURL() on null in
/extensions/wikia/Lilly/LillyHooks.class.php on line 67


saipetch:
LGTM

PR #8909, opened 2015-10-27
SOC-1507 clear attribute cache when deleting avatars

jsutterfield:
When deleting an avatar, we weren't clearing the attribute cache as we do when an attribute is saved. This was resulting in the old attribute being shown until that cache expires. Clear the cache after a successful delete to fix that.

Ticket: https://wikia-inc.atlassian.net/browse/SOC-1507


armonr:
:+1:

garthwebb:
LGTM

PR #8910, opened 2015-10-27
VE-2004: Silencing loadHTML warning

inez:
I carefully researched the topic and it seems that there is no way to instruct DOMDocument to accept "lower" quality HTML. Specifically in HTML <div> or <figure> (block elements) are not allowed inside <p> (which accepts only inline elements) - hence it's complaining about.

Based on comments from stackoverflow it seems that such practice (silencing it) is commong practice in this situation.

(I also now run regex on the entire DIV which makes it more robust.)


pauloslund:
LGTM

PR #8911, opened 2015-10-27
VE-1997: Add chevron to pi-collapse header

pauloslund:


extensions/wikia/PortableInfobox/styles/PortableInfobox.scss
```diff
@@ -150,7 +151,31 @@
...
+		}
+	}
+}
+
 .pi-collapse-closed {
+	border-bottom: none;
+
+	.pi-header::after {
+		margin-top: -($infobox-chevron-size / 2);
+		-webkit-transform: rotate(135deg);
```
inez: I know that properties here are supposed to by ordered alphabetically - so I think "-..." comes first - but not sure - maybe check other examples?

pauloslund: Based on our SCSS guidelines it seems vendor prefixed properties can be grouped with their standard property, however I'm switching this to use a mixin as per guidelines

PR #8912, opened 2015-10-27
Reject authentication fallback to MW by default

drsnyder:
This changes the feature flag to reject fallbacks to the MW session for authentication by default. This means that a valid helios access token is required for successful authentication.

https://wikia-inc.atlassian.net/browse/SERVICES-889

The corresponding config change is https://github.com/Wikia/config/pull/1437.

@Wikia/services-team


nmonterroso:
LGTM

PR #8914, opened 2015-10-27
INT-242: Achievements: reduce number of badges for Wookiee

mari-mcmurtrie:
We want to show only edit and add photo achievements and hide everything else. We want to do it only for Japanese Starwars Wikia. Control comes from a new WikiFactory variable called "wgAchievementsEditAddPhotoOnly". If this is set to TRUE, then we do the followings.
1) Show only edit and add image badges in Special:AchievementsCustomize
2) Award only edit and add image badges so that only those badges appear in user profile, wiki activities, etc.
3) Challenge badge list in user profile, show only edit and add photo badges.
Ticket: https://wikia-inc.atlassian.net/browse/INT-242


mari-mcmurtrie:
@lgarczewski will you please CR this? It is kind of patchy work (if statement here and there) but I did not see any other ways to do it. Basically, if the global variable "wgAchievementsEditAddPhotoOnly", then we only want to enable "edit" and "add photo" badges. The rest we hide completely.

extensions/wikia/AchievementsII/services/AchAwardingService.class.php
```diff
@@ -42,10 +42,13 @@ public function migration($user_id)  {
 	 * @param Int $badge_type_id
 	 */
 	public function awardCustomNotInTrackBadge($user, $badge_type_id) {
-		wfProfileIn(__METHOD__);
+		global $wgExternalSharedDB, $wgAchievementsEditAddPhotoOnly;
 
-		global $wgExternalSharedDB;
+		if( !empty($wgAchievementsEditAddPhotoOnly) && ($badge_type_id != BADGE_EDIT && $badge_type_id != BADGE_PICTURE)) {
```
inez: Inconsistent usage of "(" and ")"

extensions/wikia/AchievementsII/services/AchAwardingService.class.php
```diff
@@ -42,10 +42,13 @@ public function migration($user_id)  {
...
+			return;
+		}
 
+		wfProfileIn(__METHOD__);
```
inez: You shouldn't move wfProfileIn down - instead you should add another wfProfileOut

mari-mcmurtrie:
@inez thanks for CR! I fixed formatting, etc. 
@lgarczewski for CR when you have time, too. Thank you!

inez:
@mari-mcmurtrie I have concerns about approach you are taking here - basically it wraps already pretty complex and nested code in another level of nesting. Have you consider any other approaches?
Two things to consider: maybe make changes only just before saving badges - some sort of badge filter method? maybe make it more generic: so have an array (as a configuration) of type of badges that are allowed instead of boolean variable?

This comes from my heart and from believe that we should always leave the code in the better shape than it was :)

mari-mcmurtrie:
@inez I worked on this as we discussed. Will you take a look? thanks!

extensions/wikia/AchievementsII/services/AchAwardingService.class.php
```diff
@@ -466,26 +486,39 @@ private function processAllInTrack() {
...
 				$this->mCounters[BADGE_LOVE][COUNTERS_COUNTER] = 1;
 			}
 		}
 		$this->mCounters[BADGE_LOVE][COUNTERS_DATE] = date('Y-m-d');
+	}
 
-		wfProfileOut(__METHOD__);
+	private function incrementCounterBy($badge_type_id, $value) {
+		$this->mCounters = is_array($this->mCounters) ? $this->mCounters : $this->mUserCountersService->getCounters();
+		if (empty($this->mCounters[$badge_type_id])) {
```
inez: Should be "empty" or "isset"?

extensions/wikia/AchievementsII/services/AchAwardingService.class.php
```diff
@@ -466,26 +486,39 @@ private function processAllInTrack() {
...
 			} else {
 				$this->mCounters[BADGE_LOVE][COUNTERS_COUNTER] = 1;
 			}
 		}
 		$this->mCounters[BADGE_LOVE][COUNTERS_DATE] = date('Y-m-d');
+	}
 
-		wfProfileOut(__METHOD__);
+	private function incrementCounterBy($badge_type_id, $value) {
+		$this->mCounters = is_array($this->mCounters) ? $this->mCounters : $this->mUserCountersService->getCounters();
```
inez: if mCounters is not an array then mUserCountersService is not set either - so what's the point of this conditional check?

extensions/wikia/AchievementsII/services/AchAwardingService.class.php
mari-mcmurtrie: So the code is refactored and the old code was using empty(). But after you pointed out, I investigated and !isset() makes more sense because 0 is true to empty(). The code was if(empty) then set to 0. Then, 0 will be evaluated to empty() again, kind of waste.

mari-mcmurtrie:
@Inez, thanks for your help. Please CR when you have time!

extensions/wikia/AchievementsII/services/AchAwardingService.class.php
```diff
@@ -44,11 +44,16 @@ public function migration($user_id)  {
...
+
+		if ( !empty( $wgAchievementsEditAddPhotoOnly ) && ( $badge_type_id != BADGE_EDIT && $badge_type_id != BADGE_PICTURE ) ) {
+			wfProfileOut(__METHOD__);
+			return;
+		}
 
 		$this->mUser = $user;
 
-		if( self::canEarnBadges( $this->mUser ) ) {
+		if ( self::canEarnBadges( $this->mUser ) ) {
```
inez, line 57: What's going on with those extra spaces? It seems like you are trying to format code properly but it's not consistent:
`if ( self::canEarnBadges( $this->mUser ) ) {`
`if ($badge === false) {`
`if(!empty($wgAchievementsEditAddPhotoOnly)) {`

extensions/wikia/AchievementsII/services/AchUserProfileService.class.php
```diff
@@ -214,10 +214,13 @@ public function getChallengesAnnotated() {
 			}
 		}
 
-		global $wgEnableAchievementsForSharing;
+		global $wgEnableAchievementsForSharing, $wgAchievementsEditAddPhotoOnly;
 
 		$challengesAnnotated = array();
 		foreach($challenges as $badge_type_id => $badge_lap) {
+			if(!empty($wgAchievementsEditAddPhotoOnly) && ($badge_type_id != BADGE_EDIT && $badge_type_id != BADGE_PICTURE)) {
```
inez: You have pretty much exact same statement in AchAwardingService.class.php. Please consider having this logic in one place.

extensions/wikia/AchievementsII/specials/templates/SpecialCustomize.tmpl.php
```diff
@@ -25,9 +25,13 @@
 	<div class="errorbox"><strong><?= $errorMsg ?></strong></div>
 <?endif;?>
 <?
-global $wgCityId, $wgScriptPath, $wgExternalSharedDB, $wgJsMimeType;
+global $wgCityId, $wgScriptPath, $wgExternalSharedDB, $wgJsMimeType, $wgAchievementsEditAddPhotoOnly;
+
 $tracks = array();
 foreach($config->getInTrackStatic() as $badgeTypeId => $trackData){
+	if (!empty($wgAchievementsEditAddPhotoOnly) && ($badgeTypeId != BADGE_EDIT && $badgeTypeId != BADGE_PICTURE)) {
```
inez: And same as above.

extensions/wikia/AchievementsII/services/AchAwardingService.class.php
mari-mcmurtrie: I try to format consistent at least inside a function. This file is already using inconsistent format everywhere. I can work on to fix it in this entire file.

mari-mcmurtrie:
@inez, Will you again CR when you have time? Thanks!!

extensions/wikia/AchievementsII/AchConfig.class.php
```diff
@@ -346,6 +351,14 @@ public function isSecret($badgeTypeId) {
 		return in_array($badgeTypeId, self::$mSecretBadges);
 	}
 
+	public function shouldShow($badgeTypeId) {
+		global $wgAchievementsEditAddPhotoOnly;
+		if (empty($wgAchievementsEditAddPhotoOnly) || in_array($badgeTypeId, self::$mEditAddphotoOnlyBadges)) {
```
Grunny: This should all be spacey per our PHP coding conventions. Same throughout the changes.

extensions/wikia/AchievementsII/services/AchAwardingService.class.php
```diff
@@ -44,11 +44,17 @@ public function migration($user_id)  {
 	public function awardCustomNotInTrackBadge($user, $badge_type_id) {
 		wfProfileIn(__METHOD__);
 
-		global $wgExternalSharedDB;
+		global $wgExternalSharedDB, $wgAchievementsEditAddPhotoOnly;
```
Grunny: The `$wgAchievementsEditAddPhotoOnly` global isn't used in this method.

Grunny:
Apart from the style issues, looks OK as a fix for the issue. A lot of this code needs cleanup, but that's outside the scope of this.

PR #8915, opened 2015-10-28
SOC-926 + Include FacebookLocale module in ResourceLoader

armonr:
- Original module to localize facebook plugins
- Add the module to the Resource Loader as well as the Asset Manager
- Fixes a JS error for users in countries not supported by Facebook localization

/CC @bkoval @kvas-damian @lizlux @garthwebb


resources/wikia/Resources.php
```diff
@@ -89,6 +96,7 @@
 			'wikia.window',
 			'wikia.mw',
 			'wikia.nirvana',
+			'wikia.fbLocale',
```
armonr, line 99: This dependency must also be added here for loader JS. This should eliminate any race condition involved with the loading of this module.

bkoval:
Does this mean we'll be loading the same javascript code twice on a page view?

armonr:
@bkoval This is really a dependency of loader module, so it must be loaded here as other dependencies are. Will it be attempted to be loaded twice? I think it is as the rest of these modules will.

lizlux:
LGTM

resources/wikia/modules/facebookLocale.js
```diff
@@ -0,0 +1,286 @@
...
+		 */
+		function getSdkUrl(inputLanguageCode) {
+			var matchingLanguages,
+				languageCode = inputLanguageCode.length === 2 ? inputLanguageCode.toLowerCase() : 'en',
+				geoCountryCode = Geo.getCountryCode(),
+				countryCode = '';
+			if (geoCountryCode) {
+				matchingLanguages = languagesOfCountry[geoCountryCode.toUpperCase()];
+				if (matchingLanguages && languageCode in matchingLanguages) {
+					countryCode = matchingLanguages[languageCode];
```
garthwebb: Here `matchingLanguages` is an array (e.g. ['af', 'xh', 'zu']) not a map.  You can access it with integers but not with a string.  You probably want:

``` js
countryCode = geoCountryCode.toUpperCase();
```

resources/wikia/modules/facebookLocale.js
```diff
@@ -0,0 +1,286 @@
...
+				}
+			}
+			if (!countryCode) {
+				countryCode = defaultCountryCodes[languageCode];
+				if (!countryCode) {
+					languageCode = 'en';
```
garthwebb: This has already been set to a default on L250

resources/wikia/modules/facebookLocale.js
armonr: Absolutely what I want, indeed, and I don't understand why I have this line! Good catch, @garthwebb

resources/wikia/modules/facebookLocale.js
```diff
@@ -0,0 +1,286 @@
...
+		};
+	}
+
+	//UMD inclusive
+	if (!context.Wikia) {
+		context.Wikia = {};
+	}
+
+	//namespace
+	context.Wikia.fbLocale = facebookLocale(context);
```
garthwebb: It doesn't look like `facebookLocale` takes any arguments.  Is there something else going on here I'm not seeing?

bkoval:
LGTM :santa:

PR #8917, opened 2015-10-28
SOC-1498 | Always fetch preferences token to use in CSRF protection for CreateNewWiki

bkoval:
@grunny @macbre 

https://wikia-inc.atlassian.net/browse/SOC-1498

I realize this code needs cleanups (for a long time now) but given that my team is short on people right now I would consider a refinement ticket instead of refactoring here, especially that it's a P2 hotfix.


Grunny:
:+1: LGTM as a quick fix.

macbre:
Agree with @Grunny ;) LGTM

PR #8920, opened 2015-10-28
CE-2885 Communicate with a user on classification action on a view page

adamkarminski:

In this pull request we implement confirmation and error messages when editing a classification from a template's view page. The desired behavior is:
1. User opens a modal with an edit form by clicking on the entry point under the template's title,
2. User selects a new type and clicks save,
3. The current classification in the entry point is **instantly changed** to the selected one and:
   
   a. If the save was successful - we display a success BannerNotification
   b. If the save failed - we display an error BannerNotification **and** change the current classification to the previous one with a nice, visual indicator that it was rolled back.

/cc: @kamilkoterba @Grunny


kamilkoterba:
LGTM

PR #8921, opened 2015-10-28
Block access to the live_code_coverage directory

gbenson-wikia:
Adds an .htaccess file to the live_code_coverage directory, to prevent accidentally serving it.


ArturKlajnerok:
Lgtm :+1:

PR #8927, opened 2015-10-28
VE-2007: Show a tooltip on header hover

pauloslund:


extensions/wikia/PortableInfobox/js/PortableInfobox.js
```diff
@@ -33,9 +33,13 @@
 
 			$collapsibleGroups.each( function( index, group ) {
 				var $group = $collapsibleGroups.eq(index),
-					$header = $group.find('.pi-header');
+					$header = $group.find('.pi-header'),
+					title = $group.hasClass('pi-collapse-closed') ? 'Open' : 'Close';
```
inez, line 37: Those are hardcoded english words and we have rule of not hardcoding such stuff - use i18n mechanism if you can.

sebastianmarzjan, line 37: Great point. This is a sitewide functionality and as such needs to be translatable

PR #8929, opened 2015-10-29
INT-236: Wikia-in-your-lang pt/pt-br bug

mari-mcmurtrie:
This bug is related with CE-1220 fix. We should not have ignored "br" part of "pt-br", rather, check "pt" part of "pt-br" then decide to request Wikia-in-your-lang.

Removed unnecessary code from Javascript and php files. Added a new function to check if we need to show WIYL. 

Ticket: https://wikia-inc.atlassian.net/browse/INT-236


mari-mcmurtrie:
@adamkarminski will you please CR? You are the most familiar with this code (previous related change you made is here: https://github.com/Wikia/app/commit/d575aeafafa7e670fd948cd264b31744e8fe78d4). 
Thank you!!

extensions/wikia/WikiaInYourLang/modules/ext.wikiaInYourLang.js
```diff
@@ -47,6 +45,18 @@ require(
 			}
 		}
 
+		// Per request we should unify dialects like pt and pt-br
+		// @see CE-1220
+		function shouldShowWikiaInYourLangWithTargetAndContentLanguage(targetLanguage, contentLanguage) {
+			var targetLanguageLangCode = targetLanguage.split('-')[0],
+			    contentLanguageLangCode = contentLanguage.split('-')[0];
+
+			if (targetLanguageLangCode === contentLanguageLangCode) {
```
adamkarminski: You can just use `return targetLanguageLangCode !== contentLanguageLangCode`

mari-mcmurtrie:
@adamkarminski Thanks, that is right! I changed code as you suggested. How does this look?

adamkarminski:
LGTM, merge when tested.

mari-mcmurtrie:
Thanks @adamkarminski !!!
We tested in sandbox. Test case is documented in ticket.

PR #8930, opened 2015-10-29
Revert "Reject authentication fallback to MW by default"

ArturKlajnerok:
Reverts Wikia/app#8912


owend:
LGTM :octocat:

drsnyder:
That'll do it. Thanks.

PR #8932, opened 2015-10-29
CE-2956 Invalidate cache after changing classification of a template

adamkarminski:
/cc: @kamilkoterba @macbre

### DIFF
--- a/includes/wikia/api/TemplateClassificationApiController.class.php
+++ b/includes/wikia/api/TemplateClassificationApiController.class.php
@@ -74,6 +74,11 @@ class TemplateClassificationApiController extends WikiaApiController {
 				TemplateClassificationService::USER_PROVIDER,
 				$userId
 			);
+
+			$title = Title::newFromId( $pageId );
+			if ( $title !== null ) {
+				$title->invalidateCache();
+			}
 		} catch ( InvalidArgumentException $e ) {
 			throw new BadRequestApiException( $e->getMessage() );
 		}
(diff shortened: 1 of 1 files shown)


includes/wikia/api/TemplateClassificationApiController.class.php
```diff
@@ -74,6 +74,11 @@ public function classifyTemplate() {
 				TemplateClassificationService::USER_PROVIDER,
 				$userId
 			);
+
+			$title = Title::newFromId( $pageId );
+			if ( $title !== null ) {
```
macbre: `instanceof Title`? :)

macbre:
:+1:

kamilkoterba:
It fixes issue for now. We should consider if that's the best place if in the future service is accessed from different point not just via TemplateClassificationApiController 
LGTM

macbre:
Good point @kamilkoterba, I'd move the call to `classifyTemplate` method.

adamkarminski:
@macbre @kamilkoterba At the current code's state it could be risky since you pass `$wikiId` as an argument to the `TemplateClassificationService::invalidateCache()`. Which is, of course, good but if you create a `Title` object there, it will use `$wgCityId` which might not be equal the one passed as the argument. You'd end up invalidating the wrong cache. :)

What I think we should do is either to implement the invalidating method in the `GlobalTitle` class or create a new class that would handle page's cache manipulation globally. But let's not worry in advance. :)

kamilkoterba:
Per discussion with @Grunny and @adamkarminski we agreed that invalidating cache is not Service responsibility. Whenever you create a client that writes to Service it should take care of cache invalidation as you may want to handle cache differently.

one example could be if you set a batch of types you may want to pass invalidation to a job.
another example is here [onArticleSaveComplete hook](https://github.com/Wikia/app/blob/dev/extensions/wikia/TemplateClassification/TemplateClassification.hooks.php#L51) this update is a part of article save so invalidation is done as a part of it, you we want to invalidate it twice.

/cc @macbre @cmiller-wikia

PR #8933, opened 2015-10-29
XW-395 | Remove deprecated /api/v1/* Mercury paths

Warkot:
`/api/v1/` → `/api/mercury/`
Now Mercury API can be accessed from any client, not only mobile.

### DIFF
--- a/extensions/wikia/MercuryApi/MercuryApiHooks.class.php
+++ b/extensions/wikia/MercuryApi/MercuryApiHooks.class.php
@@ -91,8 +91,7 @@ class MercuryApiHooks {
 
 		if ( $title->inNamespaces( NS_MAIN ) ) {
 			// Mercury API call from Ember.js to Hapi.js e.g.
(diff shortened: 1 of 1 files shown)


hakubo:
LGTM :rocket:

sebastianmarzjan:
Have you checked apitests? :)
I'd also suggest taking a look at Kibana in case we have external consumers using this API

Warkot:
Good point! :smile: I wanted to run them and check if they fail, but as you mentioned on slack it might be delusive (at first glance). :sparkles:

sebastianmarzjan:
It's fixed now, so should be good to check per offline convo

Warkot:
```
Ran 329 tests in 150.213s

OK
```

…

```
Finished: SUCCESS
```

…and no references to Mercury API found in https://github.com/Wikia/api-tests repo.

sebastianmarzjan:
:+1:

PR #8934, opened 2015-10-29
CE-2958 Add auto type prefill for infobox conversion

kamilkoterba:
https://wikia-inc.atlassian.net/browse/CE-2958

@adamkarminski @Grunny

### DIFF
--- a/extensions/wikia/TemplateClassification/TemplateClassification.hooks.php
+++ b/extensions/wikia/TemplateClassification/TemplateClassification.hooks.php
@@ -95,25 +95,7 @@ class Hooks {
 			return true;
 		}
 
-		$title = $editPage->getTitle();
-
-		try {
-			$templateType = ( new \TemplateClassificationService() )->getType( $wgCityId, $title->getArticleID() );
-		} catch ( ApiException $e ) {
-			( new Logger() )->exception( $e );
-			/**
-			 * If the service is unreachable - fill the field with a not-available string
-			 * which instructs front-end tools to skip the classification part.
-			 */
-			$templateType = \TemplateClassificationService::NOT_AVAILABLE;
-		}
-
-		// Fallback to infobox on template draft conversion for not existent classification
-		if ( ( $templateType === '' || $templateType === \TemplateClassificationService::NOT_AVAILABLE )
-			&& $this->isTemplateDraftConversion( $out->getRequest(), $title )
-		) {
-			$templateType = \TemplateClassificationService::TEMPLATE_INFOBOX;
-		}
+		$templateType = $this->getTemplateTypeForEdit( $editPage->getTitle(), $wgCityId );
 
 		$editPage->addHiddenField([
 			'name' => 'templateClassificationType',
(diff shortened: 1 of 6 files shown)


extensions/wikia/TemplateClassification/TemplateClassification.hooks.php
```diff
@@ -107,6 +108,13 @@ public static function onEditPageShowEditFormFields( \EditPage $editPage, \Outpu
 			$templateType = \TemplateClassificationService::NOT_AVAILABLE;
 		}
 
+		// Fallback to infobox on template draft conversion for not existent classification
```
adamkarminski: You can actually move it before the request above. If it is a Draft and the conversion param is there it means that it is an infobox.

extensions/wikia/TemplateClassification/TemplateClassification.hooks.php
```diff
@@ -191,6 +199,16 @@ public function onEditPageLayoutExecute( \EditPageLayoutController $editPage ) {
 	}
 
 	/**
+	 * Determines whether a title is a draft in conversion mode initiated
+	 * @param IRequest $request
+	 * @param \Title $title
+	 * @return bool
+	 */
+	public function isTemplateDraftConversion( IRequest $request, \Title $title ) {
```
adamkarminski: I think that since we use it in `TemplateClassificationView` as well it should not be a part of the `Hooks` class.

includes/wikia/traits/TitleTrait.php
```diff
@@ -152,4 +152,10 @@ public function getIndirectLinks( $level = 0 ) {
 
 		return $rows;
 	}
+
+	public function isTemplateDraft() {
+		global $wgEnableTemplateDraftExt;
+
+		return !empty( $wgEnableTemplateDraftExt ) && TemplateDraftHelper::isTitleDraft( $this );
```
adamkarminski: I think that the `isTitleDraft` logic should actually be placed in this function. You can leave `TamplateDraftHelper::isTitleDraft()` as a wrapper for it so we don't change every usage of it.

extensions/wikia/TemplateClassification/TemplateClassificationView.php
```diff
@@ -27,6 +27,13 @@ public function renderTemplateType( $wikiId, \Title $title, $user, $fallbackMsg
 			return $fallbackMsg;
 		}
 
+		// Fallback to infobox on template draft for not existent classification
```
adamkarminski, line 25: Same as above - place it above the request.

includes/wikia/traits/TitleTrait.php
kamilkoterba: I wouldn't say that's the best approach as `isTitleDraft` uses messages from TemplateDraft extension and checking if global wgEnableTemplateDraftExt for every TemplateDraft internal usage doesn't make sens as you already are a part of extension.
I see it better as it is above.

adamkarminski: Ok, makes sense. Thanks!

PR #8941, opened 2015-10-29
Add a generateToken method to HeliosClient

drsnyder:
Adding `generateToken( $userId )` to the MediaWiki HeliosClient. This will allow us to generate a token and add it to the cookie when we do facebook logins in MW.

https://wikia-inc.atlassian.net/browse/SERVICES-963

@Wikia/services-team


nmonterroso:
:shipit:

PR #8942, opened 2015-10-29
SEO-134 HTML <title> as a list

gabrys:
Changing the code for generating the `<title>` tags, so that it accepts a list of parts rather than a string.

The code still acts in a compatibility mode, which means it preserves the behavior of reading the templates saved in MW messages: pagetitle.

Ideally that code should be dropped and replaced with the one that only reads the wikia-pagetitle-site for the middle part ("Site") of the "Article - Site - Wikia" template.


saipetch:
LGTM

PR #8944, opened 2015-10-29
SOC-685 Remove "Mark all pages from Weekly Digest as visited" preferece

jsutterfield:
We're removing the "Mark all pages from my Weekly Digest as visited" preference in my preferences. That's the default behavior now anyways, and we don't support changing that behavior. So just remove the preference.

Ticket: https://wikia-inc.atlassian.net/browse/SOC-685


extensions/wikia/GlobalWatchlist/GlobalWatchlist.i18n.php
```diff
@@ -1,425 +0,0 @@
-<?php
```
jsutterfield, line 1: Neither of these messages in this file are being used anymore.

extensions/wikia/GlobalWatchlist/GlobalWatchlistHooks.class.php
```diff
@@ -8,16 +8,10 @@ public static function getPreferences( $user, &$defaultPreferences ) {
 
 		$defaultPreferences['watchlistdigest'] = [
 			'type' => 'toggle',
-			'label-message' => 'tog-watchlistdigest',
-			'section' => 'watchlist/advancedwatchlist',
+			'label-message' => 'tog-watchlistdigest-v2',
+			'section' => 'emailv2/email-me-v2',
```
jsutterfield, line 12: By updating these `section` and `labe-message` values here, we can remove the code which is overriding these values [here](https://github.com/Wikia/app/pull/8944/files#diff-19b0d1c2e76c0dcf56b1bdf2b0eeab7cL160).

armonr:
LGTM

bkoval:
LGTM!

PR #8946, opened 2015-10-30
Masthead: improve detection of default avatars

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-1617
- after avatars migration even the default ones are stored with their full URL
- removing default avatars is handled by MW even when the service is enabled

@jcellary / @jsutterfield


jcellary:
Actually for normal avatars we're also storing the full url (but with an image uuid instead of original file name) - this was changed by Paweł some time ago. Before that we were storing only the id, but mobile apps requested for us to change that. I'm not sure if in such case your change will still work?

macbre:
It will - take a look at the test cases :)

jcellary:
Ok, sorry I doubted you :D LGTM

jsutterfield:
:+1:

PR #8948, opened 2015-10-30
(CE-2812) Add a custom prefix to edit interface whitelist

Grunny:
Add a prefix to edit interface whitelist that can be used for custom
messages by users.


adamkarminski:
LGTM

PR #8950, opened 2015-10-30
VE-2010: Simplified and optimized JS for collapsible groups

inez:
Now most of the logic is delayed to be execute when user actually click on the tab

### DIFF
--- a/extensions/wikia/PortableInfobox/js/PortableInfobox.js
+++ b/extensions/wikia/PortableInfobox/js/PortableInfobox.js
@@ -29,18 +29,14 @@
 
 	var CollapsibleGroup = {
 		init: function() {
-			var $collapsibleGroups = $('.pi-collapse');
-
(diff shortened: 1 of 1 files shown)


extensions/wikia/PortableInfobox/js/PortableInfobox.js
```diff
@@ -29,18 +29,14 @@
...
-
-			$collapsibleGroups.each( function( index, group ) {
-				var $group = $collapsibleGroups.eq(index),
-					$header = $group.find('.pi-header');
-
-				$header.click( function() {
-					$group.toggleClass('pi-collapse-closed');
-				});
-			});
+			$('.pi-header').click( function( e ) {
```
pauloslund, line 32: I think my only issue with this is that you are binding events for all .pi-headers, even if they don't belong to a 
`.pi-collapse` group. Might work better to do something like `$('.pi-collapse .pi-header').click(...`

PR #8953, opened 2015-10-30
Ensure the helios access_token is set during the facebook authentication flow

drsnyder:
This PR ensures that the helios `access_token` is set during the facebook authentication flow. With this PR, I'm also adding a cookie helper under the `Auth` namespace. The cookie handling is sprinkled around the code base in core, in the Helios extension and in the `UserLogin` extension. The related cookie handling in this PR should enable us to start the process of consolidating that functionality into one place.

https://wikia-inc.atlassian.net/browse/SERVICES-962

@Wikia/services-team @wladekb @michalroszka

### DIFF
--- a/lib/Wikia/src/Service/User/Auth/HeliosCookieHelper.php
+++ b/lib/Wikia/src/Service/User/Auth/HeliosCookieHelper.php
@@ -23,7 +23,7 @@ class HeliosCookieHelper implements CookieHelper {
 
 
 	/**
-		* @param HeliosClient $client
+	 * @param HeliosClient $client
 	 */
 	public function setHeliosClient( HeliosClient $client ) {
 		$this->heliosClient = $client;
--- a/lib/Wikia/src/Service/User/Auth/HeliosCookieHelper.php
+++ b/lib/Wikia/src/Service/User/Auth/HeliosCookieHelper.php
@@ -1,0 +1,79 @@
+<?php
+
+namespace Wikia\Service\User\Auth;
+
+use Wikia\HTTP\Response;
+use Wikia\Service\Helios\HeliosClient;
+
+class HeliosCookieHelper implements CookieHelper {
+
+	const ACCESS_TOKEN_COOKIE_NAME = 'access_token';
+	// This is set to 6 months,(365/2)*24*60*60 = 15768000
+	const ACCESS_TOKEN_COOKIE_TTL = 15768000;
+	const NO_COOKIE_PREFIX = '';
+
+	private $heliosClient;
+
+	/**
+	 * @Inject({Wikia\Service\Helios\HeliosClient::class})
+	 */
+	public function __construct( HeliosClient $client ) {
+		$this->heliosClient = $client;
+	}
+
+
+	/**
+		* @param HeliosClient $client
+	 */
+	public function setHeliosClient( HeliosClient $client ) {
+		$this->heliosClient = $client;
+	}
+
+
+	/**
+	 * Set the authentication cookie using the user id. This method will retrieve a new
+	 * authentication token from Helios and then set the authentication cookie.
+	 *
+	 * Warning: this should only be called once the user has been authenticated (e.g. from a 3rd party)
+	 *
+	 * @param int $userId
+	 * @param Response $response
+	 */
+	public function setAuthenticationCookieWithUserId( $userId, Response $response ) {
+		$tokenData = $this->heliosClient->generateToken( $userId );
+		$this->setAuthenticationCookieWithToken( $tokenData->access_token, $response );
+	}
+
+	/**
+	 * Set the authentication cookie for the user using the given access token.
+	 *
+	 * @param string $accessToken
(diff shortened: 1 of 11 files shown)


lib/Wikia/src/Service/User/Auth/HeliosCookieHelper.php
```diff
@@ -0,0 +1,79 @@
...
+
+use Wikia\HTTP\Response;
+use Wikia\Service\Helios\HeliosClient;
+
+class HeliosCookieHelper implements CookieHelper {
+
+	const ACCESS_TOKEN_COOKIE_NAME = 'access_token';
+	// This is set to 6 months,(365/2)*24*60*60 = 15768000
+	const ACCESS_TOKEN_COOKIE_TTL = 15768000;
+	const NO_COOKIE_PREFIX = '';
```
harnash: `NO_COOKIE_PREFIX` -> `COOKIE_PREFIX` name should reflect intent not the actual, temporary content.

lib/Wikia/src/Service/User/Auth/HeliosCookieHelper.php
```diff
@@ -0,0 +1,79 @@
...
+	/**
+	 * @Inject({Wikia\Service\Helios\HeliosClient::class})
+	 */
+	public function __construct( HeliosClient $client ) {
+		$this->heliosClient = $client;
+	}
+
+
+	/**
+		* @param HeliosClient $client
```
harnash: Indentation

drsnyder: 1f385db

ArturKlajnerok:
Lgtm :+1: . Let's just let know about it to @garthwebb & @kvas-damian

drsnyder:
I created https://wikia-inc.atlassian.net/browse/SERVICES-970 to follow up on refactoring the cookie handling.

kvas-damian:
@Wikia/services-team for future please remember about code ownership and wait for CR from Social Team before merging and pushing to production.

lib/Wikia/src/Service/User/Auth/HeliosCookieHelper.php
```diff
@@ -0,0 +1,79 @@
...
+
+use Wikia\HTTP\Response;
+use Wikia\Service\Helios\HeliosClient;
+
+class HeliosCookieHelper implements CookieHelper {
+
+	const ACCESS_TOKEN_COOKIE_NAME = 'access_token';
+	// This is set to 6 months,(365/2)*24*60*60 = 15768000
+	const ACCESS_TOKEN_COOKIE_TTL = 15768000;
+	const COOKIE_PREFIX = '';
```
kvas-damian, line 13: Having the same constants and methods in many places doesn't help in future development. Current state is really confusing - there is no `@Deprecated` annotation in original class, code is still present in original class.

lib/Wikia/src/Service/User/Auth/HeliosCookieHelper.php
```diff
@@ -0,0 +1,79 @@
...
+	/**
+	 * Set the authentication cookie using the user id. This method will retrieve a new
+	 * authentication token from Helios and then set the authentication cookie.
+	 *
+	 * Warning: this should only be called once the user has been authenticated (e.g. from a 3rd party)
+	 *
+	 * @param int $userId
+	 * @param Response $response
+	 */
+	public function setAuthenticationCookieWithUserId( $userId, Response $response ) {
```
kvas-damian, line 42: Does it matter on which `$response` object we set cookie?
In order to simplify flow I would create a getter that will always return response from main request context. What do you think about it?

bkoval:
Is it released yet? If not, could you run automated tests to make sure there's no regression? Thanks!

lib/Wikia/src/Service/User/Auth/HeliosCookieHelper.php
drsnyder, line 42: > In order to simplify flow I would create a getter that will always return response from main request context. What do you think about it?

Can you show me an example in the code. I'm not sure I understand what you mean.

drsnyder:
> Is it released yet? If not, could you run automated tests to make sure there's no regression? Thanks!

Yes, it is in production. We worked with @bekcunning to understand how the tests work. She indicated to us that that tests are currently broken so we did manual testing on sandbox-s3.

lib/Wikia/src/Service/User/Auth/HeliosCookieHelper.php
drsnyder, line 13: > Having the same constants and methods in many places doesn't help in future development. Current state is really confusing - there is no @Deprecated annotation in original class, code is still present in original class.

I created a ticket (see above) to follow up and refactor the helios "user". IMO we should decouple the cookie operations both from the user and the helios "user". As part of this we also need to revisit our cookies but I think that's probably a next step after we consolidate the functionality in one place.

PR #8955, opened 2015-11-02
SOC-1496 | Disable old login dropdown when new auth modal enabled

kvas-damian:
https://wikia-inc.atlassian.net/browse/SOC-1496#add-comment


rafalkalinski:
LGTM

PR #8956, opened 2015-11-02
CE-2928: Do not log if add/remove action was not done

lukaszkonieczny:
Do not log if add/remove action was not done


lgarczewski:
LGTM

PR #8958, opened 2015-11-02
CE-2949 Hotfix for failing user renames

adamkarminski:
The problem with renames was that it triggered a mechanism introduced by @owend that logs a user's email change. However, a `$new_email` var is empty in this case which results in DBException since neither `old_email` nor `new_email` allows `NULL` values.

A more permanent fix would be probably to untie this logging mechanisms from a user's rename process, but  would like to open a discussion with @owend.

/cc: @Grunny @lukaszkonieczny


Grunny:
:+1:

PR #8959, opened 2015-11-02
PLATFORM-1582: Fix error and exception logging in tasks

wladekb:
There were a few reports that failed tasks don't leave any trace about that fact. This should improve logging in two cases:
- tasks execution throws an exception
- Catchable Fatal Errors were never reported by our customized error handler

/cc @macbre @kvas-damian


macbre:
:+1:

kvas-damian:
LGTM

PR #8960, opened 2015-11-02
[SERVICES-959] Removed possible external reset password request.

ArturKlajnerok:
As this is no longer used or needed, I'm removing possible external access to this controller.

/cc @Wikia/services-team @jogura


nmonterroso:
Does the password reset functionality now use Helios?

ArturKlajnerok:
No it doesn't. But this controller is also used only internally by MW.

nmonterroso:
LGTM

PR #8962, opened 2015-11-02
Merging PR #8953 int release-344

drsnyder:
Merging PR #8953 into release-344 for Tuesday November 3rd. 

Blocked by the testing of #8961 on sandbox-s3. Once the test is complete and cleared then we can merge this one.

@Wikia/services-team

PR #8963, opened 2015-11-02
Reject authentication fallback to MW sessions

drsnyder:
Second try. The first attempt broke the facebook integration. That was fixed with SERVICES-962.

https://wikia-inc.atlassian.net/browse/SERVICES-889

See also
- #8953 (dev for 962)
- #8961 (release-343 with this change and #8953)
- #8962 (release-344 with this change and #8953)

@Wikia/services-team


ArturKlajnerok:
Lgtm :+1:

PR #8964, opened 2015-11-02
add boolean toggle for do-nothing cache for preference cache

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-938

module configuration for using a noop cache for the user preference service, so that all requests for a user's preferences are sent to the service

### DIFF
--- a/lib/Wikia/src/Service/User/Preferences/PreferenceModule.php
+++ b/lib/Wikia/src/Service/User/Preferences/PreferenceModule.php
@@ -2,6 +2,7 @@
 
 namespace Wikia\Service\User\Preferences;
 
+use Doctrine\Common\Cache\VoidCache;
 use Interop\Container\ContainerInterface;
 use User;
 use Wikia\Cache\BagOStuffCacheProvider;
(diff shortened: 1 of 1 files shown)


ArturKlajnerok:
Lgtm :+1:

lib/Wikia/src/Service/User/Preferences/PreferenceModule.php
```diff
@@ -14,12 +15,17 @@
...
 class PreferenceModule implements Module {
 	const PREFERENCE_CACHE_VERSION = 3;
+	const PREFERENCE_CACHE = false;
 
 	public function configure( InjectorBuilder $builder ) {
 		$builder
 			->bind( PreferenceService::class )->toClass( PreferenceServiceImpl::class )
 			->bind( PreferencePersistence::class )->toClass( PreferencePersistenceSwaggerService::class )
 			->bind( PreferenceServiceImpl::CACHE_PROVIDER )->to( function() {
+				if ( !self::PREFERENCE_CACHE ) {
```
sqreek, line 25: There should be no logic inside binding modules.
This is anti pattern and should be avoided if possible.

nmonterroso, line 25: Hm sorry I didn't see this comment until now.

Yah I agree that this probably isn't the best way to do this, but this isn't intended to live in the codebase for long. We will keep this until we're satisfied that the preference service can handle the traffic load, and then remove the `PREFERENCE_CACHE` constant as well as this conditional. Does that seem ok?

sqreek, line 25: yep, until we will remember about it, I think it's ok.

Live Experience shows people doesn't remember about this kind of thing so -1
Experience in our team shows we are remember this kind of thing so +2

PR #8965, opened 2015-11-03
(CE-2841) Disable starting transactions on slave connections

Grunny:
This disables automatically starting transactions when making queries on
slave connections.

/cc @macbre @drozdo


macbre:
LGTM

PR #8966, opened 2015-11-03
Nirvana: Decrease the noise in reported transactions

macbre:
We're getting many "interesting" requests that end up as unique NewRelic transactions. Decrease the noise by setting transaction name to `api/nirvana/notFound` when either controller or method can not be found.

```
/wikia.php?controller=Lightbox&method=getThumbImagesaaa
TransactionTrace: notification: ["onTypeChange","api\/nirvana\/notFound"]

/wikia.php?controller=Lightbox&method=getThumbImages
TransactionTrace: notification: ["onTypeChange","api\/nirvana\/Lightbox"]
```

Alternatively we can move [this block](https://github.com/Wikia/app/blob/dev/wikia.php#L19-L24) to be executed only on successful requests. But I'm not sure if that won't happen too late.

@wladekb / @jcellary

### DIFF
--- a/wikia.php
+++ b/wikia.php
@@ -1,39 +1,39 @@
 <?php
 
 // This is from google translate, just return early.
-if ( $_SERVER['REQUEST_METHOD'] == 'OPTIONS') {
-	header ( "HTTP/1.1 200", true, 200);
+if ( $_SERVER['REQUEST_METHOD'] == 'OPTIONS' ) {
+	header ( "HTTP/1.1 200", true, 200 );
 	return;
 }
 
 // prevent $_GET['title'] from being overwritten on API calls (BAC-906)
-define('DONT_INTERPOLATE_TITLE', true);
+define( 'DONT_INTERPOLATE_TITLE', true );
 
 // Initialise common MW code
 require ( dirname( __FILE__ ) . '/includes/WebStart.php' );
 
-if ($wgProfiler instanceof Profiler) {
-	$wgProfiler->setTemplated(true);
+if ( $wgProfiler instanceof Profiler ) {
+	$wgProfiler->setTemplated( true );
 }
 
 // Construct a tag for newrelic -- wgRequest is global in this scope
-Transaction::setEntryPoint(Transaction::ENTRY_POINT_NIRVANA);
-if ( is_object($wgRequest) ) {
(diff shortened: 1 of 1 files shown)


wikia.php
```diff
@@ -63,6 +63,14 @@
 		] );
 	}
 
+	// PLATFORM-1633: decrease the noise in reported transactions
+	$ex = $response->getException();
+
+	if ( $ex instanceof ControllerNotFoundException|| $ex instanceof MethodNotFoundException ) {
+		Transaction::setAttribute( Transaction::PARAM_CONTROLLER, 'other' );
+		Transaction::setAttribute( Transaction::PARAM_METHOD, '' );
```
wladekb, line 71: Maybe we should log such cases using WikiaLogger?

wikia.php
```diff
@@ -63,6 +63,14 @@
 		] );
 	}
 
+	// PLATFORM-1633: decrease the noise in reported transactions
+	$ex = $response->getException();
+
+	if ( $ex instanceof ControllerNotFoundException|| $ex instanceof MethodNotFoundException ) {
```
wladekb:
```
wladek@wladek-pc:~$ php -r 'echo 1 instanceof Exception;'
PHP Fatal error:  instanceof expects an object instance, constant given in Command line code on line 1
```

It may also be good to check if `$ex` is an object before doing these checks.

wladekb: nvm.

```
wladek@wladek-pc:~$ php -r '$a = 1; echo $a instanceof Exception;'
```

does not throw any error nor warning

wikia.php
macbre, line 71: Good point, added in 556934f

wikia.php
```diff
@@ -63,6 +63,19 @@
 		] );
 	}
 
+	// PLATFORM-1633: decrease the noise in reported transactions when we return HTTP 404
+	$ex = $response->getException();
+
+	if ( $ex instanceof NotFoundException ) {
```
macbre, line 69: This covers both `ControllerNotFoundException` and `MethodNotFoundException`

macbre:
I'll push it live tomorrow.

macbre:
Backported to production in e2a23a0

macbre:
Reverted c25d0e37a476c960c8fca8e23443fd2627bc4d6f to avoid catching responses that deliberately set HTTP 404 response code.

PR #8973, opened 2015-11-04
Use slave DB connections for LocalFile cache misses

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-1636

We're being hit by spikes of QPS on DB master nodes. Number of queries from `LocalFile::loadFromDB` is 30 times higher than usual. What makes this even worse is the fact that this method always uses master to perform select query. And we make ~475 mm of them daily.

This commit backports changes from MW upstream that routes most of `LocalFile::loadFromDB` queries to slaves.

Backporting wikimedia/mediawiki@bb06510 from MW 1.24

@drozdo / @michalroszka / @wladekb

### DIFF
--- a/includes/filerepo/file/LocalFile.php
+++ b/includes/filerepo/file/LocalFile.php
@@ -60,6 +60,13 @@ class LocalFile extends File {
 
 	protected $repoClass = 'LocalRepo';
 
+	/** @var int UNIX timestamp of last markVolatile() call */
+	private $lastMarkedVolatile = 0;
+
+	const LOAD_VIA_SLAVE = 2; // integer; use a slave to load the data
+
+	const VOLATILE_TTL = 300; // integer; seconds
+
 	/**
 	 * Create a LocalFile from a title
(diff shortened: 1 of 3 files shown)


includes/filerepo/file/LocalFile.php
```diff
@@ -160,6 +167,8 @@ function __construct( $title, $repo ) {
 	/**
 	 * Get the memcached key for the main data for this file, or false if
 	 * there is no access to the shared cache.
+	 *
+	 * @return string|bool
```
ghost, line 171: Not really a comment, just a sigh to be taken into consideration. Typecasting and functions returning values of multiple types. I agree, that cache keys might not be the case. But in general in such caches we would have to check for type, as strings `''` and `'0'` evaluate to boolean `false`. I am more into _return one specific type or null_ and `!is_null()` checks convention. I treat boolean values as answers to questions - that should never be `null`.

ghost:
:+1:

PR #8974, opened 2015-11-04
(CE-2959) Make sure the JS page is in review when viewing diff

Grunny:
We rely on the revision being in review in a few places, including when
approving and rejecting the changes. This change makes it so that if a
reviewer visits a diff page of a submitted JS change with the contentreview
URL parameter, we will move it to in review if it isn't already.
that the diff is in review.


adamkarminski:
LGTM

PR #8976, opened 2015-11-04
CE-2962: Change a way to add hidden field to template edit mode

lukaszkonieczny:
Change a way to add hidden field to template edit mode to work in both monobook and oasis and fix js error in monobook.

ping @kamilkoterba


extensions/wikia/TemplateClassification/TemplateClassification.hooks.php
```diff
@@ -91,17 +91,13 @@ public function onEditPageMakeGlobalVariablesScript( array &$aVars ) {
...
-		$templateType = $this->getTemplateTypeForEdit( $editPage->getTitle(), $wgCityId );
+		if ( ( new Permissions() )->shouldDisplayEntryPoint( $context->getUser(), $context->getTitle() ) ) {
+			$templateType = $this->getTemplateTypeForEdit( $editPage->getTitle(), $wgCityId );
 
-		$editPage->addHiddenField([
-			'name' => 'templateClassificationType',
-			'value' => $templateType,
-			'type' => 'hidden',
-		]);
+			$out->addHTML( \Html::hidden( 'templateClassificationType', $templateType ) );
```
adamkarminski: Is it good to just add the field to the output without a context of the form that is sent? Or maybe the field actually **is** in the scope of the form when added this way?

extensions/wikia/TemplateClassification/scripts/TemplateClassificationModal.js
```diff
@@ -79,7 +79,7 @@ function ($, mw, loader, nirvana, tracker, labeling) {
 			// Mark selected type
 			$preselectedType = $classificationForm.find('input[value="' + templateType + '"]');
 
-			if ($preselectedType.length > 0) {
+			if ($preselectedType) {
```
adamkarminski: Why has this check been altered?

extensions/wikia/TemplateClassification/TemplateClassification.hooks.php
lukaszkonieczny: Yes, this hook is run while form is created. See this function https://github.com/Wikia/app/blob/dev/includes/EditPage.php#L1764

extensions/wikia/TemplateClassification/scripts/TemplateClassificationModal.js
lukaszkonieczny: Because checking length of `undefined` causes error

extensions/wikia/TemplateClassification/TemplateClassification.hooks.php
adamkarminski: :+1:

extensions/wikia/TemplateClassification/scripts/TemplateClassificationModal.js
adamkarminski: I'd cast it to a boolean value with a `!!`, so `if (!!$preselectedType) {` etc.

adamkarminski:
LGTM

PR #8978, opened 2015-11-04
VE-1895: Add surfaceReady event to timings to fix analytics

pauloslund:
It seems that the missing events were being sent to GA, however we were sending `NaN` for the value on some of them which I believe is causing Google to ignore those events.

This PR also brings back some of the events that were missed.


inez:
LGTM

PR #8981, opened 2015-11-05
CE-3002 Fix forcing modal and make sure we don't automatically submit…

kamilkoterba:
… machine classification as user classification

Decided to explicitly add a meaning for empty string returned from service class as UNCLASSIFIED.
Added second hidden field for making sure we don't submit change if type didn't change.

https://wikia-inc.atlassian.net/browse/CE-3002

@Wikia/community-engineering

### DIFF
--- a/extensions/wikia/TemplateClassification/TemplateClassification.hooks.php
+++ b/extensions/wikia/TemplateClassification/TemplateClassification.hooks.php
@@ -26,7 +26,8 @@ class Hooks {
 
 	/**
 	 * Save template type passed from article creation
-	 * template type is stored in templateClassificationType hidden field
+	 * changed template type is stored in templateClassificationTypeNew hidden field.
+	 * Previous type is stored in templateClassificationTypeCurrent.
 	 *
 	 * @param \WikiPage $wikiPage
 	 * @param \User $user
@@ -38,13 +39,17 @@ class Hooks {
 		global $wgCityId;
 
 		$request = \RequestContext::getMain()->getRequest();
-		$type = $request->getVal( 'templateClassificationType' );
+		$typeNew = $request->getVal( 'templateClassificationTypeNew' );
+		$typeCurrent = $request->getVal( 'templateClassificationTypeCurrent' );
 
 		/**
 		 * The service was not available when the field's value was set
 		 * so we exit early to prevent polluting of the results.
 		 */
-		if ( !isset( $type ) || $type === \TemplateClassificationService::NOT_AVAILABLE ) {
+		if ( empty( $typeNew )
+			|| $typeNew === \TemplateClassificationService::NOT_AVAILABLE
+			|| $typeNew === $typeCurrent
+		) {
 			return true;
 		}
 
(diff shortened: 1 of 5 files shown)


extensions/wikia/TemplateClassification/TemplateClassification.hooks.php
```diff
@@ -98,11 +103,17 @@ public function onEditPageShowEditFormFields( \EditPage $editPage, \OutputPage $
 		$templateType = $this->getTemplateTypeForEdit( $editPage->getTitle(), $wgCityId );
 
 		$editPage->addHiddenField([
-			'name' => 'templateClassificationType',
+			'name' => 'templateClassificationTypeCurrent',
 			'value' => $templateType,
 			'type' => 'hidden',
 		]);
 
+		$editPage->addHiddenField([
```
lukaszkonieczny: This usage causes Fatal Error on Monobook. Please use approach from this fix https://github.com/Wikia/app/pull/8976/files#diff-e9367e7e28b3950efc1f77b31ae7bc06R99

extensions/wikia/TemplateClassification/scripts/TemplateClassificationModal.js
```diff
@@ -10,7 +10,7 @@ function ($, mw, loader, nirvana, tracker, labeling) {
 	'use strict';
 
 	var $classificationForm,
-		$preselectedType,
+		$preselectedType = $(),
```
lukaszkonieczny: I think we don't need to init this value as empty. Probably this fix is enough https://github.com/Wikia/app/pull/8976/files#diff-d5bde9afc6a40ff66cf72bf44a3fc4b2R82

extensions/wikia/TemplateClassification/scripts/TemplateClassificationInEdit.js
```diff
@@ -11,10 +11,12 @@ define('TemplateClassificationInEdit',
...
 		'use strict';
 
-		var $editFormHiddenTypeField;
+		var $editFormHiddenTypeFieldCurrent,
+			$editFormHiddenTypeFieldNew;
 
 		function init() {
-			$editFormHiddenTypeField = $('#editform').find('[name=templateClassificationType]');
+			$editFormHiddenTypeFieldCurrent = $('#editform').find('[name=templateClassificationTypeCurrent]');
+			$editFormHiddenTypeFieldNew = $('#editform').find('[name=templateClassificationTypeNew]');
```
lukaszkonieczny: We can cache `$('#editform')` in var. I think that filtering these fields by class name will be more performant

extensions/wikia/TemplateClassification/scripts/TemplateClassificationModalForce.js
```diff
@@ -1,19 +1,20 @@
...
 		'use strict';
 		function forceType() {
-			var $typeField = $('#editform').find('input[name=templateClassificationType]');
+			var $editform = $('#editform'),
+				$typeFieldCurrent = $editform.find('input[name=templateClassificationTypeCurrent]'),
+				$typeFieldNew = $editform.find('input[name=templateClassificationTypeNew]');
 
-			if ($typeField.length > 0) {
-				if ($typeField.val() !== '') {
+			if (($typeFieldCurrent.length > 0 && $typeFieldCurrent.val() !== '') ||
```
lukaszkonieczny: It will cause error if `$typeFieldCurrent` will be `undefined`

adamkarminski:
Tested.

PR #8985, opened 2015-11-05
ADEN-2484 New Taboola

fraszczakszymon:
New Taboola integration with modules below article and in right rail


extensions/wikia/AdEngine/AdEngine2Controller.class.php
```diff
@@ -57,4 +57,13 @@ public function ad() {
 		$this->slotName = $this->request->getVal('slotName');
 		$this->showAd = AdEngine2Service::shouldShowAd($this->pageTypes);
 	}
+
+	/**
+	 * Action to display an ad (or not)
```
nandy-andy: I can't come with a better idea. Maybe @jacekjursza has some... But since I'm not a big fan of changing now methods' name of the base class (`AdEngine2Controller::ad()` to `AdEngine2Controller::adSlot()` and `AdEngine2Controller::adEmptyContainer()` to `AdEngine2Controller::recoverableSlot()`) let's just leave a comment to our future selves ;)

```
/**
 * Action to display a recoverable ad product (or not)
 * 
 * It differs with AdEngine2Controller::ad() in two ways:
 * - no .wikia-ad class added to the element,
 * - no script tag with push just after the element. 
 */
```

jacekjursza: +1 for comment

jacekjursza:
LGTM

PR #8986, opened 2015-11-05
SEO-143 Tweaks to links extracting logic for lilly

gabrys:
- No more tracking links to (and from) Finnish starwars as it's not
  hosted on Wikia anymore
- Add English as the a language (so links to the English Star Wars are
  also tracked
- Add zh.starwars.wikia.com in addition to zh-hk.starwars.wikia.com as
  an expected target domain as both point to the Chinese Star Wars wiki


saipetch:
LGTM

PR #8988, opened 2015-11-06
SOC-1493 Make sure all boards are shown in Special:Forum

jsutterfield:
Due to a missing primary key in the `page_wikia_props` table in production, we have a bunch of duplicate entires in that table. This is because the helper function used to update that table, `wfSetWikiaPageProp`, is using a `REPLACE` query. When a table lacks a primary key or unique index, a `REPLACE` acts just like a regular `INSERT`, which is leading to the duplicates.

One of those entries which is duplicated is `WPP_WALL_ORDER_INDEX`, which is used by Special:Forum to determine the sort order of boards. These duplicates are leading to multiple boards sharing the same sort index, which leads to some of them not being shown.

This fixes that by doing 2 things:
1.) Updates the logic when getting the boards so that all boards will be shown, even if 2 share the same sort order index.
2.) Update the helper function `wfSetWikiaPageProp` so it behaves as if the table had a primary index. This is accomplished by manually deleting from the table first before inserting.

Ticket: https://wikia-inc.atlassian.net/browse/SOC-1493


includes/wikia/GlobalFunctions.php
```diff
@@ -1104,15 +1113,21 @@ function wfSetWikiaPageProp( $type, $pageID, $value, $dbname = '' ) {
...
+
+	$db->insert(
+		'page_wikia_props',
+		[
 			'page_id'  => $pageID,
 			'propname' => $type,
 			'props'    => wfSerializeProp( $type, $value )
-		),
-		__METHOD__
+		]
```
macbre: Please pass `__METHOD__` as the `$fname` argument in the two above methods.

armonr:
Besides outstanding comment, LGTM. Nice work :)

extensions/wikia/Forum/Forum.class.php
```diff
@@ -30,25 +30,23 @@ public function getBoardList( $db = DB_SLAVE ) {
...
-			$id = $title->getArticleID();
 
 			$boardInfo = $board->getBoardInfo();
 			$boardInfo['id'] = $title->getArticleID();
 			$boardInfo['name'] = $title->getText();
 			$boardInfo['description'] = $board->getDescriptionWithoutTemplates();
 			$boardInfo['url'] = $title->getFullURL();
-			$orderIndex = $orderIndexes[$id];
-			$boards[$orderIndex] = $boardInfo;
+			$boards[] = $boardInfo;
```
garthwebb, line 47: Nice, good way to solve the problem

garthwebb:
LGTM

PR #8990, opened 2015-11-06
Ce 3006 tc radio buttons checking

kamilkoterba:
https://wikia-inc.atlassian.net/browse/CE-3006

@Wikia/community-engineering


lukaszkonieczny:
LGTM

PR #8992, opened 2015-11-09
INT-220: Update Achievement translations in i18n and delete messaging wikia

mari-mcmurtrie:
We use messaging wikia for achievement translation. We want to use i18n instead, and delete messaging wikia. I updated translations in i18n and remove messaging wikia. Ticket: https://wikia-inc.atlassian.net/browse/INT-220


mari-mcmurtrie:
@perjg will you CR? thanks!

mari-mcmurtrie:
I removed Russian translation. @perjg will you check again, thanks!!

perjg:
LGTM

PR #8995, opened 2015-11-09
CE-2187 Adding maintenance script for calculating TempalteClasssifica…

kamilkoterba:
…tion metric
https://wikia-inc.atlassian.net/browse/CE-2187

Table for storage is not created yet so sketch of code for storing in DB is commented for now.
I'd like to have this reviewed to check the metric on production and than add support for storing in DB, as this may be useful for next step which is having chart with metric changes.

Script will be ran daily. Related chef PR https://github.com/Wikia/chef-repo/pull/7278

related config
https://github.com/Wikia/config/pull/1449

/cc @Wikia/community-engineering @SebastianMarzjan

### DIFF
--- a/maintenance/wikia/TemplateClassificationRecognizedMetric.php
+++ b/maintenance/wikia/TemplateClassificationRecognizedMetric.php
@@ -70,8 +70,8 @@ class TemplateClassificationRecognizedMetricMaintenance extends Maintenance {
 
 	private function isRecognized( $type ) {
 		return $type !== TemplateClassificationService::TEMPLATE_UNKNOWN
-			&& $type !== ''
-			&& $type !== 'other';
+			&& $type !== AutomaticTemplateTypes::TEMPLATE_UNCLASSIFIED
+			&& $type !== AutomaticTemplateTypes::TEMPLATE_OTHER;
 	}
 
 	private function getWamTop500( $db, \WikiaSQL $sql ) {
--- a/maintenance/wikia/TemplateClassificationRecognizedMetric.php
+++ b/maintenance/wikia/TemplateClassificationRecognizedMetric.php
@@ -1,0 +1,152 @@
+<?php
+
+require_once( __DIR__ . '/../Maintenance.php' );
+
+/**
+ * Maintenance script counts current number of recognized templates in content namespaces
+ *
+ * @author Kamil Koterba <kamil@wikia-inc.com>
+ * @ingroup Maintenance
+ */
+class TemplateClassificationRecognizedMetricMaintenance extends Maintenance {
+
+	/**
+	 * Set script options
+	 */
+	public function __construct() {
+		parent::__construct();
+
+		$this->mDescription = 'Maintenance script counts current number of recognized templates in content namespaces';
+	}
+
+	/**
+	 * Script entry point
+	 */
+	public function execute() {
+		global $wgContentNamespaces, $wgDWStatsDB;
+
+		$statsDBRead = wfGetDB( DB_SLAVE, [], $wgDWStatsDB );
+		$wamTop500 = $this->getWamTop500( $statsDBRead, ( new \WikiaSQL ) );
+
+		$top500CNRecognizedCount = 0;
+		$top500CNAllCount = 0;
+
+		foreach ( $wamTop500 as $wikia ) {
+			$wikiaDBName = WikiFactory::IDtoDB($wikia->wiki_id);
+			$wikiaDBRead = wfGetDB( DB_SLAVE, [], $wikiaDBName );
+
+			$CNTemplatesOnWikia = $this->getContentNamespacesTemplates( $wikiaDBRead, $wgContentNamespaces );
+			$recognizedTemplates  = $this->getRecognizedTemplatesOnWikia(
+				( new TemplateClassificationService() ),
+				$wikia->wiki_id
+			);
+
+			$CNRecognizedTemplates = $this->intersectSets( $CNTemplatesOnWikia, $recognizedTemplates );
+
+			$top500CNRecognizedCount += count( $CNRecognizedTemplates );
+			$top500CNAllCount += count( $CNTemplatesOnWikia );
+		}
+
+		/*
+		 * Store value in DB
+		 * TODO uncomment and test when storage is ready
+		 */
+//		$statsDBWrite = wfGetDB( DB_MASTER, [], $wgStatsDB ); // Change DB
+//		$timestamp = date( 'Y-m-d' );
+//		$this->setStats( $top500CNRecognizedCount, $top500CNAllCount, $timestamp, $statsDBWrite );
+
+		$this->pushGecko( $top500CNRecognizedCount, $top500CNAllCount );
+	}
+
+	private function getRecognizedTemplatesOnWikia( TemplateClassificationService $tcService, $wikiaId ) {
+		$templates = $tcService->getTemplatesOnWiki( $wikiaId );
+		foreach ( $templates as $pageId => $type ) {
+			if ( !$this->isRecognized( $type ) ) {
+				unset( $templates[$pageId] );
+			}
+		}
+		return $templates;
+	}
+
+	private function isRecognized($type) {
+		return $type !== TemplateClassificationService::TEMPLATE_UNKNOWN
+			&& $type !== ''
+			&& $type !== 'other';
+	}
+
+	private function getWamTop500($db, \WikiaSQL $sql) {
+		return $sql->SELECT( 'wiki_id' )
+			->FROM( 'fact_wam_scores' )
+			->WHERE( 'time_id' )->EQUAL_TO( date( 'Y-m-d', time() - 60 * 60 * 24 ) )
+			->ORDER_BY( 'wam_rank' )
+			->LIMIT( 500 )
+			->run($db);
+	}
+
+	/**
+	 * Remove non content templates from list of recognized templates
+	 */
+	private function intersectSets( $CNTemplatesOnWikia, $recognizedTemplates ) {
+		$CNRecognizedTemplates = [];
+		foreach( $recognizedTemplates as $pageId => $type ) {
+			if ( in_array( $pageId, $CNTemplatesOnWikia ) ) {
+				$CNRecognizedTemplates[] = $pageId;
+			}
+		}
+		return $CNRecognizedTemplates;
+	}
+
+	private function getContentNamespacesTemplates($db, $contentNamespaces) {
+		$sql = ( new \WikiaSQL() )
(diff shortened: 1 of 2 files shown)


maintenance/wikia/TemplateClassificationRecognizedMetric.php
```diff
@@ -0,0 +1,152 @@
...
+			if ( in_array( $pageId, $CNTemplatesOnWikia ) ) {
+				$CNRecognizedTemplates[] = $pageId;
+			}
+		}
+		return $CNRecognizedTemplates;
+	}
+
+	private function getContentNamespacesTemplates($db, $contentNamespaces) {
+		$sql = ( new \WikiaSQL() )
+			->SELECT()->DISTINCT( 'p2.page_id as temp_id' )
```
lgarczewski, line 101: This DISTINCT is somewhat worrying. Can I see an explain for this query please?

lgarczewski:
Overall looks good. I'd like to take a closer look at one of the queries as noted in the in-line comment above.

kamilkoterba:
@SQLReview

SQL QUERY

```
SELECT DISTINCT p2.page_id as temp_id
     FROM page AS p
         INNER JOIN templatelinks AS t
         ON
         t.tl_from =  p.page_id
         INNER JOIN page AS p2
         ON
         p2.page_title =  t.tl_title
     WHERE p.page_namespace = '0'
     AND p2.page_namespace = '10'
     AND p.page_id <> '1461'
```

EXPLAIN

```
+----+-------------+-------+--------+---------------------------------------------------------------------+------------+---------+--------------------------+-------+-------------------------------------------+
| id | select_type | table | type   | possible_keys                                                       | key        | key_len | ref                      | rows  | Extra                                     |
+----+-------------+-------+--------+---------------------------------------------------------------------+------------+---------+--------------------------+-------+-------------------------------------------+
|  1 | SIMPLE      | p     | ref    | PRIMARY,name_title                                                  | name_title | 4       | const                    | 30594 | Using where; Using index; Using temporary |
|  1 | SIMPLE      | t     | ref    | tl_from                                                             | tl_from    | 4       | gtawiki.p.page_id        |     1 | Using index                               |
|  1 | SIMPLE      | p2    | eq_ref | PRIMARY,name_title,page_random,page_len,page_redirect_namespace_len | name_title | 261     | const,gtawiki.t.tl_title |     1 | Using index                               |
+----+-------------+-------+--------+---------------------------------------------------------------------+------------+---------+--------------------------+-------+-------------------------------------------+
```

maintenance/wikia/TemplateClassificationRecognizedMetric.php
```diff
@@ -0,0 +1,152 @@
...
+				unset( $templates[$pageId] );
+			}
+		}
+		return $templates;
+	}
+
+	private function isRecognized($type) {
+		return $type !== TemplateClassificationService::TEMPLATE_UNKNOWN
+			&& $type !== ''
+			&& $type !== 'other';
```
Grunny: Is `'other'` a type likely to be checked for elsewhere? If so, it's probably worth adding a constant for it too.

maintenance/wikia/TemplateClassificationRecognizedMetric.php
```diff
@@ -0,0 +1,152 @@
...
+		global $wgGeckoboardApiKey, $wgGeckoboardPushUrl, $wgTemplateClassificationGeckometerWidgetKey;
+
+		$metric = 0;
+		if( $top500CNAllCount > 0 ) {
+			$metric = $top500CNRecognizedCount / $top500CNAllCount * 100;
+		}
+		$data = [
+				"item" => strval($metric),
+				"min" => ["value" => 0],
+				"max" => ["value" => 100]
```
Grunny: These should use single quotes and be spacey per conventions.

maintenance/wikia/TemplateClassificationRecognizedMetric.php
```diff
@@ -0,0 +1,152 @@
...
+			$wikiaDBName = WikiFactory::IDtoDB($wikia->wiki_id);
+			$wikiaDBRead = wfGetDB( DB_SLAVE, [], $wikiaDBName );
+
+			$CNTemplatesOnWikia = $this->getContentNamespacesTemplates( $wikiaDBRead, $wgContentNamespaces );
+			$recognizedTemplates  = $this->getRecognizedTemplatesOnWikia(
+				( new TemplateClassificationService() ),
+				$wikia->wiki_id
+			);
+
+			$CNRecognizedTemplates = $this->intersectSets( $CNTemplatesOnWikia, $recognizedTemplates );
```
Grunny: Variables should be lowerCamelCase per conventions (here and throughout).

maintenance/wikia/TemplateClassificationRecognizedMetric.php
```diff
@@ -0,0 +1,152 @@
...
+				unset( $templates[$pageId] );
+			}
+		}
+		return $templates;
+	}
+
+	private function isRecognized( $type ) {
+		return $type !== TemplateClassificationService::TEMPLATE_UNKNOWN
+			&& $type !== ''
+			&& $type !== 'other';
```
Grunny: Is 'other' a type that is likely to be checked for elsewhere? If so, it's probably worth adding a constant for it too.

Grunny:
Looks good. The query looks OK for part of a maintenance script run once a day, though it will likely become pretty slow on wikias with a large number of pages such as lyricwiki.

maintenance/wikia/TemplateClassificationRecognizedMetric.php
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	private function getRecognizedTemplatesOnWikia( TemplateClassificationService $tcService, $wikiaId ) {
+		$templates = $tcService->getTemplatesOnWiki( $wikiaId );
+		foreach ( $templates as $pageId => $type ) {
+			if ( !$this->isRecognized( $type ) ) {
+				unset( $templates[$pageId] );
+			}
+		}
+		return $templates;
```
sebastianmarzjan, line 68: could you use something along the lines of 
return array_filter($templates, 'self::isRecognized')
instead of L63-68 ?

maintenance/wikia/TemplateClassificationRecognizedMetric.php
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	private function isRecognized( $type ) {
+		return $type !== TemplateClassificationService::TEMPLATE_UNKNOWN
```
sebastianmarzjan, line 72: Frankly, this would probably read easier given the logic inside if it were 

```
private function isUnrecognized( $type ) {
   return in_array($type, [
        TemplateClassificationService::TEMPLATE_UNKNOWN,
        TemplateClassificationService::TEMPLATE_UNCLASSIFIED,
        TemplateClassificationService::TEMPLATE_OTHER
   ]);
}
```

maintenance/wikia/TemplateClassificationRecognizedMetric.php
```diff
@@ -0,0 +1,152 @@
...
+			&& $type !== AutomaticTemplateTypes::TEMPLATE_UNCLASSIFIED
+			&& $type !== AutomaticTemplateTypes::TEMPLATE_OTHER;
+	}
+
+	private function getWamTop500( $db, \WikiaSQL $sql ) {
+		return $sql->SELECT( 'wiki_id' )
+			->FROM( 'fact_wam_scores' )
+			->WHERE( 'time_id' )->EQUAL_TO( date( 'Y-m-d', time() - 60 * 60 * 24 ) )
```
sebastianmarzjan, line 80: What will happen if the data is not available at the time of script running (that happened a couple of times that WAM calculation was delayed)?

maintenance/wikia/TemplateClassificationRecognizedMetric.php
```diff
@@ -0,0 +1,152 @@
...
+	public function execute() {
+		global $wgContentNamespaces, $wgDWStatsDB;
+
+		$statsDBRead = wfGetDB( DB_SLAVE, [], $wgDWStatsDB );
+		$wamTop500 = $this->getWamTop500( $statsDBRead, ( new \WikiaSQL ) );
+
+		$top500CNRecognizedCount = 0;
+		$top500CNAllCount = 0;
+
+		foreach ( $wamTop500 as $wikia ) {
```
sebastianmarzjan, line 34: It's a design decision so there were some trade-offs considered. I would recommend surfacing them here. I see at least two approaches; The one proposed here is executed in sequence; the second would be moving WAM top 500 outside and running the calculation in context of one wiki would allow for parallel execution.

It may be worthwhile to add a comment about this :)

maintenance/wikia/TemplateClassificationRecognizedMetric.php
```diff
@@ -0,0 +1,152 @@
...
+			->SET( 'total_cn', $totalCNCount )
+			->run( $dbw );
+	}
+
+	private function pushGecko( $top500CNRecognizedCount, $top500CNAllCount ) {
+		global $wgGeckoboardApiKey, $wgGeckoboardPushUrl, $wgTemplateClassificationGeckometerWidgetKey;
+
+		$metric = 0;
+		if ( $top500CNAllCount > 0 ) {
+			$metric = $top500CNRecognizedCount / $top500CNAllCount * 100;
```
sebastianmarzjan, line 131: Are those floats? if not, you may want to add an explicit typecast

maintenance/wikia/TemplateClassificationRecognizedMetric.php
```diff
@@ -0,0 +1,152 @@
...
+			->WHERE( 'time_id' )->EQUAL_TO( date( 'Y-m-d', time() - 60 * 60 * 24 ) )
+			->ORDER_BY( 'wam_rank' )
+			->LIMIT( 500 )
+			->run( $db );
+	}
+
+	/**
+	 * Remove non content templates from list of recognized templates
+	 */
+	private function intersectSets( $cnTemplatesOnWikia, $recognizedTemplates ) {
```
sebastianmarzjan, line 89: could array_intersect be used to perform this logic?

maintenance/wikia/TemplateClassificationRecognizedMetric.php
```diff
@@ -0,0 +1,152 @@
...
+			'min' => [ 'value' => 0 ],
+			'max' => [ 'value' => 100 ]
+		];
+
+		$params = [
+			'api_key' => $wgGeckoboardApiKey,
+			'data' => $data
+		];
+
+		Http::post(
```
sebastianmarzjan, line 144: Please consider adding error handling here

PR #8999, opened 2015-11-09
SOC-1404 As a designer, I want two colors from Theme Designer to be used in new Discussions

rafalkalinski:
For the new discussion product, we're going with a light theme and a dark theme but each of those will also then need to use accent colors from the wiki itself so that there is some consistency.

https://wikia-inc.atlassian.net/browse/SOC-1404


kvas-damian:
LGTM

rafalkalinski:
This code is not needed anymore https://github.com/Wikia/mercury/pull/1625#discussion_r44643919

PR #9002, opened 2015-11-09
ADEN-2532 RubiconRTP cleanup

fraszczakszymon:
Plus small refactor in lookup services module


nandy-andy:
We've checked in previous sprint if Rubicon RTP is disabled and it was AFAIR. So, the red lines should be safe :)

LGTM :+1:

PR #9004, opened 2015-11-09
ADEN-2617 Fix unhandled messages from previous pageviews in messageListener on mercury

fraszczakszymon:
I investigated issue with @jacekjursza and we are going to clear unhandledMessages list every pageview, because of mercury's SPA - the list used to have old messages from previous pageviews (with not existing iframes).

//cc: @gabrys @nandy-andy


nandy-andy:
So, the `ext.wikia.adEngine.messageListener` wasn't created in Mercury times and it was keeping old iframes in the `unhandledMessages` array. By adding `adContext` dependency we can clear the array each time Mercury "opens a new page" and reloads `adContext`.

LGTM :+1:

PR #9005, opened 2015-11-10
ADEN-2607 New Taboola

nandy-andy:
New Taboola widgets implementation:
- below article widget - the same as previously,
- right rail widget.


fraszczakszymon:
LGTM :+1:

PR #9006, opened 2015-11-10
DAT-3454 create a new parser instance

dianafa:
https://wikia-inc.atlassian.net/browse/DAT-3454


sebastianmarzjan:
This looks like it just might work in an elegant way. However, please provide some more context **WHY** this is needed either in the pull request description or (preferrably) in the comment above ParserPool:create() invocation.

sebastianmarzjan:
looks good

PR #9007, opened 2015-11-10
replace READ_LOCKING with READ_LATEST and +10 to lock picking...

ghost:
https://wikia-inc.atlassian.net/browse/MAIN-5810

SELECT ... FOR UPDATE is killing our database

@macbre @drozdo


drozdo:
bueatifull workaround :)

drozdo:
MW also had problems with that:
https://gerrit.wikimedia.org/r/#/c/224647/
https://gerrit.wikimedia.org/r/#/c/224647/1/includes/User.php
https://gerrit.wikimedia.org/r/#/c/21069/1/includes/Revision.php
backporting it sounds like a good idea

PR #9012, opened 2015-11-10
CE-2955 Display TC entry point in a unified way

adamkarminski:


extensions/wikia/TemplateClassification/TemplateClassification.hooks.php
```diff
@@ -175,8 +175,8 @@ public function onEditPageLayoutExecute( \EditPageLayoutController $editPage ) {
 		$user = $editPage->getContext()->getUser();
 		$title = $editPage->getContext()->getTitle();
 		if ( ( new Permissions() )->shouldDisplayEntryPoint( $user, $title ) ) {
-			$editPage->addExtraPageControlsHtml(
-				( new View )->renderEditPageEntryPoint( $wgCityId, $title, $user )
```
kamilkoterba, line 179: I think that was the only use of this method
https://github.com/Wikia/app/blob/CE-2955/extensions/wikia/TemplateClassification/TemplateClassificationView.php#L87-87
so we can remove it.

adamkarminski, line 179: Yup, good catch!

kamilkoterba:
LGTM. Test in various browsers, wikias and resolutions before merge.

PR #9014, opened 2015-11-10
ADEN-2620 Remove floor adhesion container on consecutive pages

fraszczakszymon:
Because of Mercury SPA architecture we have to remove DOM element which contains floor adhesion on every pageview (if ad exists).

### DIFF
--- a/extensions/wikia/AdEngine/js/template/floor.js
+++ b/extensions/wikia/AdEngine/js/template/floor.js
@@ -109,6 +109,13 @@ define('ext.wikia.adEngine.template.floor', [
 		}
 	}
 
+	adContext.addCallback(function () {
+		var floor = doc.getElementById(floorId);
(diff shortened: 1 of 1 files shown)


extensions/wikia/AdEngine/js/template/floor.js
```diff
@@ -109,6 +109,13 @@ define('ext.wikia.adEngine.template.floor', [
 		}
 	}
 
+	adContext.addCallback(function () {
```
nandy-andy, line 112: I wonder if we shouldn't rename the `addCallback` method to `addMercuryCallback` since it's being used only in Mercury... But this is not anything to change at this point (in this PR) -- just a thought for future :japanese_ogre:

nandy-andy:
LGTM :+1:

PR #9015, opened 2015-11-11
(MAIN-5818) Fix for user signup confirmation page

Grunny:
Controller and method names can no longer be set in the request for
special page controllers, so instead passing a separate parameter to
forward to the necessary method.

/cc @owend @jsutterfield


owend:
LGTM, thanks!

jsutterfield:
:+1:

wladekb:
@Grunny good catch, thanks

PR #9016, opened 2015-11-11
INT-108: Added Japanese translations for special pages

mari-mcmurtrie:
https://wikia-inc.atlassian.net/browse/INT-108
We update and also add translations for Japanese special pages.


mari-mcmurtrie:
@perjg will you CR? Thank you.

perjg:
LGTM

PR #9017, opened 2015-11-11
ADEN-2615 Changed image in the message for user with ad blocking software

nandy-andy:
We want to use an image to which we've got rights before releasing the new product in production. Changes below replace the old image with the new one and adjust styling.


fraszczakszymon:
LGTM :octopus:

PR #9018, opened 2015-11-11
ADEN-2623 Slot sizes unification

fraszczakszymon:


nandy-andy:
LGTM :+1: Let's wait with merging and ask Nef if the new size `TOP_LEADERBOARD` is fine.

PR #9019, opened 2015-11-12
SOC-1619 User::newFromSession doesn't really work any more

garthwebb:
The current auth code requires the access token be correct, but newFromSession doesn't set that.  So, this code:

``` php
$u = User::newFromSession();
$u->getId();
```

Will trigger a `User::loadFromSession` which will hit an auth check that will not find the token it wants.  This in turn sets all the log out cookies.

https://wikia-inc.atlassian.net/browse/SOC-1619


kvas-damian:
LGTM

jsutterfield:
LGTM

PR #9020, opened 2015-11-12
ADEN-2602 Renamed recoverable message experiment and group names

fraszczakszymon:
Changed name of the AB test and groups in order to make it more flexible to test with other recoverable products in future.


nandy-andy:
LGTM :+1:

PR #9021, opened 2015-11-12
VE-2041: Fix infobox API to actually return proper response.

inez:
This is still work in progress - weighing pros and cons of two different approaches to this problem.

### DIFF
--- a/extensions/wikia/PortableInfobox/controllers/ApiPortableInfobox.class.php
+++ b/extensions/wikia/PortableInfobox/controllers/ApiPortableInfobox.class.php
@@ -13,12 +13,42 @@ class ApiPortableInfobox extends ApiBase {
 		if ( $arguments === null ) {
 			$this->getResult()->setWarning( "Arguments json format incorect or empty" );
 		}
-		$parser = new Parser();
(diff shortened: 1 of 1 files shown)


inez:
@pauloslund I think this could be reviewed and merged now.

extensions/wikia/PortableInfobox/controllers/ApiPortableInfobox.class.php
```diff
@@ -13,12 +13,42 @@ public function execute() {
 		if ( $arguments === null ) {
 			$this->getResult()->setWarning( "Arguments json format incorect or empty" );
 		}
-		$parser = new Parser();
-		$parser->startExternalParse( Title::newFromText( $title ), new ParserOptions(), 'text', true );
-		$frame = $parser->getPreprocessor()->newCustomFrame( $arguments );
+
+		/*
+		$fakeTemplate = 'FakeTemplate' . wfRandomString( 4 );
```
pauloslund, line 24: I think I would put a note as to why this block of commented code lives here. That way if someone else happens to come across it, they don't just remove it. Other than that though it looks good to me.

PR #9024, opened 2015-11-12
CE-2926: Prepare a way to add js import without review process

lukaszkonieczny:
PLEASE MERGE WITH: https://github.com/Wikia/config/pull/1463

Our trusted user are now able to import reviewed scripts to their wiki without review process.
For this purpose we used MediaWiki:ImportJS page.

All imported scripts should be separate by new line and contain just article name for local imports and for external imports also "dev:" prefix.


skins/oasis/modules/ContentReviewModuleController.class.php
```diff
@@ -64,5 +64,7 @@ public function executeRender( $params ) {
 		$this->setVal( 'enableTestMode', wfMessage( 'content-review-module-test-mode-enable' )->plain() );
 		$this->setVal( 'title', wfMessage( 'content-review-module-title' )->plain() );
 		$this->setVal( 'help', wfMessage( 'content-review-module-help' )->parse() );
+		$this->setVal( 'jsPages', SpecialPage::getTitleFor( 'JSPages' )->getLocalURL() );
```
adamkarminski: Could it be `jsPagesUrl`?

extensions/wikia/ContentReview/ContentReview.hooks.php
```diff
@@ -23,6 +23,28 @@ public static function register() {
...
+	 *
+	 * @param \Article $article
+	 * @param String $content
+	 * @return bool
+	 */
+	public function onArticleAfterFetchContent( \Article $article, &$content ) {
+		$title = $article->getTitle();
+		$isViewPage = empty( \RequestContext::getMain()->getRequest()->getVal( 'action' ) );
+
+		if ( \Wikia\ContentReview\ImportJS::isImportJSPage( $title ) && $isViewPage ) {
```
adamkarminski: The `Hooks` class is in the same namespace. :)

extensions/wikia/ContentReview/ContentReviewImportJS.php
```diff
@@ -0,0 +1,123 @@
...
+	 */
+	public function getImportScripts() {
+		$importScript = \WikiaDataAccess::cache(
+			$this->getImportJSMemcKey(),
+			2592000, // 30 days,
+			function() {
+				$importScript = '';
+				$title = \Title::newFromText( self::IMPORT_SCRIPTS, NS_MEDIAWIKI );
+
+				if ( $title instanceof \Title && $title->getArticleID() != 0 ) {
```
Grunny: I'd suggest changing `$title->getArticleID() != 0` to `$title->exists()`. It's the exact same check, but `exists` has more meaning to someone reading the code about what you're actually wanting to check.

extensions/wikia/ContentReview/ContentReviewImportJS.php
```diff
@@ -0,0 +1,123 @@
...
+	const IMPORT_SCRIPTS_KEY = 'content-review-importjs';
+	const IMPORT_SCRIPTS_VERSION = '1.0';
+
+	/**
+	 * Get js script to load imports
+	 *
+	 * @return String
+	 */
+	public function getImportScripts() {
+		$importScript = \WikiaDataAccess::cache(
```
adamkarminski: Minor thing, but I think that if the function is called `getImportScripts`, then the variable should be called `$importScripts` to match. :)

extensions/wikia/ContentReview/ContentReviewImportJS.php
```diff
@@ -0,0 +1,123 @@
...
+	 * @param array $scripts
+	 * @return array
+	 */
+	public function prepareImports( Array $scripts ) {
+		$imports = [];
+
+		foreach( $scripts as $key => $script ) {
+			$script = trim( $script );
+
+			if ( strtoupper( substr( $script, -3 ) ) === '.JS' ) {
```
adamkarminski: You can use `ContentReviewHelper::JS_FILE_EXTENSION`. Out of curiosity - why `strtoupper` and not `strtolower`? :D

extensions/wikia/ContentReview/ContentReviewImportJS.php
```diff
@@ -0,0 +1,123 @@
...
+	 * Create inline script with function to load imports
+	 *
+	 * @param array $imports
+	 * @return string
+	 */
+	public function createInlineScript( Array $imports ) {
+		$importScript = '';
+
+		if ( !empty( $imports ) ) {
+			$code = '(function(){'
```
adamkarminski: For better readability you could use `sprintf` i.e.
`sprintf( '(function(){$1(["$2"])})', self::IMPORT_SCRIPTS_FUNCTION, implode( ',', $imports ) );`

extensions/wikia/ContentReview/ContentReviewImportJS.php
```diff
@@ -0,0 +1,123 @@
...
+				. self::IMPORT_SCRIPTS_FUNCTION . '(["' . implode( '", "', $imports ) . '"]);'
```
Grunny: What if the script name contains a `"`?

extensions/wikia/ContentReview/ContentReviewImportJS.php
```diff
@@ -0,0 +1,123 @@
+<?php
+
+namespace Wikia\ContentReview;
+
+class ImportJS {
+	const IMPORT_SCRIPTS = 'ImportJS';
```
adamkarminski: I'd consider renaming this const, since what it ultimately is is a title of a page. :)

extensions/wikia/ContentReview/ContentReviewImportJS.php
lukaszkonieczny: There is no specific reason :) I think I saw it in `ProtectSiteJS` class ;)

Grunny: In this case, you probably don't want to change the case when checking if it's a valid JS page. For importing JS pages, they should only end in `.js` lowercase. Pages ending in `.Js` or `.JS` aren't consider valid JS pages by MediaWiki, so `substr( $script, -3 ) === ContentReviewHelper::JS_FILE_EXTENSION` should be enough.

extensions/wikia/ContentReview/ContentReviewImportJS.php
```diff
@@ -0,0 +1,128 @@
...
+	 * @return string
+	 */
+	public function createInlineScript( Array $imports ) {
+		$importScript = '';
+
+		if ( !empty( $imports ) ) {
+			$code = sprintf(
+				'(function(){%s(["%s"]);})();',
+				self::IMPORT_SCRIPTS_FUNCTION,
+				implode( '", "', $imports )
```
Grunny, line 87: The `escapeJsString` should be around each of the imports here, not the overall script content below.

adamkarminski:
Apart from the one comment left it LGTM, but let's wait for a one from Grunny as well. :)

extensions/wikia/ContentReview/ContentReviewImportJS.php
```diff
@@ -0,0 +1,130 @@
...
+	 * @param array $scripts
+	 * @return array
+	 */
+	public function prepareImports( Array $scripts ) {
+		$imports = [];
+
+		foreach ( $scripts as $key => $script ) {
+			$script = trim( $script );
+
+			if ( strtolower( substr( $script, -3 ) ) === Helper::JS_FILE_EXTENSION
```
Grunny: This should just be `substr( $script, -3 ) === Helper::JS_FILE_EXTENSION`. Changing to `strtolower` when checking allows page names like "MediaWiki:Foo.Js" or "MediaWiki:Foo.JS" through which are not valid JS pages. Only MW pages ending in lowercase `.js` are valid, so no need to transform the input.

Grunny:
Last comment otherwise LGTM.

adamkarminski:
Tested.

PR #9026, opened 2015-11-12
DAT-3493 register query page only if extension loaded

idradm:
https://wikia-inc.atlassian.net/browse/DAT-3493

@adamkarminski take a look


sebastianmarzjan:
LGTM

adamkarminski:
Awww yes! We've parallelly implemented the same change. LGTM!

PR #9027, opened 2015-11-12
SOC-1434 | Unconfirmation banner can appear stacked many times

rafalkalinski:
https://wikia-inc.atlassian.net/browse/SOC-1434
Unconfirmation banner can appear stacked many times

And right now, the BannerNotificationsController::addConfirmation checks if the message was already set.

### DIFF
--- a/extensions/wikia/BannerNotifications/BannerNotificationsController.class.php
+++ b/extensions/wikia/BannerNotifications/BannerNotificationsController.class.php
@@ -38,6 +38,14 @@ class BannerNotificationsController extends WikiaController {
 				$_SESSION[self::SESSION_KEY] = [];
 			}
 
+			// Making sure if the message is not already set in the session.
+			// We're checking only the message text and not other message parameters,
+			// because the text is the most important here.
+			if ( array_search( $message, array_column( $_SESSION[self::SESSION_KEY], 'message' ) ) !== false ) {
+				wfDebug( __METHOD__ . " - {$message}\n - already in the _SESSION array" );
+				return;
+			}
+
 			$_SESSION[self::SESSION_KEY][] = [
 				'message' => $message,
 				'type' => $type,
--- a/extensions/wikia/BannerNotifications/BannerNotificationsController.class.php
+++ b/extensions/wikia/BannerNotifications/BannerNotificationsController.class.php
@@ -41,8 +41,9 @@ class BannerNotificationsController extends WikiaController {
 			// Making sure if the message is not already set in the session.
 			// We're checking only the message text and not other message parameters,
 			// because the text is the most important here.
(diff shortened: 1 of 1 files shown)


kvas-damian:
LGTM

extensions/wikia/BannerNotifications/BannerNotificationsController.class.php
```diff
@@ -38,6 +38,14 @@ public static function addConfirmation( $message, $type = self::CONFIRMATION_CON
 				$_SESSION[self::SESSION_KEY] = [];
 			}
 
+			// Making sure if the message is not already set in the session.
+			// We're checking only the message text and not other message parameters,
+			// because the text is the most important here.
+			if ( array_search( $message, array_column( $_SESSION[self::SESSION_KEY], 'message' ) ) !== false ) {
```
garthwebb: Maybe `in_array` would be better here than `array_search`; the logic would be more direct and we don't need the extra functionality `array_search` provides (e.g. returning the index into the array where it exists).

extensions/wikia/BannerNotifications/BannerNotificationsController.class.php
```diff
@@ -38,6 +38,14 @@ public static function addConfirmation( $message, $type = self::CONFIRMATION_CON
...
+				wfDebug( __METHOD__ . " - {$message}\n - already in the _SESSION array" );
```
garthwebb: We should use `\Wikia\Logger\WikiaLogger::instance()->debug('...');` for debugging/logging rather then `wfDebug`

extensions/wikia/BannerNotifications/BannerNotificationsController.class.php
```diff
@@ -38,13 +38,22 @@ public static function addConfirmation( $message, $type = self::CONFIRMATION_CON
 				$_SESSION[self::SESSION_KEY] = [];
 			}
 
+			// Making sure if the message is not already set in the session.
+			// We're checking only the message text and not other message parameters,
+			// because the text is the most important here.
+			if ( in_array( $message, array_column( $_SESSION[self::SESSION_KEY], 'message' ) ) ) {
+				\Wikia\Logger\WikiaLogger::instance()
+					->debug( __METHOD__ . " - {$message}\n - already in the _SESSION array" );
```
garthwebb, line 46: @rafalkalinski , so there's no need to update this PR necessarily but FYI for future use; the `WikiaLogger` logging methods have a signature like:

``` php
debug( $message, $params )
```

If you want to include interesting context around a debug call, its best to put it in the $param section so it can be queried easily from Kibana, e.g.:

``` php
\Wikia\Logger\WikiaLogger::instance()->debug( "Message already in _SESSION array", [
    'method' => __METHOD__,
    'messageText' => $message,
    'issue' => 'SOC-1434', // I like adding this for filtering and context on the kibana side
] );
```

LGTM though.

PR #9035, opened 2015-11-13
(CE-2942) Enable word wrap in the Ace editor by default

Grunny:
Enable word wrap in the ace editor by default by setting the default
option.


lukaszkonieczny:
LGTM

PR #9036, opened 2015-11-13
DatabaseBase: improve SHOW PROCESSLIST debugging

macbre:
- show client IP address if the `$wgUser` object is not ready yet (use `$wgRequest`)
- show maintenance script class name when running in CLI mode (instead of `127.0.0.1`)

Compare:

``` sql
Query wikicities (8) (slave): SELECT /* AvatarsMigrator::execute AvatarsMigrator */  user_id AS id  FROM `user`   ORDER BY user_id 
```

vs

``` sql
Query wikicities (8) (slave): SELECT /* AvatarsMigrator::execute 127.0.0.1 */  user_id AS id  FROM `user`   ORDER BY user_id 
```

This should help us when debugging DB performance issues using `SHOW PROCESSLIST`

@drozdo / @wladekb / @michalroszka

### DIFF
--- a/includes/db/Database.php
+++ b/includes/db/Database.php
@@ -920,6 +920,18 @@ abstract class DatabaseBase implements DatabaseType {
 			$userName = str_replace( '/', '', $userName );
 		} else {
 			$userName = '';
+
+			# Wikia change - if the user object is not ready yet, show the client IP (see PLATFORM-1671 for examples)
+			global $wgRequest;
+			if ( $wgRequest instanceof WebRequest ) {
+				$userName = $wgRequest->getIP();
+			}
+		}
+
+		# Wikia change - log the name of the maintenance class that run this query (instead of 127.0.0.1)
+		global $maintClass;
+		if ( !empty( $maintClass ) ) {
+			$userName = str_replace( '/', '', $maintClass );
 		}
 
 		# Wikia change - begin


includes/db/Database.php
```diff
@@ -920,6 +920,18 @@ public function query( $sql, $fname = '', $tempIgnore = false ) {
 			$userName = str_replace( '/', '', $userName );
 		} else {
 			$userName = '';
+
+			# Wikia change - if the user object is not ready yet, show the client IP (see PLATFORM-1671 for examples)
+			global $wgRequest;
+			if ( $wgRequest instanceof WebRequest ) {
+				$userName = $wgRequest->getIP();
```
pawelchcki, line 927: Will this return proper IP e.g. from  X-Fastly-Client-IP ?
I'm thinking it would be best to always return intended IP so If user action causes this code to run then we should log UserIP.
Conversely if you run this code from maintenance script or tasks.wikia.com then it would be nice to log it too.

pawelchcki:
One comment otherwise LGTM :+1:

includes/db/Database.php
macbre, line 927: 1. `$wgRequest->getIP()` uses Fastly's secret header with the IP address.
2. The source host is always displayed in `SHOW PROCESSLIS` - so instead of duplicating the cron machine IP let's show the maintenance script class name.

includes/db/Database.php
```diff
@@ -920,6 +920,18 @@ public function query( $sql, $fname = '', $tempIgnore = false ) {
...
+			}
+		}
+
+		# Wikia change - log the name of the maintenance class that run this query (instead of 127.0.0.1)
+		global $maintClass;
+		if ( !empty( $maintClass ) ) {
+			$userName = str_replace( '/', '', $maintClass );
```
wladekb, line 934: how about '_' instead of '' ?

PR #9042, opened 2015-11-16
PLATFORM-1677: improve throttling and logging in EventsCleanup script

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-1677

`eventsCleanup.php` set databases on fire during its recent run. Let's perform smaller DELETE queries by making a batch a bit smaller (10 instead of 50 wikis) and wait for slaves to catch up after each delete + additional sleep for 5 seconds.

@wladekb / @michalroszka / @drozdo / @ljagiello


ljagiello:
Looks good, will see results next weekend :)

PR #9043, opened 2015-11-16
SEO-151 One <h1> per page

gabrys:
Following experiment conducted in SEO-3 and after confirming there's no negative impact of having just one `<h1>` tag on each page (for the article title, usually matching the `<title>` tag), we're making the changes persistent, converting all the additional `<h1>` tags to `<h2>`.


inez:
Lgtm

PR #9044, opened 2015-11-17
Replace deprecated validateJSDoc rule with jsDoc rule

drets:
Replace deprecated validateJSDoc rule with jsDoc rule: http://jscs.info/rule/jsDoc.html


lizlux:
:+1:

PR #9046, opened 2015-11-17
DataProvider::GetTopFiveUsers - disable in Reston

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-1648

Average query time (selects on `events_local_users` table) is at ~8 seconds in Reston.

Additionally log slow queries (that took more than 15 seconds) without sampling - the average query time form sampled log did not provide much value here

@wladekb / @michalroszka


ghost:
:+1:

PR #9049, opened 2015-11-18
(CE-3037) Hide templates classified as notice in Mercury

Grunny:
In addition to navboxes, also hide templates classified as notice in
Mercury.

/cc @Wikia/community-engineering @SebastianMarzjan

### DIFF
--- a/includes/wikia/parser/templatetypes/TemplateTypesParser.class.php
+++ b/includes/wikia/parser/templatetypes/TemplateTypesParser.class.php
@@ -22,6 +22,9 @@ class TemplateTypesParser {
 				case TemplateClassificationService::TEMPLATE_NAVBOX:
 					$text = self::handleNavboxTemplate();
 					break;
+				case TemplateClassificationService::TEMPLATE_FLAG:
+					$text = self::handleNoticeTemplate();
+					break;
 				case AutomaticTemplateTypes::TEMPLATE_REFERENCES:
 				case TemplateClassificationService::TEMPLATE_REFERENCES:
 					$text = self::handleReferencesTemplate();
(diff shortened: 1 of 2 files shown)


includes/wikia/parser/templatetypes/TemplateTypesParser.class.php
```diff
@@ -20,7 +20,9 @@ public static function onFetchTemplateAndTitle( &$text, &$finalTitle ) {
 			switch ( $type ) {
 				case AutomaticTemplateTypes::TEMPLATE_NAVBOX:
 				case TemplateClassificationService::TEMPLATE_NAVBOX:
-					$text = self::handleNavboxTemplate();
+				case TemplateClassificationService::TEMPLATE_NAV:
+				case TemplateClassificationService::TEMPLATE_FLAG:
+					$text = self::handleHiddenTemplate();
```
sebastianmarzjan: I am thinking about actually splitting on template purpose (navbox, general navigation, flag) instead of treatment. What do you think?

Grunny: Yeah, I originally did that and went back and forward a few times thinking on what is more useful: having types with shared behaviour/treatment or having handling per template purpose. I finally came down on treatment, but I was very conflicted. ;) I'll go back to separate per purpose as it does provide a little more flexibility going forward.

Grunny: @SebastianMarzjan done.

Grunny:
Navigation templates removed from the PR for now as we've decided to start with notices only.

sebastianmarzjan:
LGTM

PR #9058, opened 2015-11-18
TransactionClassifier: a separate transaction for Wall and Forum

macbre:
Wall and Forum requests were ending up in `page/other` transaction (which is the most frequent one). Improve the classification to have more more granular data in NewRelic.

@wladekb


wladekb:
lgtm

PR #9040, opened 2015-11-16
DAT-3463 handle context-links on Mercury

dianafa:
https://wikia-inc.atlassian.net/browse/DAT-3463

PR #9059, opened 2015-11-18
Merge pull request #9040 from Wikia/DAT-3463

dianafa:
DAT-3463 handle context-links on Mercury as we want to go love with tomorrow's release.


dianafa:
was LGTM'ed before :)

PR #9063, opened 2015-11-18
CE-3054 Create a maintenance script that classifies nested templates

adamkarminski:
This script is based on assumption, that if a content NS page includes a template a `{{$templateTitle` expression is there. For better accuracy both the raw page text and template title are converted to lowercase.


extensions/wikia/TemplateClassification/maintenance/classifyNonArticleTemplates.class.php
```diff
@@ -0,0 +1,128 @@
...
+		$this->logFile = $this->getOption( 'log-file' );
+
+		if ( isset( $this->logFile ) && !$this->verifyLogFile() ) {
+			$this->out( "The log file {$this->logFile} does not exist, please create it." );
+			return false;
+		}
+
+		$this->setDB( wfGetDB( DB_SLAVE ) );
+		$tcs = new TemplateClassificationService();
+		$firstLevelTemplates = $nestedTemplates = [];
```
lukaszkonieczny: `$nestedTemplates` var is not used.

extensions/wikia/TemplateClassification/maintenance/classifyNonArticleTemplates.class.php
```diff
@@ -0,0 +1,128 @@
...
+			$count = count( $linkingPages );
+
+			$this->out( "\nProcessing {$count} inclusions of template {$templateTitle} ({$templateId})" );
+			$isFirstLevel = false;
+
+			foreach ( $linkingPages as $pageId ) {
+				$pageRawTextLc = strtolower( WikiPage::newFromID( $pageId )->getRawText() );
+				$templateTitleLc = strtolower( $templateTitle );
+				if ( strpos( $pageRawTextLc, "{{{$templateTitleLc}" ) > -1 ) {
+					$firstLevelTemplates[] = $templateId;
```
lukaszkonieczny: As we are not using template id anywhere we can use just variable to count first level templates

extensions/wikia/TemplateClassification/maintenance/classifyNonArticleTemplates.class.php
```diff
@@ -0,0 +1,127 @@
...
+		foreach ( $namespacesTemplates as $templateId => $data ) {
+			$templateTitle = $data['title'];
+			$linkingPages = $data['linkingPages'];
+			$count = count( $linkingPages );
+
+			$this->out( "\nProcessing {$count} inclusions of template {$templateTitle} ({$templateId})" );
+			$isFirstLevel = false;
+
+			foreach ( $linkingPages as $pageId ) {
+				$pageRawTextLc = strtolower( WikiPage::newFromID( $pageId )->getRawText() );
```
Grunny: It's rare, but we have had stale data with no longer existent pages stuck in templatelinks, categorylinks, and similar tables before. Just as a sanity check, I'd consider making sure `WikiPage::newFromID( $pageId )` returns a `WikiPage` object as if the page ID can't be found, it will return null and this will fatal.

adamkarminski: Hmm, I was considering this but when digging inside WikiPage's code I saw no place returning `null`. Although, considering our experience, I'll go with your suggestion. :)

extensions/wikia/TemplateClassification/maintenance/classifyNonArticleTemplates.class.php
```diff
@@ -0,0 +1,127 @@
...
+
+	private function out( $s ) {
+		if ( !$this->quiet ) {
+			$this->output( "$s\n" );
+		}
+	}
+
+	private function logResults( array $results ) {
+		if ( isset( $this->logFile ) ) {
+			$data = implode( ',', array_values( $results ) );
```
Grunny: `array_values` is unneeded here. `implode( ',', array_values( $results ) ) === implode( ',', $results )`

adamkarminski: Assumptions FTW. :)

extensions/wikia/TemplateClassification/maintenance/classifyNonArticleTemplates.class.php
Grunny: Cool. For reference, `WikiPage::newFromID` returns null if the page ID isn't in the page table: https://github.com/Wikia/app/blob/dev/includes/WikiPage.php#L136

lukaszkonieczny:
LGTM

adamkarminski:
@kamilkoterba: Apart from the SQL query fail, LGTM :)

kamilkoterba:
sql fixed. pushed force.

PR #9064, opened 2015-11-18
ADEN-2661 Extending functionality of InstantGlobals URL override

jacekjursza:
Currently there is possibility to override instantGlobals via URL, but only for int values. I've added the possiblity of overriding arrays. 
The usage is:

```
?InstantGlobals.variableName=[VAL1,VAL2,VAL3]
(note: values are without quotes or single quotes)
```


fraszczakszymon:
LGTM :+1:

PR #9068, opened 2015-11-19
TransactionTraceNewrelic: add custom tracer for Rabbit MQ PHP client

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-1696

RabbitMQ timeouts causes page submits to take over 30 seconds. Let's try to use `newrelic_add_custom_tracer` to have some insights on what's going on - performance data for these calls should end up in NewRelic's transaction traces.

And report request ID as a custom parameter.

@wladekb / @SebastianMarzjan

### DIFF
--- a/includes/wikia/transaction/TransactionTraceNewrelic.php
+++ b/includes/wikia/transaction/TransactionTraceNewrelic.php
@@ -1,5 +1,7 @@
 <?php
 
+use Wikia\Util\RequestId;
+
 /**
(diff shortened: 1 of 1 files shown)


wladekb:
didn't consult the newrelic docs, otherwise lgtm

macbre:
I gave it a try in eval.php on ap-s10 (where NR agent is active) and it did not throw a fatal :)

macbre:
Merged into production branch in af80174

PR #9069, opened 2015-11-19
Ce tc links blank

kamilkoterba:
https://wikia-inc.atlassian.net/browse/CE-2981

Since it on modal there's a requirement to open in new tab.

@Wikia/community-engineering


kamilkoterba:
LGT @Grunny

PR #9070, opened 2015-11-19
PLATFORM-1694: improve monitoring of internal service calls in NewRelic

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-1694

Let's use a similar approach as in #9068 and instrument PHP calls to:
- Helios
- Phalanx
- Preferences service
- SMTP traffic when sending emails

@wladekb / @gbenson-wikia


wladekb:
lgtm

PR #9065, opened 2015-11-19
(CE-3062) Add missing messages for the context link type

Grunny:
Add missing messages for the context link type and reset the order of template
types in the dialog.

/cc @Wikia/community-engineering

PR #9073, opened 2015-11-19
revert of 468dcd1015560f1012b9ed9bd8900c1bb6ce8f40

dianafa:
Because of performance issues on Mercury the https://github.com/Wikia/app/pull/9040 has to be reverted.

### DIFF
--- a/includes/wikia/services/TemplateClassificationService.class.php
+++ b/includes/wikia/services/TemplateClassificationService.class.php
@@ -14,7 +14,6 @@ class TemplateClassificationService {
 
 	// TODO: Move types used for manual classification outside of Template Classification Service
 	// TODO: https://wikia-inc.atlassian.net/browse/CE-3017
-	const TEMPLATE_CONTEXT_LINK = 'context-link';
 	const TEMPLATE_CUSTOM_INFOBOX = 'custom-infobox';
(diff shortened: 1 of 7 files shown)


includes/wikia/services/TemplateClassificationService.class.php
```diff
@@ -40,17 +39,16 @@ class TemplateClassificationService {
...
 		self::TEMPLATE_DESIGN,
+		self::TEMPLATE_FLAG,
+		self::TEMPLATE_INFOBOX,
+		self::TEMPLATE_MEDIA,
 		self::TEMPLATE_NAV,
+		self::TEMPLATE_NAVBOX,
 		self::TEMPLATE_NOT_ART,
+		self::TEMPLATE_QUOTE,
+		self::TEMPLATE_REFERENCES,
 		self::TEMPLATE_UNKNOWN,
```
Grunny, line 52: The order of this array is important right now, as it affects the order in the dialog (not ideal, but the way it is right now). So, the previous order with infobox first is correct, see https://github.com/Wikia/app/pull/9065

dianafa, line 52: wow, ok, thank you!

sebastianmarzjan:
lgtm

PR #9074, opened 2015-11-19
add context-link to manual template classification dialog

dianafa:
add context-link to manual template classification dialog


Grunny:
:+1:

PR #9075, opened 2015-11-19
CE-3017 Migrate the user-related part of TCS to a User child class

adamkarminski:


Grunny:
LGTM.

PR #9078, opened 2015-11-19
VE-2066 | End-end experience for clicking yes/no in the EBS

katanka:


extensions/wikia/EmergencyBroadcastSystem/EmergencyBroadcastSystemController.class.php
```diff
@@ -3,4 +3,40 @@ class EmergencyBroadcastSystemController extends WikiaController {
...
 		$this->response->setVal( 'nonPortableCount', '3' ); // Temporary number for testing
 	}
+
+	public static function saveUserResponse( $val ) {
+		global $wgUser;
+		if ($val === null) {
+			// invalid call, do nothing
+		} elseif ($val === 0) { // no
+			// if user clicks no, set ebs_response to 0
+			$wgUser->setOption( 'ebs_response', 0 );
```
inez: Maybe create a static with "ebs_response" - to avoid repeating same string few times

extensions/wikia/EmergencyBroadcastSystem/EmergencyBroadcastSystemController.class.php
```diff
@@ -3,4 +3,40 @@ class EmergencyBroadcastSystemController extends WikiaController {
 	public function index( ) {
 		$this->response->setVal( 'nonPortableCount', '3' ); // Temporary number for testing
 	}
+
+	public static function saveUserResponse( $val ) {
+		global $wgUser;
+		if ($val === null) {
+			// invalid call, do nothing
```
inez, line 11: Log the error

extensions/wikia/EmergencyBroadcastSystem/js/EmergencyBroadcastSystem.js
```diff
@@ -0,0 +1,24 @@
+(function ($) {
+	'use strict';
+
+	function init() {
+		$('.ebs-primary-action').click(onClickYes);
+		$('.ebs-secondary-action').click(onClickNo);
+	}
+
+	function onClickYes() {
```
inez: Those two methods: onClickYes and onClickNo are almost identical - I don't thing they are going to change so it might make sense to unify them to one method.

PR #9080, opened 2015-11-19
VE-2061: Add basic rules to determine when to show EBS

pauloslund:


extensions/wikia/EmergencyBroadcastSystem/EmergencyBroadcastSystemController.class.php
```diff
@@ -19,8 +24,21 @@ public static function saveUserResponse( $val ) {
 			$wgUser->setOption( $option_name, $timestamp );
 		}
 	}
+	// PROTECTED
+
+	protected function isCorrectPage() {
+		$title = $this->getContext()->getTitle();
+		$specialPageName = Transaction::getAttribute(Transaction::PARAM_SPECIAL_PAGE_NAME);
```
inez: Would make sense to check of it's "any" special page first? Title::isSpecial

extensions/wikia/EmergencyBroadcastSystem/EmergencyBroadcastSystemController.class.php
```diff
@@ -19,8 +24,21 @@ public static function saveUserResponse( $val ) {
...
+
+		return $title->isContentPage() || $specialPageName === 'WikiActivity' || $specialPageName === 'Recentchanges';
+	}
+
+	protected function userIsPowerUser() {
```
inez: Two times user in the name of the method - what for?

extensions/wikia/EmergencyBroadcastSystem/EmergencyBroadcastSystemController.class.php
```diff
@@ -19,8 +24,21 @@ public static function saveUserResponse( $val ) {
...
+		$user = $this->getContext()->getUser();
+		return $user->isPowerUser();
+	}
 
-	public static function canOpenEBS( ) {
+	protected function canOpenEBS( ) {
```
inez, line 41: I always have problem with this... If something could be static why not to keep it as such - could be used in more places then - specifically I will need it to decide about adding js.

inez:
We need to check for wiki language too

pauloslund:
Can language check be done in another PR?

PR #9081, opened 2015-11-20
Forum: make permissions applied even when the extension is disabled

macbre:
See [PLATFORM-1664](https://wikia-inc.atlassian.net/browse/PLATFORM-1664)

Introduce a separate file with permissions definition and include it in both setup and `ForumDisabled.setup.php` which is loaded when the extension is, well.. disabled.

@jcellary / @michalroszka


jcellary:
Fantastico!

PR #9087, opened 2015-11-20
Special:Templates: handling a case where there are no classified templates

lgarczewski:
@lukaszkonieczny


lukaszkonieczny:
LGTM

PR #9089, opened 2015-11-20
Piggyback Fix

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-1028

When logging in with username/password, check for a `403` response from Helios as an indicator that the username/password were correct, but the login was blocked for other reasons, and allow that login attempt to succeed.

Additionally, add the piggybacked users access token to the response so the switch from the piggyback-er to the piggyback-ee and back happens properly.

See Wikia/helios#145 for associated Helios change.


includes/specials/SpecialUserlogin.php
```diff
@@ -738,8 +738,8 @@ public function authenticateUserData() {
 			$retval = self::SUCCESS;
 		}
 
-		if ( in_array( $retval, [ self::SUCCESS, self::RESET_PASS ] ) ) {
-			wfRunHooks( 'LoginSuccessModifyRetval', [ $u->getName(), $this->mPassword, &$retval ] );
+		if ( !in_array( $retval, [ self::SUCCESS, self::RESET_PASS ] ) ) {
```
nmonterroso, line 741: I repurposed this hook because it seems like it was a noop since its original intention was overwritten above with 

``` php
} elseif ( $wgEnableHeliosExt && Wikia\Helios\User::wasResetPassAuth( $this->mUsername, $this->mPassword ) ) {
```

to set the correct status of the function. @garthwebb do you remember if there was there a reason this hook invocation was left in place after adding the above `elseif`? It was added in 2da99740a249d8dbe89be6626a30c248f8b19ee4

extensions/wikia/Helios/User.class.php
```diff
@@ -218,6 +221,14 @@ public static function wasResetPassAuth( $username, $password ) {
 		return $authMethod == self::AUTH_TYPE_RESET_PW;
 	}
 
+	public static function checkAuthenticationStatus( $username, $password, $status ) {
+		if ( empty( self::$authenticationCache[$username][$password][self::STATUS_NAME] ) ) {
+			return false;
+		}
+
+		return self::$authenticationCache[$username][$password][self::STATUS_NAME] == $status;
```
drsnyder, line 227: Do you know why we are caching these (and keying by the username and password)?

nmonterroso, line 227: I'm not sure exactly but from my debugging sessions it seems that [`authenticate`](https://github.com/Wikia/app/blob/dev/extensions/wikia/Helios/User.class.php#L141) is called multiple times during a request, so the cache is to prevent mediawiki from making multiple helios calls. This is pretty hacky, but for now I think it's necessary because we haven't fully unwound the authentication flow.

extensions/wikia/Helios/User.class.php
```diff
@@ -190,13 +192,14 @@ public static function authenticate( $username, $password ) {
...
 			self::AUTH_METHOD_NAME => $authMethod,
+			self::STATUS_NAME => $status,
 		];
 
 		if ( $throwException ) {
 			throw $throwException;
 		}
 
-		if ( !empty( $loginInfo ) && $authMethod != self::AUTH_TYPE_RESET_PW ) {
+		if ( isset( $loginInfo->access_token ) && $authMethod != self::AUTH_TYPE_RESET_PW ) {
```
nmonterroso, line 200: This seemed like a bug since `loginInfo` contains error information when helios returns non-200, but doesn't have the `access_token` field.

extensions/wikia/Helios/User.class.php
drsnyder, line 227: SKETCHY!

nmonterroso, line 227: sometimes you gotta get down and dirty

lib/Wikia/src/Service/Constants.php
```diff
@@ -9,4 +9,10 @@ class Constants {
 	const HELIOS_AUTH_HEADER = "X-Wikia-UserId";
 
 	const HTTP_STATUS_NO_CONTENT = 204;
+
+	const HTTP_STATUS_SERVER_ERROR = 500;
```
sqreek: Maybe use one place to handle consts? 
WikiaResponse has some.
Anyway some order here will be nice.

lib/Wikia/src/Service/Helios/HeliosClientImpl.php
```diff
@@ -149,7 +149,7 @@ public function login( $username, $password )
 			[ 'method'	=> 'POST' ]
 		);
 
-		return $response;
+		return [$this->status, $response];
```
sqreek, line 152: why status cannot be provide by response?

nmonterroso, line 152: The way it's coded the `$response` is the json decoded response object, and there's no way to get at the integer status. I agree that we should have some sort of `HttpResponse` object that contains everything we need, but I didn't want to make such a major change to the helios implementation usage in MW to fix piggybacking.

lib/Wikia/src/Service/Constants.php
nmonterroso: Cool, I figured there must have been some existing class that had these, but it's pretty impossible to find. Thanks for pointing this out.

nmonterroso: ad99d00

ArturKlajnerok:
I agree with @sqreek that returning full response object with the status would be much cleaner approach but also understand that this would require a bigger refactor of the client and its usage, so considering we want to fix piggyback for now, LGTM :+1:

nmonterroso:
:boom:

PR #9092, opened 2015-11-20
VE 2064: Display count of non-portable infoboxes

pauloslund:


inez:
I think that looks good, do you have any thoughts in regard to cache invalidation? Also, for how long it is being cached?

extensions/wikia/EmergencyBroadcastSystem/EmergencyBroadcastSystemController.class.php
```diff
@@ -2,8 +2,9 @@
 class EmergencyBroadcastSystemController extends WikiaController {
 
 	public function index() {
-		if ( $this->isCorrectPage() && $this->isPowerUser() && $this->hasNonPortableInfoBoxes() && $this->canOpenEBS() ) {
-			$this->response->setVal( 'nonPortableCount', '3' ); // Temporary number for testing
+		$nonPortableCount = $this->countOfNonPortableInfoBoxes();
+		if ( $this->isCorrectPage() && $this->isPowerUser() && $this->canOpenEBS() && ( $nonPortableCount > 0 ) ) {
```
dianafa: What about moving `( $this->isCorrectPage() && $this->isPowerUser() && $this->canOpenEBS() )` check before we make the `getNonportableInfoboxes()` query?

extensions/wikia/EmergencyBroadcastSystem/EmergencyBroadcastSystemController.class.php
```diff
@@ -2,8 +2,9 @@
...
+			$this->response->setVal( 'nonPortableCount', $nonPortableCount ); // Temporary number for testing
```
dianafa: is it still `// Temporary number for testing` ? ;)

extensions/wikia/EmergencyBroadcastSystem/EmergencyBroadcastSystemController.class.php
inez: @dianafa Excellent point.

pauloslund:
It's using CACHE_STANDARD, which is 24 hours. Haven't had a chance to look in to how they were invalidating the cache yet

inez:
Please take a look at Insights/InsightsUnconvertedInfoboxesHooks.class.php I think that code should be modified due to your changes.

pauloslund:
Good catch, I think I'm going to take another look at this PortableInfoboxQueryService

PR #9097, opened 2015-11-21
VE-2044 | added infobox preview entry point

katanka:


extensions/wikia/EditPageLayout/js/editor/AceEditor.js
```diff
@@ -49,6 +49,7 @@ define(
 	function beforeEditorInit() {
 		$('.loading-indicator').remove();
 		$('#wpSave').removeAttr('disabled');
+		$('#infoboxpreview').click(launchEditorPreview);
```
inez: ID perhaps should be camelcased

extensions/wikia/EditPageLayout/js/editor/AceEditor.js
```diff
@@ -176,6 +177,24 @@ define(
 		$('.editpage-widemode-trigger').click(editorModeChange);
 	}
 
+	function launchEditorPreview(){
+		var editor_value = ace.getContent();
```
inez: In general we used camelcase for variable names.

extensions/wikia/EditPageLayout/js/editor/AceEditor.js
```diff
@@ -176,6 +177,24 @@ define(
...
+		var wiki_url = mw.config.get('wgServer').split('://')[1];
+		var template_name = mw.config.get('wgPageName').split(':')[1];
+
+		var encode = function(str) {
+			return encodeURI(str).replace(/'/g, '%27');
+		}
+
+		$('<form action="http://localhost:8080/api/entrypoint/" method="POST">')
```
inez: If you use jQuery .val to set values of those hidden fields then you wouldn't have to deal with encoding altogether. Something like:

```
$inputEditorValue = $('input').val(editorValue).appendTo($form);
```

extensions/wikia/EditPageLayout/templates/EditPageLayout_EditPage.php
```diff
@@ -137,6 +140,10 @@
 								<?php endif ?>
 							</nav>
 						</div>
+						<div class="module_content">
+							<br />
+							Try the new infobox editor!&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;<input type="button" id="infoboxpreview" value="Go to Editor"/>
```
inez: i18n missing

inez: Also using bunch of &nbsp; is kind ugly, any better ideas?

extensions/wikia/EditPageLayout/js/editor/AceEditor.js
inez: Shouldn't be hardcoded to localhost - can you make it a configuration variable?

extensions/wikia/EditPageLayout/js/editor/AceEditor.js
```diff
@@ -176,6 +177,24 @@ define(
 		$('.editpage-widemode-trigger').click(editorModeChange);
 	}
 
+	function launchEditorPreview(){
+		var editor_value = ace.getContent();
+		var wiki_url = mw.config.get('wgServer').split('://')[1];
+		var template_name = mw.config.get('wgPageName').split(':')[1];
```
inez: There must be a better way to get the title... look into mw.Title and mw.util.

extensions/wikia/EditPageLayout/js/editor/AceEditor.js
```diff
@@ -176,6 +177,19 @@ define(
 		$('.editpage-widemode-trigger').click(editorModeChange);
 	}
 
+	function launchEditorPreview(){
+		var editorValue = ace.getContent();
+		var wikiURL = mw.config.get('wgServer').split('://')[1];
+		var templateName = mw.config.get('wgPageName').split(':')[1];
```
inez: Can you use wgTitle? Or use mw.Title object to parse wgPageName instead of doing it with split?

extensions/wikia/EditPageLayout/js/editor/AceEditor.js
```diff
@@ -176,6 +177,19 @@ define(
...
+		var infoboxPreviewURL = 'http://localhost:8080/api/entrypoint/';
```
inez: Just a reminder to make it a config variable.

extensions/wikia/EditPageLayout/js/editor/AceEditor.js
```diff
@@ -176,6 +177,19 @@ define(
 		$('.editpage-widemode-trigger').click(editorModeChange);
 	}
 
+	function launchEditorPreview(){
+		var editorValue = ace.getContent();
```
inez: We have this not strongly enforce policy of only one var statement per scope.

extensions/wikia/EditPageLayout/templates/EditPageLayout_EditPage.php
```diff
@@ -137,6 +140,10 @@
 								<?php endif ?>
 							</nav>
 						</div>
+						<div class="module_content">
+							<br />
+							Try the new infobox editor!&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;<input type="button" id="infoboxPreview" value="Go to Editor"/>
```
inez: Let's try to avoid having nbsps, we should be able to handle it better with CSS

extensions/wikia/EditPageLayout/js/editor/AceEditor.js
```diff
@@ -176,6 +177,19 @@ define(
 		$('.editpage-widemode-trigger').click(editorModeChange);
 	}
 
+	function launchEditorPreview(){
+		var editorValue = ace.getContent(),
+			wikiaDomain = mw.config.get('wgServer').split('://')[1],
+			templateName = mw.Title(mw.config.get('wgPageName'), 'Template').title,
```
inez: @katanka mw.Title is class, isn't it? So should be used with new `new` keyword.
Would that hardcoded "Template" work for all languages?
Use getters instead of accessing fields directly.
It should be done more or less like this: `(new mw.Title("Template:123/456")).getMain()` (not sure about getMain method - maybe some other method should be used)

extensions/wikia/EditPageLayout/js/editor/AceEditor.js
```diff
@@ -176,6 +177,19 @@ define(
...
+			infoboxPreviewURL = 'http://localhost:8080/', // TODO: add to config
+			$form = $('<form>').attr({'action': infoboxPreviewURL, 'method': 'POST'});
```
inez:
```
$('<form>')
    .attr({'action': infoboxPreviewURL, 'method': 'POST'})
    .append(
        $('<textarea>').val(editorValue).attr({'name': 'editor_value'}),
        $('<input>').val(wikiaDomain).attr({'name': 'wikia_domain'}),
        $('<input>').val(templateName).attr({'name': 'template_name'})
    )
    .submit()
```

Not saying that's a better way of doing it - but just showing it could be done this way - jQuery is awesome :)

extensions/wikia/EditPageLayout/EditPageLayoutHelper.class.php
```diff
@@ -51,6 +51,7 @@ function getRequest() {
 	 */
 	function setupEditPage( Article $editedArticle, $fullScreen = true, $class = false ) {
 		global $wgHooks;
+		global $wgInfoboxPreviewURL;
```
inez: one global statement per scope please

extensions/wikia/EditPageLayout/templates/EditPageLayout_EditPage.php
```diff
@@ -137,6 +137,10 @@
 								<?php endif ?>
 							</nav>
 						</div>
+						<div class="module_content">
+							<p><?= wfMsgExt('infoboxpreview-entrypoint-message', array('parse') ) ?></p>
```
inez: Isn't wfMsgExt deprecated in favor of wfMessage?

inez: Actually, I checked and yes it is.

inez:
LGTM but I was pretty involved in it so would be great if other people can take a look too. @pauloslund @RafLeszczynski Could you please?

extensions/wikia/EditPageLayout/css/EditPageLayout.scss
```diff
@@ -599,3 +599,9 @@ body {
 		@include breakpoints-large-typography;
 	}
 }
+
+#infoboxPreview {
```
RafLeszczynski: Please consider not using id selectors in CSS

extensions/wikia/EditPageLayout/css/EditPageLayout.scss
```diff
@@ -599,3 +599,9 @@ body {
...
+	width: 150px;
```
RafLeszczynski: Properties should be in alphabetical order

extensions/wikia/EditPageLayout/js/editor/AceEditor.js
```diff
@@ -176,6 +177,19 @@ define(
 		$('.editpage-widemode-trigger').click(editorModeChange);
 	}
 
+	function launchEditorPreview(){
```
RafLeszczynski: Please consider changing method name. If I understand correct it should open a new window with infobox preview not editor preview

extensions/wikia/EditPageLayout/js/editor/AceEditor.js
```diff
@@ -176,6 +177,19 @@ define(
...
+		var editorValue = ace.getContent(),
+			wikiaDomain = mw.config.get('wgServer').split('://')[1],
+			templateName = (new mw.Title(mw.config.get('wgPageName'))).getMain(),
+			infoboxPreviewURL = mw.config.get('wgInfoboxPreviewURL'),
+			$form = $('<form>').attr({'action': infoboxPreviewURL, 'method': 'POST', 'target': '_blank'});
```
RafLeszczynski: it's just a readability / code separation optimisation but you could consider moving form creation to separate method

extensions/wikia/EditPageLayout/js/editor/AceEditor.js
```diff
@@ -176,6 +177,19 @@ define(
 		$('.editpage-widemode-trigger').click(editorModeChange);
 	}
 
+	function launchEditorPreview(){
+		var editorValue = ace.getContent(),
+			wikiaDomain = mw.config.get('wgServer').split('://')[1],
```
RafLeszczynski: try `window.location.hostname`

inez: What's the benefit of it? We know for sure that wgServer is correct - we control it (we set it, we depend on it in many places).

extensions/wikia/EditPageLayout/EditPageLayoutController.class.php
```diff
@@ -206,6 +208,12 @@ public function executeEditPage() {
 			? wfMessage( 'editpagelayout-notificationsLink-none' )->escaped()
 			: wfMessage( 'editpagelayout-notificationsLink', count( $this->notices ) )->parse();
 
+		$templateType = ( new TemplateClassificationService() )
+			->getType( $wgCityId, $this->title->getArticleID() );
```
RafLeszczynski: This method can throw an Exception. Please consider adding exception handling.

inez: @RafLeszczynski Good point. @katanka My suggestion would be to catch exception and then not display the link but also make sure to log exception (look in other place how things are being logged).

extensions/wikia/EditPageLayout/css/EditPageLayout.scss
katanka: If I use a class selector, I have to apply `!important` to some of the styles. Since this is only applying to this specific element, doesn't it make more sense to use id?

extensions/wikia/EditPageLayout/EditPageLayoutController.class.php
```diff
@@ -206,6 +210,20 @@ public function executeEditPage() {
 			? wfMessage( 'editpagelayout-notificationsLink-none' )->escaped()
 			: wfMessage( 'editpagelayout-notificationsLink', count( $this->notices ) )->parse();
 
+		try {
+			$templateType = ( new TemplateClassificationService() )
+				->getType( $wgCityId, $this->title->getArticleID() );
+		} catch ( Exception $e ) {
+			$templateType = TemplateClassificationService::TEMPLATE_UNKNOWN;
```
inez: @katanka Why do you need to set $templateType to TEMPLATE_UNKNOWN when exception occurred?
Wouldn't something like this achieve same effect in more straightforward way?

```
try {
    $templateType = ( new TemplateClassificationService() )->getType( $wgCityId, $this->title->getArticleID() );
    $this->showInfoboxPreview = $templateType == INFOBOX || $templateType == CUSTOM_INFOBOX;
} else {
    $this->showInfoboxPreview = false;
    Log('omg, something went wrong');
}
```

inez: Actually given that we have extra code reviewer let me go ahead and change it.

extensions/wikia/EditPageLayout/css/EditPageLayout.scss
```diff
@@ -599,3 +599,9 @@ body {
 		@include breakpoints-large-typography;
 	}
 }
+
+#InfoboxPreview {
+	margin-left: 50px;
+	margin-top: 10px;
+	width: 150px;
+}
```
pauloslund: I agree with @RafLeszczynski about not using an id to style this. If the style declaration needs to be more specific you can use something like `input.InfoboxPreview` or consider nesting this under `.EditPageRail`.
Also, you can center this horizontally using:

```
.EditPageRail .InfoboxPreview {
  margin: 10px auto 0;
  display: block;
}
```

Instead of the `50px` magic number

extensions/wikia/EditPageLayout/js/editor/AceEditor.js
```diff
@@ -49,6 +49,7 @@ define(
 	function beforeEditorInit() {
 		$('.loading-indicator').remove();
 		$('#wpSave').removeAttr('disabled');
+		$('#InfoboxPreview').click(launchInfoboxPreview);
```
inez: @katanka So InfoboxPreview button is going to work only when AceEditor is being used (because handler is assigned in here - AceEditor.js). So my question is: is the button displayed only when the AceEditor is being used? Because from what I see conditions for displaying InfoboxPreview button are different than conditions for displaying actually AceEditor (isCodePage).

extensions/wikia/EditPageLayout/css/EditPageLayout.scss
```diff
@@ -599,3 +599,8 @@ body {
 		@include breakpoints-large-typography;
 	}
 }
+
+#InfoboxPreview {
```
inez: @katanka Please address @RafLeszczynski and @pauloslund comments about not using ID (and perhaps using class name instead).

extensions/wikia/EditPageLayout/js/loaders/EditPageEditorLoader.js
```diff
@@ -126,7 +126,25 @@
 
 		},
 
+		launchInfoboxPreview: function() {
+			var editorValue = window.wgEnableCodePageEditor ? ace.edit("editarea").getValue() : WikiaEditor.getInstance().getContent(),
```
inez: Use `'` instead of `"`

extensions/wikia/EditPageLayout/css/EditPageLayout.scss
```diff
@@ -105,6 +105,11 @@ body {
 	width: $width-rail-epl;
 	z-index: 101; // and the loading status indicator (z-index: 100)
 
+	.InfoboxPreview {
+		margin: 10px auto 0;
```
RafLeszczynski: please use alphabetical order for style declarations

extensions/wikia/EditPageLayout/EditPageLayoutController.class.php
```diff
@@ -206,6 +210,19 @@ public function executeEditPage() {
 			? wfMessage( 'editpagelayout-notificationsLink-none' )->escaped()
 			: wfMessage( 'editpagelayout-notificationsLink', count( $this->notices ) )->parse();
 
+		try {
+			$templateType = ( new TemplateClassificationService() )
```
inez: Please consider if the call to this service should depend on wgEnableTemplateClassificationExt variable.

extensions/wikia/EditPageLayout/EditPageLayoutController.class.php
```diff
@@ -206,6 +210,24 @@ public function executeEditPage() {
 			? wfMessage( 'editpagelayout-notificationsLink-none' )->escaped()
 			: wfMessage( 'editpagelayout-notificationsLink', count( $this->notices ) )->parse();
 
+		if ( $wgEnableTemplateClassificationExt ) {
```
inez, line 213: It should be `if(!(empty($wgEnableTemplateClassificationExt))) {...}` I believe - same as in other places.

pauloslund:
LGTM

extensions/wikia/EditPageLayout/EditPageLayoutController.class.php
katanka, line 213: Looks like usage is inconsistent, it is used directly in `isCodePage()`

PR #9098, opened 2015-11-23
CloseWikiMaintenance: introduce a helper for removing wiki DFS bucket

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-1700

Maintenance script that closes old wikis used to remove upload directory in the medieval, NFS-powered times. The obsolete code was removed in 60b7121.

This pull request introduces a helper method that recursively removes wiki files (using `s3cmd --recursive del` command). Small code cleanup included :smiley: 

@michalroszka / @wladekb / @ljagiello

### DIFF
--- a/extensions/wikia/WikiFactory/Close/maintenance.php
+++ b/extensions/wikia/WikiFactory/Close/maintenance.php
@@ -420,20 +420,21 @@ class CloseWikiMaintenance {
 	 */
 	private function removeBucket( $cityid ) {
 		try {
-			$container = \Wikia\SwiftStorage::newFromWiki( $cityid )->getContainer();
-			$this->log( "Removing DFS bucket - {$container}" );
+			$swift = \Wikia\SwiftStorage::newFromWiki( $cityid );
+			$this->log( sprintf( "Removing DFS bucket /%s%s", $swift->getContainerName(), $swift->getPathPrefix() ) );
 
-			//  Remove bucket / s3cmd rb s3://BUCKET
+			// s3cmd --recursive del s3://BUCKET/OBJECT / Recursively delete files from bucket
 			$cmd = sprintf(
-				'sudo /usr/bin/s3cmd -c %s --verbose rb s3://%s/',
+				'sudo /usr/bin/s3cmd -c %s --recursive del s3://%s%s/',
 				self::S3_CONFIG,
-				$container->name
+				$swift->getContainerName(),  # e.g. 'nordycka'
+				$swift->getPathPrefix()      # e.g. '/pl/images'
 			);
 			$out = wfShellExec( $cmd, $iStatus );
-			$this->log( $cmd . ' : ' . $out );
+			$this->log( $cmd );
 
 			if ( $iStatus !== 0 ) {
-				throw new Exception( 'Failed to remove a bucket - ' . $cmd, $iStatus );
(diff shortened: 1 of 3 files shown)


macbre:
~~Removing the entire bucket will cause problems for non English wikis sharing the bucket with the English one (muppet being one of many examples). We don't want to remove English Foo Bar wiki bucket when removing non-English Foo Bar wiki version.~~

extensions/wikia/WikiFactory/Close/maintenance.php
```diff
@@ -416,6 +413,38 @@ private function getDirTree( $dir ) {
...
+			$swift = \Wikia\SwiftStorage::newFromWiki( $cityid );
+			$this->log( sprintf( "Removing DFS bucket /%s%s", $swift->getContainerName(), $swift->getPathPrefix() ) );
+
+			// s3cmd --recursive del s3://BUCKET/OBJECT / Recursively delete files from bucket
+			$cmd = sprintf(
+				'sudo /usr/bin/s3cmd -c %s --verbose del s3://%s%s/',
+				self::S3_CONFIG,
+				$swift->getContainerName(),  # e.g. 'nordycka'
+				$swift->getPathPrefix()      # e.g. '/pl/images'
+			);
```
macbre: Tested on devbox DFS using `s3cmd --dry-run`

extensions/wikia/WikiFactory/Close/maintenance.php
```diff
@@ -416,6 +413,38 @@ private function getDirTree( $dir ) {
...
+	 * @param int $cityId
+	 */
+	private function removeBucket( $cityid ) {
+		try {
+			$swift = \Wikia\SwiftStorage::newFromWiki( $cityid );
+			$this->log( sprintf( "Removing DFS bucket /%s%s", $swift->getContainerName(), $swift->getPathPrefix() ) );
+
+			// s3cmd --recursive del s3://BUCKET/OBJECT / Recursively delete files from bucket
+			$cmd = sprintf(
+				'sudo /usr/bin/s3cmd -c %s --recursive del s3://%s%s/',
```
ljagiello, line 428: This is really hacky. Can you use something more native for PHP like [boto](https://github.com/boto/boto) for Python?

In case of any config changes or directory restructure I'm afraid we will forget about `const S3_CONFIG` and this logic.

macbre, line 428: Unfortunately, CloudFiles library (from Rackspace) that we use in MediaWiki only supports removing single files and the whole buckets.

Can't think of something different than fetching the list of ALL files in the bucket (via CloudFiles API call), filtering it and removing.

And we already use `s3cmd` in this script for creating images backups.

ghost, line 428: And we can't be sure so called "more native PHP lib" will be less hacky :-) But we can definitely thing about abstracting out some things and simplify the usage of S3 across the code base: https://wikia-inc.atlassian.net/browse/PLATFORM-1704

macbre, line 428: True, true :) CloudFiles was removed in the MediaWiki upstream repo (it's quite hacky and ugly code). WMF guys implemented their own Swift client in PHP.

PR #9101, opened 2015-11-23
VE 2074: Method for building the survey url

pauloslund:


extensions/wikia/EmergencyBroadcastSystem/EmergencyBroadcastSystemController.class.php
```diff
@@ -64,4 +65,68 @@ protected function hasNonPortableInfoBoxes() {
...
+			case 'es':
+				$url = "https://docs.google.com/forms/d/1MLwwR8t-uoIOBXjjlW7YnBLOOrbH8_19Q3N5Bttqp0Y/viewform";
+				break;
+			case 'en':
+			default:
+				$url = "https://docs.google.com/forms/d/18qE5ub8qs8bkrubcN-00JxuLrfnfpJ88MyGhq1x3RpY/viewform";
+				break;
+		}
+
+		$url = $url . "?entry.2019588325={$userName}&entry.1160019288={$domain}&entry.830663849";
```
inez: Most likely those ids: "entry.2019588325" are going to be different for each language - so this is not going to work.

extensions/wikia/EmergencyBroadcastSystem/EmergencyBroadcastSystemController.class.php
```diff
@@ -64,4 +65,68 @@ protected function hasNonPortableInfoBoxes() {
...
+
+		$url = $this->googleFormUrl( $lang, $userName);
+
+		return $url;
+	}
+
+	protected function googleFormUrl( $language, $userName ) {
+		global $wgServer;
+
+		preg_match( '/\/\/(.*)/', $wgServer, $matches );
```
inez: There "must" be some better way to figure out wikia domain - can you look for it. The less regex usually the better.

extensions/wikia/EmergencyBroadcastSystem/EmergencyBroadcastSystemController.class.php
```diff
@@ -64,4 +65,68 @@ protected function hasNonPortableInfoBoxes() {
...
+
+	protected function buildSurveyUrl() {
+		global $wgContLang;
+
+		$context = $this->getContext();
+		$userName = $context->getUser()->getName();
+		$userLanguage = $context->getLanguage();
+
+		$lang = $userLanguage->mCode;
+		$supportedLanguages = ['en', 'de', 'es', 'fr', 'it', 'ja', 'pl', 'pt-br', 'ru', 'zh'];
```
inez: It's a bummer that this list have to exist twice: here, and inside the googleFormUrl.
How about you declare just an associative array:

```
$masks = array(
    'en' => 'https://docs.google.com/forms/d/1vnLsNYhWPa4lghIMyC8Eyu6gK0dZ0N3j4_RP1K3HH28/viewform..domain={0}..user={1}',
    'de' => 'https://docs.google.com/forms/d/1vnLsNYsdsdghIMyC8Eyu6gK0dZ0N3j4_RP1K3HH28/viewform..domain={0}..user={1}',
    ....
)
```

I think that would be very optimal for all use cases that we are dealing with in here.

inez:
Also, given that Trevor created all the surveys already - please use the proper URLs - instead of just English.

extensions/wikia/EmergencyBroadcastSystem/EmergencyBroadcastSystemController.class.php
inez: Also... make sure it handles it well when username contains spaces, etc.

pauloslund: They are currently the same for all languages (probably because the forms were all just duplicated from the English form), however I am not sure how Google generates the ids. I can embed them into the string.

pauloslund:
They are currently the proper URLs, however the surveys do not all have the translated content in them.

inez:
@pauloslund Can you make it mergeable please?

extensions/wikia/EmergencyBroadcastSystem/EmergencyBroadcastSystemController.class.php
```diff
@@ -64,4 +65,44 @@ protected function hasNonPortableInfoBoxes() {
...
+		}
+
+		if ( $url === null ) {
+			$url = $this->googleFormUrl( 'en', $userName );
+		}
+
+		return $url;
+	}
+
+	protected function googleFormUrl( $language, $userName ) {
```
inez: So in this approach the array below is going to be defined up to three times, I don't think there is a good reason for that.
Also, isn't accessing not defined index of an array throwing a notice?

inez: It also parse url up to three times.

PR #9102, opened 2015-11-23
SOC-1710 Fix WallNotifications query

jsutterfield:
As part of https://github.com/Wikia/app/pull/9048 we updated the query used to get notifications, making it more efficient. That pull request had a mistake however, it was missing the "DESCENDING" qualifier for the "ORDER BY". This PR adds "DESC" into the query.

Ticket: https://wikia-inc.atlassian.net/browse/SOC-1710


garthwebb:
LGTM

PR #9106, opened 2015-11-23
Machine classify user submitted templates

sebastianmarzjan:
This CR introduces an extension to current manual template classification:
Publish a template update event whenever a user classifies a template, making sure we classify it automatically too.

/cc @lgarczewski @lukaszkonieczny @Grunny @adamkarminski @idradm @RafLeszczynski @rybmat @dianafa @azarek @nikodamn @kamilkoterba

### DIFF
--- a/extensions/wikia/IndexingPipeline/PipelineEventProducer.class.php
+++ b/extensions/wikia/IndexingPipeline/PipelineEventProducer.class.php
@@ -111,6 +111,24 @@ class PipelineEventProducer {
 		return true;
 	}
 
+	/**
+	 * @desc Fires on:
+	 *  - successful save from TemplateClassificationApiController
+	 *
+	 * @param integer $pageId The affected template's pageId
+	 * @param Title $title The affected template's Title object
+	 * @return bool
+	 */
+	public static function onTemplateClassified( $pageId, $title ) {
+		$ns = self::preparePageNamespaceName( $title );
+		$revisionId = $title->getLatestRevID();
+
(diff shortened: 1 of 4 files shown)


kamilkoterba:
Classification happens also
onArticleSaveComplete
https://github.com/Wikia/app/blob/dev/extensions/wikia/TemplateClassification/TemplateClassification.hooks.php#L57-57
and on infobox draft creation
https://github.com/Wikia/app/blob/dev/extensions/wikia/TemplateDraft/controllers/TemplateDraftController.class.php#L26-26

Is it something that you want to cover as well?

lukaszkonieczny:
We introduce `UserTemplateClassificationService.class.php` (which is in https://github.com/Wikia/app/pull/9075/files and waiting for testing) where we overwrite `getType` method from `TemplateClassificationService`. Maybe we can overwrite also `classifyTemplate` method, use it in places which Kamil mentioned and add this hook there?

sebastianmarzjan:
We already cover ArticleSaveComplete ( https://github.com/Wikia/app/blob/dev/extensions/wikia/IndexingPipeline/PipelineEventProducer.class.php#L55 ).
I will add handling for the draft, as it makes perfect sense to me.

I think that what @lukaszkonieczny proposes makes sense as next step too, but I would not block this PR because of that

lukaszkonieczny:
LGTM

lgarczewski:
LG2M2. ;) Tested on human interface. Waiting for @rybmat to test the pipeline behaviour.

PR #9111, opened 2015-11-24
VE-2084: Direct user to correct page after clicking yes

pauloslund:


extensions/wikia/EmergencyBroadcastSystem/js/EmergencyBroadcastSystem.js
```diff
@@ -1,11 +1,22 @@
...
+		$( '.ebs-container' ).hide();
+		return $.nirvana.postJson( 'EmergencyBroadcastSystemController', 'saveUserResponse', { val: val } );
 	}
 
-	$('.ebs-primary-action').click(function(){ handleClick( 1 ); return false; });
-	$('.ebs-secondary-action').click(function(){ handleClick( 0 ); return false; });
-})(jQuery);
+	$( '.ebs-primary-action' ).click( function(){
+		var href = this.href;
+		var request = handleClick( 1 );
```
inez:
```
handleClick( 1 ).then( function() {
...
}
```

That way you don't not the request variable.
Btw. using then is sort of better - because it does the same thing (more or less) but it is a promise standard - according to spec.

inez: Also if you bind then you wouldn't have to ever have href variable - up to you.

PR #9113, opened 2015-11-24
SOC-1485 Fix threads affected by the duplicate page_wikia_props rows

garthwebb:
Some threads had comments with duplicate comment index.  This means that the permalink for multiple comments was the same.

This script runs against a single wiki and fixes the problem.

https://wikia-inc.atlassian.net/browse/SOC-1485


extensions/wikia/Wall/WallBaseController.class.php
```diff
@@ -46,7 +46,6 @@ public function thread() {
 
 		$this->addAsset();
 
-		$title = $this->request->getVal( 'title', $this->app->wg->Title );
```
garthwebb: The `$title` var isn't used.

extensions/wikia/Wall/WallMessage.class.php
```diff
@@ -148,7 +148,7 @@ static public function buildNewMessageAndPost( $body, $page, $user, $metaTitle =
 			$class->setOrderId( 1 );
 			$class->getWall()->invalidateCache();
 		} else {
-			$count = $parent->getOrderId( true ); // this is not work perfect with transations
+			$count = $parent->getOrderId( $userMaster = true );
```
garthwebb, line 153: This comment was meaningless as far as I could tell

maintenance/wikia/Forum/fixCommentIndexes.php
```diff
@@ -0,0 +1,308 @@
...
+ *
+ */
+
+ini_set('display_errors', 'stderr');
+ini_set('error_reporting', E_NOTICE);
+
+require_once( dirname( __FILE__ ) . '/../../Maintenance.php' );
+
+/**
+ * Class FSCKVideos
```
kvas-damian: Please remove this docblock

maintenance/wikia/Forum/fixCommentIndexes.php
```diff
@@ -0,0 +1,308 @@
...
+ */
+class FixCommentIndexes extends Maintenance {
+	static protected $verbose = false;
+	static protected $test = false;
+
+	public function __construct() {
+		parent::__construct();
+		$this->mDescription = "Pre-populate LVS suggestions";
```
kvas-damian: Please update description

maintenance/wikia/Forum/fixCommentIndexes.php
```diff
@@ -0,0 +1,308 @@
...
+		$lastIdx = 0;
+		$foundDuplicates = false;
+
+		foreach ( $comments as $commentId ) {
+			$thisIdx = wfGetWikiaPageProp( WPP_WALL_COUNT, $commentId );
+			$isDuplicate = $thisIdx == $lastIdx;
+			$lastIdx = $thisIdx;
+
+			// If we found the start of duplicate indexes, start renumbering
+			if ( $isDuplicate ) {
```
kvas-damian: Don't we need to renumber all comments after we find duplicate?
Or how does this logic fixes this issue: https://wikia-inc.atlassian.net/browse/SOC-1485?focusedCommentId=210327&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-210327

maintenance/wikia/Forum/fixCommentIndexes.php
```diff
@@ -0,0 +1,308 @@
...
+	protected $newWallCount;
+
+	/**
+	 * @param int $threadId
+	 */
+	public function __construct( $threadId ) {
+		$this->threadId = $threadId;
+
+		$this->currentIndex = $this->getCurrentIndex();
+		$this->commentCount = $this->getCommentCount();
```
kvas-damian: Can't we use local variables?

maintenance/wikia/Forum/fixCommentIndexes.php
```diff
@@ -0,0 +1,308 @@
...
+
+	/**
+	 * Sets the new index to use whenever the next comment is made on the current thread
+	 */
+	private function setNewWallCount() {
+		if ( FixCommentIndexes::isTest() ) {
+			return;
+		}
+
+		wfSetWikiaPageProp( WPP_WALL_COUNT, $this->threadId, $this->newWallCount );
```
kvas-damian: Are there any side effect of modifying this value?

maintenance/wikia/Forum/fixCommentIndexes.php
```diff
@@ -0,0 +1,308 @@
...
+		$index = wfGetWikiaPageProp( WPP_WALL_COUNT, $this->threadId );
+		return empty( $index ) ? 0 : $index;
+	}
+
+	/**
+	 * Get the count of comments made on the thread given by $threadId.  We do this as a separate query rather
+	 * call PHP's count on the comments we get later so that we can quickly update the starting comment
+	 * index before we do anything else.
+	 *
+	 * @return array
```
jsutterfield: This is actually returning an `int`, right?

maintenance/wikia/Forum/fixCommentIndexes.php
```diff
@@ -0,0 +1,308 @@
...
+		$count = ( new WikiaSQL )
+			->SELECT( 'count(*)' )->AS_( 'count' )
+			->FROM( 'comments_index' )
+			->WHERE( 'parent_comment_id' )->EQUAL_TO( $this->threadId )
+			->ORDER_BY( 'comment_id' )
+			->runLoop( $dbr, function ( &$count, $row ) {
+				$count = $row->count;
+			});
+
+		// Ensure that we're always sending an array back
```
jsutterfield: Again, returning an `int`.

maintenance/wikia/Forum/fixCommentIndexes.php
```diff
@@ -0,0 +1,308 @@
...
+ * these comments (stored in the page_wikia_props table) to be unique, e.g.:
+ *
+ *   http://creepypasta.wikia.com/wiki/Thread:510920#2
+ *                                               ...#3
+ *                                               ...#4
+ *                                               ...#5
+ *                                               ...#6
+ *
+ * Its possible that this streak of duplicates can happen within normally numbered comments.  This is because
+ * we released a fix for the problem before creating this script.  All we need to do is make sure we only update
```
jsutterfield: This bug was caused by PLATFORM-1658, right? Can we add that ticket # here for some additional context about what that fix was?

maintenance/wikia/Forum/fixCommentIndexes.php
jsutterfield: I'm wondering the same thing. If we start to renumber comments in the middle and then stop, won't they have higher indexes then older comments?

maintenance/wikia/Forum/fixCommentIndexes.php
garthwebb: Ah, thanks, that's a good addition.

maintenance/wikia/Forum/fixCommentIndexes.php
garthwebb: yes

maintenance/wikia/Forum/fixCommentIndexes.php
garthwebb: again, yes ;)

Also that ORDER_BY up there is useless...

maintenance/wikia/Forum/fixCommentIndexes.php
garthwebb: To my knowledge all WPP_WALL_COUNT does is store the unique number to be put at the end of the comment URL.  When its used for a comments page prop, its the unique number for that particular comments URL.

maintenance/wikia/Forum/fixCommentIndexes.php
garthwebb: They will definitely have higher IDs but it definitely wont matter.  The comments are ordered by date and the Wall authors could have saved themselves (and us) a lot of time and heartache by making these a combination of a random number and the current time.

All this code does is look at a string of numbers like this:

``` php
  [ 2, 3, 4, 5, 5, 5, 5, 5, 5, 6, 7, 8, 9 ]
```

then identify the duplicates, and update them to be something that won't overlap with the current numbers:

``` php
  [ 2, 3, 4, 5, 110, 111, 112, 113,114, 6, 7, 8, 9 ]
```

It also has to set the parents `WPP_WALL_COUNT` to be something that won't overlap with the numbers we just added, so it adds the count of comments to where we started renumbering duplicates, or `13 + 110 == 123`  The number of comments is used in case all comments had dups (technically it could be num `comments - 1` but we don't need to be stingy with these numbers and it gives us a lot of breathing room).

To be honest, it would have been better and simpler to use a random number plus the time for the duplicates but I didn't want to make those URLs that much different than the rest.

maintenance/wikia/Forum/fixCommentIndexes.php
garthwebb: Yes.

jsutterfield:
LGTM

maintenance/wikia/Forum/fixCommentIndexes.php
kvas-damian: Are you sure it's not used on thread history page?
http://community.wikia.com/wiki/Thread:925591?action=history
I didn't check that but I expect it can introduce gaps in the history.

maintenance/wikia/Forum/fixCommentIndexes.php
kvas-damian: The problem here is more complex. If we decide to reorder those items, then we'll break existing permalinks.
If we don't do that history page can be confusing. I think we should let Product decide what to do.

garthwebb: @kvas-damian, the links are already broken; we're just making sure they can be used in the future.

kvas-damian: I'm referring to links that are already fixed - numbers higher than duplicates.

kvas-damian:
Left one last concern - if it's not a problem then it LGTM

maintenance/wikia/Forum/fixCommentIndexes.php
garthwebb: The `WPP_WALL_COUNT` is displayed on the history page for each individual comment.  The main count (what's being set here) isn't involved with this.

maintenance/wikia/Forum/fixCommentIndexes.php
```diff
@@ -0,0 +1,262 @@
...
+
+	/**
+	 * Update the comment number used for creating links to those comments.  It doesn't matter what they are
+	 * as long as each comment has a unique number, however ascending from 2 is how it works now and makes the
+	 * most sense.
+	 *
+	 * @return int The last index we assigned to a comment
+	 */
+	private function renumberComments() {
+		$comments = $this->getCommentIds();
```
kvas-damian, line 198: User can comment while we're updating duplicated records so new comment will still be out of order. 

To avoid this problem we can:
- fetch comments one by one - even if someone post new comment then it'll be reordered
- check if new comments were added and fix them as well
- lock DB for time of update

maintenance/wikia/Forum/fixCommentIndexes.php
```diff
@@ -0,0 +1,262 @@
...
+
+			if ( !FixCommentIndexes::isTest() ) {
+				wfSetWikiaPageProp( WPP_WALL_COUNT, $commentId, $commentIdx++ );
+			}
+		}
+
+		FixCommentIndexes::debug( "done\n" );
+
+		// At the end of the loop we're 1 past the last index used
+		return $commentIdx - 1;
```
kvas-damian: `BrokenThread` is a separate class that doesn't check if it's really broken. If someone uses for good thread then this statement will return `-1` which will finally affect https://github.com/Wikia/app/pull/9113/files#diff-dcd5ddf267c5faa7d63ddb5817c6f28eR164

What do you think about additional safety check for this situation?

maintenance/wikia/Forum/fixCommentIndexes.php
```diff
@@ -0,0 +1,262 @@
...
+		$commentIdx = 0;
+		$lastIdx = 0;
+		$foundDuplicates = false;
+
+		foreach ( $comments as $commentId ) {
+			$thisIdx = wfGetWikiaPageProp( WPP_WALL_COUNT, $commentId );
+			$isDuplicate = $thisIdx == $lastIdx;
+
+			// Run this when we find the first duplicate
+			if ( $isDuplicate && !$foundDuplicates ) {
```
kvas-damian: Probably those conditions could be simplified to one flag:
`$foundDuplicates |= $thisIdx == $lastIdx`

Then just using preincrementation for updates.

jsutterfield:
LGTM

maintenance/wikia/Forum/fixCommentIndexes.php
garthwebb: BrokenThread isn't meant to check for broken threads its meant to represent a thread that is already known to be broken.

maintenance/wikia/Forum/fixCommentIndexes.php
garthwebb, line 198: That's an understood risk.  I'm trying to keep this as simple as possible to spend the minimum amount of time necessary for a one time fix of a product we'll be replacing in the next few months.

Fetching new comments and checking for new comments don't eliminate the race condition, just moves it around (there is still a nonatomic read followed by a write).  Additionally, it would add complexity to the script and will slow execution considerably.

Locking is the only way to ensure nothing updates between determining the value to use for the wall count, and setting that value.  However I would have obtain a lock on `DB_MASTER` and use that handle for both reads and writes.  This means this script would be hammering the master DB across all clusters while running and would repeatedly lock the table used for all commenting on a wiki, not just forums.

As a compromise I can reduce the race condition window.  If I did the `count(*)` on the comment index table followed immediately with `wfSetWikiaPageProp(WPP_WALL_COUNT, ... )` for the thread, there would be a much smaller window for new comments to be made.  Any replies made for the rest of the script execution will use the correct index as we renumber all replies behind it.  A reply made between the `count(*)` read and the `WPP_WALL_COUNT` write would create a repeated index, the problem this script fixes.  If this happens and if someone complains about it, we can easily re-run the script for that single thread.

kvas-damian, line 198: @garthwebb I'm not sure why fetching comments one by one won't help.
Example:
1. get comment
2. update it
3. go to first point

Edge cases:
- if user puts his comment before we fetch it, we'll fetch and update it.
- if user puts his comment after we query DB and don't get new comments, then we're done - Comment already has right index.

Finally for updating comments count on a thread we can use
`update ... set count=(select count(*) from...) where ...`

Summarizing we can implement it almost bullet proof relatively easy.

Fix that we have currently implemented can still result in comments that use duplicated ids.

garthwebb, line 198: @kvas-damian its not that simple.  On save, new comments look at the `WPP_WALL_COUNT` for the parent.  It increments that value and writes it back to the parent and then updates itself: 

``` php
$count = $parent->getOrderId( $userMaster = true );
if ( is_numeric( $count ) ) {
    $count++;
    $parent->setOrderId( $count );
    $class->setOrderId( $count );
}
```

(code here: https://github.com/Wikia/app/blob/b00dd8fce32b720af3fd646e45204effb7b9ed2e/extensions/wikia/Wall/WallMessage.class.php#L151-L156)

Note that even this wall code contains a race condition between the read of the master and the write of the new count that I have already seen hit in our older data.

So lets assume we have the following broken data for the parent and its comments shown, noting that comments always start at 2:
- parent count: 3
- comment 1 count: 2
- comment 2 count: 3
- comment 3 count: 3
- comment 4 count: 3

We have to fix _both_ parent and comment counts.  Yah, we can follow your steps:
1. get comment
2. update it
3. go to first point

but until we update the parent's count, the next comment is always going to have the wrong count.  Here's the race condition:

T) Look and find a comment with a comment ID greater than the last comment we fixed
T+1) Update the comment count to $count += 1
T+2) Look and _fail_ to find a comment ID greater than what we found in step N
T+3) _A comment is made in the system and takes the current parent count which is 3._
T+4) Now that we have failed to find any more comments in T+2, update parent count to $count

We now have a duplicate #3 count.  Looking at individual comments doesn't help because we always have a race condition in updating the parent count.  You can't do T+2 and T+4 atomically without locking the table.

Given the Wall code already lives with many bugs _and_ there are obvious race conditions already _and_ that we're replacing it _and_ that the solution agreed up on by the community team accepts broken URLs as part of the fix, I'm not interested in spending a lot of time on a bulletproof solution for something already subpar.  We just need to fix the bulk of the problem.  If  1 wiki out of 1,000 experiences the race condition and leaves a duplicated comment behind, I'm ok with that.  Odds are that only 1 out 1,000 of those duplicates would actually be noticed by somebody, so for that 1 in a 1,000,000 situation, I'm ok re-running the script for that thread.

garthwebb, line 198: Sorry, I forgot to address your point about updating the thread count, e.g.:

``` sql
update ... set count=(select count(*) from...) where ...
```

Yes, that is definitely a great idea and would work with most sane code.  However wall code is not sane.  Comments are meant to start at count=2.  However, I've seen them start at 3 in some cases.  I've also seen missing numbers in the series (even considering comments marked as deleted or removed) so its not possible to set the thread count to the count of current comments + 1 and have it be correct.

This is why the current code updates the thread count first, and then makes one final update afterward if the thread count and last comment count are different.

kvas-damian, line 198: Good point about using parent count for generating new ones. I missed that when analyzing the problem.

About:
`update ... set count=(select count(*) from...) where ...`
If count is not an option here we could use `max()`

About counters we could update scenario to:
T) Look and find a comment with a comment ID greater than the last comment we fixed
T+0.5) Update count on parent
T+1) Update the comment count to $count += 1
...

From my perspective, it's not very time consuming to solve this complex problem. We're almost there. On the other hand having this fix bulletproof would make me fell more safe when running the script, however I won't force you to change current implementation.

garthwebb, line 198: Well, `max()` won't give us the right answer either since we are dealing with a set with duplicates, e.g. '6' could be the max, but 6 might occur ten times so wouldn't really be the number we need.  What we really want is `max()` plus the number of duplicates, which we can't really do in a single SQL statement.  Oh, also, `max()` won't work on the `page_wikia_props.props` column (where the comment index is stored) because its serialized PHP data:

``` sql
mysql> select 10 > 9;
+--------+
| 10 > 9 |
+--------+
|      1 |
+--------+
mysql> select 'i:10;' > 'i:9;';
+------------------+
| 'i:10;' > 'i:9;' |
+------------------+
|                0 |
+------------------+
```

We could use string functions to extract the number part and then cast to an int, but things get really sketchy and ill-performant at this point. :/

Also, issue with the counter steps is that I can still insert a race:

```
T) Look and find a comment with a comment ID greater than the last comment we fixed
T+0.5) Update count on parent
T+0.75) Comment comes in, sets itself to $count + 1 
T+1) Update the comment count to $count += 1, same as comment in T+0.75
```

If the process to determine what the parent count should be is not atomic with writing that value back, there will always be a way to exploit it and introduce a new dupe.

_However..._

I just talked to Bert and he's ok with us renumbering all comments on a broken/duplicate thread from start to finish.  This will allow us to use your `update count=(select...)` suggestion.  Imagine this set of comments that starts on a weird number, has duplicates, and has had new comments after the problem was fixed:

```
... #6  <-- Weird first index
... #7
... #8
... #8  <-- First dup
... #8
... #9  <-- First comment after fix
... #10
```

The parent count for this list is `10` so the next comment will be `$count+1` == `11`.

If we can renumber everything, what we want this to look like is this:

```
... #2
... #3
... #4
... #5
... #6
... #7
... #8
```

The parent count for this list should be `8` so that the next comment will be `$count+1` == `9`

Here are the steps
T+0) Set the parent count via `update count=(select count(*) + 1 ... )` which atomically sets the parent to `8`
T+1) We are spammed with comments to this thread, all of which finish during this time period, and the list now looks like this:

```
... #6
... #7
... #8
... #8
... #8
... #9
... #10
... #9   <-- First spam comment; its 9 because we updated the parent count to 8 in the last step
... #10
... #11
... #12
```

The parent count is now `13`
T+2) We select all comment IDs (the IDs, not the index/count) for the thread
T+3) We get spammed with more comments:

```
... #6   <-- First comment selected by step at T+2
... #7
... #8
... #8
... #8
... #9
... #10
... #9   <-- First spam comment
... #10
... #11
... #12 <-- Last comment selected by step at T+2
... #13 <-- Second wave of spam
... #15
```

T+4) We iterate over the list of comment IDs and update the the count for each starting at `2`

```
... #6  => #2
... #7  => #3
... #8  => #4
... #8  => #5
... #8  => #6
... #9  => #7
... #10 => #8
... #9  => #9
... #10 => #10
... #11 => #11
... #12 => #12
... #13 <-- Second wave of spam
... #15
```

The point I'm hoping to make is that as long as we can set the count atomically at the beginning AND we are free to renumber the comments correctly from `#2`, it doesn't matter what comments come in while the script is running.  Assuming the intersection between "threads that have duplicate comment counts" and "threads that start at a weird comment count" is low (I think it is) we won't really be changing that many "working" URLs.

I've talked to Bert about this and he's on board with this renumbering so I'll code up the solution.

kvas-damian, line 198: @garthwebb Great idea. It's much simpler and causes lower number computations on MySql than my idea.

I'm glad you was determined to address my comment.

kvas-damian:
LGTM

PR #9114, opened 2015-11-24
ADEN-2621 Taboola targeting

fraszczakszymon:
Enabling taboola provider slots based on countries configuration and regular/recovery mode

### DIFF
--- a/extensions/wikia/AdEngine/AdEngine2Hooks.class.php
+++ b/extensions/wikia/AdEngine/AdEngine2Hooks.class.php
@@ -121,6 +121,7 @@ class AdEngine2Hooks {
 	public static function onOasisSkinAssetGroups( &$jsAssets ) {
 
 		global $wgAdDriverUseGoogleConsumerSurveys, $wgAdDriverUseTopInContentBoxad, $wgAdDriverUseTaboola;
+		$pageType = WikiaPageType::getPageType();
 
 		$jsAssets[] = self::ASSET_GROUP_ADENGINE_DESKTOP;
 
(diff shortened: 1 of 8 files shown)


extensions/wikia/AdEngine/AdEngine2Hooks.class.php
```diff
@@ -129,7 +130,7 @@ public static function onOasisSkinAssetGroups( &$jsAssets ) {
 			$jsAssets[] = self::ASSET_GROUP_LIFTIUM_EXTRA;
 		}
 
-		if ( $wgAdDriverUseGoogleConsumerSurveys && WikiaPageType::getPageType() === 'article' ) {
+		if ( $wgAdDriverUseGoogleConsumerSurveys && $pageType === 'article' ) {
```
nandy-andy: Is assigning `$pageType === 'article'` result to `$isPageAnArticle` variable and then reusing it in two conditional over-engineering? ;)

nandy-andy:
The code changes look good for me. I left only one minor note.

The only concern I have is that we're changing existing JavaScript asset: `taboola.js`. Let me think here aloud and don't hesitate if I'm wrong. Despite the fact that we bump our cachebuster with every and each release there might be pages cached on varnishes with current cachebuster and link to the old assets:
`<script src="http://slot1.images.wikia.nocookie.net/__am/1448319930/groups/-/oasis_shared_core_js,oasis_shared_js,oasis_anon_js,adengine2_oasis_exitstitial_js,adengine2_desktop_js,liftium_ads_js,liftium_ads_extra_js,adengine2_oasis_in_content_ads_js,adengine2_taboola_js,adengine2_interactive_maps_js,toc_js,qualaroo_js,spotlights_js"></script>`. But does it hurt us? Old assets work on production currently and we don't have any tight dependencies in the module being replaced. If a cached page loads old taboola then old taboola will execute and it will just work. The only risk is that as long as there are cached objects on varnishes with our old taboola (24h) we might see less data in our reports than after they gets invalidated. Therefore, I don't think we have to do here anything more :japanese_ogre:

nandy-andy:
LGTM :+1:

PR #9115, opened 2015-11-24
CE-3054 Add classification of the first-level templates as well

adamkarminski:
/cc: @SebastianMarzjan

### DIFF
--- a/extensions/wikia/TemplateClassification/maintenance/classifyNonArticleTemplates.class.php
+++ b/extensions/wikia/TemplateClassification/maintenance/classifyNonArticleTemplates.class.php
@@ -5,6 +5,7 @@ require_once( $mainDir . '/maintenance/Maintenance.php' );
 
 class ClassifyNonArticleTemplates extends Maintenance {
 
+	const TEMPLATE_TYPE_DIRECTLY_USED = 'directlyused';
 	const NONARTICLE_MAINTENANCE_PROVIDER = 'usage_classifier';
 
 	private $dryRun, $quiet, $logFile;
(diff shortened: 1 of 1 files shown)


extensions/wikia/TemplateClassification/maintenance/classifyNonArticleTemplates.class.php
```diff
@@ -82,20 +83,23 @@ public function execute() {
...
-							$wgCityId,
-							$templateId,
-							TemplateClassificationService::TEMPLATE_NOT_ART,
-							self::NONARTICLE_MAINTENANCE_PROVIDER,
-							$origin
-						);
-					} catch ( \Swagger\Client\ApiException $e ) {
-						$this->out( 'Classification failed!' );
-					}
+				$type = TemplateClassificationService::TEMPLATE_NOT_ART;
```
sebastianmarzjan, line 89: Why so?
Some templates are indeed works of art! :smiling_imp:

extensions/wikia/TemplateClassification/maintenance/classifyNonArticleTemplates.class.php
```diff
@@ -82,20 +83,23 @@ public function execute() {
...
+				try {
+					$tcs->classifyTemplate(
+						$wgCityId,
+						$templateId,
+						$type,
+						self::NONARTICLE_MAINTENANCE_PROVIDER,
+						$origin
+					);
+				} catch ( \Swagger\Client\ApiException $e ) {
+					$this->out( 'Classification failed!' );
```
sebastianmarzjan, line 102: Will you see it or maybe it's worth to log it to Kibana?

extensions/wikia/TemplateClassification/maintenance/classifyNonArticleTemplates.class.php
adamkarminski, line 89: More of them are a work of the devil! :imp:

extensions/wikia/TemplateClassification/maintenance/classifyNonArticleTemplates.class.php
lgarczewski, line 102: Meh, for a one time run like this I'd say just showing output is fine.

lgarczewski:
LGTM but:

Please test what the service returns with these data in, for both types! We don't want to screw up Mercury or Human TC display once this gets into the DB.

PR #9116, opened 2015-11-24
LOC-200 Translations of context-link and Special:Templates

adamkarminski:


kamilkoterba:
LGTM

PR #9118, opened 2015-11-24
CE-3049 made insights entry point into an actual link

lgarczewski:
@Wikia/community-engineering


kamilkoterba:
LGTM

extensions/wikia/TemplateClassification/templates/TemplateClassificationViewPageEntryPoint.mustache
```diff
@@ -1,3 +1,5 @@
 {{templateTypeLabel}}
-<span class="template-classification-type-text" data-type="{{templateType}}">{{templateTypeName}}</span>
-{{#editButton}}<a class="template-classification-edit sprite-small edit"></a>{{/editButton}}
+{{#editButton}}<a href="#" class="template-classification-edit" data-type="{{templateType}}">
+	{{templateTypeName}}<span class="template-classification-edit sprite-small edit"></span>
```
lukaszkonieczny: Please remove `template-classification-edit` class from `span` to avoid loading and opening modal twice which cause js error (because we bind this event to this class)

kamilkoterba:
this will brake unfortunately
https://github.com/Wikia/app/blob/CE-3049-htc-better-entry-links/extensions/wikia/TemplateClassification/scripts/TemplateClassificationModal.js#L196

extensions/wikia/TemplateClassification/scripts/TemplateClassificationModal.js
```diff
@@ -193,6 +193,7 @@ function ($, w, mw, loader, nirvana, tracker, labeling) {
 	function updateEntryPointLabel(templateType) {
 		$typeLabel
 			.data('type', mw.html.escape(templateType))
+			.children().first()
```
lukaszkonieczny: I think it's better to add class to first element and use `find('.className')` function here.

lgarczewski: If I'm reading this right, .children() is actually a lot faster: http://aahacreative.com/2010/07/19/jquery-fastest-method-find-descendents/

kamilkoterba: True. I'd use children() for performance and class selector to be specific.

lukaszkonieczny: Yep, agree with Kamil :+1:

kamilkoterba:
LGTM

PR #9120, opened 2015-11-24
CE-3019: Add support for imports from dev.wikia.com

lukaszkonieczny:
We want to move scripts on dev.wikia.com to MediaWiki namespace to handle scripts on all platform in the same way by using JS review tool. At the same time, we dont want to break existing imports from dev wiki. We also added fallback to Main namespace until transition is complete.

### DIFF
--- a/includes/wikia/resourceloader/ResourceLoaderGlobalWikiModule.class.php
+++ b/includes/wikia/resourceloader/ResourceLoaderGlobalWikiModule.class.php
@@ -103,11 +103,11 @@ abstract class ResourceLoaderGlobalWikiModule extends ResourceLoaderWikiModule {
 
 		$title = null;
 		$external = false;
-		$targetCityId = !empty( $options['city_id'] ) ? $options['city_id'] : 0;
+		$targetCityId = (int) $options['city_id'];
 
 		list( $text, $namespace ) = $this->parseTitle( $titleText );
 
-		if ( $targetCityId !== 0 && $wgCityId != $targetCityId && $text !== false ) {
+		if ( $targetCityId !== 0 && $wgCityId !== $targetCityId && $text !== false ) {
 			$external = true;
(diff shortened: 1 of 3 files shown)


includes/wikia/resourceloader/ResourceLoaderGlobalWikiModule.class.php
```diff
@@ -84,6 +90,57 @@ protected function createTitle( $titleText, $options = array() ) {
...
+	 */
+	private function createScriptTitle( $titleText, $options ) {
+		global $wgCityId;
+
+		$title = null;
+		$external = false;
+
+		list( $text, $namespace ) = $this->parseTitle( $titleText );
+
+		if ( !empty( $options['city_id'] ) && $wgCityId != $options['city_id'] && $text !== false ) {
```
adamkarminski: Since `$options['city_id']` is used a few times here, you could store it in a local variable and cast it to integer to avoid using soft `==` checks.

includes/wikia/resourceloader/ResourceLoaderGlobalWikiModule.class.php
```diff
@@ -84,6 +90,57 @@ protected function createTitle( $titleText, $options = array() ) {
...
+		} else {
+			$title = Title::newFromText( $text, NS_MEDIAWIKI );
+		}
+
+		// TODO: After scripts transition on dev wiki is done, remove this if statement (CE-3093)
+		if ( ( !$title || !$title->exists() ) && $options['city_id'] == Wikia\ContentReview\Helper::DEV_WIKI_ID ) {
+			$title = $this->devWikiFallback( $text, $external );
+		}
+
+		$title = $this->resolveRedirect($title);
```
adamkarminski: Spaces in the parentheses `( $title )`

adamkarminski:
LGTM :+1:

PR #9128, opened 2015-11-24
INT-271: Achievements: Remove Image Badges for Wookiee

mari-mcmurtrie:
Spec change for achievement feature.
1: Remove image badge from achievement for Wookiee. It used to be edit and adding image but now we remove image.
2: Include edit-category (special achievement).
3: Minor bug fix described in INT-270
Ticket: https://wikia-inc.atlassian.net/browse/INT-271


mari-mcmurtrie:
@Grunny Will you CR? This is an extent of the previous achievement change that I asked you to CR. Spec changed so I had to remove "add image" but include "edit-category". Thanks!!

owend:
Sure, this LGTM

PR #9135, opened 2015-11-25
CE-2820 Basic infrastructure for the Insights counters

adamkarminski:
Provides basic infrastructure for retrieving counts of items on the Insights lists.

### DIFF
--- a/extensions/wikia/Insights/services/InsightsCountService.class.php
+++ b/extensions/wikia/Insights/services/InsightsCountService.class.php
@@ -1,36 +1,59 @@
 <?php
+/**
+ * Class InsightsCountService
+ * Retrieves information on number of items on Insights lists.
+ */
 
 class InsightsCountService extends WikiaService {
 
+	/**
+	 * @var array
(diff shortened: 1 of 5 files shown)


extensions/wikia/Insights/services/InsightsCountService.class.php
```diff
@@ -0,0 +1,59 @@
...
+
+		return $response;
+	}
+
+	/**
+	 * @param string $type A type of an Insights list for which you'd like to calculate a count.
+	 * @return int|bool Returns false if a given class does not exist or the count otherwise.
+	 */
+	private function calculateCount( $type ) {
+		$className = InsightsHelper::getInsightsPages()[$type];
```
lukaszkonieczny, line 51: I think we should check `$excludedFromCount` here. So we can move this check from `getAllCounts`.

lukaszkonieczny:
LGTM

lukaszkonieczny:
This will be tested with features on which will be used

PR #9137, opened 2015-11-25
TransactionClassifier: introduce "page/blog"

macbre:
Make transactions reported to NewRelic (and other performance tools) more granular - report requests to blog namespaces as `page/blog` instead of `page/other` (which is responsible for ~9% of a total backend time of all requests).

@wladekb


wladekb:
lgtm

PR #9139, opened 2015-11-25
MAIN-5937 use getName instead mUsername in wasResetPassAuth

sqreek:
https://wikia-inc.atlassian.net/browse/MAIN-5937
authenticationCache in helios user use getName from user.
it was not consistent with check for wasResetPassAuth


wladekb:
lgtm

PR #9140, opened 2015-11-25
TransactionClassifier: report "list" query parameter of MW API requests

macbre:
Report `/api.php?action=query&list=users&ususers=macbre&usprop=groups|editcount|gender` MW API requests as `api/api/query/users` transaction for, you're right, better granularity. This specific request is used by permissions service, so let's keep an eye on its performance.

@wladekb / @jcellary


wladekb:
lgtm, sorry for the delay

PR #9146, opened 2015-11-25
MAIN-5942: Bandaid solution to problem with ContentReview extension

inez:
Apply the same if statement that is used in config to include ContentReview setup file.


gabrys:
LGTM

PR #9148, opened 2015-11-26
INT-269: Adding fake query parameter to ping page view to GA, so that we know …

mari-mcmurtrie:
…inter and intra searches separately.


mari-mcmurtrie:
@perjg this is Oasis one for INT-269. It is pretty much the same idea. Use qIntra for intra search, qInter for inter search. Will you CR? Thanks!

perjg:
LGTM

perjg:
@mari-mcmurtrie You should probably check why there is a unit test failure.

mari-mcmurtrie:
Thanks @perjg . The test failure was intermittent. I am pretty sure it was not related with my change.

PR #9149, opened 2015-11-26
This PR updates the script to work with current infrastructure

sebastianmarzjan:
- updated routing key
- added mandatory revisionId field

Additionally, I added retries as we sometimes lose connection to rabbit.

/cc @dianafa


dianafa:
strange spacing here and there but generally LGTM :dancers:

sebastianmarzjan:
(tested)

PR #9150, opened 2015-11-26
ADEN-2681 Exclude geo using 'non-' prefix

fraszczakszymon:
Changes:
- possibility to exclude region/country/continent from countryList using `non-` prefix
- fixed code style errors

### DIFF
--- a/resources/wikia/modules/geo.js
+++ b/resources/wikia/modules/geo.js
@@ -85,7 +85,12 @@
 		 * @returns {boolean}
 		 */
 		function isProperCountry(countryList) {
-			return !!(countryList && countryList.indexOf && countryList.indexOf(getCountryCode()) > -1);
+			return !!(
+				countryList &&
+				countryList.indexOf &&
+				countryList.indexOf(getCountryCode()) > -1 &&
+				countryList.indexOf('non-' + getCountryCode()) === -1
+			);
 		}
 
 		/**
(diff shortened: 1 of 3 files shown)


resources/wikia/modules/geo.js
```diff
@@ -85,7 +85,12 @@
 		 * @returns {boolean}
 		 */
 		function isProperCountry(countryList) {
-			return !!(countryList && countryList.indexOf && countryList.indexOf(getCountryCode()) > -1);
+			return !!(
+				countryList &&
+				countryList.indexOf &&
+				countryList.indexOf(getCountryCode()) > -1 &&
+				countryList.indexOf('non-' + getCountryCode()) === -1
```
nandy-andy: What do you think about keeping `'non-'` in a `negativePrefix` variable defined at the top together with `cookieName` and `geoData`? Will it make code less readable?

resources/wikia/modules/spec/geo.spec.js
```diff
@@ -29,46 +29,60 @@ describe("Geo", function () {
...
+		expect(geo.isProperContinent(['XX', 'non-XX-EU'])).toBeFalsy();
 	});
 
-	it('isProperGeo test', function() {
+	it('isProperGeo test', function () {
 		expect(geo.isProperGeo(['PL'])).toBeTruthy();
 		expect(geo.isProperGeo(['XX-EU'])).toBeTruthy();
 		expect(geo.isProperGeo(['XX'])).toBeTruthy();
 		expect(geo.isProperGeo(['PL-72'])).toBeTruthy();
+		expect(geo.isProperGeo(['non-PL'])).toBeFalsy();
```
nandy-andy, line 63: As discussed offline: let's make only `isProperGeo` public and move important tests from above here.

jacekjursza:
LGTM

PR #9151, opened 2015-11-26
Check if wiki should load safe js pages

lukaszkonieczny:
\cc @Wikia/community-engineering @inez


adamkarminski:
LGTM

inez:
LGTM :) I looked up isUsingSafeJs and it is actually exactly same condition :)

PR #9154, opened 2015-11-26
ER-9507: add an assert on Parser::mSubstWords instead of throwing a fatal

macbre:
[ER-9507](https://wikia-inc.atlassian.net/browse/ER-9507)

Report a backtrace instead of throwing a fatal:

```
PHP Fatal Error: Call to a member function matchStartAndRemove() on null in /includes/parser/Parser.php on line 3391
```

@michalroszka


macbre:
Failed unit tests are unrelated

PR #9157, opened 2015-11-26
CE-3084: Remove loop notification from Insights pages without infoboxes list

lukaszkonieczny:
Remove loop notification from Insights pages without infoboxes list


adamkarminski:
LGTM

PR #9158, opened 2015-11-26
CE-3084: Remove loop notification from Insights pages without infoboxes

lukaszkonieczny:
Remove loop notification from Insights pages without infoboxes


adamkarminski:
LGTM

PR #9164, opened 2015-11-26
Ce 3107 restrict mw ns move

kamilkoterba:
https://wikia-inc.atlassian.net/browse/CE-3107

Moving into within and out of MW namespace for whitelisted pages was allowed.
This is restricting moves for whitelisted pages as well.

Also did some refactoring of 'move' error messages to be more verbose and display allowed groups as well and in same way changed check for 'undelete' to check for 'editinterfacetrusted' right and reuse message for 'undelete' that already exists.

@Wikia/community-engineering

### DIFF
--- a/includes/Exception.php
+++ b/includes/Exception.php
@@ -374,24 +374,31 @@ class PermissionsError extends ErrorPageError {
 	public $permission, $errors;
 
 	function __construct( $permission, $errors = array() ) {
-		global $wgLang;
-
 		$this->permission = $permission;
 
 		if ( !count( $errors ) ) {
-			$groups = array_map(
-				array( 'User', 'makeGroupLinkWiki' ),
-				User::getGroupsWithPermission( $this->permission )
-			);
-
-			if ( $groups ) {
-				$errors[] = array( 'badaccess-groups', $wgLang->commaList( $groups ), count( $groups ) );
-			} else {
-				$errors[] = array( 'badaccess-group0' );
-			}
+			$errors[] = self::prepareBadAccessErrorArray( $this->permission );
 		}
 
 		$this->errors = $errors;
+	}
+
+	/**
+	 * @param string $permission Name of action
+	 * @return array error message key and message params (optionally)
+	 */
+	public static function prepareBadAccessErrorArray( $permission ) {
+		global $wgLang;
+		$groups = array_map(
+			[ 'User', 'makeGroupLinkWiki' ],
+			User::getGroupsWithPermission( $permission )
+		);
+
+		if ( $groups ) {
+			return [ 'badaccess-groups', $wgLang->commaList( $groups ), count( $groups ) ];
+		}
+
+		return [ 'badaccess-group0' ];
 	}
 
 	function report() {
--- a/includes/specials/SpecialUndelete.php
+++ b/includes/specials/SpecialUndelete.php
@@ -328,9 +328,9 @@ class PageArchive {
 	function undelete( $timestamps, $comment = '', $fileVersions = array(), $unsuppress = false ) {
 		global $wgUser;
 
-		$resultMock = '';
-		if ( !$this->title->userCan( 'undelete' ) ) {
-			throw new PermissionsError( 'editinterfacetrusted' );
+		$permErrors = $this->title->getUserPermissionsErrors( 'undelete', $wgUser );
+		if ( count( $permErrors ) ) {
+			throw new PermissionsError( 'undelete', $permErrors );
 		}
 
 		// If both the set of text revisions and file revisions are empty,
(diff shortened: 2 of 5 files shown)


includes/wikia/Wikia.php
```diff
@@ -30,6 +30,8 @@
 $wgHooks['AjaxAddScript']            [] = 'Wikia::onAjaxAddScript';
 $wgHooks['TitleGetSquidURLs']        [] = 'Wikia::onTitleGetSquidURLs';
 $wgHooks['userCan']                  [] = 'Wikia::canEditInterfaceWhitelist';
+$wgHooks['getUserPermissionsErrors'] [] = 'Wikia::canMoveMediaWikiNS';
+$wgHooks['getUserPermissionsErrors'] [] = 'Wikia::canUndeleteMediaWikiNS';
```
lgarczewski: These two could be handled in one hook since the code of the hook functions is nearly identical.

includes/Exception.php
```diff
@@ -374,26 +374,33 @@ class PermissionsError extends ErrorPageError {
...
 		}
 
 		$this->errors = $errors;
 	}
 
+	/**
+	 * @param string $permission Name of action
+	 * @return array error message key and message params (optionally)
+	 */
+	public static function prepareBadAccessErrorArray( $permission ) {
```
lgarczewski, line 390: Why has this been moved out of the constructor? Did I miss this being used somewhere else? Seems to still just be used in one place -- but then what is the purpose of this change?

includes/specials/SpecialUndelete.php
```diff
@@ -328,9 +328,9 @@ function isDeleted() {
 	function undelete( $timestamps, $comment = '', $fileVersions = array(), $unsuppress = false ) {
 		global $wgUser;
 
-		$resultMock = '';
-		if ( !$this->title->userCan( 'undelete' ) ) {
```
lgarczewski, line 332: Could you explain why the change from Title::userCan() to Title::getPermissionsErrors()? Could we not have used the hook in Title::userCan like we do with Wikia::canEditInterfaceWhitelist?

kamilkoterba, line 332: getPermissionsErrors allows us to pass errors returned from hooks to user, while userCan calls same logic but maps result to boolean saying is there where any errors.
So in general if we wany to just check permissions we chould use userCan but if we want to throw it we should check using getPermissionsErrors

includes/Exception.php
adamkarminski, line 390: It will allow us reusing just this bit without duplicating the logic. I think that in the context of all changes this one makes sense.

includes/wikia/Wikia.php
adamkarminski: Agreed.

includes/specials/SpecialUndelete.php
adamkarminski, line 332: I'm with Kamil on this one. This is a very strange and bad UX not to give users a reason for which you blocked an action for them.

includes/wikia/Wikia.php
```diff
@@ -2331,16 +2333,26 @@ static function onLocalFilePurgeThumbnailsUrls( LocalFile $file, Array &$urls )
 	 * usually return true to allow processing other hooks
 	 * return false stops permissions processing and we are totally decided (nothing later can override)
 	 */
-	static function canEditInterfaceWhitelist (&$title, &$wgUser, $action, &$result) {
+	static function canEditInterfaceWhitelist ( &$title, &$wgUser, $action, &$result ) {
 		global $wgEditInterfaceWhitelist;
 
+		// Allowed actions at this point
+		$notRelevantActions = [
```
adamkarminski: This name is **very** misleading

includes/wikia/Wikia.php
```diff
@@ -2331,16 +2333,26 @@ static function onLocalFilePurgeThumbnailsUrls( LocalFile $file, Array &$urls )
...
+			'read',
+			'move', // Is being checked in next hook canMoveMediaWikiNS
+			'undelete' // Is being checked in next hook canUndeleteMediaWikiNS
+		];
+
 		// List the conditions we don't care about for early exit
-		if ( $action == "read" || $title->getNamespace() != NS_MEDIAWIKI || empty( $wgEditInterfaceWhitelist )) {
+		if ( in_array( $action, $notRelevantActions )
+			|| $title->getNamespace() != NS_MEDIAWIKI
```
adamkarminski, line 2347: `!$title->inNamespace( NS_MEDIAWIKI )`

adamkarminski:
Address the comments and it will LGTM. :)

PR #9174, opened 2015-11-30
Content Review translations

lukaszkonieczny:
Content Review translations


kamilkoterba:
LGTM 
Although I advice to remove NL translations as these duplicate EN and can cause a trouble once EN is updated.

lukaszkonieczny:
@kamilkoterba I think we should fix issue with NL translations in crowdin because in next download it will appear again.

PR #9180, opened 2015-11-30
ADEN-2218 First run non-turtle slots with DirectGpt

fraszczakszymon:
Slots not supported by turtle provider (e.g. prefooters) were filled in by remnant provider instead of direct and here is the fix.


nandy-andy:
LGTM :+1:

PR #9181, opened 2015-11-30
CE-3014 Make all special pages related to infoboxes use TCS as a base

adamkarminski:
Requirements:
1. Special:Templates - when filtered by the "Infobox" type it displays all templates classified as Infoboxes by users or machine, either portable or non-portable,
2. Special:Insights/pageswithoutinfobox - displays all pages that do not include a template classified as Infobox by a user or machine, either portable or non-portable,
3. Special:Insights/nonportableinfoboxes - The important part here is that this list reads from Template classification and not the existing title/class/category logic we have now.


extensions/wikia/Insights/specials/UnconvertedInfoboxesPage.class.php
```diff
@@ -95,27 +95,32 @@ public function recache( $limit = false, $ignoreErrors = true ) {
...
+		$nonportableInfoboxes = [];
+
+		foreach ( $recognizedTemplates as $templateId => $type ) {
+			if ( $tcs->isInfoboxType( $type ) ) {
+				$title = Title::newFromID( $templateId );
+				if ( !$title instanceof Title ) {
+					continue;
+				}
+
+				if ( empty( PortableInfoboxDataService::newFromTitle( $title )->getData() ) ) {
```
lukaszkonieczny: Maybe we can combine these two if statements and check `$title instanceof Title && empty( PortableInfoboxDataService::newFromTitle( $title )->getData() )` ?

extensions/wikia/TemplateClassification/services/UserTemplateClassificationService.class.php
```diff
@@ -35,12 +44,37 @@ class UserTemplateClassificationService extends TemplateClassificationService {
 	 * @throws \Swagger\Client\ApiException
 	 */
 	public function getType( $wikiId, $pageId ) {
-		$templateType = parent::getType( $wikiId, $pageId );
+		return $this->getMappedType( parent::getType( $wikiId, $pageId ) );
```
lukaszkonieczny: What do you think about changing name to something like `getUserType` because not all types are mapped there

adamkarminski: Since you call `UserTemplateClassificationService` anyway I think that `getMappedType` reflects the purpose of this method better. :)

extensions/wikia/TemplateClassification/services/UserTemplateClassificationService.class.php
```diff
@@ -35,12 +44,37 @@ class UserTemplateClassificationService extends TemplateClassificationService {
...
+	}
+
+	private function getMappedType( $type ) {
+		if ( $this->isInfoboxType( $type ) ) {
+			return self::TEMPLATE_INFOBOX;
+		}
 
-		if ( !in_array( $templateType, self::$templateTypes ) ) {
-			$templateType = self::TEMPLATE_UNCLASSIFIED;
+		if ( !in_array( $type, self::$templateTypes ) ) {
```
kamilkoterba, line 74: what about reusing isUserType here

extensions/wikia/Insights/specials/UnconvertedInfoboxesPage.class.php
```diff
@@ -95,27 +95,30 @@ public function recache( $limit = false, $ignoreErrors = true ) {
...
+					&& empty( PortableInfoboxDataService::newFromTitle( $title )->getData() )
+				) {
 					$links = $title->getIndirectLinks();
-					$nonportableTemplates[] = [
+					$nonportableInfoboxes[] = [
 						$this->getName(),
 						count( $links ),
 						NS_TEMPLATE,
-						$row->title,
+						$title->getText(),
```
kamilkoterba: I think getDBkey is what you want here

extensions/wikia/Insights/specials/UnconvertedInfoboxesPage.class.php
```diff
@@ -95,27 +95,30 @@ public function recache( $limit = false, $ignoreErrors = true ) {
 	 * @return bool|mixed
 	 */
 	public function reallyDoQuery( $limit = false, $offset = false ) {
-		$dbr = wfGetDB( DB_SLAVE, [ $this->getName(), __METHOD__, 'vslow' ] );
-
-		$nonportableTemplates = ( new WikiaSQL() )
-			->SELECT( 'page_title' )->AS_( 'title' )
-			->FROM( 'page' )
-			->WHERE( 'page_namespace' )->EQUAL_TO( NS_TEMPLATE )
-				->AND_( 'page_is_redirect' )->EQUAL_TO( 0 )
```
kamilkoterba, line 104: Do we skip this condition on purpose?

extensions/wikia/TemplateClassification/specials/TemplatesSpecialController.class.php
```diff
@@ -340,6 +340,6 @@ private function getDefaultType() {
 	 * @return bool
 	 */
 	private function isUserType( $type ) {
```
kamilkoterba, line 342: Can we move this method into UserTemplateClassificationService and make UserTemplateClassificationService::$templateTypes private in return?

extensions/wikia/Insights/specials/UnconvertedInfoboxesPage.class.php
```diff
@@ -95,27 +95,30 @@ public function recache( $limit = false, $ignoreErrors = true ) {
...
+		$tcs = new UserTemplateClassificationService();
+		$recognizedTemplates = $tcs->getTemplatesOnWiki( $wgCityId );
+
+		$nonportableInfoboxes = [];
+
+		foreach ( $recognizedTemplates as $templateId => $type ) {
+			if ( $tcs->isInfoboxType( $type ) ) {
+				$title = Title::newFromID( $templateId );
+				if ( $title instanceof Title
+					&& empty( PortableInfoboxDataService::newFromTitle( $title )->getData() )
```
kamilkoterba, line 109: We have just last use of PortableInfoboxClassification::isTitleWithNonportableInfobox in our code which is in TemplateDraftHooks::onEditPageLayoutShowIntro. Can we unify this as well and remove isTitleWithNonportableInfobox? I think it's worth to unify it as well because TemplateDtaft is part of Insights loop.

extensions/wikia/TemplateClassification/services/UserTemplateClassificationService.class.php
kamilkoterba: I would even go with `mapType` as it's private method that does mapping for us.
getType and getMappedType may lead to question what's the difference and why one is private, what requires to read the code to make sure it's actually same.
but in general getMappedType is fine for me as well.

extensions/wikia/TemplateClassification/services/UserTemplateClassificationService.class.php
adamkarminski, line 74: Definitely not in the current setup. The last thing we want is to make the service, which has a higher "hierarchy", to use methods from a Special Page controller. We could move `isUserType` to the service itself, but let's not mix different scopes here. It does not bother me that much since the method itself is just this `in_array()` call.

extensions/wikia/Insights/specials/UnconvertedInfoboxesPage.class.php
adamkarminski, line 104: What do you mean? We now use TCS data as a base for the operations so I cannot say we 'skip' it.

extensions/wikia/Insights/specials/UnconvertedInfoboxesPage.class.php
adamkarminski: :+1:

extensions/wikia/TemplateClassification/specials/TemplatesSpecialController.class.php
adamkarminski, line 342: See the comment above.

extensions/wikia/Insights/specials/UnconvertedInfoboxesPage.class.php
adamkarminski, line 109: I've removed the PortableInfoboxClassification class entirely.

extensions/wikia/Insights/specials/UnconvertedInfoboxesPage.class.php
adamkarminski, line 104: Ok, check readded: https://github.com/Wikia/app/commit/a220f8c63e719ff9c54b0936fb4c50036fa84ba6

kamilkoterba:
LGTM

azarek:
Testing done! LGTM

PR #9182, opened 2015-11-30
VE-2094: Log to 3rd party logging solution if something went terribly wrong

inez:


extensions/wikia/Track/Track.php
```diff
@@ -65,6 +65,7 @@ private static function getViewJS ($param=null) {
 		window.beacon_id = result[1];
 	} else {
 		// something went terribly wrong
+		document.write('<img src="http://logs-01.loggly.com/inputs/88a88e56-77c6-49cc-af41-6f44f83fe7fe.gif?message=something%20went%20terribly%20wrong"/>');
```
hakubo: A. not something but wikia_beacon_id is missing - could we write better message for it?
B. why not log it to our infrastructure? and make it possible to query for it in kibana?

inez: For now I'm just logging the same message that was in the comment above. I'm logging it to loggly because that is quickest and most reliable solution to implement this - if we find out that this problem really happens - then we will have bigger discussion what we should do about it. I'm coming from: any logging for it is better than no logging at all.

hakubo: @jcellary - do we have something similar in our infrastructure so we could send simple logs from anywhere? sounds like something useful to have.

I get the point of making it 'quick and dirty' it's just changing something%20went%20terribly%20wrong to wikia_beacon_id is empty 
might save you time later figuring out what went so terribly wrong :)

inez:
@hakubo Yup, I changed the message. Feel free to merge if looks good to you.

ljagiello:
Could you please use internal logging platform instead of introducing another endpoint?

inez:
@ljagiello Please take a look at the associated ticket. The goal here it to quickly find out if that problem exists (if there is a problem retrieving a beacon) and if there is then some more solid solution will be worked on - otherwise this code will be just removed. Using loggly in this case is the fastest and most reliable way to achieve it.

owend:
Yeah, I'm fine with this as it's temporary (and we think about setting up a better js logging system long term)

PR #9188, opened 2015-12-01
CE-3075 Add proper theming to a success banner on classification

adamkarminski:


kamilkoterba:
LGTM

PR #9190, opened 2015-12-01
CE-3034: Posibility to add custom badges to users without review process

lukaszkonieczny:
To add custom badge, user needs to import script from dev.wikia.com/wiki/UserBadges.js (https://gist.github.com/lukaszkonieczny/c9cb046a7b0c5819a07d) and add pairs containing user name and rights on local wiki in MediaWiki:UserBadges article.

Do not merge without https://github.com/Wikia/config/pull/1495

### DIFF
--- a/extensions/wikia/ContentReview/ContentReview.i18n.php
+++ b/extensions/wikia/ContentReview/ContentReview.i18n.php
@@ -49,8 +49,14 @@ The recently submitted change to this JavaScript page (revision [$2 $3]) was rej
 Names should not contain the MediaWiki namespace prefix. Write each script on a new line. See [[Help:Including additional CSS and JS]] for more information.
 ----
 ',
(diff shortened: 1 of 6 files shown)


extensions/wikia/ContentReview/ContentReview.i18n.php
```diff
@@ -49,6 +49,10 @@
 Names should not contain the MediaWiki namespace prefix. Write each script on a new line. See [[Help:Including additional CSS and JS]] for more information.
 ----
 ',
+	'content-review-user-badges-description' => 'Here, you can easily add custom badges to users. Just separate user name and rights by colon. If there is more rights list them using comma,
```
Grunny: - `by colon` -> `by a colon`
- `If there is more rights list them using comma` -> `If there are more rights, list them using comma`

I'd also consider consistently saying "badge" rather than "rights" since these aren't actually rights.

Grunny:
Other than wording tweaks, LGTM.

kamilkoterba:
Tested. LGTM

PR #9192, opened 2015-12-01
CE-3127: Fix serving scripts from user namespace

lukaszkonieczny:
Fix serving scripts from user namespace


lukaszkonieczny:
CR in https://github.com/Wikia/app/pull/9193

PR #9196, opened 2015-12-01
Hide ads when ChangePay triggers paid event

BladeBronson:
I'm not sure if Wikia has another method to hide ads, including site skins that aren't matched by this simple jQuery selector, but please merge this if it'll help move you forward.

### DIFF
--- a/extensions/wikia/AdEngine/AdEngine2Hooks.class.php
+++ b/extensions/wikia/AdEngine/AdEngine2Hooks.class.php
@@ -322,6 +322,7 @@ changePay.init({
         console.log("canonical is: " + changePay.get_canonical_url());
     },
     paid: function(status_code, msg) {
+	$('wikia-ad').hide();
         console.log(status_code, "You are paid up, or no payment is required!");
     },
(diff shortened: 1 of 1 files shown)


extensions/wikia/AdEngine/AdEngine2Hooks.class.php
```diff
@@ -322,6 +322,7 @@ public static function onSkinAfterBottomScripts( $skin, &$bottomScriptsText ) {
         console.log("canonical is: " + changePay.get_canonical_url());
     },
     paid: function(status_code, msg) {
+	$('wikia-ad').hide();
```
inez, line 325: We don't use jQuery anymore. It's a YUI house now.

jacekjursza, line 325: @BladeBronson I'm going to close this PR since the change is inside the code that is moved to changePay-wikia.js 
BTW you missed "dot" in the CSS selector, it should be:
$('.wikia-ad').hide();

PR #9197, opened 2015-12-01
VE-2100: Be very strict about types

inez:


inez:
Dude, this is pretty bad commit message.

PR #9204, opened 2015-12-02
DatabaseBase: improve reporting of writing queries issued during read-only

macbre:
Or performed in Reston. An example of GET requests trying to perform DB writes in Reston - [VE-1956](https://wikia-inc.atlassian.net/browse/VE-1956)

`WikiaException` will cause Jira Reporter to group the tickets by the `$fname`.

@michalroszka

### DIFF
--- a/includes/db/Database.php
+++ b/includes/db/Database.php
@@ -905,7 +905,8 @@ abstract class DatabaseBase implements DatabaseType {
 				wfProfileOut( $totalProf );
 			}
 			WikiaLogger::instance()->error( 'DB readonly mode', [
-				'exception' => new Exception( $sql ),
+				'exception' => new WikiaException( $fname . ' called in read-only mode' ),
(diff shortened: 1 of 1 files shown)


includes/db/Database.php
```diff
@@ -905,7 +905,8 @@ public function query( $sql, $fname = '', $tempIgnore = false ) {
 				wfProfileOut( $totalProf );
 			}
 			WikiaLogger::instance()->error( 'DB readonly mode', [
-				'exception' => new Exception( $sql ),
+				'exception' => new WikiaException( $fname . ' called in read-only mode' ),
```
wladekb: what's the difference between `Exception` and `WikiaException` ?

wladekb: Just looked at the PR description. How is that achieved?

wladekb:
lgtm

PR #9205, opened 2015-12-02
ADEN-2687 Restrict Japan leaderboards' sizes

fraszczakszymon:
Remove all sizes except 728x90 from leaderboards in Japan

### DIFF
--- a/extensions/wikia/AdEngine/js/provider/directGpt.js
+++ b/extensions/wikia/AdEngine/js/provider/directGpt.js
@@ -16,13 +16,22 @@ define('ext.wikia.adEngine.provider.directGpt', [
 		btfQueue = [],
 		btfQueueStarted = false,
 		slotMap = {
-			CORP_TOP_LEADERBOARD:       {size: '728x90,1030x130,1030x65,1030x250,970x365,970x250,970x90,970x66,970x180,980x150', loc: 'top'},
+			CORP_TOP_LEADERBOARD:       {
+				size: '728x90,1030x130,1030x65,1030x250,970x365,970x250,970x90,970x66,970x180,980x150',
+				loc: 'top'
+			},
 			CORP_TOP_RIGHT_BOXAD:       {size: '300x250,300x600,300x1050', loc: 'top'},
 			EXIT_STITIAL_BOXAD_1:       {size: '300x250,600x400,800x450,550x480', loc: 'exit'},
 			GPT_FLUSH:                  {flushOnly: true},
-			HOME_TOP_LEADERBOARD:       {size: '728x90,1030x130,1030x65,1030x250,970x365,970x250,970x90,970x66,970x180,980x150', loc: 'top'},
+			HOME_TOP_LEADERBOARD:       {
+				size: '728x90,1030x130,1030x65,1030x250,970x365,970x250,970x90,970x66,970x180,980x150',
+				loc: 'top'
+			},
 			HOME_TOP_RIGHT_BOXAD:       {size: '300x250,300x600,300x1050', loc: 'top'},
-			HUB_TOP_LEADERBOARD:        {size: '728x90,1030x130,1030x65,1030x250,970x365,970x250,970x90,970x66,970x180,980x150', loc: 'top'},
+			HUB_TOP_LEADERBOARD:        {
+				size: '728x90,1030x130,1030x65,1030x250,970x365,970x250,970x90,970x66,970x180,980x150',
+				loc: 'top'
+			},
 			INCONTENT_1A:               {size: '300x250', loc: 'middle', pos: 'incontent_1'},
 			INCONTENT_1B:               {size: '300x250,160x600', loc: 'middle', pos: 'incontent_1'},
 			INCONTENT_1C:               {size: '300x250,160x600,300x600', loc: 'middle', pos: 'incontent_1'},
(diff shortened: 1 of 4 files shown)


extensions/wikia/AdEngine/js/provider/directGpt.js
```diff
@@ -16,13 +16,22 @@ define('ext.wikia.adEngine.provider.directGpt', [
 		btfQueue = [],
 		btfQueueStarted = false,
 		slotMap = {
-			CORP_TOP_LEADERBOARD:       {size: '728x90,1030x130,1030x65,1030x250,970x365,970x250,970x90,970x66,970x180,980x150', loc: 'top'},
+			CORP_TOP_LEADERBOARD:       {
+				size: '728x90,1030x130,1030x65,1030x250,970x365,970x250,970x90,970x66,970x180,980x150',
```
drets: :+1:

extensions/wikia/AdEngine/js/provider/directGpt.js
```diff
@@ -16,13 +16,22 @@ define('ext.wikia.adEngine.provider.directGpt', [
 		btfQueue = [],
 		btfQueueStarted = false,
 		slotMap = {
-			CORP_TOP_LEADERBOARD:       {size: '728x90,1030x130,1030x65,1030x250,970x365,970x250,970x90,970x66,970x180,980x150', loc: 'top'},
+			CORP_TOP_LEADERBOARD:       {
```
drets, line 19: :+1:

nandy-andy, line 19: formatting nazi ;-) j/k :+1:

nandy-andy:
LGTM :+1:

drets:
LGTM :+1:

extensions/wikia/AdEngine/js/provider/factoryWikiaGpt.js
```diff
@@ -1,11 +1,22 @@
...
 	'ext.wikia.adEngine.provider.gpt.helper',
+	'wikia.geo',
+	'wikia.log',
 	require.optional('ext.wikia.adEngine.lookup.services')
-], function (log, adLogicPageParams, gptHelper, lookups) {
+], function (adLogicPageParams, gptHelper, geo, log, lookups) {
 	'use strict';
+	var country = geo.getCountryCode();
+
+	function overrideSizes(slotMap, newSizes) {
```
nandy-andy, line 12: What do you think about returning a new slot map instead of changing the passed object?

fraszczakszymon, line 12: Then it will require a several duplicates (for other slots) in both providers

extensions/wikia/AdEngine/js/provider/factoryWikiaGpt.js
```diff
@@ -24,6 +35,10 @@ define('ext.wikia.adEngine.provider.factory.wikiaGpt', [
 	function createProvider(logGroup, providerName, src, slotMap, extra) {
 		extra = extra || {};
 
+		if (extra.overrideSizesPerCountry && extra.overrideSizesPerCountry[country]) {
+			overrideSizes(slotMap, extra.overrideSizesPerCountry[country]);
```
nandy-andy, line 39: Another nitpicking nAndy: how about moving the `overrideSizes()` method to GPT helper? This way we won't introduce two more dependencies here (just one more - `geo` in the helper) and it feels more like this kind of function isn't a part of factory. But I might be missing a point and you don't have to agree. Just tell me what do you think about it :japanese_ogre: (since creating helpers and utils is also quite a bad habit to some degree :D)

extensions/wikia/AdEngine/js/provider/factoryWikiaGpt.js
drets, line 12: :+1:

extensions/wikia/AdEngine/js/provider/factoryWikiaGpt.js
fraszczakszymon, line 39: You're totally right - `helper` doesn't tell anything, for sure it's worth refactoring in "some time". What about `overrideSizes`? I thought in this way:
FactoryWikiaGpt creates provider with specific slot map (which defines list of supported slots). It also has only one additional dependency (geo). I just reordered dependencies in alphabetical way. GPT helper is the last point of our integration between AdEngine and googletag. That's why decided to place it in factory.

Maybe it's worth to call for another pair of eyes :D @jacekjursza what do you think?

jacekjursza, line 39: I'm not a big fan of helpers too (it's like a bag of code that belongs nowhere and we probably shouldn't overuse it). :+1: 
In this case, the log object doesn't hurt my eyes (especially we pass it all around anyway). 
I'm more concern about mixing 2 different configuration approaches inside one function: via parameters (logGroup, providerName, src, slotMap) and via object (extra). Maybe it would be worth to unify it?
Using object approach is less convenient  while writing the code, but more convenient while reading it. 
I'm thinking about something like:

``` javascript
function createProvider(config) { ... }
```

and then using it like:

``` javascript
provider = factory.createProvider({
  'logGroup': logGroup,
  'slotName': 'DirectGpt',
  ...
});
```

what do you think?

fraszczakszymon, line 39: I think the initial idea was to make some params required and optional (`extra`). I agree that it's worth to rethink this design, but not sure if in scope of this ticket - it's a bit sensitive area and it's Q4 :)

jacekjursza, line 39: LGTM

nandy-andy:
Thanks for the discussion: LGTM :+1:

PR #9206, opened 2015-12-02
 CE-3073: Verify template type classified by user

lukaszkonieczny:
- move hook to classifying method

/cc @Wikia/community-engineering @kamilkoterba @SebastianMarzjan


extensions/wikia/TemplateClassification/services/UserTemplateClassificationService.class.php
```diff
@@ -43,4 +50,23 @@ public function getType( $wikiId, $pageId ) {
...
+	 *
+	 * @param int $wikiId
+	 * @param int $pageId
+	 * @param string $templateType
+	 * @param string $origin
+	 * @throws BadRequestApiException
+	 */
+	public function classifyTemplate( $wikiId, $pageId, $templateType, $origin ) {
+		if ( !in_array( $templateType, self::$templateTypes ) ) {
+			throw new ApiException( self::CLASSIFY_TEMPLATE_EXCEPTION_MESSAGE, self::CLASSIFY_TEMPLATE_EXCEPTION_CODE );
```
sebastianmarzjan: You may consider adding the passed $templateType to exception message

extensions/wikia/TemplateClassification/services/UserTemplateClassificationService.class.php
```diff
@@ -43,4 +50,23 @@ public function getType( $wikiId, $pageId ) {
...
+		}
+
+		parent::classifyTemplate( $wikiId, $pageId, $templateType, self::USER_PROVIDER, $origin );
+
+		wfRunHooks( 'UserTemplateClassification::TemplateClassified', [ $wikiId, $pageId, $templateType ] );
```
sebastianmarzjan: In this flow, 
https://github.com/Wikia/app/pull/9206/files#diff-51080aaef54b01dd7a86c59c827a2761R124
and
https://github.com/Wikia/app/pull/9206/files#diff-2e8dea429e8ffb59895bd46f92ed5942R139
both need to build the title object.
What is the reason behind passing $pageId instead of a valid Title?

kamilkoterba:
LGTM, but +1 to Sebastian question about title.

extensions/wikia/TemplateClassification/services/UserTemplateClassificationService.class.php
lukaszkonieczny: My assumption here is if we don't need Title instance in this method we shouldn't create it here. By passing `$pageId` you are able to create Title wherever you want. Also after removing methods from this hook we will not have leftovers like creating unnecessary Title instance on every classification.

kamilkoterba: From performance perspective we can have a gain if we create it once before hook as each call of Title::newFromID generates DB query
https://github.com/Wikia/app/blob/CE-3110/includes/Title.php#L226-226
plus there's small possibility of having all those hooks ever removed...

lukaszkonieczny: Ok, I missed part with query database. Make sense :+1:

kamilkoterba:
LGTM

PR #9213, opened 2015-12-02
SOC-1741 None of this CSS is necessary

garthwebb:
It was an earlier attempt at styling that was abandoned for existing classes

https://wikia-inc.atlassian.net/browse/SOC-1741


extensions/wikia/AssetsManager/config.php
```diff
@@ -2539,13 +2539,6 @@
 	]
 ];
 
-$config['special_user_activity_css'] = [
```
jsutterfield, line 2542: We should remove the call from the controller to load this as well: https://github.com/Wikia/app/blob/dev/extensions/wikia/UserActivity/SpecialUserActivityController.class.php#L44-L49

jsutterfield:
LGTM

PR #9219, opened 2015-12-03
CE-3120 Hide Navigation templates with block elements on mercury

adamkarminski:


Grunny:
This should have unit tests added.

adamkarminski:
@Grunny - It does now. :)

Grunny:
LGTM in general. It's an edge case, but it's possible there may be unintentional false positives with the string matching method used. For example, we have a few parser tags that would match `<p` such as `<poem` (though that is kind of a block in itself). We may want to consider tightening it up a little with a regex.

Grunny:
:+1:

azarek:
Testing done! LGTM!

PR #9222, opened 2015-12-03
CE-3124: Publish template after forced modal

lukaszkonieczny:
Publish template after forced modal


adamkarminski:
LGTM, since `$('#wpSave').click()` is a pattern that we use for publishing from preview as well. :)

kamilkoterba:
Probably $('#editform').submit() is better idea.

kamilkoterba:
In general that should work although I don't feel like publishing is responsibility of TemplateClassificationModal.js

Plus it won't work for ACE editor which is not shown for unclassified templates right now but it's something we should be aware. 

(edit)
Actually just found out that $('#wpSave') button is common for ACE and Source so it will work for both. Let's leave it like this I suggest.

kamilkoterba:
LGTM

azarek:
Testing done! LGTM

adamkarminski:
@lukaszkonieczny @kamilkoterba - LGTM, but please resolve conflicts before merge.

PR #9226, opened 2015-12-04
ADEN-2728 Recovered message enabled only on article page types

fraszczakszymon:
Restrict places where recovered message can be displayed - now only on pages with `article` type.


drets:
LGTM :+1:

jacekjursza:
LGTM

PR #9229, opened 2015-12-04
CE-3115 Scroll TC modal to top if a preselected type is Unknown

adamkarminski:

### DIFF
--- a/extensions/wikia/TemplateClassification/scripts/TemplateClassificationModal.js
+++ b/extensions/wikia/TemplateClassification/scripts/TemplateClassificationModal.js
@@ -92,8 +92,8 @@ function ($, w, mw, loader, nirvana, tracker, throbber, labeling) {
 		$preselectedTypeInput = $classificationForm.find('#template-classification-' + preselectedType);
 
 		if ($preselectedTypeInput.length !== 0) {
-			$classificationForm.find('input[checked="checked"]').removeAttr('checked');
-			$preselectedTypeInput.attr('checked', 'checked');
+			$classificationForm.find('input:checked').removeAttr('checked');
+			$preselectedTypeInput.attr('checked', true);
 		}
 
 		// Set modal content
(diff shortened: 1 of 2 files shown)


extensions/wikia/TemplateClassification/scripts/TemplateClassificationModal.js
```diff
@@ -163,19 +166,18 @@ function ($, w, mw, loader, nirvana, tracker, throbber, labeling) {
...
+
+		$('#template-classification-' + mw.html.escape(preselectedType)).focus();
+		if (preselectedType === 'unknown') {
+			modalInstance.scroll(0);
+		}
 	}
 
 	function processSave(modalInstance) {
-		var newTemplateType = $('#TemplateClassificationEditForm [name="template-classification-types"]:checked').val(),
-			oldTemplateType = '';
```
Grunny, line 172: The `track` call in the else block below still uses `oldTemplateType` which is now undefined, so it should be updated there too.

extensions/wikia/TemplateClassification/scripts/TemplateClassificationModal.js
```diff
@@ -86,11 +88,12 @@ function ($, w, mw, loader, nirvana, tracker, throbber, labeling) {
...
 		// Mark selected type
-		$preselectedType = $classificationForm.find('#template-classification-' + templateType);
+		preselectedType = templateType;
+		$preselectedTypeInput = $classificationForm.find('#template-classification-' + preselectedType);
 
-		if ($preselectedType.length !== 0) {
+		if ($preselectedTypeInput.length !== 0) {
 			$classificationForm.find('input[checked="checked"]').removeAttr('checked');
-			$preselectedType.attr('checked', 'checked');
+			$preselectedTypeInput.attr('checked', 'checked');
```
Grunny: `$preselectedTypeInput.prop('checked', true);` is preferred.

adamkarminski: Yeah, I tried it but strangely - it does not check the selected radio button. Do you think I should investigate further or leave for a future improvement (which may not happen)? :P

extensions/wikia/TemplateClassification/scripts/TemplateClassificationModal.js
adamkarminski, line 172: Fixed.

extensions/wikia/TemplateClassification/scripts/TemplateClassificationModal.js
Grunny: Worked for me from a quick test. Can you send me a link to it not working?

azarek:
Testing done! LGTM

PR #9232, opened 2015-12-04
Mark preferences read-only after a failed read

drsnyder:
This PR marks preferences as read-only after a failed read. The `UserPreference` persistence operations are then aborted when presented with read-only preferences to prevent the corruption / loss of preferences due to populating the user preferences solely with the defaults.

https://wikia-inc.atlassian.net/browse/SERVICES-943

@Wikia/services-team

### DIFF
--- a/lib/Wikia/src/Domain/User/Preferences/UserPreferences.php
+++ b/lib/Wikia/src/Domain/User/Preferences/UserPreferences.php
@@ -4,6 +4,9 @@ namespace Wikia\Domain\User\Preferences;
 
 class UserPreferences {
 
+	/** @var boolean */
+	private $readOnly = false;
+
 	/** @var GlobalPreference[] */
 	private $globalPreferences;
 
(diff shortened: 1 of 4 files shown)


lib/Wikia/src/Domain/User/Preferences/UserPreferences.php
```diff
@@ -80,4 +83,28 @@ public function getGlobalPreferences() {
...
+	 */
+	public function setReadOnly( $state ) {
+		$this->readOnly = $state;
+		return $this;
+	}
+
+	/**
+	 * @return boolean the read only state
+	 */
+	public function getReadOnly() {
```
nmonterroso: Do we need this method? For accessing a boolean I think `isReadOnly` is more intuitive.

nmonterroso:
Small comment otw LGTM

lib/Wikia/src/Domain/User/Preferences/UserPreferences.php
drsnyder: I just added that so we have consistent `get`/`set`. But you are right I think we should prefer the usage of `isReadOnly` because it's more intuitive. I went ahead and yanked it in 2a74fe7.

PR #9236, opened 2015-12-05
Don't load Special:Preferences if prefs are RO

drsnyder:
Don't load `Special:Preferences` if the preferences for the current user are read-only. This a precautionary measure to prevent defaults from being loaded and overwriting a user's previously saved preferences.

https://wikia-inc.atlassian.net/browse/SERVICES-1056

@Wikia/services-team

### DIFF
--- a/includes/Preferences.php
+++ b/includes/Preferences.php
@@ -54,6 +54,10 @@ class Preferences {
 	 * @return array|null
 	 */
 	static function getPreferences( $user, IContextSource $context ) {
+		if ( $user->arePreferencesReadOnly() ) {
+			throw new MWException("Error, preferences are read-only.");
+		}
+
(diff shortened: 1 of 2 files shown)


nmonterroso:
Is `getPreferences` the function that's called by `Special:Preferences`? Is it called by anything else?

nmonterroso:
small question otw LGTM

drsnyder:
It looks like it's also called [here](https://github.com/Wikia/app/blob/dev/includes/api/ApiOptions.php#L75). The usage appears to be the same just API related. It's followed by `setGlobalPreference` which is what we want to avoid.

nmonterroso:
Ah yah, seems right. LGTM

PR #9237, opened 2015-12-07
Fix error found by Phan static analyzer

pawelchcki:
During first run of Phan this one of the first problems it has found
# phan


macbre:
Good catch :+1:

macbre:
Better late than never ;)

pawelchcki:
yeah: )

PR #9240, opened 2015-12-07
CE-3079 Do not display redirects on the Unorganized templates list

adamkarminski:


kamilkoterba:
LGTM

PR #9246, opened 2015-12-07
CE-3116: Add bulk classification action

lukaszkonieczny:
Add bulk classification action


lukaszkonieczny:
CR done https://github.com/Wikia/app/pull/9238

PR #9247, opened 2015-12-07
TCS and CR translations

lukaszkonieczny:
Here are translations for Template Classification and change in UserBadges feature.


lgarczewski:
LGTM.

PR #9248, opened 2015-12-07
SEO-82 Refactoring for Special:NewFiles

gabrys:
This is a first approach to the Special:NewFiles refactoring.

**I moved the whole code to the WikiaNewFiles class**. It used to call an old-school special page function.

**I removed the search feature**. `$wpIlMatch` was a central variable for it. There was also a form for that. Both the searching and showing the form was depending on `$wgMiserMode = false`, which was actually set to true globally back in 2009 without being overridden per community.

**I moved the database code to a separate class**. It's WikiaNewFilesModel (parent class: WikiaModel). The code inside is just the original code shuffled around for more readability.

**Changed `SpecialNewImages::beforeDisplay` hook**. It now doesn't pass references, but rather COW copies of the structures. It's used in one place only and doesn't modify the passed arguments.

**`getPagination` and `getWikiaGallery` methods extracted**. This is to keep the `execute` method shorter.

A `get_class` based check for Oasis replaced with `$this->getSkin()->getSkinName()`. Other than above, no "new" code was added, just reshuffled.

As the next steps, I'd like to replace the database part, add a proper pagination and maybe even create a template for the output (yay!).


extensions/wikia/WikiaNewFiles/WikiaNewFiles.class.php
```diff
@@ -20,4 +65,147 @@ function execute( $par ) {
 	public function getDescription() {
 		return $this->msg( 'wikianewfiles-title' )->text();
 	}
+
+	private function getLinkedFiles( $image ) {
+		global $wgMemc;
+		$anchorLength = 60;
+
+		wfProfileIn( __METHOD__ );
+		$cacheKey = wfMemcKey( __METHOD__, md5( $image->img_name ) );
```
garthwebb, line 73: Lets take this opportunity to pull these out.  We've got `influxdb's app_agg_xhprof_funcs` for this so these just add clutter, execution time and confusion.

extensions/wikia/WikiaNewFiles/WikiaNewFiles.class.php
```diff
@@ -20,4 +65,147 @@ function execute( $par ) {
...
+
+		$nextLink = wfMessage( 'pager-older-n', $wgLang->formatNum( $limit ) )->escaped();
+		if ( $hasOlderImages && $lastTimestamp ) {
+			$wmu['next'] = $lastTimestamp;
+			$nextLink = Linker::link( $titleObj, $nextLink,
+				array( 'class' => 'navigation-older' ),
+				'until=' . $lastTimestamp . $botpar );
+		}
+
+		return '<p id="newfiles-nav">' . $botLink . ' ' . wfMessage( 'viewprevnext' )->rawParams( $prevLink, $nextLink, $dateLink )->escaped() . '</p>';
```
garthwebb: Might be nice to format this line (and a few others) to multiple lines so it doesn't scroll off the side of the screen.

includes/specials/SpecialNewimages.php
```diff
@@ -1,4 +1,8 @@
 <?php
+
+// NOTE: This file is not in use. See WikiaNewFiles.class.php instead
+
+
```
drsnyder, line 5: Should we delete the file?

gabrys, line 5: Excellent question. It's a core MW one (+ our modifications, that are 
not in use anymore). I'm not sure what's our policy in such cases. 
Should we actually revert it to the original MW one and maybe add a die 
on top it plus this comment for anyone trying to work with it later? 
@macbre @owend?

macbre, line 5: We do not have a custom to remove not used MW core files. I'd suggest removing our changes = copying `SpecialNewimages.php` from vanilla MW 1.19.

extensions/wikia/WikiaNewFiles/WikiaNewFilesModel.class.php
```diff
@@ -0,0 +1,132 @@
...
+	/**
+	 * @var string
+	 */
+	private $hideBotsSql = '';
+
+	/**
+	 * @param bool $hideBots Whether to hide images uploaded by bots or not
+	 */
+	public function __construct( $hideBots ) {
+		$this->dbr = $this->dbr = wfGetDB( DB_SLAVE );
```
macbre: Do you want to be extra sure that this assignment works? ;)

gabrys: LOL :D

justnpT:
Please run selenium tests that make use of SpecialNewFilesPageObject:
- FacebookButtonTests.java
- FollowPhotosTests.java
- ForcedLoginTests.java
- ImageServingTests.java
- ImageStorageTests.java
- LightboxTests.java
- WikiBasePageObject.java

gabrys:
There's one test case failing that also fails on production. Code review notes addressed.

PR #9250, opened 2015-12-07
VE-2069: Adjust getImages method to work well with recent changes to ImageNode

inez:
I'm not very happy with how nested this method is but it seems that it is the convention taken within PortableInfobox project.
I would much better prefer if there was more object oriented way to delegate for instance task of retrieving images list directly to NodeImage but that would be much greater refactoring.

### DIFF
--- a/extensions/wikia/PortableInfobox/services/PortableInfoboxDataService.class.php
+++ b/extensions/wikia/PortableInfobox/services/PortableInfoboxDataService.class.php
@@ -89,7 +89,7 @@ class PortableInfoboxDataService {
 	private function getImageFromOneInfoboxData( $infoboxData ) {
 		$images = [];
 		foreach ( $infoboxData as $infoboxDataField ) {
-			if ( $infoboxDataField[ 'type' ] == self::IMAGE_FIELD_TYPE && isset( $infoboxDataField[ 'data' ] ) ) {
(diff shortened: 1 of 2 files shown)


extensions/wikia/PortableInfobox/services/PortableInfoboxDataService.class.php
```diff
@@ -77,12 +77,14 @@ public function getImages() {
...
 			foreach ( $data as $field ) {
-				if ( $field[ 'type' ] == self::IMAGE_FIELD_TYPE &&
-					 isset( $field[ 'data' ] ) &&
-					 !empty( $field[ 'data' ][ 'key' ] )
-				) {
-					$images[ $field[ 'data' ][ 'key' ] ] = true;
+				if ( $field[ 'type' ] == self::IMAGE_FIELD_TYPE && isset( $field[ 'data' ] ) ) {
+					for ( $i = 0; $i < count( $field[ 'data' ] ); $i++ ) {
+						if ( !empty( $field[ 'data' ][ $i ] [ 'key' ] ) ) {
+							$images[ $field[ 'data' ][ $i ][ 'key' ] ] = true;
```
pauloslund: Maybe this `for` loop can be extracted into it's own method? May help with how nested this method is. Other than that though it looks good

PR #9252, opened 2015-12-07
INT-252 Wikia-in-your-lang new translations

mari-mcmurtrie:
Added missing Dutch translations + improved existing translations for ES, JA, and RU.


mari-mcmurtrie:
@perjg will you CR? Thanks~

perjg:
LGTM

PR #9254, opened 2015-12-08
Remove old/unused SpecialContact extension

owend:
Delete entire SpecialContact extension, we are using SpecialContact2 now.  

@Grunny


Grunny:
:+1:

PR #9259, opened 2015-12-08
(CE-2948) Expose wgWikiDirectedAtChildrenByFounder as a JS var

Grunny:
Quick solution to make this variable available to Selenium tests without
WikiFactory access.


adamkarminski:
LGTM

PR #9264, opened 2015-12-08
(CE-2970) Add logging of classification actions to Special:Log

adamkarminski:
The copy is approved by Trevor - we use only 1 message and if a template was not classified before we just display a change from Unknown.

Side note: From a quick test that I ran it seems that adding an additional request to TCS for the oldType and adding an entry to logs does not make the response time of the API significantly longer.


kamilkoterba:
LGTM

azarek:
Testing done! LGTM

Grunny:
To define this properly on Special:Log so it can be filtered, the following needs to be added:

``` php
$wgLogTypes[] = 'templateclassification';
```

By default, the log page title will use `log-name-templateclassification` and `log-description-templateclassification` as the message keys for the filtered log page. These keys can be overridden, however.

azarek:
Filtration works!

Grunny:
:+1:

PR #9265, opened 2015-12-08
INT-223: Move Achievement translation files to correct location

mari-mcmurtrie:
For Achievement, translation file (AchievementsII.i18n.php) and crowdin.conf were located under /i18n directory, which prevented crowdin script to run since the script expect these files in top dir. So I moved files in correct directory and changed reference to the file in code. Ticket: https://wikia-inc.atlassian.net/browse/INT-223

To make sure if this works, we have to check if we still have Achievement translations correctly displayed in pages such as /Special:AchievementsCustomize.


mari-mcmurtrie:
@perjg will you check this? Crowdin doc is here: https://one.wikia-inc.com/wiki/Internationalization/i18n_Tools

perjg:
LGTM

macbre:
@mari-mcmurtrie / @perjg - please update the path here as well - https://github.com/Wikia/config/blob/dev/CommonExtensions.php#L140

PR #9266, opened 2015-12-08
VE-2044 | added support to enable entry point for each language, and a toggle to disable for all languages

katanka:


extensions/wikia/EditPageLayout/EditPageLayoutController.class.php
```diff
@@ -228,6 +228,10 @@ public function executeEditPage() {
 			|| $templateType === TemplateClassificationService::TEMPLATE_INFOBOX
 			|| $templateType === TemplateClassificationService::TEMPLATE_CUSTOM_INFOBOX;
 
+		if ( !$wgInfoboxPreviewEnabled || !in_array( strtolower( $wgUser->getGlobalPreference( 'language' ) ), $wgInfoboxPreviewSupportedLanuages ) ) {
```
inez: Looks good, but would it make sense to create a method, something like getShowInfoboxPreviewState(), that cotain pretty much everything from line 213 to 232?

inez:
LGTM except for the name of the method. You are "asking" but it's not clear from the name of the method - I would think that if I call this method I actually make preview appear. Maybe something with get*?

katanka:
I disagree, `should` is used in function names quite often. Example: https://github.com/Wikia/app/blob/dev/extensions/wikia/EditorPreference/EditorPreference.class.php#L173

PR #9267, opened 2015-12-09
INT-289: Disable Google search for Oasis intra-wiki search. Even if a…

mari-mcmurtrie:
… user is assigned to test group by Optimizely, we do non-google search if it is intra search. Ticket: https://wikia-inc.atlassian.net/browse/INT-289


mari-mcmurtrie:
@perjg will you CR? I tested this already. Basically, for Oasis, GoogleSearchAPI.js runs only when a user is in the test group (see condition !window.wgGoogleSearchTest, that indicate if a user is in test or not). When a user submits a search query, we examine the request and if it is intra search, I do not proceed and behavior falls into default. That is the change I made. Thanks!

perjg:
LGTM

PR #9270, opened 2015-12-09
Monetization (SEO-157): add meta tags for Twitter Cards

saipetch:


owend:
Comment on ticket says to use og:image (at least for now) so we don't need to do that part now.

owend:
Ok, LGTM

PR #9274, opened 2015-12-09
Phalanx: use phalanx.service.sjc.consul instead of local Varnish as a proxy

macbre:
[PLATFORM-1744](https://wikia-inc.atlassian.net/browse/PLATFORM-1744)

Use consul to configure Phalanx hosts instead of relying on a local Varnish instance sending requests to the service.

And keep the config in a single place instead of three different places.

@wladekb / @pchojnacki / @michalroszka  / @artursitarski


pawelchcki:
hmm interesting shift away from config repo. :+1: 
LGTM

pchojnacki:
LGTM

PR #9275, opened 2015-12-09
block access to /lib directory

adamw-linadm:
Prevent executing php from /lib directory.

eg: http://poznan.macbre.wikia-dev.com/lib/composer/zendframework/zend-code/src/Reflection/ClassReflection.php - gives a fatal

cc @macbre


macbre:
:+1:

PR #9277, opened 2015-12-09
CE-3155+CE-3151 Add translations for TC logs and bulk classification

adamkarminski:


Grunny:
LGTM.

PR #9278, opened 2015-12-09
Do not create searchindex table only to drop it a minute later

wladekb:
This should prevent CNW from creating the "searchindex" table only to be able to drop it a few minutes later.

https://wikia-inc.atlassian.net/browse/PLATFORM-1741

/cc @macbre @drozdo


macbre:
And `MyISAM--` :)

PR #9285, opened 2015-12-09
Monetization (MON-442): MonetizationModule - do not send request to MonetizationService from Reston

saipetch:


owend:
Sure, LGTM

macbre:
Thanks for the fix!

Can we however make the change in the config instead? We do have a "disable things in Reston" [section](https://github.com/Wikia/config/blob/dev/CommonExtensions.php#L15) in `CommonExtensions.php`. I'd simply disable the MonetizationService extension there.

PR #9286, opened 2015-12-09
Revert "VE-2080: Set Ooyala video player to prioritize HTML5 over Flash"

pauloslund:
Reverts Wikia/app#9200 because of: https://wikia-inc.atlassian.net/browse/MAIN-6051


inez:
Looks awesome - but please link to a related ticket - so someone can track why we are reverting it.

PR #9294, opened 2015-12-10
DAT-3572 add makeIconLink function

dianafa:
https://wikia-inc.atlassian.net/browse/DAT-3572


includes/wikia/parser/templatetypes/handlers/InfoIconTemplate.class.php
```diff
@@ -20,16 +20,34 @@ public static function handle( $wikitext, Parser $parser ) {
...
+	 * @param $image
+	 * @return String
+	 * @throws \MWException
+	 */
+	private static function makeIconLink( $image ) {
+		$link = '';
+		$title = Title::newFromText( $image );
+		if ( $title ) {
+			$file = wfFindFile( $title );
+			if ( $file ) {
```
sebastianmarzjan: Does it make sense to check $file && $file->exists() ?

includes/wikia/parser/templatetypes/handlers/InfoIconTemplate.class.php
```diff
@@ -20,16 +20,34 @@ public static function handle( $wikitext, Parser $parser ) {
...
 	}
+
+	/**
+	 * @desc for a given wikitext image mark, create imagelink
+	 *
+	 * @param $image
+	 * @return String
+	 * @throws \MWException
+	 */
+	private static function makeIconLink( $image ) {
```
sebastianmarzjan, line 36: I need to think about it, but maybe you have an idea how to flatten this method? The nesting level is pretty high for a logically simple function.

dianafa, line 36: only idea which comes to my mind is to use `return` twice here, but it's even worse -.-

dianafa, line 36: OK, so if `$title` will be empty (param in `wfFindFile` will be empty) it will not cause error

sebastianmarzjan:
Makes Perfect Sense To Me ;-)

PR #9296, opened 2015-12-10
CE-3172 Fix for PHP Strict error - child method should have the same definition

adamkarminski:


kamilkoterba:
LGTM

PR #9297, opened 2015-12-10
i18n for content-moderator group

macbre:
i18n messages for the "Content Moderators" group that will be introduced by [PLATFORM-1734](https://wikia-inc.atlassian.net/browse/PLATFORM-1734).

@wladekb / @michalroszka


languages/messages/wikia/crowdin.conf
```diff
@@ -0,0 +1,4 @@
+[crowdin.project]
+projectid = "CoreMessagesWikia"
+format = "mediawiki"
+source.file = "MessagesEn.php"
```
macbre: Pinging @Grunny

Grunny: Looks OK for upload, but the i18n commandline tool doesn't support downloading the translations into the separate language files for MediaWiki right now, so downloading will need some manual work to move them into the right files.

Grunny: And actually, upload won't work either, since the script looks for the `en` key in the messages array, which doesn't apply to this file.

macbre: :(

So the manual upload to Crowdin is the only solution here? A fake extension with i18n file with these messages only sounds way to hacky to me ;)

Grunny: Yeah :( though when I've done this for these files before, I've just created a temporary i18n.php file locally and used it to do the upload and later download, then just copied the translations to the right files in languages/messages/wikia. It's ugly, but at least you get the conversion to and from JSON.

macbre: Good hint, will try. Thanks :)

PR #9298, opened 2015-12-11
DAT-3576, allow for overwrite thumbnail physical dimensions in portab…

RafLeszczynski:
…le infobox via wikifactory variable

https://wikia-inc.atlassian.net/browse/DAT-3576

### DIFF
--- a/extensions/wikia/PortableInfobox/services/Helpers/PortableInfoboxRenderServiceHelper.php
+++ b/extensions/wikia/PortableInfobox/services/Helpers/PortableInfoboxRenderServiceHelper.php
@@ -251,7 +251,7 @@ class PortableInfoboxRenderServiceHelper {
 	 * @return array width and height which will be displayed i.e. in the width
 	 * and height properties of the img tag
 	 */
-	private function getImageSizesToDisplay( $thumbnail ) {
+	public function getImageSizesToDisplay( $thumbnail ) {
 		global $wgPortableInfoboxCustomImageWidth;
 
 		if ( !$this->isWikiaMobile() && !empty( $wgPortableInfoboxCustomImageWidth ) ) {
--- a/extensions/wikia/PortableInfobox/services/Helpers/PortableInfoboxRenderServiceHelper.php
+++ b/extensions/wikia/PortableInfobox/services/Helpers/PortableInfoboxRenderServiceHelper.php
@@ -112,11 +112,12 @@ class PortableInfoboxRenderServiceHelper {
 		}
 
 		wfRunHooks( 'PortableInfoboxRenderServiceHelper::extendImageData', [ $data, &$ref ] );
+		$dimensions = self::getImageSizesToDisplay( $thumbnail );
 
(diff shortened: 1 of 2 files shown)


RafLeszczynski:
TODO: add / fix unit tests

extensions/wikia/PortableInfobox/services/Helpers/PortableInfoboxRenderServiceHelper.php
```diff
@@ -214,6 +214,27 @@ private function getThumbnail( $title ) {
...
+	 * @todo rethink approach to images as a part of https://wikia-inc.atlassian.net/browse/DAT-3075
+	 * @param $thumbnail \MediaTransformOutput
+	 * @return string thumbnail URL
+	 */
+	private function getPhysicalSizeThumbUrl( $thumbnail ) {
+		global $wgPortableInfoboxCustomImageWidth;
+
+		if ( !empty( $wgPortableInfoboxCustomImageWidth ) ) {
+			$file = $thumbnail->file;
+			$height = min( self::MAX_DESKTOP_THUMBNAIL_HEIGHT, $file->getHeight() );
```
sebastianmarzjan: I wonder whether we want to use self::MAX_DESKTOP_THUMBNAIL_HEIGHT or rather a proportionally scaled value to match custom width passed through the WF var

dianafa: the `wgPortableInfoboxCustomImageWidth` will be probably set to be bigger than our current 270px. Thus, if we want to scale it height proportionally instead of use `MAX_DESKTOP_THUMBNAIL_HEIGHT`, it would have to be bigger than this. I'm not sure if images taller than our current `MAX_DESKTOP_THUMBNAIL_HEIGHT` (500px) are something we want to have.

extensions/wikia/PortableInfobox/services/Helpers/PortableInfoboxRenderServiceHelper.php
```diff
@@ -214,19 +215,51 @@ private function getThumbnail( $title ) {
...
+	/**
+	 * @desc if it's not a request from mobile skin and wgPortableInfoboxCustomImageWidth
+	 * is set, the $thumbnail->getWidth() can return some big value - we need
+	 * to adjust it to DESKTOP_THUMBNAIL_WIDTH to look good in the infobox.
+	 * Also, the $height have to be adjusted here to make image look good in infobox.
+	 * @param $thumbnail
+	 * @return array width and height which will be displayed i.e. in the width
+	 * and height properties of the img tag
+	 */
+	private function getImageSizesToDisplay( $thumbnail ) {
```
sebastianmarzjan: This is a method that could IMO be public and tested. Any objections towards doing that?

dianafa: :+1

RafLeszczynski:
LGTM

PR #9300, opened 2015-12-11
Monetization (SEO-184): Update html title for Chat and ThemeDesigner

saipetch:


includes/wikia/WikiaHtmlTitle.class.php
```diff
@@ -41,6 +41,26 @@ public function __construct() {
 	}
 
 	/**
+	 * Get brand name
+	 * @return String
+	 */
+	public function getBrandName() {
```
owend, line 47: Should use escaped() instead of text() in both these functions to be safer.

owend:
Okay, LGTM

PR #9302, opened 2015-12-12
DAT-3576, fix edge case with broken aspect ratio for tall thin images…

RafLeszczynski:
… if portable infobox custom image width is set

https://wikia-inc.atlassian.net/browse/DAT-3576


RafLeszczynski:
TODO: add / fix tests

extensions/wikia/PortableInfobox/services/Helpers/PortableInfoboxRenderServiceHelper.php
```diff
@@ -247,7 +247,7 @@ public function getImageSizesForThumbnailer( $image ) {
 	 * is set, the $thumbnail->getWidth() can return some big value - we need
 	 * to adjust it to DESKTOP_THUMBNAIL_WIDTH to look good in the infobox.
 	 * Also, the $height have to be adjusted here to make image look good in infobox.
-	 * @param $thumbnail
+	 * @param $thumbnail \MediaTransformOutput
```
sebastianmarzjan, line 250: Since we rely on the type of input object (we use its methods internally) it makes sense to me to add a hard type hint here

sebastianmarzjan:
LG

PR #9304, opened 2015-12-12
MAIN-6066 We don't need CDATA in output XML; in fact, we don't need XML in output at all 

sebastianmarzjan:
https://wikia-inc.atlassian.net/browse/DAT-3578


dianafa:
LGTM

PR #9305, opened 2015-12-14
DatabaseBase::isWriteQuery - EXPLAIN should not be treated as writable query

macbre:
Backported the current regex from https://github.com/wikimedia/mediawiki/blob/master/includes/db/Database.php#L890-L892 (which adds `ROLLBACK` there as well).

@wladekb / @michalroszka


wladekb:
lgtm

PR #9307, opened 2015-12-14
ADEN-2770 Move Taboola to seperate experiment

nandy-andy:
Currently Taboola is enabled to 60% of ad-blocking traffic. We want to ramp it up but until it's in `PROJECT_43` experiment it's impossible.

Changes below moves Taboola to new experiment and updates unit tests.


fraszczakszymon:
LGTM :+1:

PR #9308, opened 2015-12-14
Security: introduce Utils::matchToken helper

macbre:
[PLATFORM-1703](https://wikia-inc.atlassian.net/browse/PLATFORM-1703)

Use `Wikia\Security\Utils::matchToken` for checking custom tokens in `Special:EmailTest`.

@wladekb / @Grunny

### DIFF
--- a/extensions/wikia/Security/classes/Utils.class.php
+++ b/extensions/wikia/Security/classes/Utils.class.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace Wikia\Security;
+
+class Utils {
+
+	/**
+	 * A more generic version of User::matchEditToken that can be used for checking custom tokens
+	 *
+	 * @see PLATFORM-1703
+	 *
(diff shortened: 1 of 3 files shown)


wladekb:
:+1:

Grunny:
As a best practice as a reusable util, I think it would be better to use `hash_equals` so it's not affected by timing attacks.

macbre:
@Grunny - #9310

extensions/wikia/Security/classes/Utils.class.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+namespace Wikia\Security;
```
drsnyder, line 3: @macbre this namespace should be in `lib/Wikia/src`. We have been trying to consolidated all of the `Wikia` namespace there so its PSR-\* compliant. This also removes the need to alter the `$wgAutoloadClasses` map.

PR #9315, opened 2015-12-15
VE-1956: Change addmediapermanent and addmediatemporary apis to supprrt only POST requests and to require tokens

inez:
This also includes usage of those apis in MW.

PLEASE DO NOT MERGE yet - because it requires corresponding change in Mercury (which uses addmediapermanent)


Grunny:
LGTM.

pauloslund:
:+1:

PR #9317, opened 2015-12-15
Create a Special:Log API function in MW to add logs

macbre:
[PLATFORM-1733](https://wikia-inc.atlassian.net/browse/PLATFORM-1733)

Introduce `LogEventsApiController` Nirvana API that accepts POST requests (with a valid Schwartz token) containing a log entry to be added.

Usage example in the ticket linked above.

@jcellary / @wladekb / @pchojnacki


includes/wikia/api/LogEventsApiController.class.php
```diff
@@ -0,0 +1,53 @@
...
+		$comment = $this->request->getVal( 'comment' );
+		$params = json_decode( $this->request->getVal( 'params', '[]' ) );
+
+		try {
+			$entry = new LogPage( $type );
+			$id = $entry->addEntry( $action, $wgTitle, $comment, $params, $wgUser );
+		}
+		catch ( DBQueryError $e ) {
+			// 1048 Column 'log_type' cannot be null
+			throw new BadRequestException( $e->getMessage() );
```
wladekb: Exposing exact error message doesn't necessarily need to be secure. Can we improve the reporting here (log the error and return general message with request id so we could correlate reports later)?

macbre: Good point - fixed in 4698697

The following will be returned to the client

``` json
{
  "exception": {
    "type": "DBQueryError",
    "message": "",
    "code": 0,
    "details": ""
  },
  "request_id": "mw567131d8e0f5d5.24916831"
}
```

includes/wikia/api/LogEventsApiController.class.php
```diff
@@ -0,0 +1,45 @@
...
+	 * @throws BadRequestException
+	 */
+	public function add() {
+		global $wgTheSchwartzSecretToken, $wgTitle, $wgUser;
+
+		if ( !$this->request->wasPosted() ) {
+			throw new BadRequestException( 'This request must be POSTed' );
+		}
+
+		if ( !Wikia\Security\Utils::matchToken( $wgTheSchwartzSecretToken, $this->request->getVal( 'token' ) ) ) {
```
drsnyder, line 29: @michalroszka, @ArturKlajnerok do you know of any reason why we decided to use `'secret'` for the [helios end points](https://github.com/Wikia/app/blob/dev/extensions/wikia/Helios/HelperController.class.php#L221) instead of `token`? Can we consolidate on one?

PR #9318, opened 2015-12-15
(CE-3132) Use _LICENSE_ variable rather than hardcoded license

Grunny:
This will match the behaviour of English and other languages. It will
make use of the "copyright" message, which takes the copyright variable
for the wikia as a parameter so the correct license is shown on non–CC-BY-SA
wikias.


adamkarminski:
LGTM!

PR #9321, opened 2015-12-16
Fix fatal error if title doesn't exist

kamilkoterba:
Found a Fatal Error possible.

@Wikia/community-engineering


Grunny:
Since we're going to do this check anyway, should we consider checking if the title exists before sending the ID to the service and adding the log entry?

PR #9322, opened 2015-12-16
PLATFORM-1727: Add extra logging around (un)deleting articles

wladekb:
We've caught some cases that deleted articles are missing the revisions. Let's add some logging to have a better overview of what's going under the hood.

https://wikia-inc.atlassian.net/browse/PLATFORM-1727

/cc @macbre


macbre:
Let's debug this mystery!

PR #9323, opened 2015-12-16
Add caching around CategoryDataService::getArticleCount

macbre:
[PLATFORM-1192](https://wikia-inc.atlassian.net/browse/PLATFORM-1192) / [Kibana dashboard](https://kibana.wikia-inc.com/#/dashboard/elasticsearch/CategoryDataService:getArticleCount)

It hits the database heavily as it's executed for the same category over and over again when processing offline tasks (with some queries reaching 45 seconds of execution time and an average at ~6 seconds).

@wladekb / @pchojnacki / @frankfarmer

### DIFF
--- a/extensions/wikia/CategoryExhibition/services/CategoryDataService.class.php
+++ b/extensions/wikia/CategoryExhibition/services/CategoryDataService.class.php
@@ -36,6 +36,8 @@ class CategoryDataService extends Service {
 	/**
 	 * Return the number of articles that are in a particular category.
 	 *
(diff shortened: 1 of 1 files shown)


wladekb:
lgtm,

I'd just double check all the places where this method is used and verify that receiving a possibly inaccurate count doesn't break anything.

macbre:
It's only used in `CategoryExhibitionSection::isCategoryExhibitionEnabled` to determine if the category is not too big to enable this extension for - `CategoryExhibitionSection` performs some heavy black magic on database ;)

PR #9328, opened 2015-12-16
Refactor of RTEParser::doHeadings to remove buggy loop

Tjorriemorrie:
I was asked how to improve this during an interview today by @rafalkalinski. I was too nervous to think of the solution during the interview, but as soon as the interview was finished I thought of a possible solution. 

My internet is too slow to download the vagrant image or clone the `app` repository, so I edited it directly on Github off the top of my head; i.e. I assume there are some CI to test this. 

I'm not familiar with your markdown but I kept the exclusion of the `h6` tag intact.

Where is the test class for this?


mszabo-wikia:
Tests for RTE? If only there were any.

PR #9329, opened 2015-12-16
SEO-182 Special:NewFiles -> Special:Images + refactoring

gabrys:
This is the next iteration of the Special:NewFiles page.

The page is renamed to Special:Images. There is a special redirect constructed using `$wgSpecialPages['Newimages'] = 'WikiaNewFilesSpecialController';` in order to redirect users from Special:NewFiles to Special:Images.

**EDIT: actually e02072d reverts the rename of the page. We'll do it separately to avoid to many changes at once.**

Translations for "Images" re-used from message `wikistats_images`. Those translations were also used to construct per-language aliases for the special page.

The following features are missing vs the original Special:NewFiles:
- include/exclude images uploaded by bots (the new code always includes those): `?hidebots=0`
  - bots contribute very minimal number of images according to our research
- supply custom limit by URL, i.e. `/Special:NewFiles/10`
- filter images shown by date: `?until=DATE`
  - users don't seem to use pagination or the custom limit that much (98% of the requests to the Special:NewFiles doesn't include any URL params)

Notable changes:
- The code was split into a few files, massively simplified and the modern Wikia-style pagination was added including the HEAD elements for better SEO.
- `...` were replaced by `…` in messages and old-school `array()`s with the shorthand `[]` notation.
- `SpecialNewImages::beforeDisplay` hook was again simplified
- `$wmu` support was removed (didn't seem to work anyway)
- Presentation logic is moved to `WikiaNewFilesSpecial_index.php`
- `getLinkedFiles` method moved to model, renamed to `addLinkingArticles` and made private. It's now called to extend the original `$row` objects for the fetched images
- all files in the extension (except i18n and not touched uploadPhotosDialog.js/.css) are now below 100 lines
- all SQL rewritten to FluentSQL


extensions/wikia/ContentFeeds/ContentFeeds.class.php
```diff
@@ -219,13 +219,13 @@ public static function dummyRecentImagesParserHook( $input, $args, $parser ) {
 		return "Życie..";
 	}
 
-	public static function specialNewImagesHook( $images, $gallery, $limit ) {
```
garthwebb, line 222: Is this the custom limit you mentioned in the PR description that users don't seem to use much?  Maybe in future PRs make a note of things like this as in-line annotations so reviewers know they're looking at the right thing.

extensions/wikia/ContentFeeds/ContentFeeds.class.php
```diff
@@ -219,13 +219,13 @@ public static function dummyRecentImagesParserHook( $input, $args, $parser ) {
...
+	public static function specialNewImagesHook( $images ) {
 		global $wgRequest;
 
 		if( $wgRequest->getVal('feed') == 'rss' ) {
 			//filter images and get the file url
 			$filteredImages = array();
```
garthwebb, line 227: Old school array!  (OSA).  Definitely on board with updating these :).  Were you updating all occurrences in the files you touched or only the specific lines of code you changed?

extensions/wikia/Paginator/Paginator.body.php
```diff
@@ -93,6 +93,12 @@ public function hasContent(){
 		return ( $this->pagesCount >= 1 );
 	}
 
+	/**
+	 * Set the currently active page. This is 0-indexed, so you may need to
+	 * set the value to $this->getRequest()->getInt( 'page' ) - 1
+	 *
+	 * @param $iPageNumber
```
garthwebb: While there isn't an explicit rule against type based variable names,  in general the MW guidelines for PHP discourage the use of prefixes (except for `wg` for globals):

> https://www.mediawiki.org/wiki/Manual:Coding_conventions/PHP#Variables

You can note the type of this variable above by adding it before the variable name, e.g.:

``` php
@param int $pageNumber
```

extensions/wikia/WikiaNewFiles/WikiaNewFilesModel.class.php
```diff
@@ -2,131 +2,77 @@
...
 
 	/**
-	 * @var string
+	 * @var Database
 	 */
-	private $hideBotsSql = '';
+	private $dbr;
 
 	/**
 	 * @param bool $hideBots Whether to hide images uploaded by bots or not
```
garthwebb: Can remove this `@param` line now

extensions/wikia/WikiaNewFiles/WikiaNewFilesModel.class.php
```diff
@@ -2,131 +2,77 @@
...
+		return intval( $count );
 	}
 
 	/**
-	 * Query the database to find out what is the timestamp of the newest image uploaded
+	 * Get the specific page of images
 	 *
-	 * @return string the timestamp of the newest image in MediaWiki format (YmdHis)
+	 * @param $limit      images per page
+	 * @param $pageNumber page number (1-indexed)
```
garthwebb: Could add `int` as type for the above params.

extensions/wikia/WikiaNewFiles/WikiaNewFilesModel.class.php
```diff
@@ -2,131 +2,77 @@
...
-			$ts = false;
-		}
-		$this->dbr->freeResult( $res );
-
-		return wfTimestamp( TS_MW, $ts );
+	public function getImagesPage( $limit, $pageNumber ) {
+		$sql = ( new WikiaSQL() )
+			->SELECT( 'img_size', 'img_name', 'img_user', 'img_user_text', 'img_description', 'img_timestamp' )
+			->FROM( 'image' )
+			->ORDER_BY( [ 'img_timestamp', 'DESC' ] )
```
garthwebb: This could be:

``` php
->ORDER_BY( 'img_timestamp' )->DESC()
```

extensions/wikia/WikiaNewFiles/WikiaNewFilesModel.class.php
```diff
@@ -2,131 +2,77 @@
...
-
-		return $images;
+	private function addLinkingArticles( $image ) {
+		$sql = ( new WikiaSQL() )
+			->SELECT( 'page.page_namespace', 'page.page_title' )
+			->FROM( 'imagelinks' )
+			->JOIN( 'page' )->ON( 'imagelinks.il_from', 'page.page_id' )
+			->WHERE( 'imagelinks.il_to' )->EQUAL_TO( $image->img_name )
+			// Get the NS_MAIN first
+			->ORDER_BY( 'page.page_namespace ASC' )
```
garthwebb: Ordering by `ASC` is default so this could be:

``` php
->ORDER_BY( 'page.page_namespace' )
```

extensions/wikia/WikiaNewFiles/WikiaNewFilesModel.class.php
```diff
@@ -2,131 +2,77 @@
...
+			->LIMIT( 2 );
+
+		$sql->cache( self::CACHE_LINKING_ARTICLES_TTL, null, true /* cache empty */ );
```
garthwebb, line 66: Oh how I wish PHP had named parameters!

gabrys, line 66: I guess what I did is the closest to named params it gets :D

extensions/wikia/ContentFeeds/ContentFeeds.class.php
gabrys, line 222: This is just a hook. The custom $limits comes from the URL.

extensions/wikia/ContentFeeds/ContentFeeds.class.php
gabrys, line 227: I only converted them in the WikiaNewFiles extension.

extensions/wikia/WikiaNewFiles/WikiaNewFilesSpecialController.class.php
```diff
@@ -0,0 +1,52 @@
...
+		// Construct gallery
+		$gallery = new WikiaNewFilesGallery( $this->specialPage->getSkin() );
+		$gallery->addImages( $images );
+
+		// View
+		$this->setVal( 'gallery', $gallery );
+		$this->setVal( 'showUi', !$this->including() );
+		$this->setVal( 'noImages', !$imageCount );
+		$this->setVal( 'emptyPage', count( $images ) === 0 );
+		$this->setVal( 'pagination', $paginator->getBarHTML( $paginatorUrl ) );
```
garthwebb: Could also set these all at once via:

```
$this->response->setValues( [
    'gallery' => $gallery,
    'showUi' => !$this->including(),
    ...
] );
```

extensions/wikia/WikiaNewFiles/WikiaNewFilesSpecialController.class.php
```diff
@@ -0,0 +1,52 @@
...
+		// Hook for ContentFeeds::specialNewImagesHook
+		wfRunHooks( 'SpecialNewImages::beforeDisplay', array( $images ) );
+
+		// Construct gallery
+		$gallery = new WikiaNewFilesGallery( $this->specialPage->getSkin() );
+		$gallery->addImages( $images );
+
+		// View
+		$this->setVal( 'gallery', $gallery );
+		$this->setVal( 'showUi', !$this->including() );
```
garthwebb: The `including` method is defined in `SpecialPage` which is the base class for `IncludableSpecialPage` which is what this class used to inherit from, so it currently doesn't exist for `$this`.  To be honest I don't know how or why you would transclude the new images special page, but maybe ask community.  If its not needed you could get rid of this whole line.

extensions/wikia/WikiaNewFiles/WikiaNewFilesSpecialController.class.php
```diff
@@ -0,0 +1,52 @@
...
+		// Fetch data from DB
+		$newFilesModel = new WikiaNewFilesModel();
+		$images = $newFilesModel->getImagesPage( self::DEFAULT_LIMIT, $pageNumber );
+		$imageCount = $newFilesModel->getImageCount();
+
+		// Pagination
+		$paginatorUrl = '?page=%s';
+		$paginator = Paginator::newFromArray( $imageCount, self::DEFAULT_LIMIT );
+		$paginator->setActivePage( $pageNumber - 1 );
+		$this->getOutput()->addHeadItem( 'Paginator', $paginator->getHeadItem( $paginatorUrl ) );
```
garthwebb: This class doesn't define `getOutput`; it should probably be:

``` php
$this->getContext()->getOutput()->addHeaditem( ... );
```

garthwebb: Ugh, just noticed that `WikiaSpecialPageController` uses `__call` to "inherit" from `SpecialPage` if the method can't be found.  Should probably still use `getContext` as you did above since its a bit more clear.  Definitely hate the `__call` method...

gabrys: Yeap. In fact I'm calling the output through getContext() above, so I'm 
gonna extract it to a variable and call the methods on the variable instead

extensions/wikia/WikiaNewFiles/WikiaNewFilesSpecialController.class.php
garthwebb: Same "ugh" as above; this ends up working because of the `__call` method defined in `WikiaSpecialPageController` that hands control over to `SpecialPage`.  Probably still worth asking about any current use of this feature.

gabrys: Good point. It's not known to be massively used, so we're going to 
mention it's going to stop working in the weekly release update and 
release it. If there are any complaints (meaning there are some use 
cases for the inclusion), the code may be easily reverted and the use 
cases included in the next version of the refactored code.

extensions/wikia/Paginator/Paginator.body.php
gabrys: done

extensions/wikia/WikiaNewFiles/WikiaNewFilesModel.class.php
gabrys: done, thanks!

extensions/wikia/WikiaNewFiles/WikiaNewFilesModel.class.php
gabrys: done

extensions/wikia/WikiaNewFiles/WikiaNewFilesModel.class.php
gabrys: done

extensions/wikia/WikiaNewFiles/WikiaNewFilesModel.class.php
gabrys: good point, done

extensions/wikia/WikiaNewFiles/WikiaNewFilesSpecialController.class.php
gabrys: converted

garthwebb:
Ok, LGTM

PR #9331, opened 2015-12-16
SOC-1540 Load first and last revisions properly

garthwebb:
This ticket was about a null value being used as an object and turned out to reveal some less than ideal ways of loading data from the DB.

Tried to improve the data loading while also updating a few other things, notably:
- Replaced Wikia::log calls with WikiaLogger calls
- Removed ProfileIn|Out calls
- Reduced conditional nesting (return early rather than wrap huge chunks of code in IF statements)

https://wikia-inc.atlassian.net/browse/SOC-1540


extensions/wikia/ArticleComments/classes/ArticleComment.class.php
```diff
@@ -49,6 +49,9 @@ class ArticleComment {
 
 	protected $minRevIdFromSlave;
 
+	/** @var bool */
```
garthwebb, line 52: This was added so that we only call the `load` method once per instance.

extensions/wikia/ArticleComments/classes/ArticleComment.class.php
```diff
@@ -130,13 +133,11 @@ static public function latestFromTitle( Title $title, array $param = [] ) {
 	}
 
 	/**
+	 * Used to store extra data in comment content
 	 *
-	 * Used to store extra data in comment contend
-	 *
-	 * @access public
```
garthwebb, line 136: Removed all these `@access` and `@static` annotations.  They add nothing over the PHP keywords used in the method signature.

extensions/wikia/ArticleComments/classes/ArticleComment.class.php
```diff
@@ -197,84 +192,124 @@ static public function newFromId( $id ) {
 	 * @return bool
 	 */
 	public function load( $master = false ) {
-		wfProfileIn( __METHOD__ );
-		$result = true;
```
garthwebb, line 201: Got rid of `$result`.  It was only used to store a true/false return value for one return statement at the end (typically because the use of wfProfileIn|Out makes it annoying to have more than one return statement).  Returning immediately with `false` in places where `$result` was set to `false`.

extensions/wikia/ArticleComments/classes/ArticleComment.class.php
```diff
@@ -197,84 +192,124 @@ static public function newFromId( $id ) {
...
+		if ( $this->isLoaded ) {
+			return true;
+		}
 
-		if ( $this->mTitle ) {
- 			// get revision ids
```
garthwebb, line 204: All of this convoluted code was moved to `loadFirstRevId` and `loadLastRevId` methods.

extensions/wikia/ArticleComments/classes/ArticleComment.class.php
```diff
@@ -197,84 +192,124 @@ static public function newFromId( $id ) {
...
-				wfProfileOut( __METHOD__ );
-				return false;
-			}
+		// Fall back to master if not on slave or if we wanted master in the first place
+		if ( empty( $this->mFirstRevId ) ) {
+			$this->mFirstRevId = $this->getFirstRevID( DB_MASTER );
+		}
 
-			// get revision objects
-			if ( $this->mFirstRevId ) {
```
garthwebb, line 238: All of this was moved to `loadFirstRevision` and `loadLastRevision` methods

extensions/wikia/ArticleComments/classes/ArticleComment.class.php
```diff
@@ -363,10 +392,6 @@ public function getTitle() {
 	public function getData( $master = false ) {
 		global $wgUser, $wgBlankImgUrl, $wgMemc, $wgArticleCommentsEnableVoting;
 
-		wfProfileIn( __METHOD__ );
-
-		$comment = false;
```
garthwebb, line 368: The `$comment` variable is set to false and used as a return value for failure until its overwritten with an actual comment below.  Updated to return false directly in those places and not bothering to define it up here just to hold a non-comment boolean.

extensions/wikia/ArticleComments/classes/ArticleComment.class.php
```diff
@@ -392,96 +417,95 @@ public function getData( $master = false ) {
...
 			return $data;
 		}
 
-		if ( $this->load( $master ) ) {
-			$sig = ( $this->mUser->isAnon() )
-				? AvatarService::renderLink( $this->mUser->getName() )
-				: Xml::element( 'a', array ( 'href' => $this->mUser->getUserPage()->getFullUrl() ), $this->mUser->getName() );
+		if ( !$this->load( $master ) ) {
+			return false;
+		}
```
garthwebb, line 425: Invert the logic above so we can return early if we can't load the comment.  The following changes reflect some of the indention change that happens because of this.

extensions/wikia/ArticleComments/classes/ArticleComment.class.php
```diff
@@ -672,39 +692,35 @@ public static function isBlog( Title $title = null ) {
...
-				'canEdit'				=> $this->canEdit(),
-				'comment'				=> htmlentities( ArticleCommentsAjax::getConvertedContent( $this->mLastRevision->getText() ) ),
-				'isReadOnly'			=> wfReadOnly(),
-				'isMiniEditorEnabled'	=> ArticleComment::isMiniEditorEnabled(),
-				'stylePath'				=> $wgStylePath,
-				'articleId'				=> $this->mTitle->getArticleId(),
-				'articleFullUrl'		=> $this->mTitle->getFullUrl(),
-			];
-			$text = F::app()->getView( 'ArticleComments', 'Edit', $vars )->render();
+		if ( !$this->load( true ) ) {
```
garthwebb, line 700: Actually check the return value of `load` and return empty string if it fails (the same return value used in the rest of this function for failure)

extensions/wikia/ArticleComments/classes/ArticleComment.class.php
```diff
@@ -713,29 +729,27 @@ public function editPage() {
...
-	 * @return Array|false TODO: Document what the array contains.
+	 * @return array|bool TODO: Document what the array contains.
 	 */
 	public function doSaveComment( $text, $user, $title = null, $commentId = 0, $force = false, $summary = '', $preserveMetadata = false ) {
 		global $wgTitle;
-		wfProfileIn( __METHOD__ );
 		$metadata = $this->mMetadata;
 
-		$this->load( true );
+		if ( !$this->load( true ) ) {
```
garthwebb, line 738: Actually check the return value of `load`

extensions/wikia/ArticleComments/classes/ArticleCommentsAjax.class.php
```diff
@@ -37,44 +34,51 @@ static public function axSave() {
...
-				$text = self::getConvertedContent( $wgRequest->getVal( 'wpArticleComment' ) );
-				$commentId = $wgRequest->getText( 'id', false );
-				$response = $comment->doSaveComment( $text, $wgUser, $title, $commentId );
-				if ( $response !== false ) {
-					$status = $response[0];
-					$article = $response[1];
-
-					return ArticleComment::doAfterPost( $status, $article, $parentId );
-				}
-			}
```
garthwebb, line 66: Eliminate the deep nesting above by returning early where possible.

extensions/wikia/ArticleComments/classes/ArticleCommentsAjax.class.php
```diff
@@ -37,44 +34,51 @@ static public function axSave() {
...
+		if ( empty( $comment ) ) {
+			return $errorResult;
 		}
 
-		return $result;
+		// Return with error if we can't load the data for this comment
+		if ( !$comment->load( true ) ) {
```
garthwebb, line 57: Actually check the return value.

extensions/wikia/ArticleComments/classes/ArticleComment.class.php
```diff
@@ -197,84 +192,124 @@ static public function newFromId( $id ) {
...
+	private function loadLastRevision() {
+		if ( !empty( $this->mLastRevision ) ) {
+			return true;
+		}
+
+		if ( $this->mLastRevId == $this->mFirstRevId ) {
+			// save one db query by just setting them to the same revision object
+			$this->mLastRevision = $this->mFirstRevision;
+		} else {
+			$this->mLastRevision = Revision::newFromId( $this->mLastRevId );
```
armonr, line 309: Consider replacing all of reads of `$this->mLastRevId` and `$this->mFirstRevId` with their corresponding loader (now that they get cached correctly) so that dependent method calls do not have to get executed in a specific order in order for everything to work.

armonr, line 309: And maybe even rename them? `load*` -> `get*`? :ice_cream:

extensions/wikia/ArticleComments/classes/ArticleComment.class.php
```diff
@@ -392,96 +417,95 @@ public function getData( $master = false ) {
...
 
-				$links['delete'] = $title->getLocalUrl( 'redirect=no&action=delete' );
-			}
+		if ( $canDelete ) {
+			$img = '<img class="remove sprite" alt="" src="' . $wgBlankImgUrl . '" width="16" height="16" />';
+			$buttons[] = $img . '<a href="' . $title->getLocalUrl( 'redirect=no&action=delete' ) . '" class="article-comm-delete">' . wfMsg( 'article-comments-delete' ) . '</a>';
 
-			// due to slave lag canEdit() can return false negative - we are hiding it by CSS and force showing by JS
-			if ( $wgUser->isLoggedIn() && $commentingAllowed ) {
-				$display = $this->canEdit() ? 'test=' : ' style="display:none"';
```
armonr, line 434: Huh, what was this `'test='` for?

armonr:
This code is actually readable now! :beers:

extensions/wikia/ArticleComments/classes/ArticleComment.class.php
garthwebb, line 309: They're called `load` rather than `get` because they load the value into the public member variable and they don't return anything, e.g. they aren't meant to be getters.  I am intentionally not returning any value here specifically because I didn't want to make a getter since these public members are accessed directly throughout this class and several times from other classes.

While making these members private, adding getters that call these load methods, and calling these getters instead would be the right thing, I don't want to make that kind of sweeping change in a ticket where I've already changed a lot of stuff.  Especially since there are a lot of places where wall and comment objects are used but are not typed at all, e.g.

``` php
$aCommentArr['level1']->getTitle();
```

Can you tell what that is?  PHPStorm can't :/.  It happens to be an ArticleComment object.  I don't want to change how we access ArticleComment and end up missing something.

extensions/wikia/ArticleComments/classes/ArticleComment.class.php
garthwebb, line 434: No clue, but the HTML for when a user could edit would be:

``` html
<span class='edit-link'test=><img class=... /><a href=...</span>
```

Which is definitely broken HTML.  My guess is that the browser was just ignoring that extra text.

armonr, line 434: :+1:

extensions/wikia/ArticleComments/classes/ArticleComment.class.php
armonr, line 309: Fair enough!

extensions/wikia/ArticleComments/classes/ArticleComment.class.php
```diff
@@ -197,84 +192,124 @@ static public function newFromId( $id ) {
...
-							$result = false;
-						}
-					}
-				} else {
-					$result = false;
-				}
-			} else {
-				$result = false;
-			}
+		return !empty( $this->mFirstRevId );
```
kvas-damian, line 258: Instead of returning result of operation, maybe we could revisionId to make flow easier to follow?

extensions/wikia/ArticleComments/classes/ArticleComment.class.php
kvas-damian, line 309: I also had a feeling that loader functions make it much harder to follow. I'd still consider using getters that return early cached values.

extensions/wikia/ArticleComments/classes/ArticleCommentsAjax.class.php
```diff
@@ -83,49 +87,52 @@ static public function axEdit() {
...
+		$result['text'] = $comment->editPage();
+
+		if ( ArticleComment::isMiniEditorEnabled() ) {
+			$result['edgeCases'] = MiniEditorHelper::getEdgeCases();
+		}
+
+		$result['emptyMsg'] = wfMessage(
+			'article-comments-empty-comment',
+			$comment->getTitle()->getLocalUrl( 'redirect=no&action=delete' )
+		)->text();
```
kvas-damian: I expect it should be escaped.

kvas-damian: Ohh. It's not so easy because we have raw HTML in the message. If you think it's not in scope of this ticket please create new one to address this issue.

extensions/wikia/ArticleComments/classes/ArticleCommentsAjax.class.php
```diff
@@ -198,36 +200,41 @@ static public function axPost() {
...
+			'counter' => $countAll
+		];
 
-			if ( $parentId ) {
-				$result['parentId'] = $parentId;
-			}
+		if ( F::app()->checkskin( 'wikiamobile' ) ) {
+			$result['counterMessage'] = wfMessage( 'wikiamobile-article-comments-counter' )
+				->params( $countAll )
+				->text();
```
kvas-damian: escaped?

extensions/wikia/ArticleComments/classes/ArticleComment.class.php
garthwebb, line 258: I'm not sure I understand the suggestion?  The idea is that this function will load member {{mFirstRevId}} with the correct value, and let the caller know if it succeeded.  The test for whether it succeeded is whether {{mFirstRevId}} is {{empty}} or not.  Are you suggesting something like this?

``` php
$operationSuccessful = !empty( $this->mFirstRevId );
return $operationSuccessful;
```

extensions/wikia/ArticleComments/classes/ArticleCommentsAjax.class.php
garthwebb: Yah, I tried escaping this at first and it broke the HTML rendering of the final message.  I don't think its in scope, but I'll take a look and see what's involved.

extensions/wikia/ArticleComments/classes/ArticleCommentsAjax.class.php
garthwebb: I didn't update this, just changed indentation.  It doesn't look like this has any HTML in it so seems safe to escape.

extensions/wikia/ArticleComments/classes/ArticleComment.class.php
garthwebb, line 309: Believe me, I really don't like loaders used like this either.  Because PHPStorm can't tell me everywhere these members are accessed (because lots of other code in Wall/Forum/Comments does not specify the types of objects) I'm worried about uses I might miss or functionality that actually depends on code being broken in a certain way, and that trying to account for those situations will blow out the scope of this ticket.

However, if you guys are willing to review an additional, potentially big diff, I can make a go of it.

garthwebb, line 309: Ok nope.  Other classes are way too involved in the inner workings of this class to make these getters work the way they should without doing a MASSIVE re-architecture.  Several classes currently expect to call `load` directly and then do some error checking if it fails.

This means that we can't lazy load `firstRevID` here and `lastRevID` there as needed because there is one place where all this code expects failure to happen, and its not when requesting these values.

Why not load everything when a constructor is called and fail to return an `ArticleComment` object?  Well for one, the `load` method takes a `useMaster` flag that determines how the data is loaded (e.g. master or slave) and we'd have to add that to every constructor as a second, awkward argument.

Also, not every comment gets or needs to be loaded currently.  Lazy loading would fix this but again, current callers are not equipped to handle errors outside `load`.

Also, in one place (`ArticleCommentList`) the code pre-loads a bunch of firstRevIds values from the DB and sets the value directly on an array of `ArticleComment` objects.  This improves performance by not doing several separate lookups but it completely ruins any abstraction.  This functionality would be better as a constructor in `ArticleComment` but it happens deep within some function calls using some nasty `for` loops.

My current updates (with `load*` methods) don't change how the code works.  I basically just took the code from one `load` method and made several smaller `load` methods.  I think this is the most we can change with out having to basically rewrite everything.

PR #9332, opened 2015-12-17
(CE-1396,CE-1853) Sunset the Top and WikiStats extensions

Grunny:
Remove the Top and WikiStats extensions. Their special pages will
now redirect to Special:Insights.


macbre:
:+1:

PR #9333, opened 2015-12-17
Support faster builds

paladox:
https://docs.travis-ci.com/user/workers/container-based-infrastructure/


mszabo-wikia:
seems duplicate of #9269

PR #9334, opened 2015-12-17
Update composer.json

paladox:


mszabo-wikia:
Hey @paladox I'm not sure about the intent of this PR. Can you clarify?

paladox:
Ah sorry, I am not sure what I was doing here as it has been so long.

I am going to close this.

What I think I was doing here is adding php lint from composer like mw does.

mszabo-wikia:
Ah thanks. Sadly that does not work with SMW autoloader :(

PR #9336, opened 2015-12-17
Made the ImpliedRedirect explicitly NOT run when action=submit which …

SeanColombo:
…fixes some problems that showed up recently. For example, if "A:S" existed, we were unable to create "A:S (1984)" which was a common use-case when a song and album shared the same name.

Fixes MAIN-5845


owend:
LGTM.  This will go live as part of the normal thursday release.

PR #9338, opened 2015-12-17
SEO-197 Bring Noindex directives back to robots.txt

gabrys:
When testing the new RobotsTxt extension we omitted the Noindex directives,
as they are not documented to be supported by any search engines.

From experimenting though, it seems they do have an impact. Searching for
"gameofthrones.wikia.com/wiki/Special:WikiActivity" brings the WikiActivity
page as the first result, the summary is missing though
(replaced with "the content was blocked by robots.txt").

We don't want those results in search results at all and this is what
Noindex was doing for us. (Another solution is to not disallow those pages
to be crawled and add a proper META tag, but this means the robots
actually visit those pages and we don't want them to).


owend:
LGTM

PR #9341, opened 2015-12-18
LookupContribsCore: do not join with a (outdated) copy of wikicities.city_list

macbre:
[PLATFORM-1768](https://wikia-inc.atlassian.net/browse/PLATFORM-1768)

`events` table is [cleaned up by the maintenance script](https://github.com/Wikia/app/blob/dev/maintenance/wikia/eventsCleanup.php#L78-78) that removes entries for old wikis. No need to make a join with an outdated copy of `wikicities.city_list` on stats DB.

@Grunny / @adamkarminski / @wladekb


Grunny:
LGTM.

PR #9343, opened 2015-12-18
Reparse unconverted infoboxes template pages

kamilkoterba:
https://wikia-inc.atlassian.net/browse/CE-2835

Haven't found root cause. All ways to add a portable infobox seems to properly add portable data to page props and items no longer appear on nonportable infoboxes list.

Adding one time script to re-parse pages from unconverted infoboxes pages. 
As a result if there's portable infobox, the portable data will be stored in page props due to Links update.

@Wikia/community-engineering, @SebastianMarzjan, @idradm, @RafLeszczynski


adamkarminski:
Should work, run it.

PR #9347, opened 2015-12-21
DAT-3296 handle both space and underscore in file namespace

dianafa:
https://wikia-inc.atlassian.net/browse/DAT-3296

### DIFF
--- a/includes/wikia/helpers/FileNamespaceSanitizeHelper.php
+++ b/includes/wikia/helpers/FileNamespaceSanitizeHelper.php
@@ -37,8 +37,14 @@ class FileNamespaceSanitizeHelper {
 					$fileNamespaces [] = $alias;
 				}
 			}
+
+			$fileNamespaces = array_map(function( $namespace ) {
+				return str_replace('_', '(_|\ )', $namespace);
+			}, $fileNamespaces);
+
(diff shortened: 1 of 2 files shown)


includes/wikia/helpers/FileNamespaceSanitizeHelper.php
```diff
@@ -37,9 +37,15 @@ private function getFilePrefixRegex( $contLang ) {
 					$fileNamespaces [] = $alias;
 				}
 			}
+
+			$fileNamespaces = array_map(function( $namespace ) {
+				return str_replace('_', '(_|\ )', $namespace);
```
sebastianmarzjan: You probably want to consider using multibyte-safe replace here

includes/wikia/helpers/FileNamespaceSanitizeHelper.php
```diff
@@ -37,6 +37,11 @@ private function getFilePrefixRegex( $contLang ) {
 					$fileNamespaces [] = $alias;
 				}
 			}
+
+			$fileNamespaces = array_map(function( $namespace ) {
+				return mb_ereg_replace('_', '(_|\ )', $namespace);
```
sebastianmarzjan, line 43: Please add a comment here :)
(that we want to match user-provided filename that may contain both spaces and underscores)

sebastianmarzjan:
One comment request and GTG

PR #9349, opened 2015-12-21
VE-1919: Remove blur-radius from VE as it was causing rendering issues

pauloslund:
The blur radius on the Visual Editors box shadow was causing VE to not render on newer versions of Microsoft Edge. Removing blur radius entirely has no change to the intended effect, while fixing rendering issues. May also cause an improvement to render times as per http://nerds.airbnb.com/box-shadows-are-expensive-to-paint/


inez:
LGTM as long as it tested well

PR #9350, opened 2015-12-21
Services 719 - ApiClient logging

nmonterroso:
@Wikia/services-team @macbre 
Adding logging for `ApiClient` similar to what's logged in [`HttpFunctions`](https://github.com/Wikia/app/blob/dev/includes/HttpFunctions.php#L77). This is accomplished by extending `ApiClient` rather than directly modifying generated code. For now I used the same mechanism (`WikiaLogger`) to log the performance data, rather than sending the data to influxdb so that it's consistent with what `HttpFunctions` does. @macbre - is there a reason that we don't send this info to influxdb?


lib/Wikia/src/Service/Swagger/ApiProviderModule.php
```diff
@@ -0,0 +1,18 @@
...
+use Wikia\Util\Statistics\BernoulliTrial;
+
+class ApiProviderModule implements Module {
+
+	const API_CLIENT_LOG_SAMPLER = "api_client_log_sampler";
+	const SAMPLE_RATE = 0.1;
+
+	public function configure(InjectorBuilder $builder) {
+		$builder
+				->bind(self::API_CLIENT_LOG_SAMPLER)->to(new BernoulliTrial(self::SAMPLE_RATE));
```
drsnyder: The spacing looks off.

lib/Wikia/src/Service/Swagger/ApiClient.php
```diff
@@ -0,0 +1,60 @@
...
+		$this->serviceName = $serviceName;
+	}
+
+	public function callApi($resourcePath, $method, $queryParams, $postData, $headerParams, $responseType=null) {
+		$start = microtime(true);
+		$response = $exception = null;
+		$code = 200;
+
+		try {
+			$response = parent::callApi($resourcePath, $method, $queryParams, $postData, $headerParams, $responseType);
```
drsnyder, line 32: Does this return the status code? It looks like it will only ever be something other than 200 if an exception is thrown.

drsnyder:
Minor comments. OTW, LGTM

lib/Wikia/src/Service/Swagger/ApiClient.php
nmonterroso, line 32: `$response` is an `array` of the json-decoded response and the headers on success (anything >= 200 && <= 299). While some of our successes are `204`, it's probably ok to log here as `200` to indicate a success. we _could_ write some logic to turn a `200 OK` or a `204 No-Content` string into the appropriate `int`, but I didn't think it mattered too much. do you think we should do that?

lib/Wikia/src/Service/Swagger/ApiProviderModule.php
nmonterroso: 52285f7

lib/Wikia/src/Service/Swagger/ApiClient.php
drsnyder, line 32: Oh, I see. It's probably fine so long as we don't need to search by `statusCode`. 

We might want to take it out since we also have `isOk` which conveys similar information.

nmonterroso, line 32: I'm inclined to keep that field since it keeps this log's fields consistent with what's in [`Http::request`](https://github.com/Wikia/app/blob/dev/includes/HttpFunctions.php#L77)

PR #9351, opened 2015-12-21
Add a temporary fix for using both 'secret' and 'token' to authenticate via the Schwartz

drsnyder:
Add a temporary fix for using both `secret` and `token` to validate the Schwartz token. We use `secret` in the `Helios` extension and `token` elsewhere. The next step will be to move helios and other clients to use `token`.

https://wikia-inc.atlassian.net/browse/SERVICES-1072

@Wikia/services-team 

There are a few other references to `token`:
- [LogEventsApiController](https://github.com/Wikia/app/blob/dev/includes/wikia/api/LogEventsApiController.class.php#L29)
- [ApiFetchBlob](https://github.com/Wikia/app/blob/dev/includes/api/wikia/ApiFetchBlob.php#L63)
- [WikiaApiCreatorReminderEmai](https://github.com/Wikia/app/blob/dev/extensions/wikia/CreateNewWiki/WikiaApiCreatorReminderEmail.php#L37)

### DIFF
--- a/extensions/wikia/Helios/HelperController.class.php
+++ b/extensions/wikia/Helios/HelperController.class.php
@@ -11,6 +11,9 @@ use Wikia\Service\User\Auth\AuthService;
 class HelperController extends \WikiaController
 {
 
+	const SCHWARTZ_PARAM = 'secret';
+	const EXTERNAL_SCHWARTZ_PARAM = 'token';
+
 	protected $authService;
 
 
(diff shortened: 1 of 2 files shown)


nmonterroso:
Is there a reason not to convert the other places you mentioned to use the newly-introduced constant? Other than that LGTM

drsnyder:
> Is there a reason not to convert the other places you mentioned to use the newly-introduced constant?

I'm still thinking over the right place to put that. I think I want to extract this into `Wikia\Service\User\Auth`. I don't want to over engineer it but I'm beginning to think we should at least consolidate this functionality. It's done a little differently everywhere.

PR #9356, opened 2015-12-22
Combined SEO changes for easier code review

gabrys:
- SEO-199 Don't HTML encode the title coming from WikiaHtmlTitle

(Also removed an unused getter method)
- SEO-194 Fix main page HTML title on Mercury

Don't add article name in front of the wiki name in the HTML title exported
by the API used by Mercury.
- SEO-181 Fix removal of hreflang link to self

The code checks the language code and removes the link from the structure
based on that language code, so there's no link to an English article on an
English article (no point linking to self).

The problem is the code checked the user language instead of the content
language and this commit fixes it by fetching the language from wgTitle.


ghost:
:+1:

PR #9357, opened 2015-12-23
SOC-1743

garthwebb:
Added sorting to the Title, edits and last edited columns of the UserActivity table.  Also cleaned up the LookupUserContribs extension that this code relies upon.

https://wikia-inc.atlassian.net/browse/SOC-1743


extensions/wikia/LookupContribs/LookupContribs.php
```diff
@@ -1,8 +1,71 @@
 <?php
-define ( 'LOOKUPCONTRIBS_HELP', 'This special page can be used to display user\'s contributions on multiple wikias. For displaying the list, give username below.' ) ;
```
garthwebb, line 2: LOOKUPCONTRIBS_HELP isn't used, deleted

extensions/wikia/LookupContribs/LookupContribs.php
```diff
@@ -1,8 +1,71 @@
...
-define ( 'LOOKUPCONTRIBS_PATH', '/' ) ;
```
garthwebb, line 3: LOOKUPCONTRIBS_PATH is useless, deleted.

extensions/wikia/LookupContribs/LookupContribs.php
```diff
@@ -1,8 +1,71 @@
...
-/* cache? no cache? */
-define ( 'LOOKUPCONTRIBS_NO_CACHE', 'false' ) ;
```
garthwebb, line 5: This double negative LOOKUPCONTRIBS_NO_CACHE doesn't really need to be here.  We should always cache.  If there is a problem with caching, we should fix it.

extensions/wikia/LookupContribs/LookupContribs.php
```diff
@@ -1,8 +1,71 @@
...
 
-require_once ( $IP . LOOKUPCONTRIBS_PATH . "extensions/wikia/LookupContribs/SpecialLookupContribs.php" ) ;
-require_once ( $IP . LOOKUPCONTRIBS_PATH . "extensions/wikia/LookupContribs/SpecialLookupContribs_hooks.php" ) ;
```
garthwebb, line 8: Not sure why we need to require these and separately.  Pulled the contents of these two files into this single file and deleted the other two.

extensions/wikia/LookupContribs/SpecialLookupContribs_ajax.php
```diff
@@ -59,12 +46,14 @@ function axData() {
...
-			$activity = $oLC->checkUserActivity( $lookupUser, $order );
+			$oLC->setOrder( $order );
+			$activity = $oLC->getUserActivity();
+
 			if ( !empty( $activity ) ) {
 				$result['iTotalRecords'] = intval( $limit );
 				$result['iTotalDisplayRecords'] = intval( $activity['cnt'] );
 
-				if ( $lookupUser === true ) {
+				if ( $lookupUser ) {
```
garthwebb, line 56: Don't do a trivial boolean check on a boolean.

extensions/wikia/LookupContribs/SpecialLookupContribs_ajax.php
```diff
@@ -59,12 +46,14 @@ function axData() {
 		if ( empty( $mode ) ) {
 			$oLC->setLimit( $limit );
 			$oLC->setOffset( $offset );
-			$activity = $oLC->checkUserActivity( $lookupUser, $order );
```
garthwebb, line 62: Eliminated the $lookupUser argument which was always false.

extensions/wikia/LookupContribs/SpecialLookupContribs_ajax.php
```diff
@@ -122,7 +111,7 @@ public static function prepareLookupContribsData( $activityData ) {
 				$row['dbname'], // wiki dbname
 				$row['title'], // wiki title
 				$row['url'], // wiki url
-				F::app()->wg->Lang->timeanddate( wfTimestamp( TS_MW, $row['last_edit'] ), true ), // last edited
+				F::app()->wg->Lang->timeanddate( wfTimestamp( TS_MW, $row['lastedit'] ), true ), // last edited
```
garthwebb, line 114: Renamed so that this is the same value as what is used by the ajax requests

extensions/wikia/LookupContribs/SpecialLookupContribs_ajax.php
```diff
@@ -146,8 +135,8 @@ public static function prepareLookupUserData( $activityData, $username ) {
 				$row['id'], // wiki Id
 				$row['title'], // wiki title
 				$row['url'], // wiki url
-				$wg->Lang->timeanddate( wfTimestamp( TS_MW, $row['last_edit'] ), true ), // last edited
-				$wg->ContLang->formatNum( $row['editcount'] ),
+				$wg->Lang->timeanddate( wfTimestamp( TS_MW, $row['lastedit'] ), true ), // last edited
+				$wg->ContLang->formatNum( $row['edits'] ),
```
garthwebb, line 139: Again, renamed so that this is the same value as what is used by the ajax requests

extensions/wikia/LookupContribs/SpecialLookupContribs_helper.php
```diff
@@ -53,14 +48,14 @@ public function setDBname( $dbname = '' ) {
 	}
 
 	public function getDBname() {
-		return LC_TEST ? F::app()->wg->DBname : $this->mDBname;
```
garthwebb, line 56: Why the test?  Removing useless debugging code.

extensions/wikia/LookupContribs/LookupContribs.php
kvas-damian, line 8: We usually try to avoid putting any code in extension setup file. I think we should follow this guideline and move `LookupContribsHooks` to it's own file.

extensions/wikia/LookupContribs/SpecialLookupContribs_helper.php
```diff
@@ -259,129 +232,51 @@ private function addWikiaInfo( $wikiaIds, &$userActivity ) {
...
-			elseif ( preg_match( '/^url:(asc|desc)$/', $order, $aMatches ) ) {
-				foreach ( $data['data'] as $aItem ) {
-					$aTemp[$aItem['url']] = $aItem;
-				}
-			}
-
-			// order by last edit
-			elseif ( preg_match( '/^lastedit:(asc|desc)$/', $order, $aMatches ) ) {
-				foreach ( $data['data'] as $aItem ) {
-					// added URL part since last edits are ints and might not be unique
```
kvas-damian, line 303: Are we sure we want to remove URL from sorting key?
I completely have no idea if it was done for purpose...

extensions/wikia/AssetsManager/config.php
```diff
@@ -2736,3 +2736,17 @@
 		'//extensions/wikia/PolldaddyTag/scripts/wikiamobile.js'
 	]
 ];
+
+$config['special_user_activity_js'] = [
+	'type' => AssetsManager::TYPE_JS,
+	'assets' => [
+		'//extensions/wikia/UserActivity/js/UserActivity.js',
```
kvas-damian, line 2738: Can't find this file

extensions/wikia/AssetsManager/config.php
```diff
@@ -2736,3 +2736,17 @@
...
+	],
+];
+
+$config['special_user_activity_scss'] = [
+	'type' => AssetsManager::TYPE_SCSS,
+	'assets' => [
+		'//extensions/wikia/UserActivity/css/UserActivity.scss',
```
kvas-damian, line 2745: Can't find this file

extensions/wikia/AssetsManager/config.php
garthwebb, line 2738: Whoops, should be there now.

extensions/wikia/AssetsManager/config.php
garthwebb, line 2745: Should be there now.

extensions/wikia/LookupContribs/SpecialLookupContribs_helper.php
garthwebb, line 303: The only reason the URL was combined with `last_edit` (and `editcount` below) is because of the extremely awkward way this sorting was being done.

I'm assuming the author only knew how to sort on an array's key or an array with simple scalar values.  To sort based on what they knew, they built a new temporary array `$aTemp` who's keys were a value from the complex (array of arrays) data structure, e.g.:

``` php
$data = [
    [ 'url' => 'http://abc', 'title' => 'abc', 'last_edit' => '20150101' ],
    [ 'url' => 'http://xyz', 'title' => 'xzy', 'last_edit' => '20150102' ],
    ...
];
```

becomes:

``` php
$aTemp = [
    '20150101' => [ 'url' => 'http://abc', 'title' => 'abc', 'last_edit' => '20150101' ],
    '20150102' => [ 'url' => 'http://xyz', 'title' => 'xzy', 'last_edit' => '20150102' ],
]
```

However, some of these values were identical (e.g. multiple pages could have the same `last_edit` date), in which case each next identical value would overwrite the last in `$aTemp`. To prevent this and include all values in `$aTemp`, they appended something that would make it unique, e.g. the URL.

Using `usort` allows us to sort the original array directly without creating a temporary array and then loading it back into the original and doesn't require that the sort key be unique in the set.

extensions/wikia/LookupContribs/LookupContribs.php
garthwebb, line 8: Fine!  ;)

extensions/wikia/LookupContribs/SpecialLookupContribs_hooks.php
```diff
@@ -1,30 +1,28 @@
...
+	 * @param User $user
+	 *
+	 * @return bool
+	 */
 	static public function ArticleSaveComplete ( $article, User $user ) {
-		global $wgDBname, $wgMemc, $wgSharedDB, $wgUser ;
-		/* unset the key for this user on this database */
-		$username = $user->getName () ;
-		$wgMemc->delete ( "$wgSharedDB:LookupContribs:normal:$username:$wgDBname" ) ;
-		$wgMemc->delete ( "$wgSharedDB:LookupContribs:final:$username:$wgDBname" ) ;
```
garthwebb, line 22: I looked as far back as Aug 26, 2008 (https://github.com/Wikia/app/commit/cf58c10fdc79fc97253e80fc970dbea2b670d243) and I could not find anywhere that actually set these memcached keys.  I can tell that `normal` and `final` are values used in `LookupContribsCore::mMode` and there is a memkey that uses `mMode`, but it looks nothing like this key.  This is definitely a great example of why one should never build their own cache keys rather than using a common method.

I updated this to clear the only `LookupContribsCore` cache possible.  The other cache value includes limit and offset values and would require clearing an impossible to know number of keys.  In this case I think its fine to let the TTL expire this key, which is what has been happening for at least 7 years anyway.

extensions/wikia/LookupContribs/SpecialLookupContribs_helper.php
```diff
@@ -259,129 +232,59 @@ private function addWikiaInfo( $wikiaIds, &$userActivity ) {
...
-			[ 'count(distinct wiki_id) as num' ],
-			$where,
-			__METHOD__
-		);
-
-		$row = $dbr->fetchObject( $res );
-		return $row->num;
+	/**
+	 * Clear the data cached by method getUserActivity
+	 */
```
garthwebb, line 237: Added this as a helper method for the hooks class to use.

extensions/wikia/LookupContribs/LookupContribs.php
kvas-damian, line 8: Thanks ;)

extensions/wikia/UserActivity/js/UserActivity.js
```diff
@@ -0,0 +1,73 @@
...
+			})
+			.click(function () {
+				updateOrderFor('title');
+			});
+		$('#edits-header')
+			.addClass(function () {
+				return updateHighlightFor('edits');
+			})
+			.click(function () {
+			updateOrderFor('edits');
```
kvas-damian: Please fix indentation

extensions/wikia/UserActivity/js/UserActivity.js
```diff
@@ -0,0 +1,73 @@
...
+	'use strict';
+
+	$(function () {
+		var orderMatch = /order=([^:]+):(desc|asc)/i,
+			results,
+			curOrder,
+			curDirection,
+			curLocation = window.location;
+
+		results = curLocation.href.match(orderMatch);
```
kvas-damian: For safety maybe we could use `window.location.search` here?

extensions/wikia/UserActivity/js/UserActivity.js
```diff
@@ -0,0 +1,73 @@
...
+		$('#last-edit-header')
+			.addClass(function () {
+				return updateHighlightFor('lastedit');
+			})
+			.click(function () {
+			updateOrderFor('lastedit');
+		});
+
+		function updateHighlightFor(column) {
+			if (typeof curOrder === 'undefined' && column === 'lastedit') {
```
kvas-damian: Maybe we should initialize `curOrder` and `column` with defaults to avoid this special condition here?

extensions/wikia/UserActivity/js/UserActivity.js
```diff
@@ -0,0 +1,73 @@
...
+				// Default to ascending for title, descending for everything else
+				curDirection = column === 'title' ? 'asc' : 'desc';
+			}
+
+			orderParam = 'order=' + column + ':' + curDirection;
+
+			if (typeof curOrder === 'undefined') {
+				newURL = curLocation.href + (curLocation.href.indexOf('?') === -1 ? '?' : '&') + orderParam;
+			} else {
+				newURL = curLocation.href.replace(orderMatch, orderParam);
```
kvas-damian: What do you think about using QueryString library to simplify URL operations?
https://github.com/Wikia/app/blob/dev/resources/wikia/modules/spec/querystring.spec.js#L306-L311

extensions/wikia/LookupContribs/SpecialLookupContribs_helper.php
kvas-damian, line 303: Ohh, didn't catch those values are used as keys. It definitely makes more sense to use `usort`.
Thanks for detailed explanation.

extensions/wikia/LookupContribs/SpecialLookupContribs_hooks.php
kvas-damian, line 22: Makes sense

extensions/wikia/UserActivity/js/UserActivity.js
garthwebb: Just what I wanted... much better!

extensions/wikia/UserActivity/UserActivityController.class.php
```diff
@@ -58,16 +54,16 @@ public function index() {
 	protected function formatItems( $items ) {
 		$flattened = [];
 		foreach ( $items as $sortKey => $contribItem ) {
-			$lastEditTS = wfTimestamp( TS_MW, $contribItem[ 'last_edit' ] );
+			$lastEditTS = wfTimestamp( TS_MW, $contribItem[ 'lastedit' ] );
+			unset($contribItem['lastedit']);
```
armonr: What is the `'lastedit'` key vs `'lastEdit'` (line 61)? A :bug: magnet? :smiley_cat:

extensions/wikia/UserActivity/UserActivityController.class.php
```diff
@@ -58,16 +54,16 @@ public function index() {
...
+
 			$localizedDate = \F::app()->wg->Lang->timeanddate( $lastEditTS, $localTZ = true );
 			$contribItem['lastEdit'] = $localizedDate;
-			unset($contribItem['last_edit']);
 
-			$editCount = $contribItem['editcount'];
+			$editCount = $contribItem['edits'];
+			unset($contribItem['edits']);
```
armonr: Could the existing keys in the helper be leveraged to be used here too?

extensions/wikia/UserActivity/UserActivityController.class.php
garthwebb: I'm trying to update these to our naming conventions but I see what you mean.  I've updated to create a new array rather than alter the key slightly.

extensions/wikia/UserActivity/UserActivityController.class.php
garthwebb: I don't think I understand this suggestion?

armonr: To be clearer, I was referring to the [constants here](https://github.com/Wikia/app/pull/9357/files#diff-b28e18ad904e400ccc6038595de3ce0bR8). It would be neat if they could be in a common place an accessed in this class too, but not a big deal.

kvas-damian:
LGTM

extensions/wikia/UserActivity/css/UserActivity.scss
```diff
@@ -0,0 +1,19 @@
...
+
+	&.down {
+	  @include arrow(6px, $color-buttons, 'down');
+	}
+  }
+
+  .sortable {
+	cursor: pointer;
+  }
+}
```
lizlux, line 23: LGTM

PR #9361, opened 2015-12-24
SEO-146 "In other languages" section for Mercury

gabrys:
Using lilly as the source for the "In other languages" section in Mercury. Also populating the <head> with <link hreflang> links using this mapping.

See also Wikia/mercury#1813

### DIFF
--- a/extensions/wikia/MercuryApi/MercuryApiController.class.php
+++ b/extensions/wikia/MercuryApi/MercuryApiController.class.php
@@ -580,7 +580,7 @@ class MercuryApiController extends WikiaController {
 		}
 
 		$url = $title->getFullURL();
-		$url = str_replace( '.rychu.wikia-dev.com', '.wikia.com', $url );
+		//$url = str_replace( '.rychu.wikia-dev.com', '.wikia.com', $url );
 
 		$lilly = new Lilly();
 		$links = $lilly->getCluster( $url );
--- a/extensions/wikia/MercuryApi/MercuryApiController.class.php
+++ b/extensions/wikia/MercuryApi/MercuryApiController.class.php
@@ -391,6 +391,11 @@ class MercuryApiController extends WikiaController {
 			}
 			$data['htmlTitle'] = $titleBuilder->getTitle();
 
+			$otherLanguages = $this->getOtherLanguages( $title );
+			if ( !empty( $otherLanguages ) ) {
+				$data['otherLanguages'] = $otherLanguages;
+			}
+
 		} catch ( WikiaHttpException $exception ) {
 			$this->response->setCode( $exception->getCode() );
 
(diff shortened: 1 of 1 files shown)


extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -566,4 +571,45 @@ private function getWikiaStatsData() {
 
 		return $wikiDetails['stats'];
 	}
+
+	private function getOtherLanguages( Title $title ) {
+		global $wgEnableLillyExt;
```
nandy-andy, line 576: Lilly :girl: :smile_cat:

extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -566,4 +571,45 @@ private function getWikiaStatsData() {
...
+
+		if ( empty( $wgEnableLillyExt ) ) {
+			return null;
+		}
+
+		$url = $title->getFullURL();
+		$url = str_replace( '.rychu.wikia-dev.com', '.wikia.com', $url );
```
nandy-andy: Is it necessary? Is it some kind of left over after development? What about other devboxes than `dev-rychu`? :)

extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -566,4 +571,45 @@ private function getWikiaStatsData() {
...
+		$url = $title->getFullURL();
+		//$url = str_replace( '.rychu.wikia-dev.com', '.wikia.com', $url );
+
+		$lilly = new Lilly();
+		$links = $lilly->getCluster( $url );
+		if ( !count( $links ) ) {
+			return null;
+		}
+
+		// Remove link to self
```
rogatty, line 591: Shouldn't this be done before `if ( !count( $links ) )`? What if there is only one link - to self?

extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -566,4 +571,45 @@ private function getWikiaStatsData() {
...
+
+	private function getOtherLanguages( Title $title ) {
+		global $wgEnableLillyExt;
+
+		if ( empty( $wgEnableLillyExt ) ) {
+			return null;
+		}
+
+		$url = $title->getFullURL();
+		//$url = str_replace( '.rychu.wikia-dev.com', '.wikia.com', $url );
```
rogatty, line 583: This isn't nice :see_no_evil:

extensions/wikia/MercuryApi/MercuryApiController.class.php
gabrys, line 591: A link cluster is always at least two links.

extensions/wikia/MercuryApi/MercuryApiController.class.php
gabrys, line 583: Prototype-quality. (Lilly is a prototype as long as we stick to Star Wars communities): easy to test on one devbox without having the whole logic for handling different environments.

PR #9362, opened 2015-12-28
SEO-146 Remove "In other languages" section from mobile skin

gabrys:
This is only triggered on Star Wars wikis (where wgEnableLillyExt is
on). Only the Star Wars wikis are known to generate this section as a
part of the content (vs using the regular interwiki links).

There's code in Mercury to generate the section based on links from
Lilly (again for Star Wars wikis only).

See also Wikia/mercury#1813 and Wikia/app#9361


rogatty:
LGTM

PR #9363, opened 2015-12-29
Just for comparison

kamilkoterba:


extensions/wikia/InsightsV2/InsightsLoopController.class.php
```diff
@@ -0,0 +1,203 @@
...
+		];
+	}
+
+	/**
+	 * Get params to generate link to insight main page in notification template
+	 */
+	private function getInsightLinkParams() {
+		return [
+			'insightsPageText' => wfMessage( 'insights-notification-see-more' )->plain(),
+			'insightsPageLink' => SpecialPage::getTitleFor( 'Insights' )
```
kamilkoterba: It returns Title object and previously here was string, so probably you're missing `SpecialPage::getTitleFor(...)->getLocalUrl()` here.

kamilkoterba:
LGTM

PR #9364, opened 2016-01-04
(CE-3194) Fix LookupContribs data for users with quotes in their names

Grunny:
No need to HTML escape data sent in an AJAX request, which results in quotes
in usernames being HTML encoded before being sent in the request.


macbre:
LGTM

PR #9366, opened 2016-01-04
make infoboxes use params same way as mediawiki, enable 0 param

idradm:


extensions/wikia/PortableInfobox/services/Parser/Nodes/Node.php
```diff
@@ -197,7 +197,7 @@ protected function getInfoboxData( $key ) {
 	protected function extractDataFromSource( \SimpleXMLElement $xmlNode ) {
 		$source = $this->getXmlAttribute( $xmlNode, self::DATA_SRC_ATTR_NAME );
 
-		return ( !empty( $source ) ) ? $this->getInfoboxData( $source )
+		return ( !empty( $source ) || $source == '0' ) ? $this->getInfoboxData( $source )
```
sebastianmarzjan, line 200: Do you think a loose check like that is sufficient, or an explicit strict equality check for both 0 (int) and '0' (string) would be safer?

extensions/wikia/PortableInfobox/controllers/PortableInfoboxParserTagController.class.php
```diff
@@ -67,7 +67,7 @@ public static function replaceInfoboxMarkers( &$parser, &$text ) {
 	 * @throws InvalidInfoboxParamsException when unsupported attributes exist in params array
 	 */
 	public function render( $markup, Parser $parser, PPFrame $frame, $params = null ) {
-		$infoboxNode = Nodes\NodeFactory::newFromXML( $markup, $this->getFrameParams( $frame ) );
+		$infoboxNode = Nodes\NodeFactory::newFromXML( $markup, $frame->getArguments() );
```
sebastianmarzjan, line 70: yay for simplicty! :+1:

extensions/wikia/PortableInfobox/tests/PortableInfoboxParserTagControllerTest.php
```diff
@@ -178,17 +186,55 @@ public function testGetLayoutDataProvider() {
...
+
+		$marker = $this->controller->renderInfobox( $text, [ ], $this->parser,
+			$this->parser->getPreprocessor()->newCustomFrame( $params ) )[ 0 ];
+		$output = $this->controller->replaceMarkers( $marker );
+
+		$result = [ ];
+		$xpath = $this->getXPath( $output );
+		$dataNodes = $xpath->query( '//aside/div[contains(@class,\'pi-data\')]' );
+		for ( $i = 0; $i < $dataNodes->length; $i++ ) {
+			$result[ $xpath->query( 'h3[contains(@class, \'pi-data-label\')]', $dataNodes->item( $i ) )->item( 0 )->nodeValue ] =
```
sebastianmarzjan: The string literals here provide little information without the context of naming; A comment or extracting them to consts would be very helpful

extensions/wikia/PortableInfobox/services/Parser/Nodes/Node.php
idradm, line 200: it's a consistent approach we use for data checks

sebastianmarzjan:
LGTM

PR #9368, opened 2016-01-04
Insights v2 refactoring

kamilkoterba:
https://wikia-inc.atlassian.net/browse/CE-3195

Refactoring started by @lukaszkonieczny during hackathon.
It covers better code organization, single responsibility concept and implementation of strategy pattern for insights pages.

**from Łukasz's email:**
here's what's changed: https://github.com/Wikia/app/compare/insights-v2...insights-v2-refactoring?expand=1

I think there is still some room to improvement for example:
1. Move code connect with Loop Notification outside models. For example each model has different method to check if item is fixed. Maybe we can create separate class (which implements some interface) to handle that for each model? What do you think?
2. How to handle additional actions (like adding button to convert infobox).
3. How to setup Insights list easly. Right now we have different global variables and setup files. Maybe we can simplify it somehow?

Looking on all this things together and adding such as special pages created to fill querycache table with data (for example class PagesWithoutInfobox) maybe we can change also file structure like:

-- models
---- Insights Type
------ InsightsTypeModel extends InsightsModel
------ InsightsTypeLoopNotification extends InsightsLoopNotification
------ InsightsTypeActions (?)
------ InsightsTypeConfig (?)
------ InsightsType.setup
------ specials/InsightsTypeQueryPage
------ template/InsightsTemplate

What do you think team? I would love to hear you feedback and suggestions :)

@Wikia/community-engineering


extensions/wikia/InsightsV2/models/InsightsFlagsModel.php
```diff
@@ -0,0 +1,121 @@
...
+			$subtype = $this->getDefaultType();
+		}
+
+		/* If still empty (no flag types on wikia) return empty list */
+		if ( empty( $subtype ) ) {
+			return [];
+		}
+
+		/* Get to list of pages marked with flags */
+		$flaggedPages = $app->sendRequest(
```
kamilkoterba, line 62: Needs try catch block

extensions/wikia/InsightsV2/models/InsightsFlagsModel.php
```diff
@@ -0,0 +1,121 @@
...
+
+		return $pages;
+	}
+
+	private function getSubTypes() {
+		$subtypes = [];
+		$app = F::app();
+
+		$params = [ 'flag_targeting' => \Flags\Models\FlagType::FLAG_TARGETING_CONTRIBUTORS ];
+		$flagTypes = $app->sendRequest( 'FlagsApiController', 'getFlagTypes', $params )->getData()['data'];
```
kamilkoterba: Needs try catch block

extensions/wikia/InsightsV2/models/InsightsFlagsModel.php
```diff
@@ -0,0 +1,121 @@
...
+		return $subtypes;
+	}
+
+	/**
+	 * Select first type ID and use as default
+	 */
+	private function getDefaultType() {
+		$app = F::app();
+		$params = [ 'flag_targeting' => \Flags\Models\FlagType::FLAG_TARGETING_CONTRIBUTORS ];
+		$flagTypes = $app->sendRequest( 'FlagsApiController', 'getFlagTypes' , $params )->getData()['data'];
```
kamilkoterba: Needs try catch block

extensions/wikia/InsightsV2/models/InsightsUnconvertedInfoboxesModel.php
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Class InsightsNonportableInfoboxesModel
+ * A class specific to a subpage with a list of pages
+ * without categories.
+ */
+class InsightsUnconvertedInfoboxesModel extends InsightsQueryPageModel {
+	const INSIGHT_TYPE = 'nonportableinfoboxes';
+
+	private static $insightConfig = [
+		'displayFixItMessage' => false,
```
kamilkoterba: Now it's 'displayfix' instead of 'displayFixItMessage' right?

extensions/wikia/InsightsV2/models/InsightsModel.php
```diff
@@ -0,0 +1,64 @@
...
+abstract class InsightsModel {
+	use Loggable;
+
+	const INSIGHTS_FLOW_URL_PARAM = 'insights';
+
+	private $template = 'subpageList';
+	protected $config;
+	protected $subtype;
+
+	abstract public function fetchArticlesData();
```
kamilkoterba, line 33: Document interface of abstract method.

extensions/wikia/InsightsV2/helpers/InsightsItemData.php
```diff
@@ -0,0 +1,122 @@
+<?php
+
+class InsightsItemData {
```
kamilkoterba, line 3: What do you think about adding namespaces?
eg. Wikia/Insights/Helpers/ItemData

extensions/wikia/InsightsV2/InsightsHelper.php
```diff
@@ -0,0 +1,200 @@
...
+		return $highlightedInsights;
+	}
+
+	/**
+	 * Returns a full URL for a known subpage and a NULL for an unknown one.
+	 * @param string $subpage slug of subpage
+	 * @param array $params params
+	 * @return string|null
+	 */
+	public static function getSubpageLocalUrl( $subpage = false, Array $params = [] ) {
```
kamilkoterba: Default value for `$subpage` doesn't make any sense here. Calling `getSubpageLocalUrl` without subpage parameter returns just null, so each usage of this method should provide a `$subpage` param.

kamilkoterba: minor: array type is not an object - can use small caps

extensions/wikia/InsightsV2/templates/Insights_flagsFiltering.php
```diff
@@ -0,0 +1,15 @@
...
+				foreach ( $flagTypes as $flagType ):
+				$selected = ( $flagType['flag_type_id'] == $selectedFlagTypeId );
+			?>
+				<option value="<?= Sanitizer::encodeAttribute( $flagType['flag_type_id'] ); ?>" <?= $selected ? 'selected="selected"' : ''; ?>><?= htmlspecialchars( $flagType['flag_name'] ) ?></option>
+			<?php
+				endforeach;
+			?>
+		</select>
+	</form>
+</div>
```
kamilkoterba: I see that this template is not used anymore. Is that on purpose?

extensions/wikia/Insights/PagesWithoutInfobox.setup.php
```diff
@@ -4,7 +4,7 @@
  * Insights - Pages without an infobox
  *
  * CAUTION
- * BELOW CLASSES HAVE DEPENDENCIES ON CLASSES FROM Insights.setup.php - remember to load them as well
+ * BELOW CLASSES HAVE DEPENDENCIES ON CLASSES FROM InsightsV2.setup.php - remember to load them as well
```
kamilkoterba: All changes within Insights directory wasn't intended, right? (Insights to InsightsV2 in comments in setup files)

extensions/wikia/InsightsV2/InsightsContext.class.php
```diff
@@ -0,0 +1,118 @@
...
+			$sortedIds = $this->sortData();
+			$content = $this->sliceData( $sortedIds );
+		}
+
+		return $content;
+	}
+
+	public function fetchData() {
+		$cacheKey = ( new InsightsCache( $this->config ) )->getMemcKey( InsightsCache::INSIGHTS_MEMC_ARTICLES_KEY );
+		$articlesData = WikiaDataAccess::cache( $cacheKey, InsightsCache::INSIGHTS_MEMC_TTL, function () {
```
kamilkoterba, line 35: After moving caching the list of articles away from fetchArticlesData to fetchData we lost caching on other usages of fetchArticlesData
e.g. getCount is used in couple of places and it's calling fetchArticlesData https://github.com/Wikia/app/blob/insights-v2-refactoring/extensions/wikia/InsightsV2/services/InsightsCountService.class.php#L73

extensions/wikia/InsightsV2/helpers/InsightsPageViews.php
```diff
@@ -0,0 +1,112 @@
...
+			$articlesData[ $articleId ]['metadata']['pvDiff'] = $pvDiff;
+
+		}
+
+		$start = microtime(true);
+
+		( new InsightsSorting( $this->config ) )->createSortingArrays( $sortingData );
+
+		$end = microtime(true) - $start;
+		var_dump($end);
```
kamilkoterba: Remove debugging code

extensions/wikia/InsightsV2/InsightsController.class.php
```diff
@@ -0,0 +1,115 @@
...
+		parent::__construct( 'Insights', 'insights', true );
+	}
+
+	/**
+	 * The main, initializing function
+	 * @throws MWException
+	 */
+	public function index() {
+		wfProfileIn( __METHOD__ );
+		$this->wg->Out->setPageTitle( wfMessage( 'insights' )->escaped() );
```
Grunny: `setPageTitle` is safe, so no need to escape the message passed to it. Also, you should be able to make use of `$this->getOutput()->setPageTitle( $this->msg( 'insights' )->plain() );`

extensions/wikia/InsightsV2/InsightsHelper.php
```diff
@@ -0,0 +1,200 @@
...
+	 * 'insights-notification-message-fixed-wantedpages'
+	 * 'insights-notification-message-fixed-withoutimages',
+	 */
+	const INSIGHT_FIXED_MSG_PREFIX = 'insights-notification-message-fixed-';
+
+	private static $defaultInsights = [
+		InsightsUncategorizedModel::INSIGHT_TYPE	=> 'InsightsUncategorizedModel',
+		InsightsWithoutimagesModel::INSIGHT_TYPE	=> 'InsightsWithoutimagesModel',
+		InsightsDeadendModel::INSIGHT_TYPE			=> 'InsightsDeadendModel',
+		InsightsWantedpagesModel::INSIGHT_TYPE		=> 'InsightsWantedpagesModel'
```
Grunny, line 68: Weird whitespace going on in this array. ;)

extensions/wikia/InsightsV2/InsightsHooks.class.php
```diff
@@ -0,0 +1,147 @@
...
+
+class InsightsHooks {
+
+	/**
+	 * Check if article is in insights flow and init script to show banner with message and next steps
+	 */
+	public static function onBeforePageDisplay( OutputPage &$out, Skin &$skin ) {
+		global $wgRequest;
+
+		$subpage = $wgRequest->getVal( 'insights', null );
```
Grunny: You can avoid the global with `$out->getRequest()->getVal(...);`.

extensions/wikia/InsightsV2/InsightsHooks.class.php
```diff
@@ -0,0 +1,147 @@
+<?php
+
+class InsightsHooks {
+
+	/**
+	 * Check if article is in insights flow and init script to show banner with message and next steps
+	 */
+	public static function onBeforePageDisplay( OutputPage &$out, Skin &$skin ) {
```
Grunny: No need for the `&` as objects are effectively passed by reference in PHP anyway.

extensions/wikia/InsightsV2/InsightsLoopController.class.php
```diff
@@ -0,0 +1,201 @@
...
+	/**
+	 * Get params to generate next item link in notification template
+	 *
+	 * @param String $subpage Insights subpage
+	 * @param array $next An array with a url and text for a link to the next item
+	 * @return array
+	 */
+	private function getInsightNextLinkParams( $subpage, $next ) {
+		return [
+			'nextArticleText' => wfMessage( 'insights-notification-next-item-' . $subpage )->plain(),
```
Grunny: For consistency, this prefix should probably also have a constant in the helper class.

extensions/wikia/InsightsV2/scripts/LoopNotification.js
```diff
@@ -0,0 +1,136 @@
...
+				if (isFixed) {
+					loopNotificationTracking.successTrack();
+				}
+			}
+		};
+
+		getMessageType = function() {
+			if (isEdit || !isFixed) {
+				return 'warn';
+			} else {
```
Grunny: I think it always looks cleaner if you remove the unnecessary else block when there is a return in the if (same in the method below), i.e.

``` javascript
getMessageType = function() {
    if (isEdit || !isFixed) {
        return 'warn';
    }
    return 'confirm';
}
```

extensions/wikia/InsightsV2/models/InsightsFlagsModel.php
lukaszkonieczny, line 62: Fixed here: https://github.com/Wikia/app/commit/2f44d8cf47dbaa64dd3ff9a0a49551d52e779f16

extensions/wikia/InsightsV2/InsightsV2.setup.php
```diff
@@ -0,0 +1,110 @@
...
+$wgHooks['wgQueryPages'][] = 'InsightsHooks::onwgQueryPages';
+$wgHooks['AfterUpdateSpecialPages'][] = 'InsightsHooks::onAfterUpdateSpecialPages';
+
+$wgExtensionFunctions[] = 'wfInsightsSetup';
+function wfInsightsSetup() {
+	global $wgRequest, $wgHooks;
+
+	if ( !empty( $wgRequest->getVal( 'insights', null ) ) ) {
+		$wgHooks['GetLocalURL'][] = 'InsightsHooks::onGetLocalURL';
+	}
```
Grunny: Minor, but I think it might be slightly cleaner to set a init method in the InsightsHooks classes that registers the hooks and has this logic, rather than having logic and global function in the setup file itself.

extensions/wikia/InsightsV2/InsightsHooks.class.php
```diff
@@ -0,0 +1,147 @@
...
+			$out->addScriptFile('/extensions/wikia/InsightsV2/scripts/InsightsLoopNotificationTracking.js');
+		}
+
+		return true;
+	}
+
+	/**
+	 * Add insight param to keep information about flow after edit
+	 */
+	public static function AfterActionBeforeRedirect( Article $article, &$sectionanchor, &$extraQuery ) {
```
Grunny: Methods should be lowerCamelCase, so `onAfterActionBeforeRedirect` or whatnot.

extensions/wikia/InsightsV2/helpers/InsightsItemData.php
lukaszkonieczny, line 3: Maybe we can consider this idea in next steps?

extensions/wikia/InsightsV2/InsightsContext.class.php
lukaszkonieczny, line 35: InsightsCountService fixed here: https://github.com/Wikia/app/commit/4ecdde3fe59fec0a2d5573483051de943c9a2b76

Did you find any other places?

extensions/wikia/InsightsV2/models/InsightsUnconvertedInfoboxesModel.php
lukaszkonieczny: Fixed here: https://github.com/Wikia/app/commit/4e858944b3dc387755003f07cecdcb0f14189ec0

kamilkoterba:
LGTM

kamilkoterba:
LGTM

PR #9373, opened 2016-01-05
PLATFORM-1725 Division by zero in imageServing.class

RafalWilinski:
https://wikia-inc.atlassian.net/browse/PLATFORM-1725

Right now, there's possibility of having both `this->proportion['h']` and `this->proportion['w']` equal to zero. This should fix this problem (division operation in line 405).


wladekb:
lgtm

PR #9374, opened 2016-01-05
MAIN-6140 Add minimum ResourceLoader cache @jj

fraszczakszymon:


ghost:
:+1:

macbre:
:+1:

inez:
@fraszczakszymon Could you please add more explanation about this change? What is it supposed to do (as opposed to what is happening currently) and is it temporary (given hardcoded value).

fraszczakszymon:
@inez Issue with details: MAIN-6140.
Before fixing apache issue (https://github.com/Wikia/chef-repo/pull/7690) many resources loaded by ResourceLoader have been cached with wrong page for 24h (__load / load.php was redirected to 404 page instead of serving real resource). Because most of JS libs were broken, our AdEngine didn't worked properly and ads weren't served to anon users. Here we wanted to bump "Last modified" flag of all resources, but also invalidating fastly cache was needed - https://github.com/Wikia/wikia-vcl/pull/698. After discussion with team we decided to keep this constant in ResourceLoader, it can be useful for similar issues in future.

PR #9377, opened 2016-01-05
SEO-75 Replace Our404Handler extension with a simpler approach

gabrys:
A two-line redirect-canon.php PHP script will do the same as the whole extension. We'll just need to point ErrorDocument 404 to this file in Apache config.

The problem is it would discard the URL params. So let's log the query strings to find out if they are worth keeping.


owend:
Had a discussion offline, this lgtm.

PR #9389, opened 2016-01-07
SEO-213 Fix cropping of the images in the Special:NewFiles

gabrys:
The constructor for WikiaPhotoGallery was not called and apparently it
sets some important defaults!

Also HTML-escaping href attributes for the links below the images.


extensions/wikia/WikiaNewFiles/WikiaNewFilesGallery.class.php
```diff
@@ -3,12 +3,9 @@
 class WikiaNewFilesGallery extends WikiaPhotoGallery {
 	const ANCHOR_LENGTH = 60;
 
-	/**
```
bkoval: Why remove phpdoc? :)

bkoval:
LGTM :+1:

PR #9391, opened 2016-01-07
(CE-3211) Rename UserBadges page to ProfileTags

Grunny:
The UserBadges page was renamed to ProfileTags to avoid confusion and
needs to be updated here as well. Also, update the instructions to use
a pipe instead of a colon.


kamilkoterba:
LGTM

PR #9394, opened 2016-01-08
DAT-3614 Always pass an *ARRAY* of params to the infobox NodeFactory

sebastianmarzjan:
https://wikia-inc.atlassian.net/browse/DAT-3614

### DIFF
--- a/extensions/wikia/PortableInfobox/controllers/PortableInfoboxParserTagController.class.php
+++ b/extensions/wikia/PortableInfobox/controllers/PortableInfoboxParserTagController.class.php
@@ -67,7 +67,8 @@ class PortableInfoboxParserTagController extends WikiaController {
 	 * @throws InvalidInfoboxParamsException when unsupported attributes exist in params array
 	 */
 	public function render( $markup, Parser $parser, PPFrame $frame, $params = null ) {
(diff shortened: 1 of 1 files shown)


dianafa:
LGTM

inez:
`PHP Fatal Error: Argument 2 passed to Wikia\PortableInfobox\Parser\Nodes\NodeFactory::newFromXML() must be of the type array, boolean given, called in /extensions/wikia/PortableInfobox/controllers/PortableInfoboxParserTagController.class.php on line`

In what case getArguments() returns boolean? I see when it would return array or null, but not a boolean.

dianafa:
`getArguments()` here: https://github.com/Wikia/app/blob/dev/includes/parser/Preprocessor_DOM.php#L1718 will return whatever we will pass to the constructor. We were passing there `false` if the args list in api call ( `/api.php?action=infobox` ) was empty - that's the reason of having false in getArguments. Follow-up PR: https://github.com/Wikia/app/pull/9402 thanks!

inez:
So the change done within this PR won't be needed after merging DAT-3614-2 and can be reverted.

dianafa:
Well, if
- `render()` is used in 3 different places so far
- and we already know that bad argument passed to `newFromXML()` can cause fatals
- and we already have prevention for this which doesn't cost us anything (1 line of code more)

I'd leave check introduced in this PR

PR #9402, opened 2016-01-09
make getFrameArguments return always an array

dianafa:
follow-up to the https://github.com/Wikia/app/pull/9394 as this is probably a reason why the error occured

https://wikia-inc.atlassian.net/browse/DAT-3622


extensions/wikia/PortableInfobox/controllers/ApiPortableInfobox.class.php
```diff
@@ -107,11 +108,17 @@ public function getVersion() {
 	}
 
 	/**
-	 * @return mixed
+	 * @return array of key-value args
```
inez, line 111: PHPDOC format is `@return datatype description` so description shoudn't be `"of key-value..."` but just `"key-value..."`
Also - seems like it got split into two lines too early

PR #9405, opened 2016-01-11
VE-2117: Track Parsoid HIT ratio

inez:
We are tracking it at a VisualEditor API level because this is the ratio that we are interested in - and VisualEditor is not the only client of Parsoid.


macbre:
:+1:

PR #9406, opened 2016-01-11
Script to move js pages in main namespace to new namespace

lukaszkonieczny:
Script to move js pages in main namespace to new namespace.
Mostly to move scripts on dev.wikia.com


extensions/wikia/ContentReview/maintenance/moveScriptsToNamespace.php
```diff
@@ -0,0 +1,94 @@
+<?php
+
+$dir = __DIR__ . "/../../../../";
+require_once( $dir . 'maintenance/Maintenance.php' );
+
+class MoveScripts extends Maintenance {
+
+	const JS_FILE_EXTENSION = '.js';
```
Grunny, line 10: This is unused.

Grunny:
Minor comment, but LGTM as a maintenance script. :)

PR #9411, opened 2016-01-11
(CE-2503) Add zhstarter to list of starter wikias

Grunny:
Add zhstarter to list of starter wikias so that newly created Chinese wikias
will get initial content from zh.starter.wikia.com.


lukaszkonieczny:
LGTM

PR #9413, opened 2016-01-11
ADEN-2805 Enable Evolve2 provider

fraszczakszymon:
Evolve turn on globally and narrowed by other "country variable".


nandy-andy:
LGTM :+1:

PR #9419, opened 2016-01-12
ADEN-2815 Evolve2 section param

fraszczakszymon:
Please do CR for single commits:
- fixed calculation of section param for Evolve2 provider (now it's based on `s0` and `s0v` params)
- changed `getMappedVertical` method name to `getName` because it's more connected to dbname

### DIFF
--- a/extensions/wikia/AdEngine/js/provider/evolve2.js
+++ b/extensions/wikia/AdEngine/js/provider/evolve2.js
@@ -1,10 +1,10 @@
 /*global define*/
 define('ext.wikia.adEngine.provider.evolve2', [
-	'ext.wikia.adEngine.evolveHelper',
 	'ext.wikia.adEngine.provider.gpt.helper',
 	'ext.wikia.adEngine.slotTweaker',
+	'ext.wikia.adEngine.utils.adLogicZoneParams',
 	'wikia.log'
-], function (evolveHelper, gptHelper, slotTweaker, log) {
+], function (gptHelper, slotTweaker, zoneParams, log) {
 	'use strict';
 
 	var logGroup = 'ext.wikia.adEngine.provider.evolve2',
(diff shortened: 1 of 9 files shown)


extensions/wikia/AdEngine/js/spec/provider/evolve2.spec.js
```diff
@@ -1,33 +1,41 @@
 /*global beforeEach, describe, it, modules, expect, spyOn*/
-describe('OpenX Provider targeting', function () {
```
nandy-andy, line 2: Good eye :bird:

extensions/wikia/AdEngine/js/provider/evolve2.js
```diff
@@ -17,7 +17,6 @@ define('ext.wikia.adEngine.provider.evolve2', [
 			'320x50,320x100,300x250': 'a',
 			'300x250': 'a'
 		},
-		section = evolveHelper.getSect(),
```
nandy-andy, line 20: It seems we don't use `getSect()` anylonger. Shouldn't we remove it from `evolveHelper`?

nandy-andy:
LGTM :sunglasses:

extensions/wikia/AdEngine/js/provider/evolve2.js
fraszczakszymon, line 20: It is used by previous evolve provider:
https://github.com/Wikia/app/blob/dev/extensions/wikia/AdEngine/js/provider/evolve.js#L32

We'll be able to remove whole evolveHelper once we decide to move on Evolve2

nandy-andy, line 20: Oh, OK. I might've done something wrong while grepping I didn't spot that. Thanks for update!

PR #9420, opened 2016-01-12
(CE-3223) Fix fatal error in InsightsV2

Grunny:
Fix fatal error in InsightsV2 due to outdated calls to cache methods
on the model rather than the cache class.

```
PHP Fatal Error: Call to undefined method
InsightsTemplatesWithoutTypeModel::updateInsightsCache()
in /extensions/wikia/InsightsV2/InsightsHooks.class.php on line 147
```

### DIFF
--- a/extensions/wikia/Flags/controllers/FlagsApiController.class.php
+++ b/extensions/wikia/Flags/controllers/FlagsApiController.class.php
@@ -566,14 +566,13 @@ class FlagsApiController extends FlagsApiBaseController {
 		if ( !empty( $wgEnableInsightsExt ) ) {
 			$flagsIds = $this->prepareFlagTypesIds( $flags );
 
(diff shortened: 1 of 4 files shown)


adamkarminski:
LGTM

extensions/wikia/Flags/controllers/FlagsApiController.class.php
```diff
@@ -570,10 +570,11 @@ private function purgeFlagInsights( Array $flags = [], $pageId = null ) {
 
 			foreach ( $flagsIds as $flagId ) {
 				$insightsFlagsModel->initModel( [ 'flagTypeId' => $flagId ] );
```
lukaszkonieczny: `initModel` method was removed as well

lukaszkonieczny:
LGTM

PR #9423, opened 2016-01-12
CE-3224: Insights methods cleanup

lukaszkonieczny:
Cleanup some bad method usage after Insights refactoring


Grunny:
LGTM. Though I'd consider logging the bad title data in case we want to clean it up.

PR #9425, opened 2016-01-12
PLATFORM-1801: update the CategoryExhibitionSection preference only when needed

macbre:
i.e. when we do change the value and it's not a default one

Avoid useless updates and user settings propagation across clusters (via `ExternalUser_Wikia::linkToLoca`) and fix a P2 - [PLATFORM-1801](https://wikia-inc.atlassian.net/browse/PLATFORM-1801).

@wladekb / @drozdo

### DIFF
--- a/extensions/wikia/CategoryExhibition/CategoryExhibitionSection.class.php
+++ b/extensions/wikia/CategoryExhibition/CategoryExhibitionSection.class.php
@@ -140,8 +140,11 @@ class CategoryExhibitionSection {
 
 		if ( in_array( $sortType, $this->allowedSortOptions ) ) {
 			if ( !$wgUser->isAnon() ) {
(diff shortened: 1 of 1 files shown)


jcellary:
As my experience with MW code is very limited my suggestion probably doesn't make much sense, but an alternative approach would be to introduce a flag like "dirty" in the user object and set it to true in setters and then when saveChanges is invoked only make an update if object was modified. That would allow for a more generic solution instead of a get and compare before each update of user data, I don't know however if that covers all cases and is easily doable.

Otherwise LGTM.

macbre:
@jcellary: your suggestion makes perfect sense.

However :) it will not cover the case with a default value that's used in `getPreferences` calls here. And apparently URLs with default values passed are the most common here.

PR #9428, opened 2016-01-12
SEO-217 Fix Our404Handler for non-English wikis

gabrys:
Return false when we want to prevent the redirect to the canonical URL
for Our404Handler special page

### DIFF
--- a/extensions/wikia/Our404Handler/SpecialOur404Handler_body.php
+++ b/extensions/wikia/Our404Handler/SpecialOur404Handler_body.php
@@ -67,4 +67,26 @@ class Our404HandlerPage extends UnlistedSpecialPage {
 		header( "X-Redirected-By: Our404Handler" );
 		header( sprintf( "Location: %s", $oTitle->getFullURL( $query ) ), true, 301 );
 	}
+
+	/**
+	 * This hook is called when about to force a redirect to a canonical URL
+	 * for a title when we have no other parameters on the URL.
+	 *
+	 * Return false when we want to prevent the redirect to the canonical URL
+	 * for Our404Handler special page (for non-English wikis)
+	 *
+	 * @see PLATFORM-811, SEO-217
+	 *
+	 * @param WebRequest $request
+	 * @param Title $title
+	 * @param OutputPage $output
+	 * @return bool
+	 */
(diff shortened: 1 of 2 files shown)


extensions/wikia/Our404Handler/SpecialOur404Handler_body.php
```diff
@@ -67,4 +67,26 @@ public function doRender404() {
 		header( "X-Redirected-By: Our404Handler" );
 		header( sprintf( "Location: %s", $oTitle->getFullURL( $query ) ), true, 301 );
 	}
+
+	/**
+	 * This hook is called when about to force a redirect to a canonical URL
+	 * for a title when we have no other parameters on the URL.
+	 *
+	 * Return false when we want to prevent the redirect to the canonical URL
+	 * for Our404Handler special page (for non-English wikis)
```
ghost, line 76: How do you distinguish between English and non-English wikis?

gabrys, line 76: Oh, basically the Apache directs the request to Special:Our404Handler, which is the canonical URL and this hook is never called. The hook is called for cases when Special:Our404Handler is called, but the canonical URL for it is other like Spezial:Our404Handler (German naming of the namespace).

extensions/wikia/Our404Handler/SpecialOur404Handler_body.php
```diff
@@ -67,4 +67,26 @@ public function doRender404() {
...
+	 * @see PLATFORM-811, SEO-217
+	 *
+	 * @param WebRequest $request
+	 * @param Title $title
+	 * @param OutputPage $output
+	 * @return bool
+	 */
+	public static function onTestCanonicalRedirect( WebRequest $request, Title $title, OutputPage $output) {
+		if ( $title->isSpecial( self::NAME ) ) {
+			return false;
```
ghost, line 87: Then the execution flow might fall to https://github.com/Wikia/app/blob/ed16f45be1bbefd8f52e7dfb753d3f084d93995b/includes/Wiki.php#L289 - is it OK?

gabrys, line 87: Precisely. The expected outcome is the special page is rendered without being redirected to the canonical URL first.

ghost:
Thanks for explanation. :+1:

PR #9430, opened 2016-01-12
SOC-1857 Clean up the module before starting work

garthwebb:
https://wikia-inc.atlassian.net/browse/SOC-1857

### DIFF
--- a/extensions/wikia/UserProfilePageV3/models/FavoriteWikisModel.class.php
+++ b/extensions/wikia/UserProfilePageV3/models/FavoriteWikisModel.class.php
@@ -83,11 +83,7 @@ class FavoriteWikisModel extends WikiaModel {
 		];
 	}
 
-	private function getTopWikisQuery( $limit = null ) {
-		if ( is_null( $limit ) ) {
-			$limit = self::MAX_FAV_WIKIS;
-		}
-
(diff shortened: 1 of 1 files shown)


kvas-damian:
LGTM

extensions/wikia/UserProfilePageV3/models/FavoriteWikisModel.class.php
```diff
@@ -27,11 +27,13 @@ class FavoriteWikisModel extends WikiaModel {
 	 * @param User $user
 	 */
 	public function __construct( User $user ) {
+		parent::__construct();
```
jsutterfield, line 36: Minor: Is there a reason to call the parent constructor here? It doesn't look like we're using any of the helper values it initializes in this PR or in the other one.

garthwebb, line 36: I think its good form unless there is a specific reason not to.  Additionally, I often see code that does:

``` php
$wg = \F::app()->wg;
```

in classes that would give them that variable if they'd just inherited from parent.

jsutterfield, line 36: Sounds good :+1:

jsutterfield:
LGTM

PR #9431, opened 2016-01-12
SOC-1857 Cache results from getTopWikis query

garthwebb:
Implemented caching of often run query.  Note that this builds off cleanup pull request #9430.

https://wikia-inc.atlassian.net/browse/SOC-1857

### DIFF
--- a/extensions/wikia/UserProfilePageV3/models/FavoriteWikisModel.class.php
+++ b/extensions/wikia/UserProfilePageV3/models/FavoriteWikisModel.class.php
@@ -83,11 +83,7 @@ class FavoriteWikisModel extends WikiaModel {
 		];
 	}
 
-	private function getTopWikisQuery( $limit = null ) {
-		if ( is_null( $limit ) ) {
-			$limit = self::MAX_FAV_WIKIS;
-		}
-
+	private function getTopWikisQuery( $limit = self::MAX_FAV_WIKIS ) {
 		$query = (new WikiaSQL())
 			->SELECT()
 				->FIELD( 'wiki_id' )
@@ -120,11 +116,9 @@ class FavoriteWikisModel extends WikiaModel {
 		}
 
 		$memKey = $this->getTopWikisMemKey();
-		$savedWikis = $this->wg->Memc->get( $memKey );
-		if ( empty( $savedWikis ) ) {
-			$savedWikis = $this->getTopWikisFromDb();
-			$this->wg->Memc->set( $memKey, $savedWikis, self::SAVED_WIKIS_TTL );
-		}
+		$savedWikis = WikiaDataAccess::cache( $memKey, self::SAVED_WIKIS_TTL, function() {
+			return $this->getTopWikisFromDb();
+		});
 
 		$wikis = array_merge( $savedWikis, $this->getEditsWikis() );
 
--- a/extensions/wikia/UserProfilePageV3/models/FavoriteWikisModel.class.php
+++ b/extensions/wikia/UserProfilePageV3/models/FavoriteWikisModel.class.php
@@ -1,8 +1,10 @@
 <?php
+
+/**
+ * Class FavoriteWikisModel
+ */
 class FavoriteWikisModel extends WikiaModel {
-	/**
-	 * @var User|null
-	 */
+	/** @var User|null */
 	private $user;
 
 	/**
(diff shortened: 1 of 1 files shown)


extensions/wikia/UserProfilePageV3/models/FavoriteWikisModel.class.php
```diff
@@ -33,67 +39,71 @@ public function __construct( User $user ) {
...
+
+		$wikis[$wikiId] = [
+			'id' => $wikiId,
+			'wikiName' => $wikiName,
+			'wikiUrl' => $wikiUrl,
+			'edits' => $editCount
+		];
+	}
+
+	private function getTopWikisQuery( $limit = null ) {
```
kvas-damian: what do you think about `$limit = self::MAX_FAV_WIKIS`?

extensions/wikia/UserProfilePageV3/models/FavoriteWikisModel.class.php
```diff
@@ -105,11 +115,18 @@ private function getTopWikisFromDb( $limit = null ) {
...
 	public function getTopWikis( $refreshHidden = false ) {
 
-		if ( $refreshHidden === true ) {
+		if ( $refreshHidden ) {
 			$this->clearHiddenTopWikis();
 		}
 
-		$wikis = array_merge( $this->getTopWikisFromDb(), $this->getEditsWikis() );
+		$memKey = $this->getTopWikisMemKey();
+		$savedWikis = $this->wg->Memc->get( $memKey );
```
kvas-damian: What do you think about using `WikiaDataAccess::cache()` here?

kvas-damian: I think we could use `SharedMemcKey` for better performance.

extensions/wikia/UserProfilePageV3/models/FavoriteWikisModel.class.php
garthwebb: Sounds better

extensions/wikia/UserProfilePageV3/models/FavoriteWikisModel.class.php
garthwebb: The `getTopWikisMemKey()` method uses `wfSharedMemKey()`. Is that what you mean?

extensions/wikia/UserProfilePageV3/models/FavoriteWikisModel.class.php
```diff
@@ -33,67 +39,67 @@ public function __construct( User $user ) {
...
-		if ( is_null( $limit ) ) {
-			$limit = self::MAX_FAV_WIKIS;
-		}
+	private function getTopWikisFromDb( $limit = self::MAX_FAV_WIKIS ) {
+		global $wgCityId, $wgSpecialsDB;
 
 		if ( empty( $wgCityId ) ) {
 			// staff/internal does not have stats db
 			$wikis = [];
-		} else if ( $wgDevelEnvironment ) {
```
jsutterfield, line 52: Do we not have to worry about this `$wgDevelEnvironment` anymore? Looks like it didn't make it into the new version.

extensions/wikia/UserProfilePageV3/models/FavoriteWikisModel.class.php
kvas-damian: Right, I misread the code ;)

kvas-damian:
LGTM

extensions/wikia/UserProfilePageV3/models/FavoriteWikisModel.class.php
garthwebb, line 52: They were using $wgDevelEnvironment as a way to cook the data in dev for what was essentially guided unit tests.  The `getTestData` method returned a set of preselected wikis, one of which was one that should be filtered.  This really belongs in a unit test somehwere.  Also, there wasn't any "strange behavior" that I could tell, so I thought it was best to get rid of this fake code path.

jsutterfield, line 52: Okay. Makes sense.

jsutterfield:
LGTM

PR #9433, opened 2016-01-12
ADEN-2712 Refactor floating medrec

drets:
Re-writing floating medrec feature (adding new AMD module), extending floating medrec to `300x600` size.


extensions/wikia/AdEngine/liftium/AdsInContent.js
```diff
@@ -1,211 +0,0 @@
...
-			!isCorporate &&
-			(isArticle || isSearchPage);
-
-	} catch (e) {
-		AIC2.disabledReason = e;
-		return false;
-	}
-}());
-
-AIC2.init = function() {
```
fraszczakszymon, line 37: Clean up please:
https://github.com/Wikia/app/blob/dev/extensions/wikia/Rail/scripts/Rail.js#L82

drets, line 37: :+1:

extensions/wikia/AdEngine/js/slot/floatingMedrec.js
```diff
@@ -0,0 +1,86 @@
...
+	var logGroup = 'ext.wikia.adEngine.slot.floatingMedrec';
+
+	log('load', 'debug', logGroup);
+
+	function init() {
+		var context = adContext.getContext(),
+			positionFixed = false,
+			margin = 10,
+			slotName = 'INCONTENT_BOXAD_1',
+			$slot = $('<div class="wikia-ad" style="position: relative;"></div>').attr('id', slotName),
```
fraszczakszymon: Is this style attr needed?

extensions/wikia/AdEngine/js/slot/floatingMedrec.js
```diff
@@ -0,0 +1,86 @@
...
+		$placeHolder.append($slot);
+
+		globalNavigationHeight = $globalNavigation.height();
+		lastAdHeight = $slot.height();
+
+		startPosition = parseInt($placeHolder.offset().top, 10) - globalNavigationHeight - margin;
+
+		stopPosition = Math.min(
+			parseInt($footer.offset().top, 10),
+			parseInt($leftSkyscraper3.offset().top, 10)) - globalNavigationHeight - 2 * margin - lastAdHeight;
```
fraszczakszymon: It will throw an exception when there is no LEFT_SKYSCRAPER_3, am I wrong?

extensions/wikia/AdEngine/js/slot/floatingMedrec.js
```diff
@@ -0,0 +1,86 @@
...
+
+		function update() {
+			if (lastAdHeight !== $slot.height()) {
+				stopPosition -= $slot.height() - lastAdHeight;
+				lastAdHeight = $slot.height();
+			}
+
+			if (positionFixed && win.scrollY <= startPosition) {
```
fraszczakszymon: Could you remind me why do we need `positionFixed` flag?

extensions/wikia/AdEngine/js/slot/floatingMedrec.js
drets: `Uncaught TypeError: Cannot read property 'top' of undefined` :+1:

skins/oasis/css/core/ads.scss
```diff
@@ -228,14 +228,9 @@ body >a >img[src$="noad.gif"] {
 	}
 }
 
-// Floating medrec
+// Floating medrec place holder
 #WikiaAdInContentPlaceHolder {
 	position: absolute;
-	#INCONTENT_BOXAD_1 {
```
fraszczakszymon, line 234: Isn't `position` needed here? (instead of inside style attr)

extensions/wikia/AdEngine/js/AdContext.js
```diff
@@ -131,6 +131,13 @@ define('ext.wikia.adEngine.adContext', [
 			!noExternals
 		);
 
+		// Floating medrec
+		context.opts.floatingMedrec = !!(
+			context.opts.showAds && context.opts.adsInContent &&
+			(context.targeting.pageType === 'article' || context.targeting.pageType === 'search') &&
```
nandy-andy: We've got three conditions that includes `context.targeting.pageType === 'article'`. What do you think about creating a small helper function in `AdContext` called `isArticle()` and reusing it in these conditions?

extensions/wikia/AdEngine/js/run/desktop.run.js
```diff
@@ -142,6 +143,10 @@ require([
 		if (exitstitial) {
 			exitstitial.init();
 		}
+
+		if (floatingMedrec) {
+			win.wgAfterContentAndJS.push(floatingMedrec.init);
```
nandy-andy: I'd rather change this line to `floatingMedrec.init()` to keep consistent with the rest since it's being executed after DOM loads anyway (lines 152-156). And I think it's important to init after DOM loads because having content and JS isn't good enough since there were issues where users' JS were breaking the functionality.

extensions/wikia/AdEngine/js/slot/floatingMedrec.js
```diff
@@ -0,0 +1,88 @@
...
+
+			if (win.scrollY >= stopPosition) {
+				$slot.css({
+					position: 'absolute',
+					top: stopPosition - startPosition + 'px'
+				});
+			}
+		}
+
+		win.addEventListener('scroll', adHelper.throttle(update, 100));
```
nandy-andy: Was is choppy with default throttle?

extensions/wikia/AdEngine/js/slot/floatingMedrec.js
```diff
@@ -0,0 +1,88 @@
...
+		stopPosition = stopPoint - globalNavigationHeight - 2 * margin - lastAdHeight;
+
+		function update() {
+			if (lastAdHeight !== $slot.height()) {
+				stopPosition -= $slot.height() - lastAdHeight;
+				lastAdHeight = $slot.height();
+			}
+
+			if (win.scrollY <= startPosition) {
+				$slot.css({
```
nandy-andy: How about just adding CSS classes instead of modifying CSS with JS ~~(here and in next 2 places below)~~? Below it's impossible because of calculation but here it could be a class like `.floating-ad-default`.

extensions/wikia/AdEngine/js/slot/floatingMedrec.js
```diff
@@ -0,0 +1,88 @@
+/*global define*/
+define('ext.wikia.adEngine.slot.floatingMedrec', [
```
nandy-andy, line 2: Looks much lighter. Sometimes I have the feeling too light :japanese_ogre: The notes below can be used as QA notes or/and as something to check before moving to testing phase. 

I can't find RTL support in the new implementation. Does it work out-of-the-box and doesn't need code implementation?

~~I can see in the new code calculations connected to global navigation but in the old code it counted only when global nav was fixed. How does it change behavior of the ad on high screens when the global nav isn't fixed?~~ Oh, it's always fixed as far as I can quickly check now... :smile_cat:

skins/oasis/css/core/ads.scss
```diff
@@ -232,8 +232,6 @@ body >a >img[src$="noad.gif"] {
 #WikiaAdInContentPlaceHolder {
 	position: absolute;
 	#INCONTENT_BOXAD_1 {
-		height: 250px;
-		width: 300px;
```
nandy-andy, line 236: And this is a tricky part... If I'm correct those sizes here were needed for good calculations. Now, before we get the ad from DFP we don't know if the height will be `250` or `600`. Without this sizes the calculations are always wrong because are made for height which equals `0`. It means `floatingMedrec` will calculate and decide the ad fits if it's 300x250 it'll probably fit but the higher one may not...

Let's think about how to solve it and we can talk after daily tomorrow. Unless you know how - then implement it please! :smiley_cat:

drets:
(talked about @nandy-andy's comments offline, nice CR :+1: )

nandy-andy:
Ditto :tangerine: However, the 2nd round of CR should be done by other dev from the team as we talked at daily today. @drets please add a short description to the PR as our [guidelines say](https://github.com/Wikia/guidelines/blob/master/Process/PullRequests.md#guidelines) :japanese_ogre:

drets:
It's the 2nd round.
But I am fine with 3d round from @jacekjursza :wink:

nandy-andy:
+1 :trolleybus:

extensions/wikia/AdEngine/js/AdContext.js
```diff
@@ -131,6 +135,13 @@ define('ext.wikia.adEngine.adContext', [
 			!noExternals
 		);
 
+		// Floating medrec
+		context.opts.floatingMedrec = !!(
+			context.opts.showAds && context.opts.adsInContent &&
+			(isArticle() || context.targeting.pageType === 'search') &&
```
jacekjursza: Good point about moving this to isArticle function. What do you think about unifying the approach and creating one function to cover all page types?
for example:
isPageType('article'), isPageType('search') etc.

jacekjursza:
One minor thing from me :)

extensions/wikia/AdEngine/js/AdContext.js
drets: :+1:

drets:
Any LGTM guys? :smirk_cat:

nandy-andy:
LGTM :sunglasses: Let's test it! :japanese_ogre:

drets:
:whale:

nandy-andy:
The failing unit test isn't connected to changes in this branch.

fraszczakszymon:
LGTM :+1:

PR #9434, opened 2016-01-12
SEO-210 Replace Our404Handler with a simple PHP script

gabrys:
Introducing a file called redirect-canonical.php.

It handles URLs that miss the /wiki/ prefix. Instead of just adding /wiki/ in front
of the URLs, the full canonical URL is constructed (using $title->getFullURL).
This limits number of redirects and redirects to the best possible URL in one run.

This is a simple replacement for Our404Handler and should be configured as follows in
Apache config:

```
ErrorDocument 404 /redirect-canonical.php
```

### DIFF
--- a/redirect-canonical.php
+++ b/redirect-canonical.php
@@ -1,0 +1,32 @@
+<?php
+
+/**
+ * redirect-canonical.php
+ *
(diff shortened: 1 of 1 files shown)


owend:
LGTM

PR #9435, opened 2016-01-13
(CE-3212) Change SWM tracking to also send events to GA

Grunny:
Change SWM tracking to also send events to GA in addition to DW.


lukaszkonieczny:
LGTM

PR #9436, opened 2016-01-13
(CE-3225) Redirect Special:Top and Special:WikiStats to Insights

Grunny:
This was previously done for the original Insights extension but the
changes didn't make it into InsightsV2.


lukaszkonieczny:
LGTM

PR #9437, opened 2016-01-13
PLATFORM-1734: i18n for Content Moderator and Discussion Moderator

macbre:
Extends #9297


macbre:
Backported to preview in 90432b9

PR #9438, opened 2016-01-13
ADEN-1558 Change the way how arrays are rendered in WikiFactory

fraszczakszymon:
Simplify the way of rendering array objects in WikiFactory. Instead of rendering:

```
array (
  0 => 'foo', 
  1 => 'bar',
)
```

now it will look like:

```
["foo","bar"]
```

Associative array shouldn't be affected so `array(15=>'foo', 20=>'bar')` will be rendered as:

```
array (
  15 => 'foo', 
  20 => 'bar',
)
```

### DIFF
--- a/extensions/wikia/WikiFactory/WikiFactory.php
+++ b/extensions/wikia/WikiFactory/WikiFactory.php
@@ -3344,14 +3344,16 @@ class WikiFactory {
 		global $$name;
 		global $preWFValues;
 
+		$value = "";
+
 		if( isset( $preWFValues[$name] ) ) {
 			// was modified, spit out saved default
-			return self::parseValue( $preWFValues[$name], $type );
+			$value = self::parseValue( $preWFValues[$name], $type );
 		} elseif( isset( $$name ) ) {
 			// was not modified, spit out actual value
-			return self::parseValue( $$name, $type );
+			$value = self::parseValue( $$name, $type );
 		}
-		return "";
+		return htmlspecialchars( $value );
 	}
 
 	/**
@@ -3368,8 +3370,8 @@ class WikiFactory {
 		if ( !isset( $variable->cv_value ) ) {
 			return "";
 		}
-
-		return self::parseValue( unserialize( $variable->cv_value ), $variable->cv_variable_type );
+		$value = self::parseValue( unserialize( $variable->cv_value ), $variable->cv_variable_type );
+		return htmlspecialchars( $value );
 	}
 
 	/**
--- a/extensions/wikia/WikiFactory/WikiFactory.php
+++ b/extensions/wikia/WikiFactory/WikiFactory.php
@@ -3329,4 +3329,73 @@ class WikiFactory {
 		$wgMemc->prefetch($keys);
 	}
 
+	/**
+	 * Renders community's value of given variable
+	 *
+	 * @access public
+	 * @static
+	 *
+	 * @param string $name name of wg variable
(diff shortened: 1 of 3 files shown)


fraszczakszymon:
Ping @macbre because you are watching this issue on JIRA :)

macbre:
LGTM, thanks :)

nandy-andy:
LGTM :+1:

extensions/wikia/WikiFactory/templates/variable.tmpl.php
```diff
@@ -99,19 +93,15 @@
 
 <?php elseif( $variable->cv_variable_type == "integer"): ?>
 
-	<input type="text" name="varValue" id="varValue" value="<?php echo unserialize( $variable->cv_value ) ?>" size="40" maxlength="255" />
+	<input type="text" name="varValue" id="varValue" value="<?php echo WikiFactory::renderValue( $variable ) ?>" size="40" maxlength="255" />
```
gabrys, line 96: it doesn't seem like a regression, but HTML-escaping the variables is missing is the template

fraszczakszymon, line 96: Fixed. Thanks :gem:

nandy-andy, line 96: Before:
wiki-factory-before
After:
wiki-factory-after

extensions/wikia/WikiFactory/WikiFactory.php
```diff
@@ -3329,4 +3329,75 @@ static public function prefetchWikisById( $ids, $what = self::PREFETCH_DEFAULT )
...
+	 * Renders wg variable
+	 *
+	 * @access public
+	 * @static
+	 *
+	 * @param object $variable parameter passed to variable.tmpl.php
+	 *
+	 * @return string
+	 */
+	static public function renderValue( $variable ) {
```
inez, line 3369: Name of this method (and description) is not accurate. It does not render anything, it just returns string.

extensions/wikia/WikiFactory/WikiFactory.php
```diff
@@ -3329,4 +3329,75 @@ static public function prefetchWikisById( $ids, $what = self::PREFETCH_DEFAULT )
...
+	 * @access public
+	 * @static
+	 *
+	 * @param string $name name of wg variable
+	 * @param string $type type of variable ($variable->cv_variable_type)
+	 *
+	 * @return string
+	 */
+	static public function renderValueOnCommunity( $name, $type ) {
+		global $$name;
```
inez, line 3344: Have you considered using $GLOBALS[$name] instead?

extensions/wikia/WikiFactory/WikiFactory.php
nandy-andy, line 3369: That's right! What are your suggestion for the new name?

inez, line 3369: getValue or getEscapedValue, I personally wouldn't put htmlspecialchars in there (but in template perhaps, even if it would have to be there few times).

nandy-andy, line 3369: I like `getEscapedValue()` when I thought about it at first ridiculously long names came to my mind. This one is short and descriptive :smile_cat: Thanks for the feedback! We'll see on Monday if we can work on it quickly.

PR #9443, opened 2016-01-13
CE-3139: Display information is infobox portable or not

lukaszkonieczny:
Display information is infobox portable or not on Special:Templates


adamkarminski:
LGTM

azarek:
LGTM

PR #9444, opened 2016-01-13
CE-3234: Fix global user script imports

lukaszkonieczny:
Fix global user script imports on wikis with disabled js support


lukaszkonieczny:
CR done in https://github.com/Wikia/app/pull/9439

PR #9447, opened 2016-01-13
SOC-1867 Make sure we use the user IDs sent into the task

garthwebb:
Previously it was querying for followers again.

https://wikia-inc.atlassian.net/browse/SOC-1867

### DIFF
--- a/extensions/wikia/Follow/FollowEmailTask.class.php
+++ b/extensions/wikia/Follow/FollowEmailTask.class.php
@@ -4,25 +4,35 @@ use Wikia\Tasks\Tasks\BaseTask;
 
 class FollowEmailTask extends BaseTask {
 
-	public function emailFollowNotifications( $initiatingUser, $aWatchers, $iUserId, $iNamespace, $sMessage, $sAction ) {
+	/**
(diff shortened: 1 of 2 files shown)


extensions/wikia/Follow/FollowEmailTask.class.php
```diff
@@ -4,25 +4,35 @@
 
 class FollowEmailTask extends BaseTask {
 
-	public function emailFollowNotifications( $initiatingUser, $aWatchers, $iUserId, $iNamespace, $sMessage, $sAction ) {
```
garthwebb, line 7: Removed useless (and in the case of $sValue below, wrong) Hungarian Notation.

includes/EmailNotification.php
```diff
@@ -103,15 +105,19 @@ public function notifyOnPageChange() {
 		}
 
 		// Build a list of users to notify
-		$watchers = [];
```
garthwebb, line 106: This default now happens via the method signature.

owend:
Cool, LGTM

jsutterfield:
:+1:

PR #9449, opened 2016-01-14
CE-3083 Skip update when there are no records to update

kamilkoterba:
https://wikia-inc.atlassian.net/browse/CE-3083

@Wikia/community-engineering


lukaszkonieczny:
LGTM

PR #9450, opened 2016-01-14
CE-3198: Add insight usage (actionable or informative) to display dif…

lukaszkonieczny:
Add insight usage (actionable or informative) to display different data (insights count)

### DIFF
--- a/extensions/wikia/InsightsV2/InsightsHelper.php
+++ b/extensions/wikia/InsightsV2/InsightsHelper.php
@@ -195,14 +195,19 @@ class InsightsHelper {
 		$highlightedInsighs = self::getHighlightedInsights();
 
 		foreach ( $insightsPages as $key => $class ) {
-			$insightsList[$key] = [
-				'subtitle' => self::INSIGHT_SUBTITLE_MSG_PREFIX . $key,
-				'description' => self::INSIGHT_DESCRIPTION_MSG_PREFIX . $key,
-				'count' => $class::INSIGHT_USAGE == InsightsModel::INSIGHTS_USAGE_ACTIONABLE
+			if ( class_exists( $class ) ) {
+				$insightModel =  new $class();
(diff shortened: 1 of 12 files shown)


extensions/wikia/InsightsV2/models/InsightsFlagsModel.php
```diff
@@ -6,8 +6,8 @@
  */
 
 class InsightsFlagsModel extends InsightsModel {
-
 	const INSIGHT_TYPE = 'flags';
+	const INSIGHT_USAGE = InsightsModel::INSIGHTS_USAGE_INFORMATIVE;
```
kamilkoterba: What about moving INSIGHT_USAGE value to config?

extensions/wikia/InsightsV2/InsightsHelper.php
```diff
@@ -198,7 +198,9 @@ public function prepareInsightsList( $limit = 0 ) {
 			$insightsList[$key] = [
 				'subtitle' => self::INSIGHT_SUBTITLE_MSG_PREFIX . $key,
 				'description' => self::INSIGHT_DESCRIPTION_MSG_PREFIX . $key,
-				'count' => $this->prepareCountDisplay( $insightsCountService->getCount( $key ) ),
+				'count' => $class::INSIGHT_USAGE == InsightsModel::INSIGHTS_USAGE_ACTIONABLE
```
kamilkoterba: And as for moving INSIGHT_USAGE to config, we could retrieve it's state from config here.

lukaszkonieczny: Actually I don't move it to config to avoid creating every insight model here. What do you think?

kamilkoterba: Discussed offline. In general I'm for code cleanness. Performance gain is most likely neglectable here.

kamilkoterba:
LGTM

azarek:
LGTM

PR #9456, opened 2016-01-15
EmailIntegrationTest: make it pass in Poznań

macbre:
Poznań devboxes do not have a dump of `wikianewsletter` database (and WikiFactory variables for this wiki) that is used by `EmailIntegrationTest` - mock `$wgUploadPath` WikiFactory variable to generate a proper URL to production Vignette.

And disable MW HTTP proxy as we want to connect to production rather than Poznań DC border.

```
1) EmailIntegrationTest::testEmailImages
Wikia.gif should return HTTP 200 -- http://vignette3.wikia.nocookie.net//images/8/89/Wikia.gif/revision/latest?cb=20150330185243
Failed asserting that false is true.
```

Pinging @jsutterfield


jsutterfield:
:+1:

PR #9457, opened 2016-01-15
CE-2533 Enable easy review escalation

adamkarminski:

This PR's goal was to create an easy way for JS reviewers to escalate a JS change. The solution is a third button - Escalate! - that sets an `escalated` flag for the given revision and sends a message to the `#js-review-tool` Slack channel.

As part of the work, all code related to a Diff Page view has been moved to a separate class.

Related config PR: https://github.com/Wikia/config/pull/1554


extensions/wikia/ContentReview/ContentReview.i18n.php
```diff
@@ -36,6 +36,7 @@
 	'content-review-status-rejected' => 'Rejected',
 	'content-review-status-live' => 'Live',
 	'content-review-status-autoapproved' => 'Auto-approved',
+	'content-review-status-escalated' => 'Escalated',
```
Grunny, line 39: Needs a QQQ doc message.

extensions/wikia/ContentReview/services/ContentReviewService.class.php
```diff
@@ -83,4 +84,53 @@ public function deletePageData( $wikiId, $pageId ) {
...
+		} else {
+			$revisionId = $diff;
+		}
+
+		$reviewModel = new ReviewModel();
+		$sqlStatus = $reviewModel->escalateReview( $wikiId, $pageId, $revisionId );
+
+		if ( !$sqlStatus ) {
+			return false;
+		} else {
```
Grunny: Since there is a return in the `if`, this `else` block is unnecessary and the code can be unindented.

extensions/wikia/ContentReview/scripts/contentReviewDiffPage.js
```diff
@@ -20,25 +20,47 @@ define(
...
 					oldid: oldid,
 					editToken: mw.user.tokens.get('editToken')
 				};
 
+			sendReviewModifyingRequest('removeCompletedAndUpdateLogs', data);
+		}
+
+		function escalateReview() {
+			var	$button = $(this),
+				$parent = $($button.parent()),
```
Grunny: This can just be `$parent = $button.parent(),`

extensions/wikia/ContentReview/scripts/contentReviewDiffPage.js
```diff
@@ -20,25 +20,47 @@ define(
...
-			$('.content-review-diff-button').on('click', removeCompletedAndUpdateLogs);
+			var $toolbar = $('.content-review-toolbar');
+			$toolbar.on('click', '.content-review-status-button', removeCompletedAndUpdateLogs);
+			$toolbar.on('click', '#content-review-escalate-button', escalateReview);
 		}
 
 		function removeCompletedAndUpdateLogs() {
 			var	$button = $(this),
-				notification,
+				$parent = $($button.parent()),
```
Grunny: This can just be `$parent = $button.parent(),`

extensions/wikia/ContentReview/controllers/ContentReviewDiffPage.class.php
```diff
@@ -0,0 +1,178 @@
...
+		$this->revisionId = $this->getCurrentlyReviewedRevisionId();
+	}
+
+	/**
+	 * Tries to initiate and render the toolbar. Triggers checks if for conditions
+	 * for displaying the toolbar being fulfilled. If not - it returns false.
+	 * Sets toolbarHtml, prepares assets and returns true otherwise.
+	 * @return bool
+	 */
+	public function initReviewerToolbar() {
```
Grunny: I think it might be better to move the rendering of the HTML and adding of assets to the output to the `addToolbarToOutput` method (and make `$toolbarHtml` a local var to that method as it isn't used anywhere else) and then change the hook to just use:

``` php
if ( $diffPage->shouldDisplayToolbar() ) {
    $diffPage->addToolbarToOutput( $output );
}
```

As I'm not sure of the reasoning for initiating that data that is only relevant to modifying the output in the check for whether or not to add the output.

extensions/wikia/ContentReview/controllers/ContentReviewDiffPage.class.php
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	/**
+	 * Returns an HTML with a toolbar displayed to reviewers.
+	 * @return string
+	 * @throws \Exception
+	 */
+	private function renderToolbar() {
+		$params = [
+			'toolbarTitle' => wfMessage( 'content-review-diff-toolbar-title' )->plain(),
```
Grunny: Since this is a `ContextSource`, you can use `$this->msg( ... )->plain()`.

adamkarminski:
Vader approves

extensions/wikia/ContentReview/ContentReviewHelper.php
```diff
@@ -239,113 +237,6 @@ public function hasPageApprovedId( CurrentRevisionModel $model, $wikiId, $pageId
...
-		);
-	}
-
-	/**
-	 * Link for adding new section on script talk page. Prefilled with standard explanation of rejection.
-	 * @param \Title $title Title object of JS page
-	 * @param int $revisionId
-	 * @return string full link to edit page
-	 */
-	public function prepareProvideFeedbackLink( \Title $title, $revisionId = 0 ) {
```
Grunny, line 317: This is still called in ContentReviewApiController on the helper object, so that would cause a fatal.

adamkarminski:
The failing test is the EmailIntegration one, not related.

PR #9462, opened 2016-01-18
SUS-12 Make the statically called put() method actually static

adamkarminski:
The method is called only statically in 3 places in our codebase. I assume that somebody just forgot about the `static` word. :)


macbre:
LGTM

PR #9464, opened 2016-01-18
SUS-15: use the actual block timestamp if possible

ghost:
- https://wikia-inc.atlassian.net/browse/SUS-15

MediaWiki UI uses incorrect block timestamp value - corresponding to the current user's local time - instead of the actual timestamp when the block has been created.

This change fixes the UI. The Phalanx service has to be modified (`match` resource) so that it returns the value of field `p_timestamp` as `timestamp`.

Corresponding Phalanx change: https://github.com/Wikia/scala-backend/pull/8

@Wikia/sustaining-team


macbre:
:+1:

PR #9467, opened 2016-01-18
SUS-8 Prevent SQL Injection and CSRF in the ManageHomePage controller

adamkarminski:
This PR covers fixes for:
1. **SQL Injection** in the `WikiaHomePageHelper::addFlag()` and `WikiaHomePageHelper::removeFlag()` methods.
2. **CSRF** in the `WikiaHomePageController::changeFlag()` method.

Pinging @macbre and the @Wikia/sustaining-team

### DIFF
--- a/extensions/wikia/CityVisualization/models/CityVisualization.class.php
+++ b/extensions/wikia/CityVisualization/models/CityVisualization.class.php
@@ -344,9 +344,7 @@ class CityVisualization extends WikiaModel {
 		$mdb = wfGetDB( DB_MASTER, [], $this->wg->ExternalSharedDB );
 
 		$result = $mdb->update( self::CITY_VISUALIZATION_TABLE_NAME,
-			[
-				'city_flags' => "(city_flags | {$flag})",
-			],
+			[ "city_flags = {$mdb->bitOr( 'city_flags', $flag )}" ],
 			[
 				'city_id' => $wikiId,
 				'city_lang_code' => $langCode,
--- a/extensions/wikia/CityVisualization/models/CityVisualization.class.php
+++ b/extensions/wikia/CityVisualization/models/CityVisualization.class.php
@@ -338,24 +338,13 @@ class CityVisualization extends WikiaModel {
 		$mdb->commit();
 	}
 
-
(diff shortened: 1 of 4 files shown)


extensions/wikia/SpecialManageWikiaHome/ManageWikiaHomeController.class.php
```diff
@@ -452,6 +452,10 @@ public function isWikiInCollection() {
 	public function changeFlag() {
 		wfProfileIn(__METHOD__);
 
+		if ( !$this->request->isValidWriteRequest( $this->getContext()->getUser() ) ) {
```
Grunny: `WikiaRequest::isValidWriteRequest` throws an exception, it doesn't return a boolean. Also, since you're in a controller, you can just call `$this->checkWriteRequest()` which will handle the user for you.

adamkarminski: Good catch, thanks!

extensions/wikia/CityVisualization/models/CityVisualization.class.php
```diff
@@ -369,16 +373,18 @@ public function getFlag($wikiId, $langCode) {
...
+		$mdb = wfGetDB( DB_MASTER, [], $this->wg->ExternalSharedDB );
+		$result = $mdb->update( self::CITY_VISUALIZATION_TABLE_NAME,
+			[ "city_flags = {$mdb->bitAnd( 'city_flags', $mdb->bitNot( $flag ) )}" ],
+			[
+				'city_id' => $wikiId,
+				'city_lang_code' => $langCode,
+			]
+		);
+		$mdb->commit( __METHOD__ );
+		wfProfileOut( __METHOD__ );
```
macbre, line 421: As we're here. Let's create a helper method (`doUpdateFlag` maybe?) that will be called from both `setFlag` and `removeFlag` - one c&p less.

macbre:
LGTM - one minor left

adamkarminski:
@macbre Done. :)

macbre:
Magnificent break, sir! ;)

PR #9469, opened 2016-01-18
RECIRC-2: AB test top fandom content in right rail

pauloslund:
The Recirculation extensions relies on a AB test being run.

AB Test Name: RECIRCULATION_RAIL
Variants: VIDEOS_MODULE, TABOOLA, POPULAR

### DIFF
--- a/extensions/wikia/Recirculation/RecirculationController.class.php
+++ b/extensions/wikia/Recirculation/RecirculationController.class.php
@@ -17,10 +17,13 @@ class RecirculationController extends WikiaController {
 	public function index() {
 		$fandomDataService = FandomDataService::getInstance();
 
-		$posts = $fandomDataService->getData( $this->type );
+		$posts = $fandomDataService->getPosts( $this->type );
 
 		if (count($posts) > 0) {
-			$this->response->setVal( 'posts', $posts );
+			$this->response->setData( [
+				'title'	=> wfMessage( 'recirculation-fandom-title' )->plain(),
+				'posts' => $posts
+			] );
 			return true;
 		} else {
 			return false;
--- a/extensions/wikia/Recirculation/RecirculationController.class.php
+++ b/extensions/wikia/Recirculation/RecirculationController.class.php
@@ -7,7 +7,7 @@ class RecirculationController extends WikiaController {
 
 	public function init() {
 		$type = $this->getVal( 'type' );
-		if (in_array($type, self::ALLOWED_TYPES)) {
+		if ( in_array( $type, self::ALLOWED_TYPES ) ) {
 			$this->type = $type;
 		} else {
 			$this->type = self::DEFAULT_TYPE;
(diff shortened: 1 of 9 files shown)


extensions/wikia/Recirculation/js/recirculation.js
```diff
@@ -17,10 +18,31 @@ define('ext.wikia.recirculation.recirculation', [
 		});
 	}
 
+	function fandomPosts ( type, element ) {
```
gabrys: Shouldn't this be named `injectFandomPosts`?

extensions/wikia/Recirculation/RecirculationController.class.php
```diff
@@ -0,0 +1,33 @@
...
+		$type = $this->getVal( 'type' );
+		if (in_array($type, self::ALLOWED_TYPES)) {
+			$this->type = $type;
+		} else {
+			$this->type = self::DEFAULT_TYPE;
+		}
+	}
+
+	public function index() {
+		$fandomDataService = FandomDataService::getInstance();
```
gabrys: Any reason not to actually use the old-school 'new FandomDataService()`?

extensions/wikia/Recirculation/services/FandomDataService.class.php
```diff
@@ -0,0 +1,89 @@
...
+
+	/**
+	 * Make an API request to parsely to gather posts
+	 * @param string $type
+	 * @return an array of posts
+	 */
+	private function apiRequest( $type ) {
+		switch ($type) {
+			case 'shares':
+				$endpoint = 'shares/posts';
```
gabrys, line 40: Those could be class consts, but I don't mind them like that (if only used once).

extensions/wikia/Recirculation/templates/Recirculation_index.mustache
```diff
@@ -0,0 +1,15 @@
+<section class="recirculation-rail module">
+	<h2>{{ title }}</h2>
+	<ul class="thumbnails">
+		{{#posts}}
```
gabrys, line 4: Wow, I didn't know you could do this without writing `for`

extensions/wikia/Recirculation/RecirculationHooks.class.php
```diff
@@ -31,6 +31,11 @@ static public function onGetRailModuleList( &$modules ) {
 		return true;
 	}
 
+	public static function onBeforePageDisplay( OutputPage $out, Skin $skin ) {
+		Wikia::addAssetsToOutput( 'recirculation_scss' );
```
gabrys, line 35: We could load the CSS along with the AJAX query fetching the fandom posts rather than loading this eagerly.

extensions/wikia/Recirculation/RecirculationController.class.php
pauloslund: Looked at a few other extensions and saw something similar to `getInstance`. May be over engineered for this use case though, so I have no issue with `new FandomDataService();`

gabrys:
LGTM

PR #9470, opened 2016-01-18
Aden 2688 revive

jacekjursza:
Those changes are enabling us to use never version of OpenX server - Revive. In order to use it, you need to set wiki variable wgEnableReviveSpotlights to true; In the future we are going to set this variable based on geolocation - automatically.


fraszczakszymon:
LGTM

PR #9471, opened 2016-01-18
ADEN-2845 Make Liftium more secure

nandy-andy:
Liftium is on life support and if the change in this PR makes it more secure and in parallel still provides an ad - we go with it in production environment!

It fixes an issue with XSS through `postMessage`.


fraszczakszymon:
LGTM :+1:

PR #9472, opened 2016-01-18
SEO-221 Broken XML comment in sitemaps

gabrys:
For wikis with two consecutive dashes in the domain name, the XML comment
that we output is invalid as it contains "--" within a comment, which is
invalid.

There's no much use of the comment anyway, so I'm removing it.


extensions/wikia/Sitemap/SpecialSitemap_body.php
```diff
@@ -195,7 +195,6 @@ private function generateIndex() {
 
 		$out = "";
 		$out .= "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n";
-		$out .= sprintf( "<!-- generated on fly by %s -->\n", $this->mTitle->getFullURL() );
```
saipetch, line 198: What about line 259?

saipetch:
LGTM

PR #9477, opened 2016-01-19
Fix CSRF in UserSignupSpecialController

macbre:
[SUS-20](https://wikia-inc.atlassian.net/browse/SUS-20)

Check edit token when sending confirmation email & changing user's unconfirmed email.

To increase the security, let's generate the "full" edit token when the PHP session is started (even when  the user is not logged in).

@Wikia/sustaining-team

### DIFF
--- a/includes/User.php
+++ b/includes/User.php
@@ -3949,7 +3949,7 @@ class User {
 			$request = $this->getRequest();
 		}
 
-		if ( $this->isAnon() ) {
+		if ( $this->isAnon() && session_status() !== PHP_SESSION_ACTIVE /* Wikia change (SUS-20) */ ) {
 			return EDIT_TOKEN_SUFFIX;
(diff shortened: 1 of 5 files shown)


includes/User.php
```diff
@@ -3949,7 +3949,7 @@ public function getEditToken( $salt = '', $request = null ) {
 			$request = $this->getRequest();
 		}
 
-		if ( $this->isAnon() ) {
+		if ( $this->isAnon() && session_status() !== PHP_SESSION_ACTIVE /* Wikia change (SUS-20) */ ) {
```
macbre, line 3952: @Grunny / @wladekb 

MW returns `+\` edit token for anons. Let's generate a token that is less likely to be guessed when an anon has the PHP session (i.e. it can be used to store the token).

PR #9483, opened 2016-01-19
SUS-19: check token when removing an avatar

macbre:
[SUS-19 https://wikia-inc.atlassian.net/browse/SUS-19]

Check the HTTP method and edit token when trying to remove an avatar

@Wikia/sustaining-team

### DIFF
--- a/extensions/wikia/UserProfilePageV3/UserProfilePageController.class.php
+++ b/extensions/wikia/UserProfilePageV3/UserProfilePageController.class.php
@@ -956,6 +956,10 @@ class UserProfilePageController extends WikiaController {
 		wfProfileOut( __METHOD__ );
 	}
 
+	/**
+	 * @throws BadRequestException
+	 * @return bool
+	 */
 	public function removeavatar() {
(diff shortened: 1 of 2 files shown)


extensions/wikia/UserProfilePageV3/UserProfilePageController.class.php
```diff
@@ -958,6 +958,8 @@ public function renderClosingModal() {
 
 	public function removeavatar() {
```
RafLeszczynski, line 963: just a small thing but you could add the doc with info that this function throw BadRequest Exception

RafLeszczynski:
one small remark, except this LGTM

extensions/wikia/UserProfilePageV3/UserProfilePageController.class.php
macbre, line 963: True, let's be consistent :)

0962fce

adamkarminski:
Hetfield approves

PR #9484, opened 2016-01-19
ADEN-2859 Add noProxy flag to ResourceLoader::fetchRemoveScript and bump all CB

fraszczakszymon:
Quick fix for ADEN P2 in DE

### DIFF
--- a/extensions/wikia/AdEngine/ResourceLoaders/ResourceLoaderAdEngineBase.php
+++ b/extensions/wikia/AdEngine/ResourceLoaders/ResourceLoaderAdEngineBase.php
@@ -23,7 +23,7 @@ abstract class ResourceLoaderAdEngineBase extends ResourceLoaderModule {
 	 * @return bool|MWHttpRequest|string
 	 */
 	protected function fetchRemoteScript( $url ) {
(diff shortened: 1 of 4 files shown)


extensions/wikia/AdEngine/ResourceLoaders/ResourceLoaderAdEngineBase.php
```diff
@@ -23,7 +23,7 @@
 	 * @return bool|MWHttpRequest|string
 	 */
 	protected function fetchRemoteScript( $url ) {
-		return Http::get( $url );
+		return Http::get( $url, 'default', [ 'noProxy' => true ] );
```
nandy-andy, line 26: ping @macbre

nandy-andy:
LGTM :+1:

extensions/wikia/AdEngine/ResourceLoaders/ResourceLoaderAdEngineBase.php
macbre, line 26: LGTM.

As we discussed earlier, let's debug the root cause tomorrow morning :)

PR #9486, opened 2016-01-19
RECIRC-27: Update Parsely API code so 'shares' API works

pauloslund:
There is currently a bug with Parsely's API that require the optional parameter `days` to be greater than 3.


gabrys:
LGTM

PR #9488, opened 2016-01-20
ADEN-2812 IVW3 integration

fraszczakszymon:
Implementation of new IVW3 tracking provider which will be enabled in DE, AT, CH

### DIFF
--- a/extensions/wikia/AnalyticsEngine/AnalyticsProviderIVW2.php
+++ b/extensions/wikia/AnalyticsEngine/AnalyticsProviderIVW2.php
@@ -38,8 +38,7 @@ class AnalyticsProviderIVW2 implements iAnalyticsProvider {
 					"sv" => "ke"      // FRABO-Tag deaktiviert
 				];
 
-				$ivwScriptTag = '<script src="https://script.ioam.de/iam.js"></script>';
-				$ivwScriptTag .= '<script>iom.c(' . json_encode($iamData) . ', 2);</script>';
+				$ivwScriptTag = '<script>iom.c(' . json_encode($iamData) . ', 2);</script>';
 				$ivwScriptTagEscaped = json_encode($ivwScriptTag);
 
 				$script = <<<SCRIPT
--- a/extensions/wikia/AnalyticsEngine/AnalyticsProviderIVW2.php
+++ b/extensions/wikia/AnalyticsEngine/AnalyticsProviderIVW2.php
@@ -38,6 +38,7 @@ class AnalyticsProviderIVW2 implements iAnalyticsProvider {
 					"sv" => "ke"      // FRABO-Tag deaktiviert
 				];
 
+				// iam.js library loaded in AnalyticsProviderIVW3
 				$ivwScriptTag = '<script>iom.c(' . json_encode($iamData) . ', 2);</script>';
 				$ivwScriptTagEscaped = json_encode($ivwScriptTag);
 
--- a/extensions/wikia/AnalyticsEngine/AnalyticsProviderIVW2.php
+++ b/extensions/wikia/AnalyticsEngine/AnalyticsProviderIVW2.php
@@ -38,15 +38,17 @@ class AnalyticsProviderIVW2 implements iAnalyticsProvider {
 					"sv" => "ke"      // FRABO-Tag deaktiviert
 				];
 
-				// iam.js library loaded in AnalyticsProviderIVW3
-				$ivwScriptTag = '<script>iom.c(' . json_encode($iamData) . ', 2);</script>';
-				$ivwScriptTagEscaped = json_encode($ivwScriptTag);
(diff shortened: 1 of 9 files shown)


extensions/wikia/AnalyticsEngine/AnalyticsProviderIVW3.php
```diff
@@ -0,0 +1,65 @@
...
+		$vertical = HubService::getVerticalNameForComscore( $wgCityId );
+		$url = self::$libraryUrl;
+		$iamData = json_encode( [
+				'st' => 'wikia',
+				'cp' => $vertical,
+				'sv' => 'ke',
+		] );
+
+		$ivwScriptTag = <<<CODE
+<script type="text/javascript" src="{$url}"></script>
```
nandy-andy: Can't we just use here `self::$libraryUrl`?

extensions/wikia/AnalyticsEngine/AnalyticsProviderIVW3.php
```diff
@@ -0,0 +1,65 @@
...
+<script type="text/javascript">window.iom.c({$iamData}, 2);</script>
+CODE;
+		$ivwScriptTagEscaped = json_encode($ivwScriptTag);
+
+		return <<<CODE
+<!-- Begin IVW3 Tag -->
+<script type="text/javascript">
+	if (window.Wikia && window.Wikia.geo.isProperGeo(['DE', 'AT', 'CH'])) {
```
nandy-andy: Now I'm thinking about not an instant global but just a regular WF variable which we can use here in PHP and re-use in Mercury. But can we use it in Mercury? Like get it on first view where whole Ember is loading and Mercury asks for WF config? Can we return it with this config if it gets there before sending request to IVW? If yes, let's do it! If no, let's not waste our time and use it to think how make Mercury more flexible ;-)

extensions/wikia/AnalyticsEngine/AnalyticsProviderIVW2.php
```diff
@@ -38,15 +38,17 @@ function trackEvent($event, $eventDetails = array()) {
 					"sv" => "ke"      // FRABO-Tag deaktiviert
 				];
 
-				$ivwScriptTag = '<script src="https://script.ioam.de/iam.js"></script>';
-				$ivwScriptTag .= '<script>iom.c(' . json_encode($iamData) . ', 2);</script>';
-				$ivwScriptTagEscaped = json_encode($ivwScriptTag);
+				$ivwScriptTag = json_encode('<script src="https://script.ioam.de/iam.js"></script>');
+				$ivwTrackTag = json_encode('<script>iom.c(' . json_encode($iamData) . ', 2);</script>');
```
nandy-andy, line 42: Can you contact IVW if it hurts them if we execute it twice? However, those are different data, aren't they?

extensions/wikia/AnalyticsEngine/AnalyticsProviderIVW3.php
fraszczakszymon: Can't use it in HEREDOC :(

hakubo: can we not write JavaScript in php? 
You could put that at least in mustache template and render it, that alone would already be like we lived in 2016 :)

fraszczakszymon: Right! :gem: now it's much cleaner, thanks

extensions/wikia/AnalyticsEngine/AnalyticsProviderIVW3.php
fraszczakszymon: :+1:

extensions/wikia/AnalyticsEngine/AnalyticsProviderIVW2.php
fraszczakszymon, line 42: :+1: I'll send an e-mail

extensions/wikia/AnalyticsEngine/AnalyticsProviderIVW3.php
nandy-andy: @hakubo, oh you fashion boy! :japanese_ogre: :trollface: :dash: 

Thanks for the hint! We got into "being consistent" trap. We'll try to remember it next time we touch trackers :bowtie:

extensions/wikia/AnalyticsEngine/AnalyticsProviderIVW2.php
nandy-andy, line 42: Thank you :bow:

jacekjursza:
LGTM

jacekjursza:
LGTM

PR #9490, opened 2016-01-20
WikiaPageType::isWikiaHubMain - allow a "custom" title to be passed

macbre:
[SUS-11](https://wikia-inc.atlassian.net/browse/SUS-11)

`WikiaHubsV3Hooks::onArticleFromTitle` calls `WikiaPageType::isWikiaHubMain` in the context where `$wgTitle` is not defined (namely `updateSpecialPages` script and tasks).

Allow `WikiaPageType::isWikiaHubMain` to be provided with a "custom" `Title` object to perform a check for (instead of relying on `$wgTitle`). Unit tests extended to cover this case.

@Wikia/sustaining-team


extensions/wikia/WikiaHubsV3/hooks/WikiaHubsV3Hooks.php
```diff
@@ -7,11 +7,11 @@ class WikiaHubsV3Hooks {
 	 *
 	 * @return true because it's a hook
 	 */
-	static public function onArticleFromTitle(&$title, &$article) {
+	static public function onArticleFromTitle(Title $title, &$article) {
```
adamkarminski, line 10: If the `$article` is an object you could remove the reference and hint a class for it. :)

adamkarminski:
Ultra mini minor comment, LGTM otherwise. :)

extensions/wikia/WikiaHubsV3/hooks/WikiaHubsV3Hooks.php
macbre, line 10: Removed in 851fb45

macbre, line 10: The `$article` object is replaced inside the hook - let's keep the reference.

PR #9491, opened 2016-01-20
CE-3242: Show different message if informative insight is empty

lukaszkonieczny:
Show different message if informative insight is empty

### DIFF
--- a/extensions/wikia/InsightsV2/templates/Insights_subpageList.php
+++ b/extensions/wikia/InsightsV2/templates/Insights_subpageList.php
@@ -94,18 +94,7 @@
 				<?php if ( $pagination ) : ?>
 					<?= $pagination ?>
 				<?php endif ?>
-			<?php else: ?>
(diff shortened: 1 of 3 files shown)


extensions/wikia/InsightsV2/templates/Insights_subpageList.php
```diff
@@ -94,14 +94,16 @@
...
 				<?php endif ?>
-			<?php elseif (!empty( $flagTypes ) ) : ?>
-				<p>
-					<?= wfMessage( 'insights-list-no-flag-types' )->escaped(); ?>
-				</p>
 			<?php else: ?>
-				<p>
-					<?= wfMessage( 'insights-list-no-items' )->escaped(); ?>
-				</p>
+				<?php if ( $usage == InsightsModel::INSIGHTS_USAGE_ACTIONABLE ): ?>
```
adamkarminski: I think a cleaner solution would be to perform this logic in the controller and pass a prepared message to the template. Also - use `===` if possible. :)

adamkarminski:
LGTM!

azarek:
LGTM :)

PR #9495, opened 2016-01-20
SUS-16 Force local URLs as referers while making HTTP requests

ghost:
The change applies to [interwiki pages](https://www.mediawiki.org/wiki/Manual:Interwiki) execution flow. It replaces the interwiki URL with the original requested URL in the `Referer` header.

Otherwise an HTTP request sent during the execution of http://muppet.wikia.com/wiki/WikiMedia:Common.css would have https://wikimediafoundation.org/wiki/Common.css in the `Referer` header which would produce confusing results.

The issue has been first discovered on Special:PhalanxStats claiming that some user has been blocked while visiting some wikimediafoundation.org page which does not make sense at all.

This was because the Phalanx service was receiving the URL after interwiki conversion.

@Wikia/sustaining-team 

https://wikia-inc.atlassian.net/browse/SUS-16

### DIFF
--- a/includes/HttpFunctions.php
+++ b/includes/HttpFunctions.php
@@ -472,7 +472,8 @@ class MWHttpRequest {
 		}
 
 		if ( is_object( $wgTitle ) && !isset( $this->reqHeaders['Referer'] ) ) {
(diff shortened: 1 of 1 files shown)


macbre:
:+1:

PR #9499, opened 2016-01-20
ADEN-2864 Mark SP events with nonInteraction flag

fraszczakszymon:
Mark SP events as non-interaction in order to fix bounce rate calculations


nandy-andy:
LGTM :angel:

PR #9504, opened 2016-01-20
CE-3206 i18n for GlobalShortcuts

kamilkoterba:


extensions/wikia/GlobalShortcuts/GlobalShortcutsHelper.php
```diff
@@ -81,9 +81,25 @@ public function addSpecialPageActions( array &$actions, array &$shortcuts ) {
 			}
 		}
 
+		$specialPagesMsg = wfMessage( 'specialpages' )->plain();
```
Grunny: This message isn't being escaped later. Try putting JS into the message and opening the modal. Either the modal should make sure to escape the text of the shortcuts, or they should be escaped before being passed to the modal. Same for the groupName messages below and possibly others.

Grunny:
LGTM.

extensions/wikia/TemplateClassification/TemplateClassification.setup.php
```diff
@@ -101,3 +101,8 @@
 JSMessages::registerPackage( 'TemplateClassificationHints', [
 	'template-classification-bulk-classification-*'
 ] );
+
+JSMessages::registerPackage( 'TemplateClassificationGlobalShortcuts', [
+	'template-class-global-shortcuts-caption-classify-page',
+	'template-class-global-shortcuts-press-to-explore-shortcuts',
```
Grunny: Why is `global-shortcuts-press-to-explore-shortcuts` prefixed with `template-class-`? It doesn't seem specific to TemplateClassification and it's defined in the GlobalShortcuts extension.

extensions/wikia/GlobalShortcuts/scripts/GlobalShortcutsHelp.js
```diff
@@ -19,7 +19,9 @@ define('GlobalShortcutsHelp',
 					Wikia.getMultiTypePackage({
 						mustache: 'extensions/wikia/GlobalShortcuts/templates/KeyCombination.mustache,' +
 							'extensions/wikia/GlobalShortcuts/templates/GlobalShortcutsController_help.mustache',
+						messages: 'GlobalShortcuts,TemplateClassificationGlobalShortcuts',
```
Grunny: If this is just for `template-class-global-shortcuts-press-to-explore-shortcuts`, I think it should be shortened to  `global-shortcuts-press-to-explore-shortcuts`  and defined in the `GlobalShortcuts` messages package rather than having this cross-dependency with the TemplateClassification extension where the `TemplateClassificationGlobalShortcuts` package is defined.

extensions/wikia/TemplateClassification/TemplateClassification.setup.php
kamilkoterba: I want to move it to TemplateClassification as it's used there, but I think it's better to wait for translation process to complete and then move.

PR #9507, opened 2016-01-20
SEO-232 Update isSpecial('Newimages') to isSpecial('Images')

gabrys:
There were two places in the code that checked for whether the current
page is the Newimages special page.

One of them was adding the "Add a photo button" in the page header. The
other one would enable ads on the page.

Both were converted to check for the new Special:Images page.


saipetch:
LGTM

PR #9511, opened 2016-01-20
SEO-233 Fix PHP Fatal in redirect-canonical and add logging

gabrys:
Not coding defensively, this fatal revealed another issue.

Before the changes some URLs generated a MediaWiki page saying "Error
404: Page not found | Special page | These aren't the pages you're
looking for. Move along, please".

Now they return HTTP 404 with an empty response and a PHP Fatal logged.
This commit fixes that, but also logs the original REQUEST_URI for
catching what weird URIs like that are requested.

It seems there were three main problems:
1. URLs like http://some.wikia.com/Something:123 . PHP's parse_url given
   /Something:123 and the param interprets Something as the hostname and
   123 as the port and leaves path (which we're extracting) empty.
2. Modular main pages URLs requested on desktop. For some reason the
   URLs are double-encoded, so after initial decoding, there are still "%"
   signs in them, which are forbidden in MW Titles.
3. Plain broken URLs that don't map to any valid MW title.
   Title::newFromText doesn't return an object and code fails.

This pull request adds logging for all three, replaces the forbidden
characters with "_" and as a last resort redirects to the home page.

### DIFF
--- a/redirect-canonical.php
+++ b/redirect-canonical.php
@@ -13,9 +13,22 @@
  * ErrorDocument 404 /redirect-canonical.php
  */
 
+use \Wikia\Logger\WikiaLogger;
+
 require_once( dirname( __FILE__ ) . '/includes/WebStart.php' );
 
-$path = ltrim( parse_url( $_SERVER['REQUEST_URI'], PHP_URL_PATH ), '/' );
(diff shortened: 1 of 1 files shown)


saipetch:
LGTM

PR #9513, opened 2016-01-21
CE-3204: Create AMD module for devbridge autocomplete plugin

lukaszkonieczny:
Create AMD module for devbridge autocomplete plugin and use it in global shortcuts suggestions


Grunny:
:+1:

azarek:
LGTM

PR #9516, opened 2016-01-21
Update the class generated with the new special page name

adamkarminski:
The CSS class of the "Upload new image" button is generated based on the special page's name. The fix is just updating the class selector.


nandy-andy:
It's sad :crying_cat_face: But LGTM as a quick fix. 

How about creating a ticket in which we'll add one more class to which we'll attach the event?

adamkarminski:
@nandy-andy Of course. :)

PR #9520, opened 2016-01-21
Dat 3639b use curated content data service

idradm:


includes/wikia/services/CommunityDataService.class.php
```diff
@@ -0,0 +1,112 @@
...
+	public function setCuratedContent( $data ) {
+		$ready = $this->isOldFormat( $data ) ? $this->toNew( $data ) : $data;
+		$status = WikiFactory::setVarByName( self::CURATED_CONTENT_VAR_NAME, $this->cityId, $ready );
+		if ( $status ) {
+			wfWaitForSlaves();
+		}
+		return $status;
+	}
+
+	public function getCuratedContent( $useLegacyFormat = false ) {
```
sebastianmarzjan: Can we make it an enum with two possible values, i.e. 
FORMAT_STRUCTURED and FORMAT_LEGACY ?

includes/wikia/services/CommunityDataService.class.php
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	/**
+	 * Needs to be removed after migration of wgWikiaCuratedContent
+	 * to new format
+	 *
+	 * @param $curatedContent array
+	 * @return bool true if wgWikiaCuratedContent has old format
+	 */
+	private function isOldFormat( $curatedContent ) {
```
sebastianmarzjan: Let's use 'Legacy' instead of 'Old' consistently

includes/wikia/services/CommunityDataService.class.php
```diff
@@ -0,0 +1,156 @@
...
+
+	/**
+	 * Returns both curated and optional sections merged together
+	 * @return array
+	 */
+	public function getNonFeaturedSections() {
+		$curated = $this->getCurated();
+		$optional = $this->getOptional();
+		if ( !empty( $optional ) ) {
+			$curated[] = $optional;
```
sebastianmarzjan, line 36: Are you sure about this type of assignments? It looks like getCurated returns an array of items, getOptional a single item, and it may be confusing especially because of 'merge' used in description. If there's only one item in optional, I'd suggest making it explicit; alternatively, unify the format so both are an array.

extensions/wikia/GameGuides/GameGuidesController.class.php
```diff
@@ -564,39 +561,41 @@ private function getTagCategories( $content, $requestTag ) {
...
-								return $ret;
-							} ),
-							null,
-							false,
-							//I need all of them basically
-							count( $ret )
-						)
-					);
+				} else {
+					if ( $sort == 'hot' ) {
```
sebastianmarzjan: Why the logic change here and in a few else if => else { if (() } other cases?

sebastianmarzjan:
LGTM

PR #9522, opened 2016-01-21
PLATFORM-1837: Include request id in DB query comment

wladekb:
This change will include request id in the DB query comment for easier log correlation.

https://wikia-inc.atlassian.net/browse/PLATFORM-1837

/cc @macbre


macbre:
:+1:

PR #9525, opened 2016-01-21
PLATFORM-1355: Log to Kibana information about serving an empty page

inez:
I'm checking if the length of wikitext is greater than 50 to avoid noise of logging pages which contains just a category (or few categories) and their content is meant to be empty. This simple check will not eliminiate all the "noise" but probably most of it.

I also removed the previously added logging which wasn't helpful in debugging this issue - but was sending tons of data to Kibana and making bunch of pages "not cachable"


dianafa:
now LGTM

PR #9530, opened 2016-01-21
SOC-1820 Delete attributes from service with value set to null.

jsutterfield:
This PR makes sure that if an attribute is set with a value of  `NULL`, it's deleted from the attribute service. That's the way `User.php` used to handle attributes. When we ported over that logic, we missed that case. See [here](https://github.com/Wikia/app/blob/dev/includes/User.php#L5032) (`isDefaultOption` returns true for null attributes without a default value).

This fixes the bug of a user not being able to clear out their birthday on their profile page. 

Ticket: https://wikia-inc.atlassian.net/browse/SOC-1820

### DIFF
--- a/lib/Wikia/src/Service/User/Attributes/UserAttributes.php
+++ b/lib/Wikia/src/Service/User/Attributes/UserAttributes.php
@@ -141,7 +141,10 @@ class UserAttributes {
 	}
 
 	private function attributeShouldBeDeleted( $name, $value ) {
-		return isset( $this->defaultAttributes[$name] ) && $value == $this->defaultAttributes[$name];
+		return (
+			( isset( $this->defaultAttributes[$name] ) && $value == $this->defaultAttributes[$name] ) ||
+			is_null( $value )
+		);
 	}
 
 	/**
--- a/lib/Wikia/tests/Service/User/UserAttributesTest.php
+++ b/lib/Wikia/tests/Service/User/UserAttributesTest.php
@@ -137,6 +137,18 @@ class UserAttributeTest extends PHPUnit_Framework_TestCase {
 		$this->assertNull( $userAttributes->getAttribute( $this->userId, $this->attribute1->getName() ) );
 	}
 
+	public function testDeleteAttrbutesWithNullValue() {
+		$attrNullValue = new Attribute( "someKey", null );
+		$this->setupServiceGetExpects();
+		$this->service->expects( $this->once() )
+			->method( 'delete' )
+			->with( $this->userId, $attrNullValue );
+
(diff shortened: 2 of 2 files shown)


lib/Wikia/src/Service/User/Attributes/UserAttributes.php
```diff
@@ -141,7 +141,10 @@ private function attributeShouldBeSaved( $name, $value ) {
 	}
 
 	private function attributeShouldBeDeleted( $name, $value ) {
-		return isset( $this->defaultAttributes[$name] ) && $value == $this->defaultAttributes[$name];
+		return (
+			( isset( $this->defaultAttributes[$name] ) && $value == $this->defaultAttributes[$name] ) ||
+			is_null( $value )
```
armonr, line 146: Not too familiar with attributes logic, but could this potentially conflict with the logic here https://github.com/Wikia/app/blob/SOC-1820/lib/Wikia/src/Service/User/Attributes/UserAttributes.php#L103-L105 if the default attribute for the attribute in question happens to be non-null? The `$value` parameter seems to be populated with the value of an attribute object loaded by `loadAttributes()`.

garthwebb, line 146: Good point @armonr definitely looks like something to investigate

jsutterfield, line 146: I don't think that logic should conflict. When it comes to default attributes, we only store them in the service if the user has a non-default value for them. Otherwise, we just pull them from the `$defaultAttributes` array which is loaded from a config. See [here](https://github.com/Wikia/app/blob/SOC-1820/lib/Wikia/src/Service/User/Attributes/UserAttributes.php#L58-L60). By setting an attribute with a default to null, we're effectively saying "delete whatever value the user had for this attribute in the service and use the default from here on out."

For example, if we set `fancySig` to null via  `UserAttribute::setAttribute`, then what gets stored in the attribute array for that user is the default value for `fancySig`, [which is 0](https://github.com/Wikia/app/blob/dev/includes/DefaultSettings.php#L3218). Then for subsequent calls to `$userAttributes->getAttribute( 'fancySig' )`, 0 would be returned.

When we came to actually saving the attributes, we'd call `attributeShouldBeDeleted()` for `fancySig` and see that the user has the default value set. So that means we should delete it from the service. That way we can update the default to something else and not have to worry about updating that value for all users. It's just a config change.

I may not be 100% understanding your question though. Does that explanation make sense?

armonr, line 146: Yes, make sense now. Thanks for the explanation!

armonr:
LGTM

PR #9534, opened 2016-01-22
XW-840 Use simple_html_dom instead of regexp to strip inline attributes

rogatty:
https://wikia-inc.atlassian.net/browse/XW-840

/cc @WikiaSparrow @macbre @hakubo


macbre:
LGTM

Feel free to grab me when this fix goes live. We can take a look at xhprof data for `onParserAfterTidy` method.

hakubo:
lgtm

PR #9538, opened 2016-01-22
PLATFORM-1355: This is a temporary code to track the cause of the problem

inez:
It will be removed soon and it might be modified often in the meantime - for instance to add logging of parser options used while parsing.


garthwebb:
LGTM!

PR #9539, opened 2016-01-22
PLATFORM-1355: Added logging in one more place

inez:
Apparently there are plent of handlers for "ParserAfterTidy" hook


garthwebb:
LGTM

PR #9545, opened 2016-01-25
ADEN-2873 Remove lifestyle restriction

drets:
Remove lifestyle restrictions from the exitsting OX Bidder

### DIFF
--- a/extensions/wikia/AdEngine/js/lookup/openXBidder.js
+++ b/extensions/wikia/AdEngine/js/lookup/openXBidder.js
@@ -142,10 +142,6 @@ define('ext.wikia.adEngine.lookup.openXBidder', [
 		node.parentNode.insertBefore(openx, node);
 	}
 
-	function isEnabled() {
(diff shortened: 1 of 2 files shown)


extensions/wikia/AdEngine/js/lookup/openXBidder.js
```diff
@@ -157,7 +153,6 @@ define('ext.wikia.adEngine.lookup.openXBidder', [
 	return factory.create({
 		logGroup: logGroup,
 		name: 'ox_bidder',
-		isEnabled: isEnabled,
```
nandy-andy, line 160: It seems none of our bidders implements this method any longer. What do you think about [removing it from the factory as well](https://github.com/Wikia/app/blob/dev/extensions/wikia/AdEngine/js/lookup/lookupFactory.js#L23)?

drets, line 160: :+1:

nandy-andy:
LGTM :+1:

PR #9548, opened 2016-01-25
MAIN-6290: remove block/ban token check for now

owend:


macbre:
Pinging @adamkarminski

PR #9549, opened 2016-01-26
SEO-210 More forbidden characters and fixed urlLen

gabrys:
Square and curly brackets are not what MediaWiki likes, let's filter them out before redirecting.


saipetch:
LGTM

PR #9550, opened 2016-01-26
PLATFORM-1355: Create a special kind of logger that will help me track when the output of article becomes empty

inez:


owend:
LGTM

PR #9556, opened 2016-01-26
MAIN-6280: Add HTTP caching to Recirculation controller index

pauloslund:


extensions/wikia/Recirculation/RecirculationController.class.php
```diff
@@ -20,6 +21,7 @@ public function index() {
 		$posts = $fandomDataService->getPosts( $this->type );
 
 		if ( count( $posts ) > 0 ) {
+			$this->response->setCacheValidity( self::CACHE_DURATION );
```
hakubo: I believe we did have some standards for that
@pchojnacki might now more

gabrys: There are some standards, but mostly for what is the HTTP cache we serve, not the Memcached cache times.

gabrys:
LGTM

extensions/wikia/Recirculation/RecirculationController.class.php
hakubo: and I think `setCacheValidity` is actually dealing with Varnish/Browser cache, right?

pauloslund: Yes `setCacheValidity` is dealing with Varnish/Browser cache. There is `WikiaResponse::CACHE_STANDARD`, but we decided on the 15 minute cache time to make sure we're grabbing fresh information from Parsely. Would it make sense to keep the Varnish cache for 24 hours even though we're updating the cache from Parsely every 15 minutes?

hakubo: Well there are 4 options right now :)

https://github.com/Wikia/app/blob/MAIN-6280/includes/wikia/nirvana/WikiaResponse.class.php#L47

I'll leave it to you - no hard opinions here.

Just did a quick math and
If you estimate 100 hits per minute that means 99,8% of request will be served by Varnish and data will be stale for max 20min
if you set it to 15 minutes you get 99,9% of request hitting varnish and max stale data of 30min

Good night :)

gabrys: Oh yeah, you're right about it, it is the HTTP cache we're exposing to Varnish and client. We should probably use the consts then.

pawelchcki: @hakubo I think @macbre did the work on Cache standarization actually :)
Anyway I think WikiaResponse::CACHE_VERY_SHORT (5 min) might be used instead of 15 minute cache time 

This is the ticket about it: https://wikia-inc.atlassian.net/browse/PLATFORM-57
I don't remember rationale clearly now, but I do remember that we used have  bazilion different cache expiration  times and it was quite  hard to divine  what expiries when, especially coupled with memcache expiration periods. 

So I think if its possible you could use the standard CACHE_VERY_SHORT.

gabrys: Yeap, what this gentleman said :-). (Or maybe add a 30 minutes tier 
between 5 minutes and 3 hours?)

pawelchcki: :+1: adding another standard time in WikiaResponse  seems like a swell idea

hakubo: for 100 req/min:
15 min cache - 99,99333333% requests hits varnish
30 min cache - 99,99666667% requests hits varnish

Does this change is really worth adding another standard time?

pauloslund: 5 minutes does seem too short and 3 hours too long, however what could you possibly name a standard in between `SHORT` and `VERY_SHORT`? I went with just using `WikiaResponse::CACHE_VERY_SHORT` for now. Although, I'm not totally opposed to the idea of using `WikiaResponse::CACHE_SHORT` if we think 5 minutes isn't enough time

gabrys: My vote goes to SHORTER :-)

macbre: Even the shortest caching time is better than no caching at all :) Thanks!

hakubo: True!
Sorry for starting this long conversation :) My sincere apologies :)

PR #9557, opened 2016-01-26
SEO-246 Don't publish any more language links to lilly

gabrys:
Processing the queue is disabled because of an issue found, so right now
lilly serves as a read-only database of starwars links only.

This means we don't need to hook to the links processing and post
anymore interlanguage links to lilly.


saipetch:
LGTM

PR #9558, opened 2016-01-26
RECIRC-35: Track recirculation impressions in GA

pauloslund:
There are various reasons why a user may not see the recirculation modules. By tracking impressions we will be able to properly measure funnels.

### DIFF
--- a/extensions/wikia/Recirculation/js/recirculation.js
+++ b/extensions/wikia/Recirculation/js/recirculation.js
@@ -1,17 +1,27 @@
 /*global define*/
 define('ext.wikia.recirculation.recirculation', [
 	'jquery',
+	'wikia.window',
 	'wikia.abTest',
 	'wikia.tracker',
 	'wikia.nirvana',
 	'videosmodule.controllers.rail',
 	'ext.wikia.adEngine.taboolaHelper'
-], function ($, abTest, tracker, nirvana, videosModule, taboolaHelper) {
+], function ($, w, abTest, tracker, nirvana, videosModule, taboolaHelper) {
 	'use strict';
 
 	function trackClick() {
 		tracker.track({
 			action: tracker.ACTIONS.CLICK,
+			category: 'recirculation',
+			label: 'rail',
+			trackingMethod: 'analytics'
+		});
+	}
+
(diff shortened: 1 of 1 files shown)


extensions/wikia/Recirculation/js/recirculation.js
```diff
@@ -1,12 +1,13 @@
...
 define('ext.wikia.recirculation.recirculation', [
 	'jquery',
+	'wikia.window',
 	'wikia.abTest',
 	'wikia.tracker',
 	'wikia.nirvana',
 	'videosmodule.controllers.rail',
 	'ext.wikia.adEngine.taboolaHelper'
-], function ($, abTest, tracker, nirvana, videosModule, taboolaHelper) {
+], function ($, w, abTest, tracker, nirvana, videosModule, taboolaHelper) {
```
ghost, line 10: Won't adding a param in the beginning break something else elsewhere? (backwards compatibility)

extensions/wikia/Recirculation/js/recirculation.js
```diff
@@ -52,12 +67,15 @@ define('ext.wikia.recirculation.recirculation', [
...
 				videosModule(element);
 				break;
+			default:
+				videosModule(element);
+				return;
 		}
 
-		$(element).on('click', 'a', trackClick);
+		trackImpression();
+		$(element).on('mousedown', 'a', trackClick);
```
ghost, line 78: Why mousedown instead of click? (just asking)

pauloslund, line 78: Mousedown is more reliable when the link that is clicked sends the user to an external page. See: https://github.com/Wikia/app/blob/dev/resources/wikia/modules/tracker.js#L320

extensions/wikia/Recirculation/js/recirculation.js
pauloslund, line 10: These parameters are the aliases for the dependencies defined in the array above. We don't really use this module anywhere else, so by adding `wikia.window` to the array at the beginning we can add `w` at the beginning as well.

ghost:
:+1:  go on :)

PR #9560, opened 2016-01-27
SOC-1820 

jsutterfield:
There were 2 problems with Special:Coppa, which we use to blank out underage user's information.

The first was fixed in SOC-1820, which this is branched off of. There was a bug in the logic we used to save `null` attributes. By fixing SOC-1820, Special:Coppa now deletes all the user's attributes from the attribute service as expected.

The second was an issue with `UserIdentityBox::resetUserProfile` which is what this branch specifically addresses. "Blank user profiles" on Special:Coppa was timing out because 2 database transactions were trying to operate on the same table. One transaction updates the timestamp for a row, and the second transaction deletes the row. The transaction to update the timestamp (called from `$user->saveSettings();`) doesn't do an explicit commit when it's done, so when the second transaction to delete the row comes along, the table is locked creating a deadlock.

The second transaction however (the one to delete the row), commits as soon as it's done. The fix then to swap the order of these transactions so that the one that actually commits runs first. The order doesn't matter b/c the end result is the same, the row is deleted.

Ticket: https://wikia-inc.atlassian.net/browse/SOC-1840


extensions/wikia/UserProfilePageV3/UserIdentityBox.class.php
```diff
@@ -28,6 +28,7 @@ class UserIdentityBox {
 
 	public $optionsArray = array(
 		'location',
+		'bio',
```
jsutterfield, line 31: `bio` is used by apps for their profiles. By adding this here, it's included in the attributes to be blanked out by `resetUserProfile`.

extensions/wikia/UserProfilePageV3/UserIdentityBox.class.php
```diff
@@ -577,11 +578,16 @@ public function resetUserProfile() {
 				$option = self::USER_PROPERTIES_PREFIX . $option;
 			}
 			$this->user->setGlobalAttribute( $option, null );
-
-			$this->user->saveSettings();
```
armonr, line 581: It's curious why this was being called in the loop!

armonr:
:shipit:

garthwebb:
LGTM

PR #9563, opened 2016-01-27
Dat 3623

rybmat:
fixing recognition of data tables


sebastianmarzjan:
Looks reasonable. though I think it may be addressing only the safesubst case. Can you add a test for the other  example provided in the ticket (quote)?

sebastianmarzjan:
LGTM (test failure unrelated)

PR #9572, opened 2016-01-27
PLATFORM-1355: Temporary disable some logging and create a new logging to track down if we are passing correct data to Mustache template

inez:


owend:
To boldly go :space_invader:

PR #9573, opened 2016-01-27
SEO-252 Interwiki links sometimes add an extra "/Wiki/"

gabrys:
The problematic part is the %2F in the URL. Apache treats that as an
incorrect URL
(http://httpd.apache.org/docs/2.2/mod/core.html#allowencodedslashes) and
fires the 404 handler. Before the changes the ErrorDocument was
index.php?title=Special:Our404Handler and the title was ignored if the
(URL-decoded) QUERY_STRING matched the MediaWiki $wgArticlePath.

With the updated code the 404 goes to the /redirect-canonical.php which
takes you to /wiki/(the-original-path). This code was not meant to
handle requests with /wiki/ in PATH.

We're fixing the redirect done in SpecialInterwikiDispatcher_body.php by
replacing the PHP's urlencode with MediaWiki's wfUrlencode, which
encodes fewer characters (for instance it leaves / and : intact).

Also we can extend the redirect-canonical.php to also work in the case
it's called as a broken-URL handler.

A comment mentioning the %2F breaks URLs was updated to point to the
Apache docs on AllowEncodedSlashes.

Redirect-canonical.php was also updated to output a X-Redirected-By
header (with a special version for %2F version) for easier debugging of
similar issues in the future.

### DIFF
--- a/extensions/wikia/InterwikiDispatcher/SpecialInterwikiDispatcher_body.php
+++ b/extensions/wikia/InterwikiDispatcher/SpecialInterwikiDispatcher_body.php
@@ -128,7 +128,7 @@ class InterwikiDispatcher extends UnlistedSpecialPage {
 
 						//RT#54264,#41254
 						$sArticleTitle = str_replace(' ', '_', $sArticleTitle);
-						$sArticleTitle = urlencode($sArticleTitle);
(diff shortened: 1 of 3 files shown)


saipetch:
LGTM

PR #9577, opened 2016-01-28
SUS-58 Add backlinks as PhalanxStats subtitle

adamkarminski:
@macbre It's much simpler. :)


macbre:
Looks jolly good to me :)

PR #9581, opened 2016-01-28
CE-3288: Fix translations in Global Shortcuts

lukaszkonieczny:
- Remove ImageReviewMercury from special pages list
- Add short description for Special:Templates
- Make suggestions search case insensitive


Grunny:
LGTM.

adamkarminski:
Merging since the Localization QA team will handle testing on the stable environment.

PR #9583, opened 2016-01-28
PLATFORM-1828: Log each and every queued task

wladekb:
We need more insight into the Job Queue system. This will help us know more when and why tasks are queued.

https://wikia-inc.atlassian.net/browse/PLATFORM-1828

/cc @macbre


macbre:
:+1:

PR #9584, opened 2016-01-28
(CE-775) Exclude non-existent users on Special:ListUsers

Grunny:
Make sure we don't list non-existent users on Special:ListUsers. Prevent
rows with empty usernames in the events table from being returned which
will catch the majority of these issues, and fix the check for whether or
not the user exists so it works properly.

### DIFF
--- a/extensions/wikia/Listusers/SpecialListusers_helper.php
+++ b/extensions/wikia/Listusers/SpecialListusers_helper.php
@@ -129,7 +129,7 @@ class ListusersData {
 			$dbs = wfGetDB( DB_SLAVE, array(), $this->mDBh );
 
 			/* initial conditions for SQL query */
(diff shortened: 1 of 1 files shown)


adamkarminski:
Is it worth to implement a self-cleaning mechanism that removes non-existent users?

Grunny:
@adamkarminski This was a pretty isolated event and I don't know of this ever being the case outside of these autocreated bot users, so given the legacy nature of the events_local_users table and all the scribe infrastructure, I don't think it's worth investing time in anything more complex than this simple check. :) The `user_name != 0` check in the queries will remove all the bad data from being returned in this case, and doesn't affect the query performance.

adamkarminski:
Ok, LGTM then :)

PR #9589, opened 2016-01-29
INT-322 New Upload Image / Video dialogs. Optimizely A/B test integra…

perjg:
…tion.

Overview
As part of the user testing we did in Japan, we realized that many users found it hard to even find where an image could be added inside the Visual Editor because it is probably one of the most commonly used forms of contribution, yet the button is hidden under a dropdown
Goals
The goal is to run an A/B test with a version of the VE that shows the image upload functionality right on the tool bar.


extensions/VisualEditor/VisualEditor.i18n.php
```diff
@@ -493,6 +493,11 @@
   'visualeditor-wikitext-warning-link' => 'mediawikiwiki:Special:MyLanguage/Help:Formatting',
   'visualeditor-wikitext-warning-title' => 'Wikitext markup detected',
   'visualeditor-dialog-media-insert-title' => 'Insert media',
+  'visualeditor-dialog-image-insert-title' => 'Insert an image',
+  'visualeditor-dialog-video-insert-title' => 'Insert a video',
+  'wikia-visualeditor-dialogbutton-wikiamediainsert-tooltip' => 'Media',
+  'wikia-visualeditor-dialogbutton-imageinsert-tooltip' => 'Movie',
```
inez, line 499: Is that really correct value? - "Movie" for wikia-visualeditor-dialogbutton-**image**insert-tooltip

extensions/VisualEditor/wikia/modules/ve/init/ve.init.wikia.Optimizely.js
```diff
@@ -0,0 +1,19 @@
+/*!
```
inez, line 1: Few things:
- perhaps name of this method should start with a prefix "get"
- is there really a need to create another file for it? why not to put it it in ve.utils?
- please do comply with VE code formatting standards (comment at the to of the file, and brackets around method calls and statements - if)

extensions/VisualEditor/wikia/modules/ve/init/ve.init.wikia.ViewPageTarget.js
```diff
@@ -32,45 +32,90 @@ OO.inheritClass( ve.init.wikia.ViewPageTarget, ve.init.mw.ViewPageTarget );
...
-	{
-		header: OO.ui.deferMsg( 'visualeditor-toolbar-table' ),
-		type: 'list',
-		icon: 'table-insert',
-		indicator: 'down',
-		include: [ { group: 'table' } ],
-		demote: [ 'deleteTable' ]
-	}
-];
+if (ve.init.wikia.ToolbarABTestVariantNumber() === 1) {
```
inez: Instead of having two very similar arrays I suggest considering defining array once and then modifying it slightly - in case of the experiment (that's what we've done before for similar experiment).

extensions/VisualEditor/wikia/modules/ve/ui/styles/images/icons/image.svg
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
inez, line 1: Run in through SVG optimizer to remove all the junk

extensions/VisualEditor/wikia/modules/ve/ui/styles/images/icons/video2.svg
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
inez: Same as above.

extensions/VisualEditor/wikia/modules/ve/ve.track.js
```diff
@@ -37,9 +37,13 @@ require( ['wikia.tracker'], function ( tracker ) {
 				};
 			},
 			'mwtiming.performance.system.activation': function ( data ) {
+				// INT-322 A/B test: Image / Video upload for toolbar
+				var label = (ve.init.wikia.ToolbarABTestVariantNumber() === 1) ?
```
inez: It's ok to generate new event (another one), but the original one should still be generated. In the future we will forget about the experiment and we will be curious why we had sudden dive and then come back in "edit-page-ready" events.

extensions/VisualEditor/wikia/modules/ve/ui/dialogs/ve.ui.WikiaImageInsertDialog.js
```diff
@@ -0,0 +1,230 @@
+/*!
```
inez, line 1: Files ve.ui.WikiaImageInsertDialog.js and ve.ui.WikiaVideoInsertDialog.js are huge copy & pastes which unfortunately makes this solution unacceptable. There are plenty of resource on the Internet explaining why code copy & paste is wrong but to name just few:
- from looking at this diff it's actually impossible to say what are the "slight" but important differences between those two files
- if we need to fix bugs - we need to do in two places (which is not common so it makes it more likely that one of the places gonna be missed)
- we lose history of the changes

Please consider what I was suggesting before, create new class that inherits from the already existing one and just overwrite some of the things (very common pattern in general and in VE specifically). I'm available for a meeting if you want to discuss it.

inez:
DO NOT MERGE (until code review feedback is addressed)

extensions/VisualEditor/wikia/modules/ve/ui/dialogs/ve.ui.WikiaImageInsertDialog.js
perjg, line 1: My intention here was to do what you suggested: I use inheritClass to make subclasses of WikiaMediaInsertDialog. If you were suggesting some other type of subclassing, you'll have to clarify.

For example, WikiaImageInsertDialog inherits from the original media dialog:
`
OO.inheritClass( ve.ui.WikiaImageInsertDialog, ve.ui.WikiaMediaInsertDialog );
`

I override only these two functions:
ve.ui.WikiaImageInsertDialog.prototype.initialize
ve.ui.WikiaImageInsertDialog.prototype.insertPermanentMediaCallback

And these are the only ones I copy, the other functions are used directly from WikiaMediaInsertDialog and are not copied.

I do however copy all the .static. variables in the top part of the class: I checked oo.inheritClass and I see this at the end:

`// Extend static properties - always initialize both sides
    oo.initClass( originFn );
    targetFn.static = Object.create( originFn.static );
`

so I think that means that the static properties are copied too (I had assumed only .prototype. was copied like what happens when you use JS new). I will try and remove the redundant .static. variables and see if it still works as intended.

(PS: I will address your other comments after the weekend)

extensions/VisualEditor/wikia/modules/ve/ui/styles/images/icons/image.svg
perjg, line 1: Is there a specific SVG optimizer you use?

extensions/VisualEditor/wikia/modules/ve/ui/dialogs/ve.ui.WikiaImageInsertDialog.js
inez, line 1: Static properties (in `.static.`) are inherited as well. However regardless of that I would like to encourage you to try to find better solution when you find yourself copying & pasting code - for instance method `formatPolicy` which is exactly the same in both cases.
I understand that methods `insertPermanentMediaCallback` and `initialize` are very similar but not exactly same - but it is hard to say from looking at this diff. If there is an alternation to default behavior required and currently that behavior is implemented somewhere within `intialize` for instance then I recommend extracting that logic (essentially splitting `initialize` in the base class to few smaller methods) and then overriding those small methods.
This way you not only achieve what you need to the scope of your current ticket but also improve overall quality of our codebase (most likely)

extensions/VisualEditor/wikia/modules/ve/init/ve.init.wikia.Optimizely.js
perjg, line 1: Sure, I can change the name to getToolbarABTestVariantNumber.

It can also be moved to ve.utils if you want:
My reason for making a separate file is that there are more VE A/B tests upcoming (INT-328, INT-304, INT-307), so my intention was simply to group everything together in this new file as we will likely want to run several A/B tests at the same time.

Let me know if I should still move it.

Code formatting: This file failed the linter on my first commit, but the second commit should already have fixed it. Is it possible to run the linter locally before commiting?

perjg:
@inez I've cleaned up most of your comments:
- I don't know about SVG optimizer, can you let me know which one to use?
- getToolbarABTestVariantNumber is still in a separate file, see comment above

extensions/VisualEditor/wikia/modules/ve/init/ve.init.wikia.Optimizely.js
```diff
@@ -0,0 +1,19 @@
...
+ * Utility functions for running Optimizely A/B tests with VisualEditor
+ */
+
+( function () {
+	/**
+	 * @returns {number} The A/B test variant number for the Toolbar A/B test
+	 */
+	ve.init.wikia.getToolbarABTestVariantNumber = function () {
```
inez: I'm not an expert in Optimizely but shouldn't this experiment be activate if and only if user starts loading VisualEditor (instead of activating it on any page view?)

extensions/VisualEditor/wikia/modules/ve/init/ve.init.wikia.Optimizely.js
```diff
@@ -0,0 +1,19 @@
...
+		var optimizely = window.optimizely,
+			optimizelyId = window.wgDevelEnvironment ? '4721410313' : '4701112678';
+
+		if (optimizely && optimizely.variationMap && optimizely.variationMap.hasOwnProperty(optimizelyId)) {
+			return optimizely.variationMap[optimizelyId] || 0;
```
inez: Is this a proper documented way of retrieving variation from Optimizely? I can't see that in their docs - the most similar one I see is: `optimizely.data.state.variationMap`
Btw. in Optimizely you can configure different variations to execute different JS when activated.
So for instance variation 1 could set some global variable to different value than variation 2. That's what we used to do in the past - worked great - but we were not experimenting too much.

extensions/VisualEditor/wikia/modules/ve/ui/dialogs/ve.ui.WikiaImageInsertDialog.js
```diff
@@ -0,0 +1,50 @@
...
+ */
+ve.ui.WikiaImageInsertDialog = function VeUiMWImageInsertDialog( config ) {
+	// Parent constructor
+	ve.ui.WikiaImageInsertDialog.super.call( this, config );
+};
+
+/* Inheritance */
+
+OO.inheritClass( ve.ui.WikiaImageInsertDialog, ve.ui.WikiaMediaInsertDialog );
+// OO.inheritClass( ve.ui.WikiaImageInsertDialog, ve.ui.FragmentDialog );
```
inez: No need for this comment.

extensions/VisualEditor/wikia/modules/ve/ui/dialogs/ve.ui.WikiaMediaInsertDialog.js
```diff
@@ -151,7 +152,12 @@ ve.ui.WikiaMediaInsertDialog.prototype.initialize = function () {
 
 	this.mainPage = new OO.ui.PageLayout( 'main', { $content: this.$mainPage } );
 	this.searchPage = new OO.ui.PageLayout( 'search', { $content: this.search.$element } );
-	this.pages.addPages( [ this.mainPage, this.searchPage ] );
```
inez, line 154: Instead of renaming `initialize` method and having `initializeWrapper` take parameter would it make sense to create new method called `addPages` that default implementation be `this.pages.addPages( [ this.mainPage, this.searchPage ] );` and `this.pages.addPages( [ this.searchPage ] );` after overwriting?
(not sure if the method should be called exactly addPages - but you get the idea... could be also handleAddingPapes)

extensions/VisualEditor/wikia/modules/ve/ui/dialogs/ve.ui.WikiaMediaInsertDialog.js
```diff
@@ -600,7 +613,7 @@ ve.ui.WikiaMediaInsertDialog.prototype.insertPermanentMedia = function ( cartIte
  * @param {Object} items Items to insert
  * @param {ve.dm.SurfaceFragment} fragment
  */
-ve.ui.WikiaMediaInsertDialog.prototype.insertPermanentMediaCallback = function ( items, fragment ) {
+ve.ui.WikiaMediaInsertDialog.prototype.insertPermanentMediaCallbackWrapper = function ( items, fragment, label ) {
```
inez: Why not just create a property such as `trackingPrefix` and have if very between different classes? Very common pattern in VE - and proven to works nice at scale.

extensions/VisualEditor/wikia/modules/ve/ui/dialogs/ve.ui.WikiaMediaInsertDialog.js
inez, line 154: You can also just overwrite intialize method, make it call it parent method, and then undo what you "don't want", e.g.: `this.pages.removePage(this.mainPage)`

inez, line 154: It would be great if that approach could work also for hiding what currently you are hiding with class `ve-ui-wikiaMediaQueryWidget-uploadWrapper-video`

extensions/VisualEditor/wikia/modules/ve/ui/styles/images/icons/image.svg
inez, line 1: I believe we use this one https://petercollingridge.appspot.com/svg-editor but I'm not sure about exact options - you would have to play with it a bit and look what output it generates.

extensions/VisualEditor/wikia/modules/ve/init/ve.init.wikia.Optimizely.js
perjg: Everything on optimizely.data.state is also replicated directly on the optimizely object. I can't find this explicitly in the documentation, but optimizely.variationMap is used directly throughout Mercury.

I know it is possible to inject javascript too, but I was explicitly advised to just use variationMap by the Mercury team to avoid polluting the global namespace.

extensions/VisualEditor/wikia/modules/ve/init/ve.init.wikia.Optimizely.js
perjg: Where the experiment is active depends on how the url filter is configured in Optimizely, so there is no need to check for it here:

If the experiment is active _and_ you are in the test group, getToolbarABTestVariantNumber will return 1, if it is inactive or you are not in the test group, it will return 0.

perjg:
@inez I've cleaned up your remaining comments.
The optimizely implementation is still the same. It works properly as-is, and follows the same style as Optimizely A/B tests in Mercury.

extensions/VisualEditor/wikia/modules/ve/init/ve.init.wikia.Optimizely.js
inez: Sure, but I just want you to aware. optimizely experiment might be activated right away on the page view - when user opens the page - or later with JavaScript - when user for instance opens visualeditor.

so in your case you will get more many users bucketed than users actually using visualeditor and optimizely perhaps will report that received data is conclusive before it really is.

and by activating I mean assigning to particular bucket/group.

it seems like you optimizely only for bucketing users, but not for tracking "success" / conversion ratio. just checking if that's the intention.

extensions/VisualEditor/wikia/modules/ve/ui/dialogs/ve.ui.WikiaMediaInsertDialog.js
```diff
@@ -72,6 +74,7 @@ ve.ui.WikiaMediaInsertDialog.prototype.getBodyHeight = function () {
 
 /**
  * @inheritdoc
+ * @param {boolean} Sets if the main page (insert image drop area) should be shown
```
inez: not needed anymore

extensions/VisualEditor/wikia/modules/ve/ui/dialogs/ve.ui.WikiaVideoInsertDialog.js
```diff
@@ -0,0 +1,44 @@
+/*!
+ * VisualEditor user interface WikiaVideoInsertDialog class.
+ */
+
+/**
+ * Dialog for inserting MediaWiki media objects.
+ *
+ * @class
+ * @extends ve.ui.FragmentDialog
```
inez: please correct comments like this one

extensions/VisualEditor/wikia/modules/ve/ve.track.js
```diff
@@ -199,6 +199,11 @@ require( ['wikia.tracker'], function ( tracker ) {
 		funnelEvent = data.label + '/' + data.action;
 
 		if ( funnelEvents.indexOf( funnelEvent ) !== -1 ) {
+			if (ve.init.wikia.getToolbarABTestVariantNumber() === 1 && funnelEvents === 'edit-page-ready/impression') {
+				// Extra impression event when toolbar A/B test is active
```
inez: in ve everything is tracked as an event, and subset of everything is tracked as a fake page view. after this change we will have fake page view which does not have a corresponding event. just making sure that's what you want. (usually we use events for tracking particular features metrics / success)

extensions/VisualEditor/wikia/modules/ve/init/ve.init.wikia.Optimizely.js
perjg: I think where the experiment is active depends on how you set up the URL filter in Optimizely.

The code in ve.init.wikia.Optimizely.js doesn't activate the experiment in any way, it just provides 
getToolbarABTestVariantNumber which is called on VE init from JS. 

With the alternative method you describe (setting a global variable to control the experiment) still wouldn't change this, the experiment will still run on any url that matches what is set in Optimizely. Or were you referring to some other way of controlling exactly when the experiment is activated programmatically?

I can probably get around this by adding &veaction=edit to the url filter, thus making sure it is only activated when the VE is loaded.

perjg: It might be the closure around the contents of the file which is confusing. I was just emulating code style from elsewhere, and it isn't required here since ve.init.wikia is set in ve.init.wikia.js which is loaded before anyway. I'll remove it.

perjg:
@inez I've changed the following:
- Moved the extra edit-page-ready-toolbartest to track so that it triggers both an event and a fake page view like you suggested.
- Removed the closure in the optimizely file as it wasn't needed and only made the intent of the function confusing.
- Fixed the comments you mentioned

inez:
Overall it looks good to me but the checks are failing. I still want to clarify one thing.

Regarding Optimizely I think there might still be a confusion around those two different and distinct things:
- experiment activation - bucketing
- retrieving information about the bucket to which visitor got assigned after experiment activation

Experiment might be and often is activated by matching the URL. I'm not sure if such matching would work in case of VE where veaction=edit is placed in the URL - but there is no actual page reload, it is just added there with browser history/state API. I recommend testing it (preferably in isolation - simple HTML page).
So if it works - that's fine - if it doesn't then activating experiment programmatically in JS when editor finishes loading is the way to go (unless you are ok with just activating experiment for all visitors - even if they never open VE).

Above has nothing really to do with retrieving information about the bucket - it can still be retrieved from `optimizely` super global object - or within snippet of code which for example sets variable in a global scope (that snippet would be executed if you activate experiment programmatically).

I hope this clarifies my point and the fact that experiment activation and retrieving state are two different things.

perjg:
Tests are passing again, it was just me forgetting a semicolon.

Good point about filtering by veaction in optimizely: It won't work since there is no page reload.

It looks like there is a a way of programmatically activating the experiment:
https://help.optimizely.com/hc/en-us/articles/200040225-Activation-Mode-Activating-an-experiment-dynamically-after-a-page-has-loaded

I'll try that out, the buckets be assigned only for people who load the VE:
I'll activate it in ve.init.wikia.ViewPageTarget.js, which is the first place where the A/B test is queried on loading.

inez:
Yes. We've been using "manual" activation before - it works well.
Btw. if you want you can also activate in a hook which is called when VE activates successfully. Just look for `mw.hook` in VisualEditor directory (I'm pretty sure some event tracking is listening to that hook as well).

perjg:
@inez Commit 5a3f669 switches over to activate the experiment in the 've.activate' hook, which runs just before the toolbar is built. I've made the necessary changes in Optimizely too. This should give proper 50/50 buckets for actual VE loads.

perjg:
@inez INT-328 is based on this CR (adding translations). I've squashed all the commits on this CR and pushed them to a new CR, with the extra work for INT-328 added: https://github.com/Wikia/app/pull/9667

inez:
@perjg If your goal was to just squash commits then you don't have to create a new PR, instead do this:

```
Step 1: in your feature branch: git rebase -i origin/<target branch>
Step 2: mark your main commit (usually the first one) as "pick", and other commits as "squash"
Step 3: write a summary commit message
Step 4: push to origin with force! (git push --force)
```

this way we keep more history (the original PR, with the comments).

perjg:
@inez The only difference between what I did and your steps is that I did "git checkout -b INT-328" first to create a new branch.

The reason I did that is that INT-328 requires all the changes from this PR, which I wanted to squash to make it a bit cleaner, but I wanted a separate CR for the new changes.

Is there an official policy on squashing? I.e. is the procedure you describe of squashing + force push to an open PR in github before merging ok to use for any CR?

inez:
We do not have an official policy regarding squashing, some people actually prefer to see a lot of commits in the history, I personal don't find much value in commits such as "addressing CR", "fixing typo" and would rather read one, solid commit message :)
(Btw. it's highly recommended to never push force to dev - however that sometimes happen anyways)

When it comes to situation where one feature requires few tasks to be completed and those tasks can not be separately merged to dev what we do it to create a "feature branch" based on dev, and then create branches for tickets based on that feature branch. Code is to be code reviewed when it's merged from ticket branch to feature branch, and then there is no need to code review when merging from feature branch (at the end) to dev - however I always ask someone to review anyways.

perjg:
Got it, I'll use a feature branch off dev when I know there will be a chain of CR's before I can merge back to dev.

(in this case, I think INT-328 was created after I started work on INT-322, as the original intention was just to A/B test for Japan)

PR #9592, opened 2016-01-29
CE-3292: Added translations for some special pages

lukaszkonieczny:
Added ES and RU translations for some special pages


Grunny:
LGTM.

PR #9593, opened 2016-01-29
CE-2505: Do not allow move js pages to MW or USER namespace

lukaszkonieczny:
Do not allow move js pages to MW or USER namespace

### DIFF
--- a/extensions/wikia/ProtectSiteJS/ProtectSiteJS.class.php
+++ b/extensions/wikia/ProtectSiteJS/ProtectSiteJS.class.php
@@ -74,7 +74,7 @@ class ProtectSiteJS {
 	 */
 	public static function onGetUserPermissionsErrors( \Title $title, \User $user, $action, &$result ) {
 		if ( $action === 'move' && ( $title->isJsPage() || $title->isJsSubpage() ) && !$user->isStaff() ) {
-			$result = ['badaccess-groups', \User::getGroupName( 'staff' ), 1];
+			$result = [ 'badaccess-groups', \User::getGroupName( 'staff' ), 1 ];
 			return false;
 		}
 
--- a/extensions/wikia/ProtectSiteJS/ProtectSiteJS.class.php
+++ b/extensions/wikia/ProtectSiteJS/ProtectSiteJS.class.php
@@ -63,6 +63,15 @@ class ProtectSiteJS {
 		return false;
 	}
 
+	/**
+	 * Allow moving js pages to Mediawiki or User namespace only for staff users
+	 *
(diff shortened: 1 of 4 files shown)


Grunny:
I think this might be better in the `ProtectSiteJS` so it is in the same place as the editing restrictions on JS pages and will always loaded. What do you think?

lukaszkonieczny:
Yes, that make sense :)

extensions/wikia/ProtectSiteJS/ProtectSiteJS.class.php
```diff
@@ -64,6 +64,24 @@ private static function isUserSkinJS( Title $title, User $user ) {
...
+	 *
+	 * @param Title $title
+	 * @param User $user
+	 * @param string $action
+	 * @param array $result
+	 * @return bool
+	 */
+	public static function onGetUserPermissionsErrors( \Title $title, \User $user, $action, &$result ) {
+		if ( $action === 'move' && ( $title->isJsPage() || $title->isJsSubpage() ) && !$user->isStaff() ) {
+			$result = ['badaccess-groups', \User::getGroupName( 'staff' ), 1];
```
Grunny: Array should be spacey per coding conventions.

Grunny:

Grunny:
Though, should this only be for personal (i.e. user) JS pages? If you move a MW JS page, it will need to go through content review again anyway, won't it? I'm more thinking of whether we should allow users to move JS pages within the MW namespace, i.e. MediaWiki:Test.js -> MediaWiki:Test2.js or whatnot.

lukaszkonieczny:
Per https://wikia-inc.atlassian.net/browse/CE-3107 we should not allow users to move js pages within MW namespace

onei:
Spotted this while looking for something else, and thoght I'd give my 2 cents as someone who works with JS pages more than most users on Wikia.

Not sure on what CE-3107 is exactly, but I'd agree with preventing moving pages to NS_MEDIAWIKI from other namespaces, as well as preventing moving .js pages to un-suffixed names and vice versa.

Personally I'd like to be able to rename JS pages within those restrictions which enables me as an admin to move pages to a more relevant name as the script evolves over time, standardise page names or even just fix a spelling mistake in the title. You're not going to prevent people from moving the content via copy/paste so why not just allow it in the proper way and preserve page history? Page history is super useful at times, as is a file's commit history if I was doing something similar in git.

azarek:
This looks good to me when it comes to not be able to move js to non js as any other user than staff, so this might be merged. Further steps about .js move will appear in other tickets.

PR #9594, opened 2016-01-29
CE-3086 Use a localisable message for a read-only mode

adamkarminski:
Using a message instead of a hard-coded string will allow us to localise the message. If the solution passes CR - I'll start the localisation process.

@Wikia/community-engineering  @macbre


macbre:
:+1:

adamkarminski:
Translations are in, ping @macbre :)

macbre:
Jolly good, let's merge it!

PR #9595, opened 2016-01-29
ADEN-2862 Move krux to Analytics Engine

fraszczakszymon:
Call krux at the top of page (using Analytics Engine), but after bidders.

### DIFF
--- a/extensions/wikia/AnalyticsEngine/AnalyticsProviderKrux.php
+++ b/extensions/wikia/AnalyticsEngine/AnalyticsProviderKrux.php
@@ -1,0 +1,29 @@
+<?php
+
+class AnalyticsProviderKrux implements iAnalyticsProvider {
+
+	private static $siteId = 'JU3_GW1b';
+	private static $template = 'extensions/wikia/AnalyticsEngine/templates/krux.mustache';
+
+	function getSetupHtml( $params = array() ) {
+		return null;
+	}
+
+	function trackEvent( $event, $eventDetails = array() ) {
+		switch ($event) {
+			case AnalyticsEngine::EVENT_PAGEVIEW:
+				return $this->trackPageView();
+			default:
+				return '<!-- Unsupported event for Krux -->';
+		}
+	}
+
+	private function trackPageView() {
+		return \MustacheService::getInstance()->render(
+			self::$template,
+			[
+				'siteId' => self::$siteId
+			]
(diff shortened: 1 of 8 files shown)


drets:
LGTM :+1:

extensions/wikia/AnalyticsEngine/AnalyticsProviderKrux.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+class AnalyticsProviderKrux implements iAnalyticsProvider {
+
+	private static $siteId = 'JU3_GW1b';
```
nandy-andy, line 5: Can we check in how many places we've got this id? Do you think it's fine keeping it in public repo?

extensions/wikia/AssetsManager/config.php
```diff
@@ -198,10 +198,13 @@
 	'skin' => [ 'oasis', 'wikiamobile' ],
 	'assets' => array(
 		'//extensions/wikia/AdEngine/js/AdContext.js',
+		'//extensions/wikia/AdEngine/js/AdLogicPageParams.js',
+		'//extensions/wikia/AdEngine/js/AdLogicPageViewCounter.js',
 		'//extensions/wikia/AdEngine/js/AdTracker.js',
 		'//extensions/wikia/AdEngine/js/slot/adSlot.js',
 		'//extensions/wikia/AdEngine/js/utils/AdLogicZoneParams.js',
-		'//extensions/wikia/AdEngine/js/lookup/lookupFactory.js'
+		'//extensions/wikia/AdEngine/js/lookup/lookupFactory.js',
```
nandy-andy: Shouldn't it be few places up? ;)

extensions/wikia/AnalyticsEngine/AnalyticsProviderKrux.php
fraszczakszymon, line 5: In single repo we have only one occurrence of this id. I think it's ok because it's a part of request sent to krux for library (it's visible anyway).

nandy-andy, line 5: Oh, yeah! You're right. Thanks :)

nandy-andy:
LGTM - just few minor notes :japanese_ogre:

PR #9597, opened 2016-01-29
XW-41 | Use new format of Ember component placeholder for single images

rogatty:
https://wikia-inc.atlassian.net/browse/XW-41


extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
```diff
@@ -1,20 +1,89 @@
...
 	];
 
-	const ICON_MAX_SIZE = 48;
 	const CACHE_VERSION = '0.0.3';
+	const CACHE_VERSION_FOR_SEO_FRIENDLY_IMAGES = 1; // for $wgEnableSeoFriendlyImagesForMobile
+
+	const ICON_MAX_SIZE = 48;
+	// Line height in Mercury
+	const ICON_SCALE_TO_MAX_HEIGHT = 20;
+	const MAX_MERCURY_IMAGE_WIDTH = 985;
```
rogatty: This name might be misleading. How about `IMAGE_FALLBACK_SIZE`?

extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
```diff
@@ -322,4 +450,57 @@ private static function isIconSize( $sizeParam ) {
...
+			$width = intval( $maxHeight * $originalWidth / $originalHeight );
+		}
+
+		return [
+			'height' => $height,
+			'width' => $width
+		];
+	}
+
+	private static function getMediaDetailWithSizeFallback(
```
rogatty, line 504: `getMediaDetailsWithSizeFallback`?

extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
```diff
@@ -322,4 +450,57 @@ private static function isIconSize( $sizeParam ) {
...
+		$title, $mediaDetailConfig, $fallbackSize=self::MAX_MERCURY_IMAGE_WIDTH
+	) {
+		$mediaDetail = WikiaFileHelper::getMediaDetail( $title, $mediaDetailConfig );
+		if ( $mediaDetail['width'] == 0 ) {
```
rogatty: If we don't want to use `===` (kinda makes sense) then maybe this should be `if ( empty(...) )`?
http://php.net/manual/en/function.empty.php

extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
```diff
@@ -322,4 +450,57 @@ private static function isIconSize( $sizeParam ) {
...
+			$mediaDetail['width'] = $fallbackSize;
+
+			\Wikia\Logger\WikiaLogger::instance()->error(
+				'Media width was empty - fallback to fallbackSize',
```
rogatty: How about considering https://docs.google.com/document/d/1rGZ0we9E0iNFu2Y5SO89m3mCK2uALvvCwEUQQ_OS8qE and using more meaningful message here?

E.g. "ArticleAsJson - Media width was empty, using fallback size". Also keys should be underscored, so `media_details` and `fallback_size`.

extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
bognix, line 504: I wanted to be consistent with `WikiaFileHelper::getMediaDetail`

extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
bognix: On the other hand this describes what this value represents not where it's used.
If it's misleading I can add comment in `getMediaDetailWithSizeFallback` why we use this value, makes sense?

rogatty: Ok, but I would add a comment here that it's the max viewport width seen in GA on mobile devices.

extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
rogatty, line 504: Probably someone made a typo there? It returns more than one detail, after all.

PR #9600, opened 2016-01-30
VE-2123: Track if PCRE errors are actually happening

inez:
This is a step towards starting throwing exceptions when those errors are happening

### DIFF
--- a/includes/parser/Parser.php
+++ b/includes/parser/Parser.php
@@ -1264,6 +1264,11 @@ class Parser {
 					[0-9Xx]                 # check digit
 					\b)
 			)!xu', array( &$this, 'magicLinkCallback' ), $text );
+
+		if ( preg_last_error() !== 0 ) {
+			\Wikia\Logger\WikiaLogger::instance()->error( 'PCRE error', [ 'preg_last_error' => preg_last_error() ] );
+		}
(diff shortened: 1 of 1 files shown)


inez:
@macbre Could you CR this for me please?

includes/parser/Parser.php
```diff
@@ -1264,6 +1264,11 @@ function doMagicLinks( $text ) {
 					[0-9Xx]                 # check digit
 					\b)
 			)!xu', array( &$this, 'magicLinkCallback' ), $text );
+
+		if ( preg_last_error() !== 0 ) {
+			\Wikia\Logger\WikiaLogger::instance()->error( 'PCRE error', [ 'preg_last_error' => preg_last_error() ] );
```
macbre, line 1269: Please add `"exception" => new Exception()` field to the context. This will allow us to have a full backtrace.

inez, line 1269: Well, it would, but I'm not interested yet in having full backtrace (especially that I know where the logger is called). If the idea is that every error should log an Exception then probably it should implemented in the logger itself.

macbre, line 1269: Ok, up to you :)

PR #9601, opened 2016-01-30
VE-2124: Make sure to set $wgArticleAsJson back to false even if Exception is thrown

inez:
ArticleApiController might be called via FauxRequest which does not separates scopes between caller and callee - so this potentially could affect caller scope.


macbre:
:+1:

PR #9603, opened 2016-01-30
VE-2125: There is no need to use mw.loader.use anymore because mw.loader.using returns promise as well

inez:
This is related to my change from couple months ago, where I made mw.loader.using get callback but also return promise.


inez:
DO NOT CR YET - still working on it

wladekb:
@inez thanks for taking the time to clean it up!

inez:
@wladekb It's good for the code review now. Thank you.

wladekb:
@inez sorry for the late response, lgtm

PR #9608, opened 2016-02-01
XW-1036 | Make sure we test localStorage API

vforge:
https://github.com/Wikia/mercury/pull/1916

/cc @nandy-andy


nandy-andy:
LGTM :+1: Thanks for the comment - it's longer than I expected ;)

The `VideoPageToolAssetTest.php:163` likely isn't connected to the changes.

:trollface: comment: have you greped our code to check if there is no JS piece of code getting anything from `localStorage` from `'foo'` key? :trollface:

vforge:
Yes, to both of them :)

rogatty:
LGTM

PR #9620, opened 2016-02-02
MAIN-6319: move logEmailChanges hook to a global location

owend:
This hook was only installed where EditAccount was active (community wiki), but it should be global.

@Grunny


Grunny:
LGTM.

PR #9623, opened 2016-02-02
DAT-3637 Migration script from Promote to Curated Content

sebastianmarzjan:
This is going to be ran only once, for the purpose of migrating content from Special:Promote to Curated Content.

### DIFF
--- a/extensions/wikia/CuratedContent/maintenance/SpecialPromoteToCuratedContent.php
+++ b/extensions/wikia/CuratedContent/maintenance/SpecialPromoteToCuratedContent.php
@@ -42,21 +42,21 @@ class SpecialPromoteToCuratedContentMigrator extends Maintenance {
 		if ( empty( $curatedContentData['community_data']['image_id'] ) ) {
 			$originalImageId = $curatedContentData['community_data']['image_id'];
 
-			// update image
+			// fetch image Id for update from corporate wiki for a given language
 			$image = ( new PromoImage( PromoImage::MAIN, $wgDBname ) );
 			$corporateWikiId = $cv->getTargetWikiId( $wgLang->getCode() );
-			$imageTitle = $image->getOriginFile()->getTitle();
+			$originFile = $image->getOriginFile($corporateWikiId);
+			$imageName = $image->getPathname();
+			$imageTitle = $originFile->getTitle();
+			$imageUrl = $originFile->getUrl();
 			$imageId = $imageTitle->exists() ? $imageTitle->getArticleID() : null;
 
 			if ( !empty( $imageId ) ) {
-				// we are good to go; image exists on the destination wiki
-			} else {
-				// image does not exist on this community
-				// we need to upload the image to the local wiki
+				// image exists on the corporate site. Since it's the final source of truth
+				// for apps (and unapproved images cannot be there) we upload it to the local wiki
 				// we first need to fetch it from the corporate wiki
-				$imageUrl = $image->getOriginFile( $corporateWikiId )->getUrl();
-				$imageName = $image->getPathname();
-
+				// and now we upload it, disabling the Special:Promote upload prevention
+				// implemented as validation hook
 				$this->disableUploadValidationHook();
 				$result = $this->uploadImageToLocalWiki( $imageName, $imageUrl );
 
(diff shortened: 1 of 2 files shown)


extensions/wikia/CuratedContent/maintenance/SpecialPromoteToCuratedContent.php
```diff
@@ -0,0 +1,163 @@
...
+	public function execute() {
+		global $wgCityId, $wgLang, $wgDBname;
+
+		$dryRun = $this->hasOption('dry-run');
+
+		$change = false;
+		$cv = new CityVisualization();
+		$communityDataService = new CommunityDataService( $wgCityId );
+
+		$curatedContentData = [
```
rybmat, line 27: I wrote a getter for whole structure of data, see https://github.com/Wikia/app/blob/DAT-3638b/includes/wikia/services/CommunityDataService.class.php#L148

extensions/wikia/CuratedContent/maintenance/SpecialPromoteToCuratedContent.php
```diff
@@ -0,0 +1,163 @@
...
+		if ( empty( $curatedContentData['community_data']['image_id'] ) ) {
+			$originalImageId = $curatedContentData['community_data']['image_id'];
+
+			// update image
+			$image = ( new PromoImage( PromoImage::MAIN, $wgDBname ) );
+			$corporateWikiId = $cv->getTargetWikiId( $wgLang->getCode() );
+			$imageTitle = $image->getOriginFile()->getTitle();
+			$imageId = $imageTitle->exists() ? $imageTitle->getArticleID() : null;
+
+			if ( !empty( $imageId ) ) {
```
rybmat, line 54: is it necessary? I don't like empty blocks

extensions/wikia/CuratedContent/maintenance/SpecialPromoteToCuratedContent.php
sebastianmarzjan, line 27: Yep, I can use that - it'd be useful to have that in dev though not to introduce artificial dependency between those two stories

extensions/wikia/CuratedContent/maintenance/SpecialPromoteToCuratedContent.php
sebastianmarzjan, line 54: I leaned on the side of explicity here.

extensions/wikia/CuratedContent/maintenance/SpecialPromoteToCuratedContent.php
```diff
@@ -0,0 +1,163 @@
...
+			);
+			$change = true;
+		}
+
+		if ( $dryRun ) {
+			$this->output( "\nDRY RUN: Skipping save of Curated Content\n" );
+		} else {
+			if ( $change ) {
+				$this->output( "\nSaving Curated Content\n" );
+				$result = $communityDataService->setCuratedContent( $curatedContentData );
```
rybmat, line 101: As you suggested me to add "reason" parameter to this method, https://github.com/Wikia/app/blob/DAT-3638b/includes/wikia/services/CommunityDataService.class.php#L14

sebastianmarzjan, line 101: Yes :)
As in https://github.com/Wikia/app/pull/9623/files#diff-49cc6a229edc8757921429da24e5a25eR27 - I can either rebase my brach onto yours, or modify per those two comments once DAT-3638b is merged

extensions/wikia/CuratedContent/maintenance/SpecialPromoteToCuratedContent.php
```diff
@@ -0,0 +1,163 @@
...
+
+		$curatedContentData = [
+			'curated' => $communityDataService->getCurated(),
+			'featured' => $communityDataService->getFeatured(),
+			'optional' => $communityDataService->getOptional(),
+			'community_data' => $communityDataService->getCommunityData()
+		];
+
+		if ( empty( $curatedContentData['community_data']['description'] ) ) {
+			$originalDescription = $curatedContentData['community_data']['description'];
```
rybmat: shouldn't be this assignment in else and the one from else here? Otherwise I don't understand something :p

sebastianmarzjan: Good catch :-)

rybmat:
LGTM, however in my opinion it would be nice to use code from provided links in comments

idradm:
it's good enough for one time migration script

PR #9627, opened 2016-02-02
SUS-76: force JSON response format in VideoHandlerController

macbre:
[SUS-76](https://wikia-inc.atlassian.net/browse/SUS-76)

Force the JSON response format as there's no HTML template for `getEmbedCode` and `getPremiumEmbedCode` methods and some JS entry points do not set `format=json` URL parameter.

@Wikia/sustaining-team


kvas-damian:
LGTM

PR #9628, opened 2016-02-02
VideoPageToolAssetTest: make the integration test more reliable

macbre:
[SUS-118](https://wikia-inc.atlassian.net/browse/SUS-118)

Actually wait for DB slave to catch up so that `VideoPageToolAsset::newAsset` can load a row from slave node.

Use `wfWaitForSlaves()` instead of `sleep(1)`. When I wait for a bus it does not always take exactly 5 minutes ;)

@Wikia/sustaining-team


kvas-damian:
LGTM

PR #9631, opened 2016-02-02
ADEN-2917 No proxy for external asset requests

nandy-andy:
As we are going to get rid of local varnish and point MW proxy to border servers we should update proxy configuration for all external calls from MW.

//cc @macbre @artursitarski


fraszczakszymon:
LGTM :+1:

PR #9632, opened 2016-02-02
SEO-115 Favicon is missing on section redirects

gabrys:
There seems to be an issue with Firefox handling of favicons in
combination with pdating the location.hash. "Reloading" the
`<link rel="shortcut icon">` seems enough to fix it


saipetch:
LGTM

PR #9633, opened 2016-02-02
VE-2126: Log VE<->Parsoid request duration and allow for switching caching off

inez:
Switching caching off is needed just for our future experiment to find out if degraded performance is actually affecting our VE related metrics.
Having duration logged will let us verify that indeed performance is degraded.


pauloslund:
LGTM (after the settings variable we talked about)

PR #9634, opened 2016-02-03
Helios Service Resolution and Cookie Deletion

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-1139

Change the DI container to resolve Helios the same way other services are resolved, rather than relying on a known host (`helios.service.consul`) and a known port. This allows us to use a different service more easily when failing over to Reston (the new `unavailable` service). There's a config change required for this, as well: Wikia/config#1600

### DIFF
--- a/includes/User.php
+++ b/includes/User.php
@@ -1151,14 +1151,14 @@ class User {
 		}
 
 		if ( !$this->isUserAuthenticatedViaAuthenticationService() ) {
-			global $wgAuthForceLogoutOnBadCredentials;
+			global $wgAuthForceLogoutOnFailedAuthSessionFallback;
 
 			/*
 			 * when failing over to Reston (readonly), users should be anonymous but their session
 			 * should resume without them having to do anything when we go back to r/w. See
 			 * SERVICES-1125
 			 */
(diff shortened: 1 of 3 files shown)


drsnyder:
Hm. We'll need to make the same change in the api-gateway [here](https://github.com/Wikia/api-gateway/blob/master/templates/src/config.lua#L5). Instead of using the value at that key we'll have to use the service lookup result.

nmonterroso:
@drsnyder oh.. does the api gateway have the ability to do a service lookup using consul?

drsnyder:
Not directly. All of the service lookup is done via the templates [here](https://github.com/Wikia/api-gateway/tree/master/templates). So the lua code doesn't lookup the service directly but it's config will be updated and nginx will be restarted if something changes in consul.

In this case, if we "switched" `prod.helios` to `prod.unavailable`, `consul-template` will pick it up, re-write the config, and reload nginx.

nmonterroso:
I've created https://wikia-inc.atlassian.net/browse/SERVICES-1147 to track the API gateway part. Let's figure something out there and keep this discussion to the MW portion.

ArturKlajnerok:
Lgtm :+1:

PR #9635, opened 2016-02-03
Fix ExactTarget_* class autoloading

drsnyder:
The celery tasks for the `ExactTargetUpdate` extension are spewing errors like the following regarding `ExactTarget_RetrieveRequest` (but not limited to that class):

```
PHP Fatal Error: Class 'ExactTarget_RetrieveRequest' not found in /usr/wikia/slot1/9585/src/extensions/wikia/ExactTargetUpdates/api/ExactTargetApiHelper.php on line 64
```

This is because the `ExactTarget_*` classes (about 200) are declared in [exacttarget_soap_client.php](https://github.com/Wikia/app/blob/dev/extensions/wikia/ExactTargetUpdates/lib/exacttarget_soap_client.php) along with the `ExactTargetSoapClient` class. The autoloading in the setup class only references [`ExactTargetSoapClient`](https://github.com/Wikia/app/blob/dev/extensions/wikia/ExactTargetUpdates/ExactTargetUpdates.setup.php#L78) which means you have to reference `ExactTargetSoapClient` (e.g. by creating a new object) before you can reference any of the `ExactTarget_*`. This creates a load order dependency that results in class not found exceptions if it's not followed.

This PR addresses the issue by moving all of the `ExactTarget_*` classes into their own files one per class. The class loading in the setup file is also adjusted to include each class separately. Finally a test was added to confirm the loading works as expected.

Finally, I removed the `testDeleteSubscriber` test. It was testing the mock and was not confirming the expected result.

https://wikia-inc.atlassian.net/browse/SUS-129
https://wikia-inc.atlassian.net/browse/SUS-128

@Wikia/sustaining-team 

TODO:
- [x] Fix the tests
- [x] Add tests with a little more substance one or more classes that depend on the `ExactTarget_*` classes


extensions/wikia/ExactTargetUpdates/tests/ExactTargetApiTest.php
```diff
@@ -60,30 +60,6 @@ function testPrepareSoapVarsShouldReturnSoapVarsArray() {
...
-		$oReflection = new ReflectionClass($mockApiSubscriber);
-		$oReflectionProperty = $oReflection->getProperty('helper');
-		$oReflectionProperty->setAccessible(true);
-		$oReflectionProperty->setValue($mockApiSubscriber, new Wikia\ExactTarget\ExactTargetApiHelper());
-
-		// Run tested method
-		$mockApiSubscriber->deleteRequest($aApiParams);
-	}
-
-	/**
```
kvas-damian, line 86: This test was in fact testing `prepareSubscriberObjects` method of `\Wikia\ExactTarget\ExactTargetApiHelper`. What do you think about adding unit tests for that method?

kvas-damian:
@drsnyder I believe you used some automated tool to extract those classes. It's not possible for a human to catch errors there ;)

extensions/wikia/ExactTargetUpdates/tests/ExactTargetApiTest.php
drsnyder, line 86: Using a mock of the class you want to test is problematic. Its not a good practice and usually indicates there is an issue with the design.

I think it would be better to test `prepareSubscriberObjects` directly. I'll look at implementing that.

drsnyder:
Yeah, that's too much to do by hand. It went something like this

``` perl
while ($string =~ /(?:^|\s*)(class ExactTarget_\w+ {[^}]*}$)/mpg) {
 # take out the name of the class
 # write the class file with $string & close
}
```

extensions/wikia/ExactTargetUpdates/tests/ExactTargetApiTest.php
kvas-damian, line 86: Direct test for `prepareSubscriberObjects` is exactly what tried to say ;)

PR #9636, opened 2016-02-03
SUS-74 `UpdateThumbnailTask::retryThumbUpload()` was throwing exceptions

ghost:
# SUS-74
## Synopsis
- https://wikia-inc.atlassian.net/browse/SUS-74
- and the exception was `WikiaForeignDBFile: write operations are not supported`
- Kibana: https://kibana.wikia-inc.com/index.html#/dashboard/elasticsearch/SUS-74
## Notes

When it comes to the `ForeignDBRepo` and `ForeignDBViaLBRepo` classes, write operations are not supported _by design_. Every write operation ends up with an exception. This is correct behaviour which implies that _usage_ is incorrect, not the code.

```
function store( $srcPath, $dstZone, $dstRel, $flags = 0 ) {
        throw new MWException( get_class($this) . ': write operations are not supported' );
}
function publish( $srcPath, $dstRel, $archiveRel, $flags = 0 ) {
        throw new MWException( get_class($this) . ': write operations are not supported' );
}
function deleteBatch( $sourceDestPairs ) {
        throw new MWException( get_class($this) . ': write operations are not supported' );
}
```

Similar situation is with `ForeignDBFile` and the abstract `File` classes (call `readOnlyError`).

The only classes supporting `publish` are `LocalFile` and `FileRepo`.

Going up the backtrace, an object of a wrong class is passed to `VideoFileUploader::resetThumbnail()` which then calls the `publish()` method.

Going up again, we have similar situation in `VideoHandlerHelper::resetVideoThumb()`.

Finally, going up, the instance is created in `UpdateThumbnailTask::retryThumbUpload()` with `WikiaFileHelper::getVideoFileFromTitle( $title );`.

Going up, `WikiaFileHelper::getVideoFileFromTitle()` creates this instance calling `WikiaFileHelper::getFileFromTitle()` which does it calling `wfFindFile()` which calls `RepoGroup::singleton()->findFile()`. The last two are core methods which I assume to work fine.

Taking into account, that the only classes supporting `publish` are `LocalFile` and `FileRepo`, I am getting to a conclusion that `WikiaFileHelper::getVideoFileFromTitle()` is expected to return an instance of either `LocalFile` or `FileRepo`. In other words, the entire task has to be run in the context of one particular wiki for which the file in topic is local rather than foreign.

**Question:** when would `wfFindFile()` return something else than `LocalFile` and `FileRepo`?

The answer to the above question is pretty much the fix here.

So the file finder first tries to find the file in `$wgLocalFileRepo`. On failure, it iterates through `$wgForeignFileRepos`. Let's check which repos do those variables point to.

```
michal@task-s2:~$ SERVER_ID=663 php /usr/wikia/slot1/current/src/maintenance/eval.php 
> var_dump($wgLocalFileRepo, $wgForeignFileRepos);
array(11) {
  ["class"]=>
  string(14) "WikiaLocalRepo"
...
  ["class"]=>
    string(23) "WikiaForeignDBViaLBRepo"
```

Per

```
* This task is run when a video is uploaded but the provider does not have a
* thumbnail for us to use. This gets triggered the first time a thumbnail
* cannot be found, and is queued up again at longer intervals until we either
* get a thumbnail from the provider, or exhaust all of our attempts.
```

Answer: the task is supposed to give up at some point and dealing with foreign repos is one of those situations.
## Done
- remove unused parameter from `processResponse()` methods (spotted a PHP strict standards whinings in Kibana)
- log exceptions along with the context
## ToDo
- [ ] verify when is the task scheduled
- [ ] idea: once foreign repo is detected, the entire task should be re-run with SERVER_ID for which the foreign repo is local (another task scheduled, and the current task may terminate with success)
## Code Review

@Wikia/sustaining-team

### DIFF
--- a/extensions/wikia/VideoHandlers/apiwrappers/ApiWrapper.class.php
+++ b/extensions/wikia/VideoHandlers/apiwrappers/ApiWrapper.class.php
@@ -1,5 +1,4 @@
 <?php
-
 abstract class ApiWrapper {
 
 	const RESPONSE_FORMAT_JSON = 0;
@@ -636,9 +635,10 @@ abstract class PseudoApiWrapper extends ApiWrapper {
 		// override me!
 	}
 
-	protected function processResponse( $response, $type ) {
+	protected function processResponse( $response ) {
 		// override me!
 	}
+
 }
 
 /**
--- a/extensions/wikia/VideoHandlers/VideoFileUploader.class.php
+++ b/extensions/wikia/VideoHandlers/VideoFileUploader.class.php
@@ -254,13 +254,19 @@ class VideoFileUploader {
 		// to fetch a thumbnail
 		try {
 			$upload = $this->uploadBestThumbnail( $thumbnailUrl, $delayIndex );
+
+			// Publish the thumbnail file (some filerepo classes do not support write operations)
+			$result = $file->publish( $upload->getTempPath(), File::DELETE_SOURCE );
 		} catch ( Exception $e ) {
+			WikiaLogger::instance()->error( __METHOD__, [
+				'thumbnailUrl' => $thumbnailUrl,
+				'delayIndex' => $delayIndex,
+				'file_obj' => $file,
+				'exception' => $e
+			]);
 			wfProfileOut(__METHOD__);
 			return Status::newFatal($e->getMessage());
 		}
-
-		// Publish the thumbnail file
-		$result = $file->publish( $upload->getTempPath(), File::DELETE_SOURCE );
 
 		wfProfileOut(__METHOD__);
 		return $result;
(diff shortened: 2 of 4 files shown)


kvas-damian:
LGTM

extensions/wikia/VideoHandlers/apiwrappers/ApiWrapper.class.php
```diff
@@ -636,9 +635,10 @@ protected function getInterfaceObjectFromType( $type ) {
 		// override me!
 	}
 
-	protected function processResponse( $response, $type ) {
+	protected function processResponse( $response ) {
```
macbre, line 638: Minor one: let's remove the method body - this will "force" all inheriting classes to define this method - PHP Storm will even help you here :)

PR #9637, opened 2016-02-03
PLATFORM-1836 Disable Special:PasswordReset

RafalWilinski:
https://wikia-inc.atlassian.net/browse/PLATFORM-1836

With @jcellary we decided to divide this task into 2 stages:
1. Disable Special:PasswordReset page and see what the response is. This change is easy to revert because of the fact that only one line gets affected.
1. After period of time we verify that this page is really unused. If it is, we run this script removing all translations: https://gist.github.com/RafalWilinski/06e08cf92f2f8ae2dbb7 and SpecialPasswordReset.php file

/cc @frankfarmer @lgarczewski


jcellary:
@RafalWilinski are you sure this will only disable the display of SpecialPasswordReset.php
page? Searching the code revealed that this variable is also used in "SpecialUserlogin.php". We don't want to disable password reset as a whole. We just want to remove the special page, because we have an alternative flow for resetting password. I believe your change will also disable the link in "SpecialUserlogin.php" which I don't think is desired and will not disable the ability to access the PasswordReset special page.

RafalWilinski:
Page has been disabled in a "better way".

jcellary:
Looks much better - LGTM :)

PR #9638, opened 2016-02-03
SUS-66: Video upload: title exists but article is null

macbre:
[SUS-66](https://wikia-inc.atlassian.net/browse/SUS-66)

Let's change `$article = Article::newFromID( $articleId );` to constructing `Article` instance from `Title object`.

My bet here is the slave lag - `Article::newFromID` uses slave node to fetch page information by id. By passing the Title instance to the `Article` constructor we can skip the database query (that can likely hit the slave that misses the row for newly created article - causing the fatal that was replaced with an exception throw).

@Wikia/sustaining-team


drsnyder:
LGTM

ghost:
LGTM

PR #9647, opened 2016-02-04
SOC-1840

jsutterfield:
This PR is a copy of https://github.com/Wikia/app/pull/9560 which targets dev as opposed to SOC-1820.


armonr:
My response is still :shipit:

PR #9648, opened 2016-02-04
CE-3301 Modify RecentChanges API to return a URL to a user avatar

adamkarminski:
@Wikia/community-engineering

### DIFF
--- a/includes/api/ApiQueryRecentChanges.php
+++ b/includes/api/ApiQueryRecentChanges.php
@@ -411,7 +411,7 @@ class ApiQueryRecentChanges extends ApiQueryGeneratorBase {
 			}
 
 			if ( $this->fld_useravatar ) {
-				$vals['useravatar'] = AvatarService::getAvatarUrl( $vals['user'], AvatarService::AVATAR_SIZE_MEDIUM );
+				$vals['useravatar'] = AvatarService::getAvatarUrl( $row->rc_user_text, AvatarService::AVATAR_SIZE_MEDIUM );
 			}
 
 			if ( !$row->rc_user ) {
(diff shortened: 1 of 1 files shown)


includes/api/ApiQueryRecentChanges.php
```diff
@@ -408,6 +410,10 @@ public function extractRowInfo( $row ) {
 				$vals['userid'] = $row->rc_user;
 			}
 
+			if ( $this->fld_useravatar ) {
+				$vals['useravatar'] = AvatarService::getAvatarUrl( $vals['user'], AvatarService::AVATAR_SIZE_MEDIUM );
```
Grunny: If someone sets the props as `userid|useravatar`, this code would be executed but `$vals['user']` would be undefined as the `user` prop wasn't set, so it's probably better to use `$row->rc_user_text` here.

adamkarminski: Done!

lukaszkonieczny:
LGTM

includes/api/ApiQueryRecentChanges.php
```diff
@@ -639,6 +646,7 @@ public function getParamDescription() {
 				'Include additional pieces of information',
 				' user           - Adds the user responsible for the edit and tags if they are an IP',
 				' userid         - Adds the user id responsible for the edit',
+				' useravatar     - Adds an avatar of a user responsible for the edit. It requires the user or the userid param to be present.',
```
kamilkoterba, line 649: :+1:

kamilkoterba:
LGTM

azarek:
LGTM

PR #9652, opened 2016-02-04
SUS-80: filter out "not fully uploaded" videos from getVideoList response

macbre:
[SUS-80](https://wikia-inc.atlassian.net/browse/SUS-80)

Because of the way videos upload was fixed before SUS-66 was applied (#9638), rows to `page` and `video_info` tables were added, but the actual video was never uploaded (i.e. `image` table row was missing).

@Wikia/sustaining-team


ghost:
:+1:

PR #9656, opened 2016-02-04
Fix <body> tag in Oasis

wladekb:
Fix for lack of space between `<body>` attributes eg.

```
<body class="mediawiki ltr ns-0 ns-subject page-Muppet_Wiki  oasis-one-column oasis-breakpoints wikinav2 notifications skin-oasis mainpage  user-logged background-fixed background-not-tiled background-dynamic"itemscope itemtype="http://schema.org/WebPage">
```

/cc @macbre


macbre:
One space to fix'em all :) LGTM

PR #9659, opened 2016-02-05
DAT-3624 Fix for broken references in infobox image captions

sebastianmarzjan:
Since cite stores reference list in global state, do not parse template args when they are not needed.
Tests added in: https://github.com/Wikia/api-tests/pull/168


idradm:
LGTM, as long this travis build fail is irrelevant

PR #9661, opened 2016-02-05
Dat 3693 extract jsonformat tests to separate suite

idradm:


extensions/wikia/JsonFormat/tests/DivContainingHeadersVisitorTest.php
```diff
@@ -10,6 +10,7 @@ public function setUp() {
 	}
 
 	/**
+	 * @group ExternalDependencyIntegration
```
sebastianmarzjan: ContractualResponsibilitiesValidation ?

tests/Makefile
```diff
@@ -41,6 +41,11 @@ phpunit-externals: prepare
 	${CMD_PHPUNIT} \
 		--group ExternalIntegration
 
+# runs tests for externally used parts
+phpunit-dependents: prepare
```
sebastianmarzjan: phpunit-contractual-responsibilities ?

sebastianmarzjan:
Is it possible to add to the asserts some more information what a failure of a given test case means?

sebastianmarzjan:
Plus, I'm thinking about a way to actually solve the instability of the test - at this stage, it's just changing the accounting

sebastianmarzjan:
Let's :ship: and monitor if the issue was actually caused by test coupling

idradm:
Merging to dev then

PR #9663, opened 2016-02-05
SUS-81 - getVideoList: rely on CDN cache only

macbre:
[SUS-81](https://wikia-inc.atlassian.net/browse/SUS-81)

`VideoHandlerController::getVideoList` is now cached on both memcache and CDN (for 24 h both). It returns the list of recently uploaded videos on a give wiki. And it should ideally be purged on each video upload.

Due to the way memcache keys are formatted (they include a wide range of different parameters) let's rely on CDN cache solely and use surrogate keys to purge all `VideoHandlerController::getVideoList` responses for a given wiki in one purge request.

This pull request introduces [purging by surrogate keys from MediaWiki](https://github.com/Wikia/app/commit/c2ad3541e421c53155f9e88b933a57cfca1be8ce) in `CeleryPurge` class and `EdgeCachePurger`  that wraps calls to `celery_workers.purger.purge` Python celery worker.
- trigger hooks in `VideoInfo` class when making changes to `video_info` table
- when these hooks are run purge the surrogate key via `CeleryPurge::purgeBySurrogateKey`

Celery worker will write the following to `celeryd` log:

```
[2016-02-05 14:01:27,897: WARNING/Worker-1] Purging service 5NzYXXXXXHUzWP via surrogate key dev-macbre-plpoznan-VideoHandlerController-getVideoList done with HTTP 200 in 0.007651
```

Responses will be tagged with `X-Surrogate-Key` and `Surrogate-Key` (and prefixed with `$wgCachePrefix`):

```
X-Surrogate-Key: dev-macbre-plnordycka-VideoHandlerController-getVideoList
```

@Wikia/sustaining-team / @wladekb (as it touches MediaWiki core)


includes/cache/wikia/CeleryPurge.class.php
```diff
@@ -71,6 +71,30 @@ static function onRestInPeace() {
...
+		$caller = self::getPurgeCaller();
+		wfDebug( "Purging backtrace: " . wfGetAllCallers( false ) . "\n" );
+
+		WikiaLogger::instance()->info( 'varnish.purge', [
+			'key' => $key,
+			'service' => $service
+		] );
+
+		( new AsyncCeleryTask() )
+			->taskType( 'celery_workers.purger.purge' )
```
drsnyder: This should probably be a task in `\Wikia\Tasks\Tasks`. Something like

``` php
( new EdgeCachePurger( [ $key ], $service ) )->queue();
```

This would encapsulate the constants and the sequence of method calls and also make this operation more `grep` friendly.

drsnyder: Oh. As I look at this more I'm not sure what it's doing. What is the difference between this and the [purge](https://github.com/Wikia/app/pull/9663/files#diff-a3b45c6089cdc13c1ebda814abf04577R221) above?

extensions/wikia/VideoHandlers/VideoHandlers.setup.php
```diff
@@ -123,6 +123,9 @@
 	$wgHooks['ParserFirstCallInit'][] = 'VideoHandlerHooks::initParserHook';
 }
 
+$wgHooks['VideoInfoSaveToCache'][] = 'VideoHandlerHooks::onClearVideoCache';
+$wgHooks['VideoInfoInvalidateCache'][] = 'VideoHandlerHooks::onClearVideoCache';
```
kvas-damian: `onClearVideoCache` looks strange here. I was under impression that we're using `on` prefix when we related to hook name. Maybe we could use `clearVideoCache` here?

includes/cache/wikia/CeleryPurge.class.php
```diff
@@ -71,6 +71,30 @@ static function onRestInPeace() {
...
+			->setArgs( [], [ $key ], $service )
+			->setPriority( PurgeQueue::NAME )
```
kvas-damian: Why do we set queue name as a priority?

extensions/wikia/VideoHandlers/VideoHandlers.setup.php
macbre: e06964b

includes/cache/wikia/CeleryPurge.class.php
macbre: Good call on introducing `EdgeCachePurger` task.

As for the difference between `purge` and `purgeBySurrogateKey` - the first one accepts URLs only and simply puts then in appropriate "buckets" for purge a proper Fastly service.

macbre: 88c5dac

PR #9667, opened 2016-02-08
INT-328 New Upload Image / Video dialogs.

perjg:
Contains commits from INT-322 and INT-328.
See https://github.com/Wikia/app/pull/9589


extensions/VisualEditor/wikia/modules/ve/init/ve.init.wikia.Optimizely.js
```diff
@@ -0,0 +1,32 @@
...
+
+	return 0;
+};
+
+/**
+ * Hook for activating experiment on VE load
+ */
+mw.hook( 've.activate' ).add( function () {
+	window.optimizely.push(['activate', ve.init.wikia.getToolbarABTestId()]);
+});
```
inez: Please format correctly.

extensions/VisualEditor/wikia/modules/ve/ui/styles/images/icons/image.svg
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
inez: I don't think optimizer was used for this one.

extensions/VisualEditor/wikia/modules/ve/ui/styles/ve.ui.Icons-vector.scss
```diff
@@ -391,6 +391,12 @@
 .oo-ui-icon-media {
 	@include ve-icon(media, wikia);
 }
+.oo-ui-icon-image {
+	@include ve-icon(image, wikia);
+}
+.oo-ui-icon-video {
```
inez, line 397: Could you explain a little bit why there is video and video2?

extensions/VisualEditor/wikia/modules/ve/ui/styles/images/icons/image.svg
perjg: I used the optimizer with the same "conservative" settings for both this and video2.svg.
I think this icon is just more complex than the video icon.

inez: I don't think we need ids, etc., just look at the different SVGs.

extensions/VisualEditor/wikia/modules/ve/ui/styles/ve.ui.Icons-vector.scss
perjg, line 397: There was already a file video.svg, so I named the new file video2.svg

I don't know what the original video.svg icon is used for (it is not referenced in this scss file), it may be referenced elsewhere, so I didn't want to remove it and risk breaking something.

inez, line 397: If it's not being used it would be good to just remove it - otherwise this break pretty nice relation of class name and file name, e.g. "oo-ui-icon-image" => "image"

extensions/VisualEditor/wikia/modules/ve/init/ve.init.wikia.Optimizely.js
perjg: I guess you are referring to missing spaces after brackets, I've fixed that.

This formatting issue is not caught by the linter that runs after push.
Is there a way to run the linter locally before committing?

extensions/VisualEditor/wikia/modules/ve/init/ve.init.wikia.Optimizely.js
```diff
@@ -0,0 +1,32 @@
...
+
+	return 0;
+};
+
+/**
+ * Hook for activating experiment on VE load
+ */
+mw.hook( 've.activate' ).add( function () {
+	window.optimizely.push( ['activate', ve.init.wikia.getToolbarABTestId()] );
+});
```
inez: space missing

extensions/VisualEditor/wikia/modules/ve/init/ve.init.wikia.Optimizely.js
```diff
@@ -0,0 +1,32 @@
...
+	}
+
+	return 0;
+};
+
+/**
+ * Hook for activating experiment on VE load
+ */
+mw.hook( 've.activate' ).add( function () {
+	window.optimizely.push( ['activate', ve.init.wikia.getToolbarABTestId()] );
```
inez: spaces missing

extensions/VisualEditor/wikia/modules/ve/init/ve.init.wikia.Optimizely.js
inez: I'm sorry that maybe linter is not catching it up for you automatically right now - I will look into it soon - perhaps this week.
If you want to run linter locally just take a look at `.travis.yml` and follow the same steps.

extensions/VisualEditor/wikia/modules/ve/ui/styles/ve.ui.Icons-vector.scss
perjg, line 397: I've searched the code, and I am pretty sure it is not used. I suppose there is a chance it is used with a jquery.css call somewhere outside scss files, but I think it is unlikely.

I'll remove it.

inez, line 397: Thank you.

perjg:
@inez Ok to merge?

extensions/VisualEditor/wikia/modules/ve/ui/dialogs/ve.ui.WikiaVideoInsertDialog.js
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * @inheritdoc
+ */
+ve.ui.WikiaVideoInsertDialog.prototype.initialize = function () {
+	// Parent method
+	ve.ui.WikiaVideoInsertDialog.super.prototype.initialize.call( this );
+
+	this.pages.removePages( [ this.mainPage ] );
+	this.query.upload.uploadButton.toggle();
```
inez: aren't there getters for upload and uploadButton?

extensions/VisualEditor/wikia/modules/ve/init/ve.init.wikia.Optimizely.js
```diff
@@ -0,0 +1,32 @@
...
+};
+
+/**
+ * @returns {number} The A/B test variant number for the Toolbar A/B test
+ */
+ve.init.wikia.getToolbarABTestVariantNumber = function () {
+	var optimizely = window.optimizely,
+		optimizelyId = ve.init.wikia.getToolbarABTestId();
+
+	if ( optimizely && optimizely.variationMap && optimizely.variationMap.hasOwnProperty( optimizelyId ) ) {
```
inez, line 42: Not super important but I would consider adding method `getVariation` or so directly in OptimizelyBlocking.js - so we likely would avoid code duplication in the future.

inez:
Overall looks good to me - but please make sure to address remaining comment (about using getters) and write more meaningful commit message - also we try to have the prefix match exactly the ticket id - for instance: 'INT-328' instead of 'Int 328'.

perjg:
@inez "Int 328" is just the title of the PR, all the commits are prefixed with "INT-328". I've renames the PR for clarity.

I'll look into using getters like you suggest.

extensions/VisualEditor/wikia/modules/ve/ui/dialogs/ve.ui.WikiaVideoInsertDialog.js
perjg: upload has a getter, but I can't find any for uploadButton in ve.ui.WikiaUploadWidget.

perjg:
Test failure says "Exception: Can't connect to phalanx service on http://phalanx.service.consul:4666"
Is there a way to re-run the tests?

inez:
@perjg Yes - there is a way to re-run the tests - just go to test results (by clicking "Details") and look for "Rebuild" button/link on the left side navbar.
It's possible that tests are failing because your branch is pretty old - let me try merging dev into it.

inez:
Ok - I merged dev so test are running again - and there were no conflicts while merging dev - so it should not negatively affect your changes in any way.

perjg:
@inez There was a spec change in the issue, so I had to make an adjustment to how the experiment is activated:

We have four separate experiments with different engagement:
Japanese: 100 %, English: 10 %, Spanish: 25 % and German: 25 %.

All experiments are set up with manual activation in Optimizely, and when running on production, all four will be activated.
(Each experiment has mutually exclusive url filtering, so only one will be activated at a time since optimizely will only activate experiments with matching targetting conditions).

In order to simplify the code, I switched to using variant Javascript for all of the experiments.
Variant #1 (new toolbar) has this code:
var toolbarABTestVariantNumber = 1;

Details here:
https://help.optimizely.com/hc/en-us/articles/200040225-Activation-Mode-Activating-an-experiment-dynamically-after-a-page-has-loaded#change_mode

extensions/VisualEditor/wikia/modules/ve/init/ve.init.wikia.Optimizely.js
```diff
@@ -0,0 +1,29 @@
...
+/*!
+ * Utility functions for running Optimizely A/B tests with VisualEditor
+ */
+
+/**
+ * @returns {number} The A/B test variant number for the Toolbar A/B test
+ */
+ve.init.wikia.getToolbarABTestVariantNumber = function () {
+	// toolbarABTestVariantNumber is set for variant #1 for all Optimizely toolbar experiments
+	if ( typeof ( toolbarABTestVariantNumber ) === 'undefined' ) {
```
inez: So is it now that different variants now are setting different value of toolbarABTestVariantNumber?

extensions/VisualEditor/wikia/modules/ve/init/ve.init.wikia.Optimizely.js
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Hook for activating experiment on VE load
+ */
+mw.hook( 've.activate' ).add( function () {
+	// Production tests are split by language. This is done because each language has a separate engagement percentage
+	var optimizelyIds = window.wgDevelEnvironment ?
+		[ '4721410313' ] : [ '4701112678', '5003080344', '5003533755', '5013830116' ];
+
+	optimizelyIds.forEach( function (element) {
+		window.optimizely.push( [ 'activate', element ] );
```
inez: So for all languages all experiments will be activated? And GA + Optimizely integration is somehow smart enough to figure out which experiment is should report in the custom values? (That's what I'm assuming it's going on because I don't see you reporting coversion within Optimizely - so it must be within GA)

extensions/VisualEditor/wikia/modules/ve/init/ve.init.wikia.Optimizely.js
perjg: There are still only two UI variants, so each of the language Optimizely tests set toolbarABTestVariantNumber to 1 for Variant #1, which triggers the new toolbar to show.

extensions/VisualEditor/wikia/modules/ve/init/ve.init.wikia.Optimizely.js
perjg: optimizely.push:
According to what I see in the Optimizely documentation for optimizely.push: "will activate any running manual experiments whose targeting conditions are satisfied by the current visitor on this page"

So unless I am misunderstanding, this should be ok. It is of course possible to check for window.location.href and then only call optimizely.push for the matching one, but I thought I could avoid that.

There is no change in the issue for GA reporting, so reporting is the same as before, e.g. dialog-image-insert, dialog-video-insert.

perjg:
Latest commit moves the initialization of the A/B test to before the first call to getToolbarABTestVariantNumber. ve.activate hook runs after toolbar array has been built, so it was activated too late.

perjg:
@inez Looks like I may have misunderstood a detail about manual activation mode:
https://help.optimizely.com/hc/en-us/articles/200040225-Activation-Mode-Activating-an-experiment-dynamically-after-a-page-has-loaded#change_mode

This call:
window.optimizely.push(["activate"]);

Will activate only experiments with a matching url filter (but may also activate other non-toolbar related A/B tests, so we don't want to do that)

No such guarantee seems to be given for 
window.optimizely.push(["activate", EXPERIMENT_ID]);

So to be safe, it is probably best to activate only the experiment that matches the correct URL filter.

I've pushed another commit to change to this behavior. The code is a bit more verbose, but it is guaranteed to activate only the correct A/B test.

Note: The entry for sandbox-japan is just for testing, I will remove it before I merge.

extensions/VisualEditor/wikia/modules/ve/init/ve.init.wikia.Optimizely.js
```diff
@@ -0,0 +1,72 @@
...
+				id: 5003080344
+			}
+		];
+
+	if ( window.wgDevelEnvironment ) {
+		return 4721410313;
+	} else {
+		for ( i = 0; i < data.length; i++ ) {
+			if ( href.match( data[i].regex ) ) {
+				return data[i].id;
```
inez: If you want to make a decision bases on the language - use global variable `wgContentLanguage` or `mw.config.get('ContentLanguage')`. If you want to make a decision based on the wikiname - use global variable `wgDBname`.
Also - I think the point here is to activate one test? If that's correct then you can `break;` inside the `if` statement.

extensions/VisualEditor/wikia/modules/ve/ve.track.js
```diff
@@ -175,6 +176,18 @@ require( ['wikia.tracker'], function ( tracker ) {
 			return;
 		}
 
+		if ( ve.init.wikia.getToolbarABTestVariantNumber() === 1 && data.label === 'edit-page-ready' ) {
```
inez, line 179: Why do we need to send extra event to GA in case of the experiment? From what I recall information about the experiment is injected into custom variables anyways?
1. If the information about the experiment is injected into custom variable then there is no need to send extra event.
2. If the information about the experiment is NOT injected into custom variable then we should "duplicate" not only this event but many others (to find out if moving video/image to different location is positive).

I hope it's 1.

extensions/VisualEditor/wikia/modules/ve/ui/dialogs/ve.ui.WikiaVideoInsertDialog.js
```diff
@@ -0,0 +1,48 @@
...
+ve.ui.WikiaVideoInsertDialog.prototype.initialize = function () {
+	var uploadWidget;
+
+	// Parent method
+	ve.ui.WikiaVideoInsertDialog.super.prototype.initialize.call( this );
+
+	this.pages.removePages( [ this.mainPage ] );
+
+	uploadWidget = this.query.getUpload();
+	uploadWidget.uploadButton.toggle();
```
inez: Getter should be used for "uploadButton" - if it does not exist then it should be created (I was already commenting on it once - maybe in the other PR).

extensions/VisualEditor/wikia/modules/ve/ve.track.js
perjg, line 179: There is a custom variable for tracking what happens in the A/B test, see trackingLabel further up in this file. This takes care of everything asked for in the issues except for this requirement:

editor-ve // impression // edit-page-ready - already being tracked in GA but we need a way to bucket this data to A or B groups

This is the reason I send the extra edit-page-ready-toolbartest event in addition to edit-page-ready. Initially, I modified the edit-page-ready event, but in the original PR you advised me to send both.

perjg:
@inez Latest Commit:
- Use wgContentLanguage to determine which A/B test to activate
- Added getter for uploadDialog to ve.ui.WikiaUploadWidget.js

See also my comment on edit-page-ready-toolbartest above.

extensions/VisualEditor/wikia/modules/ve/ve.track.js
inez, line 179: You didn't really modify the original event - you were just sending another one instead of it.
I'm just making sure that you will be able to track for instance: if we are getting more saves in case of variant 1 vs. default?

perjg, line 179: Yes, that case is handled by the trackingLabel.

inez:
This LGTM. ( @perjg and I talked about figuring out later on what is "the best" approach to tracing in case of experiment but it is going to be handled in another PR in the future (if at all).)

PR #9674, opened 2016-02-08
(CE-3331) Fix block info in the userinfo API module

Grunny:
The blockedby value was returning false for local blocks as User::blockedBy()
returns the user name for local blocks, rather than the ID. So, use the Block
object instead.


kamilkoterba:
LGTM

PR #9680, opened 2016-02-09
SUS-150 Set the "Status" field of the subscribed object based on the users email preference

drsnyder:
We aren't currently respecting our user's `unsubscribed` preference in the ExactTarget "subscriber" entity. This PR addresses that by updating the `ExactTarget_Subscriber->Status` based on the user's subscription preference.

https://wikia-inc.atlassian.net/browse/SUS-150

This is a work in progress. `ExactTargetCreateUserTask` needs to be updated to support dependency injection and better testing.

@Wikia/sustaining-team 

TODO:
- [x] improve DI in `ExactTargetCreateUserTask`
- [x] change the methods to support passing down the user `unsubscribed` preference
- [ ] add a method for changing the subscription status when the users preferences are changed
- [ ] test `onUserSaveSettings`
- [ ] test `updateSubscriptionStatus`


macbre:
LGTM

drsnyder:
My only hesitation with merging this is the last commit. I hastily wired it in to capture what I had in my head before I ran out of time for the week. I think my preference would be to revert that commit and merge. That seems safer at this point instead of trying to rush the changes.

PR #9682, opened 2016-02-10
SUS-92 | Fix logging to PhalanxStats

kvas-damian:
https://wikia-inc.atlassian.net/browse/SUS-92

During investigation we found that application checks blocks in two different scenarios:
- prevent user from performing action - edit, move, create user, etc. Those checks should create new row in PhalanxStats
- check if user is blocked to modify UI - hide UI controls, show user is blocked flag, etc. Checks of this kind doesn't mean that action was stopped or something was protected by this block. They're only informational. It means that those actions shouldn't trigger adding new record to PhalanxStats

In order to achieve that goal, `$shouldLogBlockInStats` was introduced as a second parameter for `User:isBlock` method.

Most important changes happen in `PhalanxUserBlock` and `PhalanxModel`. Other changes are just passing new flag to `isBlocked` method.

/cc @Wikia/sustaining-team

### DIFF
--- a/extensions/CheckUser/CheckUser_body.php
+++ b/extensions/CheckUser/CheckUser_body.php
@@ -1035,7 +1035,9 @@ class CheckUser extends SpecialPage {
 				$s .= '</li>';
 			}
 			$s .= "</ul></div>\n";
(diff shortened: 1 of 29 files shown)


macbre:
LGTM

And :+1:  for adding comments in Message Wall code

PR #9684, opened 2016-02-10
SUS-108: check read-only state of ACTIVE_CLUSTER before performing any DB-related actions

macbre:
[SUS-108](https://wikia-inc.atlassian.net/browse/SUS-108)

This pull request backports https://github.com/wikimedia/mediawiki/commit/52010e6d21d8bdefa1c89fbc9421850185cc5011 from MW 1.27.

Let's pass reao-only state to `LoadBalancer` and `Database` instances, so that we can easily check if another cluster is in read-only mode. This will be used in CreateWiki code (that runs on c1) to check the read-only state of a different cluster (currently c7).
- pass `$wgReadOnly` (global read-only mode switch) to `LBFactory` constructor
- pass per DB section read-only (controlled by `$wgLBFactoryConf['readOnlyBySection']`) to `LoadBalancer` instance and when creating a database connection (via `Database::setLBInfo` )

Thanks to this change we can finally perform the following checks:

``` php
# set in the DB config
$wgLBFactoryConf['readOnlyBySection']['c7'] = 'We apologise, this wiki is temporarily in read-only state';

# in eval.php
> var_dump( wfGetLB('wikicities_c7')->getReadOnlyReason() )
string(57) "We apologise, this wiki is temporarily in read-only state"

> var_dump( wfGetDB(DB_MASTER, [], 'wikicities_c7')->getLBInfo( 'readOnlyReason' ) )
string(57) "We apologise, this wiki is temporarily in read-only state"

> var_dump( wfGetLB()->getReadOnlyReason() );
bool(false)
```

@Wikia/sustaining-team / @wladekb (touches MW core, namely LoadBalancer)

### DIFF
--- a/extensions/wikia/CreateNewWiki/CreateWiki.php
+++ b/extensions/wikia/CreateNewWiki/CreateWiki.php
@@ -227,6 +227,13 @@ class CreateWiki {
 		$this->mClusterDB = ( self::ACTIVE_CLUSTER ) ? "wikicities_" . self::ACTIVE_CLUSTER : "wikicities";
 		$this->mNewWiki->dbw = wfGetDB( DB_MASTER, array(), $this->mClusterDB ); // database handler, old $dbwTarget
 
+		// SUS-108: check read-only state of ACTIVE_CLUSTER before performing any DB-related actions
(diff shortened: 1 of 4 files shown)


kvas-damian:
LGTM

wladekb:
lgtm

PR #9686, opened 2016-02-10
MAIN-6395: check the cache first, dude!

wladekb:
We're being overwhelmed by queries from Wall, checking the cache first could help!

https://wikia-inc.atlassian.net/browse/MAIN-6395

/cc @macbre


macbre:
Let's try it!

PR #9689, opened 2016-02-10
RECIRC 41: AB Test an in-content recirculation widget

pauloslund:
This test will initially be run for 1% of english visitors


extensions/wikia/Recirculation/Recirculation.setup.php
```diff
@@ -21,3 +21,7 @@
 
 // i18n
 $wgExtensionMessagesFiles['Recirculation'] = __DIR__ . '/Recirculation.i18n.php';
+
+JSMessages::registerPackage('Recirculation', array(
```
nandy-andy: nitpicking: go with short array notation! :japanese_ogre:  `[]`

extensions/wikia/Recirculation/js/gruntfile.js
```diff
@@ -0,0 +1,24 @@
+module.exports = function (grunt) {
```
nandy-andy: Do you still need this gruntfile and `.gitignore` above after going with `wikia.loader` for templates?

extensions/wikia/Recirculation/js/incontent.js
```diff
@@ -0,0 +1,187 @@
...
+	function injectInContentWidget($container) {
+		var $links = $container.find('a'),
+			sections = buildSections($container.find('h2')),
+			width,
+			firstSuitableSection,
+			topTitles;
+
+		// If this page doesn't have enough content (either links or sections) we
+		// don't want to show this widget
+		if ($links.length < 8 || sections.length < 3) {
```
nandy-andy: What do you think about moving those "magic" numbers to some variables which names describe their purpose for example: `minLinksNumber` and `minSectionsNumber`?

extensions/wikia/Recirculation/js/incontent.js
```diff
@@ -0,0 +1,187 @@
...
+	'jquery',
+	'wikia.window',
+	'wikia.abTest',
+	'wikia.nirvana',
+	'wikia.mustache',
+	'ext.wikia.recirculation.tracker',
+	'ext.wikia.recirculation.utils'
+], function ($, w, abTest, nirvana, Mustache, tracker, utils) {
+	// Currently only showing for English communities
+	if (w.wgContentLanguage !== 'en') { return; }
```
nandy-andy, line 13: What do you think about adding one more dependency in form of `wikia.log` and using logger with all those early returns. This way when somebody comes to you and asks why she doesn't see the widget on a page you'll be able to easily add `?log_level=debug&log_group=...` to the page URL and see in console what is the reason. If it's expected because it gets to one of those early returns or if it's a unhandled "exception". You can find a lot of examples how to use `wikia.log` module in `AdEngine` extension for example in [AdInContentPlayer.js](https://github.com/Wikia/app/blob/3503f5f49310c5cf7912c95fd61d42b42d5ea7c2/extensions/wikia/AdEngine/js/AdInContentPlayer.js#L48).

extensions/wikia/Recirculation/js/incontent.js
```diff
@@ -0,0 +1,187 @@
...
+				abstract: 0,
+				width: 50,
+				height: 50
+			},
+			callback: renderArticles(firstSuitableSection)
+		});
+	}
+
+	function renderArticles(section) {
+		var placeholderImage = w.wgExtensionsPath + '/wikia/Recirculation/images/placeholder.png';
```
nandy-andy, line 68: Could you consider changing this approach a little bit? There is [a practice in Wikia that we create CSS classes with images as background](https://github.com/Wikia/app/search?utf8=%E2%9C%93&q=%2F*+%24wgCdnStylePath+*%2F) and the SASS files are the only ones with the  file paths. Then in JS you just add or remove the class with placeholder image. I'd just remove the `placeholderImage` from here and set `item.thumbnail` to `null` if image does not exist. Then in the mustache template only if the `thumbnail` is _not_ `null` I'd put ``, otherwise I'd left `div` with a `placeholder` class which has the placeholder image defined in the style rule. Think about it :-) :art:

extensions/wikia/Recirculation/js/incontent.js
```diff
@@ -0,0 +1,187 @@
...
+	/**
+	 * Build the sections array
+	 *
+	 * DOM/layout querying: OK
+	 * DOM modification:    forbidden
+	 *
+	 * @param {jQuery} $headers headers dividing the article to sections
+	 * @returns {Section[]}
+	 */
+	function buildSections($headers) {
```
nandy-andy, line 102: This method [look familiar](https://github.com/Wikia/app/blob/6f82a77805c4eb66f198e43632ae1e40b81209a3/extensions/wikia/AdEngine/js/slot/inContent.js#L98), does it @gabrys? :japanese_ogre: I'm pinging Rychu since those methods are similar but not exactly the same. So, it's a kind of red flag in my head which I don't know how to remove right now. If the methods were the same I'd suggest creating for example `wikia.sectionsBuilder` module and re-use its method in those two places :stuck_out_tongue_winking_eye:

extensions/wikia/Recirculation/js/incontent.js
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	function validLink(element) {
+		// Not a link to current article
+		if (element.title === w.wgTitle) {
+			return false;
+		}
+
+		// This is a pretty heavy handed Regex, and it may only work for EN communities
+		// but the idea is to not display links to special pages
```
nandy-andy, line 151: Do you mean special pages as the one in `Special` namespace (for example: `Special:WikiActivity`) or all pages within other than regular (`NS_MAIN`) namespace?

extensions/wikia/Recirculation/js/package.json
```diff
@@ -0,0 +1,12 @@
+{
```
nandy-andy: I think you can remove this file after moving to `wikia.mustache`.

extensions/wikia/Recirculation/js/templates.mustache.js
```diff
@@ -0,0 +1,4 @@
+define( 'ext.wikia.recirculation.templates.mustache', [], function() { 'use strict'; return {
```
nandy-andy: I think you can remove this file after moving to wikia.mustache.

extensions/wikia/Recirculation/js/incontent.js
pauloslund, line 68: They only issue with this is that some of the thumbnails will be pngs and reveal the placeholder behind them. We would be able to a class if thumbnail is null, but I think that starts to be a lot of logic inside of a mustache template

extensions/wikia/Recirculation/js/incontent.js
pauloslund, line 151: All pages other than NS_MAIN

extensions/wikia/Recirculation/js/incontent.js
pauloslund, line 102: This is definitely a stripped down version of the buildSections method from AdEngine. Given this is being run as a very limited AB test initially, we do have time planned in the future to come back and "optimize" the code if the experiment proves to be successful and we continue down this path with this widget

extensions/wikia/Recirculation/js/utils.js
```diff
@@ -0,0 +1,40 @@
...
+	'wikia.loader',
+	'wikia.cache'
+], function (loader, cache) {
+	'use strict';
+
+	/**
+	 * Checks if template is cached in LocalStorage and if not loads it by using loader
+	 * @returns {$.Deferred}
+	 */
+	function loadTemplate(templateLocation) {
```
nandy-andy, line 12: Uhm, I think we could make something useful for all company here ;) How about moving this method, a little bit changed to `wikia.loader`? And then re-use it in `extensions/wikia/Recirculation/js/incontent.js` module this way:

```
if (abTest.inGroup('RECIRCULATION_INCONTENT', 'YES')) {
   loader
     .loadTemplate(
         'RecirculationAssets_',
         'extensions/wikia/Recirculation/templates/inContent.client.mustache'
     )
     .then(function(loadedTemplate) {
         template = loadedTemplate;
         injectInContentWidget($container);
     });
}
```

The method in `wikia.loader` will have this signature then:

```
function loadTemplate(cacheKey, templateLocation, cacheTtl)
```

where default `cacheTtl = 86400`.

And in future while touching some of [those pieces of code](https://github.com/Wikia/app/search?utf8=%E2%9C%93&q=function+loadTemplate%28) we could replace custom `loadTemplate()` calls there with `wikia.loader`'s `loadTemplate()` :gem: 

@macbre @lizlux @hakubo @RafLeszczynski and @kvas-damian what do you think about it?

extensions/wikia/Recirculation/js/incontent.js
nandy-andy, line 68: That's a fair point and as always a trade-off: is it better to keep the path in JS to avoid logic in template or add not that much complicated logic to the template and keep path in SCSS ;) It's fine to think about it with next iteration since you want to release it only to small amount of users now.

extensions/wikia/Recirculation/js/incontent.js
nandy-andy, line 102: Understandable. Thanks!

extensions/wikia/Recirculation/js/incontent.js
nandy-andy, line 151: Then it's fine for now and for English :+1: In future you might want to implement a logic using global `wgNamespaceNumber` but then you'll still have to find a way to know what number is a content namespace id.

nandy-andy:
LGTM as for code changes which are about to be released for 1% of users :see_no_evil: as long as in next iteration we'll rethink and possibly address some of the "bigger" comments ;-)

I'd love to see some unit tests but I understand it's a step "after experiment phase" :-)

PR #9691, opened 2016-02-11
ADEN-2837 Get RPFL targeting params based on their value

fraszczakszymon:
Instead of checking whether rubicon has responded (to all slots or by timeout) check the real value of provided targeting params. It's not affecting other bidders because they are returning params based on priceMap calculated in onResponse callback.


drets:
LGTM :+1:

jacekjursza:
LGTM

PR #9694, opened 2016-02-11
CE-3358 Add RevisionUpvotesApiController to get number of upvotes

kamilkoterba:
https://wikia-inc.atlassian.net/browse/CE-3342

Storage for upvotes and API to retrieve it.

for SQL review
`getCount`

```
SELECT count(*) AS upvotes_count
     FROM upvote_revisions
     WHERE wiki_id = '1' AND revision_id = '1';
```

`userUpvoted`

```
SELECT upvote_id
     FROM upvote_revisions
     WHERE wiki_id = '1' AND revision_id = '1' AND from_user = '1'  
     LIMIT 1;
```

@Wikia/community-engineering


Grunny:
LGTM.

includes/wikia/services/RevisionUpvotesService.class.php
```diff
@@ -0,0 +1,283 @@
...
+		$db = $this->getDatabaseForWrite();
+
+		$db->begin();
+
+		$query = sprintf(
+			'INSERT INTO `%s` SET `upvote_id` = %d, `from_user` = %d;',
+			self::UPVOTE_TABLE,
+			$upvoteId,
+			$fromUser
+		);
```
Grunny: As a best practice, this should be using `$db->insert( ... );` or FluentSql.

includes/wikia/services/RevisionUpvotesService.class.php
```diff
@@ -0,0 +1,283 @@
...
+		$db->query( $query );
+
+		$query = sprintf(
+			'INSERT INTO `%s` SET `user_id` = %d, `total` = 1, `new` = 1
+			ON DUPLICATE KEY UPDATE `total` = `total` + 1, `new` = `new` + 1, `notified` = false;',
+			self::UPVOTE_USERS_TABLE,
+			$userId
+		);
```
Grunny, line 281: Likewise, this should be using `$db->upsert( ... );` or FluentSql.

includes/wikia/api/RevisionUpvotesApiController.class.php
```diff
@@ -0,0 +1,110 @@
...
+	 * @requestParam int revisionId
+	 *
+	 * @throws MissingParameterApiException
+	 * @throws NotFoundException
+	 * @throws BadRequestApiException
+	 */
+	public function addUpvote() {
+		$request = $this->getRequest();
+
+		if ( !$request->wasPosted() ) {
```
Grunny: What do you think of for the two write methods, replacing this and the `validateUser` call with instead so you're using the available method for validating the post and token already?

``` php
if ( !$user->isLoggedIn() ) {
    throw new UnauthorizedException();
}
$this->checkWriteRequest();
```

includes/wikia/api/RevisionUpvotesApiController.class.php
```diff
@@ -0,0 +1,102 @@
...
+	/**
+	 * Add upvote for given revision made by current user
+	 *
+	 * @requestParam int revisionId
+	 *
+	 * @throws MissingParameterApiException
+	 * @throws NotFoundException
+	 * @throws BadRequestApiException
+	 */
+	public function addUpvote() {
```
Grunny, line 14: I think for adding and removing votes, it would be good to have an explicit response (even just `{"success": true}`) that makes it clear to a consumer of the API that the request succeeded, rather than just relying on the absence of an exception. Maybe in the case of adding an upvote, it might even be worth returning the inserted ID which is usable when removing the added upvote.

includes/wikia/services/RevisionUpvotesService.class.php
```diff
@@ -0,0 +1,290 @@
...
+						'upvoteId' => $row->upvote_id
+					];
+				}
+
+				$upvotes[$row->revision_id]['upvotes'][] = [
+					'from_user' => $row->from_user
+				];
+			} );
+
+		foreach( $upvotes as $revisionId => $upvote ) {
```
Grunny: Minor: space needed between the `foreach` and the opening parenthesis per coding conventions.

includes/wikia/services/RevisionUpvotesService.class.php
```diff
@@ -0,0 +1,290 @@
...
+						'upvoteId' => $row->upvote_id
+					];
+				}
+
+				$upvotes[$row->upvote_id]['upvotes'][] = [
+					'from_user' => $row->from_user
+				];
+			} );
+
+		foreach( $upvotes as $upvoteId => $upvote ) {
```
Grunny: Same here.

Grunny:
LGTM.

PR #9695, opened 2016-02-11
SUS-153: Phalanx button is prepared only for adding new filter

macbre:
[SUS-153](https://wikia-inc.atlassian.net/browse/SUS-153)

Introduce `phalanx-modify-block` i18n message that will be displayed instead of ``phalanx-add-block` when in "modify a Phalanx block" mode.

And replace deprecated `wfMsg` with `wfMessage` + `escaped()` in cd8d0cb

@Wikia/sustaining-team


kvas-damian:
LGTM

PR #9700, opened 2016-02-11
(LOC-255) Fix translations of empty article comment

Grunny:
Message was switched to use wikitext, but the translations were not updated.


lukaszkonieczny:
LGTM

PR #9702, opened 2016-02-12
PLATFORM-1836 Disable Special:PasswordReset for real

RafalWilinski:
https://wikia-inc.atlassian.net/browse/PLATFORM-1836

After disabling this special page it's time to remove it for real but let's wait till real end of the week.

/cc @jcellary.


RafalWilinski:
@jcellary I think it's good time to check it and give it a go. As far as I know you're not much experienced in MW so should I find someone else to review this?

RafalWilinski:
cc @macbre @wladekb

RafalWilinski:
Could someone verify if I can remove these also? https://github.com/Wikia/app/blob/dev/maintenance/language/messages.inc#L481-L498

RafalWilinski:
From my understanding of the topic purging these translations (from MessagesXX.php files) _shouldn't_ affect something else than `Special:PasswordReset` which is also going to be removed.

wladekb:
lgtm,
I think you should also remove that section as well:
https://github.com/Wikia/app/blob/dev/maintenance/language/messages.inc#L481-L498

RafalWilinski:
I'll test that.

RafalWilinski:
`$wgMessageStructure['passwordreset']` array is not used anywhere so I think it's safe to remove it.

PR #9704, opened 2016-02-12
PLATFORM-1828: Dedup SMWUpdateJob tasks

wladekb:
We've had a long list of SMWUpdateJob tasks in the queue recently. At least part of these tasks seem to be duplicates of other ones which previously have been deduplicated by the built-in Mediawiki job queue. Current implementation doesn't make any dupe checks. Let's fix it!

https://wikia-inc.atlassian.net/browse/PLATFORM-1828

/cc @macbre @Grunny


macbre:
:+1:  (assuming the `dupCheck` is handled on Celery level)

wladekb:
Phalanx integration tests are broken only.

PR #9710, opened 2016-02-15
INT-304 Support for CreateBox and InputBox with VE

perjg:
Support for using VE with CreateBox and InputBox extensions.
- CreateBox changes to launch VE when user's editor pref is VE
- InputBox changes to launch VE when user's editor pref is VE
- VE changes to support Preload parameter


extensions/VisualEditor/ApiVisualEditor.php
```diff
@@ -163,6 +163,16 @@ protected function getHTML( $title, $parserParams ) {
 			$timestamp = $latestRevision->getTimestamp();
 		} else {
 			$content = '';
+			$preloadTitle = Title::newFromText( urldecode( $parserParams['preload'] ) );
```
inez: I believe you should not have to urldecode - ApiMain/Base should already have handle it for you.

extensions/VisualEditor/ApiVisualEditor.php
```diff
@@ -163,6 +163,16 @@ protected function getHTML( $title, $parserParams ) {
...
+
+			if ( isset( $preloadTitle ) && $preloadTitle->userCan( 'read' ) ) {
+				$rev = Revision::newFromTitle($preloadTitle);
+
+				if (is_object($rev)) {
+					$content = $this->parseWikitextFragment( $preloadTitle, $rev->getText() );
```
inez: This is POST so it does not take adventage of caching - consider calling here getHTML (itself) and passing $preloadTitle as a parameter - and then eventually overriding oldid and or timestamp in the response, e.g.:

```
$out = $this->getHtml( Title::newFromText( $parserParams['preload'] ) );
$out['oldid'] = 0;
return $out;
```

With some appropriate "checks" around.

extensions/VisualEditor/ApiVisualEditor.php
```diff
@@ -359,6 +369,10 @@ public function execute() {
 			$parserParams['oldid'] = $params['oldid'];
 		}
 
+		if ( isset( $params['preload'] ) ) {
+			$parserParams['preload'] = $params['preload'];
```
inez: "preload" is not a parser parameter so should not be added to $parserParams array

extensions/VisualEditor/modules/ve-mw/init/ve.init.mw.Target.js
```diff
@@ -293,6 +293,28 @@ ve.init.mw.Target.static.integrationType = 'page';
 /* Static Methods */
 
 /**
+ * Process page query arguments
+ * @returns {object} The processed parameters
+ */
+function processArguments() {
```
inez: No need for this whole method - just look how other parameters are accessed - for instance "veaction": `ve.init.wikia.ViewPageTarget.init.js` - line 396

inez:
Have you consider making no changes to API at all and just making changes to the client - so request different articles when "preload" parameter is present (and valid)?

This approach overall is not safe from dirty diffs (to find out more about it read about selective serializer). I think it's ok as a experiment but in a long run we have to re-consider it.

extensions/InputBox/InputBox.hooks.php
```diff
@@ -68,8 +66,12 @@ public static function onMediaWikiPerformAction(
 		}
 		unset( $params['prefix'] );
 		$params['title'] = $title;
-		
-		global $wgScript;
+
+		if ( EditorPreference::isVisualEditorPrimary() && $wgUser->isLoggedIn() ) {
+			$params['veaction'] = $params['action'];
```
inez: any reason why not to just assign "edit" as a value?

extensions/VisualEditor/ApiVisualEditor.php
perjg: Title::newFromText returns null if I don't urldecode.

perjg:
Latest commit:
- Removed processArguments and use mw.Uri instead
- Added extra parameter to getHTML and removed preload from parserArguments
- Used suggested alternative method for retrieving HTML for preload page
- Changed assignment in InputBox.hooks.php

Not changed:
- Overall approach
- I still call urldecode, otherwise I get a null return value. See above comment.

extensions/VisualEditor/modules/ve-mw/init/ve.init.mw.Target.js
```diff
@@ -1014,6 +1014,10 @@ ve.init.mw.Target.prototype.load = function ( additionalModules ) {
 		data.oldid = this.revid;
 	}
 
+	if (uri.query.preload) {
```
inez: There are missing spaces - not sure why linter did not complain about it.

extensions/VisualEditor/ApiVisualEditor.php
```diff
@@ -133,6 +133,24 @@ protected function requestParsoid( $method, $title, $params ) {
 		return $req->getContent();
 	}
 
+
+	protected function getPreloadHTML( $params ) {
+		$preloadTitle = Title::newFromText( urldecode( $params['preload'] ) );
```
inez: I believe this should still work without urldecode - just like it works here - https://github.com/Wikia/app/blob/INT-304/extensions/VisualEditor/ApiVisualEditor.php#L366. Is it possible that it's somehow encoded twice?

extensions/VisualEditor/ApiVisualEditor.php
```diff
@@ -367,7 +385,12 @@ public function execute() {
 		wfDebugLog( 'visualeditor', "called on '$page' with paction: '{$params['paction']}'" );
 		switch ( $params['paction'] ) {
 			case 'parse':
-				$parsed = $this->getHTML( $page, $parserParams );
+				if ( isset( $params['preload'] )) {
+					$parsed = $this->getPreloadHTML( $params );
```
inez: If $parsed is false it will later return "Error contacting the Parsoid server" (https://github.com/Wikia/app/blob/INT-304/extensions/VisualEditor/ApiVisualEditor.php#L541) but this might be not the case (getPreloadHTML returns false for different reason).
Also, I think you should be passing to getPreloadHTML already just a Title (object) or title (string) of the article you want to retrieve.

inez:
I still think that approach of making this API more complicated than it is (additional parameter, additional conditional logic) might be not the best way to go given that it's not really needed. It seem that it could be handled completely on the client (if preload parameter is present then call "parse" API with different "page" value), which needs to be modified anyways. Are there some downsides to doing it that way?

extensions/VisualEditor/ApiVisualEditor.php
perjg: In the url it looks like preload=MediaWiki:Character-preload

When I process it PHP side $params['preload'] is MediaWiki%3ACharacter-preload

Title::newFromText expects the decoded name.

I don't do any other processing of $params['preload'], so it should be handled the same way as all other elements in $params.

Is colon (:) a valid part of a plain MediaWiki page, or is it only used for "special" pages such as the MediaWikia: page here?

extensions/VisualEditor/ApiVisualEditor.php
perjg: I moved the extraction of the title into the call of getPreloadHTML like suggested.

The error you mention will happen if the preload parameter is pointing to a page that doesn't exist.
I tested it (see error section here: http://visualeditor.pgroland.wikia-dev.com/wiki/Test1) and it doesn't end up with a user-facing error, it just shows empty contents in VE.

I could rewrite the code like so:

```
            if ( isset( $params['preload'] )) {
                $parsed = $this->getPreloadHTML( $urldecode( $params['preload'] ) );
            }

            if ( !$parsed ) 
                $parsed = $this->getHTML( $page, $parserParams );
            }
```

Which might be safer, but I don't see any difference in the user experience. Is there something else going on (logging?) that means that approach is better?

inez: This is an API and VE just happens to be the main client of it - but it
doesn't mean that we should return incorrect error because VE is going to
handle it the same way (creating a ticket to have VE handle it in more
robust way perhaps would be more appropriate action).
The more I look at it the more it looks that API was already good and
changes should be done on the client.

perjg:
@inez I've changed the two immediate things you asked for:
- Fixed broken formatting
- Moved extraction of title parameter to call to getPreloadHTML rather than inside it

I'm not sure about the parsoid error you mention, see my comment above.

As for your suggestion of an alternative implementation of the task, I'd like to make sure that I understand exactly what you mean before trying out other alternative implementations.
Let me send you an email about it.

extensions/VisualEditor/ApiVisualEditor.php
```diff
@@ -164,8 +164,8 @@ protected function getHTML( $title, $parserParams ) {
 		} else {
 			$content = '';
 			$timestamp = wfTimestampNow();
-			$oldid = 0;
-		}
+  			$oldid = 0;
+  		}
```
inez: Looks like accidentally added whitespaces.

extensions/VisualEditor/ApiVisualEditor.php
```diff
@@ -368,6 +364,7 @@ public function execute() {
 		switch ( $params['paction'] ) {
 			case 'parse':
 				$parsed = $this->getHTML( $page, $parserParams );
+
```
inez: Looks like accidentally added new line.

extensions/VisualEditor/ApiVisualEditor.php
```diff
@@ -657,6 +654,7 @@ public function getAllowedParams() {
 			'html' => null,
 			'cachekey' => null,
 			'pst' => false,
+			'preload' => null,
```
inez: This is not needed anymore.

extensions/VisualEditor/modules/ve-mw/init/ve.init.mw.Target.js
```diff
@@ -1014,6 +1014,10 @@ ve.init.mw.Target.prototype.load = function ( additionalModules ) {
 		data.oldid = this.revid;
 	}
 
+	if ( uri.query.preload ) {
```
inez: Probably you want to add something like: `mw.config.get('wgArticleId') === 0` (btw. this was already mentioned in the PR/CR)

extensions/VisualEditor/modules/ve-mw/init/ve.init.mw.Target.js
```diff
@@ -1014,6 +1014,10 @@ ve.init.mw.Target.prototype.load = function ( additionalModules ) {
 		data.oldid = this.revid;
 	}
 
+	if ( uri.query.preload && mw.config.get('wgArticleId') === 0 ) {
```
inez: Missing spaces.

inez:
LGTM (except for missing spaces - left separated comment about it - not critical tho)

extensions/InputBox/InputBox.hooks.php
```diff
@@ -52,24 +52,38 @@ public static function onMediaWikiPerformAction(
...
 		if( $wiki->getAction( $request ) !== 'edit' ){
 			# not our problem
 			return true;
 		}
+
+		$params = $request->getValues();
+
+		if ( $request->getText( 'preload', '' ) !== '' ){
+			# Make sure any call with preload uses VE when it matches the user's prefs
+			if ( EditorPreference::isVisualEditorPrimary() && $wgUser->isLoggedIn() ) {
```
Grunny: You can avoid the use of the global `$wgUser` as the current user is passed to this method in the `$user` variable.

extensions/InputBox/InputBox.hooks.php
```diff
@@ -52,24 +52,38 @@ public static function onMediaWikiPerformAction(
...
+		global $wgScript, $wgUser;
+
 		if( $wiki->getAction( $request ) !== 'edit' ){
 			# not our problem
 			return true;
 		}
+
+		$params = $request->getValues();
+
+		if ( $request->getText( 'preload', '' ) !== '' ){
```
Grunny: There should be a space between the closing parenthesis and the opening brace: `) {`.

extensions/InputBox/InputBox.hooks.php
```diff
@@ -52,24 +52,38 @@ public static function onMediaWikiPerformAction(
...
+			# Make sure any call with preload uses VE when it matches the user's prefs
+			if ( EditorPreference::isVisualEditorPrimary() && $wgUser->isLoggedIn() ) {
+				$params['veaction'] = 'edit';
+				unset( $params['action'] );
+
+				$output->redirect( wfAppendQuery( $wgScript, $params ), '301' );
+				return false;
```
Grunny: Can `preload` and `prefix` both be used for a single input box? If so, this would stop `prefix` from working if `preload` was set, right?

perjg: In our use case, we only use preload.

However, there is nothing preventing somebody for adding both preload and prefix, so I should probably add in a check for that.

How about the following at line 70:

`
if( $request->getText( 'prefix', '' ) !== '' && isset( $params['title'] ) ) {
    $title = $params['prefix'];
    $title .= $params['title'];
    $params['title'] = $title;
    unset( $params['prefix'] );
}
`

Grunny: Yeah, I'd make the early return if neither prefix nor preload are set, i.e.

``` php
if ( $request->getText( 'prefix', '' ) === '' && $request->getText( 'preload', '' ) === '' ) {
    // Fine
    return true;
}
```

Then put each of the params changes in if blocks, i.e. something like:

``` php
if (
    !empty( $params['preload'] )
    && EditorPreference::isVisualEditorPrimary()
    && $user->isLoggedIn()
) {
    $params['veaction'] = 'edit';
    unset( $params['action'] );
}

if ( !empty( $params['prefix'] ) ) {
    $title = $params['prefix'];
    if ( isset( $params['title'] ) ) {
        $title .= $params['title'];
    }
    unset( $params['prefix'] );
    $params['title'] = $title;
}

$output->redirect( wfAppendQuery( $wgScript, $params ), '301' );
return false;
```

What do you think?

perjg: Looks good. I'll try it out.

perjg: Shouldn't it be this for the early return:

``` php
if ( $request->getText( 'prefix', '' ) === '' && $request->getText( 'preload', '' ) === '' ) {
    // Fine
    return true;
}
```

The code sets prefix='' in all requests, so it will always be there, but mostly be empty string.
For preload, the parameter is only present when the InputBox has preload data.

What's the difference between these three?

``` php
$request->getText( 'prefix', '' ) === ''
empty( $params['prefix'] )
isset( $params['title'] )
```

perjg: I tested, and it looks to me like I could simplify the checks to all use the same format.

If my understanding is correct, checking for `$request->getText( 'prefix', '' ) === ''` is equivalent to checking for `empty( $params['prefix'] )`. 

For reference, the url want to pick up generally looks like this:
http://visualeditor.pgroland.wikia-dev.com/index.php?action=edit&preload=MediaWiki%3ACharacter-preload&editintro=&summary=&nosummary=&prefix=&minor=&title=fdsfdsf3e423&create=Create+page

This sets prefix explicitly to an empty string, and it doesn't seem to matter if an empty parameter is present, or if it is not in the url at all, empty will still return true.

Grunny: > I tested, and it looks to me like I could simplify the checks to all use the same format.
> If my understanding is correct, checking for $request->getText( 'prefix', '' ) === '' is equivalent to checking for empty( $params['prefix'] ).

Yep, it is. The above was just untested GitHub comment code ;).

perjg: The existing code was pretty inconsistent in that respect too, which is what confused me into thinking they might not be the same.

Grunny:
LGTM.

PR #9717, opened 2016-02-15
Merge pull request #9597 from Wikia/XW-41

bognix:
XW-41 | Use new format of Ember component placeholder for single images


rogatty:
LGTM

PR #9719, opened 2016-02-16
SOC-1778 | Discussion User's IP History log

bkoval:
@garthwebb @armonr @jsutterfield Could you take a look?
https://wikia-inc.atlassian.net/browse/SOC-1778


extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLog.i18n.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+$messages = array();
+
+$messages['en'] = array(
```
bkoval, line 5: There's no Crowdin conf file as i18n is not a requirement for prod, but all the strings are here to make it easier in the future.

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -0,0 +1,218 @@
...
+			}
+		}
+
+		return $records;
+	}
+
+	private function getQuery( $userId ) {
+		$pagination_size = self::PAGINATION_SIZE;
+		return <<<JSON_BODY
+{
```
bkoval, line 179: I'm not so sure what to do with it. I was thinking about just converting it to a normal PHP array and then using json_encode, but I'm not sure if it improves readability.

bkoval:
Seems like Image Service cropping test fails and it doesn't relate to any of the work here.

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLog.i18n.php
armonr, line 5: :+1:

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -0,0 +1,230 @@
...
+class SpecialDiscussionsLogController extends WikiaSpecialPageController {
+	// @todo
+	//1. Pagination
+	//2. Add date range input field
+
+	const KIBANA_BASEURL = 'http://query.es.service.sjc.consul:9200';
+	const PAGINATION_SIZE = 50;
+	const DAYS_RANGE = 15;
+	const HTTP_STATUS_OK = 200;
+	const NO_USER_MATCH_ERROR = "Provided username did not match any user";
```
armonr, line 14: Minor: Should this use `discussionslog-no-user-match-error` now?

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
armonr, line 179: We could, but I think this is fine for now until this needs to start handling different types of queries.

armonr:
Super minor comment, but looks great! Thanks @bkoval :100:

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -0,0 +1,230 @@
...
+		}
+	}]
+}
+JSON_BODY;
+	}
+
+	private function checkAccess() {
+		if ( !$this->wg->User->isLoggedIn() || !$this->wg->User->isAllowed( 'forumadmin' ) ) {
+
+			$this->wg->Out->clearHTML();
```
jsutterfield: Updating the output inside of this method feels like a side effect to me. Since this method is called `checkAccess`, I think that's all we should have it do and move this output logic after the the check.

Also, there's this nifty [PermissionsErrors.php](https://github.com/Wikia/app/blob/dev/includes/Exception.php#L373) which could be useful. Here's an example of its usage: [Usage](https://github.com/Wikia/app/blob/dev/extensions/wikia/Email/SpecialSendEmailController.class.php#L32).

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -0,0 +1,230 @@
...
+
+/**
+ * Discussion user log page
+ */
+class SpecialDiscussionsLogController extends WikiaSpecialPageController {
+	// @todo
+	//1. Pagination
+	//2. Add date range input field
+
+	const KIBANA_BASEURL = 'http://query.es.service.sjc.consul:9200';
```
jsutterfield: One of the AC is to "Remove any hardcoded values", I think this counts as hard-coded, yea? ;)

We actually have a class to resolve these urls: [ConsulUrlProvider.java](https://github.com/Wikia/app/blob/dev/lib/Wikia/src/Service/Gateway/ConsulUrlProvider.php). It takes a service tag ('query') and optional dc ('sjc'). So using it would look something like this:

``` PHP
global $wgConsulUrl;
$urlProvider = new  Wikia\Service\Gateway\ConsulUrlProvider( $wgConsulUrl, 'query', 'sjc' );
$esUrl = $urlProvider->getUrl( 'es' );
```

extensions/wikia/SpecialDiscussionsLog/UserLogRecord.class.php
```diff
@@ -0,0 +1,79 @@
+<?php
+
+/**
+ * Class UserLogRecord
+ */
+class UserLogRecord {
```
jsutterfield, line 6: So since this class doesn't have any behaviors, just vanilla getters/setters, it's really more of a data structure. What if we made all of these fields public and just treated it like a c-style struct? That would be cleaner IMO.

``` PHP
class UserLogRecord {
    public $ip;
    public $location;
    public $language;
    public $userAgent;
    public $app;
    public $timestamp;
    public $userId;
    public $userName;
}
```

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -0,0 +1,230 @@
...
+		try {
+			$userId = $this->getUserIdByUsername( $userName );
+		} catch ( Exception $e ) {
+			$hasUserError = true;
+		}
+
+		$userLogRecords = $this->aggregateLogSearches( $userId, $userName );
+
+		if ( count( $userLogRecords ) == 0 ) {
+			$hasNoUserLogRecords = true;
```
jsutterfield, line 89: I think we should be able to handle this case in the template. See [here](http://stackoverflow.com/questions/11653764/mustache-how-to-detect-array-is-not-empty)

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -0,0 +1,230 @@
...
+		$userId = null;
+		$userErrorMessage = null;
+
+		try {
+			$userId = $this->getUserIdByUsername( $userName );
+		} catch ( Exception $e ) {
+			$hasUserError = true;
+		}
+
+		$userLogRecords = $this->aggregateLogSearches( $userId, $userName );
```
jsutterfield: How about we check the state of `$hasUserError` before making these queries?

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -0,0 +1,230 @@
...
+					sprintf( 'Request to elasticsearch failed: %s', $requestException->getMessage() )
+				);
+				continue;
+			}
+
+			if ( $response->getStatusCode() !== self::HTTP_STATUS_OK ) {
+				$logger->warning(
+					sprintf( 'Elasticsearch request error; status code %d', $response->getStatusCode() )
+				);
+				continue;
```
jsutterfield: I feel like this line, and L154 should be a `break` rather than a `continue`. If these requests are failing, it doesn't seem like we should hit ES 14 more times.

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
bkoval, line 14: Well, I was thinking about it, but this const is use just as an exception message, it's not user-facing. I assumed it makes more sense to always have it in one form, so there's an easier way to construct a query to find these in Kibana etc. What do you think?

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
bkoval: Agreed.

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
bkoval: :+1:

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
bkoval, line 89: Agreed. Unfortunately, there seems to be a problem with this type of check in our php implementation of mustache. I created https://wikia-inc.atlassian.net/browse/MAIN-6445 to deal with it after we upgrade to the new version of the mustache extension.

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
bkoval: Done.

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
bkoval: Done.

extensions/wikia/SpecialDiscussionsLog/UserLogRecord.class.php
bkoval, line 6: Done :)

jsutterfield:
LGTM!

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -0,0 +1,229 @@
+<?php
+
+/**
+ * Discussion user log page
+ */
+class SpecialDiscussionsLogController extends WikiaSpecialPageController {
+	// @todo
```
kvas-damian, line 7: What about thins TODO? 
I believe we should resolve it or create a ticket to address that.

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -0,0 +1,229 @@
...
+		if ( $this->wg->request->wasPosted() ) {
+			$userName = $this->wg->request->getVal( 'username', null );
+			$output .= $this->getUserLog( $userName );
+		}
+
+		$this->wg->Out->addHtml( $output );
+	}
+
+	private function getInputForm() {
+		return \MustacheService::getInstance()->render(
```
kvas-damian: Maybe we could use default Nirvana way of rendering templates + `sendRequest`?
it's much easier to follow while all controllers use this pattern.

extensions/wikia/SpecialDiscussionsLog/templates/SpecialDiscussionsLog_userLog.mustache
```diff
@@ -0,0 +1,33 @@
...
+{{^hasUserError}}
+	{{#hasNoUserLogRecords}}
+		<p>{{noUserLogRecordsMessage}}</p>
+	{{/hasNoUserLogRecords}}
+	{{^hasNoUserLogRecords}}
+		<p>{{logTableCaption}}</p>
+		<table class="article-table">
+			<thead>
+			<tr>
+				<th style="padding-right:8px">{{ipAddressHeader}}</th>
```
kvas-damian: Is it possible to move styling to CSS file?

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -0,0 +1,229 @@
...
+		"@timestamp": {
+			"order": "desc"
+		}
+	}]
+}
+JSON_BODY;
+	}
+
+	private function checkAccess() {
+		if ( !$this->wg->User->isLoggedIn() || !$this->wg->User->isAllowed( 'forumadmin' ) ) {
```
Grunny: Admins also have `forumadmin` and should not by default have access to IP address information of users as that is private. Shouldn't this be using the `specialdiscussionslog` permission you define above? If not, an existing permission to check for whether someone should be allowed to view private IP address information is `checkuser`.

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
bkoval, line 7: https://wikia-inc.atlassian.net/browse/SOC-2010 There you go :)

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
bkoval: Right, it should use 'specialdiscussionslog' permissions, good catch! And sorry for that ;)

bkoval:
LGTM

armonr:
@bkoval: Added this for sorting ff9c06e8fd6f2441a05a4aed8eb6e9cef4687908

PR #9720, opened 2016-02-16
[PLATFORM-1483] User permissions lib

jcellary:
This PR is centralizing permission handling inside of media wiki, so that later on we can move that logic to the user permissions service.

For that purpose a new permission service lib has been introduced (similarly to user preferences) and all logic from the User class is moved to the lib. Methods in the User class now only serve as a proxy. 

From an altered lines perspective most of changes are moving permission definition to a single file located in config repo.

This PR needs to be merged together with https://github.com/Wikia/config/pull/1620


extensions/wikia/SpecialBatchUserRights/SpecialBatchUserRights.php
```diff
@@ -42,6 +49,17 @@ public function __construct() {
...
+	 * @return UserPermissions
+	 */
+	private function userPermissions() {
+		if ( is_null( $this->permissionsService ) ) {
+			$this->permissionsService = Injector::getInjector()->get( PermissionsService::class );
+		}
+
+		return $this->permissionsService;
+	}
+
```
macbre: This one is repeated in quite a few places. Let's consider introducing a trait that will provide `userPermissions` method.

extensions/3rdparty/LyricWiki/Hook_PreventBlanking.php
```diff
@@ -3,8 +3,6 @@
 require_once 'extras.php';
 
 $wgHooks['EditFilter'][] = 'stopBlanking';
-$wgAvailableRights[] = 'allowedtoblank';
-$wgGroupPermissions['sysop']['allowedtoblank'] = true;
 $wgExtensionMessagesFiles["PreventBlanking"] = "$LW/Hook_PreventBlanking.i18n.php";
```
wladekb, line 6: If you are removing the rights, is this still in use?

extensions/3rdparty/LyricWiki/LyricFind/LyricFind.setup.php
```diff
@@ -31,12 +31,6 @@
 // LyricFind indexing
 $wgHooks['ParserBeforeStrip'][] = 'LyricFindHooks::onParserBeforeStrip';
 
-// edit permissions & view-source protection
-// @see http://www.mediawiki.org/wiki/Manual:$wgNamespaceProtection
-$wgGroupPermissions['*']['editlyricfind'] = false;
-$wgGroupPermissions['staff']['editlyricfind'] = true;
-$wgGroupPermissions['sysop']['editlyricfind'] = true;
-
 $wgHooks['AlternateEdit'][] = 'LyricFindHooks::onAlternateEdit';
```
wladekb, line 34: Please find out if we can remove everything related to LyricFind. AFAIK we stopped using that service X years ago.
@owend @SeanColombo

SeanColombo, line 34: @wladekb nope! We stopped using Gracenote, but we VERY MUCH use LyricFind curreclty and that's how we stay licensed.

I know it's confusing because it changes between a bunch of things with similar names, but LyricFind _is_ our current sub-licensor :)

extensions/wikia/ContentReview/ContentReview.hooks.php
```diff
@@ -341,26 +338,6 @@ public function onShowDiff( \DifferenceEngine $diff, &$notice ) {
...
-			$disabled = true;
-		}
-
-		return true;
-	}
-
-	public function onUserAddGroup( \User $user, $group ) {
-		global $wgUser;
-
-		if ( $group === 'content-reviewer' && ( !$wgUser->isAllowed( 'content-review' ) || !$wgUser->isStaff() ) ) {
```
wladekb, line 357: where this logic will be placed now?

extensions/wikia/GameGuides/GameGuides_setup.php
```diff
@@ -42,16 +38,7 @@
 //$wgAutoloadClasses[ 'GameGuidesSpecialContentController'] =  "{$dir}/GameGuidesSpecialContentController.class.php" ;
 //$wgSpecialPages[ 'GameGuidesContent' ] =  'GameGuidesSpecialContentController';
 
-$wgGroupPermissions['*']['gameguidescontent'] = false;
-$wgGroupPermissions['staff']['gameguidescontent'] = true;
-$wgGroupPermissions['helper']['gameguidescontent'] = true;
-
-if ( $wgGameGuidesContentForAdmins ) {
-	$wgGroupPermissions['sysop']['gameguidescontent'] = true;
```
wladekb, line 50: how are we going to support this kind of logic?

extensions/wikia/Listusers/SpecialListusers_helper.php
```diff
@@ -22,6 +25,11 @@ class ListusersData {
 	var $mDBh;
 	var $mTable;
 
+	/**
+	 * @var UserPermissions
+	 */
+	private $permissionsService;
```
wladekb: `$userPermissionsService` ? so it will be consistent with the accessor method name `userPermissions` ?

extensions/wikia/PowerUser/PowerUser.class.php
```diff
@@ -69,6 +71,22 @@ function __construct( \User $oUser ) {
 	}
 
 	/**
+	 * @var UserPermissions
+	 */
+	private $permissionsService;
```
wladekb: `$userPermissionsService` ? (and all the other classes too)

extensions/wikia/PowerUser/PowerUser.class.php
```diff
@@ -122,10 +140,10 @@ public function addPowerUserSetOption( $sProperty ) {
 	public function addPowerUserAddGroup( $sProperty ) {
 		if ( in_array( $sProperty, self::$aPowerUserProperties )
 			&& $this->bUseGroups
-			&& !in_array( self::GROUP_NAME, \UserRights::getGlobalGroups( $this->oUser ) )
+			&& !in_array( self::GROUP_NAME, $this->userPermissions()->getExplicitGlobalUserGroups( $this->oUser->getId() ) )
```
wladekb: it would be much more readable if we could write something like:
`$this->userPermissions()->hasGlobalGroup( $this->oUser->getId(), self::GROUP_NAME )`
or:
`$this->userPermissions()->user( $this->oUser )->hasGlobalGroup( self::GROUP_NAME )` (or `forUser` or `withUser` or... ) - it would be convenient if such a method would accept both integer IDs and MW's User objects (or at least we had a chance to pass a User object to a separate method)

can we add a handful of helper methods to the user permissions service class and use them wherever possible?

-and-

we could have two private methods:
- `userPermissionsService()` that returns UserPermissionsService object
- `userPermissions( $user )` that returns a proxy to UserPermissionsService bound to a specific user (that would shorten constructs like this one: `$this->userPermissions()->user( $this->oUser )`

extensions/wikia/PowerUser/PowerUser.class.php
```diff
@@ -69,6 +71,22 @@ function __construct( \User $oUser ) {
...
 
 	/**
+	 * @var UserPermissions
+	 */
+	private $permissionsService;
+
+	/**
+	 * @return UserPermissions
+	 */
+	private function userPermissions() {
```
wladekb: this method and the private field is repeated unchanged in many classes, good candidate for `trait` ?

macbre: https://github.com/Wikia/app/pull/9720#discussion_r53135072 :)

includes/Autopromote.php
```diff
@@ -28,45 +28,6 @@ public static function getAutopromoteGroups( User $user ) {
 	}
 
 	/**
-	 * Get the groups for the given user based on the given criteria.
```
wladekb: will we use Autopromote for anything after this PR is merged? have you considered removing everything related to autopromote?

macbre:
One general suggestion: this pull request is still in progress, but already quite big. Can we perform a cleanup of code
- Removal of unused extension 3rdparty/SpecialProtectSite (not to confuse with wikia/SpecialProtectSite)
- Removal of `WIKIA_ENV_INTERNAL` (retired wikia one)

as a part of a separate pull request to make things a bit simpler and release these (quite big) changes in batches?

includes/User.php
```diff
@@ -5237,4 +4613,256 @@ protected function logFallbackToMediaWikiSessionRejection( $from ) {
...
+		return self::userPermissions()->getGroupsChangeableByGroup( $group );
+	}
+
+	/**
+	 * Add the user to the given group.
+	 * This takes immediate effect.
+	 * @param $group String Name of the group to add
+	 */
+	public function addGroup( $group ) {
+		global $wgUser;
```
wladekb: why are you changing a behavior here?
the original method operated on the current object, not on the currently logged-in user

the same below in a few places

includes/User.php
```diff
@@ -5237,4 +4613,256 @@ protected function logFallbackToMediaWikiSessionRejection( $from ) {
...
+		return self::userPermissions()->addUserToGroup( $wgUser, $this, $group );
```
wladekb: use `RequestContext::getMain()->getUser()` instead of `$wgUser`
the same below

includes/specials/SpecialListgrouprights.php
```diff
@@ -113,11 +125,13 @@ public function execute( $par ) {
...
 			}
 
+			$groupArr = $this->userPermissions()->getGroupsChangeableByGroup( $group );
+
 			$revoke = isset( $wgRevokePermissions[$group] ) ? $wgRevokePermissions[$group] : array();
-			$addgroups = isset( $wgAddGroups[$group] ) ? $wgAddGroups[$group] : array();
-			$removegroups = isset( $wgRemoveGroups[$group] ) ? $wgRemoveGroups[$group] : array();
-			$addgroupsSelf = isset( $wgGroupsAddToSelf[$group] ) ? $wgGroupsAddToSelf[$group] : array();
-			$removegroupsSelf = isset( $wgGroupsRemoveFromSelf[$group] ) ? $wgGroupsRemoveFromSelf[$group] : array();
+			$addgroups = isset( $groupArr['add'] ) ? $groupArr['add'] : array();
```
wladekb: how about enforcing `getGroupsChangeableByGroup` to return all those keys and then we could skip checking if they exist here?

includes/specials/SpecialStatistics.php
```diff
@@ -223,8 +243,7 @@ private function getGroupStats() {
 			$countUsers = SiteStats::numberingroup( $groupname );
 			if( $countUsers == 0 ) {
 				// wikia change start
-				global $wgWikiaGlobalUserGroups;
-				if( is_array( $wgWikiaGlobalUserGroups ) && in_array( $groupname, $wgWikiaGlobalUserGroups ) ) {
+				if( in_array( $groupname, $this->userPermissions()->getGlobalGroups() ) ) {
```
wladekb: the helper methods would be quite nice here and there...

includes/specials/SpecialUserrights.php
```diff
@@ -504,6 +511,11 @@ protected static function getAllGroups() {
 		return User::getAllGroups();
 	}
 
+	private function isGlobalNonEditableGroup( $group ) {
```
wladekb: I've seen this before, worth moving it to a shared place?

extensions/3rdparty/LyricWiki/Hook_PreventBlanking.php
jcellary, line 6: Not sure I understand - I'm not removing any rights, I'm just moving all definitions to one single place in config.

lib/Wikia/src/Service/User/Permissions/PermissionsService.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+namespace Wikia\Service\User\Permissions;
+
+interface PermissionsService {
+	public function getGlobalGroups();
```
wladekb: would be quite helpful to have phpdocs describing parameter types and return values

extensions/wikia/ContentReview/ContentReview.hooks.php
jcellary, line 357: onUserAddGroup is part of a check inside of the user permissions service. As for GroupCheckboxes I moved that logic directly to the special page displaying it

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
```diff
@@ -0,0 +1,832 @@
...
+		}
+		return $this->globalExplicitUserGroups[$userId];
+	}
+
+	private function setGlobalUserGroupsArray( $userId, $groups ) {
+		$this->globalExplicitUserGroups[$userId] = $groups;
+	}
+
+	private function loadGlobalGroups() {
+		$this->globalGroups[] = 'content-reviewer';
```
wladekb:
```
$this->globalGroups = [
   'one',
   'two',
   ....
```

or even move it to the property declaration

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
```diff
@@ -0,0 +1,832 @@
...
+		$this->globalGroups[] = 'restricted-login';
+		$this->globalGroups[] = 'council';
+		$this->globalGroups[] = 'authenticated';
+		$this->globalGroups[] = 'wikiastars';
+		$this->globalGroups[] = 'restricted-login';
+		$this->globalGroups[] = 'voldev';
+	}
+
+	private function loadImplicitGroups() {
+		$this->implicitGroups[] = '*';
```
wladekb: same here as above

extensions/wikia/GameGuides/GameGuides_setup.php
jcellary, line 50: I moved these lines as is to the new place, so the if will stay as is. You're right however that I can double check when is the wgGameGuidesContentForAdmins set to make sure we don't load rights before the value of this variable is set.

extensions/wikia/Listusers/SpecialListusers_helper.php
jcellary: True. I'll convert it to a trait as Macbre suggested

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
```diff
@@ -0,0 +1,832 @@
...
+			'refreshspecial',
+			'replacetext',
+			'spamregex',
+			'tboverride', 	// Implies tboverride-account
+			'tboverride-account', 	// For account creation
+			'torunblocked',
+			'abusefilter-bypass',
+			'platinum',
+			'sponsored-achievements',
+			'achievements-exempt',
```
wladekb: achievements are not enabled everywhere and this right should probably not be listed on wikis that don't have it

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
```diff
@@ -0,0 +1,832 @@
...
+			'achievements-explicit',
+			'admindashboard',
+			'commentmove',
+			'commentedit',
+			'commentdelete',
+			'becp_user',
+			'blog-comments-toggle',
```
wladekb: same for blogs etc.

includes/Autopromote.php
jcellary: Yes, autopromote per se is still used. What is not used is AutopromoteONCE - that is a subset of the whole autopromote logic.

lib/Wikia/src/Service/User/Permissions/PermissionsService.php
```diff
@@ -0,0 +1,43 @@
...
+
+	public function getImplicitGroups();
+
+	public function getExplicitGroups();
+
+	public function getPermissions();
+
+	public function getExplicitUserGroups( $userId );
```
wladekb: why some of the methods accept IDs and some others User objects?

jcellary:
You're right Macbre, I'll create a separate PR for the cleanups and release it step by step.

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
```diff
@@ -0,0 +1,832 @@
...
+
+		return array_unique( array_merge (
+			$this->getExplicitLocalUserGroups( $userId ),
+			$this->getExplicitGlobalUserGroups( $userId )
+		) );
+	}
+
+	public function getExplicitLocalUserGroups( $userId ) {
+		$this->loadLocalGroups( $userId );
+		return $this->getLocalUserGroupsArray( $userId );
```
wladekb: I'd suggest naming those methods in such a way it's more obvious which ones load the data and which one don't
I wouldn't guess which on of them does: `getExplicitLocalUserGroups` or `getLocalUserGroupsArray`

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
```diff
@@ -0,0 +1,832 @@
...
+	public function getExplicitGroups() {
+		return $this->explicitGroups;
+	}
+
+	public function getExplicitUserGroups( $userId ) {
+		if ( empty( $userId ) ) {
+			return [];
+		}
+
+		return array_unique( array_merge (
```
wladekb, line 31: calculated every time....

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
```diff
@@ -0,0 +1,832 @@
...
+	 * Get the list of explicit and implicit group memberships this user has.
+	 * This includes all explicit groups, plus 'user' if logged in,
+	 * '*' for all accounts, and autopromoted groups
+	 * @param \User $user
+	 * @param $reCacheAutomaticGroups
+	 * @return Array of String internal group names
+	 */
+	public function getEffectiveUserGroups( \User $user, $reCacheAutomaticGroups = false ) {
+
+		return array_unique( array_merge(
```
wladekb, line 98: calculated every time... (please also look at other methods when fixing it)

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
```diff
@@ -0,0 +1,832 @@
...
+	static private function getSharedDB( $db = DB_SLAVE ) {
+		global $wgExternalSharedDB;
+		return wfGetDB( $db, [], $wgExternalSharedDB );
+	}
+
+	private function loadGlobalUserGroups( $userId ) {
+		if ( !empty( $userId ) && $this->getGlobalUserGroupsArray( $userId ) == false ) {
+
+			$fname = __METHOD__;
+			$globalGroups = \WikiaDataAccess::cache(
```
wladekb, line 152: this should be also invalidated when `User::invalidateCache` is called

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
```diff
@@ -0,0 +1,832 @@
...
+		return (bool)$wgWikiaIsCentralWiki;
+	}
+
+	private function addUserToGlobalGroup( \User $user, $group ) {
+		if ( !$this->isCentralWiki() ) {
+			return false;
+		}
+
+		// Purge cache first so in case of DB failure we don't leave inconsistent data in cache
+		\WikiaDataAccess::cachePurge( self::getMemcKey( $user->getId() ) );
```
wladekb: race condition here... 
one can read data from DB between you cache purge and writing the new value to DB, especially when the data is read from slave

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
```diff
@@ -0,0 +1,832 @@
...
+				),
+				__METHOD__,
+				array( 'IGNORE' ) );
+		}
+		$this->invalidateUserGroupsAndPermissions( $user->getId() );
+
+		return true;
+	}
+
+	private function canGroupBeAdded( \User $userPerformingChange, \User $userToChange, $group ) {
```
wladekb: `$userPerformingChange` is a bit long, how about `$performer`, `$doer` or something else which is shorter?

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
```diff
@@ -0,0 +1,832 @@
...
+			'add-self' => array(),
+			'remove-self' => array()
+		);
+		$addergroups = $this->getEffectiveUserGroups( $user );
+
+		foreach( $addergroups as $addergroup ) {
+			$groups = array_merge_recursive(
+				$groups, $this->getGroupsChangeableByGroup( $addergroup )
+			);
+			$groups['add']    = array_unique( $groups['add'] );
```
wladekb: I guess we could do a single pass of `array_unique` calls after the `foreach` loop, is that correct?

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
jcellary: That's a case with a lot of above mentioned permissions, but that was one of intended changes - that we make permissions definition global not extension based. When moving the logic to the external service there will be no notion of extensions so if you're in a specfic group you simply get all rights defined for it.

lib/Wikia/src/Service/User/Permissions/PermissionsService.php
jcellary: Because some function need only the id, however some other use methods defined in the user class. But you're right it can be confusing - I'll take a second look if we can standardize it, till now I simply based it on actual needs.

jcellary:
@wladekb I believe I addressed all your comments, let me know if you have anything else. Otherwise I'll run some more tests and be getting close to merging it

extensions/3rdparty/LyricWiki/Hook_PreventBlanking.php
wladekb, line 6: yeah, makes sense, nvm

extensions/3rdparty/LyricWiki/LyricFind/LyricFind.setup.php
wladekb, line 34: @SeanColombo Thanks for confirmation!

extensions/wikia/ContentReview/ContentReview.hooks.php
wladekb, line 357: Please give a direct link to the relevant part of special page code, this PR is big!

extensions/wikia/GameGuides/GameGuides_setup.php
wladekb, line 50: thanks for explanation, should be fine

extensions/wikia/PowerUser/PowerUser.class.php
```diff
@@ -12,9 +12,13 @@
 namespace Wikia\PowerUser;
 
 use Wikia\Logger\Loggable;
+use Wikia\DependencyInjection\Injector;
+use Wikia\Service\User\Permissions\PermissionsAccessor;
+use Wikia\Service\User\Permissions\PermissionsService;
 
 class PowerUser {
 	use Loggable;
+	use PermissionsAccessor;
```
wladekb: The name of the trait doesn't suggest you will access a service, how about renaming it to `PermissionsServiceAccessor` ?

extensions/wikia/PowerUser/PowerUser.class.php
```diff
@@ -122,10 +126,11 @@ public function addPowerUserSetOption( $sProperty ) {
 	public function addPowerUserAddGroup( $sProperty ) {
 		if ( in_array( $sProperty, self::$aPowerUserProperties )
 			&& $this->bUseGroups
-			&& !in_array( self::GROUP_NAME, \UserRights::getGlobalGroups( $this->oUser ) )
+			&& !$this->permissionsService()->isUserInGroup( $this->oUser, self::GROUP_NAME )
```
wladekb: here we have `permissionsService()` ...

extensions/wikia/PowerUser/PowerUser.class.php
```diff
@@ -174,7 +179,7 @@ public function removePowerUserRemoveGroup( $sProperty ) {
 			&& $this->bUseGroups
 			&& $this->isGroupForRemoval( $sProperty )
 		) {
-			\UserRights::removeGlobalGroup( $this->oUser, self::GROUP_NAME );
+			$this->userPermissions()->removeUserFromGroup( $this->oUser, self::GROUP_NAME );
```
wladekb: ... and here `userPermissions()` - is something different returned or it's just in the middle of rename?

includes/specials/SpecialUserrights.php
```diff
@@ -45,15 +106,6 @@ public function userCanExecute( User $user ) {
 		return $this->userCanChangeRights( $user, false );
```
wladekb: You have converted method `userCanChangeRights` to static. Please make sure you adjust all the places where this is called to call it statically.

extensions/wikia/PowerUser/PowerUser.class.php
jcellary: missed that, thanks

extensions/wikia/PowerUser/PowerUser.class.php
jcellary: renamed

includes/User.php
```diff
@@ -309,6 +246,17 @@ private function userAttributes() {
 	}
 
 	/**
+	 * @return PermissionsService
+	 */
+	private static function permissionsService() {
```
wladekb, line 251: `use PermissionsAccessor` ?

wladekb, line 251: Oh, this is static here. why is it so?

lib/Wikia/src/Service/User/Permissions/PermissionsAccessor.php
```diff
@@ -0,0 +1,24 @@
...
+
+	/**
+	 * @var PermissionsService
+	 */
+	private $permissionsService;
+
+	/**
+	 * @return PermissionsService
+	 */
+	private function permissionsService() {
```
wladekb: I'd suggest `protected` here. Otherwise if class `A` uses `PermissionsAccessor` and the subclass `B` will also want to do it, it might have some problems IIRC.

lib/Wikia/src/Service/User/Permissions/PermissionsDefinition.php
```diff
@@ -0,0 +1,55 @@
+<?php
+
+namespace Wikia\Service\User\Permissions;
+
+interface PermissionsDefinition {
```
wladekb: `...Configuration` ?

lib/Wikia/src/Service/User/Permissions/PermissionsDefinitionImpl.php
```diff
@@ -0,0 +1,414 @@
...
+		'platinum',
+		'sponsored-achievements',
+		'achievements-exempt',
+		'achievements-explicit',
+		'admindashboard',
+		'commentmove',
+		'commentedit',
+		'commentdelete',
+		'becp_user',
+		'blog-comments-toggle',
```
wladekb: was it somehow addressed that this shouldn't be on this list on wikis that don't have blogs enabled (or you're sure we can change this behavior) ?

lib/Wikia/src/Service/User/Permissions/PermissionsService.php
```diff
@@ -0,0 +1,122 @@
...
+	 * Gets the object containing definitions of groups and permissions in the system
+	 * @return PermissionsDefinition
+	 */
+	public function getDefinitions();
+
+	/**
+	 * @param \User $user
+	 * @return string[] Get list of global and local groups to which the given user has been explicitly assigned to.
+	 */
+	public function getExplicitUserGroups( \User $user );
```
wladekb: all the methods in this class accept a `User` object and operate on user, can we get rid of all the `User`'s in the method names?

includes/User.php
jcellary, line 251: Because some methods that need it in the user class are static as well

lib/Wikia/src/Service/User/Permissions/PermissionsDefinitionImpl.php
jcellary: Yeah, that can be a problem - I want to send this version to com sup, and I explicitly wrote about this change asking if it's acceptable so we'll see what they will say.

jcellary:
@wladekb I made changes according to your comments, let me know if you have any other suggestions

lib/Wikia/src/Service/User/Permissions/PermissionsService.php
```diff
@@ -0,0 +1,122 @@
...
+	 */
+	public function removeUserFromGroup( \User $performer, \User $userToChange, $group );
+
+	/**
+	 * Checks whether the given user has the requested permission
+	 * @param \User $user
+	 * @param $permission string Name of permission
+	 * @return bool True if the user has the provided permission
+	 */
+	public function doesUserHavePermission( \User $user, $permission );
```
wladekb: `hasPermission` would be much shorter and IMHO such a naming is popular within Mediawiki ( @macbre agree? ) ?

lib/Wikia/src/Service/User/Permissions/PermissionsService.php
```diff
@@ -0,0 +1,122 @@
+<?php
+
+namespace Wikia\Service\User\Permissions;
+
+interface PermissionsService {
```
wladekb: I can see you didn't reply to one of my suggestions to introduce a "bound PermissionsService" using:
`$this->permissionsService()->user( $someUserObject )->_aMethodFromThisInterface_()`

Could you share your point of view on this topic?

lib/Wikia/tests/Service/User/UserPermissionsIntegrationTest.php
```diff
@@ -0,0 +1,293 @@
...
+	 */
+	const TEST_WIKI_NAME = "firefly";
+
+	/**
+	 * @var string
+	 */
+	const TEST_STAFF_USER_NAME = "PermissionTest";
+
+	protected function setUp() {
+		$this->staffUserId = \User::idFromName( self::TEST_STAFF_USER_NAME );
```
wladekb, line 41: How about defining this user as a "fixture' and mock all the dependent data so we don't have to have a working database for running this test?

lib/Wikia/tests/Service/User/UserPermissionsIntegrationTest.php
```diff
@@ -0,0 +1,293 @@
...
+
+		parent::setUp();
+	}
+
+	function testShouldReturnExplicitGroups() {
+		\WikiaDataAccess::cachePurge( PermissionsServiceImpl::getMemcKey( $this->staffUserId ) );
+
+		$groups = $this->permissionsService()->getExplicitUserGroups( $this->staffUser );
+		$this->assertContains("staff", $groups);
+		$this->assertContains("bureaucrat", $groups);
```
wladekb, line 54: How about some `notContains` (if that's the name) ?

tests/phpunit/includes/UserTest.php
```diff
@@ -17,90 +17,14 @@ class UserTest extends MediaWikiTestCase {
 	public function setUp() {
 		parent::setUp();
 
-		$this->savedGroupPermissions = $GLOBALS['wgGroupPermissions'];
```
wladekb: we don't care about files in `tests/phpunit`

extensions/wikia/PowerUser/PowerUser.class.php
wladekb: Any answer?

extensions/wikia/PowerUser/PowerUser.class.php
wladekb: Any answer?

includes/User.php
wladekb: That one hasn't been addressed yet.

includes/specials/SpecialListgrouprights.php
wladekb: Any answer?

includes/specials/SpecialStatistics.php
wladekb: Any answer?

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
wladekb: IMHO in such case we need a single source of truth (simple directory?) for the permission names so we can assure there will be no clash in the future when someone adds a new permission not knowing that such one already exists. Makes sense?

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
wladekb, line 31: Any answer?

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
wladekb, line 98: Any answer?

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
```diff
@@ -0,0 +1,391 @@
...
+
+	public function __construct() {
+		$this->permissionsConfiguration = new PermissionsConfigurationImpl();
+	}
+
+	public function getConfiguration() {
+		return $this->permissionsConfiguration;
+	}
+
+	public function getExplicitGroups(\User $user ) {
```
wladekb: you could declare
`use User;`
just after the namespace statement at the top of the file and then you can use `User` instead of `\User` in this file.

extensions/wikia/ContentReview/ContentReview.hooks.php
jcellary, line 357: So for this particular check, I talked to Adam and we agreed it's sufficient to specify that only members of content-reviewer group are allowed to assign this group (and skip the is staff check). So this is part of this definition:
https://github.com/Wikia/app/pull/9720/files#diff-8bf819f58bde706eaeeb9261498f35e5R384

extensions/wikia/PowerUser/PowerUser.class.php
jcellary: The answer is that I already converted it to the trait that you suggested :)

extensions/wikia/PowerUser/PowerUser.class.php
jcellary: I made the suggested change of unifying parameters - now all methods get the user objects instead of user id/user object. As for the proxy you suggested, I considered that, however part of methods take the user object for which we're performing a check, part take a user on which the operation is being performed and some both, so It wouldn't be possibly to do straightforward binding of the whole API to a single user object as its interpretation depends on the function context.

includes/User.php
jcellary: Ok, I can change that, but you could explain why using wgUser is bad? Just because we want to avoid globals in general is there some timing issue? I'm asking because I use wgUser in other places as well

includes/specials/SpecialListgrouprights.php
jcellary: Good point, actually this is already enforced, so I can just remove the checks

includes/specials/SpecialStatistics.php
jcellary: Yes, the answer is no.

...ok I replaced all in_array( getXXXgroups )

includes/specials/SpecialUserrights.php
jcellary: moved

lib/Wikia/src/Service/User/Permissions/PermissionsService.php
jcellary: done

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
jcellary: done

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
jcellary: done

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
jcellary: The single source of truth is this file. In future this file will be part of the user permissions service.

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
jcellary: same comment as above

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
jcellary: done, these are deleted now

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
jcellary, line 152: done

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
jcellary: done

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
jcellary: done

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
jcellary: done

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
jcellary, line 31: The underlying logic caching its result the only thing that is recalculated is array unique/merge. Does it really affect performance that much?

Btw. I believe I answered this comment in another spot, but can't find it now...

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
jcellary, line 98: see above

extensions/wikia/PowerUser/PowerUser.class.php
jcellary: done

includes/specials/SpecialUserrights.php
jcellary: done

lib/Wikia/src/Service/User/Permissions/PermissionsAccessor.php
jcellary: done

lib/Wikia/src/Service/User/Permissions/PermissionsDefinition.php
jcellary: done

lib/Wikia/src/Service/User/Permissions/PermissionsService.php
jcellary: done

lib/Wikia/src/Service/User/Permissions/PermissionsService.php
jcellary: done

lib/Wikia/src/Service/User/Permissions/PermissionsService.php
jcellary: I answered to that suggestion in your previous comment. The problem is that we can have the user as the operation performer, the user being changed or both. So we can't bind a single user object.

lib/Wikia/tests/Service/User/UserPermissionsIntegrationTest.php
jcellary, line 41: We can - then this won't be an integration test but a unit test. I think both is useful in general.

tests/phpunit/includes/UserTest.php
jcellary: What do you mean - they aren't used? So maybe we should just delete them, as otherwise other people like me will waste time on adjusting them.

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
jcellary: Ok I guess either way is fine (current solution is one line shorter :D ). So do we have a convention that we want to avoid '\' ? If yes I can change that, as I use it quite a bit in other places.

jcellary:
@wladekb I made changes according to your comments, let me know if you have any other suggestions

extensions/wikia/PowerUser/PowerUser.class.php
wladekb: That particular question was about the name of the variable...

includes/specials/SpecialUserrights.php
```diff
@@ -524,17 +553,14 @@ private function groupCheckboxes( $usergroups, $user ) {
 			$set = in_array( $group, $usergroups );
 			# Should the checkbox be disabled?
 			$disabled = !(
-				( $set && $this->canRemove( $group ) ) ||
-				( !$set && $this->canAdd( $group ) ) );
+				( $set && self::canRemove( $user, $group, $this->isself ) ) ||
```
wladekb: The canAdd() and canRemove() eventually call PermissionsService::getChangeableGroups() every time which in turn does a heavy array mangling.
We call canAdd()/canRemove() in a loop that goes through all the existing groups, I can see some room for optimizing for reducing cpu waste here

includes/specials/SpecialUserrights.php
```diff
@@ -21,18 +21,79 @@
...
+			$user->getUserPage(),
+			$reason,
+			array(
+				self::makeGroupNameListForLog( $oldGroups ),
+				self::makeGroupNameListForLog( $newGroups )
+			)
+		);
+	}
+
+	public static function isGlobalNonEditableGroup( $globalGroups, $group ) {
```
wladekb, line 68: The `$globalGroups` argument is a static list so we could fetch it ourselves instead of requesting callers to do that in every single place where this method is called

includes/specials/SpecialUserrights.php
```diff
@@ -392,7 +427,7 @@ function switchForm() {
 	 * @return Array:  Tuple of addable, then removable groups
 	 */
 	protected function splitGroups( $groups ) {
-		list( $addable, $removable, $addself, $removeself ) = array_values( $this->changeableGroups() );
+		list( $addable, $removable, $addself, $removeself ) = array_values( $this->getUser()->changeableGroups() );
```
wladekb, line 428: Any reason why you removed the method `changeableGroup()` and inlined its content in a few places in this file?

lib/Wikia/src/Service/User/Permissions/PermissionsConfiguration.php
```diff
@@ -0,0 +1,76 @@
...
+	 *
+	 * @param $groups string[] List of internal group names
+	 * @return string[] List of permission key names for given groups combined
+	 */
+	public function getGroupPermissions( $groups );
+
+	/**
+	 * Get all the groups who have a given permission
+	 *
+	 * @param $role String Role to check
```
wladekb: Apparently I don't have any idea what the "role" is. Could you expand the PHPDoc so there is more information about it?

lib/Wikia/src/Service/User/Permissions/PermissionsService.php
wladekb: You can still have a few fluent-ish entry points like `->user($user)` and `->as($user)` and so on but I can see you're not a great supporter of this idea. Then let's leave it as it is.

lib/Wikia/src/Service/User/Permissions/PermissionsService.php
```diff
@@ -0,0 +1,122 @@
...
+	 * Gets the object containing definitions of groups and permissions in the system
+	 * @return PermissionsConfiguration
+	 */
+	public function getConfiguration();
+
+	/**
+	 * @param \User $user
+	 * @return string[] Get list of global and local groups to which the given user has been explicitly assigned to.
+	 */
+	public function getExplicitGroups(\User $user );
```
wladekb: Space mismatch inside parens (no space at the beginning vs one at the end), autoformat this file?

lib/Wikia/src/Service/User/Permissions/PermissionsService.php
```diff
@@ -0,0 +1,122 @@
...
+	 */
+	public function isInGroup(\User $user, $group );
+
+	/**
+	 * Get the list of explicit and implicit group memberships this user has.
+	 * This includes all explicit groups, plus 'user' if logged in,
+	 * '*' for all accounts, and autopromoted groups
+	 * @param \User $user
+	 * @param $reCacheAutomaticGroups
+	 * @return Array of String internal group names
```
wladekb: you've been using "string[]" somewhere above, can we keep one consistent way to specify that?

lib/Wikia/src/Service/User/Permissions/PermissionsService.php
```diff
@@ -0,0 +1,122 @@
...
+	 */
+	public function getEffectiveGroups(\User $user, $reCacheAutomaticGroups = false );
```
wladekb: `$recache...` (small c), and the same a few lines below

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
```diff
@@ -0,0 +1,391 @@
...
+	private $globalExplicitUserGroups = [];
+
+	/** @var string[string] - key is the user id */
+	private $implicitUserGroups = [];
+
+	/** @var string[string] - key is user id */
+	private $userPermissions = [];
+
+	public function __construct() {
+		$this->permissionsConfiguration = new PermissionsConfigurationImpl();
```
wladekb, line 23: no injection here?

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
wladekb: I just surveyed one more person (hi @macbre ) and let's stay with the current usage of `\User`

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
```diff
@@ -0,0 +1,391 @@
...
+	}
+
+	public function getConfiguration() {
+		return $this->permissionsConfiguration;
+	}
+
+	public function getExplicitGroups(\User $user ) {
+		return array_unique( array_merge (
```
wladekb, line 31: coming back to "calculated every time", for sure `array_unique()` is O(n), I didn't check if that's somewhere called (directly or indirectly) in a loop but if yes that smells like a pure waste of cpu for me, anything we can do about it?

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
```diff
@@ -0,0 +1,391 @@
...
+			if ( !empty( $userId ) ) {
+				$this->userPermissions[ $userId ] = array_values( $permissions );
+			}
+		}
+		return $permissions;
+	}
+
+	private function loadLocalGroups( $userId ) {
+		if ( !empty( $userId ) && !isset( $this->localExplicitUserGroups[ $userId ] ) ) {
+			$dbr = wfGetDB( DB_MASTER );
```
wladekb, line 126: just wondering if that can also be cached in memcached as the global groups are, any opinion?

lib/Wikia/tests/Service/User/UserPermissionsIntegrationTest.php
wladekb, line 41: Ok, so let's stick with current implementation

tests/phpunit/includes/UserTest.php
wladekb: They aren't used. And yes, we should remove them from the codebase.

extensions/wikia/PowerUser/PowerUser.class.php
jcellary: Ok, this code is deleted now and move to the trait, so I'm assuming the issue is resolved

tests/phpunit/includes/UserTest.php
jcellary: ok deleted whole tests/phpunit folder

lib/Wikia/tests/Service/User/UserPermissionsIntegrationTest.php
jcellary, line 54: good point, I added some

includes/specials/SpecialUserrights.php
jcellary: You' right I moved the call to getChangeableGroups outside of the loop

includes/specials/SpecialUserrights.php
jcellary, line 68: That would require this function to not be static as user permissions from the trait is non static and hence can't be accessed. That is why I changed this method to get all required data in parameters

includes/specials/SpecialUserrights.php
jcellary, line 428: The whole code in changeableGroups in this class was a repetition of logic inside of the user class (which I moved to user permissions service), so I deleted the local copy and replaced it with a call to the central definition.

lib/Wikia/src/Service/User/Permissions/PermissionsConfiguration.php
jcellary: I missed that when moving the function - it should be called permission not role (AKA right)

lib/Wikia/src/Service/User/Permissions/PermissionsService.php
jcellary: done

lib/Wikia/src/Service/User/Permissions/PermissionsService.php
jcellary: yes

lib/Wikia/src/Service/User/Permissions/PermissionsService.php
jcellary: done

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
jcellary, line 23: Hmm I guess I could do it, although this is the actual implementation of configuration which ties to the implementation of user permission, so I'm not sure it's necessary here, as this is a pair of implementations.

When I convert it to the new service I'll have a shared data source for both, so I'm sure I'll be then sharing data between them and some internal mechanisms, so direct composition actually seems like the best solution

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
jcellary, line 31: I could store it in internal variable, similarly to the groups read from the DB but then I need to remember about purging it when the underlying data changes and I'm not convinced it's worth it. I would totally agree if this was a DB operation etc. but here I'm not convinced it's worth the extra code complexity, especially since in practice these arrays rarely have more then a few elements.

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
jcellary, line 126: Possibly. One thing I would keep in mind though is that this code will be deleted in a few weeks and replaced with an invocation of the user permissions service. So I would suggest to leave this implementation as is (as it's tested and working like that for years) and then we I make step 2 of moving it to the service, think about an appropriate caching strategy.

includes/specials/SpecialUserrights.php
wladekb, line 68: `$globalGroups` is part of a global configuration
I'd still prefer the method to take it even if it required sth like:
`Injector::getInjector()->get( PermissionsService::class )->....`
(due to the non-static trait issue you described), I'm assuming that implementation of`Injector::getInjector()->get( ... )` would cache the result and return the object very quickly

jcellary, line 68: Paweł told me that the injector is quite slow actually and should be replaced. I'm not sure about caching though.

jcellary:
Note unit tests are failing here, because this branch requires changes in config that are part of this PR: https://github.com/Wikia/config/pull/1620

wladekb:
lgtm

macbre:
```
1) Wikia\Service\User\Permissions\UserPermissionsIntegrationTest::testShouldReturnExplicitGroups
Failed asserting that an array contains 'bureaucrat'.

/var/lib/jenkins/workspace/unitTests_php_branches/app/lib/Wikia/tests/Service/User/UserPermissionsIntegrationTest.php:54
/var/lib/jenkins/workspace/unitTests_php_branches/app/tests/run-test.php:26

2) Wikia\Service\User\Permissions\UserPermissionsIntegrationTest::testShouldReturnEffectiveGroups
Failed asserting that an array contains 'bureaucrat'.

/var/lib/jenkins/workspace/unitTests_php_branches/app/lib/Wikia/tests/Service/User/UserPermissionsIntegrationTest.php:71
/var/lib/jenkins/workspace/unitTests_php_branches/app/tests/run-test.php:26
```

@jcellary , these changes broke tests suite in SJC (they do pass in Poznań).

PR #9725, opened 2016-02-16
RECIRC-31: Add Taboola Above Article (leaderboard) for Adblock traffic

pauloslund:
This requires adding

``` js
'TOP_LEADERBOARD_AB': {
  'recovery': ['XX'],
  'regular': []
}
```

to the wikifactory variable `wgAdDriverTaboolaConfig`


nandy-andy:
LGTM :japanese_ogre: 

Unfortunately, I could not check the php unit tests errors. But most of your changes were on front-end side and the phpUnit tests passes on my devbox when I'm on your branch :smiley_cat: 

Thanks for adding JS unit tests :+1: 

Just a general comment: once you're done with AB tests let us know to disable our AB test. I'd advocate also at that point to move the Taboola code outside the `AdEngine` extension.

PR #9728, opened 2016-02-16
Remove the dimension test for the rendered image from vignette from the ImageServingCroppingTest

drsnyder:
Remove the dimension test for the rendered image from vignette from the ImageServingCroppingTest. 

We have experimented with integration / validation testing in vignette in the past. We should revisit that issue and add some tests to confirm that vignette behaves as advertised. 

The ticket owner will determine where is the best place to run this kind of test.

https://wikia-inc.atlassian.net/browse/SERVICES-1168

@Wikia/platform

### DIFF
--- a/extensions/wikia/ImageServing/tests/ImageServingCroppingTest.php
+++ b/extensions/wikia/ImageServing/tests/ImageServingCroppingTest.php
@@ -9,17 +9,14 @@ class ImageServingCroppingTest extends WikiaBaseTest {
 
 	const FILE_NAME = 'Wiki-wordmark.png';
 
-	private $tmpFile;
-
 	public function setUp() {
 		$this->setupFile =  __DIR__ . '/../imageServing.setup.php';
 		parent::setUp();
 
 		$this->mockGlobalVariable('wgEnableVignette', true);
-		$this->mockGlobalVariable('wgUploadPath', 'http://images.wikia.com/firefly/images');
 	}
 
-	public function testCropping() {
+	public function testCroppingPath() {
 		// requested crop size is 50 x 50
 		$im = new ImageServing(null, 50);
 		$file = wfFindFile(self::FILE_NAME);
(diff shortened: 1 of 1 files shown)


pawelchcki:
I think this test is useful as it tests the integration of MW vignette client with vignette. I would really appreciate if you would clarify why do you believe that this test shouldn't be run on produciton.

drsnyder:
>  I would really appreciate if you would clarify why do you believe that this test shouldn't be run on production.

Just to be clear, I didn't say it shouldn't be run on production. I think it should NOT be run against the dev instance of vignette. 

When you open a PR on the app repo the tests run against the dev instance of vignette which I think is not the right place for this.

pawelchcki:
> > I would really appreciate if you would clarify why do you believe that this test shouldn't be run on production.
> 
> Just to be clear, I didn't say it shouldn't be run on production. I think it should NOT be run against the dev instance of vignette.

Ahh sorry I 'meant dev its getting late here. :)

Do we have similar integration tests that run e.g. against some pandora service ?

drsnyder:
> Do we have similar integration tests that run e.g. against some pandora service ?

Not that I'm aware of. We have been talking about it though.

In general, I think it would be great if we can catch these kinds of faults before users (other developers) do in their pull requests. Unless you are familiar with the service it's difficult to retrace the connections that may have caused a test to fail.

drsnyder:
Any objections to merging?

pawelchcki:
While this test could certainly be improved. I would like to avoid situation where we do not have any integration tests for checking vignette-mw cooperation.

E.g. here is a example of integration tests for phalanx: 
https://github.com/Wikia/app/blob/8265d1f84d51914cb63af44876baea7ae4b06403/extensions/wikia/PhalanxII/tests/PhalanxServiceTest.php

I completely agree that dev vignette fail should be caught by external monitoring. But I still think this test is valuable while maybe not perfect.

drsnyder:
Ok, we can wait until there's more progress on SERVICES-1168

PR #9731, opened 2016-02-17
AvatarsMigrator: handle --dfs-dc=res

macbre:
[PLATFORM-1842](https://wikia-inc.atlassian.net/browse/PLATFORM-1842)

This allows us to perform re-migration of avatars using DFS cluster in Reston.

@jcellary 

P.S. This was already fire tested on cron-s5, nothing exploded :)


macbre:
Failed integration test not related to the change:

```
HubsRssIntegrationTest::testHubsRssLinks with data set #2 ('Entertainment', false)
Cannot read RSS file from Entertainment hub
Failed asserting that false is not equal to false.`
```

PR #9734, opened 2016-02-17
Sus 157 add unit tests

sqreek:
add unit tests for getNotificationMessage.
notice it's not checking is mock works correctly cause msgId is build [this way](https://github.com/Wikia/app/blob/7ee047b1d27e5f24d156770323eba6bfd8a51671/extensions/wikia/WallNotifications/WallNotificationsController.class.php#L252-L252)

@Wikia/sustaining-team please add your opion about this test.

I create another PR for that cause I'm feel this test will not prevent us from `copy-paste` messages in i18n, and there problem with wrong message was


ghost:
:+1: LGTM

extensions/wikia/WallNotifications/tests/WallNotificationsControllerTest.php
```diff
@@ -29,7 +31,59 @@ public function testIndex( $isUserLoggedIn, $isUserAllowed, $wgAtCreateNewWikiPa
...
+			[
+				'parent_id' => 1245,
+				'notify' => [ 'grouped' => self::getNotifyGroupMock( $notifyGrouped ) ]
+			];
+
+		$this->mockMessage( $expectedMessageId, $expectedMessageTemplate );
+
+		//when
+		$resp = $this->app->sendRequest( 'WallNotificationsController', 'Notification', $params );
+		var_dump( $resp->getVal( 'debug' ) );
```
kamilkoterba: What this debug tells?
if we want to see it in the output maybe some caption would be welcomed here.
BTW, is it really a way we solve adding debug info to unit tests output?

kamilkoterba:
LGTM

extensions/wikia/WallNotifications/tests/WallNotificationsControllerTest.php
ghost: Seems some leftover. Let's remove it.

sqreek: This is not a var_dump you are looking for. Go away. There is nothing to see here.

PR #9737, opened 2016-02-17
SEO-185 Small TwitterCard updates

gabrys:
- Use wfShortenText for trimming the strings
- Fallback to the HTML title for the twitter:description field
- Use the HTML title for the twitter:title fields


saipetch:
LGTM

PR #9738, opened 2016-02-18
Http::get - use noProxy for external requests

macbre:
[PLATFORM-1856](https://wikia-inc.atlassian.net/browse/PLATFORM-1856)

As we're planning to get rid of local Varnish instance running on Apache nodes, we want to use  border nodes as HTTP proxies. However, we do not want to use them when making external HTTP requests - hence this change

Introduce `ExternalHttp` helper class with `get` and `post` methods that will set `noProxy` option to true internally and clearly state what kind of HTTP request we want to perform here. In the near future we'll add more headers for tracing internal HTTP requests.

@wladekb / @artursitarski

### DIFF
--- a/extensions/3rdparty/LyricWiki/LyricFind/LyricFindTrackingService.class.php
+++ b/extensions/3rdparty/LyricWiki/LyricFind/LyricFindTrackingService.class.php
@@ -122,10 +122,7 @@ class LyricFindTrackingService extends WikiaService {
 
 		wfDebug(__METHOD__ . ': ' . json_encode($data) . "\n");
 
-		$resp = Http::post($url, [
-			'postData' => $data,
-			'noProxy' => true
-		]);
+		$resp = ExternalHttp::post($url, ['postData' => $data]);
 
 		if ($resp !== false) {
 			wfDebug(__METHOD__ . ": API response - {$resp}\n");
@@ -204,10 +201,7 @@ class LyricFindTrackingService extends WikiaService {
 
 		wfDebug(__METHOD__ . ': ' . json_encode($data) . "\n");
 
-		$resp = Http::post($url, [
-			'postData' => $data,
-			'noProxy' => true
-		]);
+		$resp = ExternalHttp::post($url, ['postData' => $data]);
 
 		if ($resp !== false) {
 			wfDebug(__METHOD__ . ": API response - {$resp}\n");
--- a/extensions/3rdparty/LyricWiki/LyricFind/LyricFindTrackingService.class.php
+++ b/extensions/3rdparty/LyricWiki/LyricFind/LyricFindTrackingService.class.php
@@ -122,7 +122,10 @@ class LyricFindTrackingService extends WikiaService {
 
 		wfDebug(__METHOD__ . ': ' . json_encode($data) . "\n");
 
-		$resp = Http::post($url, ['postData' => $data]);
+		$resp = Http::post($url, [
+			'postData' => $data,
+			'noProxy' => true
+		]);
 
 		if ($resp !== false) {
 			wfDebug(__METHOD__ . ": API response - {$resp}\n");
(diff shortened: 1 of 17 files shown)


extensions/wikia/Recirculation/services/FandomDataService.class.php
```diff
@@ -55,7 +55,7 @@ private function apiRequest( $type ) {
 		}
 
 		$url = $this->buildUrl( $endpoint, $options );
-		$data = Http::get( $url );
+		$data = Http::get( $url, null, [ 'noProxy' => true ] );
```
macbre: Pinging @pauloslund

extensions/3rdparty/LyricWiki/LyricFind/LyricFindTrackingService.class.php
```diff
@@ -122,7 +122,10 @@ public function track($amgId, $gracenoteId, Title $title) {
 
 		wfDebug(__METHOD__ . ': ' . json_encode($data) . "\n");
 
-		$resp = Http::post($url, ['postData' => $data]);
+		$resp = Http::post($url, [
```
wladekb: I wonder about creating something like `ExternalHttp::post` which would also help us when we get to implementing HTTP headers for request tracing, what do you think? or maybe leave `Http` as it is and introduce `InternalHttp` ?

macbre: We perform more internal requests than external ones (by both number of requests and places in the code that make them). I was thinking about introducing `ExternalHttp` helper class with `get` and `post` methods. With the "HTTP headers for request tracing" we have a second use case for this helper class. Let's implement it.

macbre: Introduced in 6b045ed

macbre:
```
LyricFindTrackingServiceTest::testTrackResponseCode with data set #4 (1234, array(array(206)), true)
API response code should match expected value
Failed asserting that false matches expected true.
```

wladekb:
lgtm

extensions/wikia/AdEngine/ResourceLoaders/ResourceLoaderAdEngineBase.php
```diff
@@ -23,7 +23,7 @@
 	 * @return bool|MWHttpRequest|string
 	 */
 	protected function fetchRemoteScript( $url ) {
-		return Http::get( $url, 'default', [ 'noProxy' => true ] );
+		return ExternalHttp::get( $url );
```
macbre, line 26: @nandy-andy - please review

extensions/wikia/AdEngine/ResourceLoaders/ResourceLoaderAdEngineSourcePointRecoveryModule.php
```diff
@@ -41,11 +41,10 @@ protected function getFallbackDataWhenRequestFails() {
 	protected function fetchRemoteScript( $url ) {
 		global $wgSourcePointApiKey;
 
-		$content = Http::get( $url,
+		$content = ExternalHttp::get( $url,
```
macbre, line 44: @nandy-andy - please review

extensions/wikia/Recirculation/services/FandomDataService.class.php
```diff
@@ -55,7 +55,7 @@ private function apiRequest( $type ) {
 		}
 
 		$url = $this->buildUrl( $endpoint, $options );
-		$data = Http::get( $url );
+		$data = ExternalHttp::get( $url );
```
macbre, line 58: @pauloslund

nandy-andy:
It looks good. However, I'm surprised there is no files from `WikiaMaps` extension. I'll double-check tomorrow and get back to it.

macbre:
I did not replace calls to `Http::get` with `noProxy` option set that DID not perform external requests. This was the case mostly for internal services calls (to bypass the local Varnish instance causing problems).

nandy-andy:
Oh, ok. Thanks for update! :) Therefore, it looks good to me :+1:

gabrys:
Why not doing this the other way round. I assume connecting to an internal HTTP service is a special case, not connecting to an external one

extensions/wikia/Recirculation/services/FandomDataService.class.php
pauloslund, line 58: :+1:

macbre:
@rychu, external traffic is actually something quite extraordinary in our case.

PR #9739, opened 2016-02-18
SUS-180 In WallNotification remove strong typing

sqreek:
It seems php have some issue with strong typing here.
This should fix tests, and I will take some time to investigate why this is not working
//cc @Wikia/sustaining-team


kamilkoterba:
AFAIK php doesn't support strong typing for simple types. Only arrays and objects.

LGTM

ghost:
:+1: LGTM

PR #9747, opened 2016-02-18
Removed wgRevokePermissions

jcellary:
Another cleanup related to user permissions - the variable wgRevokePermissions was set only for vstf and forum. I confirmed with comsup, that we don't need that. Also this variable doesn't seem to be set in wiki factory DB in any local wiki.

Hence I'm deleting it completely to have less logic to port to user permissions service.


macbre:
LGTM

PR #9748, opened 2016-02-18
ADEN-2998 Monobook spotlights

nandy-andy:
We don't have `InstantGlobals` in Monobook and we started serving a JS error on this skin. Changes below fixes those error and also remove spotlights assets from Monobook.

### DIFF
--- a/extensions/wikia/AssetsManager/config.php
+++ b/extensions/wikia/AssetsManager/config.php
@@ -1020,7 +1020,6 @@ $config['monobook_js'] = array(
 		'//resources/wikia/modules/lazyqueue.js',
 		'//extensions/wikia/JSMessages/js/JSMessages.js',
 		'#group_imglzy_js',
-		'#group_spotlights_js',
 		'//resources/wikia/libraries/ghostwriter/gw.min.js',
 		'//skins/shared/scripts/onScroll.js',
 		'//extensions/wikia/VideoHandlers/js/VideoBootstrap.js',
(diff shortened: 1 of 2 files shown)


drets:
LGTM :+1: 

One question: maybe is it better to call `window.Wikia.geo.isProperGeo(window.Wikia.InstantGlobals.wgReviveSpotlightsCountries);` in `try...catch` statement?

nandy-andy:
It will simplify the logic. Let's try it.

jacekjursza:
@nandy-andy, @drets 
http://stackoverflow.com/questions/2631001/javascript-test-for-existence-of-nested-object-key

there might be some issues with try.catch approach:
"I don't think try/catch is a good way to test for existence of an object: try/catch is meant to handle exceptions, not normal conditions such as the test here. I think (typeof foo == "undefined") at each step is better -- and in general, there's probably some refactoring required if you're working with such deeply nested properties. Also, try/catch will cause a break in Firebug (and in any browser where break-on-error is turned on) if an exception is thrown. – Sam Dutton Nov 9 '10 at 12:00" 
- it's old post, but still it would be worth to check how different browser behave

drets:
FYI browser support: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch#Browser_compatibility

nandy-andy:
And: https://github.com/Fyrd/caniuse/issues/2102 :) But I've checked it in my Firefox with Firebug when changing `wgReviveSpotlightsCountries` to `wgReviveSpotlightsCountriesss` and it didn't break the page nor Firebug.

We'll definitely refactor it within [ADEN-2828](https://wikia-inc.atlassian.net/browse/ADEN-2828). So, let's merge it and push the fix.

PR #9752, opened 2016-02-18
SUS-208 in wall return all replies at once

sqreek:
Ticket: [SUS-208](https://wikia-inc.atlassian.net/browse/SUS-208)

Changed:
All replies for a thread is loaded by one query.
Remove unnecessery one database call for every thread

//cc @Wikia/sustaining-team


extensions/wikia/Wall/WallMessage.class.php
```diff
@@ -61,6 +61,37 @@ static public function newFromId( $id, $master = false ) {
...
+	static public function newFromIds( $ids ) {
+		wfProfileIn( __METHOD__ );
+
+		$titles = Title::newFromIDs( $ids );
+		$wallMessages = [ ];
+		$correctIds = [ ];
+
+		//double check if all titles are correct
+		foreach ( $titles as $title ) {
+			if ( $title instanceof Title && $title->exists() ) {
```
ghost: Calling `Title::newFromIds()` we expect `$titles` to be an array of `Title`s. Therefore the `instanceof` check is not needed or we should log cases when an item is not a `Title`. I would even go with assertions but it might be too aggressive at the moment. So please, log non-Titles as errors with label `SUS-165` to Kibana.

extensions/wikia/Wall/WallMessage.class.php
```diff
@@ -61,6 +61,37 @@ static public function newFromId( $id, $master = false ) {
...
+				$wallMessages[] = WallMessage::newFromTitle( $title );
+				$correctIds[] = $title->mArticleID;
+			}
+		}
+
+		$retryIds = array_diff( $ids, $correctIds );
+		foreach ( $retryIds as $id ) {
+			$title = Title::newFromId( $id, Title::GAID_FOR_UPDATE );
+			if ( $title instanceof Title && $title->exists() ) {
```
ghost, line 88: https://github.com/Wikia/app/pull/9752/files#r53467909 - the same thing here.

extensions/wikia/Wall/WallMessage.class.php
sqreek: I will leave this place without log, and remove instance of check. In this place we should never have empty title.

extensions/wikia/Wall/WallMessage.class.php
sqreek, line 88: added logging here

extensions/wikia/Wall/WallMessage.class.php
```diff
@@ -61,6 +63,39 @@ static public function newFromId( $id, $master = false ) {
...
+
+		$titles = Title::newFromIDs( $ids );
+		$wallMessages = [ ];
+		$correctIds = [ ];
+
+		//double check if all titles are correct
+		foreach ( $titles as $title ) {
+			if ( $title->exists() ) {
+				$wallMessages[] = WallMessage::newFromTitle( $title );
+				$correctIds[] = $title->mArticleID;
```
idradm: Could you use getter instead of direct access?

extensions/wikia/Wall/WallThread.class.php
```diff
@@ -110,11 +112,11 @@ private function getReplyIdsFromDB( $dbr, $afterId = null ) {
 				$list[] = $oRow->comment_id;
 			} );
 
-		$lastId = end( $list );
-
-		// TODO: add pagination or a limit here because this will return all replies
-		return empty( $lastId ) ? $list :
-			array_merge( $list, $this->getReplyIdsFromDB( $dbr, $lastId ) );
+		if ( count( $list ) < self::FETCHED_REPLIES_LIMIT ) {
```
idradm: Why not short if notation?

ghost:
:+1: LGTM

idradm:
What about some tests?

ghost:
As for tests, I think it makes sense to start small and _very_ simple. It's still pretty hard to say what does the code do and whether it is correct or not. We can provide test coverage for parts we know for sure.

idradm:
LGTM

PR #9754, opened 2016-02-18
Add logging to track down when/where bad avatar values are saved

jsutterfield:


lib/Wikia/src/Service/User/Attributes/UserAttributes.php
```diff
@@ -125,6 +128,20 @@ public function save( $userId ) {
...
 	}
 
+	private function logIfBadAvatarVal( $value, $userId ) {
+
+		if ( $value == "" || preg_match( '/^http/', $value ) ) {
+			return;
+		}
+
+		$e = new \Exception;
+		$this->info( 'USER_ATTRIBUTES saving_bad_avatar_val', [
```
macbre: Please log it as an `error`, so that we can get these automatically reported as Jira issue.

And you can simply replace `trace` entry in the context with `'exception' => new \Exception()`. Monolog formatter will do the job for you and return a backtrace.

macbre:
:+1:

macbre:
Thanks!

PR #9755, opened 2016-02-19
XW-1120 | extend UserApi with poweruser types

tomasznapieralski:
https://wikia-inc.atlassian.net/browse/XW-1120

/cc @Wikia/x-wing

### DIFF
--- a/includes/wikia/api/UserApiController.class.php
+++ b/includes/wikia/api/UserApiController.class.php
@@ -24,10 +24,12 @@ class UserApiController extends WikiaApiController {
 	public function getDetails() {
 		wfProfileIn( __METHOD__ );
 		$ids =  $this->request->getVal( 'ids' );
+
 		if ( empty( $ids ) ) {
 			wfProfileOut( __METHOD__ );
 			throw new InvalidParameterApiException( 'ids' );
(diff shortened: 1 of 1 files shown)


includes/wikia/api/UserApiController.class.php
```diff
@@ -35,9 +37,17 @@ public function getDetails() {
 		$ids = array_slice( $ids, 0, static::USER_LIMIT );
 		//users are cached inside the service
 		$users = (new UserService())->getUsers( $ids );
+
 		$items = array();
+
 		foreach ( $users as $user ) {
 			$userName = $user->getName();
+			$powerUserTypes = [];
+			$powerUserTypesRaw = ( new \Wikia\PowerUser\PowerUser( $users[0] ) )->getTypesForUser();
```
hakubo: 1. `( new \Wikia\PowerUser\PowerUser( $user ) )->getTypesForUser();` as otherwise you would always send data for first user
2. Can we just return ( new \Wikia\PowerUser\PowerUser( $user) )->getTypesForUser(); as I get it it returns something like:
   ['type1', 'type2']

I think we don't have to do anything with it

PR #5762, opened 2014-12-12
PLATFORM-654: get data from ResourceTiming

macbre:
Aggregate it and send it via Bucky on:
- DOMReady
- window.load
- 5 seconds after window.load

[Supported desktop browsers](http://caniuse.com/#feat=resource-timing): IE10+, Fx 34+, Chrome 31+ and Opera 26+

@Wikia/platform-team

PR #9758, opened 2016-02-19
bucky_resources_timing: report per-domain performance stats for 3rd party assets

macbre:
When investigating the front-end performance regression in [PLATFORM-1903](https://wikia-inc.atlassian.net/browse/PLATFORM-1903) it turned out that we do not have detailed information on 3rd party assets that we fetch on each page load.

This PR extends Bucky collector of ResourceTiming stats (introduced in #5762). Every request to 3rd party asset (with Bucky sampling applied) will now be reported:

```
ResourceTiming:  domain-w.visualdna.com.count: 1
ResourceTiming:  domain-w.visualdna.com.time: 181
ResourceTiming:  domain-ev.visualdna.com.count: 2
ResourceTiming:  domain-ev.visualdna.com.time: 933
ResourceTiming:  domain-segs.btrll.com.count: 10
ResourceTiming:  domain-segs.btrll.com.time: 2222
ResourceTiming:  domain-maps.googleapis.com.count: 4
ResourceTiming:  domain-maps.googleapis.com.time: 80
```

This data will be available in the same InfiuxDB time series as the rest of stats from ResourceTiming (namely `Mediawiki_App_Rum_ResourceTiming_60m`). Hopefully this data will allow us to identify the slowest 3rd party assets - see the ticket linked above for more details (graphs included).

@wladekb / @nandy-andy (pinging you as this can be useful from ads serving perspective)


extensions/wikia/Bucky/js/spec/bucky_resources_timing.spec.js
```diff
@@ -34,6 +34,22 @@ describe('BuckyResourcesTiming', function () {
 		});
 	});
 
+	it('correctly get the domain from URL', function() {
```
wladekb, line 37: how about test for a URL that wouldn't be valid so we make sure this will be handled properly and doesn't break anything else?

wladekb:
lgtm, left 1 comment only

extensions/wikia/Bucky/js/spec/bucky_resources_timing.spec.js
macbre, line 37: Good point, addressed in f6d18a7

PR #9760, opened 2016-02-19
SERVICES-1163: Added consul tags to Phalanx service address.

harnash:
Better and final solution would be to use swagger client but that is outside of the scope of this quickfix.

ping: @macbre


macbre:
LGTM, let's wait for tests to pass.

PR #9765, opened 2016-02-21
SUS-205 VideoHandlerController::getVideoList() prone to fatals on invalid parameters

ghost:
# SUS-205 VideoHandlerController::getVideoList() prone to fatals on invalid parameters
## Links
- https://wikia-inc.atlassian.net/browse/SUS-205
## Description

The `limit` parameter is specified in the URL and was not handled properly. This change fixes it in the following way:
- non-integer input is converted to integer as PHP does it
- 0 means no limit, queries with maximum limit are slow, so 0 is replaced with 1
- maximum limit is honored
## Reviewers

@macbre @Grunny @Wikia/sustaining-team


Grunny:
LGTM.

macbre:
:+1:

PR #9769, opened 2016-02-22
SUS-205 Upgrade FluentSQL library

ghost:
# SUS-205 Upgrade FluentSQL library
## Links
- https://wikia-inc.atlassian.net/browse/SUS-205
- Wikia/fluent-sql-php#5
## Description

The `limit` parameter is specified in the URL and was not handled properly. This change fixes it in the following way:
- non-integer input is converted to integer as PHP does it
- 0 means no limit, queries with maximum limit are slow, so 0 is replaced with 1
- maximum limit is honored

`LIMIT()` and `OFFSET()` are prone to fatal errors in non-integer input. Considering that
parameters for those are often taken from URLs, this is happening quite often and we don't want unhandled errors.

I suggest to allow only numeric values by default. If a developers need a value of different type - for example a string representig a subquery - they must state is explicitly by using `FORCE_NON_NUMERIC_LIMIT` and `FORCE_NON_NUMERIC_OFFSET` respectively as the second parameter.

```
LIMIT( '(SELECT COUNT(1) FROM user)', \FluentSql\SQL::FORCE_NON_NUMERIC_LIMIT )
```

I was not sure how MySQL-specific this code can be. I assumed as agnostic as possible. MySQL does not allow non-integer LIMIT and OFFSET apart from prepared statements and stored programs. I believe other engines might allow using functions or subqueries.

I was thinking about providing support for alternative LIMIT/OFFSET syntax, `LIMIT 5, 10` in `LIMIT()` which is now only allowed with `\FluentSql\SQL::FORCE_NON_NUMERIC_LIMIT`, but there was no such usage cases in Wikia code.
## Reviewers

@Grunny @Wikia/sustaining-team

### DIFF
--- a/lib/composer/wikia/fluent-sql-php/src/SQL.php
+++ b/lib/composer/wikia/fluent-sql-php/src/SQL.php
@@ -15,6 +15,13 @@ use FluentSql\Exception\SqlException;
 use FluentSql\Functions\SqlFunction;
 
 class SQL {
+
+	/** @const FORCE_NON_NUMERIC_LIMIT whether to force using non numeric LIMIT clause, e.g. a subquery */
+	const FORCE_NON_NUMERIC_LIMIT = true;
+
+	/** @const FORCE_NON_NUMERIC_OFFSET whether to force using non numeric OFFSET clause, e.g. a subquery */
+	const FORCE_NON_NUMERIC_OFFSET = true;
+
 	/**
 	 * order in which relevant functions are called so future calls can modify them
 	 * (like ->EQUAL_TO() modifying ->WHERE())
@@ -871,14 +878,14 @@ class SQL {
 	 * @param $limit
 	 * @return SQL
 	 */
-	public function LIMIT($limit) {
-		$this->limit = new Clause\Limit($limit);
+	public function LIMIT($limit, $allowNonNumeric = false) {
+		$this->limit = new Clause\Limit($limit, $allowNonNumeric);
 
 		return $this->called($this->limit);
 	}
 
-	public function OFFSET($offset) {
-		$this->offset = new Clause\Offset($offset);
+	public function OFFSET($offset, $allowNonNumeric = false) {
+		$this->offset = new Clause\Offset($offset, $allowNonNumeric);
 
 		return $this->called($this->offset);
 	}
@@ -1457,8 +1464,6 @@ class SQL {
 		}
 
 		if ($autoIterate) {
-			\Wikia\Util\Assert::true( $result instanceof \ResultWrapper, 'FluentSQL::query - did not get ResultWrapper - instead got ' . get_class($result) );
-
 			$data = $this->autoIterate($result, $callback);
 		} else {
 			$data = $callback($result);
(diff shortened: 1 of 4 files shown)


lib/composer/wikia/fluent-sql-php/tests/FluentSql/NonNumericLimitOffsetTest.php
```diff
@@ -0,0 +1,69 @@
...
+ */
+
+namespace FluentSql;
+
+class NonNumericLimitOffsetTest extends FluentSqlTestBase {
+
+	/**
+	 * numeric LIMIT
+	 */
+	public function test1() {
```
idradm, line 15: Is this test naming a convention in the lib? Maybe those could be more descriptive?

lib/composer/wikia/fluent-sql-php/src/SQL.php
```diff
@@ -871,14 +878,14 @@ public function HAVING($column) {
 	 * @param $limit
 	 * @return SQL
 	 */
-	public function LIMIT($limit) {
-		$this->limit = new Clause\Limit($limit);
+	public function LIMIT($limit, $allowNonNumeric = false) {
```
idradm, line 881: Should we enable non numeric values in limit at all? Is it used anywhere as such?

lib/composer/wikia/fluent-sql-php/tests/FluentSql/NonNumericLimitOffsetTest.php
sqreek, line 15: :+1:

lib/composer/wikia/fluent-sql-php/src/SQL.php
ghost, line 881: See: https://github.com/Wikia/fluent-sql-php/pull/5#discussion_r53674974

lib/composer/wikia/fluent-sql-php/tests/FluentSql/NonNumericLimitOffsetTest.php
ghost, line 15: That was the naming scheme used here https://github.com/Wikia/fluent-sql-php/blob/master/tests/FluentSql/ComplexQueryTest.php - but yeah, names could be more descriptive. This needs to be addressed as a contribution to https://github.com/Wikia/fluent-sql-php

macbre:
As we're getting this code from a separate git repository, it makes sense to use composer to get this library instead of manual copy & paste of files.

Add we do have an entry for `"wikia/fluent-sql-php": "1.0.*",` in `/composer.json`

sqreek:
or git submodule :wink:

ghost:
I agree. The only problem is that Wikia changes have to be applied. I mean changes that introduce dependencies on Wikia/app - e.g. WikiaLogger. So the process is: update the code in /lib/ and reapply Wikia-specific changes.

sqreek:
:+1:

PR #9770, opened 2016-02-22
UnusedTemplatesHandler: do not use transaction to wrap just a single query

macbre:

Do not use transactions to wrap a single query. This affects replication when adding many rows inside a transaction.

@adamkarminski / @kamilkoterba / @drozdo


macbre:
Jenkins fails with the following:

```
ERROR: Step ‘Archive the artifacts’ failed: no workspace for unitTests_php_branches #27813
```

adamkarminski:
LGTM, thanks!

PR #9771, opened 2016-02-22
AllinfoboxesQueryPage: remove transaction wrapping DELETE + INSERT queries

macbre:

This transaction affects replication as it deletes and then inserts many rows. See the ticket for more details.

@idradm  / @drozdo


idradm:
LGTM

PR #9773, opened 2016-02-22
SpecialUpload: only validate the token when we have one

macbre:
[PLATFORM-1881](https://wikia-inc.atlassian.net/browse/PLATFORM-1881)

99% of `hash_equals(): Expected user_string to be a string, null given` warnings came from `SpecialUpload` class that tries to validate the token on the first requests to `Special:Upload`. This obviously makes no sense - add a check for not null token before validating it.

@wladekb

### DIFF
--- a/includes/specials/SpecialUpload.php
+++ b/includes/specials/SpecialUpload.php
@@ -110,7 +110,7 @@ class SpecialUpload extends SpecialPage {
 
 		// If it was posted check for the token (no remote POST'ing with user credentials)
 		$token = $request->getVal( 'wpEditToken' );
-		$this->mTokenOk = !is_null( $token ) && $this->getUser()->matchEditToken( $token ); # Wikia change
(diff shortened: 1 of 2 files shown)


includes/specials/SpecialUpload.php
```diff
@@ -110,7 +110,7 @@ protected function loadRequest() {
 
 		// If it was posted check for the token (no remote POST'ing with user credentials)
 		$token = $request->getVal( 'wpEditToken' );
-		$this->mTokenOk = $this->getUser()->matchEditToken( $token );
+		$this->mTokenOk = !is_null( $token ) && $this->getUser()->matchEditToken( $token ); # Wikia change
```
wladekb: I'd add this check just before call to `hash_equals()` and return `false`.
We'd get more cases covered instead this one.
Sounds reasonable?

macbre: Added `!is_null` checks before both calls to `hash_equals` in User class - 8b705d8

PR #9774, opened 2016-02-22
ADEN-2976 Change floating medrec loc param value

fraszczakszymon:
Use `loc=hivi` instead of `loc=middle` for FMR slot.


nandy-andy:
LGTM :+1:

PR #9775, opened 2016-02-22
XW-1153 Remove redundant check from MercuryApi

rogatty:
https://wikia-inc.atlassian.net/browse/XW-1153

`$redirectTargetTitle` shouldn't ever be a `null` because we check if `$title->isRedirect()` before calling `handleRedirect()` and `Article::getRedirectTarget()` returns `null` if `!$title->isRedirect()`.

Also, I've tested it on article with:

```
#REDIRECT [[Special:NonExistent]]
```

(http://concf.wikia.com/wiki/Redirect_empty_target) and it seems to work just fine.


vforge:
LGTM

hakubo:
LGTM

PR #9776, opened 2016-02-22
RECIRC-52: Add experiment name and group to tracking event label

pauloslund:


extensions/wikia/Recirculation/js/spec/tracker.spec.js
```diff
@@ -0,0 +1,60 @@
+/*global beforeEach, describe, it, modules, expect, spyOn*/
+describe('ext.wikia.recirculation.tracker', function () {
+	'use strict';
+	function noop() {}
+
+	var experimentName = 'RECIRCULATION',
+		defaults = {
```
gabrys: You don't need this var now :-)

gabrys:
LGTM

PR #9778, opened 2016-02-22
SEO-288 Include the Special:Images in the My Tools selection

gabrys:
My Tools selection lists the selected special pages using their
description (getDescription) as the label for the translation.

We're adding a translation of 'images' label to read it from
'wikianewfiles-title'.

This makes the "Images" appear in the selection of the tools suggested
for users.

I also updated the content of the following help page:

http://community.wikia.com/wiki/Help:Toolbar/List_of_available_tools

Help in other languages should be also updated.


saipetch:
LGTM

PR #9782, opened 2016-02-23
(INT-362) Dismissing WikiaInYourLang should dismiss it for all pages

Grunny:
Dismissing WikiaInYourLang on one page should dismiss it for all pages on a
wiki.

/cc @adamkarminski


adamkarminski:
LGTM

PR #9783, opened 2016-02-23
ADEN-3011 Enable SRA in Evolve2 provider

fraszczakszymon:
Enable Single Request Architecture in Evolve2 provider.


nandy-andy:
Code changes LGTM but somehow one Travis build failed. Can you investigate it, please? It seems `npm` failed (not connected to our changes) but if it reproducible easily we'll probably have to inform somebody.

PR #9784, opened 2016-02-23
CE-3295: Update recent changes API to return upvotes for revisions

lukaszkonieczny:
We added possibility to upvote for user edit/change. We want to get data about upvotes for each revision on Recent Wiki Activity on mobile, so we updated existing API to meet out requirements.

### DIFF
--- a/includes/api/ApiQueryRecentChanges.php
+++ b/includes/api/ApiQueryRecentChanges.php
@@ -40,7 +40,7 @@ class ApiQueryRecentChanges extends ApiQueryGeneratorBase {
 			$fld_user = false, $fld_userid = false, $fld_useravatar = false,
 			$fld_flags = false, $fld_timestamp = false, $fld_title = false, $fld_ids = false,
 			$fld_sizes = false, $fld_redirect = false, $fld_patrolled = false,
-			$fld_loginfo = false, $fld_tags = false, $token = array();
+			$fld_loginfo = false, $fld_tags = false, $fld_upvotes, $token = array();
 
 	private $tokenFunctions;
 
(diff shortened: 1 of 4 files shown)


includes/api/ApiQueryRecentChanges.php
```diff
@@ -525,6 +531,27 @@ private function parseRCType( $type ) {
 		}
 	}
 
+	private function combineAdditionalData( ApiResult $result, array $revisionsIds ) {
```
Grunny: Do we expect to combine more data? It's minor, but just thinking whether this would be a little better as `addUpvoteData` or similar to be specific as to what it does at this time, and move the check to the `run` method, i.e.

``` php
if ( this->fld_upvotes ) {
    $this->addUpvoteData( $result, $revisionIds );
}
```

includes/api/ApiQueryRecentChanges.php
```diff
@@ -525,6 +531,27 @@ private function parseRCType( $type ) {
...
+		global $wgCityId;
+
+		if ( $this->fld_upvotes ) {
+			$upvotes = ( new RevisionUpvotesService() )->getRevisionsUpvotes( $wgCityId, $revisionsIds );
+			foreach( $revisionsIds as $index => $id ) {
```
Grunny: Minor: space after `foreach`.

includes/api/ApiQueryRecentChanges.php
```diff
@@ -525,6 +531,27 @@ private function parseRCType( $type ) {
...
+				$upvote = [];
+				$upvoteCount = 0;
+				$path = ['query', $this->getModuleName(), $index];
```
Grunny: Minor: should be spacey.

Grunny:
One suggestion, otherwise LGTM.

Grunny:
LGTM.

adamkarminski:
LGTM.

adamkarminski:
LGTM

PR #9787, opened 2016-02-23
SUS-156 refactor TitleBatch

sqreek:
#### [SUS-156](https://wikia-inc.atlassian.net/browse/SUS-156)

[This](https://github.com/Wikia/app/compare/SUS-156_refactor_TitleBatch?expand=1#diff-e78a1087dff03f2f50e3c1032d474ffdR161) is fix for SUS-156.

Rest is refactoring for getBoardList use one TitleBatch instead using wrappers over TitleBatch static methods (getListTitles).
- [x] Add tests for TitleBatch
- [x] Add tests for getBoardList


mixth-sense:
:+1: - I don't see anything suspicious here

sqreek:
@Wikia/sustaining-team tests added

includes/wikia/cache/TitleBatch.php
```diff
@@ -128,31 +135,37 @@ public function getAll() {
 	}
 
 	/**
+	 * @param $pageId
+	 */
+	public function getById( $pageId ) {
+		if ( empty( $this->titles[ $pageId ] ) ) {
+			WikiaLogger::instance()->warning( "Title Batch does not have cached title", [ "pageId" => $pageId ] );
```
wladekb: I'd use the exact spelling of "TitleBatch" here for easier grepping

includes/wikia/cache/TitleBatch.php
```diff
@@ -19,16 +22,17 @@ class TitleBatch {
...
 	 */
 	public function __construct( $titles ) {
-		foreach ($titles as $k => $title) {
+		$this->titles = [ ];
+		foreach ( $titles as $k => $title ) {
 			if ( $title instanceof Title ) {
-				// nothing to do...
-			} elseif ( is_string($title) ) {
-				$titles[$k] = Title::newFromText($title);
+				$this->titles[ $title->getArticleID() ] = $title;
```
wladekb, line 30: This will change the behavior of TitleBatch. Having the following code:

```
$titles = [
  // assuming both titles don't exist and getArticleID() returns 0 for them
  Title::newFromText('NonExistent1'),
  Title::newFromText('NonExistent2'),
];
$titlesFromTitleBatch = (new TitleBatch($titles))->getAll();
```

previously `getAll()` returned both titles, the new version will return only the second one.
Please make sure you don't break anything with this change.

includes/wikia/tests/TitleBatchTest.php
```diff
@@ -0,0 +1,76 @@
...
+			(object)[ 'page_id' => $expectedId1, 'propname' => $expectedProperty, 'props' => serialize( $expectedPropValue1 ) ],
+			(object)[ 'page_id' => $expectedId2, 'propname' => $expectedProperty, 'props' => serialize( $expectedPropValue2 ) ],
+		];
+
+		$mockDb->expects( $this->once() )
+			->method( 'select' )
+			->with( $fields, $table, $cond, 'TitleBatch::getWikiaProperties' )
+			->will( $this->returnValue( $result ) );
+
+		$mockGetDB = $this->getGlobalFunctionMock( 'wfGetDB' );
```
wladekb: Can be simplified if the check for DB_SLAVE is not relevant by replacing the entire block by a single line:

```
$this->mockGlobalFunction('wfGetDB',$mockDb);
```

extensions/wikia/Forum/tests/ForumTest.php
```diff
@@ -0,0 +1,61 @@
...
+	public function testGetBoardListShouldReturnBoardsWithoutLeftoversInWikiaProps() {
+
+		$titlesFromNamespace = [ ];
+		foreach ( [ 1, 2, 3, 4 ] as $key ) {
+			$titlesFromNamespace[ $key ] = $this->getTitleMock( $key );
+		}
+		$wikiaPropsReturn = [ 1 => 4, 2 => 7, 3 => 6, 5 => 2 ];
+		$expectedBoardsTitle = [ 1, 2, 3 ];
+
+		$titleBatchMock = $this->getMock( 'TitleBatch', [ 'getById', 'getWikiaProperties' ], [ ], '', false );
```
wladekb: Mocking classes with the explicit list of methods is a bit dangerous because the methods will be mocked even if they do not exist in the target class. As long as you do not need all other methods of TitleBatch to act normally I'd recommend removing that list here.

extensions/wikia/Forum/tests/ForumTest.php
```diff
@@ -0,0 +1,61 @@
...
+
+	private function getTitleMock( $key ) {
+		$titleMock = $this->getMock( 'Title', [ 'getArticleID' ] );
+		$titleMock->expects( $this->any() )
+			->method( 'getArticleID' )
+			->willReturn( $key );
+		return $titleMock;
+	}
+
+	private function getBoardMock( $title ) {
```
wladekb: Argument is unused.

extensions/wikia/Forum/tests/ForumTest.php
```diff
@@ -0,0 +1,61 @@
...
+				} ) );
+
+		$titleBatchMock->expects( $this->once() )
+			->method( 'getWikiaProperties' )
+			->willReturn( $wikiaPropsReturn );
+
+		$this->mockStaticMethod( 'TitleBatch', 'newFromConds', $titleBatchMock );
+
+		$this->mockStaticMethodWithCallBack( 'ForumBoard', 'newFromTitle',
+			function ( $Title ) {
```
wladekb: In PHP you should follow camelCase except for class names and namespaces.

includes/wikia/cache/TitleBatch.php
```diff
@@ -19,16 +22,17 @@ class TitleBatch {
 	 * @param $titles array List of Title objects or title texts
 	 */
 	public function __construct( $titles ) {
-		foreach ($titles as $k => $title) {
+		$this->titles = [ ];
```
wladekb, line 27: please add a todo above:

```
// @todo: load article ids (if they aren't loaded yet) in a more efficient way than one by one
```

wladekb:
I've left a few comments inline. I'm glad to see one bug fixed in the `TitleBatch::getWikiaProperties` by adding a missing condition:
`'p.page_id' => $this->getArticleIds()`

However I'm probably missing the more important point. Could you help me understand what gets improved by this refactor?

includes/wikia/cache/TitleBatch.php
sqreek, line 30: `getAll()` is used only in one place. I will add additional check for that

sqreek:
@wladekb Most important bug was in `TitleBatch::getWikiaProperties`
Also 
- added `getById`
- `newFromConds` return `TitleBatch` object instead array
- `getWikiaProperties` return always 1D array

wladekb:
lgtm

owend:
There was a broken test that was blocking this but it's been fixed.

PR #9790, opened 2016-02-23
Fight stupid content thefts

gabrys:


frankfarmer:
heh worth a shot.

more generally, it'd be interesting to silently log any case where `!location.host.match(/wikia.com$/)`

relates to https://wikia-inc.atlassian.net/browse/OPS-7901

gabrys:
The issue is they strip out any `<script>` tags, so you don't have a lot of room (without polluting the actual wikia pages)

PR #9793, opened 2016-02-24
SEO-292 Update WikiaMobile wgWikiaMobileNavigationBlacklist

gabrys:
Prevent both Special:NewFiles and Special:Images from appearing in the
WikiaMobile nav menu.


saipetch:
LGTM

PR #9794, opened 2016-02-24
SEO-290 Update link generated in CityVisualization

gabrys:
I wanted to remove the extension, but it seems it's actually used by the
next version of hubs code, so I'm just updating the link to point to the
new URL for Special:Images


saipetch:
LGTM

PR #9795, opened 2016-02-24
SEO-294 Clean up NewFiles -> Images: Update VideoEmbedTool

gabrys:
There's a method in VET called recentUploads that currently doesn't
work. You can asset it using an URL like this:

http://muppet.rychu.wikia-dev.com/index.php?action=ajax&rs=VET&method=recentlyUploaded

This returns a 500 response, because it calls a function that doesn't
exist.

Apparently (I checked the logs) this function is not even used, so the
only way to get this error to visit the URL manually. (The only errors
logged Kibana are those coming from me testing).

Long story short: I'm removing this method.


saipetch:
LGTM

PR #9800, opened 2016-02-24
In WikiaBaseTest add mockStaticMethodWithCallBack

sqreek:
Media wiki has a lot of static methods that act as constructors such as `Revision::newFromId`

In some tests we want to return a specific instance depending on params instead of using one single mock

@Wikia/engineers

### DIFF
--- a/includes/wikia/tests/core/WikiaBaseTest.class.php
+++ b/includes/wikia/tests/core/WikiaBaseTest.class.php
@@ -225,6 +225,34 @@ abstract class WikiaBaseTest extends PHPUnit_Framework_TestCase {
 	}
 
 	/**
+	 * Mock a static method using callback
+	 *
+	 * Example:
+	 *
+	 * $this->mockStaticMethodWithCallBack( 'Title', 'newFromID',
+	 *   function ( $titleId ) {
+	 *     $titleMock = $this->getMock( 'Title', [ 'getArticleID' ] );
+	 *     $titleMock->expects( $this->any() )
+	 *       ->method( 'getArticleID' )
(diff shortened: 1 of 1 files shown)


includes/wikia/tests/core/WikiaBaseTest.class.php
```diff
@@ -225,6 +225,34 @@ protected function mockStaticMethod($className, $methodName, $retVal) {
...
+	 * );
+	 * $this->assertEquals( 7, Title::newFromID( 7 )->getArticleID() );
+	 * $this->assertEquals( 12, Title::newFromID( 12 )->getArticleID() );
+	 * $this->assertEquals( 123, Title::newFromID( 123 )->getArticleID() );
+	 *
+	 * @param $className
+	 * @param $methodName
+	 * @param $callBack
+	 */
+	protected function mockStaticMethodWithCallBack( $className, $methodName, $callBack) {
```
macbre: You can provide a type for $callback argument - `callable $callback`.

macbre:
LGTM (one minor comment above)

alistra:
@sqreek please stop pinging wikia/engineers, as this are also people that don't know web development or PHP. Please create a group wikia/webdev or something

wladekb:
FYI the same can be achieved in a more PHPUnit-like way:

```
$staticMethodMock = $this->getStaticMethodMock('ClassName', 'method');
$staticMethodMock
    ->expects($this->any())
    ->method('method')
    ->willReturnCallback($callback);
```

however we have a few helpers like the one you suggest so maybe adding a new one for this case might be reasonable. Any opinions?

PR #9801, opened 2016-02-24
CE-3265 Add an information to the RevisionApi if a user has upvoted a revision

adamkarminski:
@Wikia/community-engineering

### DIFF
--- a/includes/wikia/api/RevisionApiController.class.php
+++ b/includes/wikia/api/RevisionApiController.class.php
@@ -8,10 +8,11 @@ class RevisionApiController extends WikiaApiController {
 	 * Gets diff between two revisions.
 	 * Also includes data about revision and article
 	 *
-	 * @param int $oldId old revision id (required)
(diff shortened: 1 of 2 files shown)


includes/wikia/api/RevisionApiController.class.php
```diff
@@ -31,6 +33,7 @@ public function getRevisionsDiff() {
 
 		$avatar = $request->getBool( 'avatar' );
 		$oldRev = $request->getBool( 'oldRev' );
+		$upvote = $request->getBool( 'upvote' );
```
lukaszkonieczny, line 36: see https://github.com/Wikia/app/pull/9784/files#diff-784f2baa3bf95847c76ff321f96a6544R101 and https://github.com/Wikia/app/commit/6530d005cec8613479ba735400a29d4a82acb738. I added needed data about revision there, so we don't need to query database (we will get data about users).

PR #9805, opened 2016-02-24
MAIN-6512 Call track method using tracker

drets:
Fix `trackVerboseImpression` method call by using `tracker` object


nandy-andy:
LGTM :+1:

PR #9808, opened 2016-02-24
SEO-280 Use <h2> tags for user badges rather than <h1>

gabrys:
There should be only one <h1> per page.

And those headers look much better as <h2> anyway (being consistent with
other modules in the rail).


saipetch:
LGTM

PR #9810, opened 2016-02-24
SEO-279 Don't emit hreflang links on redirect pages

gabrys:
Apparently $wgTitle->isRedirect() is not how you check if the page is a
redirect, because the $wgTitle is already the target title in this case.

I haven't found a better way of checking this than inspecting
mRedirectedFrom member of $wgArticle.

Additional background is Google reports an error in Search Console when
you have a page A hreflang-linking to page B, which doesn't link back to
page A.

This is the case if you have a page C that's a MediaWiki redirect to
page A. That page has a hreflang link to B, but B links back to A.

The solution is the page C should not have the hreflang link emitted.


saipetch:
LGTM

PR #9811, opened 2016-02-25
SEO-63 Clean up robots.txt

gabrys:
- Remove redirect-robots.php
- Extend RobotsTxt class to allow overriding the whole file
- Have tests for the logic
- Move $wgExperimentalRobotsTxt logic to wikia-robots-txt.php and use the logic above for that


saipetch:
LGTM

PR #9814, opened 2016-02-25
Separate phpunit reports

ludwikkazmierczak:
Reports needs to be separated, cause when running two different tasks (like `phpunit` `phpunit-infrastructure`) report is being overwritten by the last task run

Review? @mixth-sense @macbre @wladekb

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -49,7 +49,6 @@ phpunit-externals: prepare
 # runs tests for externally used parts
 phpunit-contractual-responsibilities: prepare
 	${CMD_PHPUNIT} \
-	externals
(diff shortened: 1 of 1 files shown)


tests/Makefile
```diff
@@ -1,8 +1,10 @@
 # Wikia's Makefile for executing unit tests
 
+# path to logs folder
+LOGS_PATH=junit=build/logs
```
macbre: Shouldn't it be `LOGS_PATH=build/logs` ?

ludwikkazmierczak: Right ;)

macbre:
:+1:

sebastianmarzjan:
:shipit:

PR #9815, opened 2016-02-25
Update run-test.php to avoid notices

sebastianmarzjan:
## Description

As part of work on https://wikia-inc.atlassian.net/browse/DAT-3693 I noticed that run-test.php throws 

```

Notice: Undefined index: exclude-group in /usr/wikia/source/app/tests/run-test.php on line 9

Call Stack:
    0.0001     235720   1. {main}() /usr/wikia/source/app/tests/run-test.php:0

```

on each run when exclude-group param is not provided. This PR fixes that annoyance.
## Reviewers

@ludwikkazmierczak @mixth-sense

### DIFF
--- a/tests/run-test.php
+++ b/tests/run-test.php
@@ -6,9 +6,8 @@ ini_set('display_errors', 1);
 $wgNoDBUnits = false;
 
 $params = getopt('', ['exclude-group::', 'slow-list']);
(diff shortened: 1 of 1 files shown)


ludwikkazmierczak:
LGTM

PR #9822, opened 2016-02-25
XW-1183 return 404 response for not existing articles

bognix:
## Links
- http://wikia-inc.atlassian.net/browse/XW-1183
- https://github.com/Wikia/mercury/pull/2054


extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -433,6 +433,13 @@ public function getPage() {
 				if ( !$isMainPage ) {
 					$titleBuilder->setParts( [ $data['article']['displayTitle'] ] );
 				}
+			} else if ( !$article instanceof Article ) {
```
rogatty, line 436: We use `elseif` above. Let's stick to one of them (I prefer `else if`, but whatever).

rogatty:
Minor thing and LGTM.

vforge:
LGTM

PR #9829, opened 2016-02-26
PagesWithoutInfobox: do not start a transaction

macbre:

It breaks the replication when `updateSpecialPages.php` is run across the cluster

@adamkarminski / @lukaszkonieczny


ghost:
:+1:

PR #9838, opened 2016-02-26
Remove unused tests/phpunit folder

jcellary:
This code is not used any more, so deleting per offline conversation with @wladekb and @macbre


macbre:
LGTM

wladekb:
Yeah, I think there's a few more dirs in `tests/` that we don't use, I'm not sure who could be the best person to judge it?

jcellary:
Not sure - do you have any proposals what else isn't used?

PR #9844, opened 2016-02-26
Wall refactor use const in templates

sqreek:
In Wall/Forum, use const instead magic in templates.

@Wikia/sustaining-team 

Good luck guys


kamilkoterba:
LGTM.

Although I'm not a fan of using `${ForumConst::SOME_CONST}` in templates in this particular context I can agree on this solution as it eases Wall debugging. Especially when templates here are using vars like `$body` `$id` `$error` `$action` etc.

If we move're moving to conventions field I'd suggest using usual approach like `$wallMyName` and I believe each var should be unique enough so it would be easy to find where it's generated.

sqreek:
Closing in favor for https://github.com/Wikia/app/pull/9976.

PR #9845, opened 2016-02-26
RECIRC-61: Sanity check Discussions Service $rawPosts as an array

pauloslund:


pauloslund:
Checked and given the LGTM by @gabrys

PR #9852, opened 2016-02-29
(CE-3282) Get time IDs starting from last week's Sunday

Grunny:
Get time IDs starting from last week's Sunday rather than the Sunday
of the current week so that we have page view data for four complete
weeks.


adamkarminski:
LGTM

PR #9853, opened 2016-02-29
AMQPMessage: send more context as RabbitMQ message properties

macbre:
Sending the app name and `correlation_id` (`X-Request-Id` in our case) should help us find the root cause of [PLATFORM-1943](https://wikia-inc.atlassian.net/browse/PLATFORM-1943) (and find what sends messages with no payload to RabbitMQ).

Pinging @wladekb as it touches the topic of unified logging and connecting various pieces of our infrastructure that is used during end-user request handling.

It will affect messages pushed by `AsyncTaskList::queue` and `IndexingPipeline`.


macbre:
Travis reports:

```
npm ERR! EEXIST, symlink '../esprima/bin/esparse.js'
File exists: ../esprima/bin/esparse.js
Move it away, and try again. 
```

I believe it can be ignored, pinging @ludwikkazmierczak  as it's not the first time this build fails on Jenkins with the same error.

wladekb:
Unfortunately I don't think that's a good place to make this change. This is a library managed by composer and I'm not really sure if composer looks for local changes and will react appropriately. Could you find some better place?

macbre:
True, I wanted to avoid copy & pasting, but let's avoid modifying the `/lib`.

extensions/wikia/IndexingPipeline/ConnectionBase.class.php
```diff
@@ -42,7 +43,9 @@ public function publish( $routingKey, $body ) {
 			$channel->basic_publish(
 				new AMQPMessage( json_encode( $body ), [
 					'delivery_mode' => self::DURABLE_MESSAGE,
-					'expiration' => self::MESSAGE_TTL
+					'expiration' => self::MESSAGE_TTL,
+					'app_id' => 'mediawiki',
+					'correlation_id' => RequestId::instance()->getRequestId(),
```
macbre, line 48: Pinging @SebastianMarzjan

macbre:
Merged into today's upcoming release in 6de97f0

PR #9855, opened 2016-02-29
(INT-367) Prevent double escaping of CuratedContent description messages

Grunny:
The CuratedContent description messages are currently escaped when they
are output in the template. However, because the wfMessage calls do not
have a specific output format, they are first run through the MediaWiki
parser, leading them to be double escaped. Since we only need magic words
to be expanded in one of the messages, switch the output format to plain
and text depending on the need and leave the escaping in the template.


lukaszkonieczny:
LGTM

lukaszkonieczny:
LGTM :)

PR #9858, opened 2016-02-29
SEO-50 Admin dashboard title

gabrys:
OutputPage::setHTMLTitle now calls WikiaHtmlTitle::generateTitle().

It sets HTML title based on the information passed from OutputPage,
adding the HTML title structure for admin dashboard.

Later we should expand this method, to add structure for images,
videos, categories, blogs, etc.

At that point, the method should be promoted to a separate class
with a set of tests.


saipetch:
LGTM

PR #9862, opened 2016-03-01
DAT-3877 do not use max-width in infobox label styles

dianafa:
https://wikia-inc.atlassian.net/browse/DAT-3877


extensions/wikia/PortableInfobox/styles/PortableInfoboxMixins.scss
```diff
@@ -71,7 +69,7 @@
 		&:not(:first-child) {
 			@include flex-basis($infobox-width * 2 / 3);
 
-			padding-left: $padding;
+			padding-left: 10px;
```
RafLeszczynski: Try to avoid magic numbers. 
IMO having it in variable was better.

Except that LGTM

RafLeszczynski:
LGTM

PR #9865, opened 2016-03-01
DAT 3793, combine infobox builder with manual template classification modal

RafLeszczynski:
https://wikia-inc.atlassian.net/browse/DAT-3793

### DIFF
--- a/extensions/wikia/PortableInfoboxBuilder/PortableInfoboxBuilderHooks.class.php
+++ b/extensions/wikia/PortableInfoboxBuilder/PortableInfoboxBuilderHooks.class.php
@@ -91,7 +91,7 @@ class PortableInfoboxBuilderHooks {
 	public static function onCustomEditor( $page, $user ) {
 		$title = $page->getTitle();
 
-		if ( self::isEditableInfobox( $title, $user ) ) {
+		if ( self::canUseInfoboxBuilder( $title, $user ) ) {
 			$url = SpecialPage::getTitleFor( 'InfoboxBuilder', $title->getText() )->getInternalURL();
 			F::app()->wg->out->redirect( $url );
 			return false;
@@ -122,8 +122,11 @@ class PortableInfoboxBuilderHooks {
 	 * @param $title
 	 * @return bool
 	 */
-	private static function isEditableInfobox( $title, $user ) {
+	private static function canUseInfoboxBuilder( $title, $user ) {
 		return self::isInfoboxTemplate( $title )
+		&& ( new \PortableInfoboxBuilderService() )->isValidInfoboxArray(
+			\PortableInfoboxDataService::newFromTitle( $title )->getInfoboxes()
+		)
 		&& ( new \Wikia\TemplateClassification\Permissions() )->userCanChangeType( $user, $title );
 	}
 }
--- a/extensions/wikia/PortableInfoboxBuilder/PortableInfoboxBuilderHooks.class.php
+++ b/extensions/wikia/PortableInfoboxBuilder/PortableInfoboxBuilderHooks.class.php
@@ -35,8 +35,11 @@ class PortableInfoboxBuilderHooks {
 		$title = $skin->getTitle();
 
 		if ( $title && $title->isSpecial( PortableInfoboxBuilderSpecialController::PAGE_NAME ) ) {
+			// remove the special page name from the title and returl url to the template
+			// passed after the slash, i.e.
+			// Special:InfoboxBuilder/TemplateName/Subpage => Template:TemplateName/Subpage
 			$vars['templatePageUrl'] = Title::newFromText(
-				implode( '/', array_slice( explode( '/', $title->getSubpageText() ), 1 ) ),
+				implode( '/', array_slice( explode( '/', $title->getText() ), 1 ) ),
 				NS_TEMPLATE )->getFullUrl();
 		}
 
--- a/extensions/wikia/PortableInfoboxBuilder/PortableInfoboxBuilderHooks.class.php
+++ b/extensions/wikia/PortableInfoboxBuilder/PortableInfoboxBuilderHooks.class.php
@@ -35,7 +35,7 @@ class PortableInfoboxBuilderHooks {
 		$title = $skin->getTitle();
 
 		if ( $title && $title->isSpecial( PortableInfoboxBuilderSpecialController::PAGE_NAME ) ) {
-			// remove the special page name from the title and returl url to the template
+			// remove the special page name from the title and return url to the template
 			// passed after the slash, i.e.
 			// Special:InfoboxBuilder/TemplateName/Subpage => Template:TemplateName/Subpage
 			$vars['templatePageUrl'] = Title::newFromText(
--- a/extensions/wikia/PortableInfoboxBuilder/PortableInfoboxBuilderHooks.class.php
+++ b/extensions/wikia/PortableInfoboxBuilder/PortableInfoboxBuilderHooks.class.php
@@ -4,17 +4,6 @@ class PortableInfoboxBuilderHooks {
 	const INFOBOX_BUILDER_SPECIAL_PAGE = 'Special:InfoboxBuilder';
 
 	/**
-	 * Adds infobox builder helper js assets to Template Classification on Edit page
-	 *
-	 * @return true
-	 */
-	public function onTCAfterEditPageAssets() {
-		\Wikia::addAssetsToOutput( 'portable_infobox_builder_template_classification_helper_js' );
-
-		return true;
-	}
-
-	/**
 	 * @param Skin $skin
 	 * @param string $text
 	 *
--- a/extensions/wikia/PortableInfoboxBuilder/PortableInfoboxBuilderHooks.class.php
+++ b/extensions/wikia/PortableInfoboxBuilder/PortableInfoboxBuilderHooks.class.php
@@ -39,7 +39,6 @@ class PortableInfoboxBuilderHooks {
 
 			$vars['templatePageUrl'] = Title::newFromText(
 				self::getUrlPath( $titleText ),
-				$title->getBaseText(),
 				NS_TEMPLATE
 			)->getFullUrl();
 		}
--- a/extensions/wikia/PortableInfoboxBuilder/PortableInfoboxBuilderHooks.class.php
+++ b/extensions/wikia/PortableInfoboxBuilder/PortableInfoboxBuilderHooks.class.php
@@ -47,6 +47,31 @@ class PortableInfoboxBuilderHooks {
 		return true;
 	}
 
+
+	/**
+	 * Add global variables for Javascript
+	 * @param array $aVars
+	 * @return bool
+	 */
+	public function onEditPageMakeGlobalVariablesScript( array &$aVars ) {
+		$context = \RequestContext::getMain();
+		$title = $context->getTitle();
+		if (
+			( new \Wikia\TemplateClassification\Permissions() )->shouldDisplayEntryPoint( $context->getUser(), $title )
+			&& \RequestContext::getMain()->getRequest()->getVal( 'action' ) === 'edit'
+		) {
+			$aVars['isTemplateBodySupportedInfobox'] =
+				( new \PortableInfoboxBuilderService() )->isValidInfoboxArray(
+					\PortableInfoboxDataService::newFromTitle( $title )->getInfoboxes()
+				);
+			$aVars['tcBodyClassName'] = \Wikia\TemplateClassification\Hooks::TC_BODY_CLASS_NAME;
+			$aVars['infoboxBuilderPath'] = \SpecialPage::getTitleFor( 'InfoboxBuilder', $title->getText() )
+				->getFullURL();
+
+		}
+		return true;
+	}
+
 	/**
 	 *
 	 * @param $page \Article|\Page
--- a/extensions/wikia/PortableInfoboxBuilder/PortableInfoboxBuilderHooks.class.php
+++ b/extensions/wikia/PortableInfoboxBuilder/PortableInfoboxBuilderHooks.class.php
@@ -35,11 +35,11 @@ class PortableInfoboxBuilderHooks {
 		$title = $skin->getTitle();
 
 		if ( $title && $title->isSpecial( PortableInfoboxBuilderSpecialController::PAGE_NAME ) ) {
-			// remove the special page name from the title and return url to the template
-			// passed after the slash, i.e.
-			// Special:InfoboxBuilder/TemplateName/Subpage => Template:TemplateName/Subpage
+			$titleText = $title->getText();
+
 			$vars['templatePageUrl'] = Title::newFromText(
-				implode( '/', array_slice( explode( '/', $title->getText() ), 1 ) ),
+				self::getUrlPath( $titleText ),
+				$title->getBaseText(),
 				NS_TEMPLATE
 			)->getFullUrl();
 		}
@@ -47,6 +47,16 @@ class PortableInfoboxBuilderHooks {
 		return true;
 	}
 
+	/**
+	 * remove the special page name from the title and return name of the template
+	 * passed after the slash without the namespace, i.e.
+	 * Special:InfoboxBuilder/TemplateName/Subpage => TemplateName/Subpage
+	 * @param $titleText
+	 * @return string
+	 */
+	private static function getUrlPath( $titleText ) {
+		return implode( '/', array_slice( explode( '/', $titleText ), 1 ) );
+	}
 
 	/**
 	 * Add global variables for Javascript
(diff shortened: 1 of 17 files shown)


extensions/wikia/TemplateClassification/scripts/TemplateClassificationModal.js
```diff
@@ -201,13 +204,43 @@ function ($, w, mw, loader, nirvana, tracker, throbber, labeling) {
...
+			newTemplateType === 'infobox' &&
+			modalMode !== forceClassificationModalMode;
+	}
+
+	/**
+	 * @desc redirects to infobox builder tool
+	 * @param {String} title
+	 */
+	function redirectToInfoboxBuilder(title) {
+		w.location = w.location.origin + '/wiki/Special:InfoboxBuilder/' + title;
```
sebastianmarzjan: Hmm, this path looks like something that's better passed from the app to support i18n (namespace and special page aliases)

extensions/wikia/TemplateClassification/TemplateClassification.hooks.php
```diff
@@ -98,12 +106,20 @@ public function onEditPageShowEditFormFields( \EditPage $editPage, \OutputPage $
...
 			$types = $this->getTemplateTypeForEdit( $editPage->getTitle(), $wgCityId );
 
-			$out->addHTML( \Html::hidden( 'templateClassificationTypeCurrent', $types['current'] ) );
-			$out->addHTML( \Html::hidden( 'templateClassificationTypeNew', $types['new'] ) );
+			$out->addHTML( \Html::hidden( 'templateClassificationTypeCurrent', $types['current'],
+				['autocomplete' => 'off'] ) );
+			$out->addHTML( \Html::hidden( 'templateClassificationTypeNew', $types['new'],
+				['autocomplete' => 'off'] ) );
+
+			// add additional class to body for new templates in order to hide editor while TCS modal is visible
```
sebastianmarzjan: Do we want to do it only for new templates? How we cover existing template with no set type?

dianafa: couldn't we use the same logic for adding class as for displaying modal - that is always when TC modal is displayed - add class to body and hide background.

Nitpick: it's `TC` modal, not `TC Storage` modal.

extensions/wikia/TemplateClassification/TemplateClassification.hooks.php
```diff
@@ -98,12 +106,20 @@ public function onEditPageShowEditFormFields( \EditPage $editPage, \OutputPage $
 		global $wgCityId;
 
 		$context = \RequestContext::getMain();
+		$title = $context->getTitle();
 
-		if ( ( new Permissions() )->shouldDisplayEntryPoint( $context->getUser(), $context->getTitle() ) ) {
+		if ( ( new Permissions() )->shouldDisplayEntryPoint( $context->getUser(), $title ) ) {
 			$types = $this->getTemplateTypeForEdit( $editPage->getTitle(), $wgCityId );
```
dianafa, line 106: if we decided to move $title to a separate var, why not `$context->getUser()` ?

extensions/wikia/PortableInfoboxBuilder/js/PortableInfoboxBuilderPonto.js
```diff
@@ -19,7 +19,7 @@ define('wikia.infoboxBuilder.ponto', ['wikia.window', 'ponto'], function (window
 		 * @returns {Boolean}
 		 */
 		this.redirectToTemplatePage = function (title) {
-			window.location = window.location.origin + '/wiki/Template:' + title;
+			w.location = w.location.origin + '/wiki/Template:' + title;
```
dianafa: how this gonna work for templates like: http://marvel.wikia.com/wiki/Marvel_Database:Comic_Template ?

extensions/wikia/TemplateClassification/styles/TemplateClassification.scss
```diff
@@ -128,3 +128,8 @@ $tc-name-color-checked: $color-links;
 		text-align: right;
 	}
 }
+
+// for new templates in order to hide editor while TCS modal is visible
+.show-template-classification-modal .WikiaPage {
+	opacity: 0;
+}
```
dianafa, line 136: do we wanna go with kind of `fade-in` or other sweet UX stuff when revealing source editor? ^^

extensions/wikia/TemplateClassification/TemplateClassification.hooks.php
```diff
@@ -79,11 +80,18 @@ public function onArticleSaveComplete( \WikiPage $article, \User $user, $text, $
...
-		if ( ( new Permissions() )->shouldDisplayEntryPoint( $context->getUser(), $context->getTitle() )
+		if ( ( new Permissions() )->shouldDisplayEntryPoint( $context->getUser(), $title )
 			&& $this->isEditPage()
 		) {
 			$aVars['enableTemplateClassificationEditorPlugin'] = true;
+			$aVars['isSupportedInfoboxTemplateBody'] =
+				( new \PortableInfoboxBuilderService() )->isValidInfoboxArray(
+					\PortableInfoboxDataService::newFromTitle( $title )->getInfoboxes()
+				);
+			$aVars['tcsBodyClassName'] = self::TCS_BODY_CLASS_NAME;
```
dianafa: see my comment below, I'd change `tcs` to `tc`

extensions/wikia/PortableInfoboxBuilder/js/PortableInfoboxBuilderPonto.js
sebastianmarzjan: It's not, and actually we're purposefully supporting only NS 10. However, https://github.com/Wikia/app/pull/9865/files#diff-d5bde9afc6a40ff66cf72bf44a3fc4b2R234 still holds true in regard to the NS name.

extensions/wikia/TemplateClassification/TemplateClassification.hooks.php
sebastianmarzjan, line 106: Because $context-getUser() it's used only once, while $context->getTitle() would have two invocations in this part of code

extensions/wikia/PortableInfoboxBuilder/js/PortableInfoboxBuilderPonto.js
```diff
@@ -9,24 +9,23 @@ define('wikia.infoboxBuilder.ponto', ['wikia.window', 'ponto'], function (window
...
 
 		/**
 		 * redirects to template page
-		 * @param {String} title
 		 * @returns {Boolean}
 		 */
-		this.redirectToTemplatePage = function (title) {
-			window.location = window.location.origin + '/wiki/Template:' + title;
+		this.redirectToTemplatePage = function () {
+			w.location = w.templatePageUrl;
```
rogatty, line 21: Is it possible that this won't be available? Should we check it first to be sure we don't redirect to undefined?

extensions/wikia/PortableInfoboxBuilder/tests/PortableInfoboxBuilderServiceTest.php
```diff
@@ -51,6 +58,15 @@ public function testUpdateDocumentation($data, $expected) {
 			$data['newDoc'], $data['oldContent']));
 	}
 
+
+	public function infoboxArrayValidityDataProvider() {
+		return [
+			[ [], true, 'Empty infobox array is not valid' ],
```
rogatty: Should it say `Empty infobox array is valid`?

extensions/wikia/PortableInfoboxBuilder/tests/PortableInfoboxBuilderServiceTest.php
```diff
@@ -51,6 +58,15 @@ public function testUpdateDocumentation($data, $expected) {
...
+			[ [ '<infobox><data source="source1"></data></infobox>' ], true, 'Single infobox array is valid' ],
+			[ [ '<infobox><data source="source1"></data></infobox>', '<infobox><data source="source2"></data></infobox>' ], false, 'Multi- infobox array is not valid' ],
+		];
```
rogatty, line 73: Should we also test a case where there is one infobox with unsupported markup?

extensions/wikia/TemplateClassification/TemplateClassification.hooks.php
```diff
@@ -79,11 +80,20 @@ public function onArticleSaveComplete( \WikiPage $article, \User $user, $text, $
...
 	public function onEditPageMakeGlobalVariablesScript( array &$aVars ) {
 		$context = \RequestContext::getMain();
+		$title = $context->getTitle();
 		// Enable TemplateClassificationEditorPlugin
-		if ( ( new Permissions() )->shouldDisplayEntryPoint( $context->getUser(), $context->getTitle() )
+		if ( ( new Permissions() )->shouldDisplayEntryPoint( $context->getUser(), $title )
 			&& $this->isEditPage()
 		) {
 			$aVars['enableTemplateClassificationEditorPlugin'] = true;
+			$aVars['isSupportedInfoboxTemplateBody'] =
```
rogatty: `isTemplateBodySupportedInfobox`?

extensions/wikia/PortableInfoboxBuilder/tests/PortableInfoboxBuilderServiceTest.php
dianafa, line 73: we have separate tests for that

rogatty, line 73: `isValidInfoboxArray` does this check:

```
return empty( $infoboxes ) || ( count( $infoboxes ) === 1 && $this->isSupportedMarkup( $infoboxes[0] ) );
```

If this is supposed to be a proper unit test then we should test against these conditions. Also, `isSupportedMarkup` result should be mocked and passed from provider. Then you wouldn't say we have separate tests for that ;)

extensions/wikia/TemplateClassification/scripts/TemplateClassificationModal.js
```diff
@@ -201,13 +203,42 @@ function ($, w, mw, loader, nirvana, tracker, throbber, labeling) {
...
+		return w.isSupportedInfoboxTemplateBody &&
+			newTemplateType === 'infobox' &&
+			modalMode !== forceClassificationModalMode;
+	}
+
+	/**
+	 * @desc redirects to infobox builder tool
+	 */
+	function redirectToInfoboxBuilder() {
+		w.location = w.infoboxBuilderPath;
```
rogatty: Is it possible that this won't be available? Should we check it first to be sure we don't redirect to undefined?

extensions/wikia/TemplateClassification/scripts/TemplateClassificationModal.js
```diff
@@ -201,13 +203,42 @@ function ($, w, mw, loader, nirvana, tracker, throbber, labeling) {
 			});
 		}
 
-		if (modalMode === 'addTypeBeforePublish' && newTemplateType) {
+		if (modalMode === forceClassificationModalMode && newTemplateType) {
 			$('#wpSave').click();
+		} else if (shouldRedirectToInfoboxBuilder(newTemplateType, modalMode)) {
+			throbber.show(modalInstance.$content);
+			redirectToInfoboxBuilder(mw.config.get('wgTitle'));
```
rogatty: `redirectToInfoboxBuilder()` no longer takes any params.

rogatty:
Maybe there is something I'm missing but won't this code redirect users to Infobox Builder even when it's not enabled?

extensions/wikia/PortableInfoboxBuilder/js/PortableInfoboxBuilderPonto.js
sebastianmarzjan, line 21: We might, thought this is always appended in InfoboxBuilder.
I wonder what the right behavior would be in such a case. Reload?

extensions/wikia/TemplateClassification/scripts/TemplateClassificationModal.js
sebastianmarzjan: Checked in https://github.com/Wikia/app/pull/9865/files#diff-d5bde9afc6a40ff66cf72bf44a3fc4b2R192

extensions/wikia/PortableInfoboxBuilder/js/PortableInfoboxBuilderPonto.js
sebastianmarzjan, line 21: I would not support it for now - on InfoboxBuilder page this var is always set

extensions/wikia/PortableInfoboxBuilder/tests/PortableInfoboxBuilderServiceTest.php
```diff
@@ -37,6 +37,18 @@ public function testMarkupSupport( $markup, $expected, $message ) {
 	}
 
 	/**
+	 * @dataProvider infoboxArrayValidityDataProvider
+	 */
+	public function testInfoboxArrayValidity( $array, $isSupportedMarkupReturnValue, $expected, $message ) {
+		$builderServiceMock = $this->getMockBuilder( 'PortableInfoboxBuilderService' )
+			->setMethods( [ 'isSupportedMarkup' ] )
+			->getMock();
+		$builderServiceMock->method('isSupportedMarkup')->willReturn($isSupportedMarkupReturnValue);
```
rogatty: Minor - missing spaces.

sebastianmarzjan:
You are right, we were redirecting to IB even when it was disabled. Fixed.

rogatty:
LGTM

extensions/wikia/PortableInfoboxBuilder/PortableInfoboxBuilderHooks.class.php
```diff
@@ -22,4 +22,98 @@ public static function onSkinAfterBottomScripts( $skin, &$text ) {
...
+
+		if ( $title && $title->isSpecial( PortableInfoboxBuilderSpecialController::PAGE_NAME ) ) {
+			// remove the special page name from the title and return url to the template
+			// passed after the slash, i.e.
+			// Special:InfoboxBuilder/TemplateName/Subpage => Template:TemplateName/Subpage
+			$vars['templatePageUrl'] = Title::newFromText(
+				implode( '/', array_slice( explode( '/', $title->getText() ), 1 ) ),
+				NS_TEMPLATE
+			)->getFullUrl();
+		}
```
dianafa, line 56: looks like a candidate for unit tests

sebastianmarzjan, line 56: Added in https://github.com/Wikia/app/commit/d93c27d513861d24c3d54c0b6e8d82796054c480

dianafa:
LGTM

extensions/wikia/PortableInfoboxBuilder/PortableInfoboxBuilderHooks.class.php
```diff
@@ -22,4 +22,111 @@ public static function onSkinAfterBottomScripts( $skin, &$text ) {
...
+	 * @param Array $vars - (reference) js variables
+	 * @param Array $scripts - (reference) js scripts
+	 * @param Skin $skin - skins
+	 * @return Boolean True - to continue hooks execution
+	 */
+	public static function onWikiaSkinTopScripts( &$vars, &$scripts, $skin ) {
+		$title = $skin->getTitle();
+
+		if ( $title && $title->isSpecial( PortableInfoboxBuilderSpecialController::PAGE_NAME ) ) {
+			$templateTitleText = self::getUrlPath( $titleText = $title->getText() );
```
rogatty: I don't think we should do this.

> Space is cheap, and you're a fast typist

https://www.mediawiki.org/wiki/Manual:Coding_conventions/PHP#Assignment_expressions

rogatty: Also, I don't see `$titleText` being used at all.

extensions/wikia/PortableInfoboxBuilder/PortableInfoboxBuilderHooks.class.php
```diff
@@ -4,6 +4,17 @@ class PortableInfoboxBuilderHooks {
...
 
 	/**
+	 * Adds infobox builder helper js assets to Template Classification on Edit page
+	 *
+	 * @return true
+	 */
+	public function onTCAfterEditPageAssets() {
+		\Wikia::addAssetsToOutput( 'portable_infobox_builder_template_classification_helper_js' );
+
+		return true;
```
rogatty, line 14: Let's either remove this line or check the value returned here in `TemplateClassification.hooks::onBeforePageDisplay`. I suggest the former one because we don't use it anyway.

sebastianmarzjan, line 14: Lack of explicit bool return value will cause a MW throw (hook requirement)

rogatty, line 14: Oh, that's so nice and logical! Thanks for info.

dianafa, line 14: welcome in MW :)

PR #9866, opened 2016-03-01
[PLATFORM-1922] Changed german translations for moderators

jcellary:
Changed german translations for content/discussion moderators

@Grunny


Grunny:
:+1:

PR #9867, opened 2016-03-01
SEO-202 Cache Special:Images and expose it to robots

gabrys:
For a fraction of our wikis (1/10) we want Special:Images exposed to
robots. To avoid a potential performance drop, we're also enabling
caching on those pages (5 minutes TTL). The caching is only enabled for
anonymous users (including bots), so the logged in users should be
unaffected.


saipetch:
LGTM

PR #9868, opened 2016-03-02
SUS-254 P2 Fix Notifications Loading

owend:
https://wikia-inc.atlassian.net/browse/SUS-254

parseInt returns NaN on failure, which is what we should be testing for (not undefined).  seems like there are cases when checking notifications on external/foreign wikis that this value is not set so it sends NaN to the server. 

also removed username= param which is unused by the controller.

@mixth-sense @Wikia/sustaining-team


ghost:
Makes sense. Thanks! :+1:

ukaka:
It passes QA. Selenium test which was failing before is passing after fix.

PR #9869, opened 2016-03-02
SEO-283 Remove XML Sitemaps from www.wikia.com and hubs

gabrys:
When $wgEnableSpecialSitemapExt is false, there no point in robots.txt
pointing robots to the /sitemap-index.xml, because it's gonna serve 404
saying no such special page.

Robots.txt now check the variable and only include the Sitemap rule if
the extension serving the sitemap is enabled.


saipetch:
LGTM

PR #9871, opened 2016-03-02
Merge pull request #9862 from Wikia/DAT-3877

dianafa:
DAT-3877 do not use max-width in infobox label styles


RafLeszczynski:
LGTM after all tests pass

PR #9873, opened 2016-03-02
ADEN-3017 IVW params tweaks

drets:
- for the specific pages send request to IVW with `cp` param in the following format: `<vertical>/<language>/<isHomepage>/<theme>`
- for the regular pages send request to IVW with `cp` param in the `<vertical>` format

### DIFF
--- a/extensions/wikia/AnalyticsEngine/AnalyticsProviderIVW3.php
+++ b/extensions/wikia/AnalyticsEngine/AnalyticsProviderIVW3.php
@@ -66,13 +66,11 @@ class AnalyticsProviderIVW3 implements iAnalyticsProvider {
 	}
 
 	static public function getCMKey() {
-
-		$language = self::getMappedLanguage();
-		$pageType = self::getPageType();
 		$theme = self::getTheme();
 		$vertical = self::getVertical();
 
-		return $theme !== '' ? implode( '/', [$vertical, $language, $pageType, $theme] ) : $vertical;
+		return $theme !== '' ?
+			implode( '/', [$vertical, self::getMappedLanguage(), self::getPageType(), $theme] ) : $vertical;
 	}
 
 	static private function getVertical() {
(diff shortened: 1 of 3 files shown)


extensions/wikia/AnalyticsEngine/AnalyticsProviderIVW3.php
```diff
@@ -45,8 +60,78 @@ private function trackPageView() {
...
 			]
 		);
 	}
+
+	static public function getCMKey() {
+
+		$language = self::getMappedLanguage();
+		$pageType = self::getPageType();
+		$theme = self::getTheme();
+		$vertical = self::getVertical();
```
nandy-andy, line 72: What is the reason to keep those variables?

drets, line 72: readability?

nandy-andy, line 72: How come? Would it be less readable without the vars? You can write it without them in a readable form IMHO :smile_cat:

nandy-andy:
The code looks so-so :japanese_ogre: Before moving to test I've got few suggestions:
- in rare edge-cases it happens `$wgTitle` is `null`; let's handle this exception or just use [`Assert` pattern](https://github.com/Wikia/app/blob/dev/includes/wikia/WikiaPageType.class.php#L249) to be notified if it starts failing more often in the places when we call directly a method on a variable which is suppose to be `Title` instance,
- take a look on our guidelines for PHP and remove unecessary white spaces,
- consider changing static methods to regular one which may help in future unit testing them.

drets:
Thank you for CR :+1: 
- `$wgTitle`:  
  In place which you pointed as example they got a title using `self::getTitle();` which may return [null](https://github.com/Wikia/app/blob/dev/includes/wikia/WikiaPageType.class.php#L17).
  I am not an expert of MediaWiki engine but I believe I use a different way for getting title and it can not be [null](https://github.com/Wikia/app/blob/dev/api.php#L83), so doesn't make sense to use Assert pattern.
- formatting:  
  Fixed; I used this [mediawiki-php-style-helper](https://github.com/Wikia/guidelines/blob/master/PHP/CodingConventions.md#mediawiki-php-style-helper) script
- static:  
  Since we call `getCMKey` method [here](https://github.com/Wikia/app/blob/ADEN-3017-ivw-params/extensions/wikia/MercuryApi/models/MercuryApi.class.php#L142) it doesn't make sense to change static method to regular one; and with static methods we also may create the tests in the future.

extensions/wikia/AnalyticsEngine/AnalyticsProviderIVW3.php
drets, line 72: done

drets:
Talked in Slack - going to add 2 assertions :wink:

jacekjursza:
LGTM

PR #9877, opened 2016-03-03
SEO-256 Local HTML Sitemap MVP

gabrys:
If $wgUseSpecialAllpagesAsLocalSitemap is set:
- Change the styling of the "from A to B" link in Special:Allpages
- Allow Special:Allpages in robots.txt
- Allow Special:Allpages to be followed, but not indexed in <meta>
- Include a link saying "Local Sitemap" in the global footer pointing to Special:AllPages (there's additional caching for global footer contents, so the change might not be instanteneous)

Additionally:
- Mark the previous/next page on the page listing with rel="nofollow"


saipetch:
LGTM

PR #9878, opened 2016-03-03
PLATFORM-1949: Standardize tracing mechanics

wladekb:
Huh, these commits do (unfortunately) a few things:
- unify the terminology to use trace_id in places where we had request_id and correlation_id (this may break existing queries and dashboards, let me know about such cases if you find any)
- scan HTTP headers for the standard ones we defined to be sent from service to service
- append this information to all the logs
- extend WikiaLogger's WebProcessor to support dynamically updated contexts

https://wikia-inc.atlassian.net/browse/PLATFORM-1949
https://wikia-inc.atlassian.net/browse/PLATFORM-1948

/cc @macbre @owend @pchojnacki @jcellary 
- I'd like your input and comments on this, feel free to suggest anything as it's in the early stage:
  Am I missing something? Should I do something in a different way? High-level input is also welcome.


includes/wikia/transaction/TransactionTraceNewrelic.php
```diff
@@ -38,7 +38,8 @@ function __construct() {
 
 		// report request ID as a custom request parameter for easier debugging of slow transactions
 		if ( function_exists( 'newrelic_add_custom_parameter' ) ) {
-			newrelic_add_custom_parameter( 'requestId', RequestId::instance()->getRequestId() );
+			newrelic_add_custom_parameter( 'requestId', WikiaTracer::instance()->getTraceId() );
```
macbre: Do we want to keep both custom NewRelic params here? Or is `requestId` going to stay here just for a while?

wladekb: I thought of just a transition period only. I didn't want to break anything that could rely on this but actually it's unlikely something is consuming that information automatically so I could remove it now. Would you agree?

owend: I don't think anything is "automatically" consuming this.  So if the field changes everywhere, it should be fine to change it here too.

includes/HttpFunctions.php
```diff
@@ -506,16 +506,11 @@ public function execute() {
...
-		// pass Request ID to internal requests
-		$this->setHeader( Wikia\Util\RequestId::REQUEST_HEADER_NAME, Wikia\Util\RequestId::instance()->getRequestId() );
-		// PLATFORM-1473: pass X-Wikia-Internal-Request
-		$this->setHeader( WebRequest::WIKIA_INTERNAL_REQUEST_HEADER, 'mediawiki' );
-
 		// Wikia change - begin - @author: wladek
-		// Append extra headers for internal requests, currently only X-Request-Origin-Host
-		if ( $this->internalRequest ) {
-			$this->setHeader( Wikia\Util\RequestId::REQUEST_HEADER_ORIGIN_HOST, wfHostname() );
+		$tracerInstance = \Wikia\Tracer\WikiaTracer::instance();
```
owend: This API seems a bit complex, forcing us to use the public/internal logic outside the class.  Since setting headers is one of the things WikiaTracer is for, why not a helper function to do this.  Since setHeader just uses $reqHeader which is an array, we could pass that in as a variable.

lib/Wikia/src/Tracer/WikiaTracer.php
```diff
@@ -0,0 +1,214 @@
...
+	public function updateFromMediawiki() {
+		global $wgRequest, $wgUser;
+		if ( is_null( $this->clientIp ) && $wgRequest && $wgRequest->getIP() ) {
+			$this->clientIp = $wgRequest->getIP();
+		}
+		if ( is_null( $this->clientBeaconId ) && $this->getRequestHeader( self::LEGACY_BEACON_HEADER_NAME ) ) {
+			$this->clientBeaconId = $this->getRequestHeader( self::LEGACY_BEACON_HEADER_NAME );
+		}
+		if ( is_null( $this->userId ) && $wgUser && $wgUser->isLoggedIn() ) {
+			$this->userId = (string)$wgUser->getId();
```
macbre, line 149: Any reason why this is casted to a string?

wladekb, line 149: According to our API guidelines all IDs should be strings, just in case we decide we need to put some additional info in something that used to be plain number.

extensions/wikia/IndexingPipeline/ConnectionBase.class.php
```diff
@@ -45,7 +45,7 @@ public function publish( $routingKey, $body ) {
 					'delivery_mode' => self::DURABLE_MESSAGE,
 					'expiration' => self::MESSAGE_TTL,
 					'app_id' => 'mediawiki',
-					'correlation_id' => RequestId::instance()->getRequestId(),
+					'trace_id' => WikiaTracer::instance()->getTraceId(),
```
macbre: I'd keep the old name here - `correlation_id` is a standard name for [AMQPMessage property](https://github.com/Wikia/app/blob/master/lib/composer/videlalvaro/php-amqplib/doc/AMQPMessage.md)

lib/Wikia/src/Tracer/WikiaTracer.php
macbre, line 149: Ok.

https://wikia-inc.atlassian.net/wiki/display/PLAT/Logging does not specify this requirement explicitly. Worth updating this doc?

macbre:
LGTM

lib/Wikia/src/Logger/WebProcessor.php
```diff
@@ -33,11 +37,31 @@ public function __invoke(array $record) {
...
+	 * @param ContextSource $context
+	 * @param string $type
+	 * @return int amount of context arrays for the current type, including $context
+	 */
+	public function pushContextSource(ContextSource $contextSource, $type=self::RECORD_TYPE_CONTEXT) {
 		if (!isset($this->contextStack[$type])) {
 			$this->contextStack[$type] = [];
 		}
+		if (is_null($this->listener)) {
+			$this->listener = function($cs) {
```
macbre: Let's provide class of `$cs` argument here.

PR #9880, opened 2016-03-03
PLATFORM-1861: fix php stricts

macbre:
[PLATFORM-1861](https://wikia-inc.atlassian.net/browse/PLATFORM-1861)

Fix `PHP Strict Standards: Declaration of X should be compatible with` warnings as we're preparing for PHP7 migration.

@wladekb


wladekb:
lgtm

PR #9881, opened 2016-03-03
SUS-254: fix notifications for admins/owners

owend:
- https://wikia-inc.atlassian.net/browse/SUS-254

Don't like the duplicated logic here, but it fixes the problem...

@Wikia/sustaining-team


ghost:
:+1:

PR #9883, opened 2016-03-03
SUS-160: use POST for ajax calls which have effects

owend:
https://wikia-inc.atlassian.net/browse/SUS-160

Just did a quick grep/code-review pass to find all instances. 

@Wikia/sustaining-team


ghost:
:+1: `sed` but true :-)

ukaka:
Passed QA. Selenium tests for Special:Forum also pass.
LGTM.

PR #9884, opened 2016-03-04
SUS-281 Handle retrieve responses with status MoreDataAvailable

fraszczakszymon:
Changes:
- handle retrieve responses that should return more than 2500 objects (with status `MoreDataAvailable`)
- prevent logging request/response body to elasticsearch (they were too big, up to 6MB)
- change adapters interface to accept only arrays

@Wikia/sustaining-team


ghost:
:+1:

PR #9887, opened 2016-03-04
(CE-2836) Force the use of user Wikia for edits when reverting flags

Grunny:
Force the use of user Wikia for edits to pages when restoring flag templates.

/cc @lukaszkonieczny


lukaszkonieczny:
LGTM

PR #9889, opened 2016-03-04
RECIRC 66: Setup experiment to test recirculation placement

pauloslund:


extensions/wikia/Recirculation/js/helpers/ContentLinksHelper.js
```diff
@@ -0,0 +1,140 @@
...
+		}
+	}
+
+	function findTopTitles($links) {
+		var links = buildLinks($links),
+			titles = getSortedKeys(links),
+			topTitles = [],
+			i = 0;
+
+		while (topTitles.length < 3) {
```
gabrys: `return titles.slice(0, 3);` should work equally good

extensions/wikia/Recirculation/js/helpers/ContentLinksHelper.js
```diff
@@ -0,0 +1,140 @@
...
+			if (validLink(element)) {
+				links[element.title] = links[element.title] || 0;
+				links[element.title] ++;
+			}
+		});
+
+		return links;
+	}
+
+	function getSortedKeys(obj) {
```
gabrys: this should work: `return Object.keys(obj).sort()`

extensions/wikia/Recirculation/js/helpers/ContentLinksHelper.js
```diff
@@ -0,0 +1,140 @@
...
+			return false;
+		}
+
+		return true;
+	}
+
+	function buildLinks($links) {
+		var links = [];
+
+		$links.each(function(index, element) {
```
gabrys: $links.filter(validLink)

extensions/wikia/Recirculation/js/helpers/ContentLinksHelper.js
```diff
@@ -0,0 +1,140 @@
...
+			if (validLink(element)) {
+				links[element.title] = links[element.title] || 0;
```
gabrys: Consider saving both the value and the title as the value:

```
links[element.title] = links[element.title] || (0, element.title);
```

Then, sorting this is very easy:

```
links.sort(function (a, b) { return b[0] - a[0]; })
```

And then extracting the second property:

```
links.sort(function (a, b) { return b[0] - a[0]; }).map(function (a) { return a[1];  })
```

And then getting the first 3 elements:

```
links.sort(function (a, b) { return b[0] - a[0]; }).map(function (a) { return a[1];  }).slice(0, 3)
```

extensions/wikia/Recirculation/js/helpers/FandomHelper.js
```diff
@@ -0,0 +1,63 @@
...
+		return deferred.promise();
+	}
+
+	function formatData(data) {
+		var items = [];
+
+		$.each(data.posts, function(index, item) {
+			if (items.length < 3) {
+				item.thumbnail = item.image_url;
+				items.push(item);
```
gabrys: trailing white space

extensions/wikia/Recirculation/js/helpers/FandomHelper.js
```diff
@@ -0,0 +1,63 @@
...
+			}
+		});
+
+		return {
+			title: data.title,
+			items: items
+		};
+	}
+
+	function injectHtml(type, element) {
```
gabrys: Doesn't seem to be used

extensions/wikia/Recirculation/js/views/incontent.js
```diff
@@ -0,0 +1,108 @@
...
+	/**
+	 * Build the sections array
+	 *
+	 * DOM/layout querying: OK
+	 * DOM modification:    forbidden
+	 *
+	 * @param {jQuery} $headers headers dividing the article to sections
+	 * @returns {Section[]}
+	 */
+	function buildSections($headers) {
```
gabrys: This code generates more info that's needed to get the first suitable section, I would inline it in there and remove unneeded code.

extensions/wikia/Recirculation/js/views/incontent.js
```diff
@@ -0,0 +1,108 @@
...
+			$start,
+			$end,
+			section;
+
+		for (i = 0, len = $headers.length; i < len + 1; i += 1) {
+			intro = (i === 0);
+			$start = !intro && $headers.eq(i - 1);
+			$end = $headers.eq(i);
+			section = {
+				intro: intro,
```
gabrys: ... for example this `intro` property is not used. I think you just need to find first all-width `<h2>` header and insert your module above/below it.

extensions/wikia/Recirculation/js/views/rail.js
```diff
@@ -0,0 +1,46 @@
...
+	'ext.wikia.recirculation.tracker',
+	'ext.wikia.recirculation.utils'
+], function ($, w, log, Mustache, tracker, utils) {
+
+	var logGroup = 'ext.wikia.recirculation.views.rail';
+
+	function render(data) {
+		var deferred = $.Deferred();
+
+		utils.loadTemplate('extensions/wikia/Recirculation/templates/client/rail.mustache')
```
gabrys: The whole utils.loadTemplate pattern seems repeated in the view. Is there a simple way to avoid this duplication?

extensions/wikia/Recirculation/styles/discussions.scss
```diff
@@ -0,0 +1,181 @@
+@import 'skins/shared/mixins/flexbox';
+@import 'skins/shared/color';
+
+// DISCUSSIONS
+$discussion-height: 348px;
```
gabrys: Isn't this variable to generic? Shouldn't it be more `recirculation-discussion` or something like that?

extensions/wikia/Recirculation/templates/client/footer.mustache
```diff
@@ -0,0 +1,14 @@
+<div class="recirculation-horizontal recirculation-footer">
+	<h3>{{ title }}</h3>
+
+	<div class="items">
+		{{#items}}
+		<div class="item">
+			<a title={{ title }} href="{{ url }}">
```
gabrys: `"{{ title }}"`

extensions/wikia/Recirculation/templates/client/incontent.mustache
```diff
@@ -0,0 +1,14 @@
+<div class="recirculation-horizontal recirculation-incontent">
+	<h3>{{ title }}</h3>
+
+	<div class="items">
+		{{#items}}
+		<div class="item">
+			<a title={{ title }} href="{{ url }}">
```
gabrys: `"{{title}}"`

extensions/wikia/Recirculation/templates/client/footer.mustache
```diff
@@ -0,0 +1,14 @@
...
+				<div class="thumbnail"><img src="{{ thumbnail }}" /></div>
```
gabrys: I don't think we're using the self-closing XHTML-style tags anymore. Plus I think `alt` property is still required for img tags.

extensions/wikia/Recirculation/templates/client/incontent.mustache
```diff
@@ -0,0 +1,14 @@
...
+				<div class="thumbnail"><img src="{{ thumbnail }}" /></div>
```
gabrys: I don't think we're using the self-closing XHTML-style tags anymore. Plus I think `alt` property is still required for img tags.

extensions/wikia/Recirculation/js/helpers/ContentLinksHelper.js
pauloslund: Added this filter earlier on in the process

extensions/wikia/Recirculation/js/helpers/FandomHelper.js
pauloslund: The place it was being used had a type... :(

gabrys:
LGTM :panda_face:

PR #9892, opened 2016-03-07
ADEN-3064 No breakpointsLayout module on Mercury

drets:
Hotfix ads on Mercury
(we don't have `breakpointsLayout` module on Mercury - so make it optional)


jacekjursza:
LGTM (but please check the test results)

nandy-andy:
The failed PHP unit tests seems like an issue with recent changes to user preferences. I've contacted @jcellary to find out if they're aware of that.

PR #9900, opened 2016-03-07
[PLATFORM-1985] Introduced shared memcache for permissions data

jcellary:
Changed caching of user groups to be shared between envs (sandbox/previe/prod) etc. to avoid problem that someone edit someone rights on sandbox and in prod that user doesn't have the new group, as it was cached without it and purging only affected one env.

@wladekb


wladekb:
lgtm

PR #9902, opened 2016-03-07
SEO-302 Allow Googlebot to crawl Android apps

gabrys:
Given APIs are for robots, we should just add an explicit Allow rule for
api.php that overrides the later generic block for the action param
(meant for blocking accessing ?action=edit ?action=purge, ?action=raw
etc).

Now because Googlebot doesn't follow standard, which tells that the first
matching rule wins, we also need to add a "more specific" rule to make
Googlebot crawl api.php URLs with "action" param.

Therefore we're adding code that causes the following two rules to be
emitted in robots.txt:

```
Allow: /api.php?
Allow: /api.php?*action=
```

The first one is meant for standards-compliant bots (enabling them to
use any api.php URLs). The second is just for Googlebot (to win over
Disallow /_?_action= rule).

### DIFF
--- a/extensions/wikia/RobotsTxt/RobotsTxt.class.php
+++ b/extensions/wikia/RobotsTxt/RobotsTxt.class.php
@@ -32,6 +32,17 @@ class RobotsTxt {
 	}
 
 	/**
+	 * Allow a specific path
+	 *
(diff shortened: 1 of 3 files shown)


saipetch:
LGTM

PR #9905, opened 2016-03-08
Aden 2983 unused slots

jacekjursza:
Cleaning up code after reviewing unused ad - slots.

### DIFF
--- a/extensions/wikia/Lightbox/scripts/Lightbox.js
+++ b/extensions/wikia/Lightbox/scripts/Lightbox.js
@@ -76,13 +76,6 @@
 			// cache re-used DOM elements and templates for this modal instance
 			Lightbox.cacheDOM();
 
-			// Init ads in Lightbox
-			if ($('#MODAL_RECTANGLE').length && Lightbox.ads.showAds()) {
-				Lightbox.openModal.lightbox.addClass('show-ads');
-				window.adslots2.push('MODAL_RECTANGLE');
-				Lightbox.ads.adModalRectangleShown = true;
-			}
-
 			// Set up carousel
 			Lightbox.setUpCarousel();
 
@@ -536,8 +529,6 @@
 			}
 		},
 		ads: {
-			// is MODAL_RECTANGLE ad shown?
-			adModalRectangleShown: false,
 			// preload ad after this number of unique images/videos are shown
 			adMediaCountPreload: 2,
 			// show an ad after this number of unique images/videos are shown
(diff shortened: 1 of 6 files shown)


extensions/wikia/Lightbox/scripts/Lightbox.js
```diff
@@ -855,7 +846,7 @@
 			var types = Lightbox.carouselTypes,
 				deferredList = [],
 				// show-ads class appears when there is going to be a MODAL_RECTANGLE ad
```
drets: Should we remove comment above?

extensions/wikia/Lightbox/scripts/Lightbox.js
```diff
@@ -76,13 +76,6 @@
 			// cache re-used DOM elements and templates for this modal instance
 			Lightbox.cacheDOM();
 
-			// Init ads in Lightbox
-			if ($('#MODAL_RECTANGLE').length && Lightbox.ads.showAds()) {
-				Lightbox.openModal.lightbox.addClass('show-ads');
```
drets, line 81: Maybe we also need to remove: https://github.com/Wikia/app/blob/ADEN-2983-unusedSlots/extensions/wikia/Lightbox/styles/Lightbox.scss#L402-L435 ?

drets:
I am not sure if it's worth to touch old liftium and remove those guys: https://github.com/Wikia/app/blob/ADEN-2983-unusedSlots/extensions/wikia/AdEngine/js/provider/liftium.js#L24-L25 ?

extensions/wikia/Lightbox/scripts/Lightbox.js
jacekjursza, line 81: I'm not sure. Aren't it used for the second ad in the lightbox?

jacekjursza:
Maybe let's wait for it with full Liftium cleanup?

extensions/wikia/Lightbox/scripts/Lightbox.js
jacekjursza: Good catch, thx ;)

drets:
I agree that we can wait for full Liftium cleanup but I am curious what @nandy-andy thinks about it :wink:

extensions/wikia/Lightbox/scripts/Lightbox.js
drets, line 81: I'd say it was used only by `MODAL_RECTANGLE` but, definitely, it's worth to do happy-path testing (after removing those styles).

drets:
If we shouldn't remove styles which may be related to `MODAL_RECTANGLE` rectangle ad - LGTM :+1:

drets:
LGTM :+1:

PR #9908, opened 2016-03-08
VOLDEV-129: Add i18n for Closed Wikia links

mszabo-wikia:
Introduced closed-wiki-policy-url message for specifying different URLs for different languages.
@Grunny

### DIFF
--- a/extensions/wikia/WikiFactory/Close/templates/close-info.tmpl.php
+++ b/extensions/wikia/WikiFactory/Close/templates/close-info.tmpl.php
@@ -26,7 +26,7 @@ if ( $bShowDumps ) {
 <table width="90%" align="center" id="closed-urls">
     <tr>
         <td width="50%" style="text-align:center"><a href="http://www.wikia.com/Special:CreateWiki"><?= wfMessage( 'closed-wiki-create-wiki' )->parse() ?></a></td>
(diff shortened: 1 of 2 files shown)


extensions/wikia/WikiFactory/Close/templates/close-info.tmpl.php
```diff
@@ -26,7 +26,7 @@
 <table width="90%" align="center" id="closed-urls">
     <tr>
         <td width="50%" style="text-align:center"><a href="http://www.wikia.com/Special:CreateWiki"><?= wfMessage( 'closed-wiki-create-wiki' )->parse() ?></a></td>
-        <td width="50%" style="text-align:center"><a href="/wiki/Wikia:Closed_Wikia"><?= wfMessage( 'closed-wiki-policy' )->escaped() ?></a></td>
+        <td width="50%" style="text-align:center"><a href="<?= wfMessage( 'closed-wiki-policy-url' )->escaped() ?>"><?= wfMessage( 'closed-wiki-policy' )->escaped() ?></a></td>
```
Grunny: Since it's setting the URL, as a best practice it is good to use `->useDatabase( false )` so it can't be overridden locally (though not as problematic now that local messages can't be edited unless whitelisted).

mszabo-wikia:
@Grunny good catch, fixed :)

Grunny:
LGTM.

PR #9909, opened 2016-03-08
PLATFORM-1199: clean up shared wiki data when running Close/maintenance.php

macbre:
[PLATFORM-1199](https://wikia-inc.atlassian.net/browse/PLATFORM-1199)

Thanks to this pull requests we will clean up the stats-related data when running `Close/maintenance.php` script and avoid issues like [PLATFORM-1849](https://wikia-inc.atlassian.net/browse/PLATFORM-1849).

Once this one is deployed we can get rid of `eventsCleanup.php` script run via cron (as it does the same job).

@wladekb . @mixth-sense


extensions/wikia/WikiFactory/Close/maintenance.php
```diff
@@ -460,6 +440,62 @@ private function removeBucket( $cityid ) {
...
+	 * @throws MWException
+	 */
+	private function doTableCleanup( DatabaseBase $db, $table, $city_id, $wiki_id_column = 'wiki_id' ) {
+		$db->delete( $table, [ $wiki_id_column => $city_id ], __METHOD__ );
+
+		$this->log( sprintf( "#%d: removed %d rows from %s.%s table", $city_id, $db->affectedRows(), $db->getDBname(), $table ) );
+
+		// throttle delete queries
+		if ( $db->affectedRows() > 0 ) {
+			wfWaitForSlaves( $db->getDBname() );
```
wladekb, line 494: why do we need it here?

wladekb:
lgtm, 1 inlined question only

extensions/wikia/WikiFactory/Close/maintenance.php
macbre, line 494: The same approach is used in `eventsCleanup` script. We do not want to cause slave lag by massive delete queries affecting many rows.

wladekb, line 494: lgtm

PR #9910, opened 2016-03-08
Bucky: report last two domain segments only

macbre:
[PLATFORM-1903](https://wikia-inc.atlassian.net/browse/PLATFORM-1903)

Reporting 3rd party assets stats increased the traffic that we generate to InfluxDB. Let's improve grouping of domain names by reporting only the last two segments: `vignette1.wikia.nocookie.net` will be reported as `nocookie.net`.

@nandy-andy / @wladekb


wladekb:
lgtm

PR #9912, opened 2016-03-08
PLATFORM-1938: extract our custom changes in Monolog

macbre:
#8684 introduced custom changes in Monolog. Let's extract them and introduce `Status` object formatter.

Example:

``` php
> Wikia\Logger\WikiaLogger::instance()->info( 'Message', [ 'foo' => \Status::newFatal( 'error message' ) ]);
```

``` json
{
  "@timestamp": "2016-03-08T15:47:01.143529+00:00",
  "@message": "Message",
  "@fields": {
    "db_name": "wikia",
    "city_id": "177",
    "maintenance_file": "/usr/wikia/source/app/maintenance/eval.php",
    "maintenance_class": "CommandLineInc",
    "request_id": "mw56def3f47f81b0.83588911"
  },
  "@context": {
    "foo": {
      "is_ok": false,
      "errors": [
        [
          "error message"
        ]
      ],
      "message": "&lt;error message&gt;",
      "value": null
    }
  }
}
```

The next step here is the upgrade of Monolog to the most recent version (https://github.com/Seldaek/monolog/commit/b941eac988decda854947b297c8357c5dac612e8) - see [PLATFORM-1938](https://wikia-inc.atlassian.net/browse/PLATFORM-1938) - will be required by PHP7.

@jcellary / @wladekb / @mixth-sense


lib/Wikia/src/Logger/WikiaLogger.php
```diff
@@ -212,11 +212,16 @@ public function setWebProcessor(WebProcessor $processor) {
...
 	public function defaultLogger() {
-		return new Logger(
+		$logger = new Logger(
 			'default',
 			[$this->getSyslogHandler()],
 			[$this->getWebProcessor()]
 		);
+
+		// custom formatting of MediaWiki's Status class
+		$logger->pushProcessor( new StatusProcessor() );
```
wladekb: . 1. 
AFAIK this can be simplified to the form of:

```
$logger = new Logger(
  'default',
  [$this->getSyslogHandler()],
  [$this->getWebProcessor(), new StatusProcessor()]
);
```

. 2. all the other required objects got their own `getX()` methods that create them if necessary. I'd suggest to stick with this convention in this file

macbre: Addressed in 1f950f159c29d3a8b2c8b27c1f1589887db0097a

wladekb:
lgtm

PR #9917, opened 2016-03-08
[PLATFORM-1991 Expanded groups in SpecialListGroupRights with implicit groups

jcellary:
This fix is adding showing of rights associated with implicit groups on the ListGroupRights special page. They were being shown in the past, but accidentally hidden away in a recent change.

@wladekb

### DIFF
--- a/includes/specials/SpecialListgrouprights.php
+++ b/includes/specials/SpecialListgrouprights.php
@@ -60,7 +60,10 @@ class SpecialListGroupRights extends SpecialPage {
 				'</tr>'
 		);
 
-		$allGroups = User::getAllGroups();
+		$allGroups = array_unique( array_merge(
+			User::getAllGroups(),
+			User::getImplicitGroups()
+		) );
 		asort( $allGroups );
 
 		foreach ( $allGroups as $group ) {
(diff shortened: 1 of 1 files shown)


includes/specials/SpecialListgrouprights.php
```diff
@@ -60,7 +60,10 @@ public function execute( $par ) {
 				'</tr>'
 		);
 
-		$allGroups = User::getAllGroups();
+		$allGroups = array_unique( array_merge(
+			User::getAllGroups(),
```
wladekb: `getAllGroups` suggests it returns all groups and not (I presume) "explicit" only
or I'm missing something?

jcellary: You are right, I had the same thought - that is exactly the reason why in the new internal service I renamed it to getExplicitGroups. I didn't want to change the API of the User class however to avoid problems and the need to refactor all places that use it to create even more havoc, so this name is simply in the form it existed in the past.

What I can do here however is to use the new service instead of the User class, which has much better naming and decrease the number of dependencies on the old badly named User API

wladekb: Oh, I missed this method is called on the `User` class.

wladekb:
lgtm

PR #9918, opened 2016-03-09
SUS-183: Return null if getApiWrapper returns null

inez:
Method getApiWrapper internally tries to figure our video provider and video id - if it fails it means that provided URL is incorrect - so we should not attempt to "uplod" such a broken URL.


owend:
Looks like ApiWrapperFactory::getApiWrapper can return false or null so probably better to make an empty() check here or refactor that function to consistently return null.

inez:
@owend Good catch!

inez:
Ready for CR again.

owend:
LGTM

PR #9921, opened 2016-03-09
SUS-196: check for readonly mode in addVideo method

owend:
ninja squirrel grabbing another ticket from @mixth-sense :shipit: 

wfMsg is deprecated but the rest of the function uses it.  If this bugs you ping me and I'll fix them all.


ghost:
:+1: LGTM

As a side note: I can hardly express, how much I love functions that change the state instead of returning values. They are so cute.

As for messaging stuff, I'll create a general ticket for SUS to review all code we own and replace `wfMsg` functions family with `wfMessage`.

ukaka:
LGTM. Passed QA.

PR #9922, opened 2016-03-09
remove wgVideoHandlersVideosMigrated

owend:
I was doing code review for removing variables and noticed that this one is enabled everywhere.  Removing the variable entirely allows us to also delete it from configuration.  

@saipetch @mixth-sense @wladekb 

This var is defined in wikifactory (as true in all cases) so it can be deleted from the db as well (but I haven't done that part yet)


saipetch:
LGTM

PR #9923, opened 2016-03-09
SUS-74 Prevent updating video thumbnail from foreign wiki

fraszczakszymon:
https://wikia-inc.atlassian.net/browse/SUS-74

@mixth-sense


ghost:
We're missing unit tests here. @fraszczakszymon - let's write them tomorrow. Unit tests will give all of us better understanding of the code as they are a different way of phrasing the contract the code must meet. It's too much stuff to proceed without tests.

PR #9924, opened 2016-03-09
PLATFORM-1725 Division by zero in imageServing.class

RafalWilinski:
Simple prevention of dividing by zero. Should not cause the same issue that `max(1, proportion['w'])` caused (-1 was used for keeping original aspect ratio).


macbre:
Can we add unit tests here? :)

RafalWilinski:
Sure, I was also thinking about it.

RafalWilinski:
ping @macbre

macbre:
:+1:

PR #9934, opened 2016-03-09
SOC-2114 | Helios antispoof check now skips exact match on username

kvas-damian:
https://wikia-inc.atlassian.net/browse/SOC-2114

**Problem**
Helios returns 2 different error message when username is already taken

**Reason**
Helios has many validators for username, including:
- internal check if username is already taken (exact match check)
- antispoof check using Wikia API

**Solution**
Antispoof API method provided for Helios skips exact match username when checking for username conflicts. 

Ping @harnash

### DIFF
--- a/extensions/AntiSpoof/SpoofUser.php
+++ b/extensions/AntiSpoof/SpoofUser.php
@@ -45,22 +45,32 @@ class SpoofUser {
 	/**
 	 * Does the username pass Unicode legality and script-mixing checks?
 	 *
+	 * @param $skipExactMatch - when true exact match username is not returned
+	 *
 	 * @return array empty if no conflict, or array containing conflicting user names
 	 */
-	public function getConflicts() {
+	public function getConflicts( $skipExactMatch = false ) {
 		$dbr = self::getDBSlave();
 
 		// Join against the user table to ensure that we skip stray
 		// entries left after an account is renamed or otherwise munged.
-		/* Wikia Change - begin : Quote the table names, otherwise the select method
-		   tries to prefix the tablenames with the current user DB */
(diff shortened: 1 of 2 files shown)


extensions/AntiSpoof/SpoofUser.php
```diff
@@ -45,22 +45,32 @@ public function getNormalized() {
...
-		   tries to prefix the tablenames with the current user DB */
+		/* Wikia Change - begin :
+			Quote the table names, otherwise the select method
+				tries to prefix the tablenames with the current user DB
+			Added $skipExactMatch */
+		$conds = array(
+			'su_normalized' => $this->mNormalized,
+			'su_name=user_name',
+		);
+		if ( $skipExactMatch ) {
```
kvas-damian, line 65: This method is used during user rename process, so we need to change it conditionally.

garthwebb:
Ok, LGTM

rafalkalinski:
LGTM

extensions/wikia/Helios/HelperController.class.php
```diff
@@ -43,7 +43,7 @@ public function checkAntiSpoof()
 		$spoofUser = new \SpoofUser( $username );
 
 		// Allow the given user name if it is legal and does not conflict with other names.
-		$this->response->setVal( 'allow', $spoofUser->isLegal() && ! $spoofUser->getConflicts() );
+		$this->response->setVal( 'allow', $spoofUser->isLegal() && ! $spoofUser->getConflicts( true ) );
```
bkoval: I don't think there's a need for a space after exclamation mark :)

sqreek:
LGTM

PR #9935, opened 2016-03-09
ADEN-3061 add logging

jacekjursza:
Adding logging on how often we init Krux


nandy-andy:
LGTM :+1:

drets:
I'd rename `wikiaTracker` to `tracker` for keeping [consistency](https://github.com/Wikia/app/blob/ADEN-3061-kruxLogging/extensions/wikia/AdEngine/js/recovery/message.js#L48-L55) and it looks like [value](https://github.com/Wikia/app/blob/ADEN-3061-kruxLogging/resources/wikia/modules/krux.js#L75) is redundant.

But it's really minor issues, so LGTM :+1:

PR #9936, opened 2016-03-09
ADEN-3015 Change adType=collapse result to be tracked as collapse …

fraszczakszymon:
... instead of success

https://wikia-inc.atlassian.net/browse/ADEN-3015


jacekjursza:
lgtm

PR #9939, opened 2016-03-09
ResourceLoaderUserModule: add an assert to find the root cause of PLATFORM-1980

macbre:
[PLATFORM-1980](https://wikia-inc.atlassian.net/browse/PLATFORM-1980)

`Title::makeTitleSafe` is getting not valid input - let's add an assert here.

`PHP Fatal Error: Call to a member function getPrefixedDBkey() on null in /includes/resourceloader/ResourceLoaderUserModule.php on line 40`

@wladekb

### DIFF
--- a/includes/resourceloader/ResourceLoaderUserModule.php
+++ b/includes/resourceloader/ResourceLoaderUserModule.php
@@ -37,6 +37,12 @@ class ResourceLoaderUserModule extends ResourceLoaderGlobalWikiModule {
 			// Get the normalized title of the user's user page
 			$username = $context->getUser();
 			$userpageTitle = Title::makeTitleSafe( NS_USER, $username );
+
+			// PLATFORM-1980
+			Wikia\Util\Assert::true( $userpageTitle instanceof Title, __METHOD__, [
+				'username' => $username
+			] );
+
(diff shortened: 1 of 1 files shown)


macbre:
Travis failed once again with

```
npm ERR! EEXIST, symlink '../esprima/bin/esparse.js'
File exists: ../esprima/bin/esparse.js
Move it away, and try again. 
```

@ludwikkazmierczak, do we know who actually owns tests run on Travis?

includes/resourceloader/ResourceLoaderUserModule.php
```diff
@@ -37,6 +37,12 @@ protected function getPages( ResourceLoaderContext $context ) {
 			// Get the normalized title of the user's user page
 			$username = $context->getUser();
```
wladekb, line 38: `$context->getUser()` return the unchecked value of the "user" GET parameter in the URL. Anyone can craft a request that would be an invalid title in Mediawiki world. However this can also come from some bad escaping thereof or a bug in `makeTitleSafe`. Let's then find the root cause... :)

wladekb:
lgtm

includes/resourceloader/ResourceLoaderUserModule.php
macbre, line 38: Agree. Too much information is better than none :)

PR #9940, opened 2016-03-09
XW-1206 | prepare getDimensions API

tomasznapieralski:
https://wikia-inc.atlassian.net/browse/XW-1206

@Wikia/x-wing


extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -221,6 +221,50 @@ public function getWikiVariables() {
...
+		$article = Article::newFromID( $title->getArticleId() );
+
+		if ( $article instanceof Article && $title->isRedirect() ) {
+			$title = $this->handleRedirect( $title, $article, [] )[0];
+		}
+
+		$wikiCategoryNames = WikiFactoryHub::getInstance()->getWikiCategoryNames( $wgCityId );
+		$wikiCategoryNames = join( ',', $wikiCategoryNames );
+
+		$adContext = ( new AdEngine2ContextService() )->getContext( $title, 'mercury' );
```
rogatty, line 239: We use `adsContext` everywhere else. What do you think about unifying it?

tomasznapieralski, line 239: only Mercury uses `ads`, not `ad` context, i can unify this

rogatty, line 239: Oh, you're right! Sorry for not double checking. In that case I wouldn't bother ;)

tomasznapieralski, line 239: :+1:

extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -221,6 +221,50 @@ public function getWikiVariables() {
...
+		$dimensions[17] = WikiFactoryHub::getInstance()->getWikiVertical( $wgCityId )['short'];
+		$dimensions[18] = $wikiCategoryNames;
+		$dimensions[19] = WikiaPageType::getArticleType( $title );
+
+		$dimensions[23] = in_array( 'poweruser_lifetime', $powerUserTypes ) ? 'yes' : 'no';
+		$dimensions[24] = in_array( 'poweruser_frequent', $powerUserTypes ) ? 'yes' : 'no';
+		$dimensions[25] = strval( $title->getNamespace() );
+
+		$this->response->setVal( 'dimensions', $dimensions );
+		$this->response->setContentType( 'application/javascript; charset=utf-8' );
```
hakubo: we should not send some custom script
it does sound error prone and wierd

in Mercury we should already have jquery
so we could call this API with $.getJson

vforge: But then we can't use script.js for this script - and as script.js aren't using regular promises, we couldn't load it in pararel with other scripts. Maybe you have some idea how to fix it?

bognix:
LGTM

bognix:
LGTM

PR #9941, opened 2016-03-09
MAIN-6612 / SUS-293: add missing "use" statement

owend:
Re-apply fix for SUS-158 with proper "use" statement. 
Problem can be replicated in dev with instructions in ticket

@mixth-sense @macbre  @Wikia/sustaining-team

### DIFF
--- a/includes/wikia/services/ApiService.class.php
+++ b/includes/wikia/services/ApiService.class.php
@@ -114,7 +114,6 @@ class ApiService extends Service {
 	 * @return array $options
 	 */
 	public static function loginAsUser() {
-		global $wgEnableHeliosExt;
 		$app = F::app();
 
 		$options = array();
@@ -135,8 +134,8 @@ class ApiService extends Service {
 			$cookie .= $app->wg->CookiePrefix.$key.'='.$value.';';
 		}
 
-		if ( $wgEnableHeliosExt ) {
-			$token = \Wikia\Helios\User::getAccessToken($app->wg->Request);
+		if ( $app->wg->EnableHeliosExt ) {
+			$token = \Wikia\Helios\User::getAccessToken( $app->wg->Request );
 			if ( !empty($token) ) {
 				$cookie .= HeliosCookieHelper::ACCESS_TOKEN_COOKIE_NAME . "=" . $token . ";";
 			}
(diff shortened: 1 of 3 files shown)


includes/wikia/services/ApiService.class.php
```diff
@@ -112,6 +114,7 @@ private static function getHostByDbName( $dbName ) {
 	 * @return array $options
 	 */
 	public static function loginAsUser() {
+		global $wgEnableHeliosExt;
```
ghost: Note the `$app = F::app()` in the following line. Using `$app->wg->EnableHeliosExt` later seems more convenient that the `global` reference. And consistent with the rest of the code refering to `wgCookiePrefix` and `wgRequest`.

Also, would it be overengineering if we referred to `wgExternalAuthType` or something similar rather than to Helios specifically? Probably yes, since we refer to `HeliosCookieHelper` class constant anyway. So, just a thought.

includes/wikia/services/ApiService.class.php
```diff
@@ -132,6 +135,12 @@ public static function loginAsUser() {
 			$cookie .= $app->wg->CookiePrefix.$key.'='.$value.';';
 		}
 
+		if ( $wgEnableHeliosExt ) {
```
ghost: We are mixing two whitespace conventions in the following block. Please, unify.

includes/wikia/services/ApiService.class.php
```diff
@@ -112,6 +114,7 @@ private static function getHostByDbName( $dbName ) {
 	 * @return array $options
 	 */
 	public static function loginAsUser() {
```
ghost, line 116: Please, add unit tests covering different `$options` returned by this method in different situations. Writing tests is a different way of "phrasing" the contract the tested code is supposed to do. It's often easier to understand the code by reading its tests. :-)

owend:
Will write unit tests.  Hold off on merging until then.

includes/wikia/services/ApiService.class.php
```diff
@@ -132,6 +134,12 @@ public static function loginAsUser() {
 			$cookie .= $app->wg->CookiePrefix.$key.'='.$value.';';
 		}
 
+		if ( $app->wg->EnableHeliosExt ) {
+			$token = \Wikia\Helios\User::getAccessToken( $app->wg->Request );
+			if ( !empty($token) ) {
```
ghost: Just me being a pain in the ass about MW whitespacing conventions :-)

ghost:
:+1: LGTM

Although I am concerned about the way we test things. But that's another story way beyond the scope of this particular PR.

PR #9942, opened 2016-03-09
RECIRC-49: Setup AbTest.js to run properly in Mercury

pauloslund:
Some global variables reference are not available inside of Mercury. This PR adds alternatives


gabrys:
Scary RegExp, but otherwise it LGTM

PR #9943, opened 2016-03-09
Removed helios client credentials

ArturKlajnerok:
Since we don't implement Oauth2 and client credentials are not required this should simplify implementation and help avoid errors with credentials mismatch.

/cc @Wikia/services-team @wladekb @owend @mixth-sense


wladekb:
seems legit

nmonterroso:
LGTM

PR #9944, opened 2016-03-10
SEO-50 Update the title for Special:AdminDashboard

gabrys:
The title is now "AdminDashboard - AdminDashboard - Xyz Wiki - Wikia".
We want it to be "AdminDashboard - Xyz Wiki - Wikia".


gabrys:
LGTM

PR #9954, opened 2016-03-10
ADEN-3072 Nielsen tracking in oasis

nandy-andy:
As in documentation: loading library moved to the head with `WikiaSkinTopScripts` hook, wrapped in lambda function and write in the way that it's loaded asynchronously. And tracking called when library and dom are loaded


drets:
LGTM :+1:

PR #9956, opened 2016-03-10
[PLATFORM-1994] Fixed PHP notice in user permissions

jcellary:
Added a check preventing generation of a php notice that an array key is not set

@wladekb


wladekb:
lgtm

PR #9957, opened 2016-03-11
Update ApiService to support Helios user token

owend:
MAIN-6612 / SUS-293 / SUS-158

WallNotification uses ApiService::foreignCall but was not passing user credentials. 
ApiService didn't support Helios token for user credentials
This patch updates ApiService and adds unit tests

@Wikia/sustaining-team


owend:
Going to close this and move this code back under #9941

I renamed my local branch forgetting that this one was already pushed.

PR #9964, opened 2016-03-11
PLATFORM-1788: use 30s timeout for Swift's PUT requests

macbre:
[PLATFORM-1788](https://wikia-inc.atlassian.net/browse/PLATFORM-1788)

It's better than no timeout at all that causes `MigrateImagesBetweenSwiftDC` script to hang on copying files from SJC to RES

@jcellary  / @ljagiello


jcellary:
Le GTM

macbre:
Þank you :)

PR #9971, opened 2016-03-14
SOC-1496 | Rewrite force login modal into force login pop-up

Bloodmoon24:
## Links
- https://wikia-inc.atlassian.net/browse/SOC-1496
- https://github.com/Wikia/mercury/pull/2165 - Mercury part
## Description

Rewritten login/register modal into pop-up.
## Reviewers

@rafalkalinski  @bkoval


kvas-damian:
What do you think about removing not used styles: https://github.com/Wikia/app/blob/SOC-1496/extensions/wikia/AuthModal/css/AuthModal.scss ?

extensions/wikia/AuthModal/js/AuthModal.js
```diff
@@ -52,49 +35,42 @@ define('AuthModal', ['jquery', 'wikia.window'], function ($, window) {
...
-		authIframe.src = url + (url.indexOf('?') === -1 ? '?' : '&')  + modalParam;
-		//for the selenium tests:
-		authIframe.id = 'auth-modal-iframe';
-		authIframe.onload = function () {
-			if (typeof onPageLoaded === 'function') {
-				onPageLoaded();
-			}
-		};
-		modal.appendChild(authIframe);
+		authPopUp = window.open(src, '_blank', 'width='+popUpWidth+',height='+popUpHeight+',top='+popUpTop+',left='+popUpLeft);
```
kvas-damian: Spaces...

extensions/wikia/AuthModal/js/AuthModal.js
```diff
@@ -52,49 +35,42 @@ define('AuthModal', ['jquery', 'wikia.window'], function ($, window) {
...
-	}
-
 	function close (event) {
 		if (event) {
 			event.preventDefault();
 		}
-
-		if (modal) {
+		if (authPopUp) {
 			track({
```
kvas-damian: Please make sure that tracking is called only once when you call this method. See: `onbeforeunload`

Bloodmoon24: I was wondering if this is needed. Yeah `onbeforeunload` will handle all.

extensions/wikia/AssetsManager/config.php
```diff
@@ -2641,10 +2641,7 @@
 
 $config['auth_modal_scss'] = array(
 	'type' => AssetsManager::TYPE_SCSS,
-	'skin' => [ 'oasis' ],
-	'assets' => [
-		'//extensions/wikia/AuthModal/css/AuthModal.scss',
-	]
+	'skin' => [ 'oasis' ]
```
bkoval: What's the point of having this package if it doesn't contain any assets? :)

extensions/wikia/AuthModal/js/AuthModal.js
```diff
@@ -52,49 +35,38 @@ define('AuthModal', ['jquery', 'wikia.window'], function ($, window) {
...
-		authIframe.src = url + (url.indexOf('?') === -1 ? '?' : '&')  + modalParam;
-		//for the selenium tests:
-		authIframe.id = 'auth-modal-iframe';
-		authIframe.onload = function () {
-			if (typeof onPageLoaded === 'function') {
-				onPageLoaded();
-			}
-		};
-		modal.appendChild(authIframe);
+		authPopUp = window.open(src, '_blank', 'width=' + popUpWidth + ',height=' + popUpHeight + ',top=' + popUpTop + ',left=' + popUpLeft);
```
bkoval: I would consider creating a separate method like getPopupSpecs() so we can minimize a size of this method and make it clearer what's happening here. Also, modalParam, popUpHeight and popUpMaxWidth look like consts so they could be moved to the class level.

extensions/wikia/AuthModal/js/AuthModal.js
```diff
@@ -52,49 +35,38 @@ define('AuthModal', ['jquery', 'wikia.window'], function ($, window) {
...
 	}
 
-	function loadPage (url, onPageLoaded) {
-		var authIframe = window.document.createElement('iframe'),
-			modalParam = 'modal=1';
+	function loadPage (url) {
+		var modalParam = 'modal=1',
+			pageWidth = window.innerWidth,
+			popUpHeight = 670,
+			popUpMaxWidth = 768,
```
rafalkalinski: perfect candidates for a module const :)

extensions/wikia/AuthModal/js/AuthModal.js
bkoval: Also, we should consider renaming the object here to something like AuthPopUpWindow as it is an object of a type 'Window'.

extensions/wikia/AuthModal/js/AuthModal.js
```diff
@@ -52,49 +35,38 @@ define('AuthModal', ['jquery', 'wikia.window'], function ($, window) {
...
-		authIframe.id = 'auth-modal-iframe';
-		authIframe.onload = function () {
-			if (typeof onPageLoaded === 'function') {
-				onPageLoaded();
-			}
-		};
-		modal.appendChild(authIframe);
+		authPopUp = window.open(src, '_blank', 'width=' + popUpWidth + ',height=' + popUpHeight + ',top=' + popUpTop + ',left=' + popUpLeft);
 
+		if (authPopUp) {
```
bkoval: Let's also check for the 'closed' property on the newly created window, as the popup might've been opened and closed.

This part needs to be thoroughly tested in all the supported browsers, let's make sure it happens :)

extensions/wikia/AuthModal/js/AuthModal.js
```diff
@@ -52,49 +35,38 @@ define('AuthModal', ['jquery', 'wikia.window'], function ($, window) {
 		return track;
 	}
 
-	function onKeyUp (event) {
```
rafalkalinski, line 55: wouldn't it be useful with the new modal?

bkoval:
There's no point of calling this extension "AuthModal" anymore since it's not a modal now. Could we adjust the naming? :)

extensions/wikia/AuthModal/js/AuthModal.js
Bloodmoon24, line 55: No.

bkoval, line 55: A word of explanation maybe? ;)

Bloodmoon24:
@bkoval  what about `wgEnableNewAuthModal` param? Shall I change it to `wgEnableNewAuthPopUp` ? Everything or nothing.

extensions/wikia/AuthModal/js/AuthModal.js
rafalkalinski: This comment is wrapped, because of code change, but is still not addressed.

rafalkalinski:
just for a case:
https://github.com/Wikia/app/pull/9971#discussion_r56076613

bkoval:
@kvas-damian Can we easily rename a WikiFactory variable? If it's not easy and with the fact that we're aiming at removing it after increasing popup usage to 100% communties, it might not be worth it. However, if it's a straightforward change, let's do it!

kvas-damian:
Renaming wikiFacotry variable seems to be not straight forward.
We could rename it in db, but there is memcache over it, so there is no way for quick rename.
To avoid problems during renaming process we'd need to support both variables - new and old one. I believe it's safer to get back to renaming/removing this variable when we release it side-wide.

Please create a ticket for that.

extensions/wikia/AuthModal/js/AuthModal.js
Bloodmoon24: Thanks!

Bloodmoon24:
Ticket for renaming has been created https://wikia-inc.atlassian.net/browse/SOC-2171

extensions/wikia/AuthModal/js/AuthModal.js
Bloodmoon24, line 55: After a discussion with designer he said that this is new window, and it shouldn't have this functionality.

extensions/wikia/AuthModal/js/AuthModal.js
```diff
@@ -52,51 +57,23 @@ define('AuthModal', ['jquery', 'wikia.window'], function ($, window) {
...
-		}
-
-		$(window).off('.authModal');
-	}
+		authPopUpWindow = window.open(src, '_blank', getPopUpWindowSpecs());
 
-	function onPageLoaded () {
-		if (modal) {
-			$(modal).removeClass('loading');
+		if (authPopUpWindow || authPopUpWindow.closed) {
```
rafalkalinski: I would vote for **!**authPopUpWindow.closed

Bloodmoon24: Yes, indeed!

bkoval:
LGTM :package:

PR #9972, opened 2016-03-14
CE-3486 Add a method to get an anti-CSRF token and adjust the API

adamkarminski:

Since the [API:Info](https://www.mediawiki.org/wiki/API:Info) requires users to be able to edit a specific article to get their editToken - I've implemented a simple method to that does just what we need - returns a token to protect against CSRF.

A small tweak to `variationId` is introduced to allow a zero value for it and retrieving a `userId` on the backend side.

### DIFF
--- a/includes/wikia/api/UserFeedbackStorageApiController.class.php
+++ b/includes/wikia/api/UserFeedbackStorageApiController.class.php
@@ -39,7 +39,7 @@ class UserFeedbackStorageApiController extends WikiaApiController {
 	 * @responseParam string editToken
 	 */
 	public function getEditToken() {
-		$this->response->setFormat('json');
+		$this->response->setFormat( WikiaResponse::FORMAT_JSON );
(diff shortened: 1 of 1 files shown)


Grunny:
We can and should still check the edit token even if the anon is unable to edit, as it's not actually specific to edits, it is simply the main anti-CSRF token for a user.

Also, checking `HTTP_HOST` doesn't protect against anything, as `HTTP_HOST` is the domain that serves the API, not the domain the request came from.

includes/wikia/api/UserFeedbackStorageApiController.class.php
```diff
@@ -35,6 +36,14 @@ public function saveUserFeedback() {
 	}
 
 	/**
+	 * @responseParam string editToken
+	 */
+	public function getEditToken() {
+		$this->response->setFormat('json');
```
Grunny: `$this->response->setFormat( WikiaResponse::FORMAT_JSON );`

lukaszkonieczny:
LGTM

PR #9977, opened 2016-03-14
Fix Sass support in ResourceLoader

Grunny:
The sass_params URL var is never set, but the Sass params are available in
the ResourceLoaderContext so use it to get the params instead.

/cc @macbre


macbre:
LGTM

PR #9978, opened 2016-03-14
MAIN-6651: add ConfigException used by VE/ConfigFactory

owend:
When VE was originally back-ported from upstream mediawiki, this exception class was missing.


Grunny:
:+1:

PR #9979, opened 2016-03-14
ADEN-3016 fix for BackgroundChanger

jacekjursza:
This is a fix for BackgroundChanger.js - which was not prepared for switching backgrounds multiple times. The problem was that every time you switched background it needed to reload whole CSS stylesheet. For first switch it is OK, but if you switch to background that was already loaded - it was a waste of resources.

The solution is to provide local cache loaded backgrounds, disable not used stylesheets and enable the needed one.

### DIFF
--- a/skins/oasis/js/BackgroundChanger.js
+++ b/skins/oasis/js/BackgroundChanger.js
@@ -79,12 +79,11 @@ define('wikia.backgroundchanger', function()  {
 		if (nodeIndex > -1) {
 			$('link[data-background-changer]').attr('disabled', true);
 			$('link[data-background-changer="'+nodeIndex+'"]').attr('disabled', false);
-
 			onCssLoadCallBack();
 		} else {
-			loadedCss.push(sassUrl);
 			// load CSS and apply class changes to body element after loading
 			$.getCSS(sassUrl, function(link) {
+				loadedCss.push(sassUrl);
 				$(link).attr('data-background-changer', loadedCss.length-1);
 				onCssLoadCallBack();
 			});
--- a/skins/oasis/js/BackgroundChanger.js
+++ b/skins/oasis/js/BackgroundChanger.js
@@ -45,7 +45,7 @@ define('wikia.backgroundchanger', ['wikia.document'], function(doc)  {
 		// preload adskin image
 		imagePreload.src = options.skinImage;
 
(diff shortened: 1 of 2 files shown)


skins/oasis/js/BackgroundChanger.js
```diff
@@ -45,8 +45,7 @@ define('wikia.backgroundchanger', function()  {
 		// preload adskin image
 		imagePreload.src = options.skinImage;
 
-		// load CSS and apply class changes to body element after loading
-		$.getCSS(sassUrl, function() {
+		var onCssLoadCallBack = function(options) {
```
fraszczakszymon: Is it needed to pass here `options` again? It's accessible within `load` method

skins/oasis/js/BackgroundChanger.js
```diff
@@ -72,7 +71,26 @@ define('wikia.backgroundchanger', function()  {
 			} else {
 				$('body').removeClass('background-dynamic background-not-tiled background-fixed');
 			}
-		});
+		};
+
+		if (!$.isArray(doc.loadedCss)) {
+			doc.loadedCss = [];
```
fraszczakszymon: Is this if needed? Do we need to put this variable in `document` object? Can't it be declared in `wikia.backgroundchanger` module as `var loadedCss = []`?

skins/oasis/js/BackgroundChanger.js
```diff
@@ -72,7 +71,26 @@ define('wikia.backgroundchanger', function()  {
...
+
+		if (nodeIndex > -1) {
+			$('link.backgroundChanger').attr('disabled', true);
+			$('link.backgroundChanger.cssNode' + nodeIndex).attr('disabled', false);
+			onCssLoadCallBack(options);
+		} else {
+			doc.loadedCss.push(sassUrl);
+			// load CSS and apply class changes to body element after loading
+			$.getCSS(sassUrl, function(link) {
+				link.className = 'backgroundChanger cssNode' + nodeIndex;
```
fraszczakszymon: What do you think about `classList.add('...')`?

fraszczakszymon:
LGTM :+1:

fraszczakszymon:
LGTM :+1:

PR #9981, opened 2016-03-14
RECIRC-70: Update click events with placement information

pauloslund:


pauloslund:
Got the LGTM from @gabrys on Slack

PR #9984, opened 2016-03-15
(CE-3417) Make ImportJS respect ?usesitejs=0

Grunny:
Using the ?usesitejs=0 URL parameter results in wgUseSiteJs being set to
false for a request. ImportJS should respect this parameter so we need to
check wgUseSiteJs before adding ImportJS scripts.


Grunny:
Unit test failures are unrelated.

lukaszkonieczny:
LGTM

adamkarminski:
Tested, LGTM

PR #9987, opened 2016-03-15
PLATFORM-1909: mark failed dumps in the database

macbre:
[PLATFORM-1909](https://wikia-inc.atlassian.net/browse/PLATFORM-1909)

Emit non-zero exit code when we fail to generate or upload XML content dump. Then we can mark a failed dump attempt in the database.

Thanks to this change whoever requested the dump will see the correct timestamp on [Special:Statistics](http://monsterhunter.wikia.com/wiki/Special:Statistics) and can always require dump generation once more.

@jcellary

### DIFF
--- a/extensions/wikia/WikiFactory/Dumps/maintenance/DumpsOnDemandCron.php
+++ b/extensions/wikia/WikiFactory/Dumps/maintenance/DumpsOnDemandCron.php
@@ -4,11 +4,7 @@ include __DIR__ . '/../../../../../maintenance/Maintenance.php';
 class DumpsOnDemandCron extends Maintenance {
     
     const PIDFILE = '/var/run/MediaWikiDumpsOnDemandCron.pid';
-    
-    public function __construct() {
-        parent::__construct();
-    }
-    
+
     public function execute() {
 
         global $wgExternalSharedDB;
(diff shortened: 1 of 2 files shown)


extensions/wikia/WikiFactory/Dumps/maintenance/DumpsOnDemandCron.php
```diff
@@ -49,13 +45,19 @@ public function execute() {
 
         $this->output( "INFO: Creating dumps for Wikia #{$sWikiaId}.\n" );
 
-        $sCommand = sprintf( 'SERVER_ID=177 php %s/extensions/wikia/WikiFactory/Dumps/runBackups.php --conf %s --id=%d --both --tmp --s3', $IP, $wgWikiaLocalSettingsPath, $sWikiaId );
-
-        wfShellExec( $sCommand, $iStatus );
+        $sCommand = sprintf( 'SERVER_ID=177 php %s/extensions/wikia/WikiFactory/Dumps/runBackups.php --conf %s --id=%d --both --tmp --s3 2>&1', $IP, $wgWikiaLocalSettingsPath, $sWikiaId );
```
jcellary: For an outsider I have no idea what is id 177, a comment would be very useful here :)

jcellary:
One small comment left, rest LGTM

extensions/wikia/WikiFactory/Dumps/maintenance/DumpsOnDemandCron.php
macbre: It's a [community wiki[(http://community.wikia.com/wiki/Special:WikiFactory/177) - I'll use a const here instead (I believe we have one).

PR #9989, opened 2016-03-15
CE-3454 Check if the templatesWithoutType array is empty before INSERT

adamkarminski:
Fix for a DB error occuring when we tried to insert an empty set of data.

@Wikia/community-engineering


Grunny:
LGTM.

PR #9991, opened 2016-03-15
PLATFORM-2025: dedicated mysql users for maintenance scripts and tasks

macbre:
[PLATFORM-2025](https://wikia-inc.atlassian.net/browse/PLATFORM-2025)

Spikes of tasks and maintenance script runs can affect our end-user requests by causing `Too many connections` errors. Let's use dedicated DB users when running background (`Maintenance` sub-classes) / offline tasks (`TaskRunnerMaintenance` class) by using `getDatabaseCredentials` method.

Additionally extend the context logged along with messages when DB connection, SQL query or DB error occurs (add `db_user` field).

@wladekb / @drozdo / @owend / @mixth-sense


macbre:
```
There was 1 failure:

1) HubsRssIntegrationTest::testHubsRssLinks with data set #2 ('Entertainment', '<?xml version="1.0" encoding="UTF-8"?>
<rss xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:media="http://search.yahoo.com/mrss/" version="2.0">
<channel><title>Wikia Entertainment Feed</title><description>From Wikia community - Entertainment</description><link xmlns="http://www.w3.org/2005/Atom" href="http://www.wikia.com/rss/Entertainment" rel="self"/><link>http://www.wikia.com/rss/Entertainment</link><language>en</language><generator>MediaWiki 1.19.7</generator><lastBuildDate>Thu, 01 Jan 1970 00:00:00 +0000</lastBuildDate></channel></rss>
')
Failed asserting that 0 is greater than 0.
```

Pinging @lukaszkonieczny

wladekb:
lgtm

PR #9993, opened 2016-03-15
XW-1206 | fallback nicely when title is empty

bognix:
@Wikia/x-wing


bognix:
unit tests failure not related with the change

extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -224,41 +224,51 @@ public function getWikiVariables() {
...
+		$titleExists = true;
+		$dimensions = [];
 
-		if ( $article instanceof Article && $title->isRedirect() ) {
-			$title = $this->handleRedirect( $title, $article, [] )[0];
+		// Exception is thrown when empty title is send
+		// In that case we don't want to set dimensions which depend on title
+		// Title parameter is empty for URLs like /main/edit, /d etc. (all pages outside /wiki/ space)
+		try {
+			$title = $this->getTitleFromRequest();
```
vforge, line 235: So we're using `getTitleFromRequest` to override value from `wgTitle` (few lines above)

extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -224,41 +224,51 @@ public function getWikiVariables() {
...
+		} catch (Exception $ex) {
+			$titleExists = false;
+		}
+
+		if ( !empty ( $titleExists ) ) {
```
vforge: Can we have "smart defaults" to those `dimension`s when we don't have Title?

rogatty: I don't think there is a need for `$titleExists` variable. Just move what's in this block into `try { ... }` block.

extensions/wikia/MercuryApi/MercuryApiController.class.php
bognix, line 235: resolved

vforge:
LGTM

rogatty:
LGTM

rogatty:
LGTM

vforge:
LGTM

PR #9997, opened 2016-03-16
[PLATFORM-1999] Added global variable wgWikiaGlobalUserGroups

jcellary:
https://wikia-inc.atlassian.net/browse/PLATFORM-1999

In last refactoring I removed wgWikiaGlobalUserGroups global variable, as I refactored all its usages in code. Unfortunately it turns out it is being queried by backend scripts through our api and used to feed data that is then used by listusers special page. So this change reintroduces this variable.


macbre:
LGTM

macbre:
Backend code context - https://github.com/Wikia/backend/blob/f314ed12a997003f5c150a8086daa21154ffea1e/lib/Wikia/User.pm#L531-L555

PR #7843, opened 2015-07-14
CreateNewWiki: refactor starter import

macbre:
https://wikia-inc.atlassian.net/browse/PLATFORM-1305

**TL;DR** - make databases and QA team happy with how fast CreateNewWiki works :)
### The current state

The way CreateNewWiki works currently is not the best possible solution. We basically generate on-the-fly dump of the selected starter wiki and import it to the newly created wiki. Once that's done we run  cleanup SQL queries that remove old article revisions. It all causes heavy load on database (queries to master, replication of many rows to all slaves). And it all takes time - the current average is ~30 sec (with spikes reaching over a minute).
### The proposed solution

On a daily basis generate two kind of dumps for each starter database (handled by `dumpStarters.php` script):
- content XML dump with only the recent revisions
- SQL dump with tables storing the relations between articles, templates, categories and images

These dumps are bzip2 compressed and stored on DFS:

```
$ s3cmd -c /etc/s3cmd/poz_dev.cfg ls s3://starter/dumps/
2015-07-14 10:05       998   s3://starter/dumps/aastarter.sql.bz2
2015-07-14 10:05      2892   s3://starter/dumps/aastarter.xml.bz2
2015-07-14 10:05      4657   s3://starter/dumps/destarter.sql.bz2
2015-07-14 10:05     24300   s3://starter/dumps/destarter.xml.bz2
2015-07-14 10:05      3638   s3://starter/dumps/esstarter.sql.bz2
2015-07-14 10:05     22100   s3://starter/dumps/esstarter.xml.bz2
...
```

Change CreateNewWiki controller to call `importStarter.php` maintenance script that imports these dumps (content first, followed by SQL dump) and updates the newly created wiki statistics (namely article count).

Tested these changes on a sandbox and the AJAX call now takes ~16 seconds.

@drozdo / @michalroszka / @wladekb / @Wikia/community-engineering

PR #10002, opened 2016-03-17
CreateWiki: we no longer need to check for mysql binaries

macbre:
Since #7843 we use XML and SQL dumps that MW imports using PHP's mysql client

@artursitarski / @mixth-sense / @wladekb / @Wikia/sustaining-team (per code ownership)


artursitarski:
:+1: thank you @macbre

wladekb:
lgtm

PR #10007, opened 2016-03-17
[PLATFORM-2016] Added sysop and bureaucrat forced group management rights

jcellary:
https://wikia-inc.atlassian.net/browse/PLATFORM-2016

Added forcing of some group management controls independently of any local wiki customizations. Previously the content of $wgAddGroupsLocal replaced completely $wgAddGroups which caused these groups to be lacking some important permissions.


macbre:
:+1:

PR #10008, opened 2016-03-17
[PLATFORM-2000 Fixed group promotion in chat

jcellary:
https://wikia-inc.atlassian.net/browse/PLATFORM-2000

This fixes a problem with promoting a user to chat moderator - the problem was that chat does not set the user in the request context and overwrites it locally, so the default mechanism in User class which detects who is performing the change did not work properly. After this change we pass the appropriate user object directly to the user permissions service.

Pinging @owend


owend:
Code LGTM.  I wrote up instructions on how to test this so I will wait to merge until it's confirmed fixed.

PR #10009, opened 2016-03-17
[PLATFORM-1631] Removing unused extensions

jcellary:
https://wikia-inc.atlassian.net/browse/PLATFORM-1631

Removing unused extensions. Detailed list can be found in https://docs.google.com/a/wikia-inc.com/spreadsheets/d/1K9tlcudw9LP8ofqMgSgNUiJsHfM1CfzJKT6ZGCJoQiU/edit?usp=sharing

Corresponding config PR: https://github.com/Wikia/config/pull/1685

Checklist:
- [x] Go through list of permissions that were part of these extensions and remove them from the central permission repository if not needed
- [x] Got through tables and see if some of them can be removed as well


sebastianmarzjan:
Are you sure authimage won't make images on one inacessible? I think I've seen un use this code there.
It may be a conscious trade-off given migration to Confluence

jcellary:
So for One, we decided to make a fork and already deleted the "internal" environment from the dev branch of config repo. So app/config repos in dev branch are already incapable of running One. I checked for usage of the wgEnableAuthImages variable and it seems to only be set for the "internal" environment, that is wikia one (at least before I deleted that configuration a few weeks back). Wikifactory table also showed no traces of it being used on any other wiki. So to conclude I believe it should be safe to delete, but let me know if you can pin point me to a place where it is being used.

jcellary:
Ok I don't want this to get too big, so I'd like to merge this PR as is with 3 extensions removed. We can then open another PR for other removals.

@macbre would appreciate if you could do a CR.

macbre:
SecurePasswords and AuthImage were used specifically on One Wikia. And LinkSuggest from MW core is never included. Hence, LGTM

PR #10010, opened 2016-03-17
MAIN-6654: quick fix for exceeding API limit during maps geocoding

owend:


owend:
Actually, don't merge this, thought of a better solution.

PR #10011, opened 2016-03-17
MAIN-6654: cache API calls to google geocode API

owend:
This API currently has a limit of 2500 calls per day, which we hit.  This fix does 2 things: 

1) it will cache the data locally which should in most cases allow us to get all the data we need. 
2) if we do hit the API limit, it returns an empty result which isn't displayed.  this prevents the map rendering from crashing.  

@Wikia/sustaining-team

### DIFF
--- a/extensions/Maps/includes/Maps_Geocoder.php
+++ b/extensions/Maps/includes/Maps_Geocoder.php
@@ -111,7 +111,21 @@ abstract class MapsGeocoder {
 	 * @return array or false
 	 */
 	public function geocode( $address ) {
-		$response = ExternalHttp::get( $this->getRequestUrl( $address ) ); # Wikia change
+
+		// MAIN-6654: cache requests to this API because it has daily limits
+		global $wgMemc;
+		$key = wfMemcKey( "geocoder", $address );
+		$response = $wgMemc->get( $key );
+		if ( empty($response) ) {
+			$response = ExternalHttp::get( $this->getRequestUrl( $address ) ); # Wikia change
+			// MAIN-6654: make sure we do not return or store a bad response because it causes errors later
+			if ( strpos( $response, "OVER_QUERY_LIMIT" ) !== false ) {
+				$response = false;
+			}
+			if ( $response !== false ) {
+				$wgMemc->set( $key, $response, 86400 * 30 );  // 30 days
+			}
+		}
 
 		if ( $response === false ) {
 			return false;


extensions/Maps/includes/Maps_Geocoder.php
```diff
@@ -111,7 +111,21 @@ public function hasAlias( $alias ) {
...
-		$response = ExternalHttp::get( $this->getRequestUrl( $address ) ); # Wikia change
+
+		// MAIN-6654: cache requests to this API because it has daily limits
+		global $wgMemc;
+		$key = wfMemcKey( "geocoder", $address );
+		$response = $wgMemc->get( $key );
+		if ( empty($response) ) {
+			$response = ExternalHttp::get( $this->getRequestUrl( $address ) ); # Wikia change
+			// MAIN-6654: make sure we do not return or store a bad response because it causes errors later
+			if ( substr( $response, "OVER_QUERY_LIMIT" ) !== false ) {
```
sqreek: Can we check exactly for status instead using substr?
probability for occurrence this string in correct response is low, but let's check it in proper way.
Also is there a different status for OVER_QUERY_LIMIT? `429 Too Many Requests` maybe

sqreek: After double  checking, I don't think this solution help us.
Caching is already here: https://github.com/Wikia/app/blob/24375329bce0086d595ea9a3d6851993dd1abcea/extensions/Maps/includes/Maps_Geocoders.php#L197-L197.

owend: That caching is "in memory"  in array so it's only for 1 request. 
This "get" call only returns the response string, it buries the response code.  It's used in a lot of other places so I don't think it would be easy to change the API.

extensions/Maps/includes/Maps_Geocoder.php
```diff
@@ -111,7 +111,21 @@ public function hasAlias( $alias ) {
...
+				$response = false;
+			}
+			if ( $response !== false ) {
+				$wgMemc->set( $key, $response, 86400 * 30 );
```
inez: Add comment how long it is (86400 \* 30)

extensions/Maps/includes/Maps_Geocoder.php
inez: This is not how `substr` is supposed to be used - I think you are looking to use `strpos`.
It could not work with `substr` being used - so how was it tested? (It should be tested by mocking the response from  the API.)
@sqreek: @owend is correct - the caching layer that was already provided wasn't shared between requests.

inez: If you are "OVER_QUERY_LIMIT" it might be not a bad idea to cache that response anyways - to avoid bunch of requests - which are known to fail.

owend: Nice catch.  I tested it on sandbox and then used the wrong function when copying over.   Re-testing and updating the PR now.  

As far as the caching strategy, I realized that caching for 1 day is pointless since there's a daily API limit so I increased the duration to 30 days, and we probably don't want to cache a bad response that long.

extensions/Maps/includes/Maps_Geocoder.php
```diff
@@ -111,7 +111,21 @@ public function hasAlias( $alias ) {
...
+		$key = wfMemcKey( "geocoder", $address );
+		$response = $wgMemc->get( $key );
+		if ( empty($response) ) {
+			$response = ExternalHttp::get( $this->getRequestUrl( $address ) ); # Wikia change
+			// MAIN-6654: make sure we do not return or store a bad response because it causes errors later
+			if ( strpos( $response, "OVER_QUERY_LIMIT" ) !== false ) {
+				$response = false;
+			}
+			if ( $response !== false ) {
+				$wgMemc->set( $key, $response, 86400 * 30 );  // 30 days
```
ghost, line 126: what about using constants? https://github.com/Wikia/app/blob/f28a3b789cf45b808fee3c075ecc6195d1db9bc0/includes/wikia/nirvana/WikiaResponse.class.php#L47

owend, line 126: This is a third party extension, I'd prefer to avoid adding dependencies on wikia classes/constants.

ghost, line 126: Good point. I must have missed the fact that it was a 3rd party code. Please, continue.

PR #10013, opened 2016-03-18
SUS-266 Code to prevent (false) CSRF detection on Special:Forum

gabrys:
Changes:
- Marked `Forum::createDefaultForum` method as deprecated. We want to remove it after SUS-302 and SUS-303 are completed
- Removed some `wfProfileIn` and `wfProfileOut` calls from `createOrEditBoard`
- Renamed `hasAtLeast` method to `hasMoreThan` as it checks if there's MORE than a given number of pages

Next steps:
- Once SUS-302 and SUS-303 are done we can remove the `Forum::createDefaultForum`. Created SUS-315 for that.

### DIFF
--- a/extensions/wikia/Forum/Forum.class.php
+++ b/extensions/wikia/Forum/Forum.class.php
@@ -1,6 +1,5 @@
 <?php
 use Wikia\Logger\WikiaLogger;
-use Wikia\Util\GlobalStateWrapper;
 
 /**
  * Forum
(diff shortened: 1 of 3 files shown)


gabrys:
@macbre I'd like your opinion on this one

mszabo-wikia:
Sorry to barge in on this one, but per what I found in VOLDEV-142, it should be entirely possible to get rid of the `$wgUser` swap by just using `WikiPage::doEdit`: https://github.com/Wikia/app/compare/dev...WikiaVolDev:VOLDEV-142

gabrys:
Thanks for the hint. Unless someone else wants to do it before me, I'll
investigate this tomorrow morning

extensions/wikia/Security/classes/CSRFDetector.class.php
```diff
@@ -36,6 +37,23 @@ public static function onRequestWasPosted() {
...
+	 * @param \Closure $closure
+	 * @param string $reason
+	 */
+	public static function disableCheck( \Closure $closure, $reason ) {
+		self::$disabledReason = $reason;
+		$closure();
+		self::$disabledReason = false;
+	}
+
+	/**
```
macbre: My heart is bleeding, but let's try this approach :)

extensions/wikia/Forum/Forum.class.php
```diff
@@ -334,29 +332,41 @@ public function deleteBoard( $board ) {
...
+	 * Note this function returns true even if the boards were not successfully created.
+	 *
+	 * @return bool true if tried creating the boards, false if there were some boards already
+	 */
+	public function createDefaultBoards() {
+		if ( $this->hasMoreThan( NS_WIKIA_FORUM_BOARD, 0 ) ) {
+			return false;
+		}
+
+		WikiaDataAccess::cachePurge( wfMemcKey( 'Forum_hasMoreThan', NS_WIKIA_FORUM_BOARD, 0 ) );
```
macbre: Shouldn't we purge the cache **after** the first board is created? This will prevent unlikely (but still possible) race conditions.

owend:
LGTM

PR #10017, opened 2016-03-18
Move the default avatars message and use a more modern method to get …

adamkarminski:
…the array

The issue caused a `relative-path parameter missing` exception on some devboxes. It makes more sense to keep default values in a file that is included by default.

Also, let's use a more modern method to parse a message as an array.

@Grunny @rogatty

### DIFF
--- a/extensions/wikia/UserProfilePageV3/Masthead.class.php
+++ b/extensions/wikia/UserProfilePageV3/Masthead.class.php
@@ -122,7 +122,7 @@ class Masthead {
 		wfProfileIn( __METHOD__ );
 
 		$this->mDefaultAvatars = array();
-		$images = getMessageForContentAsArray( 'blog-avatar-defaults' );
+		$images = getMessageAsArray( 'blog-avatar-defaults' );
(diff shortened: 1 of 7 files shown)


extensions/wikia/UserProfilePageV3/Masthead.class.php
```diff
@@ -122,7 +122,7 @@ public function getDefaultAvatars( $thumb = "" ) {
 		wfProfileIn( __METHOD__ );
 
 		$this->mDefaultAvatars = array();
-		$images = getMessageForContentAsArray( 'blog-avatar-defaults' );
+		$images = getMessageAsArray( 'blog-avatar-defaults' );
```
Grunny: `getMessageAsArray` is not "For Content" like the old method was. `getMessageForContentAsArray` should be updated to use the new methods like `getMessageAsArray` uses and make use of  `->inContentLanguage()` on the Message object so it has the same behaviour.

Grunny:
LGTM.

PR #10020, opened 2016-03-18
wfShellExec: do not run scripts via ulimit4.sh

macbre:
[OPS-8226](https://wikia-inc.atlassian.net/browse/OPS-8226)

It causes 

```
setaffinity: Invalid argument
Segmentation fault
```

when run in containers.

@adamw-wikia / @wladekb


adamw-linadm:
LGTM

PR #10022, opened 2016-03-18
(SEC-54) Disable NewWindowLinks functionality

Grunny:
Disable NewWindowLinks functionality due to the risks of allowing this
with user generated content.

The tag will still work so that existing uses of the tag will work, just
without the links opening in a new tab.


adamkarminski:
LGTM

PR #10024, opened 2016-03-18
SUS-291: VideoHandler::getVideoList should always return json

owend:
Spent some time on it but could not reproduce the error manually.  We will just have to look at the error logs to verify that this is fixed. 

@Wikia/sustaining-team


sqreek:
LGTM

PR #10025, opened 2016-03-21
PLATFORM-2033: forward client data to wfShellExec

macbre:
[PLATFORM-2033](https://wikia-inc.atlassian.net/browse/PLATFORM-2033)

Allow client data to be passed to maintenance scripts run via `wfShellExec` using environment variables.

``` php
> echo wfShellExec("run_maintenance --id=177 --script 'updateSpecialPages.php --list' --ip $IP/maintenance")
wfShellExec: WIKIA_TRACER_X_TRACE_ID='mw56efc375ce98e0.87969882' WIKIA_TRACER_X_REQUEST_ID='mw56efc375ce98e0.87969882' WIKIA_TRACER_X_CLIENT_IP='127.0.0.1' WIKIA_TRACER_X_REQUEST_ORIGIN_HOST='dev-macbre' WIKIA_TRACER_X_WIKIA_INTERNAL_REQUEST='mediawiki' run_maintenance --id=177 --script 'updateSpecialPages.php --list' --ip /usr/wikia/source/app/maintenance
```

Introduce `BeforeWfShellExec` hook.

@wladekb

### DIFF
--- a/includes/GlobalFunctions.php
+++ b/includes/GlobalFunctions.php
@@ -2913,6 +2913,8 @@ function wfShellExec( $cmd, &$retval = null, $environ = array() ) {
 
 	wfInitShellLocale();
 
+	wfRunHooks( 'wfShellExecBeforeEnviron', [ &$cmd, &$environ ] ); // Wikia change
+
 	$envcmd = '';
 	foreach( $environ as $k => $v ) {
 		if ( wfIsWindows() ) {
--- a/includes/GlobalFunctions.php
+++ b/includes/GlobalFunctions.php
@@ -2913,7 +2913,7 @@ function wfShellExec( $cmd, &$retval = null, $environ = array() ) {
 
 	wfInitShellLocale();
 
-	wfRunHooks( 'wfShellExecBeforeEnviron', [ &$cmd, &$environ ] ); // Wikia change
+	wfRunHooks( 'BeforeWfShellExec', [ &$cmd, &$environ ] ); // Wikia change
 
 	$envcmd = '';
 	foreach( $environ as $k => $v ) {
--- a/lib/Wikia/src/Tracer/RequestId.php
+++ b/lib/Wikia/src/Tracer/RequestId.php
@@ -1,7 +1,7 @@
 <?php
 
-/*
- * Generate unique request ID or use the value when passed as HTTP request header
+/**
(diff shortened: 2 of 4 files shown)


includes/GlobalFunctions.php
```diff
@@ -2913,6 +2913,8 @@ function wfShellExec( $cmd, &$retval = null, $environ = array() ) {
 
 	wfInitShellLocale();
 
+	wfRunHooks( 'wfShellExecBeforeEnviron', [ &$cmd, &$environ ] ); // Wikia change
```
wladekb: 1. AFAIR it's usually `BeforeSomething` which I'd like to see to be followed here.
2. I guess the `Environ` suffix is supposed to note that you'll get environment in the arguments - I'd skip it - looking at the line with `wfRunHooks` makes it obvious it's getting passed to handlers

lib/Wikia/src/Tracer/RequestId.php
```diff
@@ -43,16 +43,21 @@ function getRequestId() {
 			return $this->requestId;
 		}
 
-		// try to get the value from request X-Request-Id header
+		// try to get the value from request's X-Request-Id header or env variable
 		// we can't simply use $wgRequest as it's not yet set at this point
 		// this method is called on WikiFactoryExecuteComplete hook
 		// TODO: handle X-Trace-Id header as well
 		$headerValue = !empty( $_SERVER['HTTP_X_REQUEST_ID'] ) ? $_SERVER['HTTP_X_REQUEST_ID'] : false;
+		$envValue = !empty( $_ENV['X_TRACE_ID'] ) ? $_ENV['X_TRACE_ID'] : false;
```
wladekb: Here and in the WikiaTracer. I'm wondering about means to prevent accidental name clash in various situations. How about adding a distinct prefix which is very unlikely to occur and would give us more certainty that won't be randomly defined by anything else?

wladekb:
I like the idea of env vars here. Left 2 comments though.

includes/GlobalFunctions.php
macbre: Addressed in a9b5330

lib/Wikia/src/Tracer/RequestId.php
macbre: Added `WIKIA_TRACER_` prefix for all env variables in f74fd64

PR #10027, opened 2016-03-21
ADEN-3114 Fix modil define method

fraszczakszymon:
Allow defining modules without id as RequireJS allows it: http://requirejs.org/docs/api.html#define

//cc @nandy-andy @hakubo

### DIFF
--- a/resources/wikia/libraries/modil/modil.js
+++ b/resources/wikia/libraries/modil/modil.js
@@ -125,6 +125,7 @@
 	 *
 	 * @public
 	 *
+	 * @example define(function () { return {hello: 'World'}; });
 	 * @example define('mymod', function () { return {hello: 'World'}; });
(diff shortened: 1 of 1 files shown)


resources/wikia/libraries/modil/modil.js
```diff
@@ -134,6 +138,17 @@
...
 	define = function (id, dependencies, definition, defMock) {
+		//no id, it's the definition or dependencies
+		if (id instanceof funcType && !dependencies && !definition) {
+			definition = id;
+			dependencies = nil;
+			id = getRandomId();
+		} else if (id instanceof arrType && dependencies && !definition) {
+			definition = dependencies;
+			dependencies = id;
+			id = getRandomId();
```
nandy-andy, line 151: It seems we'll have `id` defined almost all the time now, so changing JSDocs and the error message below may be helpful. What do you think?

PR #10030, opened 2016-03-21
Include the namespace when detecting infoboxes for Builder

sebastianmarzjan:
Introduce an explicit check against namespace when determining Infobox Builder support to avoid false positives

@idradm


extensions/wikia/PortableInfoboxBuilder/services/PortableInfoboxBuilderHelper.class.php
```diff
@@ -9,20 +9,15 @@ class PortableInfoboxBuilderHelper {
...
 	public static function isInfoboxTemplate( $title ) {
-		$tc = new TemplateClassificationService();
-		$isInfobox = false;
-
-		try {
-			$type = $tc->getType( F::app()->wg->CityId, $title->getArticleID() );
-			$isInfobox = ( $type === TemplateClassificationService::TEMPLATE_INFOBOX );
-		} catch ( Swagger\Client\ApiException $e ) {
-			// If we cannot reach the service assume the default (false) to avoid overwriting data
+		if ( $title->getNamespace() === NS_TEMPLATE ) {
```
idradm: You could use `inNamespace()` method available in title object, and make it a single return statement.

sebastianmarzjan: Nice suggestion, done in 49c5a9d

extensions/wikia/PortableInfoboxBuilder/services/PortableInfoboxBuilderHelper.class.php
```diff
@@ -50,16 +41,17 @@ public static function getUrlPath( $titleText ) {
...
 	 */
 	public static function canUseInfoboxBuilder( $title, $user ) {
+		$infoboxes = \PortableInfoboxDataService::newFromTitle( $title )->getInfoboxes();
+
 		return PortableInfoboxBuilderHelper::isInfoboxTemplate( $title )
-			&& ( new \PortableInfoboxBuilderService() )->isValidInfoboxArray(
-				\PortableInfoboxDataService::newFromTitle( $title )->getInfoboxes()
-			)
-			&& ( new \Wikia\TemplateClassification\Permissions() )->userCanChangeType( $user, $title );
+			   && ( $title->isKnown() && !empty( $infoboxes ) )
```
sebastianmarzjan, line 52: what is the case that isKnown() covers for us that exists() doesn't?

sebastianmarzjan:
One question, the changes starting at 63228e3 look good to me

dianafa:
LG

extensions/wikia/PortableInfoboxBuilder/services/PortableInfoboxBuilderHelper.class.php
sebastianmarzjan, line 52: OK, no difference in our case

PR #10035, opened 2016-03-21
Bring back the list of global groups in siteinfo api

wladekb:
ApiQuery SiteInfo was missing the list of global groups after the recent changes around permissions. This is going to bring it back.

/cc @owend @jcellary

### DIFF
--- a/extensions/wikia/WikiaApi/WikiaApiQuerySiteinfo.php
+++ b/extensions/wikia/WikiaApi/WikiaApiQuerySiteinfo.php
@@ -66,7 +66,15 @@ class WikiaApiQuerySiteinfo extends ApiQuerySiteinfo {
 		$data = array ();
 		foreach ($this->variablesList as $id => $variableName) {
 			$data[$id] = array( 'id' => $variableName );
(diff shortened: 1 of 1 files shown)


extensions/wikia/WikiaApi/WikiaApiQuerySiteinfo.php
```diff
@@ -66,7 +66,15 @@ protected function appendVariables($property) {
 		$data = array ();
 		foreach ($this->variablesList as $id => $variableName) {
 			$data[$id] = array( 'id' => $variableName );
-			$value = (array_key_exists($variableName, $GLOBALS) && !is_null($GLOBALS[$variableName])) ? $GLOBALS[$variableName] : "";
+			switch ($variableName) {
+				case 'wgWikiaGlobalUserGroups':
+					/** @var \Wikia\Service\User\Permissions\PermissionsServiceImpl $permissionsService */
```
nmonterroso, line 71: I think we should try and use the interface rather than the implementation:

``` php
/** @var \Wikia\Service\User\Permissions\PermissionsService $permissionsService */
```

wladekb, line 71: Good point. Fixed in bc8f1b6

PR #10036, opened 2016-03-21
RECIRC-77: Update Taboola style in the right rail

pauloslund:
Just a minor update to bring Taboola styles in line with our other right rail experiments


saipetch:
LGTM

PR #10039, opened 2016-03-22
[SERVICES-1268] Removing legacy helios code

ArturKlajnerok:
Removing code related to helios login shadow mode. 

/cc @Wikia/services-team @mixth-sense @jcellary

### DIFF
--- a/extensions/wikia/Helios/User.class.php
+++ b/extensions/wikia/Helios/User.class.php
@@ -303,8 +303,6 @@ class User {
 	 * @return bool true - hook handler
 	 */
 	public static function onUserCheckPassword( $id, $username, $hash, $password, &$result, &$errorMessageKey ) {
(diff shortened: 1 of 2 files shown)


ArturKlajnerok:
@jcellary @mixth-sense @macbre I understand that we can safely remove `wgEnableHeliosExt` from the codebase or is there anything else you know about that could require old MW authentication path ?
We plan to remove old MW authentication in further steps and I wanted to make sure that thing like `one.wikia.com` or something else is not blocking us.

extensions/wikia/Helios/User.class.php
```diff
@@ -314,30 +312,10 @@ public static function onUserCheckPassword( $id, $username, $hash, $password, &$
 			$heliosException = $e;
```
macbre, line 312: Do we log these exceptions?

PR #10041, opened 2016-03-22
SUS-69 A param and a wg var for Ooyala HTML5 player

gabrys:
Introducing a wg var (and an URL param) to control the Flash/HTML5
preference for the Ooyala video player.

This way we can enable/disable this per-wiki and better test the
differences between the players.


owend:
LGTM.  And since this is a prerequisite for any QA/testing that we want to do let's just merge it.

PR #10042, opened 2016-03-22
Prevent smart banner on Windows Phone devices.

Silic0nS0ldier:
I've found that on Windows Phone/Mobile devices, due to "Andoid" being in the user agent the banner would appear. (reason for this craziness was developers blocking a finally properly working IE on Windows Phone 8)

To get around the issue, I've added a check for the presence of "Windows Phone" in the user agent string.


mszabo-wikia:
WikiaMobile is no longer used

PR #10045, opened 2016-03-22
XW-1283 | Return contentNamespaces in MercuryApi's getWikiVariables

Warkot:
A step to support colon-ized articles in Mercury.

@Wikia/x-wing


rogatty:
LGTM

PR #10046, opened 2016-03-22
SUS-320 add editor reset for overridden NewMessageForm

sqreek:
[Ticket](https://wikia-inc.atlassian.net/browse/SUS-320)

NewMessageForm overridden default behaviour and do not reset wikia editor.
this change add explicit reset in overridden function

//cc @Wikia/sustaining-team @wladekb


wladekb:
lgtm

PR #10047, opened 2016-03-22
SUS-320 add editor reset for overridden NewMessageForm

sqreek:
[P2 Ticket](https://wikia-inc.atlassian.net/browse/SUS-320)

NewMessageForm overridden default behaviour and do not reset wikia editor.
this change add explicit reset in overridden function


pawelchcki:
LGTM

PR #10048, opened 2016-03-22
RECIRC 73: Add new design variations for Fandom in the right rail

pauloslund:


extensions/wikia/Recirculation/js/utils.js
```diff
@@ -44,8 +44,10 @@ define('ext.wikia.recirculation.utils', [
 	}
 
 	function buildLabel(element, label) {
-		var slot = $(element).parent().index() + 1;
-		
+		// This is a work around to make sure we only get the index based on actual list items
+		var $listItem = $(element).parent();
+		var slot = $listItem.closest('ul').children('li').index($listItem) + 1;
```
gabrys: I think it would be more straight-forward to assign the numbers in the place we build the list and use data params for it.

extensions/wikia/Recirculation/js/views/incontent.js
```diff
@@ -59,9 +59,12 @@ define('ext.wikia.recirculation.views.incontent', [
 		}
 	}
 
-	return {
-		render: render,
-		setupTracking: setupTracking,
-		findSuitableSection: findSuitableSection
+	return function() {
```
gabrys, line 61: What's the reason for this? Are you having multiple "instances" of this module?

extensions/wikia/Recirculation/js/experiments/placement.js
```diff
@@ -123,10 +133,36 @@ require([
 			.fail(function() {});
 	}
 
-	function setupLegacyTracking() {
-		tracker.trackVerboseImpression(experimentName, 'rail');
-		$(railSelector).on('mousedown', 'a', function() {
-			tracker.trackVerboseClick(experimentName, utils.buildLabel(this, 'rail'));
+	function injectSubtitle($html) {
+		$html.find('.trending').after('<h2>'+ $.msg('recirculation-fandom-subtitle') +'</h2>');
```
gabrys: Probably better expressed by `$('<h2></h2>').text($.msg('recirculation-fandom-subtitle'))` (no HTML injection)

extensions/wikia/Recirculation/js/views/rail.js
```diff
@@ -28,8 +48,17 @@ define('ext.wikia.recirculation.views.rail', [
 		}
 	}
 
-	return {
-		render: render,
-		setupTracking: setupTracking
+	// Add a line break after the first word in the title
+	function formatTitle(title) {
```
gabrys, line 48: Out of curiosity: why?

extensions/wikia/Recirculation/templates/client/rail.mustache
```diff
@@ -1,13 +1,14 @@
-<section class="recirculation-rail module">
-	<h2>{{ title }}</h2>
+<section class="recirculation-rail module group-{{ group }}">
+	<h2>{{{ title }}}</h2>
```
gabrys: This looks fishy. Is there any reason to have the unescaped version other than to accept the `<br>` as added by `formatTitle`?

extensions/wikia/Recirculation/js/views/incontent.js
pauloslund, line 61: Primarily to keep consistency between how views are created. I use a constructor function so I can setup pass in different options as seen here: https://github.com/Wikia/app/blob/RECIRC-73/extensions/wikia/Recirculation/js/experiments/placement.js#L64-L67

extensions/wikia/Recirculation/js/views/rail.js
pauloslund, line 48: Design decision
screen shot 2016-03-22 at 3 33 48 pm

extensions/wikia/Recirculation/templates/client/rail.mustache
pauloslund: It's just for the `<br>`. Another option would be to modify the title post render, which I'm not really a fan of either. Suggestions are appreciated :)

extensions/wikia/Recirculation/js/views/incontent.js
gabrys, line 61: :+1:

extensions/wikia/Recirculation/templates/client/rail.mustache
gabrys: Well, out of all the possibilities not many of them fit nicely to Mustache templating. It's probably fine how it is, I would express more explicitly the title is actually HTML naming it `titleHtml`.

extensions/wikia/Recirculation/templates/client/rail.mustache
```diff
@@ -1,13 +1,14 @@
...
 	<ul class="thumbnails">
 		{{#items}}
-			<li class="title-thumbnail">
+			<li class="rail-item {{classes}}" data-flag="{{ flag }}" data-index="{{ index }}">
```
gabrys, line 5: Try `{{@index}}` (http://stackoverflow.com/questions/5021495/in-mustache-how-to-get-the-index-of-the-current-section). Would remove the need for `addIndex`

gabrys:
LGTM

PR #10049, opened 2016-03-22
SUS-134 Content Warning message is impossible to bypass on smaller browser windows

gabrys:
Two changes here:
- Removal of `position: fixed` for the ContentWarning message
- Added `height: 0` for the WikiaMainContent if ShowContentWarning

This makes the warning scrollable, but at the same time eliminates the
possibility to scroll away from it on long articles.

Also reindented the file to used tabs instead of spaces.

You might want to use the `w=1` param for GitHub: https://github.com/Wikia/app/pull/10049/files?w=1


owend:
LGTM

PR #10051, opened 2016-03-23
WikiaTracer: introduce span_id

macbre:
[PLATFORM-1948](https://wikia-inc.atlassian.net/browse/PLATFORM-1948)

Log both `trace_id` and `span_id`:
- `span_id` is sub-request specific ID (used to separate sub-requests made to compose a response for the end-user)
- `trace_id` comes from the parent call and is used to group messages from all sub-requests

@wladekb

### DIFF
--- a/lib/Wikia/src/Tracer/RequestId.php
+++ b/lib/Wikia/src/Tracer/RequestId.php
@@ -59,14 +59,22 @@ class RequestId {
 			wfDebug( __METHOD__ . ": from env variable\n" );
 		}
 		else {
-			// return 23 characters long unique ID + mw prefix
-			// e.g. mw5405bb3d129e76.46189257
-			$this->requestId = uniqid( 'mw', true );
+			$this->requestId = self::generateId();
 			wfDebug( __METHOD__ . ": generated a new one\n" );
 		}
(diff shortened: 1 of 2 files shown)


macbre:
```
HubsRssIntegrationTest::testHubsRssLinks with data set #2 ('Entertainment', '<?xml version="1.0" encoding="UTF-8"?>
<rss xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:media="http://search.yahoo.com/mrss/" version="2.0">
<channel><title>Wikia Entertainment Feed</title><description>From Wikia community - Entertainment</description><link xmlns="http://www.w3.org/2005/Atom" href="http://www.wikia.com/rss/Entertainment" rel="self"/><link>http://www.wikia.com/rss/Entertainment</link><language>en</language><generator>MediaWiki 1.19.7</generator><lastBuildDate>Thu, 01 Jan 1970 00:00:00 +0000</lastBuildDate></channel></rss>
')
Failed asserting that 0 is greater than 0.
```

Unrelated

wladekb:
lgtm

sqreek:
it seems taht

PR #10052, opened 2016-03-23
CE-3539: Dont count upvotes for anonymous users

lukaszkonieczny:
Dont count upvotes for anonymous users.

### DIFF
--- a/includes/wikia/services/RevisionUpvotesService.class.php
+++ b/includes/wikia/services/RevisionUpvotesService.class.php
@@ -49,7 +49,7 @@ class RevisionUpvotesService {
 
 		$rowRemoved = $db->affectedRows();
 
-		if ( $userId > 0 ) {
+		if ( $this->shouldStoreUserData( $userId ) ) {
 			( new \WikiaSQL() )
 				->UPDATE( self::UPVOTE_USERS_TABLE )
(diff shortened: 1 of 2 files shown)


includes/wikia/services/RevisionUpvotesService.class.php
```diff
@@ -49,12 +49,14 @@ public function removeUpvote( $id, $fromUser, $userId ) {
 
 		$rowRemoved = $db->affectedRows();
 
-		( new \WikiaSQL() )
-			->UPDATE( self::UPVOTE_USERS_TABLE )
-			->SET_RAW( 'total', 'IF(`total` > 0, `total` - 1, 0)' )
-			->SET_RAW( 'new', 'IF(`new` > 0, `new` - 1, 0)' )
-			->WHERE( 'user_id' )->EQUAL_TO( $userId )
-			->run( $db );
+		if ( $userId > 0 ) {
```
adamkarminski: I just wonder - would it be better to create an `shouldStoreUpvote()` that could hold more conditions if necessary and would have an explicit naming?

adamkarminski:
One question but LGTM in general.

adamkarminski:
LGTM

PR #10056, opened 2016-03-23
CE-3453: Check if upvotes exist for given revision

lukaszkonieczny:
Check if upvotes exist for given revision


kamilkoterba:
LGTM

PR #10057, opened 2016-03-23
ADEN-3097 Extend Evolve2 parameters

fraszczakszymon:
Changes:
- fix incrementing pos value on mercury consecutive pages
- add new slot params `wloc`, `wpos` and `wsrc`
- remove `sect` and `site` slot params

https://wikia-inc.atlassian.net/browse/ADEN-3097


extensions/wikia/AdEngine/js/provider/evolve2.js
```diff
@@ -102,6 +109,11 @@ define('ext.wikia.adEngine.provider.evolve2', [
 		log(['fillInSlot', slot.name, 'done'], 'debug', logGroup);
 	}
 
+	resetPosTargeting();
```
nandy-andy, line 112: I'm thinking if there is better place for this line and 3 below... More cleaner way to do it. It's a minor note - no action needs to be taken. Yet ;)

nandy-andy:
LGTM :+1:

PR #10058, opened 2016-03-23
XW | Handle invalid Vignette URLs gracefully

rogatty:
Without `try catch` block we have fatal errors on devboxes where broken URLs are common. Don't know if it happens on production too but better to be safe.

/cc @Warkot


Warkot:
LGTM :shipit:

PR #10059, opened 2016-03-23
RECIRC-88: Update scrolldepth test to use new view style

pauloslund:
This PR also updates a property naming regression that was introduced


gabrys:
LGTM

PR #10060, opened 2016-03-24
CE-3452: Return user id as int

lukaszkonieczny:
Return user id as int


kamilkoterba:
LGTM. test fail seems unrelated.

PR #10063, opened 2016-03-24
MAIN-6750 In UserPreference check isset before use array value

sqreek:
[MAIN-6750](https://wikia-inc.atlassian.net/browse/MAIN-6750)

ping @macbre

### DIFF
--- a/lib/Wikia/src/Domain/User/Preferences/UserPreferences.php
+++ b/lib/Wikia/src/Domain/User/Preferences/UserPreferences.php
@@ -91,7 +91,7 @@ class UserPreferences {
 	 * @return LocalPreference[]
 	 */
 	public function getLocalPreferencesForWiki( $wikiId ) {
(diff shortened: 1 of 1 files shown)


lib/Wikia/src/Domain/User/Preferences/UserPreferences.php
```diff
@@ -91,8 +91,9 @@ public function getLocalPreferences() {
 	 * @return LocalPreference[]
 	 */
 	public function getLocalPreferencesForWiki( $wikiId ) {
-		if ( is_array( $this->localPreferences[ $wikiId ] ) ) {
-			return array_values( $this->localPreferences[ $wikiId ] );
+		$localPreference = $this->localPreferences[ $wikiId ];
```
macbre: Are you sure that this one will not report a notice when `$wikiId` key does not exist?

And isn't this notice a sign of a bigger issue? For instance: we failed to load preferences for a given wiki

macbre:
LGTM

PR #10066, opened 2016-03-24
ADEN-3116 Flag slots with tierNONE

fraszczakszymon:


extensions/wikia/AdEngine/js/lookup/rubiconFastlane.js
```diff
@@ -191,13 +223,12 @@ define('ext.wikia.adEngine.lookup.rubiconFastlane', [
...
-		var slotName;
-
-		for (slotName in slots) {
-			if (slots.hasOwnProperty(slotName)) {
-				addSlotPrice(slotName, slots[slotName].getAdServerTargeting());
-			}
+		if (!win.rubicontag || !win.rubicontag.getAllSlots) {
+			return;
 		}
+		win.rubicontag.getAllSlots().forEach(function (slot) {
```
nandy-andy: Does `getAllSlots()` always return an array?

fraszczakszymon: :+1:
It seems it does, but I'm not sure so I added additional check.

drets:
LGTM :+1:

PR #10067, opened 2016-03-24
SUS-234 Maintain the watch status for the forum board when editing it

gabrys:
While editing a page you can specify if you want to watch the page or
not. When issuing an edit programmatically you need to pass the watch
status, or otherwise the user will unfollow the page after any changes
we want to contribute to the user (which is very counter-intuitive).

This fix just reads the watch status for the board article and passes it
to the edit page class, so the status is maintained after you edit board
description or title.


owend:
Nice.  LGTM.  :gem: :golf:

PR #10068, opened 2016-03-25
DAT-3969 update special pages cache on create step

sebastianmarzjan:
## DESCRIPTION

This PR introduces updating of special page lists on wiki creation to ensure all pages imported from starter are properly indexed.
The time consequence is that this will extend the execution of the async task by ~1 second.

/CC @macbre for visibility


sebastianmarzjan:
LGTM, tested

macbre:
Better approach would be to include `querycache` table in SQL dump that is made for starters databases and imported for newly created wikis. Why duplicating the work on each wiki creation when we already have this data?

PR #10070, opened 2016-03-25
DAT-4009 Infobox Builder handle redirects when no template title set

dianafa:
do not merge without mercury part: Wikia/mercury#2240

https://wikia-inc.atlassian.net/browse/DAT-4009

@SebastianMarzjan


sebastianmarzjan:
LGTM

PR #10074, opened 2016-03-25
SEO (SEO-308): RobotsTxt - add functions to set custom rules

saipetch:
SEO (SEO-308, SEO-260):
- add functions to set custom rules 
- block discussion user pages
- block redirect pages


gabrys:
LGTM

PR #10075, opened 2016-03-25
SOC-2123 Add additional logging when MW tries to save attributes for non-existent users

jsutterfield:
There are currently values in the attribute service for non-existent users and we're not sure why. This ticket adds additional logging to help track down the issue.

We already have logging when MW receives errors from the attribute service, but this PR makes it more fine-grained by passing the entire exception which will provide us with a stack trace, as well as giving a more detailed message about what was happening when an error occurred (we're only curious about save errors right now).

This should help us narrow down the conditions when this bad data was being saved, as the attribute service throws a 404 when an attempt to save data for a non-existent user happens. Those 404 manifest as exceptions in MW which are being caught and logged here.

Ticket: https://wikia-inc.atlassian.net/browse/SOC-2123

### DIFF
--- a/lib/Wikia/src/Service/User/Attributes/AttributeKeyValueService.php
+++ b/lib/Wikia/src/Service/User/Attributes/AttributeKeyValueService.php
@@ -47,7 +47,7 @@ class AttributeKeyValueService implements AttributeService {
 
 			return $ret;
 		} catch ( \Exception $e ) {
-			$this->logError( $userId, $e, "USER_ATTRIBUTE error saving to service" );
+			$this->logError( $userId, $e, "USER_ATTRIBUTES error saving to service" );
 			return false;
 		}
 	}
@@ -66,7 +66,7 @@ class AttributeKeyValueService implements AttributeService {
 		try {
 			$attributeArray = $this->persistenceAdapter->getAttributes( $userId );
 		} catch ( \Exception $e ) {
-			$this->logError( $userId, $e, "USER_ATTRIBUTE error getting from service" );
+			$this->logError( $userId, $e, "USER_ATTRIBUTES error getting from service" );
 		}
 
 		return $attributeArray;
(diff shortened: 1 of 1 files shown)


lib/Wikia/src/Service/User/Attributes/AttributeKeyValueService.php
```diff
@@ -87,15 +87,15 @@ public function delete( $userId, Attribute $attribute ) {
...
 		}
 	}
 
-	private function logError( $userId, \Exception $e ) {
-		$this->error( $e->getMessage(), [
+	private function logError( $userId, \Exception $e, $msg ) {
+		$this->error( $msg , [
 			'user' => $userId,
-			'exceptionType' => get_class( $e ),
+			'exception' => $e
```
jsutterfield, line 98: By passing the entire exception, we get access to the stack trace. As in [these logs](https://kibana.wikia-inc.com/index.html#dashboard/temp/AVOwK8vewuxBx3joFaRw).

lib/Wikia/src/Service/User/Attributes/AttributeKeyValueService.php
```diff
@@ -47,7 +47,7 @@ public function set( $userId, Attribute $attribute ) {
 
 			return $ret;
 		} catch ( \Exception $e ) {
-			$this->logError( $userId, $e );
+			$this->logError( $userId, $e, "USER_ATTRIBUTE error saving to service" );
```
armonr: Should they begin with `USER_ATTRIBUTES` (plural) to match with https://kibana.wikia-inc.com/index.html#/dashboard/temp/AVOwK8vewuxBx3joFaRw ?

jsutterfield: Yep, that was my intention. Good catch. I'll update.

PR #10076, opened 2016-03-26
SUS-327 Fix the poll bars in speech bubbles

gabrys:
This removes a part of the speech bubble fix done in SUS-166. It removes
the `height: auto` on all `divs` within `.speech-bubble-message`. This
doesn't seem to break the original fix for `SUS-166` and also fixes the
current issue.

Also updated the sizing of bars a little bit to reflect the bigger font.
Made the poll bars 19px instead of 12px high and added 4px or margin
below them.

There's also a fix for `wfFixMalformedHTML` that makes the poll not to
load the CSS and JS while in the speech bubble if the poll is the only
content of the forum/wall message (an article comment probably too).


ghost:
:+1: LGTM

PR #10077, opened 2016-03-27
Lyr 237 2

SeanColombo:
Added territory=US parameter to the LyricFind API calls. Now they will return successfully as desired.

We'd like this to be released before April 1st so that we will have the automatic reporting restored for all of Q2 2016.

The test-case that this works is that this page now returns a status code 200:
http://lyrics.sean.wikia-dev.com/wikia.php?controller=LyricFind&method=track&title=Cake%3AGuitar&amgid=0&gracenoteid=0&rand=55725133


owend:
LGTM.  We already did our release for today but I can merge this to release branch.

PR #10078, opened 2016-03-28
RECIRC-81: Add Lateral as a recommendation provider

pauloslund:
This PR exposes some weaknesses in how the placement test is currently being run. However, we are coming upon the end of the test soon and this code will soon be deprecated some of the "solutions" are a little "creative"

### DIFF
--- a/extensions/wikia/Recirculation/js/experiments/placement.js
+++ b/extensions/wikia/Recirculation/js/experiments/placement.js
@@ -10,6 +10,7 @@ require([
 	'ext.wikia.recirculation.views.footer',
 	'ext.wikia.recirculation.helpers.contentLinks',
 	'ext.wikia.recirculation.helpers.fandom',
+	'ext.wikia.recirculation.helpers.lateral',
 	'ext.wikia.recirculation.helpers.googleMatch',
 	'ext.wikia.adEngine.taboolaHelper',
 	require.optional('videosmodule.controllers.rail')
@@ -24,6 +25,7 @@ require([
 	footerView,
 	contentLinksHelper,
 	fandomHelper,
+	lateralHelper,
 	googleMatchHelper,
 	taboolaHelper,
 	videosModule
@@ -45,6 +47,18 @@ require([
 	}
 
 	switch (group) {
+		case 'LATERAL_FANDOM':
+			helper = lateralHelper();
+			view = railView();
+			isRail = true;
+			break;
+		case 'LATERAL_COMMUNITY':
+			helper = lateralHelper({
+				type: 'community'
+			});
+			view = railView();
+			isRail = true;
+			break;
 		case 'DESIGN_ONE':
 		case 'DESIGN_TWO':
 		case 'DESIGN_THREE':
--- a/extensions/wikia/Recirculation/js/experiments/placement.js
+++ b/extensions/wikia/Recirculation/js/experiments/placement.js
@@ -170,18 +170,19 @@ require([
 		}).loadData()
 			.then(incontent.render)
 			.then(incontent.setupTracking(experimentName))
+			.then(function() {
+				afterRailLoads(function() {
+					var rail = railView();
+					lateralHelper({
+						type: 'fandom',
+						count: 5
+					}).loadData()
+						.then(rail.render)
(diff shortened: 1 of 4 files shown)


extensions/wikia/Recirculation/js/helpers/LateralHelper.js
```diff
@@ -0,0 +1,114 @@
...
+		var title = options.type === 'fandom' ? $.msg('recirculation-fandom-title') : $.msg('recirculation-incontent-title');
+
+		return {
+			title: title,
+			items: items.slice(0, options.count)
+		};
+	}
+
+	return function(config) {
+		$.extend(options, config);
```
gabrys: I'm seeing this again. I want to say it may be quite confusing.

You're passing some options from the code that requires the module and `$.extends` overrides the module's `options` with what you passed. So if you use the module twice, you override the same `options` twice.

You could wrap all the "private" functions in here and have `options` scope inside here so each module "run" would have a separate `options` copy.

pauloslund: I agree with you, and this is actually an issue I ran in to while developing (that's why this entire section gets called after the first set of data is loaded https://github.com/Wikia/app/blob/RECIRC-81/extensions/wikia/Recirculation/js/experiments/placement.js#L190-L201). For some reason I wanted to stay away from wrapping all of the private functions inside of the export, can't really think of why though.

extensions/wikia/Recirculation/js/experiments/placement.js
```diff
@@ -130,7 +148,27 @@ require([
 		helper.loadData()
 			.then(view.render)
 			.then(view.setupTracking(experimentName))
-			.fail(function() {});
+			.fail(handleError);
+	}
+
+	function handleError() {
+		// If there is an error somewhere we render the control group with no tracking
+		if (errorHandled) {
```
gabrys, line 156: I'm a little bit lost in the error handling especially in case with Lateral both modules. I assume it's possible one of the modules errs and then we still fall back to the control group, but the other module still shows.

Unless there's a good reason (plenty of fails) I would just track the errors here and leave the users with empty modules instead :-)

extensions/wikia/Recirculation/js/helpers/LateralHelper.js
```diff
@@ -0,0 +1,114 @@
...
+], function ($, w, abTest) {
+	var libraryLoaded = false;
+
+	function loadLateral(callback) {
+		if (libraryLoaded && callback && typeof callback === 'function') {
+			callback(w.lateral);
+		}
+
+		var lateralScript = document.createElement('script'),
+			url = 'https://assets.lateral.io/recommendations.js';
```
gabrys: I don't think this variable is really needed (it's used in one place only)

extensions/wikia/Recirculation/js/helpers/LateralHelper.js
```diff
@@ -0,0 +1,114 @@
+/*global define*/
+define('ext.wikia.recirculation.helpers.lateral', [
+	'jquery',
+	'wikia.window',
+	'wikia.abTest',
+], function ($, w, abTest) {
+	var libraryLoaded = false;
+
+	function loadLateral(callback) {
+		if (libraryLoaded && callback && typeof callback === 'function') {
```
gabrys: Checking for both truthiness of callback and its type is a bit redundant and silences "callback passed, but not a function" error. I would just check for truthiness and expect it to be a function if present.

extensions/wikia/Recirculation/js/helpers/LateralHelper.js
```diff
@@ -0,0 +1,114 @@
...
+			$.each(data, function(index, item) {
+				if (!item.image) {
+					return;
+				}
+				item.thumbnail = item.image;
+				item.index = index;
+				items.push(item);
+			});
+
+			var title = options.type === 'fandom' ? $.msg('recirculation-fandom-title') : $.msg('recirculation-incontent-title');
```
gabrys: - multiple var statements
- use of ternary operator, I think a plain if/else is more readable

gabrys:
Some minor comments left. LGTM

PR #10080, opened 2016-03-29
ADEN-3121 Enable Krux based on real ESRB rating and from PHP level

fraszczakszymon:
Changes:
- run krux code when ESRB rating != 'ec'
- move basic checking to php level and don't attach code if not needed


jacekjursza:
LGTM

PR #10081, opened 2016-03-29
Fix one error found by Phan static analyzer

pawelchcki:
Some time ago I played with Phan I found this one error
/cc: @macbre


macbre:
LGTM

PR #10084, opened 2016-03-29
DAT-4018 europa theme var switch

idradm:
Portable Infobox Europa theme can be switched on/off by wiki factory variable.

https://wikia-inc.atlassian.net/browse/DAT-4018


sebastianmarzjan:
LG(TM)

PR #10090, opened 2016-03-29
WScribeClient: improve errors logging

macbre:
[PLATFORM-1955](https://wikia-inc.atlassian.net/browse/PLATFORM-1955)

Use a proper severity and emit log messages with context when reporting Scribe errors.

And get rid of `Wikia::log` calls that always emit log messages at `debug` level.

@wladekb


wladekb:
lgtm

PR #10093, opened 2016-03-30
Jpn 371 Adding a query for most active users in the last n days

lizlux:


extensions/wikia/CommunityPage/models/CommunityPageSpecialUsersModel.class.php
```diff
@@ -60,7 +60,84 @@ function () use ($user) {
...
+				global $wgExternalSharedDB;
+				$db = wfGetDB( DB_SLAVE );
+				$sqlData = ( new WikiaSQL() )
+					->SELECT( 'user_name, user_id, count(rev_id) AS revision_count' )
+					->FROM ( 'revision FORCE INDEX (user_timestamp)' )
+					->LEFT_JOIN( $wgExternalSharedDB . '.user')->ON( '(rev_user <> 0) AND (user_id = rev_user)' )
+					->LEFT_JOIN( 'user_groups ON (user_id = ug_user)' )
+					->WHERE( 'user_id' )->IS_NOT_NULL()
+					->AND_( 'rev_timestamp > DATE_SUB(now(), INTERVAL '. $interval . ')' )
+					->AND_( '(ug_group IS NULL or (ug_group <> "bot"))')
```
perjg, line 88: Is only <> working, and not !=?
They are supposed to be the same :)

lizlux, line 88: looks like it's a bug that they are not :(

PR #10095, opened 2016-03-30
JPN-360 Recent activity feed

perjg:


extensions/wikia/CommunityPage/models/CommunityPageSpecialUsersModel.class.php
```diff
@@ -248,4 +248,27 @@ public function getRecentlyJoinedUsers( $days = 14 ) {
 
 		return $out;
 	}
+
+	public function getRecentActivityData() {
```
lizlux: I'd say this should go on the wiki model instead of the user model, since it's pertaining more to wiki data that user data. That's the distinction I had in my head when I split these two models out anyway.

lizlux: Or maybe even in the controller

PR #10103, opened 2016-03-30
DAT-4030 Update image and copy for Europa in WikiFeatures

sebastianmarzjan:


idradm:
https://github.com/Wikia/app/pull/10104

PR #10113, opened 2016-03-30
SOC-409 don't send emails to unconfirmed email addresses

jsutterfield:
Ticket: https://wikia-inc.atlassian.net/browse/SOC-409

Don't send emails to users who haven't confirmed their email address.

This is accomplished by adding an additional check inside of `EmailController:: assertCanEmail` which checks if target user's email is confirmed before sending.

The AC also state that confirmation emails and password reminder emails (which includes fb disconnect), should be sent no matter what. That will still be the case since the controllers for those emails override the `assertCanEmail` method which you can see in the PR below. Those are the only 2 email controllers which override the `assertCanEmail` method and do not call the base class version as well.


kvas-damian:
LGTM

armonr:
LGTM too

PR #10114, opened 2016-03-31
JPN-404 Only use VE when parameters include 'preload'

perjg:


extensions/CreateBox/CreateBox.php
```diff
@@ -132,8 +132,11 @@ function acRedirect( $title, $action ) {
 	global $wgRequest, $wgOut, $wgUser;
 
 	$actionParam = 'action';
+	$params = $wgRequest->getValues();
+	$preload = $params['preload'];
+	$preloadIsMediaWikiPage = !empty( $preload ) && strpos($preload, 'MediaWiki:') === 0;
```
sebastianmarzjan: I recommend to check with Title:inNamespace instead of strpos due to possibility of encountering language-specific namespace aliases - even though for NS_MEDIAWIKI it is not a common thing, you can find some examples in the languages dir of the application repo.

Grunny:
This doesn't fix the redirect loop issue, it just limits the issue to only occur on MediaWiki namespace preloads. I'd suggest something like the following solution in InputBoxHooks instead, which will fix it for all preloads:

``` php
$redirectToVE = !empty( $params['preload'] ) &&
    EditorPreference::isVisualEditorPrimary() &&
    $user->isLoggedIn();

if ( empty( $params['prefix'] ) && !$redirectToVE ) {
    // Fine
    return true;
}

if ( $redirectToVE ) {
    $params['veaction'] = 'edit';
    unset( $params['action'] );
}
```

You then shouldn't need to make any changes to CreateBox.

perjg:
@Grunny Thanks for the suggestion: My patch for CreateBox was not needed, and it is likely that this could cause an redirect loop when combined with the InputBox hook.

I've also cleaned up the logic in the InputBox hook like you suggested.

My original reasoning for adding the check for MediaWiki namespace was to avoid any issue where VE would load for namespaces that do not support editing through VE. I've taken that part out of the code now.

Grunny:
LGTM.

PR #10115, opened 2016-03-31
SOC-1369 Fix URLs in message wall reply emails

jsutterfield:
Ticket: https://wikia-inc.atlassian.net/browse/SOC-1369

This PR fixes links in the message wall reply emails which were incorrect. They were pointing to pages which only showed the individual replies (eg [this](http://community.wikia.com/wiki/Thread:918372)), rather than a page showing the whole thread (eg [this](http://community.wikia.com/wiki/Thread:918186)).

The problem was that we were using the pageId of the reply, rather than the pageId of the thread to construct the URLs. This is fixed by using the `parentId` value being passed into the controller, rather than `threadId` like we'd been using.

### DIFF
--- a/extensions/wikia/Email/Controller/WallMessageController.class.php
+++ b/extensions/wikia/Email/Controller/WallMessageController.class.php
@@ -182,7 +182,7 @@ class OwnWallMessageController extends WallMessageController {
 	}
 }
 
-class ReplyWallMessageController extends OwnWallMessageController {
+class ReplyWallMessageController extends WallMessageController {
 
 	/** @var \Title */
 	protected $containingThread;
--- a/extensions/wikia/Email/Controller/WallMessageController.class.php
+++ b/extensions/wikia/Email/Controller/WallMessageController.class.php
@@ -166,6 +166,7 @@ class OwnWallMessageController extends WallMessageController {
 	 * @return string
 	 */
 	protected function getSummary() {
+		jmark();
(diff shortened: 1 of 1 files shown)


extensions/wikia/Email/Controller/WallMessageController.class.php
```diff
@@ -182,7 +182,15 @@ protected function getSubject() {
 	}
 }
 
-class ReplyWallMessageController extends OwnWallMessageController {
+class ReplyWallMessageController extends WallMessageController {
```
jsutterfield, line 185: This was extending `getEmailSpecificFormFields`, but there's no need to. It overrides to the only 2 methods that class has. Extend the base wall message class instead.

extensions/wikia/Email/Controller/WallMessageController.class.php
```diff
@@ -182,7 +182,15 @@ protected function getSubject() {
...
+
+	/** @var \Title */
+	protected $containingThread;
+
+	public function initEmail() {
+		parent::initEmail();
+		$this->containingThread = \Title::newFromText( $this->getVal( 'parentId' ), NS_USER_WALL_MESSAGE );
```
jsutterfield, line 192: `parentId` is already available to us in the request. See [here](https://github.com/Wikia/app/blob/dev/extensions/wikia/WallNotifications/WallNotifications.class.php#L419).

kvas-damian:
LGTM

armonr:
:shipit:

PR #10116, opened 2016-03-31
Jpn 405 Adding the first modal interaction

lizlux:
Still a bit of a work in progress but I got a good start on the modal interaction, including showing the leaderboard view.


extensions/wikia/CommunityPage/CommunityPageSpecialController.class.php
```diff
@@ -105,13 +106,13 @@ protected function getTopAdminsData() {
 
 		$remainingAdminCount = count ( $topAdmins ) - 2;
 
-		return [
+		$this->response->setData( [
```
lizlux, line 109: Had to do this so the controller method could be used as a nirvana endpoint.

PR #10120, opened 2016-03-31
(CE-3571) Insert new infobox into querycache on creation

Grunny:
When a new portable infobox is created, add it to the querycache for the
AllInfoboxes list and purge its cache.

### DIFF
--- a/extensions/wikia/PortableInfobox/PortableInfoboxHooks.class.php
+++ b/extensions/wikia/PortableInfobox/PortableInfoboxHooks.class.php
@@ -111,4 +111,60 @@ class PortableInfoboxHooks {
 
 		return true;
 	}
+
+	/**
+	 * Insert a newly created infobox into querycache, and purge the list of
+	 * infoboxes.
+	 *
+	 * @param  Page     $page          The created page object
+	 * @param  User     $user          The user who created the page
+	 * @param  string   $text          Text of the new article
+	 * @param  string   $summary       Edit summary
+	 * @param  int      $minoredit     Minor edit flag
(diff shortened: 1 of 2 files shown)


extensions/wikia/PortableInfobox/PortableInfoboxHooks.class.php
```diff
@@ -111,4 +111,60 @@ public static function onBacklinksPurge( Array $articles ) {
...
+	}
+
+	/**
+	 * Check whether or not a page has an infobox.
+	 *
+	 * @param  Title   $title Title to check
+	 * @return boolean
+	 */
+	private static function hasInfobox( Title $title ) {
+		return $title->inNamespace( NS_TEMPLATE ) &&
```
adamkarminski, line 163: What if it's a custom template namespace like the `Marvel_Database:` one?

adamkarminski:
One question but the code LGTM :)

extensions/wikia/PortableInfobox/PortableInfoboxHooks.class.php
Grunny, line 163: This isn't handled by the AllInfoboxes list at the moment, so we're matching that criteria, as we [only retrieve template namespace pages](https://github.com/Wikia/app/blob/afc755da607e12364/extensions/wikia/PortableInfobox/querypage/AllinfoboxesQueryPage.php#L85) when rebuilding the data, which is why Marvel has almost no infoboxes in the list: http://marvel.wikia.com/api.php?action=query&list=allinfoboxes&limit=100&format=jsonfm :)

PR #10122, opened 2016-03-31
Ensure all ForumBoard are showed even if they are not havesort order in wikia props

sqreek:


pawelchcki:
LGTM

PR #10123, opened 2016-03-31
ensure all ForumBoard are showed even if they are not have sort order in wikia props

sqreek:


sqreek:
https://github.com/Wikia/app/pull/10122

PR #10124, opened 2016-03-31
Revert "ensure all ForumBoard are showed even if they are not have sort order in wikia props"

ljagiello:
Reverts Wikia/app#10123

https://www.dropbox.com/s/8sbq4wtiaqpu96s/Screenshot%202016-03-31%2013.06.16.png?dl=0


frankfarmer:
:shipit:

PR #10125, opened 2016-03-31
SOC-2258 write global watchlist additions as a bulk operation

jsutterfield:
Ticket: https://wikia-inc.atlassian.net/browse/SOC-2258

Whenever a new staff blogpost is added, our dbs are getting flooded with GlobalWatchlistTask::addWatchers tasks. The current implementation of that task is very inefficient, doing one db write for each user and maintaining a handle on the db the entire time.

This PR updates that task to do a single bulk write for all users at once.

### DIFF
--- a/extensions/wikia/GlobalWatchlist/GlobalWatchlistTask.class.php
+++ b/extensions/wikia/GlobalWatchlist/GlobalWatchlistTask.class.php
@@ -60,11 +60,14 @@ class GlobalWatchlistTask extends BaseTask {
 						$revision->getTimestamp(),
 						$revision->getTimestamp()
 					];
+
+					if ( count( $watchersToAdd ) >= self::MAX_BULK_WRITES ) {
+						$this->addWatchersToDb( $watchersToAdd );
+						$watchersToAdd = [];
+					}
 				}
 
-				foreach ( array_chunk( $watchers, self::MAX_BULK_WRITES ) as $watchersArray ) {
-					$this->addWatchersToDb( $watchersArray );
-				}
+				$this->addWatchersToDb( $watchersToAdd );
 			}
 		}
 	}
--- a/extensions/wikia/GlobalWatchlist/GlobalWatchlistTask.class.php
+++ b/extensions/wikia/GlobalWatchlist/GlobalWatchlistTask.class.php
@@ -62,7 +62,9 @@ class GlobalWatchlistTask extends BaseTask {
 					];
 				}
 
-				$this->addWatchersToDb( $watchersToAdd );
+				foreach ( array_chunk( $watchers, self::MAX_BULK_WRITES ) as $watchersArray ) {
+					$this->addWatchersToDb( $watchersArray );
+				}
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


extensions/wikia/GlobalWatchlist/GlobalWatchlistTask.class.php
```diff
@@ -43,34 +43,50 @@ public function addWatchers( $databaseKey, $nameSpace, array $watchers ) {
 
 			// Skip revisions that doesn't exist
 			if ( !empty( $revision ) ) {
+				$watchersToAdd = [];
 				$globalWatchlistBot = new GlobalWatchlistBot();
-
-				$db = wfGetDB( DB_MASTER, [ ], \F::app()->wg->ExternalDatawareDB );
 				foreach ( $watchers as $watcherID ) {
 					if ( $globalWatchlistBot->shouldNotSendDigest( $watcherID ) ) {
-						$this->clearGlobalWatchlistAll( $watcherID );
```
jsutterfield, line 51: Doing this `clearGlobalWatchlistAll()` here actually isn't necessary. When we run the `sendWeeklyDigest.php` script (which runs once a week), we'll call `clearGlobalWatchlistAll()` for every user in the db. So removing it from here to simplify.

See [here](https://github.com/Wikia/app/blob/dev/extensions/wikia/GlobalWatchlist/GlobalWatchlistBot.class.php#L13-L18).

extensions/wikia/GlobalWatchlist/GlobalWatchlistTask.class.php
```diff
@@ -43,34 +43,50 @@ public function addWatchers( $databaseKey, $nameSpace, array $watchers ) {
...
+			GlobalWatchlistTable::COLUMN_TIMESTAMP
+		];
+
+		$db = wfGetDB( DB_MASTER, [ ], \F::app()->wg->ExternalDatawareDB );
+		( new WikiaSQL() )
+			->INSERT()->INTO( GlobalWatchlistTable::TABLE_NAME, $columns )
+			->VALUES( $watchers )
+			// Do nothing on duplicate key - we already have that record in place
+			->ON_DUPLICATE_KEY_UPDATE(
+				[ GlobalWatchlistTable::COLUMN_TIMESTAMP => GlobalWatchlistTable::COLUMN_TIMESTAMP ]
```
jsutterfield, line 92: I changed this from `GlobalWatchlistTable::COLUMN_USER_ID` to `GlobalWatchlistTable::COLUMN_TIMESTAMP`. Since we're doing a bulk write, we don't have access to the $watcherId value like we did before. I tried updating to 

``` PHP
[ GlobalWatchlistTable::COLUMN_USER_ID => GlobalWatchlistTable::COLUMN_USER_ID ]
```

and it was throwing the following error `ERROR 1062 (23000): Duplicate entry '869155-0-0-FBPlugins' for key 'wikia_user'`. Updating this to use a non-indexed key fixes it and since this is essentially a no-op, I think it should be fine for our needs.

extensions/wikia/GlobalWatchlist/GlobalWatchlistTask.class.php
```diff
@@ -43,34 +43,50 @@ public function addWatchers( $databaseKey, $nameSpace, array $watchers ) {
...
+						\F::app()->wg->CityId,
+						$databaseKey,
+						$nameSpace,
+						$revision->getId(),
+						$revision->getTimestamp(),
+						$revision->getTimestamp()
+					];
 				}
+
+				$this->addWatchersToDb( $watchersToAdd );
```
armonr, line 70: Should this maybe divided into batches of say 1K or 10K?

jsutterfield, line 70: That's a good idea. I tried to check the flower logs to see how many users are actually being sent over during these giant tasks, but they've expired so I don't know. 1k seems like a sane default though. I'll update.

armonr:
LGTM!

PR #10128, opened 2016-04-01
(JPN-404) Only redirect to VE if VE is enabled for that namespace

Grunny:
The previous code was always redirecting to VE even if VE was disabled
for a namespace. This checks if the namespace has VE enabled first which
needs to occur after making changes to the title with the prefix parameter.

/cc @perjg


perjg:
@Grunny I checked that our test cases for using InputBox/CreateBox with VE still work properly with this change.

LGTM.

PR #10129, opened 2016-04-01
SOC-2164 Discussion moderators should not be permitted to delete blog

armonr:
Adds a hook to the pre-deletion action to make sure that users have delete permission prior to deleting a blog article. As a result, Discussion moderators will not be able to use a workaround to delete blog articles; they are and still should be able to delete blog comments.

/CC @jsutterfield

### DIFF
--- a/extensions/wikia/Blogs/BlogLockdown.php
+++ b/extensions/wikia/Blogs/BlogLockdown.php
@@ -165,8 +165,20 @@ class BlogLockdown {
 	 */
 	public static function onBeforeDeletePermissionErrors( &$article, &$title, &$user, &$permission_errors ) {
 		// Only users with delete permission can delete a blog article
+
+		$accessErrorKey = 'badaccess-group0';
 		if ( $title->getNamespace() == NS_BLOG_ARTICLE && !$user->isAllowed( 'delete' ) ) {
-			$permission_errors[] = [ 'badaccess-group0' ];
+			$errorExists = false;
+			foreach ( $permission_errors as $errors ) {
+				if ( in_array( $accessErrorKey, $errors ) ) {
+					$errorExists = true;
+					break;
+				}
+			}
+
+			if ( !$errorExists ) {
+				$permission_errors[] = [ $accessErrorKey ];
+			}
 		}
 
 		return true;
--- a/extensions/wikia/Blogs/BlogLockdown.php
+++ b/extensions/wikia/Blogs/BlogLockdown.php
@@ -152,4 +152,23 @@ class BlogLockdown {
 
 		return $return;
 	}
+
+	/**
+	 * Checks permission before delete action on blog articles
+	 *
+	 * @param Article $article
+	 * @param Title $title
+	 * @param User $user
+	 * @param array $permission_errors
+	 *
+	 * @return boolean because it's a hook
+	 */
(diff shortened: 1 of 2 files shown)


extensions/wikia/Blogs/Blogs.php
```diff
@@ -112,6 +112,9 @@ function wfBlogsOnBeforeInitialize( &$title, &$article, &$output, User $user, $r
 // Usages of images on blogs on file pages
 $wgHooks['FilePageImageUsageSingleLink'][] = 'BlogsHelper::onFilePageImageUsageSingleLink';
 
+// Checking that user is permitted to delete blog articles
+$wgHooks['BeforeDeletePermissionErrors'][] = 'BlogLockdown::onBeforeDeletePermissionErrors';
```
jsutterfield, line 116: It looks like `BeforeDeletePermissionErrors` will work, but I think the `userCan` might be more appropriate since we're doing a perm check. It gets called right before via [this method](https://github.com/Wikia/app/blob/dev/includes/Article.php#L1388) if you follow the calls down. Thoughts?

extensions/wikia/Blogs/BlogLockdown.php
```diff
@@ -152,4 +152,23 @@ public static function userCan( $title, $user, $action, &$result ) {
...
+	 * @param Article $article
+	 * @param Title $title
+	 * @param User $user
+	 * @param array $permission_errors
+	 *
+	 * @return boolean because it's a hook
+	 */
+	public static function onBeforeDeletePermissionErrors( &$article, &$title, &$user, &$permission_errors ) {
+		// Only users with delete permission can delete a blog article
+		if ( $title->getNamespace() == NS_BLOG_ARTICLE && !$user->isAllowed( 'delete' ) ) {
```
jsutterfield, line 170: `delete` seems like a really general permission name. Will this get us into trouble if the user is allowed to delete other things like regular wiki pages or article comments? Do you know what delete specifically refers to?

armonr, line 170: The "delete" permission works like the "super delete" permission in MW, so if there is no fine-grained permission to delete a specific type of content, then it falls back on "delete"; It is given to a few groups which you find [here](https://github.com/Wikia/config/blob/dev/PermissionsDefinesBeforeWikiFactory.php) if you search for `'delete'`.

extensions/wikia/Blogs/Blogs.php
armonr, line 116: It is literally impossible to do this because this method gets called multiple times for a single requests and with separate action types. That was my first possible solution to no avail.

jsutterfield:
LGTM

extensions/wikia/Blogs/BlogLockdown.php
```diff
@@ -165,8 +165,20 @@ public static function userCan( $title, $user, $action, &$result ) {
...
+			foreach ( $permission_errors as $errors ) {
+				if ( in_array( $accessErrorKey, $errors ) ) {
+					$errorExists = true;
+					break;
+				}
+			}
+
+			if ( !$errorExists ) {
+				$permission_errors[] = [ $accessErrorKey ];
+			}
```
jsutterfield, line 181: LGTM!

PR #10131, opened 2016-04-01
JPN-394 Use string for task page call to action for both task page 

perjg:
and community page.

With this change the communitypageexperiment-entry-join is now unused, but I didn't bother removing it since the whole extension will be removed once the experiment is over.


Grunny:
LGTM.

PR #10132, opened 2016-04-01
Jpn 360 Reverting any changes we made to Grunny's experiment page 

lizlux:
This PR reverts anything we did to mess up the CommunityPageExperiment extension so we can merge JPN-360 to dev safely as long as the new extension is disabled. 

Matching config PR https://github.com/Wikia/config/pull/1721/files. 

See https://github.com/Wikia/app/compare/JPN-413 for what the new diff will be between our work and production after this PR is merged. Notice how the experiment extension is unchanged.


Grunny:
LGTM.

PR #10135, opened 2016-04-01
ensure all ForumBoard are showed even if they are not have sort order in wikia props

sqreek:

@Wikia/sustaining-engineering-team

### DIFF
--- a/extensions/wikia/Forum/Forum.class.php
+++ b/extensions/wikia/Forum/Forum.class.php
@@ -27,8 +27,19 @@ class Forum extends Walls {
 	const LEN_TOO_SMALL_ERR = -2;
 
 	public function getBoardList( $db = DB_SLAVE ) {
-		$titlesBatch = $this->getTitlesForNamespace( $db, NS_WIKIA_FORUM_BOARD );
+		$titlesBatch = TitleBatch::newFromConds( [ ],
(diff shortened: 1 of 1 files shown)


pawelchcki:
Looks good
But how do we know it wont melt down the DB ?

sqreek:
```
SELECT /* Forum::getBoardList 91.102.115.105 - 9d7552d7-6ac7-4a31-be9f-8a47fc7580d7 */  page.*  FROM `page`  WHERE page.page_namespace = '2000'  ORDER BY page_title
```

pawelchcki:
If I understand this correctly - before you had to fetch page id by separate query for each entry?  And now its prepopulated ? so you only fetch page.ids only once ?

pawelchcki:
Ok we discussed differences with previous PR offline. 
So now it LGTM

ghost:
Was it about the `JOIN`?

pawelchcki:
https://github.com/Wikia/app/pull/10122/commits/799f85890f30d936b775abbf69bcea4df38ae4bd#diff-246b95c08953161176632a5dbc3a5fffL44

sqreek:
Long story short:
previously I modified getTitlesForNamespace and removed join there.
That cause all wall got Cartesian result.

here I only load all boards for special:Forum, and add missing order for those that dont contains order in page_wikia_props.

PR #10136, opened 2016-04-01
Removed the ToneFuze ads from LyricWiki. This fixed a bug when runnin…

SeanColombo:
…g them on Mercury which was causing white-pages on Artist pages. It sounded like we weren't making much 39772 from them anyway.

### DIFF
--- a/extensions/3rdparty/LyricWiki/Tag_Lyric.php
+++ b/extensions/3rdparty/LyricWiki/Tag_Lyric.php
@@ -58,7 +58,6 @@ $wgExtensionCredits["parserhook"][]=array(
 if(isset($wgScriptPath))
 {
 	#Instruct mediawiki to call LyricExtension to initialise new extension
-	$wgExtensionFunctions[] = "lyricTag";
 	$wgHooks['ParserFirstCallInit'][] = "lyricTag_InstallParser";
 	$wgHooks['BeforePageDisplay'][] = "lyricTagCss";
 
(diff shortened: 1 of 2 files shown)


extensions/3rdparty/LyricWiki/Tag_Lyric.php
```diff
@@ -66,19 +65,10 @@
...
-	// Keep track of whether this is the first <lyric> tag on the page - this is to prevent too many Ringtones ad links.
-	global $wgFirstLyricTag;
-	$wgFirstLyricTag = true;
-}
-
 function lyricTag_InstallParser( $parser ) {
 	#install hook on the element <lyric>
 	$parser->setHook("lyric", "renderLyricTag");
 	$parser->setHook("lyrics", "renderLyricTag");
-	$parser->setHook("tonefuze", "renderToneFuzeTag");
```
rogatty, line 81: Are we sure that there are no more `<tonefuze>` tags in use? I guess it wouldn't be that bad if we found out after removing this code and seeing it inlined in some articles.

rogatty:
LGTM

extensions/3rdparty/LyricWiki/Tag_Lyric.php
SeanColombo, line 81: FWIW: I did poke around my devbox to find templates that used this, then updated those templates in production on LyricWiki at that time. Hopefully we'll be good, but the community should catch it pretty quickly if we failed.

PR #10141, opened 2016-04-01
LF changed it's response-code for SUCCESS_NO_LYRICS from 106 to 104 s…

SeanColombo:
…ometime in the last week or so. The requests should already be going through fine, but we'll be returning 404s to our client (which makes it artificially look like an issue to us) so fixing this up will let us get to a state where 404s will actually be 404s.


owend:
LGTM

PR #10142, opened 2016-04-01
SEO-208 Cache Special:AllPages similarly to Special:Images

gabrys:
Cache for anonymous users only. Cache for 5 minutes. The robots
policy remains noindex, follow.


saipetch:
LGTM

PR #10143, opened 2016-04-01
SOC-2269 make sure we have watchers before calling addWatchersToDb

jsutterfield:
Ticket: https://wikia-inc.atlassian.net/browse/SOC-2269

Since we're now doing bulk updates, I forgot the case where all users can be filtered out in the check [here](https://github.com/Wikia/app/blob/dev/extensions/wikia/GlobalWatchlist/GlobalWatchlistTask.class.php#L51), which results in invalid SQL when we make the last call [here](https://github.com/Wikia/app/blob/dev/extensions/wikia/GlobalWatchlist/GlobalWatchlistTask.class.php#L70).

This PR fixes that by checking if we have watchers before calling `addWatchersToDb`.


armonr:
:+1:

PR #10156, opened 2016-04-05
Aden 3051 advertisement label

jacekjursza:
Adding 'advertisement' label (along with language-specific translations) to one of Wikia adv slots (incontent leaderboard).

### DIFF
--- a/extensions/wikia/AdEngine/js/slot/inContent.js
+++ b/extensions/wikia/AdEngine/js/slot/inContent.js
@@ -1,9 +1,10 @@
 /*global define*/
 define('ext.wikia.adEngine.slot.inContent', [
 	'ext.wikia.adEngine.adTracker',
+	'JSMessages',
 	'wikia.log',
 	'wikia.window'
-], function (adTracker, log, win) {
+], function (adTracker, msg, log, win) {
 	'use strict';
 
 	var logGroup = 'ext.wikia.adEngine.slot.inContent',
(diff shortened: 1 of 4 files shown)


skins/oasis/css/core/ads.scss
```diff
@@ -122,6 +122,26 @@ $taboola-small-breakpoint: "#{$screen} and (max-width: 420px)";
...
+#INCONTENT_WRAPPER {
+  text-align: center;
+}
+
+#INCONTENT_LEADERBOARD {
+  display: inline-block;
+}
+
+#INCONTENT_LEADERBOARD:before {
+  color: #656d77;
```
nandy-andy: Where does this color comes from? Is there a variable we can re-use here? How does it look on light and dark wikis? If it's just copied from fandom we might want to assign it to a variable and mention in the ticket it's now constant.

skins/oasis/css/core/ads.scss
```diff
@@ -122,6 +122,26 @@ $taboola-small-breakpoint: "#{$screen} and (max-width: 420px)";
...
+  content: attr(data-label);
```
nandy-andy, line 135: :gem:

extensions/wikia/AdEngine/js/slot/inContent.js
```diff
@@ -14,7 +15,7 @@ define('ext.wikia.adEngine.slot.inContent', [
 	 * Adds dynamically new slot in the right place and sends tracking data
 	 */
 	function init(slotName) {
-		var adHtml = '<div id="' + slotName + '" class="wikia-ad default-height"></div>',
+		var adHtml = '<div id="INCONTENT_WRAPPER"><div id="' + slotName + '" class="wikia-ad default-height" data-label="' + msg('adengine-advertisement') + '"></div></div>',
```
nandy-andy, line 18: Would it be over-engineering if we move it to a mustache template?

jacekjursza, line 18: Not sure if it's worth since in order to use mustache template I would need to load it via ajax call

nandy-andy:
LGTM :+1:

PR #10157, opened 2016-04-05
SOC-2280 | QueryString don't encode params two times

kvas-damian:
QueryString module encodes values on `setVal()` so there is no need for escaping it again in `toString()`

https://github.com/Wikia/app/blob/ee56454d5f1191dabbf93c3b9c58082d7d311ca9/resources/wikia/modules/querystring.js#L165-165

Ping @mixth-sense  @macbre @hakubo


macbre:
:+1:

PR #10162, opened 2016-04-05
ADEN-3017 Add language to default IVW cmKey

fraszczakszymon:


fraszczakszymon:
JS unit tests are failing because there is no node on machines. I've run them locally and they've passed.

drets:
LGTM 👍

PR #10165, opened 2016-04-05
[PLATFORM-2006] Fixed checking for user rights on special contributions

jcellary:
https://wikia-inc.atlassian.net/browse/PLATFORM-2006

In a change some time ago I accidentally started to check for user permissions of the target user instead of the currently logged in user when determining whether a link to rights management should be provided. Also fixed setting of the isself parameter.


macbre:
:+1:

PR #10168, opened 2016-04-05
SOC-1358 Throttle category add emails

jsutterfield:
Ticket: https://wikia-inc.atlassian.net/browse/SOC-1358

There have been reports by users of mass floods of category add emails (up to 35k) which can be triggered when a commonly used template is edited to include new categories. This PR adds a throttle so we only send up to 10 category add emails every 30 mins.

### DIFF
--- a/extensions/wikia/Email/Controller/CategoryAddController.class.php
+++ b/extensions/wikia/Email/Controller/CategoryAddController.class.php
@@ -72,8 +72,8 @@ class CategoryAddController extends EmailController {
 
 	private function assertEmailsNotThrottled() {
 		global $wgMemc;
-		$emailsSent = $wgMemc->get( $this->getTargetUserCacheKey() );
-		if ( !empty( $emailsSent['sent'] ) && $emailsSent['sent'] >=  self::EMAILS_PER_THROTTLE_PERIOD ) {
+		$emailsSent = (int) $wgMemc->get( $this->getTargetUserCacheKey() );
+		if ( $emailsSent >=  self::EMAILS_PER_THROTTLE_PERIOD ) {
 			throw new Check( 'Attempt to send too many emails' );
 		}
 	}
@@ -85,15 +85,11 @@ class CategoryAddController extends EmailController {
 	private function incrementEmailsSentCount() {
 		global $wgMemc;
 		$emailsSent = $wgMemc->get( $this->getTargetUserCacheKey() );
-		if ( $emailsSent == false ) {
-			$emailsSent = [
-				'sent' => 0,
-				'ttl' => time() + self::THROTTLE_PERIOD
-			];
+		if ( !is_int( $emailsSent ) ) {
+			$wgMemc->set( $this->getTargetUserCacheKey(), 0, time() + self::THROTTLE_PERIOD );
 		}
-		$emailsSent['sent']++;
 
-		$wgMemc->set( $this->getTargetUserCacheKey(), $emailsSent, $emailsSent['ttl'] );
+		$wgMemc->incr( $this->getTargetUserCacheKey() );
 	}
 
 	private function getTargetUserCacheKey() {
--- a/extensions/wikia/Email/Controller/CategoryAddController.class.php
+++ b/extensions/wikia/Email/Controller/CategoryAddController.class.php
@@ -73,7 +73,7 @@ class CategoryAddController extends EmailController {
 	private function assertEmailsNotThrottled() {
 		global $wgMemc;
 		$emailsSent = $wgMemc->get( $this->getTargetUserCacheKey() );
-		if ( $emailsSent != false && $emailsSent['sent'] >=  self::EMAILS_PER_THROTTLE_PERIOD ) {
+		if ( !empty( $emailsSent['sent'] ) && $emailsSent['sent'] >=  self::EMAILS_PER_THROTTLE_PERIOD ) {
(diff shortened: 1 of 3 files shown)


extensions/wikia/Email/Controller/CategoryAddController.class.php
```diff
@@ -58,6 +65,41 @@ protected function assertValidPageAddedToCategory() {
...
 
+	public function assertCanEmail() {
+		parent::assertCanEmail();
+		$this->assertEmailsNotThrottled();
+	}
+
+	private function assertEmailsNotThrottled() {
+		global $wgMemc;
+		$emailsSent = $wgMemc->get( $this->getTargetUserCacheKey() );
+		if ( $emailsSent != false && $emailsSent['sent'] >=  self::EMAILS_PER_THROTTLE_PERIOD ) {
```
kvas-damian: Please use strict type comparison.
Additionally I would consider checking if `$emailsSent` is valid and contains right keys.

extensions/wikia/Email/Controller/CategoryAddController.class.php
```diff
@@ -58,6 +65,41 @@ protected function assertValidPageAddedToCategory() {
...
+				'ttl' => time() + self::THROTTLE_PERIOD
+			];
+		}
+		$emailsSent['sent']++;
+
+		$wgMemc->set( $this->getTargetUserCacheKey(), $emailsSent, $emailsSent['ttl'] );
+	}
+
+	private function getTargetUserCacheKey() {
+		return $this->targetUser->getId() . ":sentCategoryAddEmailCount";
```
kvas-damian, line 96: this key is a bit misleading while we store an array here.

extensions/wikia/Email/Controller/CategoryAddController.class.php
```diff
@@ -58,6 +65,41 @@ protected function assertValidPageAddedToCategory() {
...
+		$emailsSent = $wgMemc->get( $this->getTargetUserCacheKey() );
+		if ( $emailsSent == false ) {
+			$emailsSent = [
+				'sent' => 0,
+				'ttl' => time() + self::THROTTLE_PERIOD
+			];
+		}
+		$emailsSent['sent']++;
+
+		$wgMemc->set( $this->getTargetUserCacheKey(), $emailsSent, $emailsSent['ttl'] );
```
kvas-damian: What do you think about using [`incr()`](https://github.com/Wikia/app/blob/97373e2f5e217093b3308f662892b369a75734ba/includes/objectcache/MemcachedClient.php#L669-669) method?
According to my knowledge it doesn't affect ttl time.

extensions/wikia/Email/tests/CategoryAddControllerTest.php
```diff
@@ -0,0 +1,74 @@
+<?php
+
+use \Email\Controller\CategoryAddController;
+
+class CategoryAddControllerTest extends WikiaBaseTest {
```
kvas-damian, line 5: What do you think about adding second test case for happy path?

extensions/wikia/Email/Controller/CategoryAddController.class.php
armonr: Probably

``` php
if ( !empty( $emailsSent[ 'sent' ] ) && $emailsSent['sent'] >=  self::EMAILS_PER_THROTTLE_PERIOD ) {
```

extensions/wikia/Email/tests/CategoryAddControllerTest.php
jsutterfield, line 5: :+1:

extensions/wikia/Email/Controller/CategoryAddController.class.php
jsutterfield: You're right, I did some testing and that doesn't affect ttl. That's a cleaner solution.

extensions/wikia/Email/Controller/CategoryAddController.class.php
jsutterfield, line 96: I updated the value of the cache to just store the emails sent, so now this is appropriate :)

kvas-damian:
LGTM

extensions/wikia/Email/Controller/CategoryAddController.class.php
```diff
@@ -58,6 +65,37 @@ protected function assertValidPageAddedToCategory() {
...
+
+	protected function afterSuccess() {
+		$this->incrementEmailsSentCount();
+	}
+
+	private function incrementEmailsSentCount() {
+		global $wgMemc;
+		$emailsSent = $wgMemc->get( $this->getTargetUserCacheKey() );
+		if ( !is_int( $emailsSent ) ) {
+			$wgMemc->set( $this->getTargetUserCacheKey(), 0, time() + self::THROTTLE_PERIOD );
```
armonr, line 89: Should the second argument be `1`?

jsutterfield, line 89: Nope, we initialize to 0, then it's incremented to 1 on L92.

armonr, line 89: Right. I mistakenly registered that line as an else clause :dizzy_face:

armonr:
LGTM

PR #10169, opened 2016-04-06
JPN-360 Community Page

perjg:
Initial work for new community page.

Note: This is a work in progress. The new extension we are developing will not be enabled before we reach MVP. However, we would like to merge the current work to dev in order to avoid having a single gigantic PR at the end.

The new extension will not be active before we set wgEnableCommunityPageExt.
We are testing this separately on sandbox, see JPN-413.


perjg:
@Grunny @owend  Sorry about the size of this PR: It is the collection of all work performed on the permanent community page by Liz and I during her 2 week stay in Japan.

I'd like to get all this merged back to dev sooner rather than later. Production won't be affected as it is still set up to run Grunny's community page experiment on some select Japanese communities.

Going forward, I'll try to make smaller PR's on individual tasks or small groups of tasks.

extensions/wikia/CommunityPage/CommunityPageSpecialController.class.php
```diff
@@ -0,0 +1,243 @@
...
+			'inviteFriendsText' => $this->msg( 'communitypage-invite-friends' )->plain(),
+			'headerWelcomeMsg' => $this->msg( 'communitypage-tasks-header-welcome' )->plain(),
+			'pageListEditText' => $this->msg( 'communitypage-page-list-edit' )->plain(),
+			'thisMonthText' => $this->msg( 'communitypage-this-month' )->plain(),
+			'showMonthlySummary' => $isMember,
+			'showAdminsSummary' => !$isMember,
+			'statPagesTitle' => $this->msg( 'communitypage-pages' )->plain(),
+			'statPagesNumber' => 45,
+			'statPageViewsTitle' => $this->msg( 'communitypage-pageviews' )->plain(),
+			'statPageViewsNumber' => '1,049',
```
Grunny: Why hardcoded?

extensions/wikia/CommunityPage/scripts/ext.communityPage.js
```diff
@@ -0,0 +1,120 @@
+require([
+	'jquery',
+	'wikia.ui.factory',
+    'wikia.mustache',
```
Grunny: Spaces -> tabs per conventions.

extensions/wikia/CommunityPage/CommunityPageSpecialController.class.php
perjg: This was for demonstration purposes: It is the stuff that we haven't hooked up to real data yet.

extensions/wikia/CommunityPage/CommunityPageSpecialController.class.php
```diff
@@ -0,0 +1,243 @@
...
+			'statEditorsNumber' => 23,
+		] );
+	}
+
+	/**
+	 * Set context for contributorsModule template. Needs to be passed through the index method in order to work.
+	 * @return array
+	 */
+	public function getTopContributorsData() {
+		$userContribCount = $this->usersModel->getUserContributions( $this->wg->user );
```
Grunny: `$this->getUser()` would be preferable to `$this->wg->User` on a special page.

extensions/wikia/CommunityPage/CommunityPageSpecialController.class.php
```diff
@@ -0,0 +1,243 @@
...
+
+		// remove user styles and js
+		$this->getOutput()->disallowUserJs();
+
+		$this->response->setValues( [
+			'adminWelcomeMsg' => $this->msg( 'communitypage-tasks-admin-welcome' )->text(),
+			'pageListEmptyText' => $this->msg( 'communitypage-page-list-empty' )->plain(),
+			'showPopupMessage' => true,
+			'popupMessageText' => 'This is just a test message for the popup message box',
+			'userIsMember' => CommunityPageSpecialHelper::userHasEdited( $this->wg->User ),
```
Grunny: `$this->getUser()` would be preferable to `$this->wg->User` on a special page.

extensions/wikia/CommunityPage/CommunityPageSpecialController.class.php
```diff
@@ -0,0 +1,243 @@
...
+	 * @return array
+	 */
+	public function getTopAdminsData() {
+		$topAdmins = CommunityPageSpecialUsersModel::filterGlobalBots(
+			// get all admins who have contributed in the last two years ordered by contributions
+			CommunityPageSpecialUsersModel::getTopContributors( 10, '2 YEAR', true )
+		);
+		$topAdminsDetails = $this->getContributorsDetails( $topAdmins );
+
+		$remainingAdminCount = count ( $topAdmins ) - 2;
```
Grunny: Minor coding convention issue: Shouldn't be a space after `count` (`count( $topAdmins )`). Same elsewhere in the changes.

extensions/wikia/CommunityPage/CommunityPageSpecialController.class.php
```diff
@@ -0,0 +1,243 @@
...
+			'members' => $allMembers,
+		] );
+	}
+
+	/**
+	 * Set context for recentActivityModule template. Needs to be passed through the index method in order to work.
+	 * @return array
+	 */
+	protected function getRecentActivityData() {
+		$data = $this->sendRequest('LatestActivityController', 'executeIndex')->getData();
```
Grunny: Out coding conventions are ["spacey"](https://www.mediawiki.org/wiki/Manual:Coding_conventions/PHP#Spaces) so throughout the changes make sure you're consistently using the spacey style, i.e. `$this->sendRequest( 'LatestActivityController', 'executeIndex' )->getData();` We also have a [script](https://github.com/Wikia/guidelines/blob/master/PHP/bin/php-mediawiki-stylize) to help format things (though it isn't perfect).

extensions/wikia/CommunityPage/CommunityPageSpecialController.class.php
```diff
@@ -0,0 +1,243 @@
...
+		$this->response->setData([
+			'allText' => $this->msg( 'communitypage-modal-tab-all' )->plain(),
+			'allCount' => $memberCount,
+			'adminsText' => $this->msg( 'communitypage-modal-tab-admins' )->plain(),
+			'adminsCount' => $adminData['adminCount'],
+			'leaderboardText' => $this->msg( 'communitypage-modal-tab-leaderboard' )->plain(),
+		]);
+	}
+
+	protected function addAssets() {
```
Grunny: Why not `private`? Are you expecting to extend this class?

perjg: Not really, it is just an old habit of mine to prefer protected over private to make any future subclassing easier.

I can change to private if that is preferred in our codebase.

extensions/wikia/CommunityPage/CommunityPageSpecialController.class.php
perjg: Is there also a linter script I can use? I use the guidelines hooks, but they don't seem to catch this kind of stuff.

extensions/wikia/CommunityPage/models/CommunityPageSpecialUsersModel.class.php
```diff
@@ -0,0 +1,371 @@
...
+				$sqlData = ( new WikiaSQL() )
+					->SELECT( '*' )
+					->FROM ( 'wikia_user_properties' )
+					->WHERE ( 'wup_property' )->EQUAL_TO( 'firstContributionTimestamp' )
+					->AND_ ( 'wup_user' )->EQUAL_TO( $user->getID() )
+					->AND_ ( 'wup_value > DATE_SUB(now(), INTERVAL 2 YEAR)' )
+					->ORDER_BY( 'wup_value DESC' )
+					->run( $db, function ( ResultWrapper $result ) {
+						$out = [];
+						while ( $row = $result->fetchRow() ) {
```
Grunny: Rather than doing your own loop, just use WikiaSQL's `runLoop` rather than `run`, i.e. something like (untested):

``` php
$sqlData = ( new WikiaSQL() )
    // ...query built here...
    ->runLoop( $db, function ( &$sqlData, $row ) {
        $sqlData[] = [
            'wup_value' => $row->wup_value,
        ];
    } );
```

extensions/wikia/CommunityPage/models/CommunityPageSpecialUsersModel.class.php
```diff
@@ -0,0 +1,371 @@
...
+						return $out;
+					} );
+
+				return $sqlData;
+			}
+		);
+
+		if (count( $data ) > 0) {
+			return $data[0]['wup_value'];
+		} else {
```
Grunny: Minor: I think it's cleaner without the `else`:

``` php
if ( count( $data ) > 0 ) {
    return $data[0]['wup_value'];
}
return null;
```

extensions/wikia/CommunityPage/models/CommunityPageSpecialUsersModel.class.php
```diff
@@ -0,0 +1,371 @@
...
+					->WHERE( 'user_id' )->IS_NOT_NULL()
+					->AND_( 'rev_timestamp > DATE_SUB(now(), INTERVAL '. $interval . ')' )
+					->AND_( '(ug_group IS NULL or (ug_group <> "bot"))' . $adminFilter)
+					// TOOD: also filter by glboal bot user ids?
+					->GROUP_BY( 'user_name' )
+					->ORDER_BY( 'revision_count' )->DESC()
+					->LIMIT( $limit )
+					->run( $db, function ( ResultWrapper $result ) {
+						$out = [];
+						while ( $row = $result->fetchRow() ) {
```
Grunny: Once again, can use `runLoop`.

extensions/wikia/CommunityPage/models/CommunityPageSpecialUsersModel.class.php
```diff
@@ -0,0 +1,371 @@
...
+				$db = wfGetDB( DB_SLAVE, [], $wgExternalSharedDB );
+
+				$sqlData = ( new WikiaSQL() )
+					->SELECT( 'user_id' )
+					->FROM ( 'user' )
+					->LEFT_JOIN( 'user_groups ON (user_id = ug_user)' )
+					->WHERE( 'ug_group' )->EQUAL_TO( 'bot-global' )
+					->run( $db, function ( ResultWrapper $result ) {
+						$out = [];
+						while ( $row = $result->fetchRow() ) {
```
Grunny: `runLoop`

extensions/wikia/CommunityPage/models/CommunityPageSpecialUsersModel.class.php
```diff
@@ -0,0 +1,371 @@
...
+				$db = wfGetDB( DB_SLAVE );
+
+				$sqlData = ( new WikiaSQL() )
+					->SELECT( 'count(rev_id) AS revision_count' )
+					->FROM( 'revision' )
+					->WHERE('rev_user')->EQUAL_TO( $userId )
+					->AND_( 'rev_timestamp > DATE_SUB(now(), INTERVAL ' . $interval . ')' )
+					->run( $db, function ( ResultWrapper $result ) {
+						$out = 0;
+						while ( $row = $result->fetchRow() ) {
```
Grunny: `runLoop`

extensions/wikia/CommunityPage/models/CommunityPageSpecialUsersModel.class.php
```diff
@@ -0,0 +1,371 @@
...
+					->SELECT( '*' )
+					->FROM ( 'wikia_user_properties' )
+					->WHERE ( 'wup_property' )->EQUAL_TO( 'firstContributionTimestamp' )
+					->AND_ ( "wup_value > DATE_SUB(now(), INTERVAL 14 DAY)" )
+					->ORDER_BY( 'wup_value DESC' )
+					->LIMIT( $limit )
+					->run( $db, function ( ResultWrapper $result ) {
+						$out = [];
+
+						while ( $row = $result->fetchRow() ) {
```
Grunny: `runLoop`

extensions/wikia/CommunityPage/models/CommunityPageSpecialUsersModel.class.php
```diff
@@ -0,0 +1,371 @@
...
+					->SELECT( '*' )
+					->FROM ( 'wikia_user_properties' )
+					->LEFT_JOIN( 'user_groups')->ON( 'wup_user = ug_user' )
+					->WHERE ( 'wup_property' )->EQUAL_TO( 'firstContributionTimestamp' )
+					->AND_ ( "wup_value > DATE_SUB(now(), INTERVAL 2 YEAR)" )
+					->ORDER_BY( 'wup_value DESC' )
+					->run( $db, function ( ResultWrapper $result ) {
+						$out = [];
+
+						while ( $row = $result->fetchRow() ) {
```
Grunny: `runLoop`

extensions/wikia/CommunityPage/models/CommunityPageSpecialUsersModel.class.php
```diff
@@ -0,0 +1,371 @@
...
+				global $wgExternalSharedDB;
+				$db = wfGetDB( DB_SLAVE );
+				$adminFilter = '';
+				if ( $onlyAdmins ) {
+					$adminFilter = ' AND (ug_group = "sysop")';
+				}
+
+				$sqlData = ( new WikiaSQL() )
+					->SELECT( 'user_name, user_id, count(rev_id) AS revision_count' )
+					->FROM ( 'revision FORCE INDEX (user_timestamp)' )
```
Grunny, line 86: What was the reason for calculating this with the local revision table, rather than using the rollups in the datawarehouse that calculate user edit counts for the last week, month, etc.?

extensions/wikia/CommunityPage/CommunityPageSpecialController.class.php
Grunny: As a general OOP guideline, it's best to default to private unless otherwise needed so the class is better encapsulated and you can change the internal implementation without affecting anything that uses or extends your class. Protected should be used for things you know you want to be accessible/overridable by subclasses, and it can always be switched to protected if there is a need to extend the class.

extensions/wikia/CommunityPage/CommunityPageSpecialController.class.php
Grunny: Unfortunately, I don't think we have a linter for our PHP conventions.

extensions/wikia/CommunityPage/models/CommunityPageSpecialUsersModel.class.php
perjg, line 86: I don't remember the detailed reason, but I think it was because we need this data to be available for 1 week, 1 month and 2 years, so using the revision table is more flexible.

Maybe @lizlux remembers?

perjg:
@Grunny I've addressed all of your comments.

extensions/wikia/CommunityPage/models/CommunityPageSpecialUsersModel.class.php
lizlux, line 86: We were going back and forth for a bit on which table to use, and I think the answer should be whichever table is more efficient (while still accurate of course). It's possible that the data warehouse table is more efficient. The main concern I saw with the data warehouse was that it was unclear how best to filter out bots and IP addresses. I was hoping that this ticket https://wikia-inc.atlassian.net/browse/JPN-403 would address that issue as this query is also fairly slow. It looks like other queries were optimized but not this one. 

My suggestion is to keep that ticket open and revisit this query to make sure we're using the best one for the job. It shouldn't be a blocker for merging this code though. Maybe a "to do" comment with a ticket number is in order.

extensions/wikia/CommunityPage/CommunityPageSpecialController.class.php
```diff
@@ -0,0 +1,243 @@
...
+
+	/**
+	 * Set context for recentActivityModule template. Needs to be passed through the index method in order to work.
+	 * @return array
+	 */
+	private function getRecentActivityData() {
+		$data = $this->sendRequest( 'LatestActivityController', 'executeIndex' )->getData();
+		$recentActivity = [];
+
+		foreach ($data['changeList'] as $activity) {
```
Grunny: The code still isn't spacey in many places like here. I'd suggest running the PHP formatter from our guidelines repo over the files.

extensions/wikia/CommunityPage/CommunityPageSpecialHelper.class.php
```diff
@@ -0,0 +1,15 @@
...
+	 * Returns true if the current user has edited the current wiki
+	 *
+	 * @param User $user
+	 * @return bool
+	 */
+	public static function userHasEdited( User $user ) {
+		$firstRev = ( new CommunityPageSpecialUsersModel() )->getFirstRevisionDate( $user );
+		return !is_null( $firstRev );
+	}
+}
```
Grunny: Missing new line.

includes/wikia/services/DataMartService.class.php
```diff
@@ -394,7 +394,7 @@ function () use ( $app, $wikiId, $userIds, $periodId, $rollupDate ) {
 
 		return $events;
 	}
-
+
```
Grunny: ? :)

Grunny:
Other than coding conventions and the minor comments, looks OK for the initial version minus the incomplete parts.

perjg:
@Grunny I ran the script as you suggested. Hopefully that should take care of the remaining style issues.

Grunny:
:+1:

PR #10171, opened 2016-04-06
DAT-4034 handle navigation templates after they are parsed

idradm:


includes/wikia/parser/templatetypes/handlers/NavigationTemplate.class.php
```diff
@@ -16,11 +18,38 @@ class NavigationTemplate {
...
-		return $text;
+		return $html;
+	}
+
+	/**
+	 * @param $text
+	 * @return int
+	 */
+	private static function mark( $text ) {
+		$m = crc32( self::$mark );
```
sebastianmarzjan: Why do you want to use CRC here? Since self::$mark is a constant, maybe we can either use it directly and separate markers from text with separator like \x1F ?

includes/wikia/parser/templatetypes/tests/NavigationTemplateTest.php
```diff
@@ -58,4 +77,34 @@ public function getNavigationTemplates() {
...
+				"26880747 aksdjlfkj alksjdldf\nlkjsdl d",
+				"If block element in navigation template it should be removed"
+			],
+			[
+				"\x7f26880747akjsdlkjflk <div>asdf</div>kasjdlfkjdks ksdjlafkj26880747\x7f test",
+				" test",
+				"Single nav template with block should be removed"
+			],
+			[
+				"\x7f26880747akjsdlkjflk <div>asdf</div>kasjdlfkjdks ksdjlafkj26880747\x7f test",
```
sebastianmarzjan: Isn't it the duplicate of above case? Maybe you wanted to test for inline element and non-removal?

sebastianmarzjan:
LGTM; Please optionally consider adding a test case with and inline html tag too

PR #10178, opened 2016-04-06
CleanEventsLocalUsersMaintenance: remove entries that do not have matching user ID and name

macbre:
[PLATFORM-2094](https://wikia-inc.atlassian.net/browse/PLATFORM-2094)

Remove `events_local_users` rows that do not have matching user ID and name as they cause duplicated entries and confuse admins and ComSup members.

@wladekb

### DIFF
--- a/maintenance/wikia/cleanEventsLocalUsers.php
+++ b/maintenance/wikia/cleanEventsLocalUsers.php
@@ -6,10 +6,61 @@ require_once( __DIR__ . '/../Maintenance.php' );
  * Remove incorrect entries from specials.events_local_users
  *
  * @see PLATFORM-2061
+ * @see PLATFORM-2094
  */
 class CleanEventsLocalUsersMaintenance extends Maintenance {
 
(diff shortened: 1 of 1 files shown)


maintenance/wikia/cleanEventsLocalUsers.php
```diff
@@ -6,10 +6,62 @@
...
+				'user_id > 0'
+			],
+			__METHOD__
+		);
+
+		while ( $row = $res->fetchRow() ) {
+			$user_name = User::newFromId( $row['user_id'] )->getName();
+
+			// user_id and user_name do not match, remove this entry
+			if ( $user_name !== $row['user_name'] ) {
```
wladekb, line 39: How about users that were renamed? There can be valid entries that won't meet this condition.

macbre, line 39: The row will be removed, but the script that generates this data will put it back (using data from local `revision` table).

PR #10181, opened 2016-04-07
[PLATFORM-2054] Added bot and rollback groups to self removable groups

jcellary:
https://wikia-inc.atlassian.net/browse/PLATFORM-2054

Per com sups request added bot and rollback groups to self removable groups.


macbre:
:+1:

PR #10183, opened 2016-04-07
CE-3584: Move items from dropdown to VE toolbar

lukaszkonieczny:
Move items from dropdown to VE toolbar


extensions/VisualEditor/wikia/modules/ve/init/ve.init.wikia.ViewPageTarget.js
```diff
@@ -73,16 +73,6 @@ ve.init.wikia.ViewPageTarget.static.toolbarGroups = [
 	}
 ];
 
-ve.init.wikia.activateToolbarABTest();
-
-if ( ve.init.wikia.getToolbarABTestVariantNumber() === 1 ) {
```
Grunny, line 78: Since we're removing the AB test, I think we should cleanup/remove all the references to the toolbar AB test (i.e. in the track and optimizely modules as well).

Grunny:
LGTM.

Grunny:
Travis is failing because of `'abTestData' is defined but never used.` You just need to remove the now unused variable in ve.track.js.

PR #10185, opened 2016-04-07
SOC-1302 | tracking for the force login popup enhancement

rafalkalinski:
https://wikia-inc.atlassian.net/browse/SOC-1302
## Reviewers

@kvas-damian @bkoval


extensions/wikia/AuthModal/js/AuthModal.js
```diff
@@ -17,12 +19,31 @@ define('AuthModal', ['jquery', 'wikia.window'], function ($, window) {
...
 				close();
 				if (typeof onAuthSuccess === 'function') {
 					onAuthSuccess();
 				}
 			}
+
+			if (e.data.beforeunload && !closeTrackTimeoutId) {
+				// to avoid tracking 'close' action whenever the window is reloaded;
+				closeTrackTimeoutId = setTimeout(function () {
+					clearTimeout(closeTrackTimeoutId);
```
kvas-damian: Why do we need to clear timeout? While it's executed it's already cleared I believe.

kvas-damian: Maybe we should clear previous timeouts before setting this one?

extensions/wikia/AuthModal/js/AuthModal.js
```diff
@@ -95,12 +115,17 @@ define('AuthModal', ['jquery', 'wikia.window'], function ($, window) {
 			}
 
 			if (window.wgEnableNewAuthModal) {
+				if (params.forceLogin) {
+					popUpWindowParam += '&forceLogin=1';
```
kvas-damian: Can we avoid using/changing variables defined in parent scope?

extensions/wikia/AuthModal/js/AuthModal.js
```diff
@@ -17,12 +19,31 @@ define('AuthModal', ['jquery', 'wikia.window'], function ($, window) {
 		$(window).on('message.authPopUpWindow', function (event) {
 			var e = event.originalEvent;
 
-			if (typeof e.data !== 'undefined' && e.data.isUserAuthorized) {
+			if (typeof e.data === 'undefined') {
```
bkoval: Yeah, this check wasn't the best at the beginning, I think checking if e.data is of not of type object would make more sense? Or if it's truthy.

extensions/wikia/AuthModal/js/AuthModal.js
rafalkalinski: I'm removing clearTimeout, but there is no need for clear previous timeouts, because of closeTrackTimeoutId is checked in the condition above

kvas-damian:
LGTM

PR #10188, opened 2016-04-07
RECIRC 94: Improve loading of Lateral

pauloslund:
There are currently instances where the Lateral script is getting loaded twice, which is causing issues with analytics data. This change makes it so the call to load lateral will only get run once, and callbacks get added to a queue to be run after Lateral has finished loading. This PR also adds client side error logging for the placement test.

### DIFF
--- a/extensions/wikia/Recirculation/js/helpers/LateralHelper.js
+++ b/extensions/wikia/Recirculation/js/helpers/LateralHelper.js
@@ -30,7 +30,7 @@ define('ext.wikia.recirculation.helpers.lateral', [
 
 	function processQueue(lateral) {
 		while(queue.length > 0) {
-			var callback = queue.pop();
+			var callback = queue.shift();
 
 			callback(lateral);
 		}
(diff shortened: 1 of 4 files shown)


extensions/wikia/Recirculation/js/helpers/LateralHelper.js
```diff
@@ -4,29 +4,35 @@ define('ext.wikia.recirculation.helpers.lateral', [
...
+		}
+	}
 
-		// This function is called when the Lateral script has loaded
-		w.onLoadLateral = function(lateral) {
-			w.lateral = lateral;
-			libraryLoaded = true;
+	function processQueue(lateral) {
+		while(queue.length > 0) {
+			var callback = queue.pop();
```
gabrys: `pop` will start with the item that was `push` most recently. `shift` should be better (well, if that's really a queue at least ;-) )

gabrys:
LGTM otherwise

extensions/wikia/Recirculation/js/helpers/LateralHelper.js
pauloslund: Good point! Updated

PR #10189, opened 2016-04-08
SOC-1987 SOC-2081DiscussionsLog IP Address lookup + eliminate duplicate records

armonr:
- IP lookup capability https://wikia-inc.atlassian.net/browse/SOC-1987
- Eliminate duplicates https://wikia-inc.atlassian.net/browse/SOC-2081

And lots of refactoring...

/CC @bkoval @jsutterfield


extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -9,203 +12,306 @@ class SpecialDiscussionsLogController extends WikiaSpecialPageController {
...
+	const NO_USER_MATCH_ERROR = 'discussionslog-no-user-match-error';
 
 	const DEFAULT_TEMPLATE_ENGINE = \WikiaResponse::TEMPLATE_ENGINE_MUSTACHE;
 
+	private $logger;
+	private $users;
+
 	public function __construct() {
 		parent::__construct( 'DiscussionsLog', '', false );
+		$this->users = [ ];
```
bkoval: I don't think there's a need for a space inside when it's empty :)

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -9,203 +12,306 @@ class SpecialDiscussionsLogController extends WikiaSpecialPageController {
...
 
-		$this->inputForm = $this->sendSelfRequest( 'inputForm', [ 'username' => $userName ] );
-		if ( !empty( $userName ) ) {
-			$this->userLog = $this->sendSelfRequest( 'userLog', [ 'username' => $userName ] );
+		if ( $userName ) {
+			$searchQuery = new UserQuery( $userName );
+		} else if ( $ipAddress ) {
+			$searchQuery = new IpAddressQuery( $ipAddress );
+		} else {
+			$searchQuery = null;
```
bkoval: How about initializing searchQuery with null and getting rid of the last else?

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -9,203 +12,306 @@ class SpecialDiscussionsLogController extends WikiaSpecialPageController {
...
+		}
+
+		$queryStringParams = [ ];
+		if ( $searchQuery ) {
+			$queryStringParams[ $searchQuery->getKeyName() ] = $searchQuery->getKey();
```
bkoval: Do we do spacing inside square brackets in mediawiki? :)

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -9,203 +12,306 @@ class SpecialDiscussionsLogController extends WikiaSpecialPageController {
...
+			throw new \InvalidArgumentException( wfMessage( 'discussionslog-multiple-input-error' )->escaped() );
+		}
+
+		return [ $userName, $ipAddress ];
+	}
+
+	private function requestUserLog( SearchQuery $searchQuery ) {
+		$this->setVal(
+			'userLog',
+			$this->sendSelfRequest( 'userLog', [ $searchQuery->getKeyName() => $searchQuery->getKey() ] ) );
```
bkoval: I'm not sure what guidelines say, but maybe moving last bracket and semicolon to the new line could work?

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -9,203 +12,306 @@ class SpecialDiscussionsLogController extends WikiaSpecialPageController {
...
-				"bool": {
-					"should": [{
-						"query_string": {
-							"query":"mobile_app.event.user_id:$userId"
-						}
-					}]
-				}
-			}
+	private function performSearchRequest( $url, $query ) {
+		$client = new \GuzzleHttp\Client();
```
bkoval, line 223: Why not use our Http class for doing calls? :)

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -9,203 +12,306 @@ class SpecialDiscussionsLogController extends WikiaSpecialPageController {
...
-			"query": {
-				"bool": {
-					"should": [{
-						"query_string": {
-							"query":"mobile_app.event.user_id:$userId"
-						}
-					}]
-				}
-			}
+	private function performSearchRequest( $url, $query ) {
```
bkoval: This method looks like a getter, maybe it would be benefitial to reflect that in the name, what do you think @armonr ?

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -9,203 +12,306 @@ class SpecialDiscussionsLogController extends WikiaSpecialPageController {
...
+			$userLogRecord->user = $user;
+
+			$ipHash[ $ipHashKey ] = true;
+			$records[ $rawTimestamp ] = $userLogRecord;
 		}
+	}
 
-		return true;
+	private function checkAccess() {
+		$action = 'specialdiscussionslog';
```
jsutterfield: How about we make this a class constant rather than a local var?

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -9,203 +12,306 @@ class SpecialDiscussionsLogController extends WikiaSpecialPageController {
...
 
 	public function index() {
 
-		if ( !$this->checkAccess() ) {
-			throw new \PermissionsError( 'specialdiscussionslog' );
-		}
+		$this->checkAccess();
 
 		$this->setHeaders();
+		$this->response->addAsset( '/extensions/wikia/SpecialDiscussionsLog/css/DiscussionsLog_Forms.scss' );
```
jsutterfield: This works, but I think the preferred way to do this is to create a new entry in [AssetsManager/config.php](https://github.com/Wikia/app/blob/dev/extensions/wikia/AssetsManager/config.php) and then add the asset via the name in the config. Like [this](https://github.com/Wikia/app/blob/dev/extensions/wikia/UserActivity/SpecialUserActivityController.class.php#L44).

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -9,203 +12,306 @@ class SpecialDiscussionsLogController extends WikiaSpecialPageController {
...
 		}
 	}
 
+	/**
+	 * Returns an array containing userName and ipAddress
+	 * @return array
+	 */
+	private function readAndValidateInput() {
+		$userName = $this->getVal( UserQuery::getKeyName(), null );
+		$ipAddress = $this->getVal( IpAddressQuery::getKeyName(), null );
```
jsutterfield: No need to pass `null` here for these second arguments, that's the default.

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -9,203 +12,306 @@ class SpecialDiscussionsLogController extends WikiaSpecialPageController {
...
+					[ 'exception' => $e ]
+				);
+				continue;
+			}
+			$rawTimestamp = $record->{'mobile_app.event.timestamp'};
+			$userLogRecord = new UserLogRecord();
+			$userLogRecord->app = $app;
+			$userLogRecord->ip = $ip;
+			$userLogRecord->language = $record->{'mobile_app.event.device_language'};
+			$city = $record->{'mobile_app.geo_ip.city'};
```
jsutterfield, line 279: How about we get rid of this local variable and just use `$record->{'mobile_app.geo_ip.city'}` for `$city` on the line below? For me the visual consistency of setting something to `$userLogRecord` on every line beats trying to break up a longer line with a local var.

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -9,203 +12,306 @@ class SpecialDiscussionsLogController extends WikiaSpecialPageController {
...
+			throw new \InvalidArgumentException( wfMessage( self::NO_USER_MATCH_ERROR )->escaped() );
 		}
 
-		return $user->getId();
+		$this->users[ $user->getId() ] = $user;
+
+		return $user;
 	}
 
 	public function userLog() {
```
jsutterfield, line 108: I realized that having this as public is actually a security concern since it's available via our API. I was able to access this while logged out by going to `http://jsutterfield.james.wikia-dev.com/wikia.php?controller=SpecialDiscussionsLog&method=userLog&username=james` and see the results.

2 ways I can think to fix this. Either we put a permissions check at the start of this function, or move the logic to get the userLog info into a separate private method and only have this responsible for passing those values to the template.

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -9,203 +12,306 @@ class SpecialDiscussionsLogController extends WikiaSpecialPageController {
...
-		}
+		$this->checkAccess();
 
 		$this->setHeaders();
+		$this->response->addAsset( '/extensions/wikia/SpecialDiscussionsLog/css/DiscussionsLog_Forms.scss' );
 
 		$this->wg->Out->setPageTitle( wfMessage( 'discussionslog-pagetitle' )->escaped() );
 
-		$userName = $this->wg->request->getVal( 'username', null );
+		list( $userName, $ipAddress ) = $this->readAndValidateInput();
```
jsutterfield: I think this could be simplified by making use of the `getParams()` method on the request object. That would get rid of this unpacking and then repacking we're doing of query params. I also don't think we have to check if there was a searchQuery before calling the userLog, the checks down in there make sure that nothing is rendered to the UI unless values were passed in.

I'm thinking something like this:

``` PHP
...
    $this->validateInput();
    $this->setVal( 'inputForm', $this->sendSelfRequest( 'inputForm', $this->request->getParams() ) );
    $this->setVal( 'userLog', $this->sendSelfRequest( 'userLog', $this->request->getParams() );
}
```

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -9,203 +12,306 @@ class SpecialDiscussionsLogController extends WikiaSpecialPageController {
...
+		$ipAddress = $this->getVal( IpAddressQuery::getKeyName(), null );
+
+		if ( !empty( $userName ) && !empty( $ipAddress ) ) {
+			throw new \InvalidArgumentException( wfMessage( 'discussionslog-multiple-input-error' )->escaped() );
+		}
+
+		return [ $userName, $ipAddress ];
+	}
+
+	private function requestUserLog( SearchQuery $searchQuery ) {
```
jsutterfield: Can we get by without this method? It seems to be just a proxy for a request to `userLog`. I see that it takes an interface, but we've already validated the data as part of `readAndValidateInput()` so this additional layer feels redundant.

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -9,203 +12,306 @@ class SpecialDiscussionsLogController extends WikiaSpecialPageController {
...
+		if ( !$user ) {
+			throw new \InvalidArgumentException( wfMessage( self::NO_USER_MATCH_ERROR )->escaped() );
+		}
+
+		return $user;
+	}
+
+	private function getUserById( $userId ) {
+		if ( !$userId ) {
+			throw new \InvalidArgumentException( wfMessage( self::NO_USER_MATCH_ERROR )->escaped() );
```
jsutterfield: What do you think about not localizing these error methods that only go to our logs? L164 looks like the only place this message bubbles up to the user, and it should definitely be translated there, but for the rest of these it'll make searching for errors harder since we'll have to do it in multiple languages.

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -9,203 +12,306 @@ class SpecialDiscussionsLogController extends WikiaSpecialPageController {
...
+				->escaped();
+
+			$this->userErrorMessage = wfMessage( self::NO_USER_MATCH_ERROR )->escaped();
+
+		} else if ( !empty( $ipAddress ) ) {
+			$userLogRecords = $this->aggregateLogByIpAddress( $ipAddress );
 
-		if ( count( $userLogRecords ) == 0 ) {
-			$hasNoUserLogRecords = true;
+			$this->logTableCaption = wfMessage( 'discussionslog-table-ip-caption' )
```
jsutterfield: Some of these response values are being set indirectly via member variables, and others are set explicitly inside of `$this->response->setValues(`. Can we set all of these these explicitly to be consistent?

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -9,203 +12,306 @@ class SpecialDiscussionsLogController extends WikiaSpecialPageController {
...
+				->params( [ $ipAddress ] )
+				->escaped();
+
+			$this->noUserLogRecordsMessage = wfMessage( 'discussionslog-no-ip-activity-error' )
+				->params( $ipAddress )
+				->escaped();
+
+			$this->userErrorMessage = wfMessage( 'discussionslog-no-ip-match-error' )->escaped();
```
jsutterfield: Is there any code path where this error message would be displayed? It looks like `$hasUserError` has to be set to `true` in order for this to show up, but I think that only can get set to true if we're doing a lookup via username above.

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -9,203 +12,306 @@ class SpecialDiscussionsLogController extends WikiaSpecialPageController {
...
+		return $user;
+	}
+
+	private function getUserById( $userId ) {
+		if ( !$userId ) {
+			throw new \InvalidArgumentException( wfMessage( self::NO_USER_MATCH_ERROR )->escaped() );
+		}
+
+		// Look at cache first
+		if ( !empty( $this->users[ $userId ] ) ) {
```
jsutterfield, line 94: Good call on the local cache :+1:

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
armonr: What's wrong with having the last else? It's a fallback rather than re-assigning the value in the first two cases :)

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
armonr: True, but this method knows how to construct the userLog parameters from `$searchQuery` and no one else has to worry about what `'userLog'` is!

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
armonr, line 279: In this case, I think I prefer to stay DRY :droplet:

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
armonr, line 223: I tried!
1. I cannot get it to work here!
2. There will be much less control over error handling
And honestly, I prefer GuzzleHttp because it is very reliable.

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
bkoval: Oh, nothing is wrong there! I just had my mind set on minimizing the number of lines :)

extensions/wikia/SpecialDiscussionsLog/templates/SpecialDiscussionsLog_userLog.mustache
```diff
@@ -1,15 +1,18 @@
...
+	<div class="log-error-message">{{userErrorMessage}}</div>
+{{/userErrorMessage}}
+{{^userErrorMessage}}
 	{{#hasNoUserLogRecords}}
-		<p>{{noUserLogRecordsMessage}}</p>
+		<div class="log-error-message">{{noUserLogRecordsMessage}}</div>
 	{{/hasNoUserLogRecords}}
 	{{^hasNoUserLogRecords}}
-		<p>{{logTableCaption}}</p>
+		<div class="log-table-caption">
```
bkoval: There's a paragraph inside a div here. Maybe we could get rid of one since it seems like the div contains nothing else except the stated paragraph? :neckbeard: 

Also, there is an html tag called "caption" (https://developer.mozilla.org/pl/docs/Web/HTML/Element/caption), which seems to serve the exact same purpose as this div. The only thing here is that it should be placed inside the table.

extensions/wikia/SpecialDiscussionsLog/templates/SpecialDiscussionsLog_inputForm.mustache
```diff
@@ -1,5 +1,11 @@
-<form method="get">
+<form method="get" class="form-username">
 	<label for="username">{{userNameLabel}}</label>
 	<input id="username" name="username" type="text" value="{{userName}}" required>
 	<input type="submit" value="{{viewLogsAction}}">
 </form>
+<div class="form-divider">{{orLabel}}</div>
+<form method="get" class="form-ipaddress">
+	<label for="ipaddress">{{ipAddressLabel}}</label>
+	<input id="ipaddress" name="ipaddress" type="text" value="{{ipAddress}}" required>
```
bkoval: Hey, I think we're using lower camelcase for ID's: https://github.com/Wikia/guidelines/blob/master/SCSS/CodingConventions.md#class-and-id-naming-conventions

extensions/wikia/SpecialDiscussionsLog/search/IpAddressQuery.php
```diff
@@ -0,0 +1,48 @@
...
+
+class IpAddressQuery implements SearchQuery {
+
+	private $searchKey;
+
+	public function __construct( $searchKey ) {
+		$this->searchKey = $searchKey;
+	}
+
+	function getKey() {
```
jsutterfield: `IpAddressQuery` and `UserQuery` no longer need `$searchKey`, `__construct()`, or `getKeyName()`.

armonr: `getKeyName` is still used; I'll get rid of the others.

jsutterfield:
LGTM!

bkoval:
LGTM, awesome work!

PR #10195, opened 2016-04-08
CE-3438: Remove key shortcut for Start a Wiki

lukaszkonieczny:
Remove key shortcut for Start a Wiki

@Wikia/community-engineering

### DIFF
--- a/extensions/wikia/GlobalShortcuts/scripts/AddDefaultShortcuts.js
+++ b/extensions/wikia/GlobalShortcuts/scripts/AddDefaultShortcuts.js
@@ -22,7 +22,7 @@ require(['GlobalShortcuts', 'PageActions', 'mw', 'wikia.loader'],
 		function addShortcuts() {
 			var INITIAL_SHORTCUTS = {
 				'general:StartWikia': {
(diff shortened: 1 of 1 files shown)


extensions/wikia/GlobalShortcuts/scripts/AddDefaultShortcuts.js
```diff
@@ -22,7 +22,7 @@ require(['GlobalShortcuts', 'PageActions', 'mw', 'wikia.loader'],
 		function addShortcuts() {
 			var INITIAL_SHORTCUTS = {
 				'general:StartWikia': {
```
adamkarminski, line 24: Isn't it better to remove this whole section?

kamilkoterba, line 24: I think it's good idea to keep it, then it will be accessible via actions manager without key shortcut. The goal behind this ticket is to prevent accidental key presses.

kamilkoterba:
LGTM

adamkarminski:
Tested, LGTM

PR #10199, opened 2016-04-11
[PLATFORM-2011] Fixed changing list of groups

jcellary:
https://wikia-inc.atlassian.net/browse/PLATFORM-2011

Added possibility to add/remove list of groups instead of a single value due to an edge case that when removing rights one by one, one can lack permissions to remove the next group because of removal of a previous one. Now all validation is done first and then actual removal takes place.

### DIFF
--- a/includes/specials/SpecialUserrights.php
+++ b/includes/specials/SpecialUserrights.php
@@ -244,19 +244,20 @@ class UserrightsPage extends SpecialPage {
 	 * @return Array: Tuple of added, then removed groups
 	 */
 	function doSaveUserGroups( $user, $groupsToAdd, $groupsToRemove, $reason = '' ) {
+		$groups = $user->getGroups();
 		$globalGroups = $this->permissionsService()->getConfiguration()->getGlobalGroups();
 		$changeable = $this->getUser()->changeableGroups();
 		$validGroupsToAdd = [];
 		$validGroupsToRemove = [];
 
 		foreach ( $groupsToAdd as $group ) {
-			if( self::canAdd( $globalGroups, $changeable, $group, $this->isself ) ) {
+			if( self::canAdd( $globalGroups, $changeable, $group, $this->isself ) && !in_array( $group, $groups ) ) {
 				$validGroupsToAdd[] = $group;
 			}
 		}
 
 		foreach ( $groupsToRemove as $group ) {
-			if( self::canRemove( $globalGroups, $changeable, $group, $this->isself ) ) {
+			if( self::canRemove( $globalGroups, $changeable, $group, $this->isself ) && in_array( $group, $groups ) ) {
 				$validGroupsToRemove[] = $group;
 			}
 		}
--- a/includes/specials/SpecialUserrights.php
+++ b/includes/specials/SpecialUserrights.php
@@ -261,6 +261,7 @@ class UserrightsPage extends SpecialPage {
 
 		$success = true;
 
+		//First add, then remove, otherwise addition may not be possible due to lack of permissions
 		if( $add ) {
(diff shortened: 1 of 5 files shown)


includes/specials/SpecialUserrights.php
```diff
@@ -256,36 +256,36 @@ function doSaveUserGroups( $user, $add, $remove, $reason = '' ) {
...
 
 		$newGroups = array_unique( $newGroups );
+		sort( $newGroups );
+
+		if( !$success ) {
+			$newGroups = $user->getGroups();
+			sort( $newGroups );
+			\Wikia\Logger\WikiaLogger::instance()->error(
+				'Error occurred while changing groups. Groups requested to add: ' . json_encode( $add )
+				. ", groups request to remove: " . json_encode( $remove ) );
```
macbre: I'd change the logging here to meet our new guidelines. Use `__METHOD__ . ' - 'Error occurred while changing groups`` as a message and  log the rest as a context. This way we can easily find the problems.

The same applies to the logging at line 206.

includes/specials/SpecialUserrights.php
```diff
@@ -256,36 +256,36 @@ function doSaveUserGroups( $user, $add, $remove, $reason = '' ) {
...
+		}
 
-		// Ensure that caches are cleared
-		$user->invalidateCache();
```
macbre, line 285: No cache invalidation needed anymore?

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
```diff
@@ -283,19 +293,29 @@ private function removeFromLocalGroup( \User $user, $group ) {
...
+		$result = true;
+		try {
+			foreach ( $groupList as $group ) {
+				if ( in_array( $group, $this->permissionsConfiguration->getGlobalGroups() ) ) {
+					$result = $this->removeFromGlobalGroup( $userToChange, $group ) && $result;
+				} else {
+					$result = $this->removeFromLocalGroup( $userToChange, $group ) && $result;
+				}
+			}
+		} finally {
```
macbre, line 318: Can safely ignore exceptions caught inside this `try` block?

includes/specials/SpecialUserrights.php
jcellary, line 285: It's being done inside of PermissionsServiceImpl

https://github.com/Wikia/app/blob/dev/lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php#L249

macbre, line 285: :+1:

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
jcellary, line 318: That's a good question - the only problem (at least for now, until we migrate to external service) that I can think of here would be a problem with DB - I'm not sure if there's something more we should do here in such case? I think that in such case the exception should simply propagate higher - that's basically also the way it was working before. I added the finally statement to make sure we don't have some stale data cached locally. Is there something more that you'd suggest?

macbre, line 318: If it's only DB-related exception that can be thrown here, then we're fine. Their constructor reports them to ELK.

macbre:
LGTM

jcellary:
@macbre I improved logging according to your suggestion - I also then noticed that global groups weren't filtered out properly when doing a save, so I refactored that part as well. Let me know if it's ok now (especially the way I use wikia logger)

macbre:
LGTM

PR #10203, opened 2016-04-11
PLATFORM-2090 Update protection level text to include Content Mods

RafalWilinski:
https://wikia-inc.atlassian.net/browse/PLATFORM-2090
ping @macbre


macbre:
LGTM

PR #10204, opened 2016-04-11
ADEN-3192 Fix blocked BTF slots

fraszczakszymon:
Handle ATF slots when they are collapsed and unblock BTF queue.

https://wikia-inc.atlassian.net/browse/ADEN-3192


nandy-andy:
LGTM

PR #10209, opened 2016-04-12
DatabaseBase: log if we're within DB transaction

macbre:
Make debugging issues like [PLATFORM-2118](https://wikia-inc.atlassian.net/browse/PLATFORM-2118) a bit easier by logging `transaction_level` field stating it the query was performed within a transaction or not.

@wladekb / @drozdo


drozdo:
Good idea!

PR #10211, opened 2016-04-12
[SUS-415] Automatically include all i18n files in LocalisationCache rebuild script. 

ArturKlajnerok:
We should include all localisation files (i18n & aliases) in LocalisationCache rebuild process, this will alow us to build cache file that contains all the messages keys and we will be able to remove manual includes like here :  https://github.com/Wikia/config/blob/dev/CommonExtensions.php#L122

https://wikia-inc.atlassian.net/browse/SUS-415

/cc @mixth-sense

### DIFF
--- a/maintenance/rebuildLocalisationCache.php
+++ b/maintenance/rebuildLocalisationCache.php
@@ -157,7 +157,11 @@ class RebuildLocalisationCache extends Maintenance {
 		exit( $exitcode );
 	}
 
-	public function includeAllMessagesFiles() {
+	/**
+	 * Helper function to include all localisation files to localisation cache rebuild process.
+	 * It looks for i18n and alias files from whole code directory and registers them in global array.
+	 */
+	private function includeAllMessagesFiles() {
 		global $wgExtensionMessagesFiles;
 
 		$directory = new RecursiveDirectoryIterator('/usr/wikia/source/app/');
--- a/maintenance/rebuildLocalisationCache.php
+++ b/maintenance/rebuildLocalisationCache.php
@@ -31,6 +31,11 @@
 require_once( dirname( __FILE__ ) . '/Maintenance.php' );
 
 class RebuildLocalisationCache extends Maintenance {
+
+	const I18N_FILE_EXTENSION = ".i18n.php";
+	const PHP_FILE_EXTENSION = ".php";
+	const LOCALISATION_FILE_REGEX = "/^.+\.(i18n|aliases|alias).php$/i";
+
 	public function __construct() {
 		parent::__construct();
 		$this->mDescription = "Rebuild the localisation cache";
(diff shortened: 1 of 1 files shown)


maintenance/rebuildLocalisationCache.php
```diff
@@ -151,6 +158,28 @@ function ( $item ) use ( $firstCodes ) {
 	}
 
 	/**
+	 * Helper function to include all localisation files to localisation cache rebuild process.
+	 * It looks for i18n and alias files from whole code directory and registers them in global array.
+	 */
+	private function includeAllMessagesFiles() {
+		global $wgExtensionMessagesFiles;
+
+		$directory = new RecursiveDirectoryIterator('/usr/wikia/source/app/');
```
ghost: Please, do not hardcode paths. Use `$IP`.

maintenance/rebuildLocalisationCache.php
```diff
@@ -151,6 +158,28 @@ function ( $item ) use ( $firstCodes ) {
...
+		$iterator = new RecursiveIteratorIterator($directory);
+		$files = new RegexIterator($iterator, self::LOCALISATION_FILE_REGEX, RecursiveRegexIterator::GET_MATCH);
+
+		foreach ($files as $filepath) {
+			if (!isset($filepath[0])) continue;
```
ghost: When does it happen?

maintenance/rebuildLocalisationCache.php
```diff
@@ -151,6 +158,28 @@ function ( $item ) use ( $firstCodes ) {
...
+
+			if (strpos($filepath[0], self::I18N_FILE_EXTENSION) !== false ) {
+				$wgExtensionMessagesFiles[basename($filepath[0], self::I18N_FILE_EXTENSION)] = $filepath[0];
```
ghost: How about `$key = basename(...);` and a single `$wgExtensionMessagesFiles[$key]` assignment?

maintenance/rebuildLocalisationCache.php
ArturKlajnerok: Used `$IP` in : a85c80c

maintenance/rebuildLocalisationCache.php
ArturKlajnerok: Good catch, `$key` it is :  5025a76

maintenance/rebuildLocalisationCache.php
ArturKlajnerok: It should not happen, as if file exists and is passed to loop, there will always be `[0]` that contains full path and `[1]` the substring that matches to regex. Changed in : 725ee7e

maintenance/rebuildLocalisationCache.php
```diff
@@ -151,6 +158,28 @@ function ( $item ) use ( $firstCodes ) {
 	}
 
 	/**
+	 * Helper function to include all localisation files to localisation cache rebuild process.
+	 * It looks for i18n and alias files from whole code directory and registers them in global array.
+	 */
+	private function includeAllMessagesFiles() {
+		global $wgExtensionMessagesFiles;
```
ghost: It is perfectly fine to use `global $wgExtensionMessagesFiles, $IP;` which we normally do everywhere else in the codebase.

maintenance/rebuildLocalisationCache.php
```diff
@@ -151,6 +158,28 @@ function ( $item ) use ( $firstCodes ) {
 	}
 
 	/**
+	 * Helper function to include all localisation files to localisation cache rebuild process.
+	 * It looks for i18n and alias files from whole code directory and registers them in global array.
+	 */
+	private function includeAllMessagesFiles() {
```
ghost, line 164: Make sure, the method follows MW coding conventions. Namely, it's all about spaces after ('s and before )'s, as they are present in the above `execute()` methods.

maintenance/rebuildLocalisationCache.php
```diff
@@ -31,6 +31,11 @@
 require_once( dirname( __FILE__ ) . '/Maintenance.php' );
```
ghost: Let's use `__DIR__` here.

ghost:
@ArturKlajnerok - just a few cosmetic changes, otherwise it LGTM. :+1:

maintenance/rebuildLocalisationCache.php
ArturKlajnerok, line 164: As far I've checked there's now space between `(` and `)` if there's no params in method.

maintenance/rebuildLocalisationCache.php
ArturKlajnerok: Changed in : 351ca0b

maintenance/rebuildLocalisationCache.php
ArturKlajnerok: Changed in : 351ca0b

ghost:
Excellent, :+1:

PR #10217, opened 2016-04-13
SOC-2233 use isset rather than is_null

jsutterfield:
Ticket: https://wikia-inc.atlassian.net/browse/SOC-2233

Make sure a key exists before checking if it's null.

### DIFF
--- a/lib/Wikia/src/Service/User/Attributes/UserAttributes.php
+++ b/lib/Wikia/src/Service/User/Attributes/UserAttributes.php
@@ -51,11 +51,11 @@ class UserAttributes {
 	public function getAttribute( $userId, $attributeName, $default = null ) {
 		$attributes = $this->loadAttributes( $userId );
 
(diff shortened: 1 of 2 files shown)


lib/Wikia/src/Service/User/Attributes/UserAttributes.php
```diff
@@ -51,11 +51,11 @@ public function getAttributes( $userId ) {
 	public function getAttribute( $userId, $attributeName, $default = null ) {
 		$attributes = $this->loadAttributes( $userId );
 
-		if ( !is_null( $attributes[$attributeName] ) ) {
+		if ( isset( $attributes[$attributeName] ) ) {
```
jsutterfield, line 54: We need to make sure that the key both exists and is not null, which conveniently enough is exactly what `isset` does: http://php.net/manual/en/function.isset.php

macbre:
:+1:

armonr:
:shipit:

PR #10218, opened 2016-04-13
SOC-2268 don't log all attributeservice exceptions as errors

jsutterfield:
Ticket: https://wikia-inc.atlassian.net/browse/SOC-2268

As part of SOC-2123 I added additional logging to the AttributeKeyValueService.php. All of that logging was set at the error level, which is incorrect. There are some cases where exceptions are thrown which don't indicate an error. Some of these include when MW tries to GET a non-existent attribute, or an unauthenticated user tries to save an attribute.

This PR updates that logging so we only log `PersistenceExceptions` at error, since those indicate an error actually contacting the service, everything else has been downgraded to info.

### DIFF
--- a/lib/Wikia/src/Service/User/Attributes/AttributeKeyValueService.php
+++ b/lib/Wikia/src/Service/User/Attributes/AttributeKeyValueService.php
@@ -109,7 +109,7 @@ class AttributeKeyValueService implements AttributeService {
 		if ( $e instanceof PersistenceException ) {
 			$this->error( $msg, $context );
 		} else {
-			$this->info( $msg, $context );
+			$this->warning( $msg, $context );
 		}
 	}
 
--- a/lib/Wikia/src/Service/User/Attributes/AttributeKeyValueService.php
+++ b/lib/Wikia/src/Service/User/Attributes/AttributeKeyValueService.php
@@ -7,6 +7,7 @@ use Wikia\Logger\Loggable;
 use Wikia\Persistence\User\Attributes\AttributePersistence;
 use Wikia\Util\WikiaProfiler;
 use Wikia\Service\User\Attributes;
+use Wikia\Service\PersistenceException;
 
 class AttributeKeyValueService implements AttributeService {
 
@@ -105,7 +106,7 @@ class AttributeKeyValueService implements AttributeService {
 	 */
 	private function logException( $userId, \Exception $e, $msg ) {
 		$context = [ 'user' => $userId, 'exception' => $e ];
-		if ( is_a( $e, 'Wikia\Service\PersistenceException' ) ) {
+		if ( $e instanceof PersistenceException ) {
 			$this->error( $msg, $context );
 		} else {
 			$this->info( $msg, $context );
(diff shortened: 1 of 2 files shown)


lib/Wikia/src/Service/User/Attributes/AttributeKeyValueService.php
```diff
@@ -87,16 +87,29 @@ public function delete( $userId, Attribute $attribute ) {
...
+	 * which can be thrown, and which this method can expect to receive: PersistenceException, ForbiddenException,
+	 * NotFoundException, and UnauthorizedException. Log PersistenceExceptions at error level, and all others
+	 * at info. PersistenceExceptions indicate a problem was encountered contacting the service, or that we received
+	 * an unexpected HTTP response like a 500. The other exceptions indicate a resources wasn't found, or that the
+	 * request was refused which is not an error.
+	 * @param $userId
+	 * @param \Exception $e
+	 * @param $msg
+	 */
+	private function logException( $userId, \Exception $e, $msg ) {
```
jsutterfield, line 107: These exceptions come from [here](https://github.com/Wikia/app/blob/dev/lib/Wikia/src/Persistence/User/Attributes/AttributePersistenceSwagger.php#L120).

lib/Wikia/src/Service/User/Attributes/AttributeKeyValueService.php
```diff
@@ -87,16 +87,29 @@ public function delete( $userId, Attribute $attribute ) {
...
+		$context = [ 'user' => $userId, 'exception' => $e ];
+		if ( is_a( $e, 'Wikia\Service\PersistenceException' ) ) {
```
macbre: Why not `instanceof`?

jsutterfield: No real reason. I googled "check if object is type of class php" and the `is_a` function is the first thing that came up. It looks like `intanceof` is a better approach though. I'll update.

lib/Wikia/src/Service/User/Attributes/AttributeKeyValueService.php
```diff
@@ -105,7 +106,7 @@ public function delete( $userId, Attribute $attribute ) {
 	 */
 	private function logException( $userId, \Exception $e, $msg ) {
 		$context = [ 'user' => $userId, 'exception' => $e ];
-		if ( is_a( $e, 'Wikia\Service\PersistenceException' ) ) {
+		if ( $e instanceof PersistenceException ) {
```
macbre, line 109: :+1:

garthwebb:
LGTM

lib/Wikia/src/Service/User/Attributes/AttributeKeyValueService.php
```diff
@@ -87,16 +88,29 @@ public function delete( $userId, Attribute $attribute ) {
...
+	 * @param $userId
+	 * @param \Exception $e
+	 * @param $msg
+	 */
+	private function logException( $userId, \Exception $e, $msg ) {
+		$context = [ 'user' => $userId, 'exception' => $e ];
+		if ( $e instanceof PersistenceException ) {
+			$this->error( $msg, $context );
+		} else {
+			$this->info( $msg, $context );
```
garthwebb: Maybe this should be `warn` like you mentioned in the ticket?

garthwebb:
LGTM

PR #10219, opened 2016-04-13
SOC-2185 save/delete requested-closure-date as both a preference and …

jsutterfield:
Ticket: https://wikia-inc.atlassian.net/browse/SOC-2185

In preparation for migrating `requested-closure-date` to the user preference service, we need to start writing/deleting it from the service. This ensure that when we migrate the existing values from the `user_properties` table, new values will be added and deleted from both places so they'll be consistent. Note that we're not writing over `requested-closure`. The plan is to get rid of this redundant data after the migration since we can assume `requested-closure` == `true` if we have a value for  `requested-closure-date`.


kvas-damian:
LGTM

PR #10220, opened 2016-04-13
JPN-355 Filter out Admin and Anon from Community page experiment

perjg:
getRandomAdmins was not filtering out blocked and anon admins. Added extra filtering to make sure that   they're filtered out before randomizing.

### DIFF
--- a/extensions/wikia/CommunityPageExperiment/CommunityPageExperimentHelper.class.php
+++ b/extensions/wikia/CommunityPageExperiment/CommunityPageExperimentHelper.class.php
@@ -51,13 +51,10 @@ class CommunityPageExperimentHelper {
 		}
 
 		// Filter out blocked and anon admins
-		foreach ( $adminIds as $key => $adminId ) {
+		$adminIds = array_filter( $adminIds, function ( $adminId ) {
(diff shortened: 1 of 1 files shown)


perjg:
@Grunny A small patch to your community page experiment extension. Can you have a look?

extensions/wikia/CommunityPageExperiment/CommunityPageExperimentHelper.class.php
```diff
@@ -49,11 +49,25 @@ public function getRandomAdmins( $num = self::DEFAULT_NUM_ADMINS ) {
 		if ( empty( $adminIds ) ) {
 			return [];
 		}
+
+		// Filter out blocked and anon admins
+		foreach ( $adminIds as $key => $adminId ) {
+			$admin = User::newFromId( $adminId );
+
+			if ( $admin->isBlocked() || $admin->isAnon() ) {
+				unset( $adminIds[$key] );
```
Grunny: I think `array_filter` would be a bit cleaner than this loop.

Grunny:
One suggestion, otherwise LGTM for the experiment code.

PR #10230, opened 2016-04-14
JPN-360-2

perjg:
Work done for the community page the last week.


perjg:
@Grunny Can you please have a look at this when you have time.

extensions/wikia/CommunityPage/CommunityPageSpecialController.class.php
```diff
@@ -69,28 +71,48 @@ public function header() {
...
-				CommunityPageSpecialUsersModel::getTopContributors( 50, '1 MONTH' )
+				CommunityPageSpecialUsersModel::getTopContributors( 50 )
 			);
 		// get details for only 5 of the remaining contributors
 		$contributorDetails = $this->getContributorsDetails( array_slice( $contributors, 0, 5 ) );
 
+		$userRank = '-';
+		$editors = count( $contributors );
+
+		if ($editors == 0) {
```
Grunny: Minor: Always prefer strict comparisons (`$editors === 0`) unless you really need it to be not strict, and also should be spacey per conventions.

extensions/wikia/CommunityPage/CommunityPageSpecialController.class.php
```diff
@@ -69,28 +71,48 @@ public function header() {
...
+			$editors = '-';
+		}
+
+		if ( $userContributionCount > 0 ) {
+			$rank = 1;
+
+			foreach ( $contributors as $contributor ) {
+				if ( $contributor['userId'] == $this->wg->user->getId() ) {
```
Grunny, line 93: Would it be worth including the rank in the top contributor data and having the `$contributors` array indexed by user IDs for easier lookups to make this a little cleaner?

extensions/wikia/CommunityPage/templates/topAdmins.mustache
```diff
@@ -10,6 +10,12 @@
 			</div>
 		</li>
 	{{/admins}}
+	{{^admins}}
+		<div class="zero">
+			{{noAdminText}}
+			<a href="{{{noAdminHref}}}">{{noAdminContactText}}</a>
```
Grunny: Why is `noAdminHref` unescaped?

extensions/wikia/CommunityPage/scripts/ext.communityPage.js
```diff
@@ -81,40 +105,78 @@ require([
...
+		if (tabToActivate === activeTab) {
+			return;
+		}
+
+		getModalNavHtml().then(function (navHtml) {
+			var html;
+
+			// Switch highlight to new tab
+			// fixme: Loading indicator should be via a template.
+			html = navHtml + $.msg('communitypage-modal-loading');
```
Grunny: Since you're using `.html` you need to HTML escape `$.msg('communitypage-modal-loading')`. The message `communitypage-modal-loading` also doesn't seem to be defined in the i18n.php file?

Also, no need to separate the var declaration from the assignment here.

extensions/wikia/CommunityPage/CommunityPageSpecialHooks.class.php
```diff
@@ -14,4 +14,63 @@ public static function getHTMLBeforeWikiaPage( &$html ) {
...
+	 * @param $watchthis
+	 * @param $sectionanchor
+	 * @param $flags
+	 * @param $revision
+	 * @param $status
+	 * @param $baseRevId
+	 * @return bool
+	 */
+	public static function onArticleSaveComplete(
+		&$article, &$user, $text, $summary, $minoredit, $watchthis,
```
Grunny: No need for the `&` on `$article` and `$user` as objects are effectively passed by reference in PHP anyway.

extensions/wikia/CommunityPage/CommunityPageSpecialHooks.class.php
```diff
@@ -14,4 +14,63 @@ public static function getHTMLBeforeWikiaPage( &$html ) {
...
+		$sectionanchor, &$flags, $revision, &$status, $baseRevId ) {
```
Grunny: On multiline definitions like this, we switch to Allman-style braces to make the separation between the definition and the body clearer. i.e.

``` php
public static function onArticleSaveComplete(
    $article, $user, $text, $summary, $minoredit, $watchthis,
    $sectionanchor, &$flags, $revision, &$status, $baseRevId
) {
    // Function body
}
```

extensions/wikia/CommunityPage/CommunityPageSpecialHooks.class.php
```diff
@@ -14,4 +14,63 @@ public static function getHTMLBeforeWikiaPage( &$html ) {
...
+		// Early exit for edits that do not affect any cached item
+		if ($user->isAnon()) {
```
Grunny: Should be spacey per conventions.

extensions/wikia/CommunityPage/CommunityPageSpecialController.class.php
perjg, line 93: That sounds like a good idea.
I'd like to investigate that separately, and make a separate change for it if I decide to do it that way.

extensions/wikia/CommunityPage/CommunityPageSpecialController.class.php
perjg: Sorry, forgot running the stylize script for a few of the commits.

perjg:
@Grunny Addressed most of your comments. I left the one about refactoring the ranking. I'd like to look into that separately.

extensions/wikia/CommunityPage/scripts/ext.communityPage.js
```diff
@@ -81,40 +105,76 @@ require([
...
 
+	function switchCommunityModalTab(tabToActivate) {
+		if (tabToActivate === activeTab) {
+			return;
+		}
+
+		getModalNavHtml().then(function (navHtml) {
+			// Switch highlight to new tab
+			// fixme: Loading indicator should be via a template.
+			var html = navHtml + encodeURIComponent($.msg('communitypage-modal-loading'));
```
Grunny: URL encoding with `encodeURIComponent` is not HTML escaping. You should use `mw.html.escape` instead.

perjg:
@Grunny Now using mw.html.escape :)

Grunny:
:+1:

PR #10232, opened 2016-04-14
SER-35 when account is closing remove all tokens from helios

sqreek:
[SER-35](https://wikia-inc.atlassian.net/browse/SER-35)

Add forceLogout in HeliosClient.
Add call to forceLogout in closeAccount.

//cc @Wikia/services-team

### DIFF
--- a/extensions/wikia/EditAccount/SpecialEditAccount_body.php
+++ b/extensions/wikia/EditAccount/SpecialEditAccount_body.php
@@ -18,6 +18,9 @@ if ( !defined( 'MEDIAWIKI' ) ) {
 	exit( 1 );
 }
 
+use Wikia\DependencyInjection\Injector;
+use Wikia\Service\Helios\HeliosClient;
+
 class EditAccount extends SpecialPage {
 	/** @var User */
 	var $mUser = null;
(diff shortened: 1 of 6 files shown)


lib/Wikia/src/Service/Helios/HeliosClientImpl.php
```diff
@@ -140,6 +143,26 @@ public function login( $username, $password )
...
+	 * @internal param $username
+	 * @return null
+	 */
+	public function forceLogout( $userId ) {
+		return $this->request(
+			sprintf( 'users/%s/tokens', $userId ),
+			[ ],
+			[ ],
+			[ 'method' => 'DELETE',
+				'headers' => [
```
harnash: Indentation broke here

lib/Wikia/src/Service/Helios/HeliosClientImpl.php
```diff
@@ -140,6 +143,26 @@ public function login( $username, $password )
...
+					'THE-SCHWARTZ' => $this->schwartzToken
```
harnash: Would be nice to put this Header into some constant (or there is one already?)

harnash:
Few minor comments. Otherwise LGTM.

extensions/wikia/EditAccount/SpecialEditAccount_body.php
```diff
@@ -411,6 +414,10 @@ public static function closeAccount( $user = '', $changeReason = '', &$mStatusMs
 		// if they are connected from facebook, disconnect them
 		self::disconnectFromFacebook( $user );
 
+		/** @var HeliosClient $heliosClient */
+		$heliosClient = Injector::getInjector()->get(HeliosClient::class);
+		$heliosClient->forceLogout($user->getId());
```
harnash: I would think about the case when request fail. What will happen? Whole process will be reverted? Partially done? Will there be a sane message to the user or 5xx?

sqreek: If it fails, it will be notice in kibana.
I asked in ticket for expecting behavior, but at this moment, everything will works as was before this commit with additional call to helios.

PR #10237, opened 2016-04-15
WikiaTracer: fix PHP Strict Standards: Only variables should be passed by reference

macbre:
[PLATFORM-2119](https://wikia-inc.atlassian.net/browse/PLATFORM-2119)

@wladekb


wladekb:
lgtm

PR #10249, opened 2016-04-18
[PLATFORM-2128] Removed Batch User Rights

jcellary:
https://wikia-inc.atlassian.net/browse/PLATFORM-2128

Also made minor refactor to Special:UserRights - made previously shared methods private and fixed one small bug with missing parenthesis in a logical check.

https://github.com/Wikia/config/pull/1737


macbre:
:+1:

PR #10252, opened 2016-04-18
CE-3603 VisualEditor editing tour experiment

kamilkoterba:
https://wikia-inc.atlassian.net/browse/CE-3603
## Description

This is experiment targeted at newly signed up users and returning users without edits to verify how introducing a tour will impact their number of edits.

This code will be removed after about two weeks [CE-3630](https://wikia-inc.atlassian.net/browse/CE-3630)
## Reviewers

@Wikia/spitfires


extensions/wikia/VisualEditorTourExperiment/scripts/VisualEditorTourExperimentInit.js
```diff
@@ -0,0 +1,62 @@
...
+		function init() {
+			if (isEnabled()) {
+				clearEntrypointPopover();
+				(new VETour(veTourConfig)).start();
+			}
+		}
+
+		function isEnabled() {
+			return isExperimentVariation() &&
+				(isNewlyregistered() || isUserwithoutedit) &&
```
Grunny: `isUserwithoutedit` should be a call to the function.

extensions/wikia/VisualEditorTourExperiment/scripts/VisualEditorTourExperiment.js
```diff
@@ -0,0 +1,119 @@
...
+
+		Tour.prototype._setDisabled = function() {
+			if(!$.cookie('vetourdisabled')) {
+				$.cookie('vetourdisabled', 1, {expires: 30});
+			}
+		}
+
+		Tour.prototype._setupTour = function (assets) {
+			$('body').on('click', '.ve-tour-next', this.next.bind(this));
+			$('body').on('click', '.ve-tour-experiment .close', this.close.bind(this));
```
Grunny: Put `$('body')` in a variable since you use it more than once.

extensions/wikia/VisualEditorTourExperiment/scripts/VisualEditorTourExperiment.js
```diff
@@ -0,0 +1,134 @@
...
+		Tour.prototype._setDisabled = function() {
+			if(!$.cookie('vetourdisabled')) {
+				$.cookie('vetourdisabled', 1, {expires: 30});
+			}
+		}
+
+		Tour.prototype._setupTour = function (assets) {
+			$('body').on('click', '.ve-tour-next', this.nextHandle.bind(this));
+			$('body').on('click', '.ve-tour-prev', this.prevHandle.bind(this));
+			$('body').on('click', '.ve-tour-experiment .close', this.close.bind(this));
```
Grunny: Put `$('body')` in a variable since you use it more than once.

extensions/wikia/VisualEditorTourExperiment/scripts/VisualEditorTourExperiment.js
```diff
@@ -0,0 +1,144 @@
...
+
+		Tour.prototype.start = function() {
+			this.step = -1;
+			loader({
+				type: loader.MULTI,
+				resources: {
+					mustache: 'extensions/wikia/VisualEditorTourExperiment/templates/VisualEditorTourExperiment_content.mustache',
+				}
+			}).done(this._setupTour.bind(this));
+		}
```
Grunny: Missing semi-colons on all of these.

extensions/wikia/VisualEditorTourExperiment/scripts/VisualEditorTourExperiment.js
```diff
@@ -0,0 +1,144 @@
...
+		Tour.prototype.close = function() {
+			this._setDisabled();
+			this.destroyStep(this.step);
+			track({
+				action: tracker.ACTIONS.CLICK,
+				label: this.labelPrefix + 'close'
+			});
+			track({
+				action: tracker.ACTIONS.CLICK,
+				label: this.labelPrefix + 'close-' + this.step
```
Grunny: Do we need two tracking events for this or can we just sum the `close-<step>` events?

extensions/VisualEditor/wikia/modules/ve/ve.track.js
```diff
@@ -58,6 +58,9 @@ require( ['wikia.tracker'], function ( tracker ) {
 				};
 			},
 			'mwtiming.performance.user.saveComplete': function ( data ) {
+				require(['VisualEditorTourExperimentInit'], function (veTourInit) {
```
Grunny: What about if this experiment isn't enabled on a given wiki?

And it might be better to make this generic for all our experiments, i.e. track for any of our variations and include which type of user they are (new or returning).

extensions/wikia/VisualEditorTourExperiment/scripts/VisualEditorTourExperimentInit.js
```diff
@@ -0,0 +1,83 @@
...
+		}
+		
+		function isUserwithoutedit() {
+			return $.cookie('userwithoutedit');
+		}
+
+		function overridePopoverZIndex() {
+			if (!popoverZIndexOverriden) {
+				popoverZIndexOverriden = true;
+				$('body').append('<style type="text/css">.popover {z-index: 5000014;}</style>');
```
Grunny: If you need to add CSS with JS, I'd suggest using `mw.util.addCSS( ... );`, but is there a way to avoid this (i.e. being able to set a class on the popovers defined here)?

extensions/wikia/VisualEditorTourExperiment/scripts/VisualEditorTourExperimentInitArticleEntry.js
```diff
@@ -0,0 +1,61 @@
...
+	function ($, mw, veTourInit, tracker) {
+		'use strict';
+
+		var track = tracker.buildTrackingFunction({
+			category: 've-editing-tour',
+			trackingMethod: 'analytics'
+		});
+
+		function initEntry() {
+			var $editButton = $('#ca-ve-edit');
```
Grunny, line 8: Open question on an edge case: is there a chance a returning user without edits has set the source editor in their preferences? Probably unlikely, but worth a thought.

extensions/wikia/VisualEditorTourExperiment/styles/VisualEditorTourExperimentInit.scss
```diff
@@ -0,0 +1,45 @@
...
+
+	.close {
+		@include image-replacement();
+		background: url( $close-image ) center;
+		background-size: 16px 16px;
+		display: block;
+		float: right;
+		height: 16px;
+		outline: none;
+		cursor: pointer;
```
Grunny: Should be alphabetised.

extensions/wikia/VisualEditorTourExperiment/scripts/VisualEditorTourExperimentInitArticleEntry.js
kamilkoterba, line 8: hence check is below mw.config.get('wgVisualEditorPreferred')

Grunny, line 8: Bah, missed that. ;)

Grunny:
:+1:

PR #10258, opened 2016-04-18
RECIRC-96: Be explicit when using `.outerWidth()`

pauloslund:
the version of jQuery UI that we use on some communities breaks the `.outerWidth()` functionality of jQuery objects when called without a parameter, causing it to return an object instead of the actual width. Adding `false` is a quick fix until we decide to upgrade jQuery UI.

Fixed in version 1.8.22 of jQuery UI:
http://stackoverflow.com/a/13805897


gabrys:
LGTM!

PR #10261, opened 2016-04-19
VOLDEV-138: Show proper error message for anonymous users if they cannot comment

mszabo-wikia:
@garthwebb @kvas-damian 
Per Bert Hall's testing, this change ensures that anonymous users are shown proper error messages if they are not allowed to comment.

The major change introduced is that creating a new comment now requires the generic 'edit' permission, along with 'commentcreate', to ensure expected behavior on wikias with anonymous editing disabled. The user, however, does _not_ have to be able to edit the specific page they are commenting on (on e.g. blog articles, this would not be feasible).
https://wikia-inc.atlassian.net/browse/VOLDEV-138
https://wikia-inc.atlassian.net/browse/SOC-1627


garthwebb:
LGTM

PR #10262, opened 2016-04-19
CE-3637: Use static functions

lukaszkonieczny:
https://wikia-inc.atlassian.net/browse/CE-3637

@macbre @SebastianMarzjan


macbre:
Looks jolly good to me

PR #10264, opened 2016-04-19
SEO-380 Special:AllPages and Special:Images no longer allowed in robots.txt

gabrys:
While working on SEO-351 we broke the robots.txt rules for allowing
Special:AllPages (the local sitemap) and Special:Images.

Fixed them now and added the unit tests around the trigger variables.


saipetch:
LGTM

PR #10274, opened 2016-04-20
Make Special:HealthCheck a bit lighter

macbre:
[PLATFORM-2131](https://wikia-inc.atlassian.net/browse/PLATFORM-2131)
- skip the internal POST request
- skip load checks
- leave the `host_disabled` check

@wladekb / @owend / @ljagiello


wladekb:
I remember there was a reason for this POST check, let me find it out before merging...

macbre:
`WebStart.php` has a check for read-only mode - it simply dies when MediaWiki is asked to handle POST request in read-only mode. However, we were not making POST request from HealthCheck when read-only was on.

wladekb:
I couldn't find an exact reason for that so let's proceed with that.

wladekb:
lgtm

owend:
LGTM

PR #10278, opened 2016-04-20
CE-3604 Look for cookie truthness instead of value

kamilkoterba:
Cookies are stored as strings, I think it's better to check for it's pressence instead.

@Wikia/spitfires


Grunny:
:+1:

PR #10280, opened 2016-04-20
CE-3604 User profile fill step by step experiment

kamilkoterba:
https://wikia-inc.atlassian.net/browse/CE-3604

As a freshly signed up user, I want to fill my profile step-by-step providing one information at the time.

Last thing in progress is setting up experiment conditions. Let's start reviewing the code.

@Wikia/spitfires


extensions/wikia/SpitfiresContributionExperiments/templates/MyProfileTourExperimentStep3.mustache
```diff
@@ -0,0 +1,8 @@
+<div class="my-profile-step step-3"></div>
+<a class="my-profile-header-go-back">BACK</a>
+<h3 class="my-profile-welcome-header">Your profile is almost ready!</h3>
+<form class="my-profile-form">
+	<label class="my-profile-label" for="my-profile-gaming-platforms">Tell us a few words about yourself:</label>
+	<textarea class="my-profile-textarea" id="my-profile-gaming-platforms">{{answer}}</textarea>
```
Grunny, line 6: Same ID as the previous one about gaming platforms.

Also, do we need to load four different templates that are basically the same, or can we load one template and have the header and messages be part of the config that is passed to the template?

extensions/wikia/SpitfiresContributionExperiments/Hooks.class.php
```diff
@@ -0,0 +1,66 @@
...
+
+		return true;
+	}
+
+	private function manageUserActivityGroupCookie( \User $user ) {
+		$newlyRegistered = isset( $_COOKIE[ self::NEWLY_REGISTERED_USER ] );
+
+		if ( !$newlyRegistered ) {
+			$userEditCount = $user->getEditCount();
+			$userWithoutEditCookie = empty( $_COOKIE[ self::WITHOUT_EDIT_USER ] );
```
Grunny: `$userWithoutEditCookie` would suggest this would be true if the variable is set, where this is the opposite.

extensions/wikia/SpitfiresContributionExperiments/scripts/my-profile-tour.js
```diff
@@ -0,0 +1,177 @@
...
+				},
+				confirmCloseModal: function () {
+					if (!saved && Object.keys(answers).length) {
+						saveProfile();
+					} else {
+						if (saved) {
+							window.location.href = wgServer + '/wiki/' + userPage;
+						}
+						return true;
+					}
```
Grunny, line 30: I think it would look cleaner with an explicit `return false;` for this function to make it clear it doesn't close the modal.

extensions/wikia/SpitfiresContributionExperiments/scripts/my-profile-tour.js
```diff
@@ -0,0 +1,177 @@
...
+			$profileModal.on('click', '.my-profile-header-go-back', function () {
+				addAnswer();
+				currentStep--;
+				templateData.answer = answers[currentStep];
+				modalContent.html(mustache.render(templates[currentStep], templateData));
+
+				tracker.trackVerboseClick(experimentName, 'go-back-to-step-' + currentStep);
+			});
+
+			$profileModal.on('click', '.my-profile-next-pages a', function () {
```
Grunny: Since this involves navigating away, the preferred event is `.on('mousedown touchstart', ...` to allow the browser a little more time to send the tracker calls through.

extensions/wikia/SpitfiresContributionExperiments/scripts/my-profile-tour.js
```diff
@@ -0,0 +1,177 @@
...
+				uiFactory.init(['modal']).then(createModalComponent);
+			});
+		}
+
+		function createModalComponent(uiModal) {
+			uiModal.createComponent(modalConfig, processInstance);
+		}
+
+		function processInstance(modalInstance) {
+			var modalContent = modalInstance.$element.find('.my-profile-content'),
```
Grunny: This is a jQuery varaible so should be prefixed with `$`.

extensions/wikia/SpitfiresContributionExperiments/templates/MyProfileTourExperimentStep3.mustache
kamilkoterba, line 6: Those templates are loaded interchangeably so ID won't be duplicated in DOM. Besides your proposal is obviously a good idea but given the time and experimental nature and throw away code here I think we can treat this optional.

extensions/wikia/SpitfiresContributionExperiments/scripts/challenge.js
```diff
@@ -0,0 +1,138 @@
...
+], function ($, mw, tracker, cache, loader, mustache) {
+	'use strict';
+
+	var experimentName = 'contribution-experiments',
+		experimentId = 5685550912,
+		dismissCookieName = 'contribchallengedismissed';
+
+	function init() {
+		if (
+			mw.user.anonymous() ||
```
kamilkoterba: @Grunny Isn't it redundant check to what happens in optimizely preconditions?

Grunny: Yep, this code pre-dated the pre-conditions, it doesn't hurt since it's a super cheap check, but we can remove it.

extensions/wikia/SpitfiresContributionExperiments/Hooks.class.php
```diff
@@ -39,7 +39,11 @@ public function onUserLoadFromHeliosToken( \User $user ) {
 	}
 
 	public function onBeforePageDisplay( \OutputPage $out, \Skin $skin ) {
-		if ( isset( $_COOKIE[ self::NEWLY_REGISTERED_USER ] ) || isset( $_COOKIE[ self::WITHOUT_EDIT_USER ] ) ) {
+		$user = \RequestContext::getMain()->getUser();
```
Grunny, line 42: `OutputPage` is a context source, so you can just use `$out->getUser();` (and actually `Skin` is too so `$skin->getUser();` would also work :)).

extensions/wikia/SpitfiresContributionExperiments/Hooks.class.php
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	public function onUserLoadFromHeliosToken( \User $user ) {
+		$this->manageUserActivityGroupCookie( $user );
+
+		return true;
+	}
+
+	public function onBeforePageDisplay( \OutputPage $out, \Skin $skin ) {
+		$user = \RequestContext::getMain()->getUser();
```
Grunny, line 42: `OutputPage` is a context source, so you can just use `$out->getUser();` (and actually `Skin` is too so `$skin->getUser();` would also work :)).

adamkarminski:
CE-3605 (Challenge experiment) tested - LGTM 💾

kamilkoterba:
CE-3604 (User profile completion) tested LGTM 💾

PR #10281, opened 2016-04-20
PLATFORM-2088 Fixing /Wikis/Details API inconsistencies

RafalWilinski:
https://wikia-inc.atlassian.net/browse/PLATFORM-2088

In swagger, "language" is replaced with "lang",
API now returns "domain", "hub", "topic" and "title" computed from other tables.

cc @macbre


includes/wikia/services/WikiService.class.php
```diff
@@ -817,18 +817,29 @@ public function getDetails( Array $wikiIds = null ) {
...
+					'city_list.city_vertical',
 					'city_visualization.city_headline',
 					'city_visualization.city_description',
 					'city_visualization.city_main_image',
 					'city_visualization.city_flags',
+					'city_verticals.vertical_name',
+					'city_domains.city_domain',
+					'city_cat_mapping.cat_id',
+					'city_cat_mapping.city_id',
+					'city_cats.cat_name'
```
macbre, line 837: Do we need to get all these columns? Not all of them seems to be used when fetching results in PHP.

macbre, line 837: Can you run `EXPLAIN` on these query and provide the result here?

includes/wikia/services/WikiService.class.php
```diff
@@ -841,6 +852,22 @@ public function getDetails( Array $wikiIds = null ) {
...
 						'city_list.city_id = city_visualization.city_id'
+					),
+					'city_verticals' => array(
+						'LEFT JOIN',
+						'city_list.city_vertical = city_verticals.vertical_id'
+					),
+					'city_domains' => array(
+						'LEFT JOIN',
+						'city_list.city_id = city_domains.city_id'
+					),
```
macbre: `city_list.city_url` column keeps the "primary" wiki domain. Maybe it will be enough here to just use this column instead of a join.

RafalWilinski:
@macbre Could you please review it? I think I fixed everything.

includes/wikia/services/WikiService.class.php
```diff
@@ -817,18 +817,25 @@ public function getDetails( Array $wikiIds = null ) {
...
-					'city_list'
+					'city_list',
+					'city_verticals',
+					'city_cat_mapping',
+					'city_cats'
 				),
 				array(
 					'city_list.city_id',
 					'city_list.city_title',
+					'city_list.city_dbname',
```
macbre: `city_dbname` does not seem to be used when fetching the results below.

macbre:
Two comments left inline and you're good to go.

includes/wikia/services/WikiService.class.php
```diff
@@ -817,18 +817,25 @@ public function getDetails( Array $wikiIds = null ) {
...
+					'city_list.city_vertical',
 					'city_visualization.city_headline',
 					'city_visualization.city_description',
 					'city_visualization.city_main_image',
 					'city_visualization.city_flags',
+					'city_verticals.vertical_name',
+					'city_cats.cat_name'
 				),
 				array(
 					'city_list.city_public' => 1,
```
macbre, line 840:
``` php
'city_list.city_id IN (' . implode( ',', $wikiIds ) . ')',
```

can be rewritten to prevent SQL injection:

``` php
'city_list.city_id' => $wikiIds
```

RafalWilinski:
🎉  @macbre !

PR #10284, opened 2016-04-20
[PLATFORM-2077] Removed a leftover else statement

jcellary:
https://wikia-inc.atlassian.net/browse/PLATFORM-2077

Removed an else statement that was part of an if statement that has been previously deleted.

pinging @owend

### DIFF
--- a/skins/oasis/modules/PageHeaderController.class.php
+++ b/skins/oasis/modules/PageHeaderController.class.php
@@ -72,7 +72,7 @@ class PageHeaderController extends WikiaController {
 		}
 
 		// "Add topic"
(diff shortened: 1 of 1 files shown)


ghost:
Any retrospective thoughts, how could we avoid that (or similar issues in the future)? Apart from the obvious 100% coverage :-) I mean something feasible. :-)

jcellary:
I don't thing there's anything special - the original commit was removing 77k lines, so not much one can do to manually review :) And just to be clear, I'm really happy we did that :) The only thing that would have saved us is having selenium tests for that feature.

ghost:
Thanks! Go on! :+1: LGTM

PR #10285, opened 2016-04-21
SOC-2290 Disable email sent as a result of logging

garthwebb:
https://wikia-inc.atlassian.net/browse/SOC-2290


armonr:
Will turning off `$wgUseEnotif` stop other expected notifications to be sent out or is it the case that we never want this API to generate any notifications?

kvas-damian:
LGTM

garthwebb:
@armonr, turning off `$wgUseEnotif` will stop all email notifications via this controller which is what we want.

PR #10288, opened 2016-04-21
ADEN-3238 Remove JP slot size restriction

drets:
Let JP to have regular ad sizes in the top leaderboard


nandy-andy:
LGTM :+1: 

It means in JP we'll have [the regural sizes](https://github.com/Wikia/app/blob/ADEN-3238-remove-jp-slot-size-restriction/extensions/wikia/AdEngine/js/provider/directGpt.js#L51).

PR #10293, opened 2016-04-21
RECIRC-106: Catch Http errors in recirculation data services

pauloslund:


extensions/wikia/Recirculation/services/DiscussionsDataService.class.php
```diff
@@ -46,10 +46,17 @@ private function apiRequest() {
 		$endpoint = $wgCityId . '/forums/' . $wgCityId;
 
 		$url = $this->buildUrl( $endpoint, $options );
-		$data = Http::get( $url );
 
-		$obj = json_decode( $data, true );
-		return $obj['_embedded']['doc:threads'];
+		try {
+			$data = Http::get( $url );
```
gabrys, line 51: this doesn't throw on error, but rather return false

extensions/wikia/Recirculation/services/DiscussionsDataService.class.php
```diff
@@ -46,10 +46,17 @@ private function apiRequest() {
...
+
+			$obj = json_decode( $data, true );
```
gabrys, line 53: this doesn't throw on error, but rather return null

extensions/wikia/Recirculation/services/DiscussionsDataService.class.php
```diff
@@ -46,10 +46,17 @@ private function apiRequest() {
...
+			$posts = $obj['_embedded']['doc:threads'];
```
gabrys, line 54: this doesn't throw on error, but returns null too (plus emits a warning)

extensions/wikia/Recirculation/services/FandomDataService.class.php
```diff
@@ -58,10 +58,16 @@ private function apiRequest( $type ) {
 		}
 
 		$url = $this->buildUrl( $endpoint, $options );
-		$data = ExternalHttp::get( $url );
 
-		$obj = json_decode( $data );
-		$posts = $this->dedupePosts( $obj->data );
+		try {
+			$data = ExternalHttp::get( $url );
```
gabrys, line 63: same as above ;-)

extensions/wikia/Recirculation/services/DiscussionsDataService.class.php
pauloslund: Generally yes, however it seems to throw exceptions when the route times out as per: https://wikia-inc.atlassian.net/browse/RECIRC-106. I could most like just wrap the `Http::get` in a try catch block, but seemed to make sense just to put it all in here even if the rest of the methods don't throw exceptions

PR #10294, opened 2016-04-21
RECIRC-107: Check if rawPosts is array before working with it

pauloslund:


extensions/wikia/Recirculation/services/FandomDataService.class.php
```diff
@@ -97,15 +97,17 @@ private function dedupePosts( $rawPosts ) {
...
-		foreach ( $rawPosts as $post ) {
-			if ( count( $posts ) >= self::PARSELY_POSTS_LIMIT ) {
-				break;
-			}
-
-			$metadata = json_decode( $post->metadata );
-			if ( !empty( $metadata->postID ) && !in_array( $metadata->postID, $postIds ) ) {
-				$postIds[] = $metadata->postID;
-				$posts[] = $post;
+		if ( is_array( $rawPosts ) ) {
```
gabrys: My preference would be an early return:

```
if ( !is_array( $rawPosts) ) {
    return [];
}
```

Also I would rather make sure the `$rawPosts` passed are always an array (validate before calling this method), unless there's a reason to do it in here and not before.

extensions/wikia/Recirculation/services/FandomDataService.class.php
```diff
@@ -97,16 +101,14 @@ private function dedupePosts( $rawPosts ) {
 		$posts = [];
 		$postIds = [];
 
-		foreach ( $rawPosts as $post ) {
-			if ( count( $posts ) >= self::PARSELY_POSTS_LIMIT ) {
-				break;
-			}
+		if ( count( $posts ) >= self::PARSELY_POSTS_LIMIT ) {
+			break;
```
gabrys: I don't thing this `break` makes sense in the function scope

gabrys:
LGTM

PR #10295, opened 2016-04-21
SOC-2290 Create a new wg variable to selectively turn off old style emails

garthwebb:
https://wikia-inc.atlassian.net/browse/SOC-2290


jsutterfield:
:shipit:

PR #10299, opened 2016-04-22
SEO-311 Paginator update

gabrys:
This changeset doesn't really change anything. There's a bunch of unit
tests added and a number of small changes like making the unused methods
private, removing unused params and simplifying the way the class is
called in a few places.

The real changes will follow. This just sets a baseline for the test coverage.


saipetch:
LGTM

PR #10303, opened 2016-04-25
Port Database classes to use mysqli extension

macbre:
`mysql` extension is removed in PHP 7 (was deprecated in PHP 5.6).
- backport [MW upstream approach](https://github.com/wikimedia/mediawiki/search?utf8=%E2%9C%93&q=mysqlFieldType) - 718ea2f + 65d18c5
- use `DatabaseMysqli` class for mocking database access (instead of the old `DatabaseMysql`)
- introduce `WikiaBaseTest::getDatabaseMock` helper to avoid hardcoding DB class name in all the tests
- get rid of `DatabaseMysql` class - e8ca2ad

[PLATFORM-2151](https://wikia-inc.atlassian.net/browse/PLATFORM-2151)

@harnash / @wladekb


includes/db/DatabaseMysqli.php
```diff
@@ -305,3 +315,11 @@ public function __toString() {
...
 	}
 }
+
+
+/**
+ * Legacy support: Database == DatabaseMysqli
+ *
+ * @deprecated in 1.16
+ */
+class Database extends DatabaseMysqli {}
```
harnash: I would create ticket for cleaning this up (if there is no one already).

harnash:
LGTM.

includes/db/DatabaseMysqli.php
macbre: Good point - 3ccf849

PR #10305, opened 2016-04-25
CE-3659: Add entry point in article

lukaszkonieczny:


extensions/wikia/PotentialMemberPageExperiments/templates/PMPEntryPoint.mustache
```diff
@@ -1,4 +1,4 @@
-<div class="pmp-entry-point pmp-entry-point-{{bannerType}} clearfix">
+<div class="pmp-entry-point pmp-entry-point-{{bannerType}}">
```
Grunny, line 1: This will now output the second class as `pmp-entry-point-` since `bannerType` isn't set. Though banner type will probably be useful for the third entry point variation.

adamkarminski, line 1: Yup, that was the point - to use one template and just style it differently with this class. :)

extensions/wikia/PotentialMemberPageExperiments/scripts/entry-point-experiment.js
```diff
@@ -31,16 +33,42 @@ require([
...
 
-		$banner = $(mustache.render(resources.mustache[0], templateData));
-		$banner.insertAfter($('.header-container'))
+		$banner = $(mustache.render(template, { bannerType: group.toLowerCase() }))
 			.on('click', '.pmp-entry-point-button', onEntryPointClick)
 			.on('click', '.pmp-entry-point-close', close);
+
+		switch (group) {
+			case 'TOP': addEntryPointTop(); break;
+			case 'IN_ARTICLE': addEntryPointInArticle(); break;
```
Grunny: These should be formatted over multiple lines.

Also, I think it would be cleaner to use a more standard strategy pattern here and set the strategy in the init method that will be then used in a standard way throughout, but since it's an experiment, that's just a nice to have.

extensions/wikia/PotentialMemberPageExperiments/scripts/entry-point-experiment.js
```diff
@@ -3,44 +3,87 @@ require([
...
+					resources: {
+						mustache: '/extensions/wikia/PotentialMemberPageExperiments/templates/PMPEntryPoint.mustache',
+						styles: '/extensions/wikia/PotentialMemberPageExperiments/styles/entry-point-experiment.scss'
+					}
+				})
+			).done(addEntryPoint);
+		}
+	}
+
+	function setExperiment() {
```
adamkarminski: I'd name it `shouldSetExperiment` since it returns a boolean.

extensions/wikia/PotentialMemberPageExperiments/scripts/entry-point-experiment.js
```diff
@@ -3,44 +3,87 @@ require([
...
+
+		if (experiments.hasOwnProperty(group)) {
+			experiment = new experiments[group]();
+			return true;
+		}
+
+		return false;
 	}
 
 	function addEntryPoint(resources) {
```
adamkarminski: And I'd rename this function to `setupExperiment` not to confuse this method with an experiment's method `addEntryPoint` and move the `experiment = new experiments[group]()` line here.

adamkarminski:
Two suggestions and it will LGTM :)

extensions/wikia/PotentialMemberPageExperiments/scripts/entry-point-experiment.js
```diff
@@ -3,44 +3,87 @@ require([
...
+			TOP: function () {
+				this.type = 'top';
+				this.addEntryPoint = function () {
+					$banner.insertAfter($('#WikiaPageHeader .header-container'));
+				}
+			},
+			IN_ARTICLE: function () {
+				this.type = 'in-article';
+				this.addEntryPoint = function () {
+					var $content = $('.mw-content-text'),
```
Grunny: You can actually avoid using this since there's always an existing reference to it in `mw.util.$content` so don't need to select it again yourself.

PR #10306, opened 2016-04-25
CE-3666 Add tracking of banner impressions

adamkarminski:
@Wikia/spitfires 

Tracking impression only if 100% of the banner is in viewport.

### DIFF
--- a/extensions/wikia/PotentialMemberPageExperiments/scripts/entry-point-experiment.js
+++ b/extensions/wikia/PotentialMemberPageExperiments/scripts/entry-point-experiment.js
@@ -53,7 +53,9 @@ require([
 		if (isEntryPointInViewport()) {
 			trackBannerImpression();
 		} else {
-			$(w).on('scroll', trackEntryPointInViewport);
+			$(w).on('scroll.trackPMPEntryPoint', $.debounce(200, function() {
+				trackEntryPointInViewport();
+			}));
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


extensions/wikia/PotentialMemberPageExperiments/scripts/entry-point-experiment.js
```diff
@@ -41,14 +46,23 @@ require([
...
 			.on('click', '.pmp-entry-point-button', onEntryPointClick)
 			.on('click', '.pmp-entry-point-close', close);
+
+		bannerOffset = $banner.offset().top;
+		bannerHeight = $banner.outerHeight();
+
+		if (isEntryPointInViewport()) {
+			trackBannerImpression();
+		} else {
+			$(w).on('scroll', trackEntryPointInViewport);
```
lukaszkonieczny: What do you think about adding debounce function?

lukaszkonieczny:
lgtm

extensions/wikia/PotentialMemberPageExperiments/scripts/entry-point-experiment.js
```diff
@@ -57,15 +71,35 @@ require([
...
+	function trackEntryPointInViewport() {
+		if (isEntryPointInViewport()) {
+			trackBannerImpression();
+			$(w).off('scroll.trackPMPEntryPoint');
+		}
+	}
+
+	function trackBannerImpression() {
+		track({
+			label: '',
```
Grunny, line 91: Why is the label empty?

adamkarminski, line 91: I'm working on tracking viewability of the banner. It's just a placeholder and if I don't use it in the end I'll remove it in the upcoming PR. What do you think? :)

Grunny, line 91: Sure.

PR #10309, opened 2016-04-25
CE-3652: Add experiment for users without edit

lukaszkonieczny:


extensions/wikia/SpitfiresContributionExperiments/scripts/experiments-tracker.js
```diff
@@ -55,8 +56,16 @@ define('ext.wikia.spitfires.experiments.tracker', [
 		}
 	}
 
+	function getVariationName() {
+		if (window.optimizely.activeExperiments.indexOf(freshlyRegisteredExperimentId) !== -1) {
+			return window.optimizely.variationNamesMap[freshlyRegisteredExperimentId].toLowerCase()
+		} else {
+			return window.optimizely.variationNamesMap[usersWithoutEditExperimentId].toLowerCase()
```
Grunny: Missing semi-colons on these two returns. Also, super minor, but I think it looks cleaner to not use `else` if the `if` block has a return in it.

Grunny:
LGTM.

PR #10310, opened 2016-04-25
CE-3652: Add experiment for users without edit

lukaszkonieczny:


lukaszkonieczny:
CR done in https://github.com/Wikia/app/pull/10309

PR #10311, opened 2016-04-26
(CE-3688) Fix checks on whether Optimizely experiment is active

Grunny:
Make sure Optimizely is available and fix the conditions so that the
challenge and profile tour experiments will actually load.

/cc @Wikia/spitfires


lukaszkonieczny:
LGTM

PR #10313, opened 2016-04-26
(CE-3688) Fix checks on whether Optimizely experiment is active

lukaszkonieczny:
Make sure Optimizely is available and fix the conditions so that the
challenge and profile tour experiments will actually load.


lukaszkonieczny:
CR done in https://github.com/Wikia/app/pull/10311

PR #10314, opened 2016-04-26
Use PHP5-style constructors

macbre:
[PLATFORM-2163](https://wikia-inc.atlassian.net/browse/PLATFORM-2163) / [PHP 4 style constructors](http://php.net/manual/en/migration70.deprecated.php):

> PHP 4 style constructors (methods that have the same name as the class they are defined in) are deprecated, and will be removed in the future. PHP 7 will emit E_DEPRECATED if a PHP 4 constructor is the only constructor defined within a class. Classes that implement a __construct() method are unaffected.

https://github.com/Alexia/php7mar/pull/18

Additionally:
- remove deprecated and not used `FormatExif` class
- get rid of `extensions/wikia/RTE/ckeditor/ckeditor_php4.php` (keep and use PHP5 version only)

@harnash


harnash:
LGTM

PR #10323, opened 2016-04-27
WikiaRobotsTest: require classes via autoload entries in setup file

macbre:
Avoid race condition when the `RobotsTxtMock` class is required by the unit test class before the extension's setup file is included.

@gabrys / @harnash


gabrys:
Thanks guys @macbre @harnash

PR #10328, opened 2016-04-27
Install phpflickr/wikia composer

macbre:
[PLATFORM-2152](https://wikia-inc.atlassian.net/browse/PLATFORM-2152)

Use new [composer package](https://packagist.org/packages/wikia/phpflickr) instead of the old code from `lib/vendor`.

@harnash


harnash:
Seems legit. LGTM!

PR #10336, opened 2016-04-27
SEO-218 Update Not_a_valid_Wikia page

gabrys:
Building a custom handler for the article.

We'll enable NotAValidWikia extension on community and the code will
take it from there.

The content of the page remains a message. I'm using another label
though to allow easy transition between old and new. The old version
uses 'not_a_valid_Wikia' message, the new uses 'not-a-valid-wikia'
message (dashes and all lowercase).

There's no need for translations as of now, because you're always taken
to the English community anyway (we may want to support different
languages based on a guess from the domain name or user browser setting).

The ?from param is used to construct the query for the search page.
We're removing the last part from the domain plus optionally the second
last part if it's one of org, com, co, and similar ones. We're dropping
the leading "www" too.

The page will need to be removed from community, so the correct (404)
HTTP status is returned.

### DIFF
--- a/extensions/wikia/NotAValidWikia/NotAValidWikia.i18n.php
+++ b/extensions/wikia/NotAValidWikia/NotAValidWikia.i18n.php
@@ -1,0 +1,26 @@
+<?php
+/**
+* Internationalisation file for the NotAValidWikia extension.
+*
+* @addtogroup Languages
+*/
+
+$messages = array();
+
+$messages['en'] = array(
+	'not-a-valid-wikia' => ' 
+<div class="center" style="padding: 1em 0; font-weight:bold; font-size:120%; margin-bottom:1em">This is not the wiki you\'re looking for! We didn\'t recognize the URL you provided. Head over to search and find the wiki of your dreams [{{fullurl:homepage:Special:Search|search={{urlencode:$1}}}} here].</div>
+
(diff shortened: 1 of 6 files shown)


extensions/wikia/NotAValidWikia/NotAValidWikia.i18n.php
```diff
@@ -0,0 +1,26 @@
+<?php
+/**
+* Internationalisation file for the NotAValidWikia extension.
+*
+* @addtogroup Languages
+*/
+
+$messages = array();
+
+$messages['en'] = array(
```
saipetch: consider adding $messages['qqq'] ?

extensions/wikia/NotAValidWikia/NotAValidWikiaArticle.class.php
```diff
@@ -0,0 +1,29 @@
...
+		global $wgRequest, $wgOut;
+
+		$wgOut->setPageTitle( $this->getTitle()->getText() );
+
+		// Construct the search query from the from param
+		// (which is the domain we're redirected from)
+		$fromDomain = $wgRequest->getVal( 'from' );
+
+		// Extract the interesting part from the domain
+		$interestingPart = preg_replace(
```
saipetch: consider using preg_match() ?

saipetch:
LGTM

PR #10340, opened 2016-04-28
lib/vendor: get rid of PHP4-style constructors

macbre:
And remove `lib/vendor/DB.php` (leftover after `lib/vendor/DB` removal).

@harnash


harnash:
LGTM

PR #10345, opened 2016-04-28
WikiaSpecialVersion: fix " Non-static method should not be called statically

macbre:
```
Deprecated: Non-static method WikiaSpecialVersion::getGitBranch() should not be called statically in /usr/wikia/source/app/extensions/wikia/SpecialVersion/WikiaSpecialVersion.class.php on line 34
```

@harnash


harnash:
PR title is a bit misleading but LGTM! :)

macbre:
Yep, the error message is a bit confusing :)

PR #10346, opened 2016-04-28
SUS-490 | Adjusting CNW background gradient

bkoval:
https://wikia-inc.atlassian.net/browse/SUS-490

Gradient got an OK from a UX person and a ticket reporter as a UX improvement.


harnash:
LGTM

PR #10350, opened 2016-04-28
Removed ApiGate. We shifted business goals, and since API keys aren't…

SeanColombo:
… required, not many users were using it, no staff were checking the stats that it kept, so it wasn't too terribly useful anymore.  The PR for config changes should be accepted before the PR that will remove this code.


macbre:
LGTM

PR #10352, opened 2016-04-28
XW-1387 | remove documentTitle, make displayTitle consistent for category pages

tomasznapieralski:
https://wikia-inc.atlassian.net/browse/XW-1387

//cc @Wikia/x-wing

### DIFF
--- a/extensions/wikia/MercuryApi/MercuryApiController.class.php
+++ b/extensions/wikia/MercuryApi/MercuryApiController.class.php
@@ -333,15 +333,11 @@ class MercuryApiController extends WikiaController {
 	 * @return void
 	 */
 	public function getPage() {
+		global $wgContLang;
+
 		try {
 			$title = $this->getTitleFromRequest();
 			$data = [ ];
-
-			/**
-			 * On article Main Pages and on Curated Main Pages, page title and Wiki name are equal
-			 * and to avoid duplicates we don't use page title
-			 */
-			$documentTitle = '';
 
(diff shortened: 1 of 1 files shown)


extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -370,16 +366,21 @@ public function getPage() {
...
 						if ( MercuryApiCategoryHandler::hasArticle( $this->request, $article ) ) {
-							$data['article'] = MercuryApiArticleHandler::getArticleJson( $this->request, $article );
 							$data['details'] = MercuryApiArticleHandler::getArticleDetails( $article );
+							$data['article'] = MercuryApiArticleHandler::getArticleJson( $this->request, $article );
+
+							// Remove namespace prefix from displayTitle, so it can be consistent with title
+							$namespacePrefix = $wgContLang->getNamespaces()[$data['ns']];
+
+							if ( strpos( $data['article']['displayTitle'], $namespacePrefix ) === 0 ) {
+								$data['article']['displayTitle'] = substr( $data['article']['displayTitle'], strlen( $namespacePrefix ) + 1 );
```
kamilkoterba: Can't we provide title from API without NS prefix to avoid removal? Ideally it should be calculated in parser where prefixed display title is created.

kamilkoterba:
LGTM

tomasznapieralski:
tested, works with old and new app version

PR #10359, opened 2016-04-29
CE-3697 Prepare a Japanese version of the experiment

adamkarminski:
### Links
- https://wikia-inc.atlassian.net/browse/CE-3697
### Description

This PR prepares the Potential Member Page Entry Points experiment to be run in different languages. A new experiment for Japanese has been created.

It's ready to be run on one GA slot by including a language code in the tracking category: `potential-member-page-experiment-{langCode}`.
### Reviewers

@Wikia/spitfires

### DIFF
--- a/extensions/wikia/PotentialMemberPageExperiments/PotentialMemberPageExperiments.i18n.php
+++ b/extensions/wikia/PotentialMemberPageExperiments/PotentialMemberPageExperiments.i18n.php
@@ -24,7 +24,7 @@ $messages['qqq'] = [
  * Japanese - ja
  */
 $messages['ja'] = [
-	'pmp-entry-point-message' => 'このコミュニティはあなたのようなファンによって創られています。',
+	'pmp-entry-point-message' => 'このコミュニティは<strong>あなたのようなファンによって創られています</strong>。',
 	'pmp-entry-point-button' => 'コミュニティを助ける',
 	'pmp-entry-point-button-url' => 'http://ja.community.wikia.com/wiki/%E5%88%9D%E3%82%81%E3%81%A6%E3%81%AE%E6%8A%95%E7%A8%BF%E3%82%92%E3%81%99%E3%82%8B',
 ];
--- a/extensions/wikia/PotentialMemberPageExperiments/PotentialMemberPageExperiments.i18n.php
+++ b/extensions/wikia/PotentialMemberPageExperiments/PotentialMemberPageExperiments.i18n.php
@@ -1,0 +1,30 @@
+<?php
+
+$messages = [];
+
+/**
+ * English - en
+ */
+$messages['en'] = [
+	'pmp-entry-point-message' => 'This wiki is <strong>built by</strong> fans like <strong>you</strong>',
+	'pmp-entry-point-button' => 'Help out',
+	'pmp-entry-point-button-url' => 'http://community.wikia.com/wiki/Tips_on_Getting_Started',
+];
+
+/**
+ * Documentation - qqq
+ */
+$messages['qqq'] = [
+	'pmp-entry-point-message' => 'A message shown next to an entry point to the Potential Member Page.',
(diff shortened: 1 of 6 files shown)


extensions/wikia/PotentialMemberPageExperiments/scripts/entry-point-experiment.js
```diff
@@ -74,15 +87,26 @@ require([
...
+		var experiment = experiments[experimentGroup],
+			templateData;
 
 		loader.processStyle(resources.styles);
+		mw.messages.set(resources.messages);
+
+		templateData = {
+			bannerType: experiment.type,
+			language: contentLanguage,
+			messageText: mw.message('pmp-entry-point-message').plain(),
```
adamkarminski, line 103: @Grunny - I wasn't sure about the usage of `plain()`and `{{{messageText}}}` here. Is there a better way to deal with messages including HTML tags? Thanks

extensions/wikia/PotentialMemberPageExperiments/PotentialMemberPageExperiments.i18n.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+$messages = [];
+
+/**
+ * English - en
+ */
+$messages['en'] = [
+	'pmp-entry-point-message' => 'This wiki is <strong>built by</strong> fans like <strong>you</strong>',
```
kamilkoterba, line 9: Can we use `This wiki is '''built by''' fans like '''you'''` and parse?

kamilkoterba:
Except the comment LGTM

extensions/wikia/PotentialMemberPageExperiments/scripts/entry-point-experiment.js
Grunny, line 103: In general, messages should not contain HTML (both because of security and because translators shouldn't be required to produce valid HTML). In a case like `<strong>` you'd use wikitext `'''` and parse (and `<strong>` is valid wikitext as well), but parsing doesn't work in JS, so you'd have to retrieve the parsed messages from the server side. Since it's an experiment though, we can make a temporary exception.

extensions/wikia/PotentialMemberPageExperiments/scripts/entry-point-experiment.js
```diff
@@ -74,15 +87,26 @@ require([
 	}
 
 	function experimentEnabled() {
+		experimentGroup = abTest.getGroup(experimentNames[contentLanguage]);
```
Grunny: Isn't this already set in the `init` method?

extensions/wikia/PotentialMemberPageExperiments/scripts/entry-point-experiment.js
```diff
@@ -74,15 +87,26 @@ require([
...
+			templateData;
 
 		loader.processStyle(resources.styles);
+		mw.messages.set(resources.messages);
+
+		templateData = {
+			bannerType: experiment.type,
+			language: contentLanguage,
+			messageText: mw.message('pmp-entry-point-message').plain(),
+			buttonText: mw.message('pmp-entry-point-button').escaped(),
```
Grunny: `buttonText` is escaped in the template with `{{buttonText}}` so this will be double escaped. Can just be `.plain()`.

extensions/wikia/PotentialMemberPageExperiments/scripts/entry-point-experiment.js
adamkarminski: Yes, I missed this leftover, thanks!

extensions/wikia/PotentialMemberPageExperiments/scripts/entry-point-experiment.js
adamkarminski: Corrected

Grunny:
Testing LGTM.

PR #10373, opened 2016-05-03
RECIRC-18: Filter Fandom by relevant vertical

pauloslund:
Currently using hard coded tag map arrays for the experiment. If the variation performs well we will want to update this to something more manageable.


gabrys:
LGTM

PR #10379, opened 2016-05-04
Get rid of handlebars

macbre:
[PLATFORM-2155](https://wikia-inc.atlassian.net/browse/PLATFORM-2155)

"As we're here" kind of pull request. php7mar detected the use of PHP7 reserved keyword in Handlebars library.

``` md
#### /usr/wikia/source/app/lib/vendor/Handlebars/String.php
* reservedNames
 * Line 29: `class String extends BaseString`
```

After some grepping it turned out that this library is not used at all and is simply a leftover from Venus times. Can be safely removed. [API help pages](http://www.wikia.com/api/v1) use Handlebars (however client-side rendering is used there).

@bognix / @kvas-damian


bognix:
LGTM! Go for it!

kvas-damian:
LGTM :shipit:

PR #10380, opened 2016-05-04
CE-3697 Force linebreak for JA translation

kamilkoterba:
Per discussion in https://wikia-inc.atlassian.net/browse/CE-3697 adding linebrak

@Wikia/spitfires


lukaszkonieczny:
lgtm

PR #10384, opened 2016-05-05
JPN-416 Show admin status and contributors correctly in modal

perjg:
The "top contributors" tab was not implemented according to the design.

The following two issues were fixed (by adding a separate template for the modal tab from the one used on the main page contributor module):
- Show admin status beneath the name (on main page # of contributions is shown instead)
- Show # of contributions right justified by the name


perjg:
@kamilkoterba PR for JPN-416.

kamilkoterba:
LGTM

kamilkoterba:
Tested LGTM

PR #10387, opened 2016-05-05
ADEN-3257 SP for logged users

jacekjursza:
Disable AdBlock content control for logged in users

### DIFF
--- a/extensions/wikia/ARecoveryEngine/templates/ARecoveryEngineApiController_getBootstrap.mustache
+++ b/extensions/wikia/ARecoveryEngine/templates/ARecoveryEngineApiController_getBootstrap.mustache
@@ -4,11 +4,16 @@ window._sp_ = {
     config: {
         enable_rid: true,
         content_control_callback: function() {
-            var msg = '<h2>Ad blocker interference detected!</h2>';
+            var logGroup = 'ext.wikia.aRecoveryEngine.recovery.bootstrap',
+                msg = '<h2>Ad blocker interference detected!</h2>';
                 msg += '<h3>If you have added a new rule to your ad blocker that interferes with the loading of ads on your pages, ';
                 msg += 'this Interference Screen is the expected behaviour. <br />';
                 msg += 'Removing the rule(s) should load the page as expected.</h3>';
-            document.getElementById('WikiaArticle').innerHTML = msg;
+            if (!window.wgUserName) {
+                document.getElementById('WikiaArticle').innerHTML = msg;
+            } else {
+                window.Wikia.log(['isContentControlEnabled', false], 'debug', logGroup);
+            }
         }
     }
 };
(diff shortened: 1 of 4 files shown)


extensions/wikia/ARecoveryEngine/templates/ARecoveryEngineApiController_getBootstrap.mustache
```diff
@@ -4,11 +4,16 @@ window._sp_ = {
     config: {
         enable_rid: true,
         content_control_callback: function() {
-            var msg = '<h2>Ad blocker interference detected!</h2>';
+            var logGroup = 'ext.wikia.aRecoveryEngine.recovery.bootstrap',
+                msg = '<h2>Ad blocker interference detected!</h2>';
```
drets: What about making the same message [here](https://github.com/Wikia/app/blob/155c0f83b95b1e26c7e21796ce1fe6af347087ba/extensions/wikia/ARecoveryEngine/css/recoveryLock.scss#L7)?

extensions/wikia/ARecoveryEngine/templates/ARecoveryEngineApiController_getBootstrap.mustache
```diff
@@ -4,11 +4,16 @@ window._sp_ = {
...
                 msg += '<h3>If you have added a new rule to your ad blocker that interferes with the loading of ads on your pages, ';
                 msg += 'this Interference Screen is the expected behaviour. <br />';
                 msg += 'Removing the rule(s) should load the page as expected.</h3>';
```
drets, line 11: What about making the same message [here](https://github.com/Wikia/app/blob/155c0f83b95b1e26c7e21796ce1fe6af347087ba/extensions/wikia/ARecoveryEngine/css/recoveryLock.scss#L7) as part of this PR?

jacekjursza, line 11: I agree that there should be one message, but. There is an issue with that. Basically, we can not generate HTML tags trough CSS. IMO we should wait till we get a real message from ComDev and then think how to implement it in both places. I wouldn't touch it now to not duplicate work.

fraszczakszymon:
Please remove lock css for logged in users as well:
https://github.com/Wikia/app/blob/dev/extensions/wikia/ARecoveryEngine/ARecoveryEngineHooks.class.php#L23

fraszczakszymon:
One more thing, we are recovering only TOP_LEADERBOARD and TOP_RIGHT_BOXAD. Should we add HOME_ / CORP_ slots?
https://github.com/Wikia/app/blob/dev/extensions/wikia/AdEngine/js/provider/directGpt.js#L57

fraszczakszymon:
LGTM 👍

PR #10389, opened 2016-05-05
JPN-432 Fix issue where modal tab could get stuck in 'loading' state

perjg:
Deleted redundant leftover code from before modal loading refactor.

This caused a bug where the tab would be stuck in loading state when invoking the modal a second time.


kamilkoterba:
LGTM

PR #10391, opened 2016-05-05
XW-1368 | Remove /api/mercury from purged URLs

rogatty:
https://wikia-inc.atlassian.net/browse/XW-1368

Since https://github.com/Wikia/mercury/pull/2363 we no longer call `/api/mercury` URLs, instead we use MediaWiki's `/wikia.php` directly. No need to purge those old URLs.

/cc @Wikia/x-wing

### DIFF
--- a/extensions/wikia/MercuryApi/MercuryApiHooks.class.php
+++ b/extensions/wikia/MercuryApi/MercuryApiHooks.class.php
@@ -2,11 +2,6 @@
 
 class MercuryApiHooks {
 
-	const SERVICE_API_ROOT = '/';
-	const SERVICE_API_BASE = 'api/mercury/';
-	const SERVICE_API_ARTICLE = 'article/';
-	const SERVICE_API_CURATED_CONTENT = 'main/section/';
-
 	/**
 	 * @desc Get number of user's contribution from DB
 	 *
@@ -87,20 +82,8 @@ class MercuryApiHooks {
 	 * @return bool
 	 */
 	static public function onTitleGetSquidURLs( Title $title, Array &$urls ) {
-		global $wgServer;
-
 		if ( $title->inNamespaces( NS_MAIN ) ) {
-			// Mercury API call from Ember.js to Hapi.js e.g.
-			// http://elderscrolls.wikia.com/api/mercury/article/Morrowind
-			$urls[] =
-				$wgServer .
-				self::SERVICE_API_ROOT .
-				self::SERVICE_API_BASE .
-				self::SERVICE_API_ARTICLE .
-				$title->getPartialURL();
-
-			// Mercury API call from Hapi.js to MediaWiki e.g.
-			// http://elderscrolls.wikia.com/wikia.php?controller=MercuryApi&method=getArticle&title=Morrowind
+			// Request from browser to MediaWiki
 			// TODO: Remove one of these two below when it is decided if we do the switch to getPage() or drop it.
 			$urls[] = MercuryApiController::getUrl( 'getArticle', [ 'title' => $title->getPartialURL() ] );
 			$urls[] = MercuryApiController::getUrl( 'getPage', [ 'title' => $title->getPartialURL() ] );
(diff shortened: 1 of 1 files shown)


extensions/wikia/MercuryApi/MercuryApiHooks.class.php
```diff
@@ -87,20 +82,8 @@ public static function onArticleRollbackComplete( WikiPage $wikiPage, User $user
...
-				self::SERVICE_API_ROOT .
-				self::SERVICE_API_BASE .
-				self::SERVICE_API_ARTICLE .
-				$title->getPartialURL();
-
-			// Mercury API call from Hapi.js to MediaWiki e.g.
-			// http://elderscrolls.wikia.com/wikia.php?controller=MercuryApi&method=getArticle&title=Morrowind
+			// Request from browser to MediaWiki
 			// TODO: Remove one of these two below when it is decided if we do the switch to getPage() or drop it.
 			$urls[] = MercuryApiController::getUrl( 'getArticle', [ 'title' => $title->getPartialURL() ] );
```
Warkot, line 88: What do you think about throwing away this one, too?

extensions/wikia/MercuryApi/MercuryApiHooks.class.php
```diff
@@ -110,17 +110,17 @@ static public function onCuratedContentSave( $sections ) {
...
-			$sectionTitle = $section['title'];
-
-			WikiaDataAccess::cachePurge( MercuryApiMainPageHandler::curatedContentDataMemcKey( $sectionTitle ) );
+			WikiaDataAccess::cachePurge( MercuryApiMainPageHandler::curatedContentDataMemcKey( $sectionLabel ) );
 
 			// Request from browser to MediaWiki
-			$encodedTitle = self::encodeURIQueryParam( $sectionTitle );
-			$urls[] = MercuryApiController::getUrl( 'getCuratedContentSection', [ 'section' => $encodedTitle ] );
+			$encodedTitle = self::encodeURIQueryParam( $sectionLabel );
+			$urls[] = MercuryApiController::getUrl( 'getCuratedContentSection' ) . '&section=' . $encodedTitle;
```
rogatty: `getUrl` method does the additional encoding so I used the same solution as is in https://github.com/Wikia/app/blob/dev/extensions/wikia/CuratedContent/CuratedContentHooks.class.php#L30

extensions/wikia/MercuryApi/MercuryApiHooks.class.php
```diff
@@ -120,44 +103,24 @@ static public function onInstantGlobalsGetVariables( array &$vars ) {
...
-				self::SERVICE_API_CURATED_CONTENT .
-				$doubleEncodedTitle;
+			WikiaDataAccess::cachePurge( MercuryApiMainPageHandler::curatedContentDataMemcKey( $sectionLabel ) );
 
-			// Request from Hapi to MediaWiki
-			$encodedTitle = self::encodeURIQueryParam( $sectionTitle );
-			$urls[] = MercuryApiController::getUrl( 'getCuratedContentSection', [ 'section' => $encodedTitle ] );
+			// Request from browser to MediaWiki
+			$encodedTitle = self::encodeURIQueryParam( $sectionLabel );
+			$urls[] = MercuryApiController::getUrl( 'getCuratedContentSection' ) . '&section=' . $encodedTitle;
```
rogatty, line 123: `getUrl` method does the additional encoding so I used the same solution as is in https://github.com/Wikia/app/blob/dev/extensions/wikia/CuratedContent/CuratedContentHooks.class.php#L30

extensions/wikia/MercuryApi/MercuryApiHooks.class.php
rogatty, line 88: We would have to fix https://github.com/Wikia/api-tests/blob/master/apitests/Mercury/test_mercury_api.py#L13 and possibly some other places too. Let's not do that now.

vforge:
LGTM

PR #10393, opened 2016-05-05
ADEN-3289 apid per wiki

nandy-andy:
Nielsen is now enabled on lego.wikia.com. We want to give possibility to add it to other wikis by setting `apid` in WikiFactory.


fraszczakszymon:
LGTM 👍

PR #10401, opened 2016-05-06
Fix missing event parameter

lukaszkonieczny:
- small refactoring

### DIFF
--- a/extensions/wikia/CommunityPage/scripts/ext.communityPage.js
+++ b/extensions/wikia/CommunityPage/scripts/ext.communityPage.js
@@ -4,8 +4,9 @@ require([
 	'wikia.mustache',
 	'communitypage.templates.mustache',
 	'wikia.nirvana',
-	'wikia.throbber'
-], function ($, uiFactory, mustache, templates, nirvana, throbber) {
+	'wikia.throbber',
+	'wikia.window'
+], function ($, uiFactory, mustache, templates, nirvana, throbber, window) {
 	'use strict';
 
 	// "private" vars - don't access directly. Use getUiModalInstance().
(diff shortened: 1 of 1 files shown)


extensions/wikia/CommunityPage/scripts/ext.communityPage.js
```diff
@@ -4,8 +4,9 @@ require([
 	'wikia.mustache',
 	'communitypage.templates.mustache',
 	'wikia.nirvana',
-	'wikia.throbber'
-], function ($, uiFactory, mustache, templates, nirvana, throbber) {
+	'wikia.throbber',
+	'wikia.window'
+], function ($, uiFactory, mustache, templates, nirvana, throbber, window) {
```
adamkarminski, line 9: I am not sure if you show overwrite the `window` var with `wikia.window`. We usually use a different name like `w`.

extensions/wikia/CommunityPage/scripts/ext.communityPage.js
```diff
@@ -165,20 +166,19 @@ require([
...
-		event.preventDefault();
-	});
-
-	$(document).on( 'click', '#modalTabLeaderboard', function () {
-		switchCommunityModalTab(tabs.TAB_LEADERBOARD);
-		event.preventDefault();
-	});
+	$(document)
+		.on( 'click', '#modalTabAll', function (event) {
+			switchCommunityModalTab(tabs.TAB_ALL);
```
adamkarminski, line 171: @kamilkoterba and @perjg moved the `event.preventDefault();` line above. You could move it as well.

adamkarminski:
After discussion - looks ok to me ;)

PR #10402, opened 2016-05-06
JPN-401 Include the local wiki nav on the Community Page and adjust UI

adamkarminski:
To include the local nav I also had to adjust the UI of the page to the new designs. Also, the monthly stats have been removed as we're going with a welcome message editable by admins instead.


extensions/wikia/CommunityPage/templates/pageHeader.mustache
```diff
@@ -0,0 +1,13 @@
+<div id="CommunityPageHeader" class="community-page-header">
```
lukaszkonieczny: Please remove id, while there are not used

lukaszkonieczny:
LGTM

kamilkoterba:
Tested, LGTM

PR #10404, opened 2016-05-06
SwiftSync: ignore id when moving rows from image_sync to image_sync_done

macbre:
`auto_increment` for `image_sync` is initialized to 1 on mysql restart causing `Error: 1062 Duplicate entry '1' for key 'PRIMARY'` when moving rows from `image_sync` to `image_sync_done` table - https://dev.mysql.com/doc/refman/5.6/en/innodb-auto-increment-handling.html#innodb-auto-increment-initialization

``` sql
mysql@master.db-statsdb.service.consul[swift_sync]>ALTER TABLE image_sync_done MODIFY COLUMN id INT(8) auto_increment;
Query OK, 43299987 rows affected (8 min 50.95 sec)
Records: 43299987  Duplicates: 0  Warnings: 0
```

@mixth-sense / @drozdo

### DIFF
--- a/extensions/wikia/SwiftSync/classes/Queue.class.php
+++ b/extensions/wikia/SwiftSync/classes/Queue.class.php
@@ -203,15 +203,15 @@ class Queue {
 		}
 		
 		$dbw = self::getDB( true );
-		$dbw->begin();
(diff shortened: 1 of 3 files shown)


drozdo:
LGTM

PR #10407, opened 2016-05-06
(SEC-62) Properly escape and validate AdminDashboard header content

Grunny:
Properly escape and validate AdminDashboard header content.

/cc @Wikia/spitfires


lukaszkonieczny:
LGTM

PR #10409, opened 2016-05-06
PLATFORM-2192: Improve AbTesting defense againt bad config

wladekb:
https://wikia-inc.atlassian.net/browse/PLATFORM-2192

/cc @bognix @nefretete @nandy-andy

### DIFF
--- a/extensions/wikia/AbTesting/js/AbTest.js
+++ b/extensions/wikia/AbTesting/js/AbTest.js
@@ -19,7 +19,7 @@
 		logCache = {},
 		beacon = window.beacon_id || getCookie('wikia_beacon_id'),
 		serverTimeString = window.varnishTime,
-		serverDate = serverTimeString ? new Date( serverTimeString ) : new Date();
+		serverDate = serverTimeString ? new Date( serverTimeString ) : new Date(),
 		serverTime = serverDate.getTime() / 1000;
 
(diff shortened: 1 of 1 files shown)


extensions/wikia/AbTesting/js/AbTest.js
```diff
@@ -207,6 +207,9 @@
 	// Returns true if value falls into provided ranges
 	function isInRanges( value, ranges ) {
 		var i, range;
+		if ( !ranges || !ranges.length ) {
```
bognix, line 210: you don't have to check the `ranges.length`
If length is 0 then it won't go into the `for` loop and it will return `false`.

You can consider using `Array.isArray` method - https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray inside if statement

wladekb, line 210: Object.isArray is available in IE9+, I wouldn't like to intentionally break it for IE8-
`ranges.length` check was supposed to emulate isArray.

bognix, line 210: I understand 👍 however we use `Array.isArray` ~170 times in our application.

wladekb, line 210: I've just run grep and it seems we rarely use `Array.isArray` blindly. I'll stick with the current code.

PR #10399, opened 2016-05-06
JPN-407 Entry point for more than 3 admins

kamilkoterba:
## Links
- https://wikia-inc.atlassian.net/browse/JPN-407
## Description

Adding other admins entry pointy with count that opens admins tab on modal.
## Reviewers

@Wikia/spitfires @perjg @d34th4ck3r

PR #10410, opened 2016-05-06
JPN-375 Tracking for community page

perjg:


extensions/wikia/CommunityPage/scripts/ext.communityPage.js
```diff
@@ -170,18 +178,52 @@ require([
...
+					label: data,
+				});
+			}
+		});
+	}
+
+	function initModalTracking() {
+		// Track clicks in contribution modal
+		$('#CommunityPageModalDialog').on('mousedown touchstart', 'a', function (event) {
+			var data = $(event.currentTarget).data('tracking');
```
Grunny: This handler function is exactly the same as the one for the module tracking above, so I'd move it out into a `handleClick` method and then reuse it for both.

extensions/wikia/CommunityPage/templates/topContributorsModal.mustache
```diff
@@ -16,10 +16,10 @@
 	{{#contributors}}
 		<li>
 			<div class="avatar-container">
-				<a href="{{profilePage}}">{{{avatar}}}</a>
+				<a data-tracking="top-contributors-modal-avatar" href="{{profilePage}}">{{{avatar}}}</a>
```
kamilkoterba: I think top-contributors-modal-user-avatar will be more clear

extensions/wikia/CommunityPage/templates/topContributorsModal.mustache
```diff
@@ -16,10 +16,10 @@
...
 			</div>
 			<div class="contributor-details">
-				{{count}}. <a href="{{profilePage}}">{{userName}}</a>
+				{{count}}. <a data-tracking="top-contributors-modal" href="{{profilePage}}">{{userName}}</a>
```
kamilkoterba: I think top-contributors-modal-user will be more clear

extensions/wikia/CommunityPage/templates/topContributors.mustache
```diff
@@ -19,7 +19,7 @@
 				<a href="{{profilePage}}">{{{avatar}}}</a>
 			</div>
 			<div class="contributor-details">
-				{{count}}. <a href="{{profilePage}}">{{userName}}</a>
+				{{count}}. <a data-tracking="top-contributors" href="{{profilePage}}">{{userName}}</a>
```
kamilkoterba: I think format like top-contributors-{modal/module}-user will be more clear which means we would need to provide different label depending whether topContributors template is used in module or modal.

What do you think about it?

extensions/wikia/CommunityPage/templates/topContributors.mustache
```diff
@@ -19,7 +19,7 @@
 				<a href="{{profilePage}}">{{{avatar}}}</a>
```
kamilkoterba: Please add tracking for avatar as well

extensions/wikia/CommunityPage/templates/recentlyJoined.mustache
```diff
@@ -2,10 +2,10 @@
...
 	{{#members}}
 		<div class="avatar-container">
-			<a href="{{profilePage}}">{{{avatar}}}</a>
+			<a data-tracking="recently-joined-avatar" href="{{profilePage}}">{{{avatar}}}</a>
 		</div>
 	{{/members}}
 {{/haveMembers}}
 <span class="recently-joined">
-	<a href="#" id="viewAllMembers">{{allMembers}}</a>
+	<a data-tracking="show-modal" href="#" id="viewAllMembers">{{allMembers}}</a>
```
kamilkoterba: In upcoming https://github.com/Wikia/app/pull/10399 we'll adding new entry point to open modal. What about changing the label to tell which entry point is clicked? e.g. show-modal-all

kamilkoterba:
@grunny, @adamk, @lukaszk, @perjg Second thought on labels. Initially we decided with Per to add one category for community-page and than distinguish by labels. Although I think it may get complicated with time. What do you think about add categories per UI group:
- community-page-contributors-module
- community-page-contributors-modal
- community-page-header
- community-page-recent-activity
- etc.

extensions/wikia/CommunityPage/templates/topContributors.mustache
perjg: There are already two templates for this: topContributors.mustache and topContributorsModal.mustache. The latter is used in the modal only and has the tracking label top-contributors-modal.

perjg: @kamilkoterba Right now, the special tracking has the tag -modal added, with no suffix for the main module. I'm ok with adding -module to the ones in the main module. I'm not so sure about adding -user to the end of everything.

kamilkoterba: Cool.
I would add user only to links that redirects to user profile page. What are your concerns?

perjg: Ok, that makes sense. I misunderstood your intention.

kamilkoterba: 👍

kamilkoterba:
LGTM. moving o testing

kamilkoterba:
Testing passed.

PR #10411, opened 2016-05-06
PLATFORM-2147: Copy WF vars to dev when importing wiki

wladekb:
We've had a long-standing issue that WF variables were not copied to dev. This had to be done because `mysqlimport --replace` removed var values for all wikis. This change attempts to fix that by importing only non-existent variables.

On the other hand this will have a side-effect of overwriting dev-only WF variables. Is it something we can live with? Supporting variable id remapping would make the code longer but it is doable. I'm open to suggestions in this point.

https://wikia-inc.atlassian.net/browse/PLATFORM-2147

/cc @owend @mixth-sense


owend:
I like this implementation. 👍  from me.   I'm okay with overwriting variables in dev, most of the time it's what we want (because some production variable changed).  Devs should be using PHP settings to set custom variables for testing.

PR #10412, opened 2016-05-06
SEO-218 Improve search query for "Not a valid Wikia" page

gabrys:
Make sure there are always some results once you click the "search"
link.


gabrys:
Doesn't work as expected. Removing changes.

PR #10415, opened 2016-05-08
Lyr 262

SeanColombo:
Modernized the LyricWiki extensions not to use mysql_ anywhere. There were some chunks of commented-out code I just swapped to mysqli_ (note the "i") but in general I switched them to using MediaWiki's database objects.

Also, extracted the common code from Soap search failures and Mobile search failures to share a base-class SearchFailuresSpecialPage so that the code only needs to be maintained in one spot.

Tested these on my devbox and all looks good. Strangely, in production with current code, no data shows up... the tables appear to be there in production dbs, so hopefully when we push, these updates will make some of this visible again.

In production, Special:LinksToRedirects and Special:ArtistRedirects have regressed to basically being blank and this is likely hurting SEO. Will be good to keep an eye on LW traffic a couple of weeks after this change to see if it results in an improvement.

### DIFF
--- a/extensions/3rdparty/LyricWiki/Special_LinksToRedirects.php
+++ b/extensions/3rdparty/LyricWiki/Special_LinksToRedirects.php
@@ -176,8 +176,15 @@ class Linkstoredirects extends SpecialPage{
 					);
 					$idToTitle = array();
 					$ids = array_unique($ids);
-					$queryString = "SELECT page_id, page_namespace, page_title FROM $TABLE_PREFIX"."page WHERE page_id IN (".implode(",", $ids).")";
-					$res2 = $dbr->query($queryString);
+
+					$res2 = $dbr->select(
+						"page",
+						array( "page_id",
+							   "page_namespace",
+							   "page_title"),
+						"page_id IN (".implode(",", $ids).")",
+						__METHOD__
+					);
 					while ($innerRow = $dbr->fetchObject($res2)) {
 						$id = $innerRow->page_id;
 						$ns = $innerRow->page_namespace;
--- a/extensions/3rdparty/LyricWiki/Special_LinksToRedirects.php
+++ b/extensions/3rdparty/LyricWiki/Special_LinksToRedirects.php
@@ -184,7 +184,9 @@ class Linkstoredirects extends SpecialPage{
 							"page_namespace",
 							"page_title"
 						],
-						"page_id" => $ids,
+						[
+							"page_id" => $ids,
+						]
 						__METHOD__
 					);
 					while ($innerRow = $dbr->fetchObject($res2)) {
--- a/extensions/3rdparty/LyricWiki/Special_LinksToRedirects.php
+++ b/extensions/3rdparty/LyricWiki/Special_LinksToRedirects.php
@@ -184,7 +184,7 @@ class Linkstoredirects extends SpecialPage{
 							"page_namespace",
 							"page_title"
 						],
-						"page_id IN (".implode(",", $ids).")",
+						"page_id" => $ids,
 						__METHOD__
 					);
 					while ($innerRow = $dbr->fetchObject($res2)) {
--- a/extensions/3rdparty/LyricWiki/Special_LinksToRedirects.php
+++ b/extensions/3rdparty/LyricWiki/Special_LinksToRedirects.php
@@ -179,9 +179,11 @@ class Linkstoredirects extends SpecialPage{
 
 					$res2 = $dbr->select(
 						"page",
-						array( "page_id",
-							   "page_namespace",
-							   "page_title"),
+						[
+							"page_id",
+							"page_namespace",
+							"page_title"
+						],
 						"page_id IN (".implode(",", $ids).")",
 						__METHOD__
 					);
--- a/extensions/3rdparty/LyricWiki/Special_LinksToRedirects.php
+++ b/extensions/3rdparty/LyricWiki/Special_LinksToRedirects.php
@@ -143,7 +143,7 @@ class Linkstoredirects extends SpecialPage{
 				while ($row = $dbr->fetchObject($res)) {
 					$totalResults++;
 					$pageId = $row->from_id;
-					$target = $row->links_to;3
+					$target = $row->links_to;
 					$allListings[] = array($pageId, $target);
 
 					// Somehow, a blank from_id can slip into the results (which messes up the query).
--- a/extensions/3rdparty/LyricWiki/Special_LinksToRedirects.php
+++ b/extensions/3rdparty/LyricWiki/Special_LinksToRedirects.php
@@ -48,8 +48,7 @@ class Linkstoredirects extends SpecialPage{
 		$CACHE_KEY = "LW_LINKS_TO_REDIRECTS";
 
 		$wgOut->setPageTitle(wfMsg('linkstoredirects'));
-
-	/*
+/*
 		// This processes any requested for removal of an item from the list.
 		if(isset($_GET['artist']) && isset($_GET['song'])){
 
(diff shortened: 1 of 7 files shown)


extensions/3rdparty/LyricWiki/server.php
```diff
@@ -2477,13 +2476,19 @@ function requestStarted($funcName, $requestData){
...
-				$retVal = mysql_insert_id( $db );
-			}
+			
+			$dbr = wfGetDB( DB_MASTER );
+			$dbr->insert(
+				'apiRequests',
+				[
+					"requestedThrough" => $REQUEST_TYPE,
+					"requestedFunction" => $dbr->addQuotes(funcName),
+					"requestData" => $dbr->addQuotes(requestData),
```
macbre: `Database::insert` properly escapes the values internally. No need for `addQuotes` calls here.

extensions/3rdparty/LyricWiki/Special_Soapfailures.php
```diff
@@ -109,13 +114,14 @@ function execute( $par ){
...
 				print_r($songResult);
 			} else {
-				$dbw = wfGetDB( DB_MASTER );
+				$artist = str_replace("'", "\\'", $artist);
+				$song = str_replace("'", "\\'", $song);
 
 				print "<html><head><title>Success</title></head><body>\n"; // TODO: i18n
 				print "Deleting record... ";
-
+				$dbw = &wfGetDB(DB_MASTER);
```
macbre: No need for a reference here (`&`).

extensions/3rdparty/LyricWiki/Special_LinksToRedirects.php
```diff
@@ -139,107 +136,105 @@ function execute( $par ){
...
+						10 => ":Template",
+						11 => "Template_talk",
+						12 => "Help",
+						13 => "Help_talk",
+						14 => ":Category",
+						15 => "Category_talk"
+					);
+					$idToTitle = array();
+					$ids = array_unique($ids);
+					$queryString = "SELECT page_id, page_namespace, page_title FROM $TABLE_PREFIX"."page WHERE page_id IN (".implode(",", $ids).")";
```
macbre: Let's be consistent with the rest of the change and use `Database::select` helper instead of providing a raw SQL query.

extensions/3rdparty/LyricWiki/Special_Soapfailures.php
```diff
@@ -171,34 +177,33 @@ function execute( $par ){
...
-						for($cnt=0; $cnt<$numRows; $cnt++){
-							$row = array();
-							$row['artist'] = mysql_result($result, $cnt, "request_artist");
-							$row['song'] = mysql_result($result, $cnt, "request_song");
-							$row['numRequests'] = mysql_result($result, $cnt, "numRequests");
-							$row['lookedFor'] = mysql_result($result, $cnt, "lookedFor");
-							$row['lookedFor'] = formatLookedFor($row['lookedFor']);
-							$data[] = $row;
-						}
+				$queryString = "SELECT * FROM {$this->TABLE_NAME} ORDER BY numRequests DESC LIMIT $MAX_RESULTS";
```
macbre: Let's be consistent with the rest of the change and use `Database::select` helper instead of providing a raw SQL query.

And please provide explicit list of columns you want to get (instead of `*`).

extensions/3rdparty/LyricWiki/server.php
```diff
@@ -2477,13 +2476,19 @@ function requestStarted($funcName, $requestData){
...
+					"requestTime" => "NOW()",
+				],
+				__METHOD__
+			);
 		}
 	}
 	return $retVal;
```
macbre, line 2495: `$retVal` is now set to `false` at the beginning of `requestStarted` function only. Is that correct?

extensions/3rdparty/LyricWiki/server.php
```diff
@@ -2477,13 +2476,19 @@ function requestStarted($funcName, $requestData){
...
-			}
+			
+			$dbr = wfGetDB( DB_MASTER );
+			$dbr->insert(
+				'apiRequests',
+				[
+					"requestedThrough" => $REQUEST_TYPE,
+					"requestedFunction" => $dbr->addQuotes(funcName),
+					"requestData" => $dbr->addQuotes(requestData),
+					"requestTime" => "NOW()",
```
macbre: `NOW()` will be passed as a string not the MySQL function call.

extensions/3rdparty/LyricWiki/server.php
SeanColombo: Thanks. Removed in next commit.

extensions/3rdparty/LyricWiki/Special_Soapfailures.php
SeanColombo: Removed in next commit.

extensions/3rdparty/LyricWiki/server.php
SeanColombo: Good catch! Changed to $dbw->timestamp( wfTimestampNow() ) in next commit.

extensions/3rdparty/LyricWiki/Special_LinksToRedirects.php
SeanColombo: Changed in next commit. Tested on devbox & still works.

extensions/3rdparty/LyricWiki/Special_Soapfailures.php
SeanColombo: Specified columns, changed to ->select() and tested on devbox... will be in the 2nd additional commit.

extensions/3rdparty/LyricWiki/Special_LinksToRedirects.php
SeanColombo: I used array() at first... please disregard... changing that to [ ] ;)

extensions/3rdparty/LyricWiki/server.php
SeanColombo, line 2495: Nope, that's definitely wrong. Needed to add ->insertId().  Also noticed that it was called "dbr" here instead of "dbw" so fixed both of those issues.

SeanColombo:
What's the process when there are a bunch of resolved comments like this, @macbre? Do we review again?

extensions/3rdparty/LyricWiki/Special_LinksToRedirects.php
```diff
@@ -176,8 +176,17 @@ function execute( $par ){
...
-					$res2 = $dbr->query($queryString);
+
+					$res2 = $dbr->select(
+						"page",
+						[
+							"page_id",
+							"page_namespace",
+							"page_title"
+						],
+						"page_id IN (".implode(",", $ids).")",
```
macbre: You can use the following instead:

``` php
[
  'page_id' => $ids
]
```

MediaWiki DB abstraction layer will replace this one (when `$ids` is an array) with `page_id IN (....)`.

extensions/3rdparty/LyricWiki/Special_LinksToRedirects.php
```diff
@@ -176,8 +176,15 @@ function execute( $par ){
...
 					$ids = array_unique($ids);
-					$queryString = "SELECT page_id, page_namespace, page_title FROM $TABLE_PREFIX"."page WHERE page_id IN (".implode(",", $ids).")";
-					$res2 = $dbr->query($queryString);
+
+					$res2 = $dbr->select(
+						"page",
+						array( "page_id",
+							   "page_namespace",
+							   "page_title"),
+						"page_id IN (".implode(",", $ids).")",
```
macbre: You can use the following instead:

```php
[
  'page_id' => $ids
]
```

MediaWiki DB abstraction layer will replace this one (when `$ids` is an array) with `page_id IN (....)`.

extensions/3rdparty/LyricWiki/Special_LinksToRedirects.php
SeanColombo: Changed in latest commit on branch.

SeanColombo: Github doesn't seem to have picked up the commit yet, but I did push it to this branch about 10 minutes ago. Will wait a bit.

SeanColombo: Github is apparently having issues right now, so this is expected.

extensions/3rdparty/LyricWiki/Special_LinksToRedirects.php
```diff
@@ -184,7 +184,7 @@ function execute( $par ){
 							"page_namespace",
 							"page_title"
 						],
-						"page_id IN (".implode(",", $ids).")",
+						"page_id" => $ids,
```
macbre: This needs to be wrapped inside an array:

``` php
[
  "page_id" => $ids
]
```

macbre:
LGTM

PR #10416, opened 2016-05-08
Lyr 264

SeanColombo:
Changed constructors from having their names match their classname to be __construct which is what PHP 7 expects.

Code runs on my devbox w/out crashing & I counted 3 individual files plus one file with 8 changes, both in the Jira Issue and in this changeset.

### DIFF
--- a/extensions/3rdparty/LyricWiki/Special_Wikify.body.php
+++ b/extensions/3rdparty/LyricWiki/Special_Wikify.body.php
@@ -6,7 +6,7 @@ class Wikify extends SpecialPage
 {
 	function __construct()
 	{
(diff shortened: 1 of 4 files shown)


extensions/3rdparty/LyricWiki/Special_Wikify.body.php
```diff
@@ -4,7 +4,7 @@
 
 class Wikify extends SpecialPage
 {
-	function Wikify()
+	function __construct()
 	{
 		SpecialPage::SpecialPage("Wikify");
```
macbre: Please change this one to `parent::__construct`.

macbre:
:+1

PR #10419, opened 2016-05-09
JPN-415 Fix global user table references

kamilkoterba:
@adamkarminski's fix for: https://wikia-inc.atlassian.net/browse/JPN-481

### DIFF
--- a/extensions/wikia/CommunityPage/models/CommunityPageSpecialUsersModel.class.php
+++ b/extensions/wikia/CommunityPage/models/CommunityPageSpecialUsersModel.class.php
@@ -33,7 +33,7 @@ class CommunityPageSpecialUsersModel {
 			wfMemcKey( self::TOP_CONTRIB_MCACHE_KEY, $limit, $weekly, $onlyAdmins ),
 			WikiaResponse::CACHE_STANDARD,
 			function () use ( $limit, $weekly, $onlyAdmins ) {
-				global $wgExternalSharedDB;
+				global $wgExternalSharedDB, $wgDBcluster;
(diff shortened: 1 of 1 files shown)


kamilkoterba:
LGTM

perjg:
There are a couple of other places in the CommunityPageSpecialUsersModel that access the external db like so:

```
global $wgExternalSharedDB;
$db = wfGetDB( DB_SLAVE, [], $wgExternalSharedDB );
```

For the change made for this task, wfGetDB is called like so:

```
global $wgExternalSharedDB;
$db = wfGetDB( DB_SLAVE );
```

@Grunny Would changing the call to wfGetDB allow the explicit `$wgExternalSharedDB . '_' . $wgDBcluster . '.user'` call to be simplified?

Grunny:
@perjg If you did that, then you'd still need to specify the local DB name for all the local tables being queried. On Wookieepedia `$db = wfGetDB( DB_SLAVE );` will return a DB connection for the `starwars` database, so if you want to join on the user table, you need to specify the wikicities database copy present on that cluster. If you did, `$db = wfGetDB( DB_SLAVE, $wgExternalSharedDB . '_' . $wgDBcluster . '.user' );` on Wookieepedia, it would return a connection using the `wikicities_c4` database, and you'd need to specify `starwars` when querying local DB data (i.e. `starwars.revision`). So, it wouldn't clean it up much. We may want to centralise some of this summarised data in the future in the datawarehouse, so this isn't needed.

Grunny:
Testing LGTM.

PR #10422, opened 2016-05-09
JPN-485 Remove modal tab data preloading

perjg:


Grunny:
LGTM.

extensions/wikia/CommunityPage/scripts/ext.communityPage.js
```diff
@@ -184,10 +184,5 @@ require([
 	$(function () {
 		// prefetch UI modal on DOM ready
 		getUiModalInstance();
```
lukaszkonieczny: Do we want to prefetch modal instance on DOM ready?

perjg: Yes, because the throbber is shown inside the modal (when loading tab data), so having the actual modal preloaded will avoid having to put up a throbber for loading the modal too.

Grunny:
LGTM, part II ;).

PR #10425, opened 2016-05-09
PLATFORM-2189: Support ascii mode in moxi

wladekb:
Gives an ability to switch moxi client into ascii mode via config variable.

https://wikia-inc.atlassian.net/browse/PLATFORM-2189

/cc @artursitarski @owend @mixth-sense @harnash


jcellary:
I'm lacking full context, but the code LGTM

PR #10428, opened 2016-05-09
SUS-520 unify assets loading

bognix:
Don't merge until SUS-520_cnw_refactor_v1 is not merge.
This PR should be created against `dev` branch.


extensions/wikia/AssetsManager/config.php
```diff
@@ -2878,3 +2878,23 @@
 		'//extensions/wikia/PortabilityDashboard/styles/PortabilityDashboard.scss',
 	],
 ];
+
+
```
bkoval: Why 2 new lines? :)

PR #10434, opened 2016-05-09
JPN-415 Fix for admins not correctly detected. With unit test

perjg:
https://wikia-inc.atlassian.net/browse/JPN-486
### Reviewers

@Wikia/spitfires @d34th4ck3r


extensions/wikia/CommunityPage/models/CommunityPageSpecialUsersModel.class.php
```diff
@@ -20,6 +20,16 @@ public function __construct() {
 	}
 
 	/**
+	 * Check if a user is an admin
+	 *
+	 * @param string $group The user's group from ug_group
+	 * @return bool
+	 */
+	public function isAdmin( $group ) {
+		return strpos( $group, 'sysop' ) !== false;
```
Grunny: Why `strpos`? `ug_group` should be an exact group name so `$group === 'sysop'` should work. Though does the query definitely always return sysop if the user is in multiple groups?

perjg: This was the original check. The function is called once for each row in a query result, with $row->ug_group. This fix was made on the assumption that ug_groups contains all the user's group space separated, which I think is not correct.

We get the data by joining wikia_user_properties and user_groups, so I guess that means that for users who have more than one group, we just get a random one.

Is there an easy and low-cost way to get all of a user's groups or an isAdmin flag when we have the user id?

Grunny: Well, you can create the user object and get all groups, but doing this for every user would be potentially expensive. Have you considered using `WikiService::getWikiAdminIds`, which should only do one efficient query if it's not in cache, and then just use `in_array` for `isAdmin`?

perjg: Thanks, that sounds like a better solution.

Grunny:
LGTM.

PR #10437, opened 2016-05-10
SUS-385 Wall and Forum UI fix

ghost:
Currently there is no visible UI response to submit click. It happens often, that users click the button one more time which results with duplicated messages. In order to prevent that we disable buttons on submit and re-enable on error.

https://wikia-inc.atlassian.net/browse/SUS-385

@vforge 

Do not merge until all the tests are green.


vforge:
LGTM, php unit test fail is not connected to this change - see https://wikia.slack.com/archives/engineering/p1462882737000189

ludwikkazmierczak:
tested and accepted

PR #10445, opened 2016-05-11
toggle logging User::invalidateCache calls

ghost:
https://wikia-inc.atlassian.net/browse/SUS-546

As part of https://wikia-inc.atlassian.net/browse/SUS-545 we need to determine, when SMW calls `User::invalidateCache()`, which produces tons of queries to shared db.

@vforge @ludwikkazmierczak


vforge:
LGTM

ludwikkazmierczak:
LGTM

PR #10449, opened 2016-05-11
Sanitize notificationKey

vforge:
https://wikia-inc.atlassian.net/browse/SUS-543

Sanitize key in WallNotificationController - make XSS impossible.

/cc @mixth-sense @Grunny

### DIFF
--- a/extensions/wikia/WallNotifications/WallNotificationControllerBase.class.php
+++ b/extensions/wikia/WallNotifications/WallNotificationControllerBase.class.php
@@ -48,10 +48,11 @@ abstract class WallNotificationControllerBase extends WikiaController {
 		wfProfileIn( __METHOD__ );
 
 		$this->response->setVal( 'alwaysGrouped', empty( $wgEnableWallExt ) && empty( $wgEnableForumExt ) );
-		$this->response->setVal( 'notificationKey', $this->request->getVal( 'notificationKey' ) );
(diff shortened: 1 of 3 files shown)


extensions/wikia/WallNotifications/WallNotificationControllerBase.class.php
```diff
@@ -48,10 +48,11 @@ public function Update() {
 		wfProfileIn( __METHOD__ );
 
 		$this->response->setVal( 'alwaysGrouped', empty( $wgEnableWallExt ) && empty( $wgEnableForumExt ) );
-		$this->response->setVal( 'notificationKey', $this->request->getVal( 'notificationKey' ) );
+		$this->response->setVal( 'notificationKey', htmlentities( $this->request->getVal( 'notificationKey' ), ENT_QUOTES ) );
```
Grunny: This is being output into a HTML attribute, so I'd prefer `Sanitizer::encodeAttribute`, and it's also generally preferable to escape as close to the output as possible, so I'd consider escaping in the template as a best practice, but this is OK as well.

Grunny:
One suggestion, but LGTM in general.

vforge:
@Grunny Thanks for the suggestion, I've noticed both approaches in the code - I decided to proceed with your tip.

Grunny:
LGTM.

Grunny:
This also needs to be handled in GlobalNavigationWallNotifications_Update.php.

Grunny:
LGTM and tested.

vforge:
@Grunny Thanks for finishing it up at night!

PR #10450, opened 2016-05-11
optimize task scheduling

ghost:
@vforge @ludwikkazmierczak @Grunny 

We have a couple of nested `foreach`'s - `$task->call()` is called multiple times whereas it only needs to be called once per `$title`. I am now adding `$title`'s to an array, removing possible duplicates and interate through the array to schedule tasks

### DIFF
--- a/extensions/wikia/SemanticMediaWiki/includes/storage/SQLStore/SMW_SQLStore3_Writers.php
+++ b/extensions/wikia/SemanticMediaWiki/includes/storage/SQLStore/SMW_SQLStore3_Writers.php
@@ -777,7 +777,7 @@ class SMWSQLStore3Writers {
 			if ( $smwgEnableUpdateJobs && ( $smwgQEqualitySupport != SMW_EQ_NONE ) ) {
 				// entries that refer to old target may in fact refer to subject,
 				// but we don't know which: schedule affected pages for update
-				$jobs = array();
+				$titles = array();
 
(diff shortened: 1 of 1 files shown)


drozdo:
LGTM, lets try it, it wont get worse

vforge:
LGTM

ludwikkazmierczak:
LGQM

extensions/wikia/SemanticMediaWiki/includes/storage/SQLStore/SMW_SQLStore3_Writers.php
```diff
@@ -777,7 +777,7 @@ protected function updateRedirects( $subject_t, $subject_ns, $curtarget_t = '',
 			if ( $smwgEnableUpdateJobs && ( $smwgQEqualitySupport != SMW_EQ_NONE ) ) {
 				// entries that refer to old target may in fact refer to subject,
 				// but we don't know which: schedule affected pages for update
-				$jobs = array();
+				$titles = array();
```
rogatty, line 780: `// wikia change start` or it doesn't make sense for just a rename?

ghost, line 780: It does. This method is going to be thoroughly refactor. That was just to verify the concept and possibly mitigate performance problem.

PR #10451, opened 2016-05-11
DAT-4156

rybmat:
A/B test - marking navboxes and navigation templates to measure if hiding them has impact on page view per session

//cc: @Wikia/west-wing


nikodamn:
LGTM

sebastianmarzjan:
:shipit:

PR #10454, opened 2016-05-11
(JPN-495) Alter PMP entry point experiments tracking labels

Grunny:
Make sure all PMP entry point tracking events have labels and include the
variation in the label as a prefix.

/cc @Wikia/spitfires


adamkarminski:
LGTM

PR #10456, opened 2016-05-12
do not set $smwgEnableUpdateJobs if already set

ghost:
https://wikia-inc.atlassian.net/browse/SUS-545

We suspect SMW falls into an infinite loop of spawning new jobs. We need to be able to stop it on per-wiki basis with WikiFactory. The plan is to add `$smwgEnableUpdateJobs` and set it to `false` as problems materialize. Other than that this change is not really a change :-)

@Grunny @ljagiello @ludwikkazmierczak @vforge


vforge:
LQGTM

PR #10465, opened 2016-05-13
SUS-542 | Fix XSS vulnerabilities in Wikia Search

vforge:
- https://wikia-inc.atlassian.net/browse/SUS-545

Sanitize attributes and strip HTML from params.

/cc @mixth-sense @Grunny

### DIFF
--- a/extensions/wikia/Search/templates/WikiaSearch_CrossWiki_index.php
+++ b/extensions/wikia/Search/templates/WikiaSearch_CrossWiki_index.php
@@ -10,7 +10,7 @@
 						<?= wfMsg('wikiasearch2-results-for', '<strong>' . $query . '</strong>'); ?>
 					<?php endif; ?>
 					<?php if (isset($hub) && $hub) : ?>
-						<?= wfMessage( 'wikiasearch2-onhub', Sanitizer::stripAllTags( $hub ) )->text() ?>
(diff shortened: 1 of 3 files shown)


extensions/wikia/Search/templates/WikiaSearch_CrossWiki_index.php
```diff
@@ -10,7 +10,7 @@
 						<?= wfMsg('wikiasearch2-results-for', '<strong>' . $query . '</strong>'); ?>
 					<?php endif; ?>
 					<?php if (isset($hub) && $hub) : ?>
-						<?= wfMsg('wikiasearch2-onhub', $hub) ?>
+						<?= wfMessage( 'wikiasearch2-onhub', Sanitizer::stripAllTags( $hub ) )->text() ?>
```
Grunny: This will still be vulnerable because of the use of `->text()`, since the wikitext will be expanded unescaped after you strip the tags, allowing an attacker to still output raw HTML. Should be `->escaped()` here and below.

vforge: Yeah, I changed it to the text and should other way around - thanks for noticing!

Grunny:
LGTM.

ludwikkazmierczak:
Tested and Accepted

PR #10467, opened 2016-05-13
[SUS-552] Removed unused CreateNewWiki code

jcellary:
https://wikia-inc.atlassian.net/browse/SUS-552

Removed unneeded fixWikia maintenance script and the job CreateWikiLocalJob which was used only by that script.

@bognix @mixth-sense @vforge


extensions/wikia/ImageReview/PromoteImageReview.setup.php
```diff
@@ -30,6 +30,7 @@
 if( function_exists('extAddBatchTask') ) {
 	extAddBatchTask($dir . "PromoteImageReviewTask.php", "promoteimagereview", "PromoteImageReviewTask");
 }
+//Possibly this line can be removed (as the usage new Wikia\\Tasks\\Tasks\\PromoteImageReviewTask was removed)
```
bognix, line 33: could you create a ticket to investigate that? or do it as part of this code change?

bognix:
the change LGTM although I don't have enough context to tell you if it's safe the remove this

extensions/wikia/ImageReview/PromoteImageReview.setup.php
jcellary, line 33: I consulted that with Mix and Paweł and Sus is already working on retiring special promote, so they'll handle that as part of that effort (for which I presume they already have a ticket)

bognix, line 33: Maybe we should add a note about this part of code to this ticket? It would be easier for them in the future to find the usages.

ludwikkazmierczak:
Tested And Accepted

PR #10471, opened 2016-05-13
JPN-458 Introduce a pattern instead of solid color to zero state hero image

adamkarminski:
Let's get rid of the hideous alternate solid color!


lukaszkonieczny:
LGTM

lukaszkonieczny:
still LGTM

PR #10473, opened 2016-05-13
Fix duplicated users

lukaszkonieczny:
One user may may belong to many groups so in results we get duplicated users


Grunny:
LGTM.

PR #10474, opened 2016-05-13
JPN-436 Update designs of the Insights module

adamkarminski:
Having the "View full list" link on the right made it hard to recognize which module it belongs to for more than 2 of them.

This PR puts it between a description and a list. Also, the top border of the first item on a list is removed.


perjg:
LGTM

PR #10489, opened 2016-05-16
ADEN-3187 Use SP detection in Yavli

drets:
Use SourcePoint detection (ad un-blocking software) and show Yavli (Taboola-like provider) only for AdBlock users.


nandy-andy:
LGTM :+1:

PR #10492, opened 2016-05-16
JPN-436 Insights modules fixes

adamkarminski:
@perjg


extensions/wikia/CommunityPage/styles/components/_insights-module.scss
```diff
@@ -23,15 +23,12 @@
 
 	.community-page-insights-module-list-item {
 		border-bottom: 1px solid $color-page-border-subtle;
+		display: flex;
```
perjg: In other places in the module we use a mixin for flexbox. I am not sure if this is still a requirement, but we should at least be consistent.

perjg:
LGTM

adamkarminski:
Tests LGTM

PR #10494, opened 2016-05-17
JPN-536: Use proper separator and fix list order

lukaszkonieczny:


extensions/wikia/CommunityPage/models/CommunityPageSpecialInsightsModel.class.php
```diff
@@ -40,14 +51,16 @@ public function getInsightsModules() {
...
+	 * 	- how data should be sorted (@see InsightsSorting::$sorting)
 	 * @return array Insight Module
 	 */
-	private function getInsightModule( $type, $sortingType = self::INSIGHTS_MODULE_SORT_TYPE ) {
+	private function getInsightModule( $type, $config ) {
 		$insightPages = $this->insightsService->getInsightPages(
 			$type,
 			self::INSIGHTS_MODULE_ITEMS,
-			$sortingType
+			$config['sortingType']
```
kamilkoterba: Would be much better to have a const for this name as it's reused in few places...

extensions/wikia/CommunityPage/models/CommunityPageSpecialInsightsModel.class.php
```diff
@@ -72,29 +85,32 @@ private function getInsightModule( $type, $sortingType = self::INSIGHTS_MODULE_S
...
+	private function addLastRevision( $insightsPages, $displayPageviews ) {
+		global $wgLang;
+
 		foreach ( $insightsPages['pages'] as $key => $insight ) {
 			$insightsPages['pages'][$key]['metadataDetails'] = $this->getArticleMetadataDetails( $insight['metadata'] );
 			$insightsPages['pages'][$key]['editlink'] = $this->getEditUrl( $insight['link']['articleurl'] );
 			$insightsPages['pages'][$key]['edittext'] = $this->getArticleContributeText( $insight['metadata'] );
 
-			if ( !empty( $insight['metadata']['pv7'] ) ) {
+			if ( $displayPageviews ) {
```
kamilkoterba: What if `$insight['metadata']['pv7']` will be empty? Sounds like potential place for throwing an error?
Simple Assert::true( !empty( $insight['metadata']['pv7'] ), 'log message'  ) would let us be notified by ErrorReported if there are many of those.

extensions/wikia/CommunityPage/models/CommunityPageSpecialInsightsModel.class.php
kamilkoterba: or if it would add too much complexity maybe easier would be to choose more unique keys and leave it this way as I see that we're already calling array properties directly in many places.... :/

kamilkoterba:
LGTM

kamilkoterba:
Tested and it works fine for popular pages. Can we also add this for number of links in wanted pages and number of user edits for the rest of modules?

lukaszkonieczny:
Sure! Good catch :+1:

kamilkoterba:
Tested. LGTM

PR #10498, opened 2016-05-17
JPN-482 Add more unique selectors prefix for entrypoint selectors

kamilkoterba:
@Wikia/spitfires @perjg I think just entry-point prefix is too generic. What do you think?


lukaszkonieczny:
My only concern is should we have two different prefixes

kamilkoterba:
Full prefix is on parent and abbreviation is added to childs

kamilkoterba:
Changed approach to have consistency around all selectors and went with full prefixes

lukaszkonieczny:
LGTM

PR #10499, opened 2016-05-17
JPN-532: Fix removing fixed item from cached insights sorting arrays

lukaszkonieczny:
\cc @Wikia/spitfires


adamkarminski:
LGTM

PR #10501, opened 2016-05-17
JPN-539 Redesign of Entry Point

perjg:
Entry point redesigned according to new zeplin design:
- NEW marker moved to top left and changed size
- Welcome text split into two lines with different font sizes.
- Wikia icon moved up slightly

Also updated latest translations from crowdin.


extensions/wikia/CommunityPage/styles/entrypoint/EntryPoint.scss
```diff
@@ -70,19 +71,26 @@ $logo-url: $logo-url-light;
...
+		line-height: 17px;
+		position: absolute;
 		text-align: center;
 		text-transform: uppercase;
+		top: 0;
+		width: 41px;
+	}
+
+	.entry-point-description-top {
+		font-size: 14px;
```
kamilkoterba: font size 14px is defined for parent already so I think is should be safe to skip it

extensions/wikia/CommunityPage/styles/entrypoint/EntryPoint.scss
```diff
@@ -36,7 +36,7 @@ $logo-url: $logo-url-light;
 		left: 50%;
 		position: absolute;
 		top: 0;
-		transform: translate(-50%, -50%);
+		transform: translate(-50%, -40%);
```
lukaszkonieczny, line 39: Why we are using `transform` here?

extensions/wikia/CommunityPage/CommunityPage.i18n.php
```diff
@@ -113,13 +114,14 @@
...
-	'communitypage-modal-tab-loadingerror' => "Error message shown when ajax call for data failed",
+	'communitypage-modal-tab-loadingerror' => 'Error message shown when ajax call for data failed',
 	'communitypage-no-contributions' => 'Text indicating that there are no contributions yet this week',
 	'communitypage-no-admins' => 'Text indicating that the wiki has no admins.',
 	'communitypage-no-admins-contact' => 'Contact us to become the first admin text',
 	'communitypage-communitycentral-link' => 'Link to the community page for this language',
 	'communitypage-no-members' => 'Text for zero state of all members list',
-	'communitypage-help-us-grow' => 'Entry point for logged in users. Invitation to help this community',
+	'communitypage-help-us-grow-prefix' => 'Entry point for logged in users. Invitation to help this community, first part of text',
+	'communitypage-help-us-grow' => 'Second part of text, including sitename',
```
kamilkoterba: I think just for clear description for translators we could repeat description of it's purpose to second part as well, e.g.
'communitypage-help-us-grow' => 'Entry point for logged in users. Invitation to help this community, second part of text, including sitename',

Also I'd recommend tie these names more explicit, like:
communitypage-help-us-grow-before-sitename
communitypage-help-us-grow-after-sitename-included
what do you think

extensions/wikia/CommunityPage/styles/entrypoint/EntryPoint.scss
perjg, line 39: Discussed with @lukaszkonieczny: Leaving it like this for now. We may change transform translate at some later point based on further discussion.

kamilkoterba:
LGTM

kamilkoterba:
Tested

PR #10506, opened 2016-05-17
JPN-538 Break line before (but not after) wiki name when it's long

kamilkoterba:
https://wikia-inc.atlassian.net/browse/JPN-538

As the requirement is a bit tricky it's not possible to implement pure css solution.
It's hacky although I'm open for suggestions.

@Wikia/spitfires @perjg


perjg:
I'm ok with the workaround here, I don't think it is possible to implement in CSS directly.

kamilkoterba:
Solved by simplification here https://github.com/Wikia/app/commit/1f41e5eace683a4c88e4af2e8d883cc5b02e293a

PR #10508, opened 2016-05-17
[Cake-127] - recirc recommendations from content-entity service

nmonterroso:
@Wikia/cake @pauloslund 
https://wikia-inc.atlassian.net/browse/CAKE-127

Adding another treatment group to the `RECIRCULATION_PLACEMENT` experiment (`CAKE_RELATED_CONTENT`) that sources recommendations from the `content-entity` service. 

I also added the `RecirculationContent` class so it's clear what's expected of the output from one of the services, to remove the guesswork I went through when figuring out what I needed to return to have it rendered properly.

We can merge this but we shouldn't enable the treatment group until we fix https://wikia-inc.atlassian.net/browse/CAKE-130 and https://wikia-inc.atlassian.net/browse/CAKE-131

note that all code under `lib/Swagger` added in this PR is autogenerated code for talking to the service

@pauloslund - can you confirm that we don't need to add anything here to take advantage of the event tracking that's being used for the other treatment groups? It seems that's abstracted at a lower level but just want to be sure


pauloslund:
This all looks good to me. You're right about event tracking. It is handled at the view layer so it will work without adding anything else here.

extensions/wikia/Recirculation/js/helpers/CakeRelatedContentHelper.js
```diff
@@ -0,0 +1,48 @@
...
+        limit: 3
+    };
+
+    function loadData() {
+        var deferred = $.Deferred(),
+            currentArticle = window.location.pathname.replace('_', ' ');
+
+        if (currentArticle.startsWith('/wiki/')) {
+            currentArticle =  currentArticle.split('/wiki/')[1];
+        }
```
drsnyder, line 15: Do we always offset at `/wiki`?

drsnyder:
LGTM

extensions/wikia/Recirculation/js/helpers/CakeRelatedContentHelper.js
nmonterroso, line 15: only if the current path starts with it

PR #10509, opened 2016-05-17
RECIRC-123: Update GENRE and TOPIC tests to show 5 results

pauloslund:
Update experiments that are currently running and add some stability to existing code


saipetch:
LGTM

PR #10510, opened 2016-05-18
PLATFORM-2200: Use critical section when invalidating user

wladekb:
We have still a lot of trouble with deadlocks with `User::invalidateCache` for automatic logging from Helios token. Let's try a solution using critical section.

https://wikia-inc.atlassian.net/browse/PLATFORM-2200

/cc @kvas-damian @owend @macbre @mixth-sense @harnash @frankfarmer


includes/wikia/WikiaDataAccess.class.php
```diff
@@ -302,6 +302,18 @@ static private function getDataAndVerify( $app, $key, $clearLocalCache = false )
 		return $result;
 	}
 
+	static public function criticalSection( $key, $lockTimeout, callable $fn, &$result = null ) {
```
kvas-damian: Why do we need to return value by reference in this method?
Maybe we could throw an exception when lock cannot be acquired?

extensions/wikia/Helios/User.class.php
```diff
@@ -107,17 +107,16 @@ public static function newFromToken( \WebRequest $request ) {
...
+						wfSetupSession( $sessionId );
 						WikiaLogger::instance()->debug( __METHOD__ . '::startSession' );
 
-						// Update mTouched on user when he starts new MW session
-						// @see SOC-1326
-						$invalidateCacheThrottleTime = $request->getSessionData( self::INVALIDATE_CACHE_THROTTLE_SESSION_KEY );
-						if ( $invalidateCacheThrottleTime === null || $invalidateCacheThrottleTime < time() ) {
-							$request->setSessionData( self::INVALIDATE_CACHE_THROTTLE_SESSION_KEY, time() + self::INVALIDATE_CACHE_THROTTLE );
+						$lockKey = wfSharedMemcKey( 'login-from-helios', 'user-invalidate-cache', 'user', $user->getId() );
+						\WikiaDataAccess::criticalSection( $lockKey, 1000/*ms*/, function () use ( $user ) {
```
kvas-damian: Do I understand it correctly that we don't allow more that one user to automatically log in at a time?

BTW: I can't find handling situation when lock cannot be acquired.

harnash: We do generate user-unique keys so that should be per user restriction only. Am I right?

wladekb: @kvas-damian Lock is per user. Moreover I assume when lock cannot be acquired it means that some other request is doing the same. In such case IMHO we can just pass since the user_touched will be bumped by another process.

kvas-damian: Yeah, I missed that. Makes sense for me.

wladekb:
@kvas-damian @harnash Please take a look at the updated code.

includes/wikia/WikiaDataAccess.class.php
```diff
@@ -302,6 +301,39 @@ static private function getDataAndVerify( $app, $key, $clearLocalCache = false )
...
+	static public function pseudoCriticalSection( $key, $timeout, callable $fn ) {
+		$key .= ':CRITICAL_SECTION';
+		list( $gotLock, $wasLocked ) = self::lock( $key, true, $timeout );
+
+		if ( !$gotLock ) {
+			throw new CannotAcquireLockException('Could not acquire critical section lock for: ' . $key);
+		}
+
+		try {
+			$result = $fn();
```
kvas-damian, line 329: What do you think about re-throwing original exception caused by this callable?

wladekb, line 329: AFAIK this will happen automatically because I have only the "finally" clause.

kvas-damian, line 329: Ohh you're right. I wasn't aware of this syntax ;)

kvas-damian:
LGTM

wladekb:
Fixed in #10520

PR #10511, opened 2016-05-18
JPN-542

perjg:
- Capitalization was incorrect for some strings.
- Added local string for Recent wiki activity string.
- Updated translations from crowdin


lukaszkonieczny:
LGTM

PR #10512, opened 2016-05-18
ADEN-3378 hashing ARecovery CSS URL as memcached key

jacekjursza:
I've added hashing function to memc-key in order to avoid too long keys.


fraszczakszymon:
LGTM 👍

PR #10515, opened 2016-05-18
XW-1225 | Remove obsolete media templates previously used in MercuryApi

rogatty:
## Links
- https://wikia-inc.atlassian.net/browse/XW-1225
- https://github.com/Wikia/config/pull/1777
- https://github.com/Wikia/mercury/pull/2452
## Description

Removed old, unused media templates as a part of clean up after SEO experiment.
## Reviewers

@Wikia/x-wing @Wikia/west-wing


extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
```diff
@@ -8,7 +8,6 @@ class ArticleAsJson extends WikiaService {
 	];
 
 	const CACHE_VERSION = '0.0.3';
-	const CACHE_VERSION_FOR_SEO_FRIENDLY_IMAGES = 1; // for $wgEnableSeoFriendlyImagesForMobile
```
bognix, line 11: should we bump cache?

rogatty, line 11: I don't think so. The code that I removed wasn't used for few weeks now. Do you think it's possible that something is still cached since then?

bognix, line 11: the cache version is going to change. The code that was enabled for few weeks was using the Cache version 1. Now it's going to use version 0.0.3. The question is if it should change to 0.0.3 or maybe we should leave 1. I'm not sure if it's going to break something or not.

rogatty, line 11: You're right, I think it's safer to use `1` as not touching is always safer :wink:

hakubo:
LGTM

dianafa:
+46 −198 👍

PR #10517, opened 2016-05-18
JPN-538 	Add linebreak before wiki name

kamilkoterba:
https://wikia-inc.atlassian.net/browse/JPN-538

Going with simpler approach than one proposed here https://github.com/Wikia/app/pull/10506
Tradeoff is that it breaks the line always that will be confirmed with product owners whether is sufficient or we should proceed with above one.

@Wikia/spitfires @perjg


lukaszkonieczny:
LGTM

PR #10521, opened 2016-05-18
SEO-420 Disable and Remove wgAddNoIndexToFilePages

gabrys:
Reverting an old-old change that allowed us to add noindex,follow robot policy to file pages


saipetch:
LGTM

PR #10524, opened 2016-05-19
SER-212: Add option to return a username back to the caller

wladekb:
This adds an option "retusername" that will post back the username processed by LoginForm.

https://wikia-inc.atlassian.net/browse/SER-218

/cc @sqreek


pawelchcki:
LGTM

PR #10525, opened 2016-05-19
Cake 141 - related content in specific order

nmonterroso:
@Wikia/cake 
https://wikia-inc.atlassian.net/browse/CAKE-141

Ensure that the ordering for content in the related content list is in the order:

```
// fandom
// discussions
// articles
(repeat)
```


drsnyder:
LGTM

PR #10531, opened 2016-05-23
DAT-4228 fix inner template by adding empty lines to marker

idradm:


sebastianmarzjan:
Frankly, I'm unsure about the potential side effects of this change (what will happen if there are already newlines at the start / end of template). Can you describe more in the mark() method?

idradm:
It can cause an empty paragraph to show up, which on mobile has some margin without any height.

PR #10533, opened 2016-05-23
SEO-428 Fix i18n for Paginator

gabrys:
Fixes PHP Warning: include(/extensions/wikia/TopListsi18n/Paginator.i18n.php): failed to open stream: No such file or directory in /includes/LocalisationCache.php on line 461


saipetch:
LGTM

macbre:
Thanks, @gabrys !

gabrys:
We didn't catch this, because the paginator labels ("Next" and "Back") 
are actually hidden using CSS (replaced with arrows).

PR #10540, opened 2016-05-24
CE-3749 Optimize getTopContributors

kamilkoterba:
https://wikia-inc.atlassian.net/browse/CE-3749

Initial PR with new query. 
getTopContributors method broken up to two separate use cases and separate cache keys to have control over keys to purge later.

@Wikia/spitfires @d34th4ck3r


extensions/wikia/CommunityPage/CommunityPageSpecialController.class.php
```diff
@@ -109,8 +109,7 @@ public function getTopContributorsData() {
 	 * @return array
 	 */
 	public function getTopAdminsData() {
-		$allAdmins = $this->getAllAdmins();
-		$allAdminsDetails = $this->getContributorsDetails( $allAdmins );
+		$allAdminsDetails = $this->getContributorsDetails( $this->usersModel->getAllAdmins() );
```
lukaszkonieczny: What do you think about caching Contributors Details instead of results of query?

extensions/wikia/CommunityPage/models/CommunityPageSpecialUsersModel.class.php
```diff
@@ -39,57 +40,74 @@ public function isAdmin( $userId, $admins ) {
...
+		$data = WikiaDataAccess::cache(
+			wfMemcKey( self::ALL_ADMINS_MCACHE_KEY ),
+			WikiaResponse::CACHE_STANDARD,
+			function () {
+				$db = wfGetDB( DB_SLAVE );
+
+				$adminIds = $this->getAdmins();
+				$botIds = $this->getBotIds();
+
+				$dateFilter = 'rev_timestamp > DATE_SUB(now(), INTERVAL 2 YEAR)';
```
lukaszkonieczny: Maybe we can remove this variable and move this condition to query

extensions/wikia/CommunityPage/CommunityPageSpecialController.class.php
kamilkoterba: I'm sceptical as we'll end up caching additional data like username and avatar which will increase number of places where we should purge this. Fortunately user and avatar are already cached, so I think performance gain is not worth the complexity increase here.

Although I'm planning to reduce number of users passed to getContributorsDetails as we don't need to walk through the whole list here.

extensions/wikia/CommunityPage/CommunityPageSpecialController.class.php
```diff
@@ -57,50 +57,28 @@ public function index() {
...
 	 */
 	public function getTopContributorsData() {
-		$userContributionCount = $this->usersModel->getUserContributions( $this->getUser() );
-		$contributors = $this->usersModel->getTopContributors( 50 );
-		// get details for only 5 of the remaining contributors
-		$contributorDetails = $this->getContributorsDetails(
-			array_slice( $contributors, 0, self::TOP_CONTRIBUTORS_LIMIT )
+		$currentUserContributionCount = $this->usersModel->getUserContributions( $this->getUser() );
+		$topContributors = $this->usersModel->getTopContributors();
+		$top5ContributorsDetails = $this->getContributorsDetails(
```
lukaszkonieczny: Maybe we can rethink this variable name because changing limit value in const will require changing var name

lukaszkonieczny:
LGTM

PR #10547, opened 2016-05-25
better related content api usage

nmonterroso:
@Wikia/cake @frankfarmer 
https://wikia-inc.atlassian.net/browse/CAKE-166

Short circuit the API calls when we know there's no related content, and add a timeout to the `curl` call.


extensions/wikia/Recirculation/services/CakeRelatedContentService.php
```diff
@@ -101,4 +111,24 @@ private function formatTitle(Content $content) {
...
+
+		return in_array(
+				$wgCityId,
+				[
+						147,    // starwars
+						130814, // gameofthrones
+						3035,   // fallout
+						2237,   // dc
+						2233,   // marvel
+						208733, // darksouls
```
ArturKlajnerok, line 126: Seems that `darksouls` is not yet imported to production database.

ArturKlajnerok:
After addressing issue mentioned in comment, Lgtm 👍

extensions/wikia/Recirculation/services/CakeRelatedContentService.php
nmonterroso, line 126: Yah, I added it because I see we have a ticket to do it. I figured since we're going to add it it's better to get ahead of that now rather than having to push out another MW release.

Do you think we should remove it and re-add later? I don't think darksouls gets as much traffic as some of the other sites on this list.

drsnyder:
lgtm

PR #10549, opened 2016-05-27
PLATFORM-2220: Automatically shorten memcached keys to be standard-compliant

wladekb:
Recently I found out that in a few cases we are trying to use memcached keys that are too long. This makes some of these data to be never cached.

This change introduces automatic and transparent key shortening by deriving a standard compliant key for the supplied one by cutting X first characters and appending a hash (sha1) of the entire key in such a way that the new key is never too long.

https://wikia-inc.atlassian.net/browse/PLATFORM-2220

/cc @owend @mixth-sense @macbre @kvas-damian

### DIFF
--- a/includes/objectcache/MemcachedClient.php
+++ b/includes/objectcache/MemcachedClient.php
@@ -1419,7 +1419,11 @@ class MWMemcached {
 			if (isset($rev_keys[$key])) {
 				$out[$rev_keys[$key]] = $value;
 			} else {
-				// @todo - it should not happen - log such cases
+				// it should never happen but if it does log it
+				\Wikia\Logger\WikiaLogger::instance()->error( __METHOD__ . ' - key not found', [
+					'missing_key' => $key,
+					'all_keys' => array_keys($rev_keys)
+				] );
 				$out[$key] = $value;
 			}
 		}
--- a/includes/objectcache/MemcachedClient.php
+++ b/includes/objectcache/MemcachedClient.php
@@ -103,6 +103,10 @@ class MWMemcached {
 
 	const MEMCACHED_ITEM_MAX_SIZE = 1048576; // 1MB
 
+	// Wikia change - begin - @author: wladek
+	const MEMCACHED_KEY_MAX_LENGTH = 245; // actual protocol limit = 250
+	// Wikia change - end
+
 	/**
 	 * Command statistics
 	 *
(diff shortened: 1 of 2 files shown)


includes/objectcache/MemcachedClient.php
```diff
@@ -1351,6 +1368,63 @@ function _flush_read_buffer( $f ) {
...
+	 * @param $rev_keys array Reverse key map obtained from call to _encode_keys()
+	 */
+	function _decode_keys( $result, $rev_keys ) {
+		$out = [];
+		foreach ($result as $key => $value) {
+			if (isset($rev_keys[$key])) {
+				$out[$rev_keys[$key]] = $value;
+			} else {
+				// @todo - it should not happen - log such cases
+				$out[$key] = $value;
```
macbre, line 1427: So let's log it :)

macbre:
:+1:

lib/Wikia/src/Logger/WikiaLogger.php
```diff
@@ -126,27 +126,38 @@ public function notice($message, Array $context=[]) {
...
 	}
 
 	public function log($level, $message, Array $context=[]) {} // NOOP
 
+	private function addStacktraceIfMissing( array &$context ) {
+		if ( !array_key_exists( 'exception', $context ) ) {
+			$context['exception'] = new \Exception();
+		}
+	}
+
```
macbre, line 160: :+1:

includes/objectcache/MemcachedClient.php
```diff
@@ -1351,6 +1368,67 @@ function _flush_read_buffer( $f ) {
...
+			$this->_encode_key( $key[1] );
+
+			return;
+		}
+
+		$key_length = strlen( $key );
+		if ( $key_length > self::MEMCACHED_KEY_MAX_LENGTH ) {
+			$suffix = '--HASH--' . sha1( $key );
+			$suffix_length = strlen( $suffix );
+			$key = substr( $key, 0, self::MEMCACHED_KEY_MAX_LENGTH - $suffix_length ) . $suffix;
```
ghost, line 1388: Will that meet the 250 character protocol constraint? Plus, it is 250 characters or 250 bytes?

lib/Wikia/src/Logger/WikiaLogger.php
ghost, line 160: Are you should it will not kill us in terms of Elasticsearch diskspace?

PR #10550, opened 2016-05-27
PLATFORM-2198: Strip domain from http_url_path log attribute

wladekb:
We are getting inconsistent values for http_url_path in the logs. This change ensures that we do not report domain in this attribute.

https://wikia-inc.atlassian.net/browse/PLATFORM-2198

/cc @macbre @mixth-sense


macbre:
:+1:

PR #10551, opened 2016-05-27
Improve rail styling on Community Page and in RightRail in general

kamilkoterba:
Spotted that we could do better with styling right rail on small widths.

@Wikia/spitfires @vforge


vforge:
LGTM

PR #10555, opened 2016-05-30
Mail_mimePart: PHP7 compatibility

macbre:
`Deprecated: Methods with the same name as their class will not be constructors in a future version of PHP; Mail_mimePart has a deprecated constructor`

@wladekb


wladekb:
lgtm

PR #10558, opened 2016-05-30
DataFeedProvider cleanup

wladekb:
https://wikia-inc.atlassian.net/browse/PLATFORM-2219

/cc @macbre


includes/wikia/GlobalFunctions.php
```diff
@@ -1327,40 +1327,31 @@ function endsWith( $haystack, $needle, $case = true ) {
...
+		return array( $input, $funcs );
+	} else {
+		$input_json = json_encode( $input );
+		foreach ( $funcs as $key => $value ) {
+			$input_json = str_replace( '"' . $key . '"', $value, $input_json );
+		}
+
+		return $input_json;
+	}
+}
```
macbre: This function is not used anywhere. Worth removing as we're here?

macbre:
:+1:

PR #10559, opened 2016-05-30
XW-1478 | Don't render infobox hero module on Mercury

Warkot:
…as it is now rendered in the article header.

⚠️  Release one release before https://github.com/Wikia/mercury/pull/2488 —app change should be on our servers and in cache before the removal of CSS rules hiding the hero image and title. One downside of this is that if an infobox has two or more `pi-title`s in it, the second will be hidden.

https://wikia-inc.atlassian.net/browse/XW-1478

@Wikia/x-wing @Wikia/west-wing

### DIFF
--- a/extensions/wikia/PortableInfobox/services/PortableInfoboxRenderService.class.php
+++ b/extensions/wikia/PortableInfobox/services/PortableInfoboxRenderService.class.php
@@ -54,6 +54,7 @@ class PortableInfoboxRenderService extends WikiaService {
 		wfProfileIn( __METHOD__ );
 
 		$helper = new PortableInfoboxRenderServiceHelper();
+		$dataBag = \Wikia\PortableInfobox\Helpers\PortableInfoboxDataBag::getInstance();
 		$infoboxHtmlContent = '';
 		$heroData = [ ];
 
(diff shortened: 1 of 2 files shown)


extensions/wikia/PortableInfobox/services/PortableInfoboxRenderService.class.php
```diff
@@ -158,22 +160,26 @@ private function renderGroup( $groupData ) {
...
+				return $this->renderItem( 'hero-mobile', $data );
 			}
-		} else {
-			$markup = $this->renderItem( 'title', $data[ 'title' ] );
+		} elseif ( !$helper->isMercury() || $firstInfoboxAlredyRendered ) {
+			return $this->renderItem( 'title', $data[ 'title' ] );
 		}
 
-		return $markup;
+		return '';
```
dianafa, line 182: i'd prefer not to make early returns in 3 places and instead return markup at the end of function :)

dianafa:
One remark about early returns and except that LGTM

extensions/wikia/PortableInfobox/services/PortableInfoboxRenderService.class.php
bognix, line 182: And I would prefer early return because it makes it easier for me to follow the logic - I read less number of lines.

Warkot, line 182: This variable could be useful if anything would be additionally done to the `$markup` just before returning it.

tomasznapieralski:
LGTM, tested

PR #10560, opened 2016-05-31
ADEN-3401 Unblock detection resources

fraszczakszymon:


extensions/wikia/AdEngine/ResourceLoaders/AdEngine2Resource.class.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+class AdEngine2Resource {
```
jacekjursza, line 3: 👍

extensions/wikia/AdEngine/ResourceLoaders/AdEngine2Resource.class.php
```diff
@@ -0,0 +1,30 @@
...
+	const RESOURCE_DATE_FORMAT = 'Y-m-d';
+
+	static public function register($name, $class) {
```
jacekjursza: I'm not a fan, but I think we should this:
https://www.mediawiki.org/wiki/Manual:Coding_conventions/PHP
as coding convention for all our PHP code

jacekjursza:
LGTM, but please fix code formatting

PR #10561, opened 2016-05-31
don't escape title - it is escaped by default in mustache

hakubo:
I have also improved whitespaces a bit

### DIFF
--- a/extensions/wikia/Chat2/js/controllers/ChatBanModalLogs.js
+++ b/extensions/wikia/Chat2/js/controllers/ChatBanModalLogs.js
@@ -1,29 +1,29 @@
-var ChatBanLogController = function() {
-    $('.chat-change-ban').click(function(e) {
-        e.preventDefault();
-		var title = mw.html.escape($.msg('chat-ban-modal-change-ban-heading')),
-            userId = $(this).data('user-id'),
-			okCallback = function(expires, reason) {
-			 	$.post(wgScript + '?action=ajax&rs=ChatAjax&method=blockOrBanChat' ,{
+var ChatBanLogController = function () {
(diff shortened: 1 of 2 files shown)


extensions/wikia/Chat2/js/controllers/ChatBanModalLogs.js
```diff
@@ -1,29 +1,29 @@
...
+			okCallback = function (expires, reason) {
+				$.post(wgScript + '?action=ajax&rs=ChatAjax&method=blockOrBanChat', {
 						userToBanId: userId,
 						time: expires,
 						reason: reason,
 						mode: 'global',
 						token: mw.user.tokens.get('editToken')
 					},
-					$.proxy(function(data) {
+					function () {
```
ghost, line 13: What is the benefit of not using the `proxy` function?

ghost:
Just one question for the sake of curiosity. Otherwise, LG, tej. 👍

extensions/wikia/Chat2/js/controllers/ChatBanModalLogs.js
hakubo, line 13: I would flip it :)
What is the benefit of USING the proxy function here?

hakubo, line 13: In addition I think if you use $.proxy with no second parameter
you bind 'this' to undefined

hakubo, line 13: nope it binds this to Window object that is available there anyway

pawelchcki:
LGTM

ludwikkazmierczak:
Tested and accepted

PR #10565, opened 2016-05-31
Add ability to override solr host for KV queries

pawelchcki:
This allows solr hostname used in KV queries (outside special:search) to be overriden separately.

### DIFF
--- a/includes/wikia/services/ArticleService.class.php
+++ b/includes/wikia/services/ArticleService.class.php
@@ -274,6 +274,8 @@ class ArticleService extends WikiaObject {
 	public function getTextFromSolr()
 	{
 		global $wgSolrHost, $wgSolrKvHost;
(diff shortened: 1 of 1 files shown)


ghost:
Since `$wgSolrHost` and `$wgSolrKvHost` are used in instance methods, I would use only one `global` in `__construct` and assign them to `protected` members. It might help us with future development and mitigation of the problem - when we want to extend the class or deal with multiple hosts.

includes/wikia/services/ArticleService.class.php
```diff
@@ -45,13 +47,21 @@ class ArticleService extends WikiaObject {
 	 */
 	protected $solrDocumentService;
 
+	private $solrHostname;
```
ghost, line 50: `protected` perhaps, as above?

pawelchcki, line 50: tbh I don't like protected variables :dancer:

PR #10568, opened 2016-05-31
Introduce CURLOPT_TIMEOUT of 5s to Solr client

pawelchcki:
Default TIMEOUT is indefinite  - 5 seconds is the max request time suggestion we pass to Solr (but its only suggestion so solr requests can last longer - a lot longer)

/cc: @owend @idradm @mixth-sense @ljagiello


ghost:
👍 LGTM

ljagiello:
LGTM

PR #10572, opened 2016-05-31
Cake 177 - contribution prototype integration

nmonterroso:
@Wikia/cake 
https://wikia-inc.atlassian.net/browse/CAKE-177
See Wikia/config#1791

Render CP articles in MW when viewing articles instead of rendering MW articles.


extensions/wikia/ContributionPrototype/CPArticleRenderer.php
```diff
@@ -0,0 +1,91 @@
...
+				'/<script src=("|\')\/(.*?)(\1)>/',
+				'<script src=$1'.$this->host.'/$2$1>',
+				$matches[1]);
+		$output->addHTML($body);
+	}
+
+	private function getArticleContent($title) {
+		return file_get_contents("{$this->host}/wiki/lol/wiki/Home");
+
+//		$internalHost = $this->urlProvider->getUrl(self::SERVICE_NAME);
```
nmonterroso, line 76: For now this is all commented out while the CP side of things gets ironed out. For now, we get the article content externally since that allows us to get it from a locally running CP instance.

nmonterroso:
~~oops, gotta fix somethign before merging~~
k

extensions/wikia/ContributionPrototype/ContributionPrototype.setup.php
```diff
@@ -0,0 +1,16 @@
...
+	}
+
+	$file = __DIR__.'/'.str_replace('\\', '/', substr($class, $len)).'.php';
+	if (file_exists($file)) {
+		require $file;
+	}
+});
+
+// take over actions
+$wgActions['view'] = ContributionPrototype\CPViewAction::class;
```
ArturKlajnerok, line 16: Is `$wgActions['view']` responsible for article rendering ?

nmonterroso, line 16: Yah. There are several actions. `view` refers to the action of article viewing, with the default being `null` which means to use `ViewAction`. We're hijacking that :)

ArturKlajnerok:
After looking at `Action` and `FormlessAction` seems like we are replacing article rendering with the PoC article, but the MW still will load or at leas try to load whole article and execute it's logic on it. So if this is fully intentional and we want to support only rendering inside MW, Lgtm 👍

nmonterroso:
Yah, we only want it to be visibly functional at this point. That's a good point, though. I knew this was happening but it didn't jump out to me as a problem. We can file that as "let's worry about that when we want to turn it on"

extensions/wikia/ContributionPrototype/CPArticleRenderer.php
```diff
@@ -0,0 +1,98 @@
...
+		// this ends up using $wgOut, but we need it for the assets manager integration, no point in duplicating :(
+		Wikia::addAssetsToOutput('contribution_prototype_scss');
+	}
+
+	private function addScripts(OutputPage $output) {
+		$output->addScript("<script src=\"{$this->publicHost}/public/assets/app.js\"></script>");
+	}
+
+	private function getArticleContent($title) {
+//		$internalHost = $this->urlProvider->getUrl(self::SERVICE_NAME);
```
drsnyder, line 76: Should this comment be removed?

nmonterroso, line 76: I intentionally left this in, because we do want to use the `urlProvider` here to get an internal url. For now there's only one instance of the CP in dev, so we can't use the url provider (since that's environment aware). When we launch a prod instance we will change this to use the internal url provided by the `UrlProvider`

PR #10580, opened 2016-06-01
CreateNewWiki / ImportStarterDataTest: assert bz2 PHP extension is there

macbre:
[PLATFORM-2224](https://wikia-inc.atlassian.net/browse/PLATFORM-2224)

Wuthout `bz2` extension we cannot decompress XML dumps.

```
Maintenance script ImportStarter was interrupted by unhandled exception.

PHP Warning: stream_filter_append(): unable to locate filter "bzip2.decompress" in /usr/wikia/slot1/11735/src/extensions/wikia/CreateNewWiki/maintenance/importStarter.php on line 59
```

@mixth-sense / @artursitarski

### DIFF
--- a/extensions/wikia/CreateNewWiki/tests/tasks/ImportStarterDataTest.php
+++ b/extensions/wikia/CreateNewWiki/tests/tasks/ImportStarterDataTest.php
@@ -19,6 +19,14 @@ class ImportStarterDataTest extends \WikiaBaseTest {
 	}
 
 	/**
+	 * @see PLATFORM-2224
+	 * @group Infrastructure
+	 */
+	public function testDependencies() {
+		$this->assertTrue( extension_loaded( 'bz2' ), 'bz2 extension should be loaded for XML dumps decompression' );
+	}
+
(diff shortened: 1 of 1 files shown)


ghost:
Do you think, a wise elder, that an assertion on the extension setup level would be a good thing? A.K.A. poor man's dependency handling :-)

macbre:
Honestly, can't think of a better place

 We can introduce a separate test class that will assert the presence of all required extensions (believe me, it's pretty easy to skip one or two and spend several hours debugging just to find out that it's missing extension to be blamed). But I believe it's better to assert dependency where it's really required.

ghost:
Agreed. It also depends on the case. When using external authentication mechanism, it's safe to assert that making HTTP requests is possible and terminate immediately if necessary. Dead program does less damage that a crippled one. We need to decide, when a starter import should be terminated and when it should be continued. Can we at least fail with a fatal error rather than sneak through a warning?

ghost:
LGTM :+1:

PR #10581, opened 2016-06-01
don't encode wgUserName twice

hakubo:
It appears that when we upgraded socket.io it got broken
as it does encode query for us

### DIFF
--- a/extensions/wikia/Chat2/js/controllers/controllers.js
+++ b/extensions/wikia/Chat2/js/controllers/controllers.js
@@ -41,28 +41,26 @@ var NodeChatSocketWrapper = $.createClass(Observable, {
 		}
 		this.authRequestWithMW(function (data) {
 			var socket = io.connect(url, {
-				'force new connection': true,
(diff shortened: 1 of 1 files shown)


extensions/wikia/Chat2/js/controllers/controllers.js
```diff
@@ -81,12 +79,7 @@ var NodeChatSocketWrapper = $.createClass(Observable, {
 	},
 
 	authRequestWithMW: function (callback) {
-		//hacky fix of fb#19714
-		//it seems socket.io decodes it -- that's why I double encoded it
-		//but maybe we should implement here authorization via user id instead of username?
-		var encodedWgUserName = encodeURIComponent(encodeURIComponent(wgUserName));
```
ghost, line 87: :heart: - thirty four levels more and we would have a hashing mechanism :-)

ghost:
:+1: LGTM

ludwikkazmierczak:
Tested and Accepted

PR #10584, opened 2016-06-01
Changes Modil to play nicer with the AMD standard, while still issuing a warning for sub-ideal practices.

gbenson-wikia:
Specifically, the AMD standard permits the first argument to be omitted.  These changes update Modil to permit this case, and emits a console warning when encountered (instead of throwing an Error).

### DIFF
--- a/resources/wikia/libraries/modil/modil.js
+++ b/resources/wikia/libraries/modil/modil.js
@@ -135,6 +135,8 @@
 	 * @throws {Error} If dependencies is not undefined but not an array
 	 */
 	define = function (id, dependencies, definition, defMock) {
+		var warning;
+
 		if (typeof id !== strType) {
 			// Slide all the args over by one to make room for the id
 			defMock = definition;
@@ -145,7 +147,7 @@
 			id = 'module_' + (new Date()).getTime() + '_' + Math.random();
 
 			// Let the developer know there's a potential problem
-			var warning = "Module id missing or not a string; assigning temporary id (" + id + "). "
+			warning = "Module id missing or not a string; assigning temporary id (" + id + "). "
 				+ (new Error().stack||'').replace(/\n/g, ' / ');
 
 			if (console.warn) {
--- a/resources/wikia/libraries/modil/modil.js
+++ b/resources/wikia/libraries/modil/modil.js
@@ -121,17 +121,18 @@
 	 *
 	 * @public
 	 *
+	 * @example define(function () { return {hello: 'World'}; });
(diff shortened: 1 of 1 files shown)


gbenson-wikia:
Pinging @hakubo and @ludwikkazmierczak for visibility on this change.  In theory, this change should be invisible.  In practice, it may cause:
- ad display problems
- other libraries which previously died silently to actually start working now, which would potentially have unintended side effects

resources/wikia/libraries/modil/modil.js
```diff
@@ -121,21 +121,38 @@
...
+			dependencies = id;
+
+			// Give it a temporary module name
+			id = 'module_' + (new Date()).getTime() + '_' + Math.random();
+
+			// Let the developer know there's a potential problem
+			var warning = "Module id missing or not a string; assigning temporary id (" + id + "). "
+				+ (new Error().stack||'').replace(/\n/g, ' / ');
+
+			if( console.warn ) {
```
lizlux: ws

resources/wikia/libraries/modil/modil.js
```diff
@@ -121,21 +121,38 @@
 	 *
 	 * @public
 	 *
+	 * @example define(function () { return {hello: 'World'}; });
+     * @example define(['dep1', 'dep2'], function (dep1, dep2) { ... });
```
lizlux: ws

lizlux:
Only wondering why, if it's this simple, no one has done this yet. In other words, what's the catch? I can't think of any so LGTM.

resources/wikia/libraries/modil/modil.js
```diff
@@ -121,21 +121,38 @@
...
+			// Slide all the args over by one to make room for the id
+			defMock = definition;
+			definition = dependencies;
+			dependencies = id;
+
+			// Give it a temporary module name
+			id = 'module_' + (new Date()).getTime() + '_' + Math.random();
+
+			// Let the developer know there's a potential problem
+			var warning = "Module id missing or not a string; assigning temporary id (" + id + "). "
```
hakubo: var should be at the top of a function

hakubo:
The catch is - this is plain wrong :) as you won't be able to import it anywhere
But if you don't need that it might a valid workaround.

gbenson-wikia:
Heh.  Well, you couldn't import it before since Modil would reject it
outright, so there's no functional change there.  ;)  In my specific case,
the important thing is that the code inside the module gets executed
without error, which it now does.

PR #10586, opened 2016-06-02
CSRFDetector: introduce markHttpMethodAccepted method to improve handling of "internal" GET requests

macbre:
In some cases we consider GET requests to be secure when they're provided with a secret token (see [PLATFORM-2207](https://wikia-inc.atlassian.net/browse/PLATFORM-2207) for an example) - simply call `\Wikia\Security\CSRFDetector::markHttpMethodAccepted` to mark such requests as secure ones and avoid reporting CSRF errors for them.

@wladekb / @Wikia/sustaining-engineering-team (per code ownership)

### DIFF
--- a/extensions/wikia/Chat2/ChatAjax.class.php
+++ b/extensions/wikia/Chat2/ChatAjax.class.php
@@ -262,6 +262,8 @@ class ChatAjax {
 	private static function authenticateServer() {
 		global $wgRequest;
 
+		\Wikia\Security\CSRFDetector::markRequestAsSecure( __METHOD__ );
+
 		return \Wikia\Security\Utils::matchToken( ChatConfig::getSecretToken(), $wgRequest->getVal( 'token' ) );
 	}
 
--- a/extensions/wikia/Chat2/ChatAjax.class.php
+++ b/extensions/wikia/Chat2/ChatAjax.class.php
@@ -257,14 +257,20 @@ class ChatAjax {
 	}
 
 	/**
+	 * When token matches accept internal GET requests, i.e. do not report them as CSRF errors (PLATFORM-2207)
+	 *
 	 * @return bool
 	 */
 	private static function authenticateServer() {
 		global $wgRequest;
 
-		\Wikia\Security\CSRFDetector::markHttpMethodAccepted( __METHOD__ );
+		$tokenMatched = \Wikia\Security\Utils::matchToken( ChatConfig::getSecretToken(), $wgRequest->getVal( 'token' ) );
 
-		return \Wikia\Security\Utils::matchToken( ChatConfig::getSecretToken(), $wgRequest->getVal( 'token' ) );
+		if ( $tokenMatched ) {
+			\Wikia\Security\CSRFDetector::markHttpMethodAccepted( __METHOD__ );
+		}
+
+		return $tokenMatched;
 	}
 
 	private static function authenticateServerOrUser() {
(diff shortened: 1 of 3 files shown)


extensions/wikia/Chat2/ChatAjax.class.php
```diff
@@ -262,6 +262,8 @@ static public function giveChatMod() {
 	private static function authenticateServer() {
 		global $wgRequest;
 
+		\Wikia\Security\CSRFDetector::markRequestAsSecure( __METHOD__ );
```
wladekb: I'd rather go with sth like:

```
$result = \Wikia\Security\Utils::matchToken( ChatConfig::getSecretToken(), $wgRequest->getVal( 'token' ) );
if ($result) {
  \Wikia\Security\CSRFDetector::markRequestAsSecure( __METHOD__ );
}
return $result
```

so that is the token does not match we don't mark the request as secure

extensions/wikia/Security/classes/CSRFDetector.class.php
```diff
@@ -36,6 +36,18 @@ public static function onRequestWasPosted() {
 	}
 
 	/**
+	 * In some cases we consider GET requests to be secure when they're provided with a secret token
+	 *
+	 * @see PLATFORM-2207
+	 */
+	public static function markRequestAsSecure( $caller ) {
```
wladekb: The name of the method is misleading in my opinion. It should either mark all internal checks as passed or be renamed to describe that it only marks the HTTP method to be acceptable

extensions/wikia/Chat2/ChatAjax.class.php
macbre: `CSRFDetector` checks two conditions:
- the token is checked
- the HTTP method is checked

`markRequestAsSecure` marks only the second condition as true - the token needs to be checked to avoid reporting CSRF error. And if the token does not match, then the different condition will avoid performing certain actions.

extensions/wikia/Security/classes/CSRFDetector.class.php
macbre: I always have problems with a good naming :) `markHttpMethodAccepted` ?

wladekb: sounds much less misleading, got for it!

extensions/wikia/Chat2/ChatAjax.class.php
wladekb: this is used in a following way:
`authenticateServer() or authenticateUser()`
which means that if it is a request from user in this specific code we won't actually check if the method was correct because it will be flagged as accepted while doing server authentication check
so I still think we should not mark method as accepted here unconditionally

extensions/wikia/Security/classes/CSRFDetector.class.php
macbre: Fixed in d1e0aa2

extensions/wikia/Chat2/ChatAjax.class.php
macbre: Ah,good catch! I'll fix it!

macbre: caa2662

wladekb:
I'm sorry for this... but I just came up with even better idea :)
Let's create `matchSecretToken` method that would mark entire request as secure if and only if the token is matched, sound reasonable?

macbre:
It does, no worries - better late than never :)

wladekb:
lgtm

ghost:
Thanks!

PR #10587, opened 2016-06-02
CloseMyAccountSpecialController: call to User::newFromConfirmationCode() is basically a token check

macbre:
[PLATFORM-2206](https://wikia-inc.atlassian.net/browse/PLATFORM-2206)

Tell `CSRFDetector` that we're fine here

@wladekb

### DIFF
--- a/extensions/wikia/CloseMyAccount/CloseMyAccountSpecialController.class.php
+++ b/extensions/wikia/CloseMyAccount/CloseMyAccountSpecialController.class.php
@@ -174,6 +174,10 @@ class CloseMyAccountSpecialController extends WikiaSpecialPageController {
 			return;
 		}
 
+		// call to User::newFromConfirmationCode() is basically a token check
+		// tell CSRFDetector that we're fine here (PLATFORM-2206)
+		\Wikia\Security\CSRFDetector::onUserMatchEditToken();
+
 		$helper->reactivateAccount( $user );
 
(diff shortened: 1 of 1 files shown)


extensions/wikia/CloseMyAccount/CloseMyAccountSpecialController.class.php
```diff
@@ -174,6 +174,10 @@ public function reactivate() {
 			return;
 		}
 
+		// call to User::newFromConfirmationCode() is basically a token check
+		// tell CSRFDetector that we're fine here (PLATFORM-2206)
+		\Wikia\Security\CSRFDetector::onUserMatchEditToken();
```
wladekb, line 179: I'd create a separate method with better name even if it would call internally this one only. Worth changing?

macbre, line 179: It's the first case when we need to call this method outside of .`CSRFDetector` Let's wait for a second one.  Sounds fine?

wladekb, line 179: I'd consider `onUserMatchEditToken()` (or hook handlers in general) private even though they are technically public. Having that assumption in mind we're basically creating a new public entry point and I'd still strive to find a more intuitive name for it (eg. `markEditTokenChecked` or something similar)

PR #10596, opened 2016-06-02
RECIRC 99: Create large impact footer experiment

pauloslund:
https://wikia-inc.atlassian.net/browse/RECIRC-99. Experiment to be run at 2% initially

screen shot 2016-06-03 at 1 40 54 pm


extensions/wikia/Recirculation/Recirculation.setup.php
```diff
@@ -30,4 +30,11 @@
 	'recirculation-incontent-title',
 	'recirculation-fandom-title',
 	'recirculation-fandom-subtitle',
+	'recirculation-impact-footer-featured-fandom-subtitle',
+	'recirculation-impact-footer-trending-tag',
+	'recirculation-impact-footer-wiki-tag',
+	'recirculation-discussions-new',
+	'recirculation-discussions-upvotes',
+	'recirculation-discussions-replies',
+	'recirculation-discussions-posts',
```
gabrys: I think you can do `recirculation-*` if that helps

extensions/wikia/Recirculation/RecirculationApiController.class.php
```diff
@@ -35,4 +35,38 @@ public function getCakeRelatedContent() {
...
+		}
+
+		$articleId = $this->request->getVal( 'articleId' );
+		$articles = $this->app->sendRequest( 'ArticlesApi', 'getTop', [
+			'abstract' => 0,
+			'expand' => 1,
+			'height' => 220,
+			'limit' => 8,
+			'namespaces' => 0,
+			'width' => 385,
```
gabrys, line 61: Why do we need to pass width and height? Are those for thumbnail sizes?

pauloslund, line 61: Yeah, it's for thumbnail sizes. These numbers specifically are the sizes required for the largest screen sizes and then the images get scaled down smaller for smaller screens.

gabrys:
LGTM

PR #10598, opened 2016-06-02
SEO-172 Only include "active" namespaces in the sitemap

gabrys:
Changing the way the namespaces to put to sitemap are selected.
## Old way

Query the database with this (not 100% precise):

```
SELECT page_namespace FROM page ORDER BY page_namespace GROUP BY page_namespace;
```

This gives as the namespaces that have at least one article in them.

Then we iterate through that list and remove the namespaces we don't want (talk pages, some excluded namespaces).

The problem with this approach is sometimes there are some pages in the database in no longer existing namespaces. Example: http://halo.wikia.com/index.php?curid=151216 (NS: 202).
## New way

Read the namespace IDs from `$wgContLang` object, filter the ones we don't want (talk pages, some excluded namespaces), then query the database to verify which of them actually have any pages:

```
SELECT page_namespace FROM page WHERE page_namespace IN (0, 6, 14, 500, 2000) ORDER BY page_namespace GROUP BY page_namespace;
```

I also cleaned up the code a bit, so it no longer includes non-used features.


extensions/wikia/Sitemap/SpecialSitemap_body.php
```diff
@@ -246,7 +231,6 @@ private function generateNamespace( Array $sitemapIndex = null, $forceUpdate = f
 			$namespaceSitemap = $this->generateNamespaceFromDb();
 			if( !empty($namespaceSitemap) ) {
 				// this sitemap was pre-generated by the maintenance script and stored in db
-				wfProfileOut( __METHOD__ );
```
saipetch, line 249: removing wfProfileIn() line 228 and wfProfileOut() in line 306 in generateNamespace() function ?

extensions/wikia/Sitemap/SpecialSitemap_body.php
```diff
@@ -272,6 +256,8 @@ private function generateNamespace( Array $sitemapIndex = null, $forceUpdate = f
 				$index[ $this->mNamespace ] [ $this->mPage ][ "end" ]
 			);
 		}
+		$scope['page_is_redirect'] = false;
```
saipetch: adding page_is_redirect in default value line 247?

saipetch:
LGTM

PR #10600, opened 2016-06-03
JPN-555 Don't show users with zero contribs in top contributors module

perjg:
The query for getting top contributors did not take into account that entries can show up in the user properties table with zero entries (happens when a user logs in).


perjg:
@d34th4ck3r Please have a look

d34th4ck3r:
LGTM

PR #10601, opened 2016-06-03
JPN-556 Use smaller font size for community page heading in Japanese

perjg:
Also Parse community page header as wikitext.

When you set your language preference to Japanese and load a community page, the welcome text should show in 24px font size, for any other language it should be 36px.


perjg:
@d34th4ck3r Please have a look

lukaszkonieczny:
LGTM

PR #10606, opened 2016-06-03
ADEN-3399 BFAA on mobile

fraszczakszymon:


extensions/wikia/AdEngine/js/AdEngine2.js
```diff
@@ -118,6 +118,7 @@ define('ext.wikia.adEngine.adEngine', [
 				slot = createSlot(queuedSlot, container, {
 					success: function (adInfo) {
 						log(['success', provider.name, slotName, adInfo], 'debug', logGroup);
+						slotTweaker.show(slotName);
```
nandy-andy, line 121: It even makes more sense now here while reading the code ;-)

extensions/wikia/AdEngine/js/provider/directGptMobile.js
```diff
@@ -10,7 +10,7 @@ define('ext.wikia.adEngine.provider.directGptMobile', [
 		'mobile',
 		{
 			INVISIBLE_HIGH_IMPACT:      {size: '1x1'},
-			MOBILE_TOP_LEADERBOARD:     {size: '320x50,320x100,300x250,300x50,1x1'},
+			MOBILE_TOP_LEADERBOARD:     {size: '320x50,320x100,300x250,300x50,320x480,1x1'},
```
nandy-andy: Weren't we supposed to remove `1x1` or we did it in the Big Fancy Ad Below the fold story? ;)

extensions/wikia/AdEngine/js/template/bfaa.js
```diff
@@ -1,48 +1,89 @@
...
+			page.classList.add('bfaa-template');
 
-		nav.style.top = '';
-		page.classList.add('bfaa-template');
-		wrapper.style.background = backgroundColor;
+			adContext.addCallback(function () {
+				page.classList.remove('bfaa-template');
+				wrapper.style.background = '';
+				page.style.paddingTop = '';
+				win.Mercury.Modules.Ads.getInstance().setSiteHeadOffset(0);
```
nandy-andy: @rogatty does the instance returned here by Mercury is always the same? We're not sure with @fraszczakszymon and for example for readability it'd be nice to assign it to a variable and then re-use it here and at the beginning of `mobileHandler.show()`. What do you think?

extensions/wikia/AdEngine/js/provider/directGptMobile.js
fraszczakszymon: It will be resolved in conflict :)

extensions/wikia/AdEngine/js/template/bfaa.js
```diff
@@ -1,48 +1,89 @@
...
+			case 'mercury':
+				page = doc.getElementsByClassName('application-wrapper')[0];
+				wrapper = doc.getElementsByClassName('mobile-top-leaderboard')[0];
+				handler = mobileHandler;
+				break;
+			default:
+				return log(['show', 'not supported skin'], 'info', logGroup);
 		}
 
+		handler.show(height, backgroundColor);
```
nandy-andy, line 91: If somehow the `handler` will be `undefined` JS will throw an error. On the other hand, we break (with `return`) the flow when skin does not match `'oasis'` or `'mercury'`, so all the other cases we want to be "loud", don't we? :)

extensions/wikia/AdEngine/js/template/bfaa.js
rogatty: `getInstance()` is defined here:
https://github.com/Wikia/mercury/blob/dev/front/common/modules/ads.js#L54

`Ads.instance` is defined here:
https://github.com/Wikia/mercury/blob/dev/front/common/modules/ads.js#L309

"Static properties" in ES6 are described here:
http://exploringjs.com/es6/ch_classes.html#_static-data-properties

So I would say that yes, instance returned here is always the same object :) Moving it to a variable should work without any issues.

extensions/wikia/AdEngine/js/template/bfaa.js
fraszczakszymon, line 91: Yup, it shouldn't be undefined and it's implemented in the same way like old modal = we have log for unsupported skin and it doesn't break anything.

nandy-andy:
LGTM :+1:

PR #10611, opened 2016-06-03
SUS-537 Always load new data using ajax for chat widget module

pawelchcki:
This change enables Chat widget to always load the data.
Before this change for module embedded in the article - the behavior was to rely on data stored in article cache so refreshing list of users required purging the article.

The data is loaded from http://sandbox-s4.community.wikia.com/wikia.php?controller=ChatRail&method=GetUsers&format=json

Which also has 2 levels of cache and the authoritative data is in memcache too.
- the data is provided asynchronously via setChatters call via chat service
- caches need to be properly purged for the module to show the list of active users

So for the chat module to show actual users list it needs to be set by chat service before hand.

This pr only fixes most immediate cause of this issue. There are still some edge cases that e.g. MW calling chat service for a list of users would fix;
e.g. if for some reason chat users list is empty but there are users in chat - module will show empty list until a time when any new users joins in.

I've considered removing cache that uses key self::getUsersInfoMemcKey() that will remove one layer of unnecessary cache. But I have @mixth-sense(mixed) feelings about this


ghost:
LGTM :+1:

PR #10615, opened 2016-06-03
updating service name

nmonterroso:
@Wikia/cake 
https://wikia-inc.atlassian.net/browse/CAKE-211

updating the related content api to use the new service name


ArturKlajnerok:
Lgtm 👍

PR #10622, opened 2016-06-06
PLATFORM-2226 Related Pages performance improvements and cleanup

macbre:
[PLATFORM-2226](https://wikia-inc.atlassian.net/browse/PLATFORM-2226)

Improve the way `RelatedPages::getPagesForCategories` method caches query results:
- the previous version of the code used only the article ID to generate the memcache key
- the query performed was quite heavy and the filtering was performed inside the query
- get rid of subquery (as suggested and implemented by @drozdo)

This code introduces caching by the list of categories to get the best matching articles. The current article (i.e. the one we get related pages for) is filtered out from the results (got from either database or memcache) on app level. The cached results can then be reused by other API calls.

Expected results: improve hit ratio and hence less heavy DB queries from `RelatedPages`.

Queries before:

``` sql
SELECT page_id, count(*) c FROM ( SELECT  cl_from AS page_id  FROM `categorylinks` JOIN `page` ON (
(page_id = cl_from AND page_namespace = 0))  WHERE (cl_to IN ( 'Gwara','Tradycja','Kuchnia_regionalna','Potrawy','Gzik' ))  LIMIT 100000   ) i WHERE page_id != 1963 GROUP BY page_id ORDER BY c desc LIMIT 20

+----+-------------+---------------+--------+---------------------------------+--------------+---------+--------------------------------+------+----------------------------------------------+
| id | select_type | table         | type   | possible_keys                   | key          | key_len | ref                            | rows | Extra                                        |
+----+-------------+---------------+--------+---------------------------------+--------------+---------+--------------------------------+------+----------------------------------------------+
|  1 | PRIMARY     | <derived2>    | ALL    | NULL                            | NULL         | NULL    | NULL                           |  111 | Using where; Using temporary; Using filesort |
|  2 | DERIVED     | categorylinks | range  | cl_from,cl_timestamp,cl_sortkey | cl_timestamp | 257     | NULL                           |  111 | Using where; Using index                     |
|  2 | DERIVED     | page          | eq_ref | PRIMARY,name_title              | PRIMARY      | 4       | plpoznan.categorylinks.cl_from |    1 | Using where                                  |
+----+-------------+---------------+--------+---------------------------------+--------------+---------+--------------------------------+------+----------------------------------------------+
```

and after (thanks @drozdo):

``` sql
SELECT /* RelatedPages::getPagesForCategories Macbre - 40c99228-84d0-4899-9139-8dac747b5fc5 */  count(page_id) AS c,page_id,page_namespace,page_title,page_is_redirect,page_len,page_latest  FROM `categorylinks` JOIN `page` ON ((page_id = cl_from) AND (page_namespace = 0))  WHERE cl_to IN ('Gwara','Gzik','Kuchnia_regionalna','Potrawy','Tradycja')  AND page_is_redirect = '0'  GROUP BY page.page_id ORDER BY c DESC LIMIT 11;

+----+-------------+---------------+--------+---------------------------------------------------------------------+--------------+---------+--------------------------------+------+-----------------------------------------------------------+
| id | select_type | table         | type   | possible_keys                                                       | key          | key_len | ref                            | rows | Extra                                                     |
+----+-------------+---------------+--------+---------------------------------------------------------------------+--------------+---------+--------------------------------+------+-----------------------------------------------------------+
|  1 | SIMPLE      | categorylinks | range  | cl_from,cl_timestamp,cl_sortkey                                     | cl_timestamp | 257     | NULL                           |  111 | Using where; Using index; Using temporary; Using filesort |
|  1 | SIMPLE      | page          | eq_ref | PRIMARY,name_title,page_random,page_len,page_redirect_namespace_len | PRIMARY      | 4       | plpoznan.categorylinks.cl_from |    1 | Using where                                               |
+----+-------------+---------------+--------+---------------------------------------------------------------------+--------------+---------+--------------------------------+------+-----------------------------------------------------------+
```

The results set order is a bit different, however the results are still correct.([examples in the ticket](https://wikia-inc.atlassian.net/browse/PLATFORM-2213)).

@wladekb / @drozdo

### DIFF
--- a/extensions/wikia/RelatedPages/RelatedPages.class.php
+++ b/extensions/wikia/RelatedPages/RelatedPages.class.php
@@ -144,7 +144,7 @@ class RelatedPages {
 					}
 
 					// limit * 2 - get more pages (some can be filtered out - RT #72703)
-					$pages = $this->getPagesForCategories( $articleId, self::LIMIT_MAX * 2, $categories );
+					$pages = $this->getPagesForCategories( $articleId, $categories );
 
 					$this->afterGet( $pages, self::LIMIT_MAX );
 				}
@@ -183,14 +183,14 @@ class RelatedPages {
 	/**
 	 * get pages that belong to a list of categories
 	 * @author Owen
+	 * @author Macbre
 	 *
 	 * @param int $articleId
-	 * @param int $limit
 	 * @param array $categories
 	 * @return array
 	 * @throws DBUnexpectedError|MWException
 	 */
-	protected function getPagesForCategories( $articleId, $limit, Array $categories ) {
+	protected function getPagesForCategories( $articleId, Array $categories ) {
 		if ( empty( $categories ) ) {
 			return [];
 		}
@@ -204,11 +204,8 @@ class RelatedPages {
 		$pages = WikiaDataAccess::cache(
 			wfMemcKey( __METHOD__, 'categories', md5( serialize( $categories ) ) ),
 			WikiaResponse::CACHE_STANDARD,
-			function() use ( $categories, $fname, $limit ) {
+			function() use ( $categories, $fname ) {
 				$dbr = wfGetDB( DB_SLAVE );
-
-				# sanitize query parameters
-				$limit = intval( $limit );
 
 				/**
 				 * SELECT count(page_id) as c, cl_from AS page_id, page_namespace, page_title, page_is_redirect, page_len, page_latest
@@ -241,7 +238,7 @@ class RelatedPages {
 					[
 						'GROUP BY' => 'page.page_id',
 						'ORDER BY' => 'c DESC',
-						'LIMIT' => $limit,
+						'LIMIT' => self::LIMIT_MAX * 2,
 					],
 					[
 						'page' => [
--- a/extensions/wikia/RelatedPages/RelatedPages.class.php
+++ b/extensions/wikia/RelatedPages/RelatedPages.class.php
@@ -181,47 +181,6 @@ class RelatedPages {
 	}
 
 	/**
-	 * get all pages for given category
-	 * @param string $category category name
-	 * @return array list of page ids
-	 */
-	protected function getPagesForCategory( $category ) {
-		global $wgMemc;
-		wfProfileIn( __METHOD__ );
-
-		$cacheKey = wfMemcKey( __METHOD__, md5($category) );
-		$cache = $wgMemc->get( $cacheKey );
-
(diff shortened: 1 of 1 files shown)


extensions/wikia/RelatedPages/RelatedPages.class.php
```diff
@@ -232,62 +232,66 @@ protected function getPagesForCategory( $category ) {
...
+				$limit = intval( $limit );
+
+				$sql = "SELECT page_id, count(*) c FROM ( $innerSQL ) i GROUP BY page_id ORDER BY c desc LIMIT $limit";
+				$res = $dbr->query( $sql, $fname );
+				while ( $row = $dbr->fetchObject( $res ) ) {
+					$pageId = (int) $row->page_id;
+					$title = Title::newFromId( $pageId );
+
+					// filter out redirect pages (RT #72662)
+					if ( $title instanceof Title && $title->exists() && !$title->isRedirect() ) {
```
wladekb: IMHO the filter for redirect could be done on the SQL level, JOIN with `page` would be needed and condition on `page_is_redirect`, but this would save us from creating Title objects that do one query per instantiation. What do you think?

wladekb: Oh, we're creating title for more than checking for a redirect, in this case we could fetch all fields from `page` table and then use `Title::newFromRow()`. Sound reasonable?

macbre: We can do that to avoid checking the LinkCache for redirect. However, we still need to instantiate  `Title` object to get the URL.

macbre: `Title::newFromRow` introduced as a part of dccaf2e

extensions/wikia/RelatedPages/RelatedPages.class.php
```diff
@@ -181,143 +171,102 @@ protected function afterGet( $pages, $limit ) {
...
+				 * LIMIT 11;
+				 *
+				 * @see PLATFORM-2226
+				 */
+				$res = $dbr->select(
+					[ 'categorylinks', 'page' ],
+					[
+						'count(page_id) AS c',
+						// columns below are required by Title::newFromRow
+						'page_id',
```
wladekb, line 216: why not `page.*` to be sure nothing is omitted by mistake ?

macbre, line 216: I like being explicit :)

PR #10624, opened 2016-06-06
MWMemcached: improve reporting of too large values (add normalized key value)

macbre:
We see quite a lot of `SERVER_ERROR object too large for cache` errors reported by memcache client. This issue mostly affects (and is caused by) article comments.

Improve logging so we can easily group by normalized keys and find features that try to set to large values in memcache

@wladekb / @artursitarski


artursitarski:
:+1:

wladekb:
lgtm

PR #10625, opened 2016-06-06
SER-178: Add Span-Id and Trace-Id headers to the service clients

harnash:
I've regenerated sources for all Swagger clients MW uses. I had to make some fixes along the way and some changes caused some incompatibilities in our code which was hopefully fixed.
Since the Client generated documentation is something we should keep I had to do some minor cleanup there too.

ping @Wikia/services-team


pawelchcki:
Its hard to do proper CR of this PR. I've looked on it through Upsource and the non generated files look ok.  
so LGTM
but we need to setup better tools to review PRs like this

lib/Wikia/src/Service/Swagger/ApiClient.php
```diff
@@ -29,9 +29,8 @@ public function callApi($resourcePath, $method, $queryParams, $postData, $header
 		$response = $exception = null;
 		$code = 200;
 
-		// adding trace headers
-		$headerParams['X-Trace-Id'] = WikiaTracer::instance()->getTraceId();
-		$headerParams['X-Span-Id'] = WikiaTracer::instance()->getSpanId();
+		// adding internal headers
+		WikiaTracer::instance()->setRequestHeaders( $headerParams, true );
```
pawelchcki, line 33: :+1:

lib/Wikia/src/Service/Helios/HeliosClientImpl.php
```diff
@@ -63,9 +63,8 @@ public function request( $resourceName, $getParams = [], $postData = [], $extraR
 		global $wgRequest;
 		$headers['X-Forwarded-For'] = $wgRequest->getIP();
 
-		// adding trace headers
-		$headers['X-Trace-Id'] = WikiaTracer::instance()->getTraceId();
-		$headers['X-Span-Id'] = WikiaTracer::instance()->getSpanId();
+		// adding internal headers
+		WikiaTracer::instance()->setRequestHeaders( $headers, true );
```
pawelchcki, line 67: :+1:

PR #10626, opened 2016-06-06
RECIRC-101: Add curated content to rail experiments

pauloslund:
This uses a custom API endpoint to load curated posts from Fandom and inject them into our rail experiments. We only want to show one post to a user at a time, and after they have seen that post they will not see it for 24 hours.


extensions/wikia/Recirculation/templates/client/rail.mustache
```diff
@@ -2,7 +2,7 @@
 	<h2>{{{ titleHtml }}}</h2>
 	<ul class="thumbnails">
 		{{#items}}
-			<li class="rail-item {{classes}}" data-flag="{{ flag }}" data-index="{{ index }}" data-source="{{source}}">
+			<li class="rail-item {{classes}} item-{{source}}" {{#id}}data-id="{{ id }}"{{/id}} data-flag="{{ flag }}" data-index="{{ index }}" data-source="{{source}}">
```
gabrys: I'm not sure what are the guidelines for the mustache files, but I'd probably add spaces inside `{{#id}}...{{/id}}`

gabrys:
LGTM

PR #10627, opened 2016-06-06
SEO-366 Test: Small-Scale, Highly Relevant Crosslinks

gabrys:
This extension replaces selected words with links to manually selected Wikia articles on other wikis.

It's only supposed to run on English wikis for anonymous users. The links mimic the regular external links.
We're not tricking users into thinking they go to the next article on the same wiki.


saipetch:
LGTM

PR #10631, opened 2016-06-07
SUS-541 ImageReview: fetch more images from database

ghost:
This change is not supposed to fix the issue or make up for a proper implementation of the tool. It is expected to reduce the severity of the issue and I want to know how much.

The list of images in batch is limited to 20. A number of images (formerly 24, now 128) is fetched from the database. Then a piece of code removes items matching certain criteria and as soon as the batch for view has 20 items, the iteration is terminated.

The change has little performance impact and only results with more data being sent from MySQL to Apache. We can live with that as this is a staff tool.

The corresponding dashboard is here https://kibana.wikia-inc.com/#/dashboard/elasticsearch/SUS-541 and the expected effect of this change is lower error rate (errors are marked red).

http://wikia-inc.atlassian.net/browse/SUS-541

@jacekjursza @ludwikkazmierczak 

Excuse me a typo in the name of the branch.

### DIFF
--- a/extensions/wikia/ImageReview/ImageReviewHelperBase.class.php
+++ b/extensions/wikia/ImageReview/ImageReviewHelperBase.class.php
@@ -14,7 +14,7 @@ abstract class ImageReviewHelperBase extends WikiaModel {
 	 * we fetch a few icons from DB, we can skip them
 	 */
 	const LIMIT_IMAGES = 20;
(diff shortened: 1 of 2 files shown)


ludwikkazmierczak:
LGTM

ludwikkazmierczak:
Tested and Accepted

PR #10635, opened 2016-06-07
JPN-551 Show 'A wikia contributor' instead of IP address for anon use…

perjg:
Show 'A wikia contributor' instead of IP address for anon users for community page.

Also fixed some formatting issues in the nearby code.

@d34th4ck3r @Wikia/spitfires


d34th4ck3r:
Is there a reason why you choose to create a new function instead of directly editing getName function on User class ?

perjg:
My intention was not to complicate the logic of the existing getName function as it already has logic for checking item load status and IP sanitation. I felt getName would be harder to understand if I just added my logic on top of that, so instead I just wrapped my logic in a new function.

However, @lukaszkonieczny pointed out to me in slack that calling it getNameWithStringForAnon is misleading, so perhaps I should give it a different name?

I'm also open to rewriting all the logic into the existing getName, something like so:

```
    public function getName( $anonString = null ) {
        if ( $this->isItemLoaded( 'name', 'only' ) ) {
            # Special case optimisation
            $name =  $this->mName;
        } else {
            $this->load();
            if ( $this->mName === false ) {
                # Clean up IPs
                $this->mName = IP::sanitizeIP( $this->getRequest()->getIP() );
            }
            $name = $this->mName;
        }


        if ( $anonString && $this->isAnon() ) {
            return $anonString;
        }

        return $name;
    }
```

perjg:
@lukaszkonieczny Is it better to use isIP directly instead? Can we assume that all anon users will always have a user name that looks like an IP address? If so I can simplify my implementation and probably drop the extra getNameWithStringForAnon function entirely?

perjg:
Updated to use isIP as discussed with @lukaszkonieczny.

lukaszkonieczny:
LGTM

perjg:
@lukaszkonieczny Based on Trevor's suggestion we changed the string to use the already existing 'oasis-anon-user' string. Can you please check Gautam's latest commit?

lukaszkonieczny:
LGTM

PR #10595, opened 2016-06-02
PLATFORM-2233: use a "classic" MemCache client on devboxes and in Reston

macbre:
[PLATFORM-2233](https://wikia-inc.atlassian.net/browse/PLATFORM-2233)

Do not use Moxi on devboxes and in Reston, but instead connect directly to [Twitter's memcache proxy](https://github.com/twitter/twemproxy). See https://github.com/Wikia/config/pull/1792

@wladekb / @artursitarski

PR #10639, opened 2016-06-07
PLATFORM-2233: get rid of Moxi

macbre:
[PLATFORM-2233](https://wikia-inc.atlassian.net/browse/PLATFORM-2233)

`MemCachedClientforWiki` is now going to be used in all environments (dev, SJC and Reston) - remove `MemcacheMoxiCluster` and `MemcacheClientShadower` classes 

And profile `MWMemcached::_connect_sock` calls.

This is a follow-up of #10595

@wladekb / @artursitarski


wladekb:
@artursitarski @macbre Do we have any hard data for performance implications of this change?

artursitarski:
@wladekb AFAIK we don't have any precise data in this area. However, it shouldn't be noticeable impact. That's true, you have to create more connections to proxy server but twemproxy->memcache connection is persistent. We disabled moxi in Reston last week and didn't notice troubling changes in response time. Base on this I think it's worth to check it also in SJC. If it harm production in observable way we can just revert this change. The less layers of abstraction the better. Give it a try

macbre:
xhprof data collector normalizes all `MWMemcached` class calls to `mwmemcached`. We can take a look there and monitor percentiles values.

wladekb:
`xhprof` now reports separate `MWMemcached` methods as well as the summarized `mwmemcached`. This should help us get reasonable data for the comparison.

macbre:
Thanks, @wladekb !

PR #10641, opened 2016-06-07
SEO-366 Cross link test update

gabrys:
- Add two more hot words (Pokemon and Dragon)
- Limit the test to NS_MAIN


saipetch:
LGTM

PR #10644, opened 2016-06-08
JPN-549 Fixed incorrect time formatting for Japanese and Chinese dates

perjg:
Fixed incorrect time formatting for dates between 30 and 365 days old for Japanese and Chinese.

@Grunny This is an ad-hoc fix that works for both Japanese and Chinese, but it may not be the ideal way to fix this problem. Please see JPN-549 for my detailed comments.


includes/wikia/GlobalFunctions.php
```diff
@@ -937,7 +937,11 @@ function wfTimeFormatAgo( $stamp, $hideCurrentYear = true ) {
 		$format = $wgLang->getDateFormatString( 'date', 'default' );
 		if ( $hideCurrentYear ) {
 			$format = trim( $format, ' ,yY' );
+
+			// For Japanese and Chinese the 年 character before the month number must also be stripped. See JPN-549
+			$format = str_replace('年', '', $format);
```
lukaszkonieczny: Maybe we could fix it globally?

perjg: Not sure what you mean by globally.

lukaszkonieczny: Oh, I did not see it's already in Global Functions. Right now this string replacement will be done for every language, not only Japanese or Chinese. Maybe you can investigate usage of date formats in Messages*.php files like here https://github.com/Wikia/app/blob/dev/languages/messages/MessagesZh_hans.php#L239

perjg: Yes, but I think it is safe to say that str_replace with '年' will have no effect except for Japanese and Chinese.

Thanks for pointing out Messages*.php. This looks like the place to replace the default data format for Japanese and Chinese, so we can drop the (金) part from the date.

However, this fix is still required, since this is about the check for $hideCurrentYear being broken for Japanese and Chinese.

Grunny: Whitespace issues. Also, couldn't you just add 年 to the trim? i.e. `trim( $format, ' ,yY年' );`

perjg:
@Grunny @lukaszkonieczny 
I've updated the PR according to Grunny's suggestion: Using ' ,yY年' directly in the trim method works fine.

I've also changed the default date format form Japanese and Chinese Simplified / Traditional to drop the weekday name at the end.

Grunny:
LGTM.

perjg:
@Grunny Made one more commit to also drop the weekday at the end for the special Japanese Nengo and Taiwan Minguo date formats.

Grunny:
:+1:

PR #10645, opened 2016-06-08
CE-3808: Calculate Rank only for logged in users

lukaszkonieczny:
cc @Wikia/spitfires @perjg @d34th4ck3r


Grunny:
LGTM.

perjg:
See also: JPN-558
This task is a request to remove anons from the contribution module, so they seem overlapping.

PR #10638, opened 2016-06-07
[PLATFORM-2210] Added patroller group

jcellary:
https://wikia-inc.atlassian.net/browse/PLATFORM-2210

Per com sup's request created a new group

https://github.com/Wikia/config/pull/1811

PR #10647, opened 2016-06-08
Revert "[PLATFORM-2210] Added patroller group"

jcellary:
Reverts Wikia/app#10638

Due to multiple PTO's in ComSup we've been asked to delay the release of this change by two releases, hence reverting.


macbre:
:+1:

PR #10648, opened 2016-06-08
SUS-563 | Use $wgSitename instead of wfMsgForContent( 'mainpage' )

rogatty:
https://wikia-inc.atlassian.net/browse/SUS-563

There is a race condition that sometimes causes `wfMsgForContent( 'mainpage' )` to have the initial value of e.g. "Main Page" (depends on wiki language) instead of site name. New value is set in `CreateNewWikiTask::moveMainPage()` to `$wgSitename` but there seems to be some caching issue and it's not always propagated when we need it. Instead of depending on cache let's use the global directly.

@mixth-sense @ludwikkazmierczak


armonr:
:shipit:

ghost:
:+1:

PR #10652, opened 2016-06-08
ADEN-3406 SP Firefox fix

drets:
Recovered ads are placed in divs without styles.

Fix:
- add the crossorigin=“anonymous” attribute to CSS link element 
- return the proper CORS header for CSS link element

CC: @macbre

### DIFF
--- a/extensions/wikia/AssetsManager/AssetsManagerServer.class.php
+++ b/extensions/wikia/AssetsManager/AssetsManagerServer.class.php
@@ -107,7 +107,7 @@ class AssetsManagerServer {
 			$headers['X-Pass-Cache-Control'] = 'public, max-age=' . $cacheDuration['client'];
 		}
 
-		if ($type === 'saas' || $type === 'sasses') {
+		if ( $type === 'sasses' && strpos( $request->getRequestURL(), urlencode( 'crossorigin=anonymous' ) ) !== false ) {
 			header( 'Access-Control-Allow-Origin: *' );
 		}
 
--- a/extensions/wikia/AssetsManager/AssetsManagerServer.class.php
+++ b/extensions/wikia/AssetsManager/AssetsManagerServer.class.php
@@ -107,6 +107,10 @@ class AssetsManagerServer {
 			$headers['X-Pass-Cache-Control'] = 'public, max-age=' . $cacheDuration['client'];
 		}
 
+		if ($type === 'saas' || $type === 'sasses') {
+			header( 'Access-Control-Allow-Origin: *' );
+		}
+
 		$headers['Last-Modified'] = gmdate('D, d M Y H:i:s \G\M\T');
 
 		foreach($headers as $k => $v) {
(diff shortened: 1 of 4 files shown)


includes/wikia/nirvana/WikiaSkin.class.php
```diff
@@ -276,15 +276,17 @@ public function getStylesWithCombinedSASS(Array &$sassFiles) {
 		$sassFiles = $this->assetsManager->getSassFilePath($sassFiles);
 
 		// get a single URL to fetch all the required SASS files
-		$sassFilesUrl = $this->assetsManager->getSassesUrl($sassFiles);
+		$sassParams = SassUtil::getSassSettings();
+		$sassParams['crossorigin'] = 'anonymous';
```
drets, line 280: Not sure if it is a proper way of adding a custom param.

macbre, line 280: I gave this approach a second thought and...  Instead of using the URL parameter and passing it through several layers of code we can use `config.php` file and define a flag for AM group to set `crossorigin` header to a given value when serving a specific group.
- less code changes
- we have better control over which assets can be served with `crossorigin` header

includes/wikia/nirvana/WikiaSkin.class.php
```diff
@@ -276,15 +276,17 @@ public function getStylesWithCombinedSASS(Array &$sassFiles) {
...
+		$sassFilesUrl = $this->assetsManager->getSassesUrl( $sassFiles, $sassParams );
 
 		// recovery unlock css
 		$unlockUrl = ARecoveryUnlockCSS::getUnlockCSSUrl();
 
 		wfDebug( sprintf( "%s: combined %d SASS files\n", __METHOD__, count($sassFiles) ) );
 
 		wfProfileOut(__METHOD__);
-		return Html::linkedStyle($sassFilesUrl) . Html::linkedStyle($unlockUrl) . implode('', $cssLinks);
+		return Html::linkedStyle( $sassFilesUrl, 'all', 'anonymous' ) . Html::linkedStyle( $unlockUrl ) . implode( '', $cssLinks );
```
fraszczakszymon, line 289: Too long line :devil:

extensions/wikia/AssetsManager/AssetsManagerServer.class.php
```diff
@@ -107,6 +107,10 @@ public static function serve(WebRequest $request) {
 			$headers['X-Pass-Cache-Control'] = 'public, max-age=' . $cacheDuration['client'];
 		}
 
+		if ( $type === 'sasses' && strpos( $request->getRequestURL(), urlencode( 'crossorigin=anonymous' ) ) !== false ) {
```
fraszczakszymon, line 110: Now `crossorigin=anonymous` can be added to any sass url and it'll be returned with additional header. Is it expected?

includes/wikia/nirvana/WikiaSkin.class.php
nandy-andy, line 280: Makes sense for me. @drets, @fraszczakszymon, @jacekjursza is this idea going to work with recovery? My assumption is it is since recovery is just a proxy for URLs generated with AM and it does not change the AM groups. What are your thoughts?

drets:
Closed in favor of https://github.com/Wikia/app/pull/10695

PR #10653, opened 2016-06-08
Adds sane defaults and logging when the server's response doesn't conform to what we're expecting to see.

gbenson-wikia:
This also addresses https://wikia-inc.atlassian.net/browse/CAKE-181 and https://wikia-inc.atlassian.net/browse/CAKE-179 at the same time.


extensions/wikia/Recirculation/services/CakeRelatedContentService.php
```diff
@@ -29,9 +30,46 @@ public function getContentRelatedTo($title, $limit=5, $ignore=null) {
 		}
 
 		$api = $this->relatedContentApi();
-		
+
 		try {
 			$filteredRelatedContent = $api->getRelatedContentFromEntityName($title, $limit + 1, "true");
+            if (is_null($filteredRelatedContent)) {
+                WikiaLogger::instance()->warning("getRelatedContentFromEntityName failed to retrieve recommendations", [
```
nmonterroso: the `Loggable` trait, which this class uses, adds `$this->warning`, so importing the logger and using `WikiaLogger::instance` is unnecessary (though it all ends up going to the same place)

nmonterroso: Also, the spacing seems off here. I'm not sure if the original file is correct and the additions are wrong, or vice versa.

nmonterroso:
minor comments otw lgtm

nmonterroso:
the spacing is still off :(

PR #10654, opened 2016-06-09
[PLATFORM-2263] Removed core SpecialListUsers

jcellary:
https://wikia-inc.atlassian.net/browse/PLATFORM-2263

Removed core implementation of the list users special page - there is a wikia extension which overrides that special page and it's enabled on all but 17 wikis. We agreed with com sup that we can turn it on all wikis and thus the core special page will not be used any more and can be deleted

SpecialActiveUsers seems to have been always redirected to the list users special page when the extension is enabled, so it can also be removed.

https://github.com/Wikia/config/pull/1817


macbre:
It's EURO 2016 after all

PR #10657, opened 2016-06-10
enable reflective cors - disable by default

hakubo:
This is needed if we want to develop other services locally
This is disabled by default https://github.com/Wikia/config/pull/1818

and can be enabled in your DevBoxSettings

I cannot send \* in origin as we do make withCredentials api calls - and \* is not allowed with such requests

### DIFF
--- a/includes/wikia/nirvana/WikiaResponse.class.php
+++ b/includes/wikia/nirvana/WikiaResponse.class.php
@@ -446,6 +446,10 @@ class WikiaResponse {
 	}
 
 	public function sendHeaders() {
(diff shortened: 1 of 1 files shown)


includes/wikia/nirvana/WikiaResponse.class.php
```diff
@@ -483,6 +487,31 @@ public function sendHeaders() {
 		}
 	}
 
+	private function reflectiveCORS($enabled) {
```
rogatty, line 490: This should have some tests.

PR #10658, opened 2016-06-10
RECIRC 125: Add E3 right rail treatment

pauloslund:
Special right rail group for E3. Will only be active for the week, then the code will be removed.


extensions/wikia/Recirculation/RecirculationContent.php
```diff
@@ -27,23 +29,25 @@ class RecirculationContent implements JsonSerializable {
...
-	public function __construct($index, $url, $thumbnail, $title, $publishDate, $author) {
-		$this->index = $index;
-		$this->url = $url;
-		$this->thumbnail = $thumbnail;
-		$this->title = $title;
-		$this->publishDate = $publishDate;
-		$this->author = $author;
+	public function __construct( array $properties = [] ) {
+		foreach($properties as $key => $value){
+			$this->{$key} = $value;
```
gabrys: Let's save that in an old-school assoc. array instead of relying on magic like that.

extensions/wikia/Recirculation/RecirculationContent.php
```diff
@@ -27,23 +29,25 @@ class RecirculationContent implements JsonSerializable {
...
+		}
+	}
+
+	public function get( $property ){
+		return isset( $this->{ $property } ) ? $this->{ $property } : '';
 	}
 
 	function jsonSerialize() {
```
gabrys: add `public` also the name suggests it's returning this JSON-ed

extensions/wikia/Recirculation/images/logomark.svg
```diff
@@ -0,0 +1,23 @@
+<svg width="20" height="19" viewBox="0 0 20 19" xmlns="http://www.w3.org/2000/svg">
```
gabrys: We should name it "fandom-logo" or something like that, so it's clear from the name what it is.

extensions/wikia/Recirculation/services/ParselyDataService.class.php
```diff
@@ -0,0 +1,197 @@
...
+
+		return $metadata;
+	}
+
+	private function buildVerticalData() {
+		global $wgCityId;
+		$wikiFactoryHub = WikiFactoryHub::getInstance();
+		$wgWikiVertical = $wikiFactoryHub->getWikiVertical( $wgCityId )['short'];
+
+		$verticalMap = [
```
gabrys, line 157: isn't this re-used by other providers?

extensions/wikia/Recirculation/templates/client/rail.mustache
```diff
@@ -7,7 +7,7 @@
 					<div class="image image-thumbnail fluid small">
 						<img src="{{ thumbnail }}" alt="{{ title }}" itemprop="thumbnail">
 					</div>
-					<div class="title" title="{{ title }}">{{ title }}</div>
+					<div class="title" title="{{{ title }}}">{{{ title }}}</div>
```
gabrys: no escaping for title="..."

extensions/wikia/Recirculation/services/ParselyDataService.class.php
pauloslund, line 157: Not currently. It's a map between the vertical of the current community and the tag name of the vertical on Fandom.

extensions/wikia/Recirculation/RecirculationContent.php
pauloslund: Added public, but the name is based on http://php.net/manual/en/jsonserializable.jsonserialize.php

gabrys:
LGTM

PR #10660, opened 2016-06-13
SUS-563 | use $wgSitename instead of wfMsgForContent( 'mainpage' )

rogatty:
https://wikia-inc.atlassian.net/browse/SUS-563

LGTMized in https://github.com/Wikia/app/pull/10648


rogatty:
It's no longer relevant.

PR #10664, opened 2016-06-13
ArticleComments: performance improvements

macbre:
[Rendering of user blog posts takes a while](https://wikia-inc.atlassian.net/browse/PLATFORM-2260). Thanks to NewRelic slow transactions details we were able to find the bottleneck.
## Problem description

Each call to `Title::checkPermissionHooks` triggers a hook that is bound by `ArticleComment` class. This section is critical in terms of performance as it's called 20+ times on each user blog view.

`ArticleComment::userCan` what not only loading comment metadata (like revisions data and the user that added a comment), but was parsing the comment / blog post content (on each call). This has a really negative impact on larger blog posts - it can take up to 90-120 sec to render them (and we observe many timeouts here).
## Solution

Improve lazy loading mechanism. Extract a part of `load` method that loads revisions and user data into a separate method called `loadRevisionsAndAuthor`. Use the new method in `isAuthor` and `editPage` to lazy-load required data when needed (without triggering costly parsing).

`ArticleComment::parseText` method makes a following call:

``` php
$wgEnableParserCache = false
```

@garthwebb - as the owning team do you recall why we disable parser cache when rendering article comments (and hence user blogs and Wall messages)?
## Results

> The Following profiler data is for [a pretty short blog post entry](http://poznan.wikia.com/wiki/Blog_u%C5%BCytkownika:Macbre/Edycja_nr_15000)

```
view-source:http://poznan.wikia.com/wiki/Blog_u%C5%BCytkownika:Macbre/Edycja_nr_15000?forceprofile=1

 15.31% 0.102219     23 - ArticleComment::userCan
```

```
# after the fix on dev
view-source:http://poznan.wikia.com/wiki/Blog_u%C5%BCytkownika:Macbre/Edycja_nr_15000?forceprofile=1
  1.36% 0.045222     23 - ArticleComment::userCan

# without the fix on dev
  4.70% 0.143464     23 - ArticleComment::userCan
  3.51% 0.107008      8 - ArticleComment::parseText
```

> For [a blog post that takes ~85 seconds to render](http://godzilla.wikia.com/wiki/User_blog:The_Boy_Who_Cried_Godzilla/Fun_with_Navboxes):

```
 87.00% 38.795542     23 - ArticleComment::userCan
```

Pinging @wladekb and @garthwebb for code review.

### DIFF
--- a/extensions/wikia/ArticleComments/classes/ArticleComment.class.php
+++ b/extensions/wikia/ArticleComments/classes/ArticleComment.class.php
@@ -48,8 +48,8 @@ class ArticleComment {
 
 	protected $minRevIdFromSlave;
 
-	/** @var bool */
-	private $isLoaded = false;
+	private $isTextLoaded = false;
+	private $isRevisionLoaded = false;
 
 	/**
 	 * @param Title $title
@@ -191,7 +191,34 @@ class ArticleComment {
 	 * @return bool
 	 */
 	public function load( $master = false ) {
-		if ( $this->isLoaded ) {
+		$ret = $this->loadRevisionsAndAuthor();
+		if ( $ret === false ) {
+			return false;
+		}
+
+		if ( $this->isTextLoaded ) {
+			return true;
+		}
+
+		$rawText = $this->mLastRevision->getText();
+		$this->parseText( $rawText );
+
+		$this->isTextLoaded = true;
+		return true;
+	}
+
+	/**
+	 * Lazy load revisions and comment's author data
+	 *
+	 * We can use this method to check user permissions instead of calling load() that parses the content (which takes time!)
+	 *
+	 * @see PLATFORM-2260
+	 *
+	 * @param bool $master
+	 * @return bool
+	 */
+	private function loadRevisionsAndAuthor( $master = false ) {
+		if ( $this->isRevisionLoaded ) {
 			return true;
 		}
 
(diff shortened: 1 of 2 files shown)


extensions/wikia/ArticleComments/classes/ArticleComment.class.php
```diff
@@ -632,7 +662,7 @@ public function canEdit() {
 	 * @return bool
 	 */
 	public function isAuthor( $user ) {
-		if ( $this->mUser ) {
+		if ( $this->loadRevisionsAndAuthor( true ) && $this->mUser ) {
```
wladekb, line 665: `$this->loadRevisionsAndAuthor()` does nothing when called for the second time and beyond, if the first call didn't request load from master then the actual data will be coming from slave while we explicitly request for master data here. Worth fixing?

macbre, line 665: I've asked myself the same question :) It used to be the same in the old implementation and we did not spot any problems. I'd keep it as it is.

macbre, line 665: I forgot to pass `$master` arg to `loadRevisionsAndAuthor` method - fixed in 10c816a.

wladekb:
lgtm

PR #10668, opened 2016-06-14
JPN-483 | Added function to remove certain user from Top Contributors

d34th4ck3r:
@Wikia/spitfires @perjg


extensions/wikia/CommunityPage/models/CommunityPageSpecialUsersModel.class.php
```diff
@@ -190,6 +191,47 @@ function () {
 		return $botIds;
 	}
 
+
+	/**
+	 * @return array list of blocked ids for Top Contributors
+	 */
+	private function getBlockedIds(){
```
Grunny: I'd consider renaming this method. "Blocked" has a specific meaning in MediaWiki when relating to users, so when I read `getBlockedIds`, I expected the method to be getting a list of "blocked" users (i.e. blocked from editing), not a blacklist of users based on user group.

extensions/wikia/CommunityPage/models/CommunityPageSpecialUsersModel.class.php
```diff
@@ -190,6 +191,47 @@ function () {
...
+					->runLoop( $globalDb, function ( &$globalIds, $row ) {
+						$globalIds[] = $row->ug_user;
+					} );
+
+				$localDb = wfGetDB( DB_SLAVE );
+
+				$localIds = ( new WikiaSQL() )
+					->SELECT( 'ug_user' )
+					->FROM ( 'user_groups' )
+					->WHERE( 'ug_group' )->IN( [ 'staff', 'util', 'helper', 'vstf' ] )
```
Grunny: These groups shouldn't be in the local table. Is this query necessary?

d34th4ck3r:
Made changes according to your comments. Please recheck.

extensions/wikia/CommunityPage/models/CommunityPageSpecialUsersModel.class.php
```diff
@@ -190,6 +191,36 @@ function () {
...
+					->FROM ( 'user_groups' )
+					->WHERE( 'ug_group' )->IN( [ 'bot', 'bot-global', 'staff', 'util', 'helper', 'vstf' ] )
+					->GROUP_BY( 'ug_user' )
+					->runLoop( $globalDb, function ( &$globalIds, $row ) {
+						$globalIds[] = $row->ug_user;
+					} );
+
+				$allBlockedIds = array_merge( $globalIds, $this->getBotIds() );
+
+				return $allBlockedIds;
```
Grunny: This is still using the "blocked" in the variable name, but I'd suggest just changing it to:

``` php
return array_merge( $globalIds, $this->getBotIds() );
```

extensions/wikia/CommunityPage/models/CommunityPageSpecialUsersModel.class.php
```diff
@@ -7,6 +7,7 @@ class CommunityPageSpecialUsersModel {
 	const ALL_ADMINS_MCACHE_KEY = 'community_page_all_admins';
 	const GLOBAL_BOTS_MCACHE_KEY = 'community_page_global_bots';
 	const ALL_BOTS_MCACHE_KEY = 'community_page_all_bots';
+	const ALL_BLOCKED_IDS_MCACHE_KEY = 'community_page_all_blocked_ids';
```
Grunny: Likewise, still using "Blocked".

Grunny:
Last two comments, otherwise LGTM. :)

extensions/wikia/CommunityPage/models/CommunityPageSpecialUsersModel.class.php
```diff
@@ -190,6 +191,48 @@ function () {
...
+					->runLoop( $globalDb, function ( &$globalIds, $row ) {
+						$globalIds[] = $row->ug_user;
+					} );
+
+				$localDb = wfGetDB( DB_SLAVE );
+
+				$localUsers = ( new WikiaSQL() )
+					->SELECT( 'ug_user' )
+					->FROM ( 'user_groups' )
+					->WHERE( 'ug_group' )->NOT_IN( [ 'bot', 'bot-global' ] )
```
Grunny: Isn't this just [the exact query from `getBotIds`](https://github.com/Wikia/app/blob/6da13d7280cb58024838318061bc83933cd2d6f5/extensions/wikia/CommunityPage/models/CommunityPageSpecialUsersModel.class.php#L174-L185), so would be redundant to calling `getBotIds` below, right?

d34th4ck3r: It's actually the exact opposite. (Notice NOT_IN)
Basically I'm trying to remove local users from blocked ids if they are in any of the global groups.

Grunny: Oh, bah, I missed that. ;)

So, you want to remove local users (who are also one of staff, helper, etc. globally) from being blacklisted if they have local rights (basically if they either have rollback or sysop rights)?

Also, you can remove `bot-global` from this query, that's only a global right, so shouldn't be in the local table.

d34th4ck3r: Right! I want to remove any local users from blacklist unless they are bot.

Done.

Grunny:
LGTM. :)

PR #10669, opened 2016-06-14
Remove RegexBlock extension

macbre:
[PLATFORM-2216](https://wikia-inc.atlassian.net/browse/PLATFORM-2216)

This extension uses `wikicities.blockedby` and `blockedby_stats` tables which are empty. The extension itself is not reachable since we enabled Phalanx (it used to provide `Special:RegexBlock`). Let's remove it.


jcellary:
I'm assuming we don't want to touch the DB and remove the unused tables :) ? @wladekb asked me once that when removing unused extensions we should also look the unused tables, so that we can potentially remove them

jcellary:
Other than the comment about tables, LGTM

macbre:
Actually, we found this extension after we renamed empty tables on wikicities and we got several reports  in the logs that the code tried to access these empty tables.

PR #10671, opened 2016-06-14
ADEN-3446 Removed all paddings and margins for taboola leaderboard

fraszczakszymon:


nandy-andy:
The failed integration test is not related to SCSS changes. Merging.

PR #10672, opened 2016-06-14
MWMemcached: do not report NOT_STORED responses as errors

macbre:
[PLATFORM-2268](https://wikia-inc.atlassian.net/browse/PLATFORM-2268) / [memcache protocol documentaion](https://github.com/memcached/memcached/blob/master/doc/protocol.txt) says:

```
After sending the command line and the data block the client awaits
the reply, which may be:

...

- "NOT_STORED\r\n" to indicate the data was not stored, but not
because of an error. This normally means that the
condition for an "add" or a "replace" command wasn't met.

...

  "add" means "store this data, but only if the server *doesn't* already
  hold data for this key".
```

@wladekb / @artursitarski

### DIFF
--- a/includes/objectcache/MemcachedClient.php
+++ b/includes/objectcache/MemcachedClient.php
@@ -1247,14 +1247,17 @@ class MWMemcached {
 
 		// Wikia change - begin
 		// @author macbre (PLATFORM-774)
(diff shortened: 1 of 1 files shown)


wladekb:
:+1:

PR #10673, opened 2016-06-14
SUS-544 | Use template instead of replacing main page's content

rogatty:
https://wikia-inc.atlassian.net/browse/SUS-544
### Wiki description

Instead of blindly replacing the first section of main page with a site description let's use a new template - `Template:SiteDescription`. This way the ComDev will be able to put it wherever they please instead of being limited by some obscure, hardcoded rules which weren't communicated anywhere else but in the code itself.
### Wiki name

This PR also removes logic which was replacing `Wiki` word in the first section's header to `$wgSitename`. I've checked all starters and only two of them still depends on it:
- http://starter.wikia.com/wiki/Main_Page?action=raw
- http://de.starter.wikia.com/wiki/Hauptseite?action=raw

All other starters either:
- use another structure or wording so the replacement doesn't work anyway
- use `{{SITENAME}}` which is preferable way to achieve the effect
### Non-code changes

Before we can merge it we should:
- create `Template:SiteDescription` on every starter
- optionally put a default description there (if we don't do that, we will show description only if user provided it)
- edit main pages in starters to include `Template:SiteDescription`
### Reviewers

@mixth-sense

### DIFF
--- a/extensions/wikia/CreateNewWiki/FinishCreateWikiController.class.php
+++ b/extensions/wikia/CreateNewWiki/FinishCreateWikiController.class.php
@@ -5,6 +5,7 @@ use \Wikia\Logger\WikiaLogger;
 class FinishCreateWikiController extends WikiaController {
 
 	const COOKIE_NAME = 'createnewwiki';
+	const SITE_DESCRIPTION_TEMPLATE_NAME = 'SiteDescription';
 
 	// form field values
 	var $params;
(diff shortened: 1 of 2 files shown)


extensions/wikia/CreateNewWiki/FinishCreateWikiController.class.php
```diff
@@ -156,24 +157,23 @@ private function getMoMMainPage( $mainPageTitle ) {
...
 	 */
-	private function getClassicMainPage( $mainArticle ) {
-		global $wgParser, $wgSitename;
-
-		$mainPageText = $mainArticle->getRawText();
-		$matches = array();
+	private function editSiteDescriptionTemplate() {
 		$description = $this->params['wikiDescription'];
 
-		if ( preg_match( '/={2,3}[^=]+={2,3}/', $mainPageText, $matches ) ) {
```
ghost, line 170: :heart: for removing this obscure piece

ghost:
:+1: LGTM

ghost:
It has been communicated to Community Support/Development and will be merged as soon as we agree on the release date.

PR #10679, opened 2016-06-15
do not set $smwgEnableUpdateJobs if already set

ghost:
This change helps to mitigate severe problems caused by one SMW flaw. The sequence of update jobs falls into an infinite loop, which is dangerous for the entire RabbitMQ messaging system.

`$smwgEnableUpdateJobs` can be set to `false` via WikiFactory for one problematic wiki, which will stop the sequence immediately. It has minor user-facing implications. There is a limited collection of articles that require mutual updating.

https://wikia-inc.atlassian.net/browse/SUS-647

@ljagiello @macbre


macbre:
:+1:

PR #10682, opened 2016-06-15
DAT-4136 category won't be duplicated on interrupted edit

idradm:


sebastianmarzjan:
Rebase to dev should fix the test failures.
The fix sense to me makes. (LG)

PR #10683, opened 2016-06-15
DAT-4469 | use proper variable + reduce log level error --> warning

tomasznapieralski:
https://wikia-inc.atlassian.net/browse/DAT-4469

//cc @Wikia/x-wing

### DIFF
--- a/extensions/wikia/MercuryApi/MercuryApiController.class.php
+++ b/extensions/wikia/MercuryApi/MercuryApiController.class.php
@@ -166,7 +166,7 @@ class MercuryApiController extends WikiaController {
 		$navigation = $this->getNavigation();
 
 		if ( empty( $navigation ) ) {
-			\Wikia\Logger\WikiaLogger::instance()->warning(
+			\Wikia\Logger\WikiaLogger::instance()->debug(
 				'Fallback to empty navigation'
 			);
 		}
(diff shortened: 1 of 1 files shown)


extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -164,8 +164,9 @@ public function getArticleComments() {
 	private function prepareWikiVariables() {
 		$wikiVariables = $this->mercuryApi->getWikiVariables();
 		$navigation = $this->getNavigation();
-		if ( empty( $navData ) ) {
-			\Wikia\Logger\WikiaLogger::instance()->error(
+
+		if ( empty( $navigation ) ) {
+			\Wikia\Logger\WikiaLogger::instance()->warning(
```
rogatty: See https://wikia-inc.atlassian.net/wiki/display/PLAT/Logging#Logging-severitylevelsMeaningfulseveritylevels

IMO this should be `debug` level as there is no action we can take here.

rogatty:
LGTM

extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -164,8 +164,9 @@ public function getArticleComments() {
 	private function prepareWikiVariables() {
 		$wikiVariables = $this->mercuryApi->getWikiVariables();
 		$navigation = $this->getNavigation();
-		if ( empty( $navData ) ) {
-			\Wikia\Logger\WikiaLogger::instance()->error(
+
+		if ( empty( $navigation ) ) {
+			\Wikia\Logger\WikiaLogger::instance()->debug(
 				'Fallback to empty navigation'
```
bognix, line 170: make it notice please. It was logged as an error before our code refactor back in January. On the other hand I am wondering if we ever going to do something with this information.

bognix:
general wonderment on my side, beside LGTM

PR #10686, opened 2016-06-15
DAT-4470 | don't send error to kibana when width or height is missing for file which doesn't exist + reduce severity

tomasznapieralski:
https://wikia-inc.atlassian.net/browse/DAT-4470

//cc @Wikia/x-wing

### DIFF
--- a/extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
+++ b/extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
@@ -551,11 +551,11 @@ class ArticleAsJson extends WikiaService {
 	) {
 		$mediaDetail = WikiaFileHelper::getMediaDetail( $title, $mediaDetailConfig );
 
-		if ( !empty( $mediaDetail['mediaType'] ) ) {
+		if ( $mediaDetail['exists'] === true ) {
 			if ( empty( $mediaDetail['width'] ) ) {
 				$mediaDetail['width'] = $fallbackSize;
 
-				\Wikia\Logger\WikiaLogger::instance()->debug(
+				\Wikia\Logger\WikiaLogger::instance()->notice(
 					'ArticleAsJson - Media width was empty - fallback to fallbackSize',
 					[
 						'media_details' => $mediaDetail,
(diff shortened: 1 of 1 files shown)


extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
```diff
@@ -550,28 +550,31 @@ private static function getMediaDetailWithSizeFallback(
...
 
-			\Wikia\Logger\WikiaLogger::instance()->error(
-				'ArticleAsJson - Media width was empty - fallback to fallbackSize',
-				[
-					'media_details' => $mediaDetail,
-					'fallback_size' => $fallbackSize
-				]
-			);
-		}
+		if ( !empty( $mediaDetail['mediaType'] ) ) {
```
bognix: are you sure we want to check here `mediaType` not check if `Title`exists?

tomasznapieralski: it doesn't matter, because `WikiaFileHelper::getMediaDetail()` returns array filled with keys, but without values if `Title` is invalid.
That way is also consistent with other `if`s in this function

rogatty: I suggest using `$mediaDetail['exists'] === true`, that's what it's for.
https://github.com/Wikia/app/blob/dev/includes/wikia/services/WikiaFileHelper.class.php#L311

extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
```diff
@@ -550,28 +550,31 @@ private static function getMediaDetailWithSizeFallback(
...
+			if ( empty( $mediaDetail['width'] ) ) {
+				$mediaDetail['width'] = $fallbackSize;
+
+				\Wikia\Logger\WikiaLogger::instance()->debug(
+					'ArticleAsJson - Media width was empty - fallback to fallbackSize',
+					[
+						'media_details' => $mediaDetail,
```
rogatty, line 561: Can you also unify it? It's snake_cased here and camelCased below.

extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
bognix: I lean toward @rogatty suggestion. Empty mediaType doesn't represent what we are trying to check here, it's just a side effect of not existing file.

extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
bognix, line 561: And when unifying please make it compatible with our API client (Mercury), so use snake_cased in both places.

bognix:
minor comments.

bognix:
LGTM

PR #10687, opened 2016-06-15
SUS-639 Un-stifle on-site event notifications for all following users

armonr:
This aims to fix two issues:
1. On each event being followed by multiple users, only the first follower probing for notifications 'wins' in getting the notification. The rest of the followers will not because the marker had no user id factored in it.
2. On multiple events by same author, the found flag is raised prematurely which can cause more notification entities removed than necessary

/CC @mixth-sense @garthwebb


ghost:
LGTM :+1:

khoffmannwikia:
LGTM too !

PR #10691, opened 2016-06-16
ArticleComment: always return a bool from loadRevisionsAndAuthor()

macbre:
Fix a regression caused by #10664 - users were not able to edit their own comments.

@ludwikkazmierczak / @wladekb


macbre:
Merged with preview in ab3bd28

PR #10695, opened 2016-06-16
ADEN-3406 Fix styles for recoverable ads on FF

drets:
Second approach for fixing styles of recoverable ads on FF (first - https://github.com/Wikia/app/pull/10652).


extensions/wikia/AssetsManager/builders/AssetsManagerSassesBuilder.class.php
```diff
@@ -21,6 +21,10 @@ public function __construct(WebRequest $request) {
 			$builder = new AssetsManagerSassBuilder($sassRequest);
 
 			$this->mContent .= "\n\n/* $file */\n" . $builder->getContent();
+
+			if (strpos($this->mOid, 'ads.scss') !== false) {
```
nandy-andy: :see_no_evil: we can still do better and try follow @macbre's [commnet](https://github.com/Wikia/app/pull/10652/files#r66602002)

includes/Html.php
```diff
@@ -627,15 +627,20 @@ public static function inlineStyle( $contents, $media = 'all' ) {
 	 *
 	 * @param $url string
 	 * @param $media mixed A media type string, like 'screen'
+	 * @param $crossorigin Configure the CORS requests for the fetched data
 	 * @return string Raw HTML
 	 */
-	public static function linkedStyle( $url, $media = 'all' ) {
-		return self::element( 'link', array(
+	public static function linkedStyle( $url, $media = 'all', $crossorigin = null ) {
```
nandy-andy: This is core MW file. We want to avoid as much as possible changing it. What do you think of creating `WikiaHtml` class which extends the possibility of this method and using it in `WikiaSkin` part?

extensions/wikia/AssetsManager/builders/AssetsManagerSassesBuilder.class.php
macbre: Just an idea here: we're using `/sasses/` AssetsManager entry point to render and fetch several SCSS in a single request. What if we make `/sasses/` entry point support both AM groups and single files? Using this approach we can introduce `ads` AM group that will have `crossorigin` setting in AM's `config.php`' - hence causing the `/sasses/` response that contains `ads` group to have a proper HTTP header set.

Do we have to set both response header and the `<link>` tag attribute to have `crossorigin` working?

#### Pros:
- you can tweak `crossorigin` setting on per-group basis
- no MW core "hacking" - adding HTML attributes etc.
- more secure as we do not rely on client-provided URL parameter that controls `crossorigin` header

#### Cons:
- can't think of any now ;)

includes/wikia/nirvana/WikiaSkin.class.php
```diff
@@ -283,8 +283,11 @@ public function getStylesWithCombinedSASS(Array &$sassFiles) {
 
 		wfDebug( sprintf( "%s: combined %d SASS files\n", __METHOD__, count($sassFiles) ) );
 
+		$main = strpos($sassFilesUrl, 'ads.scss') !== false ?
```
nandy-andy: It's worth updating `WikiaSkinTest::testGetStylesWithCombinedSASS` since we changing the method.

jacekjursza:
LGTM

fraszczakszymon:
LGTM

PR #10698, opened 2016-06-16
SUS-638 | Don't try to assign property on an array

rogatty:
https://wikia-inc.atlassian.net/browse/SUS-638

/cc @mixth-sense


ghost:
:+1: LGTM

PR #10701, opened 2016-06-16
(DAT-4494) Quick fix Portable Infobox rendering

Grunny:
After the MW security release, the format of the UNIQ parser markers was changed
and the way PIs replace them broke.

In Mercury, this is because the JSON encoding was encoding and escaping the quote
in the unique prefix and then the replace was being done on the JSON encoded string.

This is a quick and ugly fix to get PIs working again on Mercury, but needs to be refactored
with more time (in particular the quick and dirty fix for `PortableInfoboxParserTagController`
when `wgArticleAsJson` is true).

/cc @rogatty


rogatty:
Ugly but works. Needs a follow up cleaning but LGTM as a P2 fix.

PR #10703, opened 2016-06-17
Adding <br> as an allowed tag.

NoIdeaHowToCallMe:
Advantage: more possibilities for design on mobile phones.

@SebastianMarzjan

### DIFF
--- a/extensions/wikia/PortableInfobox/services/Sanitizers/NodeDataSanitizer.php
+++ b/extensions/wikia/PortableInfobox/services/Sanitizers/NodeDataSanitizer.php
@@ -1,10 +1,10 @@
 <?php
 
 class NodeDataSanitizer extends NodeSanitizer {
-	protected $allowedTags = [ 'a' ];
+	protected $allowedTags = [ 'a','br' ];
 
 	/**
-	 * @desc remove all HTML tags but links from data labels.
+	 * @desc remove all HTML tags but links and wrapping from data labels.
 	 * If label after sanitization became empty because contained only image
 	 * do not sanitize it.
 	 *


sebastianmarzjan:
@NoIdeaHowToCallMe
Hi, thanks for the suggestion!

However, given current direction of mobile rendering this needs to be solved differently by treating &lt;br&gt; tags as whitespace characters. If you feel like adjusting the PR to accommodate that approach (please add tests too) that would be great.

NoIdeaHowToCallMe:
@SebastianMarzjan 
Since I'm new here: What do you mean with a test?

sebastianmarzjan:
@NoIdeaHowToCallMe 
Sure! I am referring to unit tests, please see: https://github.com/Wikia/app/blob/dev/extensions/wikia/PortableInfobox/tests/sanitizers/NodeDataSanitizerTest.php

NoIdeaHowToCallMe:
@SebastianMarzjan 
[NodeDataSanitizerTest.txt](https://github.com/Wikia/app/files/331690/NodeDataSanitizerTest.txt)
Something like that?

NoIdeaHowToCallMe:
@SebastianMarzjan 
Since there was no reaction on the previous answer I guess I have to do it in another way. So I just cloned the wikia repository and edited it.
https://github.com/NoIdeaHowToCallMe/app/blob/dev/extensions/wikia/PortableInfobox/tests/sanitizers/NodeDataSanitizerTestWrapping.php

sebastianmarzjan:
@NoIdeaHowToCallMe
Sorry it took so long!
The file you are referring to is the right test file. However, based on the first comment regarding the overall direction for infoboxes product ( &lt;br&gt; tag is **not** to be passed directly to mobile view ) :
- &lt;br&gt; tags need to be sanitized and converted to whitespace; For example, sanitizer currently processes labels like "Previous&lt;br&gt;appearances" into "Previousappearances". This needs to be fixed, however the expected output is "Previous appearances" 

Also, when adding tests, please create new cases (add elements to the array in testSanitizeDataProvider) instead of modifying existing ones, unless it's needed for the tests to pass.

NoIdeaHowToCallMe:
@SebastianMarzjan 
Bummer! Thanks anyways. I think this thread can be closed now.

PR #10705, opened 2016-06-17
CE-3837: Introduce new modal size (content-size)

lukaszkonieczny:
We need modal which will fit into content page in all breakpoints.

@Wikia/spitfires


kamilkoterba:
LGTM

PR #10707, opened 2016-06-17
(CE-3844) Refactor fix for PI and parser tag fix

Grunny:
A security fix from MW core added quotes to the Parser's strip tag
markers in order to prevent an XSS vulnerability. When this was deployed
PIs and our Wikia tag extensions broke in Mercury because ArticlesAsJson
JSON encodes the parser output in the ParserAfterTidy, but then PI and
the tag extensions were manually replacing the Parser tag strip markers
on the already JSON encoded string within their own ParserAfterTidy methods
(which were always run after the ArticlesAsJson method). This meant that the
strip marker had the quote escaped with a backslash in the encoded string,
so the string replacements failed.

This change follows [the previous quick and dirty fix](https://github.com/Wikia/app/pull/10701) by moving the replacements
in PI and the tag extensions to just before ArticlesAsJson encodes the output.
This also removes the need from before the fix in #10701 to JSON encode the
parser output of the tags that will be replaced into the text.

/cc @Wikia/west-wing @rogatty

### DIFF
--- a/extensions/wikia/PortableInfobox/controllers/PortableInfoboxParserTagController.class.php
+++ b/extensions/wikia/PortableInfobox/controllers/PortableInfoboxParserTagController.class.php
@@ -50,7 +50,10 @@ class PortableInfoboxParserTagController extends WikiaController {
 	 * @return string
 	 */
 	public static function replaceInfoboxMarkers( &$parser, &$text ) {
-		$text = static::getInstance()->replaceMarkers( $text );
+		global $wgArticleAsJson;
+		if ( !$wgArticleAsJson ) {
+			$text = static::getInstance()->replaceMarkers( $text );
+		}
 
 		return true;
 	}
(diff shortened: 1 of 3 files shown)


idradm:
LGTM

extensions/wikia/PortableInfobox/controllers/PortableInfoboxParserTagController.class.php
```diff
@@ -50,7 +50,10 @@ public static function parserTagInit( Parser $parser ) {
 	 * @return string
 	 */
 	public static function replaceInfoboxMarkers( &$parser, &$text ) {
-		$text = static::getInstance()->replaceMarkers( $text );
+		global $wgArticleAsJson;
+		if ( !$wgArticleAsJson ) {
```
rogatty, line 55: Please add the info that it's done in `PortableInfoboxHooks::onArticleAsJsonBeforeEncode` for Mercury.

rogatty:
Great job, LGTM!

extensions/wikia/PortableInfobox/controllers/PortableInfoboxParserTagController.class.php
Grunny, line 55: Good point, comment added. :)

PR #10711, opened 2016-06-20
JPN-552 Add transparent black background for light theme

perjg:
JPN-552 Add transparent black background for light theme default community page image. Left dark theme default image alone, as it already looks good with white text.

@Wikia/spitfires @d34th4ck3r


Grunny:
LGTM.

PR #10716, opened 2016-06-21
JPN-570 Fix community page overlay for Firefox

perjg:
top: 0; is needed for correct placement in Firefox.

@Wikia/spitfires @d34th4ck3r


Grunny:
:+1: LGTM.

PR #10718, opened 2016-06-21
JPN-553 Modal for first time editors

perjg:
A modal shown whenever a user makes his first edit on a community.

Tracked using a cookie in order to make it work seamlessly both for VE and CE:
- With VE, there is no page reload, but a JS hook callback when the edit completes.
- With CE, a new page is loaded.

The cookie is set in onUserFirstEditOnLocalWiki, and then immediately deleted after being read on the client side.

@Wikia/spitfires @d34th4ck3r


extensions/wikia/CommunityPage/scripts/firstEditModal.js
```diff
@@ -0,0 +1,130 @@
...
+	 });
+
+	function getUiModalInstance() {
+		var $deferred = $.Deferred();
+
+		uiFactory.init(['modal']).then(function (uiModal) {
+			$deferred.resolve(uiModal);
+		});
+
+		return $deferred;
```
Grunny: Since `uiFactory.int` returns a promise, couldn't you just `return uiFactory.init(['modal']);` here rather than creating another promise object?

extensions/wikia/CommunityPage/scripts/firstEditModal.js
```diff
@@ -0,0 +1,130 @@
...
+			data: {
+				uselang: window.wgUserLanguage
+			},
+			format: 'json',
+			type: 'get',
+		}).then(function (response) {
+			$deferred.resolve(response);
+		});
+
+		return $deferred;
```
Grunny: Likewise, since you're not manipulating the response before resolving, couldn't this just return the `sendRequest` as that itself returns a Promise?

extensions/wikia/CommunityPage/scripts/firstEditModal.js
```diff
@@ -0,0 +1,130 @@
...
+		var cookie = cookies.get(cookieName);
+
+		// If cookie is set on load, show dialog, this happens after CE edit
+		if (cookie) {
+			openModal();
+
+			// Delete the cookie
+			cookies.set(cookieName, null, {
+				domain: wgCookieDomain,
+				path: '/',
```
Grunny: `mw.config.get('wgCookiePath')`

extensions/wikia/CommunityPage/scripts/firstEditModal.js
```diff
@@ -0,0 +1,130 @@
...
+		var cookieName = 'community-page-first-time';
+		var cookie = cookies.get(cookieName);
+
+		// If cookie is set on load, show dialog, this happens after CE edit
+		if (cookie) {
+			openModal();
+
+			// Delete the cookie
+			cookies.set(cookieName, null, {
+				domain: wgCookieDomain,
```
Grunny: `mw.config.get('wgCookieDomain')`

extensions/wikia/CommunityPage/scripts/firstEditModal.js
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	function initModalTracking() {
+		// Track clicks in modal
+		$('#CommunityPageFirstEditModal').on('mousedown touchstart', 'a', function (event) {
+			handleClick(event);
+		});
+	}
+
+	function handleClick (event) {
```
Grunny: Remove space between function name and opening parenthesis.

extensions/wikia/CommunityPage/scripts/firstEditModal.js
```diff
@@ -0,0 +1,130 @@
...
+		if (label !== null && label.length > 0) {
+			track({
+				label: label,
+			});
+		}
+	}
+
+	function openModalOnFirstEdit() {
+		var cookieName = 'community-page-first-time';
+		var cookie = cookies.get(cookieName);
```
Grunny: Our JS conventions are currently [one var declaration per scope](https://github.com/Wikia/guidelines/blob/master/JavaScript/CodingConventions.md#variables).

extensions/wikia/CommunityPage/scripts/firstEditModal.js
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	function getModalContents() {
+		var $deferred = $.Deferred();
+
+		nirvana.sendRequest({
+			controller: 'CommunityPageSpecial',
+			method: 'getFirstTimeEditorModalData',
+			data: {
+				uselang: window.wgUserLanguage
```
Grunny: `mw.config.get('wgUserLanguage')`

extensions/wikia/CommunityPage/CommunityPageSpecialHooks.class.php
```diff
@@ -64,6 +65,9 @@ public static function onArticleSaveComplete(
 	public static function onBeforePageDisplay( \OutputPage $out, \Skin $skin ) {
 		\Wikia::addAssetsToOutput( 'community_page_benefits_js' );
 		\Wikia::addAssetsToOutput( 'community_page_benefits_scss' );
+		\Wikia::addAssetsToOutput( 'community_page_new_user_modal_js' );
+		\Wikia::addAssetsToOutput( 'community_page_new_user_modal_scss' );
```
Grunny: Do both of these need to be loaded on every single page view or only when the cookie is set? Also, from the ticket, it looks like this is only for logged in users, so can the assets be restricted to only load for logged-in users?

extensions/wikia/CommunityPage/CommunityPageSpecialHooks.class.php
```diff
@@ -107,6 +111,10 @@ public static function onUserFirstEditOnLocalWiki( $userId, $wikiId ) {
 		WikiaDataAccess::cachePurge( $key );
 		CommunityPageSpecialUsersModel::logUserModelPerformanceData( 'purge', 'recently_joined' );
 
+		// Set cookie to show first edit modal to user
+		global $wgCookieDomain;
```
Grunny: I'd put the `global` at the top of the method as a standard.

extensions/wikia/CommunityPage/CommunityPageSpecialHooks.class.php
```diff
@@ -107,6 +111,10 @@ public static function onUserFirstEditOnLocalWiki( $userId, $wikiId ) {
...
+		setcookie ( self::FIRST_EDIT_COOKIE_KEY, true, time()+60, '/', $wgCookieDomain );
```
Grunny: Remove the space between the function name and the opening parenthesis.

Also, the path should be set to `$wgCookiePath` rather than hardcoded as `/`.

extensions/wikia/CommunityPage/scripts/firstEditModal.js
```diff
@@ -0,0 +1,130 @@
...
+					.html(mustache.render(templates.firstEditModal, {
+						heading: modalContents.headingText,
+						subheading: modalContents.subheadingText,
+						getStarted: modalContents.getStartedText,
+						maybeLater: modalContents.maybeLaterText,
+						getStartedLink: modalContents.getStartedLink,
+					}));
+
+				modal.show();
+				initModalTracking(modal);
```
Grunny: `initModalTracking` currently doesn't use a parameter but it is called with one here. Did you mean to pass the modal (or even `modal.$element`)?

perjg:
@Grunny Cleaned up according to your comments. I changed the assets to not load for anon users.

extensions/wikia/CommunityPage/CommunityPageSpecialHooks.class.php
```diff
@@ -71,6 +72,12 @@ public static function onBeforePageDisplay( \OutputPage $out, \Skin $skin ) {
 			\Wikia::addAssetsToOutput( 'community_page_benefits_js' );
 			\Wikia::addAssetsToOutput( 'community_page_benefits_scss' );
 		}
+
+		if ( !$out->getUser()->isAnon() ) {
```
lukaszkonieczny: I think you should check is `community-page-first-time` cookie set here to avoid unnecessary assets load

perjg: @lukaszkonieczny I can't check if the cookie is set here, because it will break the functionality for VE. With VE the workflow goes like this:
- Inject assets on page load
- On PHP side, set cookie in onUserFirstEditOnLocalWiki
- mw.hook('postEdit') is fired, check for cookie there and unset if necessary.

Grunny:
LGTM.

perjg:
@Grunny Pushed a commit for JPN-580 to this PR, as JPN-580 requires this work as a base.

See also https://github.com/Wikia/config/pull/1840 for the config change required.

extensions/wikia/CommunityPage/CommunityPageSpecialHooks.class.php
```diff
@@ -73,7 +73,7 @@ public static function onBeforePageDisplay( \OutputPage $out, \Skin $skin ) {
 			\Wikia::addAssetsToOutput( 'community_page_benefits_scss' );
 		}
 
-		if ( !$out->getUser()->isAnon() ) {
+		if ( !$out->getUser()->isAnon() && !$out->getUser()->isAllowed( 'first-edit-dialog-exempt' )) {
```
Grunny: I'd suggest putting `$out->getUser()` in a variable since it's used three times in this method now. Also, missing a space before the closing parenthesis.

Grunny:

Grunny:
:+1:

extensions/wikia/CommunityPage/styles/firstedit/FirstEditModal.scss
```diff
@@ -0,0 +1,66 @@
...
+
+	.community-page-header-image {
+		background: url($header-image) no-repeat center;
+		height: 80px;
+	}
+
+	.community-page-first-edit-heading {
+		font-size: 18px;
+		font-weight: bold;
+		padding: 20px 20px;
```
Grunny: Can just be `padding: 20px;`.

extensions/wikia/CommunityPage/scripts/firstEditModal.js
```diff
@@ -0,0 +1,125 @@
...
+				path: mw.config.get('wgCookiePath'),
+			});
+		}
+	}
+
+	$(function() {
+		openModalOnFirstEdit();
+
+		// Hook to show modal after VE edit completes (no page reload)
+		mw.hook('postEdit').add( function() {
```
Grunny: Remove the space after the opening parenthesis of add, and add a space after the function keyword.

extensions/wikia/CommunityPage/scripts/firstEditModal.js
```diff
@@ -0,0 +1,125 @@
...
+			openModalOnFirstEdit();
+		});
+
+		// Hook to show modal after comment (no page reload)
+		mw.hook('wikipage.content').add( function() {
```
Grunny: Remove the space after the opening parenthesis of add, and add a space after the function keyword.

Grunny:
Minor comments, otherwise LGTM.

perjg:
@Grunny Addressed your new comments.

Grunny:
:+1:

PR #10722, opened 2016-06-21
CE-3851 Wrap words in right rail entrypoint module

kamilkoterba:
https://wikia-inc.atlassian.net/browse/CE-3851
## Description

`word-wrap: break-word` wasn't working properly for the entry point module and it turned out that flexbox is breaking it. Fix could be to add width on flexbox element then longer words start to be wrapped properly.

Although instead I decided to simplify stylling here and remove flex stylling which also fixed word wrapping.

@Wikia/spitfires


lukaszkonieczny:
LGTM

PR #10727, opened 2016-06-22
CE-3809 Add tracking for benefits modal

kamilkoterba:
https://wikia-inc.atlassian.net/browse/CE-3809

@Wikia/spitfires


extensions/wikia/CommunityPage/scripts/CommunityPageBenefitsModal.js
```diff
@@ -57,12 +67,36 @@ define('CommunityPageBenefitsModal',
...
 			modalInstance.show();
+
+			// Send tracking event for modal shown
+			track({
+				action: tracker.ACTIONS.IMPRESSION,
+				label: 'benefits-modal-shown'
+			});
+
+			// Bind tracking links that navigate away
+			modalInstance.$element.find('[data-track-mousedown]').bind('mousedown', function (e) {
```
Grunny: `.on` has been preferred over `.bind` since jQuery 1.7.

extensions/wikia/CommunityPage/scripts/CommunityPageBenefitsModal.js
```diff
@@ -57,12 +67,36 @@ define('CommunityPageBenefitsModal',
...
+				track({label: $(e.target).data('trackMousedown')});
+			});
+
+			// Bind tracking clicks on elements with data-track attribute
+			modalInstance.$element.find('[data-track]').click(function (e) {
```
Grunny: Can this just be unified with the one above? For every `.click` a `mousedown` already happened, so I don't think this needs to be handled separately, just have a single handler.

lukaszkonieczny: Maybe we could simplify this and bind `mousedown` to all elements which need tracking?

Grunny: @lukaszkonieczny yep, that's what I meant. :)

kamilkoterba: I don't see the need to track mousedowns for elements that don't navigate away. Click is more precise. Here we won't have duplication as clicks and mousedowns are pinned to different elements.

@Grunny, @lukaszkonieczny would you agree or still think it's worth unifying all to mousedowns?

lukaszkonieczny: @Grunny I didn't refresh this PR before making a comment ;)
We will get the same information, regardless we go with clicks or mousedowns so simplifying code make sense for me

Grunny: I agree with Lukasz. :)

kamilkoterba: ok, let it be ;)

extensions/wikia/CommunityPage/scripts/CommunityPageBenefitsModal.js
```diff
@@ -57,12 +67,31 @@ define('CommunityPageBenefitsModal',
...
+				label: 'benefits-modal-shown'
+			});
+
+			// Bind tracking on elements with data-track attribute
+			modalInstance.$element.find('[data-track]').on('mousedown', function (e) {
+				track({label: $(e.target).data('track')});
+			});
+
+			// Bind tracking modal close
+			modalInstance.bind('close', function () {
```
Grunny: Likewise, `.bind` should be `.on`.

Grunny, line 89: `.on`

kamilkoterba, line 89: modalInstance is custom Wikia object that doesn't provide .on method
https://github.com/Wikia/app/blob/2d75c4bedd478d3746aaeb48e10b0440c7fa6359/resources/wikia/ui_components/modal/js/modal.js#L325-L325

Grunny, line 89: Bah, was on autopilot. ;)

Grunny:
LGTM.

extensions/wikia/CommunityPage/scripts/CommunityPageBenefitsModal.js
kamilkoterba, line 89: I knew it's not real you working in all timezones ;)

PR #10728, opened 2016-06-22
DAT-4317 | fixing navigation template handling on mercury

rybmat:
https://wikia-inc.atlassian.net/browse/DAT-4317
https://wikia-inc.atlassian.net/browse/DAT-4461

fixing issues, when wikitext was displayed on mercury articles in place where navigation template was inserted

@Wikia/west-wing


includes/wikia/parser/templatetypes/tests/NavigationTemplateTest.php
```diff
@@ -35,7 +36,7 @@ public function getNavigationTemplates() {
 		return [
 			[
 				'<a>This is a <strong>template</strong> <b>without</b> a <span>block</span> element</a>.',
-				'<a>This is a <strong>template</strong> <b>without</b> a <span>block</span> element</a>.',
+				"\n<a>This is a <strong>template</strong> <b>without</b> a <span>block</span> element</a>.\n",
```
rybmat: this is side effect of the fix, we should decide if it is ok or we should fix it somehow.

sebastianmarzjan: Yes, we need to keep the output clean - if there's no removal, the markup has to stay intact

includes/wikia/parser/templatetypes/handlers/NavigationTemplate.class.php
```diff
@@ -85,6 +85,18 @@ private static function replaceMarker( $marker, $html ) {
...
-		return sprintf( "<%s>%s</%s>", $marker, $text, $marker );
+		return sprintf( "<%s>\n%s\n</%s>", $marker, $text, $marker );
+	}
+
+	public static function removeInnerMarks( $templateWikitext ) {
+		$openMarkerRegex = "[<&lt;]\x7f" . self::MARK . "[^\x7f]+\x7f[>&gt;]\\n";
+		$closeMarkerRegex = "\\n[<&lt;]\\/\x7f" . self::MARK . ".+\x7f[>&gt;]";
+
+		if (preg_match( "/(" . $openMarkerRegex . ")(.*)(" . $closeMarkerRegex . ")/is", $templateWikitext, $inside )) {
+			$replacedOpenings = preg_replace( "/" . $openMarkerRegex . "/isU", "", $inside[2] );
```
idradm: Should new lines be removed here as well? What if regex fails? Is it possible to make those replaces in one?

includes/wikia/parser/templatetypes/TemplateTypesParser.class.php
```diff
@@ -116,7 +116,7 @@ public static function onGetTemplateDom( $title, $args, $frame, &$outputText ) {
 	 * @return bool
 	 */
 	public static function onEndBraceSubstitution( $templateTitle, &$templateWikitext, &$parser ) {
-		global $wgEnableContextLinkTemplateParsing, $wgEnableInfoIconTemplateParsing;
+		global $wgEnableContextLinkTemplateParsing, $wgEnableInfoIconTemplateParsing, $wgEnableNavigationTemplateParsing;
```
sebastianmarzjan, line 119: This looks partially redundant with https://github.com/Wikia/app/blob/dev/includes/parser/Parser.php#L3761-L3766, plus in the linked lines there's no reference to $wgEnableNavigationTemplateParsing (and should be if we want to have checks in both places)

includes/wikia/helpers/HtmlHelper.class.php
```diff
@@ -69,6 +69,23 @@ public static function removeNodes( Array $nodes ) {
 	}
 
 	/**
+	 * Removes wrapping node
+	 * eg.: "<div><p>test</p> <a>more test</a></div>" -> "<p>test</p> <a>more test</a>"
+	 *
+	 * @param DOMNode $node
+	 * @return DOMNode removed node
+	 */
+	public static function removeWrappingNode( DOMNode $node ) {
```
sebastianmarzjan: Can we call it similar to BS, like unwrapNode?

sebastianmarzjan:
I like this approach a lot better than the regex-based one. LGTM, though the test failure seems fishy - please check it before merging

includes/wikia/parser/templatetypes/TemplateTypesParser.class.php
idradm, line 119: @dianafa Is there a reason for this duplicated check? If not I will remove the one located in `Parser`.

sebastianmarzjan:
LGTM

PR #10729, opened 2016-06-22
RECIRC 153: Remove internal state from Recirculation controllers

pauloslund:
We now pass in the wiki id with each JS request, allowing us to use wgCdnApiUrl.


owend:
LGTM, varnish rules have been changed so this can be tested now, which we should do before deploy

gabrys:
LGTM

pauloslund:
Tested and working with new varnish rules

PR #10731, opened 2016-06-22
SUS-649 handle longMessage event in a user-friendly manner

ghost:
## Description

Up until today, maximum length of a chat message has been set [on the client](https://github.com/Wikia/app/blob/dev/extensions/wikia/Chat2/js/controllers/controllers.js#L121). Then somebody managed to set it to `infinity` and started posting lengthy messages with lots of emojis which killed all clients' performance.

The point is to introduce maximum limit, which I did in the model. And then I applied the limit in the server code. Finally, I added user-friendly handling of the event on the client side.
## Ticket & other links
- http://wikia-inc.atlassian.net/browse/SUS-649
- https://github.com/Wikia/chat/pull/30
- https://github.com/Wikia/app/pull/10731
## Reviewers

@nmonterroso @owend @wladekb @pchojnacki @dianafa , whoever gets here first :-)

### DIFF
--- a/extensions/wikia/Chat2/js/controllers/controllers.js
+++ b/extensions/wikia/Chat2/js/controllers/controllers.js
@@ -389,7 +389,7 @@ var NodeRoomController = $.createClass(Observable, {
 	},
 
 	onLongMessage: function (message) {
-		if (typeof message.user == 'string' && message.user == wgUserName) {
+		if (message.user === wgUserName) {
 			var newChatEntry = new models.InlineAlert({text: mw.message('chat-message-was-too-long').escaped()});
 			this.model.chats.add(newChatEntry);
 		} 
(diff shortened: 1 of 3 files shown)


ghost:
The tests failed because of the instability of the test environment. :(

```
Running 'PortableInfoboxRenderServiceTest'.................Build timed out (after 10 minutes). Marking the build as aborted.
Build was aborted
```

extensions/wikia/Chat2/js/controllers/controllers.js
```diff
@@ -386,6 +388,13 @@ var NodeRoomController = $.createClass(Observable, {
 		}), 10000);
 	},
 
+	onLongMessage: function (message) {
+		if (typeof message.user == 'string' && message.user == wgUserName) {
```
macbre: You can use `===` operator that will check type as well, hence `typeof` won't be needed (assuming `wgUserName` is a string)

macbre:
One minor comment left, looks good

extensions/wikia/Chat2/js/controllers/controllers.js
```diff
@@ -386,6 +388,13 @@ var NodeRoomController = $.createClass(Observable, {
 		}), 10000);
 	},
 
+	onLongMessage: function (message) {
+		if (message.user === wgUserName) {
```
macbre, line 392: :+1:

PR #10733, opened 2016-06-23
JPN-573 Don't show hub pages in global nav for non-English languages

perjg:
@bognix @kvas-damian @vforge

### DIFF
--- a/extensions/wikia/GlobalNavigation/GlobalNavigationHelper.class.php
+++ b/extensions/wikia/GlobalNavigation/GlobalNavigationHelper.class.php
@@ -111,6 +111,10 @@ class GlobalNavigationHelper {
 			[3] // max 3 links
 		);
 
+		if ( $wgLang->getCode() !== 'en' ) {
+			$hubsNodes = [];
+		}
+
 		// Link to WAM - Top Communities
 		$exploreDropdownLinks[] = [
 			'text' => $WAMLinkLabel->plain(),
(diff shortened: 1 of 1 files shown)


extensions/wikia/GlobalNavigation/GlobalNavigationHelper.class.php
```diff
@@ -111,6 +111,10 @@ public function getMenuNodes() {
 			[3] // max 3 links
 		);
 
+		if ( $wgLang->getCode() !== 'en' ) {
```
kvas-damian: What do you think about moving this `if` higher and don't fetch links when they aren't required?

perjg: You mean like this:

`
        if ( $wgLang->getCode() === 'en' ) {
            $hubsNodes = (...);
        } else {
            $hubsNodes = [];
        }
`
The reason I went with overriding for non-English as opposed to the way I pasted above is that other code in the file seems to prefer that method:

I see `if ( $lang !== 'en' )` and `if ( $lang !== self::DEFAULT_LANG)` other places in the file.

I'm fine with changing it, so please let me know if I should choose the newly pasted code or stick with the convention in the file.

PS: Regardless of that, I will change 'en' to self::DEFAULT_LANG, also in getWAMLinkForLang since it is inconsistent there compared to the rest of the file.

kvas-damian: Good catch with using `self::DEFAULT_LANG`

Yeah, I was thinking exactly like you pasted. Avoiding unnecessary computations makes code faster, so let's go that way.

vforge:
LGTM

perjg:
@kvas-damian Changed the if test like you suggested, and also changed to use self:: DEFAULT_LANG in two places.

kvas-damian:
LGTM

PR #10739, opened 2016-06-23
RECIRC-133: Setup test using Fandom hero units

pauloslund:
This primarily changes the e3 test into a more generic one that uses the Fandom API to grab vertical specific hero units.


gabrys:
LGTM

PR #10741, opened 2016-06-24
JPN-574 | Setting the default landing page as Community Page for registered User.

d34th4ck3r:
@Wikia/spitfires @perjg


extensions/wikia/MyHome/MyHome.class.php
```diff
@@ -181,14 +181,18 @@ public static function getInitialMainPage(Title &$title) {
...
 		if( $wgUser->isLoggedIn() && empty($wgEnableWikiaHomePageExt) ) {
 			$value = $wgUser->getGlobalPreference(UserPreferencesV2::LANDING_PAGE_PROP_NAME);
-			switch($value) {
-				case UserPreferencesV2::LANDING_PAGE_WIKI_ACTIVITY:
-					$title = SpecialPage::getTitleFor('WikiActivity');
-					break;
-				case UserPreferencesV2::LANDING_PAGE_RECENT_CHANGES:
-					$title = SpecialPage::getTitleFor('RecentChanges');
-					break;
+			if ( $value == UserPreferencesV2::LANDING_PAGE_WIKI_ACTIVITY ) {
```
lukaszkonieczny, line 184: Is there any reason you replaced `switch` by `if` statements?

d34th4ck3r, line 184: Not particularly. But I need more customized condition than simple switch case. Other option was to use if-else inside a switch case, which didn't seem to follow a specific standard. So I choose to change the whole statement.

lukaszkonieczny, line 184: It's pretty minor thing here but in my opinion is cleaner what's happening in `switch case` statement. And to avoid `if else` inside you can use ternary operator or create separate method to get proper title. But it's up to you :)

lukaszkonieczny:
LGTM

d34th4ck3r:
@Wikia/spitfires @perjg 

Added crowdin translation

Grunny:
LGTM.

PR #10744, opened 2016-06-24
Auth | Check for presence of an object

bkoval:
@kvas-damian This is sometimes throwing a js error just before redirecting to the standalone auth page.


kvas-damian:
LGTM

PR #10749, opened 2016-06-27
JPN-576 Community to-do list

perjg:
@Wikia/spitfires @d34th4ck3r


extensions/wikia/CommunityPage/models/CommunityPageSpecialCommunityTodoListModel.class.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+class CommunityPageSpecialCommunityTodoListModel {
+	public function getData() {
+		$title = Title::newFromText(
+			wfMessage( 'communitypage-todo-module-page-name' )->inContentLanguage()->plain(),
```
Grunny: What do you think of just making use of `wfMessage` and parse the contents with an output format? Also, having a customisable page name for this functionality will be difficult as each permutation would need to be whitelisted so admins could edit it. For this purpose, I'd suggest just making `Community-to-do-list` the page name and not have another message control that (MediaWiki pages have always been in English anyway).

Then, this method could just be something like (with any other functionality you need that can be added):

``` php
$message = wfMessage( 'community-to-do-list' );
$data = [
    'haveContent' => false,
];

if ( !$message->isDisabled() ) {
    $data['haveContent'] = true;
    $data['data'] = $message->parse();
}

return $data;
```

perjg: @Grunny Your suggested alternate code works well, but I am now missing the edit link which was:

`'editUrl' => $title->getFullURL( ['action' => 'edit'] ),`

I tried `Title::newFromText(` $message->plain(), NS_MEDIAWIKI );` to get the Title, but it just gives null.

Any suggestions on how to get the proper edit url in the context of your suggested code?

Grunny: I'd just do something like:

``` php
$messageKey = 'community-to-do-list';
$message = wfMessage( $messageKey );
$title = Title::newFromText( $messageKey, NS_MEDIAWIKI );
$data = [
    'haveContent' => false,
    'editUrl' => $title->getFullURL( ['action' => 'edit'] ),
];

if ( !$message->isDisabled() ) {
    $data['haveContent'] = true;
    $data['data'] = $message->parse();
}

return $data;
```

The message key could also be a constant on the class if it might need to be checked outside this method. What do you think?

perjg:
@grunny: I've updated the code with your suggestion for getData implementation, it works great.

(my mistake was that I tried to call Title::newFromText from a wfMessage rather than from a string).

extensions/wikia/CommunityPage/models/CommunityPageSpecialCommunityTodoListModel.class.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+class CommunityPageSpecialCommunityTodoListModel {
+	public function getData() {
+		$messageKey = 'community-to-do-list';
+		$message = wfMessage( $messageKey );
+		$title = Title::newFromText( $messageKey, NS_MEDIAWIKI );
+		$data = [
+			'haveContent' => false,
+			'editUrl' => $title->getFullURL( ['action' => 'edit'] ),
```
Grunny: Minor: Spaces needed within square brackets per conventions.

extensions/wikia/CommunityPage/scripts/ext.communityPage.js
```diff
@@ -251,6 +251,15 @@ require([
...
 			}
 		});
+
+		// Track clicks in the To-do List module
+		$('.community-page-todo-list-module-edit').on('mousedown touchstart', 'a', function (event) {
+			handleClick(event, 'community-page-todo-list-module-edit');
+		});
+
+		$('.community-page-todo-list-module-content').on('mousedown touchstart', 'a', function (event) {
+			handleClick(event, 'community-page-todo-list-content');
```
Grunny: I'd consider unifying these two tracking events into one, similar to what we do elsewhere in this code and use either a label or class to determine the tracking label.

Grunny:
Minor comments, otherwise LGTM. :)

perjg:
@Grunny I changed the tracking implementation a bit, but I kept the two separate selectors as I need to treat the data tag for the "edit" link and the actual todo list differently.

I also made a couple of other changes:
- Removed the link for "contact an admin" - it was based on my misunderstanding of the AC, there should be no link
- Added max-height 500px with scrollbar for todo list content.

extensions/wikia/CommunityPage/scripts/ext.communityPage.js
```diff
@@ -251,6 +256,18 @@ require([
...
+
+		// Track clicks in the To-do List module
+		$('.community-page-todo-list-module-edit').on('mousedown touchstart', 'a', function (event) {
+			handleClick(event, 'community-page-todo-list-module');
+		});
+
+		$('.community-page-todo-list-module-content').on('mousedown touchstart', 'a', function (event) {
+			track({
+				category: 'community-page-todo-list-module',
+				label: event.delegateTarget.getAttribute('data-tracking'),
```
Grunny: Ah, I see, since that's the case, I probably wouldn't bother with using the HTML data attribute for this label, as you're not going to use more than one, right?

extensions/wikia/CommunityPage/scripts/ext.communityPage.js
```diff
@@ -220,6 +220,11 @@ require([
 				category: category,
 				label: label,
 			});
+
+			console.log({
```
Grunny: Test code.

extensions/wikia/CommunityPage/scripts/ext.communityPage.js
perjg: Yeah, the AC specifically say to use the same label for all content inside the .community-page-todo-list-module-content div, so I will drop the data-tracking attribute for that one and just hardcode the data to send.

Grunny:
:+1:

Grunny:
:+1:

perjg:
@Grunny A few more changes pushed to account for changes to the Acceptance Criteria.

Grunny:
LGTM.

PR #10750, opened 2016-06-27
PLATFORM-2284: fix unit tests in PHP7

macbre:
`opcache` is here to blame, disable it via command line option when running unit tests -    f4d6f2c

@wladekb / @artursitarski


macbre:
Merged into preview branch as well

PR #10753, opened 2016-06-27
ADEN-3463 Hide spotlights for UAP

drets:
Hide spotlights according to _Universal Ad Product_ specification.

### DIFF
--- a/extensions/wikia/AdEngine/js/template/bfaa.js
+++ b/extensions/wikia/AdEngine/js/template/bfaa.js
@@ -45,7 +45,7 @@ define('ext.wikia.adEngine.template.bfaa', [
 		},
 
 		show: function (iframe) {
-			var spotlightFooter = document.getElementById('SPOTLIGHT_FOOTER');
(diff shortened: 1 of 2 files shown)


extensions/wikia/AdEngine/js/template/bfaa.js
```diff
@@ -56,6 +57,10 @@ define('ext.wikia.adEngine.template.bfaa', [
 			if (win.WikiaBar) {
 				win.WikiaBar.hideContainer();
 			}
+
+			if (spotlightFooter) {
+				slotTweaker.hideByElement(spotlightFooter.parentNode, true);
```
fraszczakszymon: Is it worth to use **Slot**Tweaker for that?

jacekjursza:
One, not related  concern: I don't like the fact that we use div IDs directly, it would be useful to have it abstracted (as we started doing so in Fandom via SlotService). But that is more general thought, not to be addressed now. 

LGTM

PR #10762, opened 2016-06-28
PLATFORM-2275: SiteStats performance improvements

macbre:
[PLATFORM-2275](https://wikia-inc.atlassian.net/browse/PLATFORM-2275)

Updates on `site_stats` table are causing serious performance issues when they're issued in parallel during heavy edits / uploads periods.

In order to resolve the root-cause let's do the following (inspired by MediaWiki 1.120 and 1.26 changes):
- make  `SiteStatsUpdate` use locking mechanism to avoid performing update queries in parallel
- avoid direct update queries in `LocalFile` - use `DeferredUpdates::addUpdate( SiteStatsUpdate::factory( [ 'images' => x ] ) );` instead
- actually do an update on database only every _n_ time (`wgSiteStatsAsyncFactor` will be set to an appriopriate value), use memcache to accumulate "deltas" and perform stats updates from several operations in one go
- make `SiteStats` getters use a cache - currently each call to `SiteStats` performs a select query.
- because not persistent memcache is used for storing deltas let's run the full update of `site_stats` table as a part of `updateSpecialPages` maintenance script (run daily for active wikis)

Let's start with `$wgSiteStatsAsyncFactor = 1` - we'll update `site_stats` on each edit / upload, but protect against parallel updates (and use memcache deltas in such cases).

Pinging @drozdo


includes/SiteStats.php
```diff
@@ -39,45 +46,29 @@ static function load( $recache = false ) {
...
-			wfInitStats();
-			wfGetDB( DB_MASTER )->commit();
-			ob_end_clean();
-
-			$row = self::doLoad( wfGetDB( DB_MASTER ) );
-		}
+		# Wikia change
+		return WikiaDataAccess::cache(
+			self::getMemcKey(),
+			WikiaResponse::CACHE_STANDARD,
```
wladekb: maybe `CACHE_LONG` ?

wladekb:
lgtm

includes/SiteStats.php
macbre: Modified in 47b861f.

PR #10764, opened 2016-06-28
Sus 648 - update ListUsers group selection

nmonterroso:
@mixth-sense 
https://wikia-inc.atlassian.net/browse/SUS-648

The problem with group numbers being out of sync between the group selection and the group result is the two being created from different queries.


ghost:
:+1: LGTM

ghost:
Tested and accepted. :-)

PR #10768, opened 2016-06-30
ADEN-3509 Fix collapsed BOTTOM_LEADERBOARD

nandy-andy:
New slot, when collapsed, did not have `wikia-ad` class and was adding empty space at the bottom of our pages. Using the regular method to generate slots helped but also expanding current functionality to make it possible for a slot not to push slot to the ad queue right away.


skins/oasis/modules/templates/Footer_Index.php
```diff
@@ -1,5 +1,5 @@
 <footer id="WikiaFooter" class="WikiaFooter <?= $showToolbar ? '' : 'notoolbar' ?>">
-	<div id="BOTTOM_LEADERBOARD"></div>
+	<?= F::app()->renderView('Ad', 'Index', ['slotName' => 'BOTTOM_LEADERBOARD', 'pageTypes' => ['homepage_logged', 'corporate', 'search', 'all_ads']]); ?>
```
drets: It will `push` the slot from the beginning which is not OK in bottom leaderboard case.

nandy-andy: Good catch :+1: Thanks for thinking of it :)

drets: Feel free to run `ads-uap-oasis` job on your code changes :wink:

drets:
LGTM :+1:

PR #10769, opened 2016-06-30
PLATFORM-2235:improve migrateImagesBetweenSwiftDC

macbre:
[PLATFORM-2235](https://wikia-inc.atlassian.net/browse/PLATFORM-2235)

Part of delete operations fail to propagate to Reston DFS nodes. Let's add logging of all pushes to the images sync queue from the app and check that we're actually syncing a successful Swift operation.

Tag log messages with `SwiftSync` to be able to easily correlate pushes and queue consumption.

@wladekb


wladekb:
lgtm

PR #10787, opened 2016-07-05
(DAT-4579) Remove action=render from the TabView tab links

Grunny:
The TabView tab links include action=render which leads to a strange
experience when clicking the links in Mercury and if you open the tab
link in a new tab. This removes action=render from the link URL and
instead adds it to the AJAX request when loading the tab content.

/cc @Wikia/x-wing


hakubo:
LGTM

rogatty:
LGTM

PR #10790, opened 2016-07-05
ADEN-3483 Sync UAP on Mercury

drets:
App part for https://github.com/Wikia/mercury/pull/2591

### DIFF
--- a/extensions/wikia/AdEngine/js/provider/gpt/helper.js
+++ b/extensions/wikia/AdEngine/js/provider/gpt/helper.js
@@ -134,6 +134,7 @@ define('ext.wikia.adEngine.provider.gpt.helper', [
 	adContext.addCallback(function () {
 		if (googleApi.isInitialized()) {
 			googleApi.setPageLevelParams(adLogicPageParams.getPageLevelParams());
+			uapContext.reset();
 		}
 	});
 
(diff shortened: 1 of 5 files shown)


extensions/wikia/AdEngine/js/provider/gpt/helper.js
```diff
@@ -76,9 +76,8 @@ define('ext.wikia.adEngine.provider.gpt.helper', [
 				slotTargeting.rv = count.toString();
 			}
 		}
-		if (uapId) {
-			slotTargeting.uap = uapId.toString();
-		}
+
+		slotTargeting.uap = uapId ? uapId.toString() : 'none';
```
fraszczakszymon, line 80: Why do we need this `none`?

Edit: OK, I think I know why - to reset value on consecutive pageviews (?). But slotTargeting is copied (here: https://github.com/Wikia/app/pull/10790/files#diff-d0456a56fa8978685b777a3c10aaf5abR65) so is this resetting needed?

drets, line 80: Yes and yes.
(for some reasons gpt lib remembers what was set before and in order to update slot targeting parameter (in requested URL) we need to set a new not-falsy value)
Do you know a better way how to reset `uap` param on consecutive pageviews on Mercury?

nandy-andy, line 80: Yep, @fraszczakszymon you're our last resort. We were looking for a way to "reset" slot targeting with pure GPT but it seems there is no way. And resetting it by setting it to falsy values results in our wrapper and implementation working correctly but the request to DFP includes the old `uap` on consecutive PV in Mercury :/

nandy-andy:
LGTM :+1: Let's test it.

nandy-andy:
LGTM

PR #10791, opened 2016-07-06
PLATFORM-2303: Fix fatal error in feed provider/renderer

wladekb:
This change should fix the fatal error:

```
PHP Fatal Error: Call to a member function getText() on null in /extensions/wikia/MyHome/renderers/FeedRenderer.php on line 688
```

https://wikia-inc.atlassian.net/browse/PLATFORM-2303

/cc @macbre @garthwebb


macbre:
:+1: 

Do we know what's causing title creation to fail?

wladekb:
@macbre There is some more information in the ticket about the case. However the direct reason is that `wfFindFile()` must return a false-y value

macbre:
Ok, thanks for the details

PR #10798, opened 2016-07-07
(SEC-68) Prevent staff members from removing the restricted-login group

Grunny:
Also removes a duplicate entry from the global groups list and alphabetises it for easier reference.

/cc @jcellary @macbre


macbre:
:+1:

jcellary:
:+1:

PR #10802, opened 2016-07-07
SOC-1892 Authorize an internal request using a Schwartz token

slayful:
Right now the Email controller can only be called within MediaWiki or when there is a current user available who is an admin.

The notification service will use a Schwartz header to authorize itself to the Email service.

See SOC-1892 for more details and discussion.

### DIFF
--- a/includes/WebRequest.php
+++ b/includes/WebRequest.php
@@ -1134,6 +1134,23 @@ HTML;
 	public function isWikiaInternalRequest() {
 		return $this->getHeader( self::WIKIA_INTERNAL_REQUEST_HEADER ) !== false;
 	}
+
+	const WIKIA_INTERNAL_AUTHORIZED_REQUEST_HEADER = 'X-Wikia-Schwartz';
+
+	/**
+	 * Use X-Wikia-Schwartz to authorize an internal request using a Schwartz token,
+	 * e.g. was sent via Http::get() helper or made by DC warmer
+	 *
+	 * @see SOC-1892
+	 *
+	 * @return bool
+	 */
+	public function isWikiaInternalAuthorizedRequest() {
+		global $wgSchwartzToken;
+
+		return $this->getHeader( self::WIKIA_INTERNAL_AUTHORIZED_REQUEST_HEADER ) ===
+			$wgSchwartzToken;
+	}
 }
 
 /**
--- a/includes/WebRequest.php
+++ b/includes/WebRequest.php
@@ -1146,10 +1146,10 @@ HTML;
 	 * @return bool
 	 */
 	public function isWikiaInternalAuthorizedRequest() {
-		global $wgSchwartzToken;
+		global $wgTheSchwartzSecretToken;
 
 		return $this->getHeader( self::WIKIA_INTERNAL_AUTHORIZED_REQUEST_HEADER ) ===
-			$wgSchwartzToken;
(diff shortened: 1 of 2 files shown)


includes/WebRequest.php
```diff
@@ -1134,6 +1134,23 @@ public function isValidWriteRequest( \User $user ) {
 	public function isWikiaInternalRequest() {
 		return $this->getHeader( self::WIKIA_INTERNAL_REQUEST_HEADER ) !== false;
 	}
+
```
armonr, line 1137: Can you please add begin/end comments denoting that this is a Wikia change? This file is originally from MediaWiki, so conventionally, within MW files, in-house changes are marked. Here is an [Example](https://github.com/Wikia/app/blob/master/includes/WebRequest.php#L1098-L1106).

armonr:
Minor comment. LGTM!

includes/WebRequest.php
garthwebb, line 1137: Eh, I think we can let that practice go.  Before we forked (over two years ago) we added those lines so that it would make it easier to merge newer versions of MW into our highly customized build.  We kept it up for about a year after that just in case we needed to merge again, however now it just adds noise and no real value.

armonr, line 1137: Makes sense.

garthwebb:
The change mechanics look good and I'll pass this.  However, I still don't fully understand the difference between 'X-Wikia-Internal-Request' and 'X-Wikia-Schwartz' and why both are needed.  I feel like we could merge or remove one.

macbre:
I second @garthwebb opinion - `X-Wikia-Internal-Request` is all you need here (assuming you're making a local HTTP requests to our PHP backend that does not go via edge nodes - you should use our border nodes as HTTP proxy).

includes/WebRequest.php
macbre, line 1137: @garthwebb, we do backport changes from MW upstream. Such comments are really handy when solving conflicts.

garthwebb, line 1137: Its inconsistent then; the PLATFORM change just above this one (isWikiaInternalRequest) isn't marked with "Wikia Change" for example.  Is there some recommendation for this?  Looking through recent commits in non-wikia directories under `includes` I see several changes that don't include any Wikia markings.  Of those that do, I see many formats:
- Start and end pairs of comments
- A single note in the function comment header
- Inline comments at the end of a line
- Many wordings "Wikia start", "Wikia change begin", "Wikia:", A Jira ticket number etc.
- Some begin/ends that include a JIRA ticket but the update is for a different ticket number

includes/wikia/nirvana/WikiaApp.class.php
```diff
@@ -633,6 +636,10 @@ public function sendRequest( $controllerName = null, $methodName = null, $params
 
 		$request = new WikiaRequest($params);
 
+		if ( false === $internal ) {
+			$internal = F::app()->wg->Request->isWikiaInternalAuthorizedRequest();
```
garthwebb, line 640: @slayful, we still need this change, but it the above line should be changed to the existing header method:

``` php
$internal = F::app()->wg->Request-> isWikiaInternalRequest();
```

armonr:
@garthwebb, @slayful: should this be closed?

mszabo-wikia:
@slayful it seems we don't need this.

PR #10804, opened 2016-07-07
RECIRC-155: Setup fallback tracking for errors

pauloslund:
We want to make sure we are tracking impressions and clicks even when they are shown the fallback


katanka:
LGTM

PR #10808, opened 2016-07-11
SOC-2203 Discussion swagger generated client + Special page for Discussions

armonr:
- Addition of Swagger generated client for the Discussion service
- New Special page for enabling Discussions only accessible by staff


lib/Swagger/src/Discussion/Api/ContributionApi.php
```diff
@@ -0,0 +1,348 @@
+<?php
```
armonr, line 1: **\* BEGINNING OF SWAGGER GENERATED CLIENT FOR DISCUSSION SERVICE ***

lib/Swagger/src/Discussion/Models/UserInfo.php
```diff
@@ -0,0 +1,308 @@
...
+     */
+    public function __toString()
+    {
+        if (defined('JSON_PRETTY_PRINT')) { // use JSON pretty print
+            return json_encode(\Swagger\Client\ObjectSerializer::sanitizeForSerialization($this), JSON_PRETTY_PRINT);
+        }
+
+        return json_encode(\Swagger\Client\ObjectSerializer::sanitizeForSerialization($this));
+    }
+}
```
armonr, line 308: **\* END OF SWAGGER GENERATED CLIENT FOR DISCUSSION SERVICE ***

extensions/wikia/SpecialDiscussions/SpecialDiscussionsController.class.php
```diff
@@ -0,0 +1,165 @@
...
+		}
+	}
+
+	private function activateDiscussions() {
+		$this->assertValidPostRequest( $this->request, $this->getUser() );
+
+		$site = new \Swagger\Client\Discussion\Models\Site(
+			[
+				'id' => $this->siteId,
+				'language_code' => F::app()->wg->ContLang->getCode(),
```
slayful: Discussion's POJO uses camelCase, not sure if that's related.

lib/Swagger/src/Discussion/Api/ContributionApi.php
slayful, line 1: I'm wondering what's the policy for those clients. Can we generate the client only for one resource or simply remove the generated classes for unnecessary endpoints?

slayful, line 1: Another option would be to manually use an HTTPClient of some sort.

armonr, line 1: I think the preference has been to use swagger client. I did ask a few in the team here about whether I should just add the partial API client, but we decided that would be suboptimal for future usage by other teams such as CAKE.

extensions/wikia/SpecialDiscussions/SpecialDiscussionsController.class.php
```diff
@@ -0,0 +1,165 @@
...
+	private $logger;
+	private $sitesApi;
+	private $siteId;
+	private $siteName;
+
+	public function __construct() {
+		parent::__construct( 'Discussions', '', false );
+		$this->sitesApi = $this->getDiscussionSitesApi();
+		$this->siteId = F::app()->wg->CityId;
+		$this->siteName = WikiFactory::getVarValueByName( 'wgSitename', $this->siteId );
```
garthwebb: The value of `$this->siteId` just came from `F::app()->wg->CityId` so is there any reason why we can't get siteName from `F::app()->wg->Sitename`?  Calling WikiFactory here is going to make a DB call.

extensions/wikia/SpecialDiscussions/SpecialDiscussionsController.class.php
```diff
@@ -0,0 +1,165 @@
...
+	public function index() {
+		$this->checkAccess();
+
+		$this->setHeaders();
+		$this->response->addAsset( 'special_discussions_scss' );
+
+		$this->wg->Out->setPageTitle( wfMessage( 'discussions-pagetitle' )->escaped() );
+
+		$editToken = $this->getVal( self::EDIT_TOKEN );
+		if ( !empty( $editToken ) ) {
```
garthwebb: What is this check for?  If its to see if the request was POSTed, it would be better to test `$this->request->wasPosted()`

extensions/wikia/SpecialDiscussions/SpecialDiscussionsController.class.php
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	private function setIndexOutput() {
+		if ( $this->getIsDiscussionsActive() ) {
+			$this->response->setVal(
+				'discussionsLink',
+				$this->sendSelfRequest(
+					'discussionsLink',
+					[
+						'discussionsLink' => $this->getDiscussionsLink(),
```
garthwebb: It doesn't look like this parameter, nor the parameter passed into the `inputForm` method on L91 are used in the methods themselves.  You could remove these parameters and setting the response could probably fit all on one line.

extensions/wikia/SpecialDiscussions/templates/SpecialDiscussions_index.mustache
```diff
@@ -0,0 +1,8 @@
+<div class="discussions-container">
+    {{#discussionsLink}}
```
garthwebb: This is mustache's version of `if`/`else` and in the index method above you also have an `if`/`else` in PHP for the same condition.  I would suggest just keeping one of them and I would probably keep the one in PHP.  You could replace all of this with a simple `{{{content}}}` include

extensions/wikia/SpecialDiscussions/SpecialDiscussionsController.class.php
```diff
@@ -0,0 +1,165 @@
...
+		$editToken = $this->getVal( self::EDIT_TOKEN );
+		if ( !empty( $editToken ) ) {
+			$this->activateDiscussions();
+		}
+
+		$this->setIndexOutput();
+	}
+
+	public function inputForm() {
+		$this->checkAccess();
```
garthwebb: The `checkAccess` method isn't meant to keep people from doing something they shouldn't be doing (that's what `assertValidPostRequest` is for) but to remind non-staff that there is nothing here for them.  For the that reason I don't think it's necessary to call it here and in the `discussionsLink` method.  Its already been called in `index` and the only way around that would be to URL hack to hit `inputForm` directly.

Alternately, the call to `checkAccess` could be removed from all these controller methods (including `index`) and added to `__construct`.  While it would still get called twice in this case because of the selfRequest stuff, it would mean it would never need to be added (or forgotten) to any new controller methods that are expected to be called by the client.

extensions/wikia/SpecialDiscussions/SpecialDiscussionsController.class.php
```diff
@@ -0,0 +1,165 @@
...
+					[
+						'activateDiscussions' => $this->getActivateDiscussionsMessage(),
+					]
+				)
+			);
+		}
+	}
+
+	private function activateDiscussions() {
+		$this->assertValidPostRequest( $this->request, $this->getUser() );
```
garthwebb: Both parameters to this instance methods are available via `$this`.  The `assertValidPostRequest` method has equal access to these variables/methods and both can be mocked in tests so I don't think they need to be arguments.

extensions/wikia/SpecialDiscussions/SpecialDiscussionsController.class.php
```diff
@@ -0,0 +1,165 @@
...
+				[
+					'siteId' => $this->siteId,
+					'error' => $e->getMessage(),
+				]
+			);
+			throw new ErrorPageError( 'unknown-error', 'discussions-activate-error' );
+		}
+	}
+
+	private function getIsDiscussionsActive() {
```
garthwebb: As a conditional test, the `get` could be dropped for `isDiscussionsActive`.

extensions/wikia/SpecialDiscussions/SpecialDiscussionsController.class.php
```diff
@@ -0,0 +1,165 @@
...
+		/** @var ApiProvider $apiProvider */
+		$apiProvider = Injector::getInjector()->get( ApiProvider::class );
+		/** @var SitesApi $api */
+		$api = $apiProvider->getApi( self::SERVICE_NAME, SitesApi::class );
+		$api->getApiClient()->getConfig()->setCurlTimeout( self::TIMEOUT );
+
+		return $api;
+	}
+
+	private function checkAccess() {
```
garthwebb: Since this operates in the same way as `assertValidPostRequest`, maybe this should be `assertCanAccess`?

extensions/wikia/SpecialDiscussions/templates/SpecialDiscussions_inputForm.mustache
```diff
@@ -0,0 +1,5 @@
+<p>{{discussionsInactiveMessage}}</p>
+<form method="post" class="form-discussions">
+    <input type="hidden" name="editToken" value="{{editToken}}">
+	<input type="submit" value="{{activateDiscussions}}"/>
```
garthwebb: Minor; indented too far.  Probably was done by PHPStorm because the `input` on the line above isn't terminated with a `/>`

extensions/wikia/SpecialDiscussions/SpecialDiscussionsController.class.php
armonr: Sure! I did not know it was available! 😆

extensions/wikia/SpecialDiscussions/SpecialDiscussionsController.class.php
armonr: That is redundant, indeed!

extensions/wikia/SpecialDiscussions/SpecialDiscussionsController.class.php
armonr: Only difference is, this method, as is, is stateless, so it can be reused for different requests, but probably YAGNI. I'll remove the params.

extensions/wikia/SpecialDiscussions/SpecialDiscussionsController.class.php
armonr: The intent was checking for authorization. Going with the latter suggestion since it is more bullet-proof!

extensions/wikia/SpecialDiscussions/SpecialDiscussionsController.class.php
garthwebb: Yup, any global variable starting with `$wg` is accessible via `F::app()->wg`

extensions/wikia/SpecialDiscussions/SpecialDiscussionsController.class.php
garthwebb: I think this is more valuable as an instance/stateful method that knows its a SpecialDiscusionController request.  Otherwise, you'd use or extend `WebRequest::isValidWriteRequest`.

PR #10810, opened 2016-07-11
PLATFORM-2313: Force 2FA for members of selected groups

wladekb:
https://wikia-inc.atlassian.net/browse/PLATFORM-2313

Depends on: 
- https://github.com/Wikia/config/pull/1857

/cc @jcellary @macbre @Grunny

### DIFF
--- a/lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
+++ b/lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
@@ -73,12 +73,12 @@ class PermissionsServiceImpl implements PermissionsService {
 			$implicitGroups = $this->implicitUserGroups[ $user->getId() ];
 		} else {
 			$implicitGroups = array( '*' );
-			if ( $user->getId() ) {
+			if ( $user->isLoggedIn() ) {
 				$implicitGroups[] = 'user';
 
 				$implicitGroups = array_unique( array_merge(
 					$implicitGroups,
-					$this->getAutomaticAccessControlGroups($user),
+					$this->getAutomaticAccessControlGroups( $user ),
 					\Autopromote::getAutopromoteGroups( $user )
 				) );
 				$this->implicitUserGroups[ $user->getId() ] = $implicitGroups;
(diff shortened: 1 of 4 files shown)


lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
```diff
@@ -88,6 +89,27 @@ public function getAutomaticGroups( \User $user, $recacheAutomaticGroups = false
 	}
 
 	/**
+	 * Returns list of access control groups that need to be added based on the explicit user groups
+	 *
+	 * @param \User $user
+	 * @return array
+	 */
+	private function getAutomaticAccessControlGroups( \User $user ) {
+		if ($user->getId()) {
```
macbre: `s/getId/isLoggedIn` ?

wladekb: This is exactly how it is checked a few lines above and I tried to be consistent. Should I change it in both places?

macbre: Yes, `isLoggedIn` is way easier to read than `getId`.

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
```diff
@@ -88,6 +89,27 @@ public function getAutomaticGroups( \User $user, $recacheAutomaticGroups = false
 	}
 
 	/**
+	 * Returns list of access control groups that need to be added based on the explicit user groups
```
jcellary: I believe it would be helpful to say not what the code is doing but why is it doing it. I.e. we need to make sure no highly privileged accounts are able to remove restricted-login group which controls 2FA.

lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
```diff
@@ -88,6 +89,27 @@ public function getAutomaticGroups( \User $user, $recacheAutomaticGroups = false
...
+	 *
+	 * @param \User $user
+	 * @return array
+	 */
+	private function getAutomaticAccessControlGroups( \User $user ) {
+		if ($user->getId()) {
+			$explicitGroups = $this->getExplicitGlobalGroups($user);
```
jcellary: You use different formatting i.e. ($user) instead of ( $user ).

Is this something that we care about?

jcellary:
One general remark is that restricted-login group is now a hybrid group that is partially an implicit group and partially not. PermissionsConfiguration enumerates implicit groups and explicit ones and restricted-login falls into the explicit bucket there. Explicit groups can be manually added to a user, hence the enumerations. However the function getAutomaticGroups now also adds 'restricted-login' alongside other implicit groups. That can be a bit confusing as it breaks the clear separation between implicit and explicit. If not for restrictions in RemovableGroups you could go to the UI, remove that group explicitly and then you'd still have it due to implicit additions, which would be very confusing for the user, so as a pattern should be avoided.

Possible solutions that come to my mind: create a separate implicit group with same permissions, convert restricted-login to a fully implicit group, leave as is and mark this logic for deletion after staging is ready.

wladekb:
@jcellary We have now an explicit group "restricted-login" and an implicit group "restricted-login-auto" that is added automatically. Does it solve your concern?

jcellary:
It does :)

PR #10814, opened 2016-07-13
RECIRC-159: Include FANDOM_TOPIC items in IMPACT_FOOTER scroller

pauloslund:
This change also starts the process of moving placement experiments into their own files in hopes of cleaning up the `placement.js` and making new groups easier to add

### DIFF
--- a/extensions/wikia/Recirculation/RecirculationHooks.class.php
+++ b/extensions/wikia/Recirculation/RecirculationHooks.class.php
@@ -46,7 +46,7 @@ class RecirculationHooks {
 	 * @return bool
 	 */
 	public static function onOasisSkinAssetGroups( &$jsAssets ) {
-		global $wgWikiaEnvironment, $wgNoExternals, $wgRecirculationTestGroup;
+		global $wgWikiaEnvironment, $wgNoExternals, $wgRecirculationTestGroup, $wgCityId;
 
 		// We only want to track this on production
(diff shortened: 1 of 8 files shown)


extensions/wikia/Recirculation/RecirculationHooks.class.php
```diff
@@ -56,7 +56,7 @@ public static function onOasisSkinAssetGroups( &$jsAssets ) {
 		if ( self::isCorrectPageType() ) {
 			$jsAssets[] = 'recirculation_js';
 
-			if ( self::canShowDiscussions() ) {
+			if ( self::canShowDiscussions( $wgCityId ) ) {
```
nmonterroso, line 59: seems like this check is being done many times - is there a way to consolidate this or are the use cases all different?

nmonterroso:
small comment otherwise lgtm

extensions/wikia/Recirculation/RecirculationHooks.class.php
pauloslund, line 59: The use cases are a little bit different as they are for different experiments, however the check could probably be removed from the `RecirculationController.php` file as the route that uses it should never get called without including the `recirculation_discussions_js` file referenced here. The Recirculation controller is getting an update in the near future and will probably be deprecated in favor of just using the RecirculationAPIController so I'll take a look at consolidating it then.

nmonterroso, line 59: cool, LGTM then 👍

PR #10816, opened 2016-07-13
SUS-673 in ChatWidget fix missing message about page type

gitkuba:
[SUS-673](https://wikia-inc.atlassian.net/browse/SUS-673)
Missing message when hovering on user avatar in chat widget was caused by missing variable 'profileType'. I've added profileType to getUserInfo method together with new variable profileTypeMsg to avoid concatenation on template. Both variables are wrapped in a conditional to determine wether Talk Page or Message Wall should be displayed.

@mixth-sense @sqreek


sqreek:
LGTM

ludwikkazmierczak:
Tested And Accepted

PR #10822, opened 2016-07-13
SOC-2737 - Improve DiscussionsLog

katanka:
Improvements to the discussions log based on feedback received during demo.
1. Use the domain instead of dbname when displaying site information.
2. Finish updating i18n strings.
3. Only display contributions in the same environment (only prod contributions on prod, only dev on dev).

https://wikia-inc.atlassian.net/browse/SOC-2737


extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -101,6 +101,13 @@ private function getUserById( $userId ) {
 			throw new \InvalidArgumentException( self::NO_USER_MATCH_ERROR );
 		}
 
+		$user->load();
+
+		// Anonymous users can't have contributions
+		if ($user->mName === false) {
```
armonr: Styling convention including spacing in MW/php app follows the [MediaWiki guidelines](https://www.mediawiki.org/wiki/Manual:Coding_conventions/PHP).

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -250,7 +257,7 @@ private function getSearchResults( $url, $query ) {
 		return $response;
 	}
 
-	private function appendRecordsFromResponse( array &$records, $response, array &$ipHash ) {
+	private function appendRecordsFromResponse( array &$records, $response, array &$domainCache ) {
```
armonr: Why not have domainCache be member of the controller class so no need to pass it around?

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -101,6 +101,13 @@ private function getUserById( $userId ) {
 			throw new \InvalidArgumentException( self::NO_USER_MATCH_ERROR );
 		}
 
+		$user->load();
```
armonr, line 106: What does this do?

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -207,7 +214,7 @@ private function aggregateLogByIpAddress( $ipAddress ) {
 	private function aggregateSearchLogs( $query ) {
 		$records = [];
 		$dayOffset = 0;
-		$ipHash = [];
```
armonr, line 210: Identical ip/domain tuples should still be eliminated, right?
Also come to think of it, should this become an ip/domain/userId tuple or else will it eliminate users to the same site within the same ip?

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -271,17 +278,30 @@ private function appendRecordsFromResponse( array &$records, $response, array &$
...
-			$wikiInfo = WikiFactory::getWikiByID($site);
+			if (!isset($domainCache[$site])) {
+				$cityTitle = WikiFactory::getWikiByID($site)->city_title;
+				$wikiDomain = preg_replace('#^https?://#', '', WikiFactory::getHostById($site));
+				if (empty($cityTitle) && empty($wikiDomain)) {
+					// most likely from a different environment (e.g. prod contribution when this is on dev, or vice versa)
+					$this->logger->warning(sprintf( 'Site not found: %d', $site ));
+					continue;
+				} else {
+					$domainCache[$site] = $cityTitle . ' (' . $wikiDomain . ')';
```
armonr: Consider using `sprintf` here too.

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
katanka, line 106: It loads the information into `$user` based on the user id.

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -271,17 +287,29 @@ private function appendRecordsFromResponse( array &$records, $response, array &$
 				continue;
 			}
 
-			$wikiInfo = WikiFactory::getWikiByID($site);
+			if ( !isset($domainCache[$site]) ) {
```
jsutterfield: Minor: There's a number of white space violations in this block. Function calls should have padding in them. Eg

``` php
// bad 
if ( !isset($domainCache[$site]) )

// good
if ( !isset( $domainCache[$site] ) )
```

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
```diff
@@ -101,6 +103,13 @@ private function getUserById( $userId ) {
 			throw new \InvalidArgumentException( self::NO_USER_MATCH_ERROR );
 		}
 
+		$user->load();
+
+		// Anonymous users can't have contributions
+		if ( $user->mName === false ) {
```
jsutterfield, line 109: How about using the `$user->isAnon();` method for this check?

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
armonr: And to fix all these automatically, you can run [`stylize.php`](https://github.com/Wikia/guidelines/blob/master/PHP/CodingConventions.md#mediawiki-php-style-helper) on the file. It's available on your devbox.

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
armonr, line 106: What information is lacking after calling `User::newFromId`?

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
katanka, line 109: That won't work, because `isAnon()` calls `isLoggedIn()` which return `$this->getId() != 0;`. Since we're providing the user id, `isAnon()` will always return `false`. Perhaps I should update the comment - I guess it's not so much that the user is anonymous, but that MW is unable to get information for that user id.

extensions/wikia/SpecialDiscussionsLog/SpecialDiscussionsLogController.class.php
katanka, line 106: `newFromId()` does not load any user information into the class besides the user id.

katanka, line 106: It seems counterintuitive, but the user class loads user info when a property is accessed, rather than when it is created. For example take the implementation of `getName()`, which this code was drawn from. https://github.com/Wikia/app/blob/dev/includes/User.php#L1817

PR #10831, opened 2016-07-15
WW-11: Run method after required modules are loaded

lukaszkonieczny:
https://wikia-inc.atlassian.net/browse/WW-11

This error is really hard to reproduce and it happens rarely. After small investigation and observation I assumed that `importWikiaScriptPages` is not defined because it's used before depended modules are loaded. So lets wait for them ;)


lukaszkonieczny:
ping @Wikia/west-wing

sebastianmarzjan:
👍

PR #10833, opened 2016-07-15
WW-57: Set cookie about flag that modal was shown, when modal is shown

lukaszkonieczny:
If user is on 4th page view, but click link before modal is shown, we should show modal on next page view.

@Wikia/west-wing


extensions/wikia/CommunityPage/scripts/CommunityPageBenefitsPageviews.js
```diff
@@ -1,28 +1,17 @@
...
-	// This cookie is checked in CommunityPageSpecialHooks::onBeforePageDisplay to avoid unnecessary script loading
-	function setModalShownCookie() {
-		cookies.set('cpBenefitsModalShown', 1, {
-			domain: mw.config.get('wgCookieDomain'),
-			expires: 2592000000, // 30 days
-			path: mw.config.get('wgCookiePath')
-		});
-	}
-
 	function trackModalImpression() {
```
sebastianmarzjan, line 27: Does this method need to be invoked here? I think that it belongs to CommunityPageBenefitsModal

sebastianmarzjan:
LGTM

PR #10834, opened 2016-07-15
JPN-589: log the reason why we're in the ready-only mode

macbre:
[JPN-589](https://wikia-inc.atlassian.net/browse/JPN-589) - make debugging it a bit easier

@d34th4ck3r / @wladekb


wladekb:
lgtm

PR #10837, opened 2016-07-15
(XW-1601) Support any English community in DesignSystem Footer API

Grunny:
Support providing a different response per wiki for English communities.
This only involves changing the License and Vertical data per community,
so the remaining response is still static for now.

/cc @Wikia/x-wing


includes/wikia/models/DesignSystemFooterModel.class.php
```diff
@@ -0,0 +1,423 @@
...
+	private function getVerticalData() {
+		$wikiVerticalData = WikiFactoryHub::getInstance()->getWikiVertical( $this->wikiId );
+		$vertical = null;
+		if ( $wikiVerticalData['id'] ) {
+			$vertical = $wikiVerticalData['short'];
+			$link = 'http://fandom.wikia.com/' . $vertical;
+		}
+
+		if ( !in_array( $vertical, $this->allowedVerticals ) ) {
+			return [];
```
rogatty: Shouldn't this be

```
return [
    'type' => 'line-text',
    'title' => [
        'key' => 'fandom-link-vertical-' . $vertical
    ]
];
```

?

Grunny: I went with not returning anything right now until we know how we want to handle them. But, yeah, we can do that if we just want to return the text without a link.

includes/wikia/models/DesignSystemFooterModel.class.php
```diff
@@ -382,18 +382,23 @@ private function getVerticalData() {
...
 		}
 
-		if ( !in_array( $vertical, $this->allowedVerticals ) ) {
+		if ( is_null( $vertical ) ) {
 			return [];
 		}
 
-		return [
+		$data = [
 			'type' => 'link-text',
```
rogatty: This won't pass the validation (we should fix the schema if it does). When there is no link the type should be `line-text`.

includes/wikia/models/DesignSystemFooterModel.class.php
```diff
@@ -0,0 +1,428 @@
...
+			$vertical = $wikiVerticalData['short'];
+			$link = 'http://fandom.wikia.com/' . $vertical;
+		}
+
+		if ( is_null( $vertical ) ) {
+			return [];
+		}
+
+		$data = [
+			'type' => 'link-text',
```
rogatty: This won't pass the validation (we should fix the schema if it does). When there is no link the type should be `line-text`.

includes/wikia/models/DesignSystemFooterModel.class.php
```diff
@@ -0,0 +1,429 @@
+<?php
+
+class DesignSystemFooterModel extends WikiaModel {
```
rogatty: In all our RFCs we use words "Global Footer". Do you think it makes sense to keep naming consistent here?

includes/wikia/models/DesignSystemFooterModel.class.php
```diff
@@ -0,0 +1,429 @@
...
+						'type' => 'translatable-text',
+						'key' => 'wikia-advertise-link-contact'
+					],
+					'href' => 'http://www.wikia.com/mediakit#contacts'
+				]
+			]
+		],
+	];
+
+	private $allowedVerticals = [
```
rogatty: This name is misleading. How about `$verticalsWithLink` or something like that?

includes/wikia/models/DesignSystemFooterModel.class.php
```diff
@@ -0,0 +1,429 @@
...
+		return $this->data;
+	}
+
+	private function setLicensingAndVertical() {
+		$params = [
+			'license' => $this->getLicenseData(),
+		];
+		$verticalData = $this->getVerticalData();
+
+		if ( !empty( $verticalData ) ) {
```
rogatty: I wonder if we should `setLicensingAndVertical` if there is no vertical data. The final result will be broken, right?

includes/wikia/models/DesignSystemFooterModel.class.php
```diff
@@ -0,0 +1,429 @@
...
+			],
+		];
+	}
+
+	private function getVerticalData() {
+		$wikiVerticalData = WikiFactoryHub::getInstance()->getWikiVertical( $this->wikiId );
+		$vertical = null;
+		if ( $wikiVerticalData['id'] ) {
+			$vertical = $wikiVerticalData['short'];
+			$link = 'http://fandom.wikia.com/' . $vertical;
```
rogatty: I see no need for this variable, actually the logic would be easier to reason about without it.

includes/wikia/models/DesignSystemFooterModel.class.php
Grunny: Sure, we can do that. I went with just Footer to keep it consistent with the API method it's providing data for (`getFooter`).

includes/wikia/models/DesignSystemFooterModel.class.php
```diff
@@ -0,0 +1,429 @@
...
+		}
+
+		if ( is_null( $vertical ) ) {
```
rogatty: When can it be a null? Is there a lot of wikis without vertical? Would it make sense to set some default?

includes/wikia/models/DesignSystemFooterModel.class.php
Grunny: Yeah, left over from when the idea was to not return a vertical at all in this case.

includes/wikia/models/DesignSystemFooterModel.class.php
Grunny: There are ~30,000 wikis without a vertical, and we currently simply don't display a vertical in the footer on them: http://freebsd.wikia.com/wiki/Main_Page We can make a default, but we'd have to decide what we want the default to be.

includes/wikia/models/DesignSystemFooterModel.class.php
Grunny: Ideally we should still be returning the license, rather than nothing on the ~30,000 wikis this affects. I'd prefer the client to handle no vertical, but we can try to come up with a default which would also solve this.

rogatty:
Will LGTM when we do something about the wikis without verticals.

includes/wikia/models/DesignSystemGlobalFooterModel.class.php
```diff
@@ -0,0 +1,428 @@
...
+			'title' => [
+				'type' => 'text',
+				'value' => $licenseText
+			],
+			'href' => $this->getLicenseUrl( $this->wikiId )
+		];
+	}
+
+	private function getLicenseUrl() {
+		$licenseUrl = WikiFactory::getVarByName( 'wgRightsUrl', $this->wikiId )->cv_value ?: $this->wg->RightsUrl;
```
hakubo, line 420: shouldn't that be the other way round?
read it from memory and if not there possibly go to DB?

hakubo:
LGTM, one minor comment

includes/wikia/models/DesignSystemGlobalFooterModel.class.php
Grunny, line 420: This API takes the wiki ID being requested as a parameter, so it's not necessarily called on the wiki DB being requested. So, we want to first check the value on the wiki in question (configured in WikiFactory), then fallback to the default.

hakubo, line 420: got it! 😶

rogatty:
LGTM

includes/wikia/models/DesignSystemGlobalFooterModel.class.php
```diff
@@ -0,0 +1,430 @@
...
+class DesignSystemGlobalFooterModel extends WikiaModel {
+	private $data = [
+		'fandom' => [
+			'header' => [
+				'type' => 'link-image',
+				'image' => 'company/logo-fandom',
+				'href' => 'http://fandom.wikia.com',
+				'title' => [
+					'type' => 'translatable-text',
+					'key' => 'fandom-header'
```
rogatty: Shouldn't this be `global-footer-fandom-header`? Same thing for all message keys.

PR #10838, opened 2016-07-15
SUS-505: Fix HTML entity broken by Brazilian Portuguese l10n

mszabo-wikia:
https://wikia-inc.atlassian.net/browse/SUS-505

@Rappy @mixth-sense


Grunny:
LGTM.

PR #10843, opened 2016-07-18
Post PHP7 migration cleanup

macbre:
[PLATFORM-2221](https://wikia-inc.atlassian.net/browse/PLATFORM-2221)

Remove various checks `if PHP 5.6.x then do this, if PHP 7 - do that`.

**Do not merge before production is fully switched to PHP7 (should happen on July 20th)**

@wladekb / @artursitarski


includes/profiler/ProfilerTideways.php
```diff
@@ -4,6 +4,7 @@
  * data collection to Tideway's php-profiler-extension. Afterwards sends the gathered data
  * to UDP formatted in similar way to regular ProfilerSimpleUDP.
  *
+ * @author wladek
```
wladekb, line 7: ;-)

wladekb:
lgtm

includes/profiler/ProfilerTideways.php
macbre, line 7: Give credit where credit is due :)

PR #10850, opened 2016-07-18
WW-64: use timestamp for benefits modal cookie, instead of "1"

kerriamber:
https://wikia-inc.atlassian.net/browse/WW-64

### DIFF
--- a/extensions/wikia/CommunityPage/scripts/CommunityPageBenefitsPageviews.js
+++ b/extensions/wikia/CommunityPage/scripts/CommunityPageBenefitsPageviews.js
@@ -14,9 +14,9 @@ require([
 		}
 	}
 
-	// Gets timestamp in format YYYY-mm-dd HH:mm:ss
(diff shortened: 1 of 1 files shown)


extensions/wikia/CommunityPage/scripts/CommunityPageBenefitsPageviews.js
```diff
@@ -14,9 +14,14 @@ require([
 		}
 	}
 
+	// Gets timestamp in format YYYY-mm-dd HH:mm:ss
+	function getTimestamp() {
+		return (new Date()).toISOString().substr(0,19).replace('T', ' ');
```
dianafa: what about `date("Y-m-d H:i:s")` ?

kerriamber: Oops, I meant to put `yyyy-MM-dd HH:mm:ss` (what you said in slack) into comment. I think this output is identical to php function though

dianafa:
LGTM 🙈

PR #10853, opened 2016-07-19
JPN-598 Smaller font sizes for Japanese in NCF modal

perjg:
Benefit Modal font size overrides for Japanese.

@d34th4ck3r @Wikia/spitfires


Grunny:
LGTM.

PR #10855, opened 2016-07-19
PLATFORM-2325: only allow internal requests to use WikiaStyleGuide controllers

macbre:
[PLATFORM-2325](https://wikia-inc.atlassian.net/browse/PLATFORM-2325)

@Grunny


Grunny:
LGTM.

PR #10856, opened 2016-07-19
SUS-541: Simplify main method in ImageReview controller

lukaszkonieczny:
Extract code from ImageReviewSpecialController::index to smaller functions

@harnash


extensions/wikia/ImageReview/ImageReviewSpecialController.class.php
```diff
@@ -280,4 +199,112 @@ protected function parseImageData($data) {
...
+
+			$order = $this->app->wg->User->getGlobalPreference('imageReviewSort');
+			return $order;
+		} else {
+			$order = -1;
+			return $order;
+		}
+	}
+
+	private function getAction() {
```
harnash: This name is a bit misleading. I would expect it to be some kind of accessor but it is more a parser. Fortunately it does not have side effects so it is not that bad. I can see that it follows convention in this class so it is up to you to change its name to something closer to what this function does.... `parseAction`?

harnash:
Two minor comments. Otherwise LGTM

extensions/wikia/ImageReview/ImageReviewSpecialController.class.php
```diff
@@ -28,100 +34,39 @@ protected function setGlobalDisplayVars() {
 	public function index() {
 		$this->setGlobalDisplayVars();
 
-		$action = $this->getPar();
-		$actions = explode('/',$action);
-		$action = array_pop($actions);
+		$this->action = $this->getAction();
+		$this->order = $this->getOrderingMethod();
+		$this->ts = $this->wg->request->getVal( 'ts' );
```
mszabo-wikia: it could be possible to use the local RequestContext via `->getContext()` instead of the Nirvana accessors

PR #10858, opened 2016-07-19
XW-1601 | fix icon names/links in Design System Global Footer Model

Warkot:
After build step of Design System SVGs have their path converted to names (`/` => `-`) and are used using `use:`.

I also fixed the names themselves—there are no `-default` suffixes in DS.

/cc @Wikia/x-wing


rogatty:
LGTM

kamilkoterba:
LGTM

PR #10861, opened 2016-07-19
SUS-541: Image stats are now cached in separate keys to avoid race conditions

harnash:
To allow global cache for image review stats we should avoid race conditions
and do not store state per user (except reviewer stats). To achieve this since
cas() function is not available for us we split the stats into sepearate keys
and use incr() decr() to change the counters. This should eliminate the race
conditions and perhaps allow values to be more up to date.

ping: @lukaszkonieczny

### DIFF
--- a/extensions/wikia/ImageReview/ImageReviewHelper.class.php
+++ b/extensions/wikia/ImageReview/ImageReviewHelper.class.php
@@ -11,9 +11,34 @@ class ImageReviewHelper extends ImageReviewHelperBase {
 
 	private $user_id = 0;
 
+	private $image_stats_cache_keys = [];
+
+	const STATS_REJECTED = 'rejected';
+
+	const STATS_UNREVIEWED = 'unreviewed';
+
+	const STATS_QUESTIONABLE = 'questionable';
+
+	const STATS_INVALID = 'invalid';
+
+	const STATS_REVIEWER = 'reviewer';
+
 	function __construct() {
 		parent::__construct();
 		$this->user_id = $this->wg->user->getId();
+
+		$stats_keys = [
+			self::STATS_REJECTED, self::STATS_UNREVIEWED, self::STATS_QUESTIONABLE,
+			self::STATS_INVALID, self::STATS_REVIEWER
+		];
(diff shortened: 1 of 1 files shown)


extensions/wikia/ImageReview/ImageReviewHelper.class.php
```diff
@@ -376,11 +393,39 @@ private function imageListAdditionalAction( $sType, DatabaseBase $oDB, Array $aV
 		] );
 	}
 
+	public function getImageStats() {
+		$stats = [];
+		foreach( $this->image_stats_mapping as $key ) {
+			$value = $this->wg->memc->get( $this->image_stats_mapping[$key] );
```
lukaszkonieczny: Did you mean `$value = $this->wg->memc->get( $key );` ?

extensions/wikia/ImageReview/ImageReviewHelper.class.php
```diff
@@ -11,9 +11,34 @@ class ImageReviewHelper extends ImageReviewHelperBase {
 
 	private $user_id = 0;
 
+	private $image_stats_mapping = [];
```
lukaszkonieczny: I'm not sure about naming because actually it stores cache keys.

extensions/wikia/ImageReview/ImageReviewHelper.class.php
```diff
@@ -376,11 +393,39 @@ private function imageListAdditionalAction( $sType, DatabaseBase $oDB, Array $aV
...
+			}
+		}
+	}
+
+	public function updateImageStats ($key, $relative_change ) {
+		if ( array_key_exists( $key, $this->image_stats_mapping ) ) {
+			if ( $relative_change > 0 ) {
+				$this->wg->memc->incr( $this->image_stats_mapping[$key], $relative_change );
+			} else {
+				$this->wg->memc->decr( $this->image_stats_mapping[$key], $relative_change );
```
lukaszkonieczny: Do we want decrement negative value?

harnash: Good catch! Fixed.

PR #10869, opened 2016-07-20
Platform 1711

pgrabowski-wikia:
If category is missing inside the URL, it now returns information about it


macbre:
As discussed offline, I'd create separate branches for each ticket to make code review a bit easier.

PR #10873, opened 2016-07-20
RECIRC-164: Map discussions cityId to alias WF variable if set

pauloslund:
On some communities that don't have discussions enabled, we want to be able to configure other communities discussions to be shown in the Recirculation modules


gabrys:
LGTM

PR #10874, opened 2016-07-20
(SEC-69) Make use of PHP 7's filtered unserialization

Grunny:
Make use of PHP 7's filtered unserialization instead of the safe unserialize
library. The allowed_classes option allows us to have the same functionality
as the safe unserialize library to prevent unserializing arbitrary class objects
with a more performant native method (which doesn't limit the length of the
string being unserialized).

/cc @macbre


macbre:
LGTM

PR #10875, opened 2016-07-20
SOC-1408 Stop writing to, and remove existing migrated attributes from, sharedDB

jsutterfield:
Ticket: https://wikia-inc.atlassian.net/browse/SOC-1408

This PR does 2 things.

First it stops writing attributes to sharedDB which have been migrated to the attribute service. When we were making the transition to using the attribute service, we wanted to write attributes in both places in case we encountered any issues and needed to roll back to using sharedDB again. Now that the attribute service has been used in production for the last 9 months without issue, we can confidently stop writing that data to sharedDB.

Second, it removes those attributes from sharedDB so that they are full migrated.


includes/User.php
```diff
@@ -2648,22 +2648,7 @@ public function getGlobalAttribute( $attribute, $default = null ) {
 	 * @return bool
 	 */
 	private function shouldGetAttributeFromService( $attributeName ) {
-		global $wgEnableReadsFromAttributeService;
-
-		if ( empty( $wgEnableReadsFromAttributeService ) ) {
```
jsutterfield, line 2653: `$wgEnableReadsFromAttributeService` is universally set to true, so it's is no longer helpful. I removed this check here, and removed that variable from CommonSettings.php.

includes/User.php
```diff
@@ -2648,22 +2648,7 @@ public function getGlobalAttribute( $attribute, $default = null ) {
...
-			return false;
-		}
-
-		// User is anonymous or nonexistant
-		if ( $this->getId() == 0 ) {
-			return false;
```
jsutterfield, line 2659: We now do this check inside of `UserAttributes.php` [here](https://github.com/Wikia/app/pull/10875/files#diff-3fe2c8a946b8da62e3fdb9ba3f5ff012R71).

lib/Wikia/src/Service/User/Attributes/UserAttributes.php
```diff
@@ -68,6 +68,10 @@ public function getAttribute( $userId, $attributeName, $default = null ) {
 
 	private function loadAttributes( $userId ) {
 
+		if ( $userId == 0 ) {
+			return [];
```
jsutterfield, line 74: By returning an empty array here for anonymous users, we prevent an unnecessary call to the attribute service. Those anon users will just use default values for attributes. See the logic [here](https://github.com/Wikia/app/blob/dev/lib/Wikia/src/Service/User/Attributes/UserAttributes.php#L56-L66).

includes/User.php
```diff
@@ -2648,22 +2648,7 @@ public function getGlobalAttribute( $attribute, $default = null ) {
 	 * @return bool
 	 */
 	private function shouldGetAttributeFromService( $attributeName ) {
```
armonr, line 2650: Maybe this function can be tossed altogether in favor of directly calling `$this->isPublicAttribute( $attributeName );`?

includes/User.php
```diff
@@ -2683,11 +2668,10 @@ public function setGlobalAttribute($attribute, $value) {
 		if ( $this->isPublicAttribute( $attribute ) ) {
 			$value = $this->replaceNewlineAndCRWithSpace( $value );
 			$this->userAttributes()->setAttribute( $this->getId(), new Attribute( $attribute, $value ) );
+		} else {
+			$this->setOptionHelper( $attribute, $value );
```
armonr, line 2672: Non-public attributes then continue to be written in `user_properties` table?

armonr:
Minor comments/questions. LGTM!

lib/Wikia/src/Service/User/Attributes/UserAttributes.php
slayful, line 74: Perhaps we should document this in the code?

lib/Wikia/src/Service/User/Attributes/maintenance/cleanupMigratedAttributs.php
```diff
@@ -0,0 +1,62 @@
...
+ * Assumptions: The global variable $wgPublicUserAttributes is an array containing the names of
+ * all fully migrated attributes. Do NOT add any attributes to that array unless you have fully
+ * migrated all of its values from wikicities.user_properties to the attribute service.
+ */
+
+require_once( __DIR__ . '/../../../../../../../maintenance/Maintenance.php' );
+
+class CleanupMigratedAttributes extends Maintenance {
+
+    private $force;
```
slayful: Perhaps we could initialise it to `false`

```
private $force = false;
```

slayful: And maybe reversing the meaning to `$dryRun` might be better?

lib/Wikia/src/Service/User/Attributes/maintenance/cleanupMigratedAttributs.php
```diff
@@ -0,0 +1,62 @@
...
+
+    private function printAttrsToBeDeleted() {
+        global $wgPublicUserAttributes;
+        $this->output( "Preparing to delete following attributes from wikicities.user_properties:\n\n");
+        $this->output( implode( $wgPublicUserAttributes, "\n" ) );
+        $this->output( "\n\n" );
+        if ( $this->force ) {
+            $this->output( "NOT DRY RUN, ATTRIBUTES WILL BE DELETED\n" );
+        }
+        sleep(5);
```
slayful: That's neat =)

lib/Wikia/src/Service/User/Attributes/maintenance/cleanupMigratedAttributs.php
```diff
@@ -0,0 +1,62 @@
...
+        $this->printSuccess();
+    }
+
+    private function printAttrsToBeDeleted() {
+        global $wgPublicUserAttributes;
+        $this->output( "Preparing to delete following attributes from wikicities.user_properties:\n\n");
+        $this->output( implode( $wgPublicUserAttributes, "\n" ) );
+        $this->output( "\n\n" );
+        if ( $this->force ) {
+            $this->output( "NOT DRY RUN, ATTRIBUTES WILL BE DELETED\n" );
```
slayful: I'd add info that deletion will begin in 5s so that the user can be more confident when interrupting the script.

lib/Wikia/src/Service/User/Attributes/maintenance/cleanupMigratedAttributs.php
jsutterfield: Thanks, I stole that from Macbre :D

includes/User.php
jsutterfield, line 2672: That's correct. Any attribute stored in that attribute service is available publicly, meaning any client can request them. The private attributes are values that we don't want to expose to the outside world, like emails. Eventually we'll update the attribute service to support these and only return them to specific clients, but for now we'll just continue writing them to the user_properties table and let MW handle them.

includes/User.php
jsutterfield, line 2650: Yea, I was thinking about that too. Personally I'd like to leave it as is since I think it's more explicit about what's going on. If we just use `$this->isPublicAttribute( $attributeName );` the reader has to infer that pubic attributes mean that we use the attribute service. Keeping the 2 calls I think tells a better story with less guessing.

armonr:
LGTM

PR #10878, opened 2016-07-21
JPN-595 Use wgCommunityPageDisableTopContributors wikifactory variabl…

perjg:
…e to turn top contributors module on/off

@Grunny @d34th4ck3r


includes/resourceloader/ResourceLoaderStartUpModule.php
```diff
@@ -98,6 +98,7 @@ protected function getConfig( $context ) {
 			'wgLegalTitleChars' => Title::convertByteClassToUnicodeClass( Title::legalChars() ),
 			// Wikia - change begin - @author: wladek
 			'wgSassParams' => SassUtil::getSassSettings(),
+			'wgCommunityPageDisableTopContributors' => $wgCommunityPageDisableTopContributors,
```
Grunny: Since this is extension-specific, it should be added to the RL config vars with a hook (i.e. `ResourceLoaderGetConfigVars`).

Grunny:
Once the JS config var is moved to a hook, will LGTM. :)

perjg:
@Grunny Moved the JS config var to a hook as suggested.

Grunny:
:+1:

PR #10879, opened 2016-07-21
SUS-541/SUS-759: Fixed counters on image review special page a.k.a D.I.R.T.

harnash:
There is an issue with D.I.R.T. that causes image counters in the header of the page to be displayed incorrectly. This fix changes the way they are being cached and clean up some code.

ping: @lukaszkonieczny @mixth-sense


extensions/wikia/ImageReview/ImageReviewHelper.class.php
```diff
@@ -11,9 +11,34 @@ class ImageReviewHelper extends ImageReviewHelperBase {
...
+
 	function __construct() {
 		parent::__construct();
 		$this->user_id = $this->wg->user->getId();
+
+		$stats_keys = [
+			self::STATS_REJECTED, self::STATS_UNREVIEWED, self::STATS_QUESTIONABLE,
+			self::STATS_INVALID, self::STATS_REVIEWER
+		];
+		foreach ( $stats_keys as $key ) {
```
sqreek: This loop has logic that for one element from list is different.
I suggest separate this loop for all keys except `STATS_REVIEWER` and add `STATS_REVIEWER` outside loop.

extensions/wikia/ImageReview/ImageReviewHelper.class.php
```diff
@@ -75,29 +100,23 @@ public function updateImageState( $images, $action = '' ) {
...
-				case ImageReviewSpecialController::ACTION_REJECTED:
-					$changedState = count( $sqlWhere[ImageReviewStatuses::STATE_APPROVED] ) + count( $sqlWhere[ImageReviewStatuses::STATE_DELETED] );
-					$stats['rejected'] -= $changedState;
-					break;
-			}
-			$this->wg->memc->set( $key, $stats, 3600 /* 1h */ );
-			// Quick hack/fix for stats going negative -- FIXME
-			if ($stats['unreviewed'] < 0)
-				$this->wg->memc->delete( $key );
+		switch ( $action ) {
```
sqreek, line 98: It will be nice if block like this put in separate method.
Also using sqlWhere is misleading a little.
also please use const in https://github.com/Wikia/app/pull/10879/files#diff-a56c48c5a45af2e5c50e11787482421fR62.

extensions/wikia/ImageReview/ImageReviewHelper.class.php
```diff
@@ -374,11 +393,39 @@ private function imageListAdditionalAction( $sType, DatabaseBase $oDB, Array $aV
 		] );
 	}
 
+	public function getImageStats() {
```
sqreek: This method return data only from cache. Is this name is obvious in file context?

extensions/wikia/ImageReview/ImageReviewHelper.class.php
```diff
@@ -374,11 +393,39 @@ private function imageListAdditionalAction( $sType, DatabaseBase $oDB, Array $aV
...
+			if ( !empty( $value ) ) {
+				$stats[$key] = $value;
+			}
+		}
+		return $stats;
+	}
+
+	public function setImageStats ( $new_stats ) {
+		foreach ( $this->image_stats_cache_keys as $key => $cache_key ) {
+			if ( array_key_exists( $key, $new_stats ) ) {
```
sqreek: What if $key will not be in $new_stats?

extensions/wikia/ImageReview/ImageReviewHelper.class.php
```diff
@@ -11,9 +11,34 @@ class ImageReviewHelper extends ImageReviewHelperBase {
 
 	private $user_id = 0;
 
+	private $image_stats_cache_keys = [];
```
sqreek: After read whole code:
Can we use separate class for mapping stats_keys into cache key names?

extensions/wikia/ImageReview/ImageReviewHelper.class.php
```diff
@@ -374,11 +393,39 @@ private function imageListAdditionalAction( $sType, DatabaseBase $oDB, Array $aV
...
+			}
+		}
+	}
+
+	public function updateImageStats ($key, $relative_change ) {
+		if ( array_key_exists( $key, $this->image_stats_cache_keys ) ) {
+			if ( $relative_change > 0 ) {
+				$this->wg->memc->incr( $this->image_stats_cache_keys[$key], $relative_change );
+			} else {
+				$this->wg->memc->decr( $this->image_stats_cache_keys[$key], -$relative_change );
```
sqreek: incr with + and decr with - :) so no difference in logic :).
Just use `$this->wg->memc->incr( $this->image_stats_cache_keys[$key], $relative_change );` without condition

extensions/wikia/ImageReview/ImageReviewHelper.class.php
```diff
@@ -387,37 +434,41 @@ public function getImageCount( $sAction = false, $iImageListCount = false ) {
...
 		$aCounts = $oDatabaseHelper->countImagesByState();
 
-		$total = array(
-			'unreviewed' => 0,
-			'questionable' => 0,
-			'rejected' => 0,
-			'invalid' => 0,
-		);
-
 		if ( array_key_exists( ImageReviewStatuses::STATE_UNREVIEWED, $aCounts ) ) {
```
sqreek: maybe we can rename aCounts

extensions/wikia/ImageReview/ImageReviewSpecialController.class.php
```diff
@@ -131,50 +76,28 @@ public function index() {
...
+		$this->setVariables();
 
-	protected function getOrderingMethod() {
-		if ($this->wg->user->isAllowed('imagereviewcontrols')) {
-			$order = (int)$this->getVal('sort', -1);
-			if ($order >= 0) {
-				$this->app->wg->User->setGlobalPreference('imageReviewSort', $order);
-				$this->app->wg->User->saveSettings();
-			}
+		$this->imageCount = array_map(function($number) {
```
sqreek, line 96: $this->imageCount is transformed from integer or floating to string and assign to same array.
It brings inconsistent where imageCount is array of numbers in first part of code and strings in second one.

extensions/wikia/ImageReview/ImageReviewSpecialController.class.php
```diff
@@ -280,4 +203,112 @@ protected function parseImageData($data) {
...
+			if ( count($images) > 0 ) {
+				$this->helper->updateImageState( $images, $this->action );
+			}
+			$this->ts = null;
+		}
+	}
+
+	private function getOrderingMethod() {
+		if ($this->wg->user->isAllowed('imagereviewcontrols')) {
+			$order = (int)$this->getVal('sort', -1);
```
sqreek: maybe read preference, check is it different and set if it's different.
In theory: Set and Read can bring race condition.

extensions/wikia/ImageReview/ImageReviewSpecialController.class.php
```diff
@@ -280,4 +203,112 @@ protected function parseImageData($data) {
...
+			return $order;
+		}
+	}
+
+	private function parseAction() {
+		$actions = explode( '/', $this->getPar() );
+		return array_pop( $actions );
+	}
+
+	private function checkUserPermissions() {
```
sqreek, line 243: this method doesn't return value when use has permission.
Please add return true in the end.

extensions/wikia/ImageReview/ImageReviewSpecialController.class.php
```diff
@@ -280,4 +203,112 @@ protected function parseImageData($data) {
...
+		} elseif ( $this->action == self::ACTION_QUESTIONABLE && !$this->accessQuestionable ) {
+			$this->specialPage->displayRestrictionError( 'questionableimagereview' );
+			return false;
+		} elseif ( $this->action == self::ACTION_REJECTED && !$this->accessRejected ) {
+			$this->specialPage->displayRestrictionError( 'rejectedimagereview' );
+			return false;
+		}
+	}
+
+	private function checkRedirect() {
```
sqreek, line 262: This method doesn't return value in all cases

extensions/wikia/ImageReview/ImageReviewHelper.class.php
harnash: https://github.com/memcached/memcached/wiki/Commands#incrdecr

You cannot incr/decr with negative values. They accept only positive.

extensions/wikia/ImageReview/ImageReviewSpecialController.class.php
harnash, line 96: @sqreek This transformation is needed for frontend to be able to display those values properly. However function operates on integers and stores them in cache as integers to allow offline update of those values easily. I could apply transformation on function output but it is more "backend" function so it should not deal with representation of this data - it is "frontend" task hence this code is in controller.

extensions/wikia/ImageReview/ImageReviewSpecialController.class.php
```diff
@@ -221,19 +221,18 @@ private function handlePostAction() {
 	}
 
 	private function getOrderingMethod() {
-		if ($this->wg->user->isAllowed('imagereviewcontrols')) {
-			$order = (int)$this->getVal('sort', -1);
-			if ($order >= 0) {
-				$this->app->wg->User->setGlobalPreference('imageReviewSort', $order);
+		if ($this->wg->user->isAllowed( 'imagereviewcontrols' )) {
+			$preferedOrder = (int)$this->app->wg->User->getGlobalPreference( 'imageReviewSort' );
```
sqreek, line 225: what happen if $preferedOrder will be null?

lukaszkonieczny, line 225: It will be converted to 0

extensions/wikia/ImageReview/ImageReviewSpecialController.class.php
ghost, line 243: This has been addressed.

ghost:
Hmm, and not even a single test for the new classes?

harnash:
@mixth-sense Unit Tests added in 1ce94f91f838c8fee6d7936ff883c35b4faadeb9

PR #10881, opened 2016-07-21
PLATFORM-2326: Fix domain attribute in Wikis/Details api

wladekb:
This makes the following change:
`"domain": "community.wikia.com",`

https://wikia-inc.atlassian.net/browse/PLATFORM-2326

/cc @macbre @aniaMu

### DIFF
--- a/includes/wikia/services/WikiService.class.php
+++ b/includes/wikia/services/WikiService.class.php
@@ -868,11 +868,11 @@ class WikiService extends WikiaModel {
 			);
 
 			while( $row = $db->fetchObject( $rows ) ) {
-				$parsedUrl = parse_url( $row->city_url );
+				$domain = parse_url( $row->city_url, PHP_URL_HOST );
 				$item = array(
 					'name' => $row->city_title,
 					'url' => $row->city_url,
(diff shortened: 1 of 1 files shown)


includes/wikia/services/WikiService.class.php
```diff
@@ -868,10 +868,11 @@ public function getDetails( Array $wikiIds = null ) {
 			);
 
 			while( $row = $db->fetchObject( $rows ) ) {
+				$parsedUrl = parse_url( $row->city_url );
```
macbre: You can pass `PHP_URL_HOST` as a second argument and get the string in return.

includes/wikia/services/WikiService.class.php
```diff
@@ -868,10 +868,11 @@ public function getDetails( Array $wikiIds = null ) {
...
 				$item = array(
 					'name' => $row->city_title,
 					'url' => $row->city_url,
-					'domain' => $row->city_url,
+					'domain' => !empty($parsedUrl['host']) ? $parsedUrl['host'] : null,
```
aniaMu: was the change domain -> url change in the getByString method?

wladekb: @aniaMu Could you point me out what `getByString` method are you referring to?

PR #10883, opened 2016-07-21
WW-37 | Insights popular pages showing only pages from content namespaces

rybmat:
Ticket: https://wikia-inc.atlassian.net/browse/WW-37

including only pages from content namespaces for Popular pages list on insights and Trending pages list on Community Page

@Wikia/west-wing

### DIFF
--- a/extensions/wikia/InsightsV2/specials/PopularPages.class.php
+++ b/extensions/wikia/InsightsV2/specials/PopularPages.class.php
@@ -73,12 +73,12 @@ class PopularPages extends PageQueryPage {
 	 * @return array
 	 */
 	public function reallyDoQuery( $limit = false, $offset = false ) {
-		global $wgCityId;
+		global $wgCityId, $wgContentNamespaces;
 
 		$topArticles = DataMartService::getTopArticlesByPageview(
 			$wgCityId,
 			null,
-			null,
+			$wgContentNamespaces,
 			false,
 			self::LIMIT
 		);


kerriamber:
LGTM :+1:

extensions/wikia/InsightsV2/specials/PopularPages.class.php
```diff
@@ -73,12 +73,12 @@ public function recache( $limit = false, $ignoreErrors = true ) {
 	 * @return array
 	 */
 	public function reallyDoQuery( $limit = false, $offset = false ) {
-		global $wgCityId;
+		global $wgCityId, $wgContentNamespaces;
 
 		$topArticles = DataMartService::getTopArticlesByPageview(
 			$wgCityId,
 			null,
-			null,
```
idradm: Does this contain main (0) namespace as well?

extensions/wikia/InsightsV2/specials/PopularPages.class.php
```diff
@@ -73,12 +73,12 @@ public function recache( $limit = false, $ignoreErrors = true ) {
...
+			$wgContentNamespaces,
```
idradm, line 81: Does this contain main (0) namespace as well?

kerriamber, line 81: @idradm It should. By default `$wgContentNamespaces` has value of `array( NS_MAIN );`

rybmat, line 81: by default yes, but if user deletes it then no. Do you think we should force it?

kerriamber, line 81: @rybmat In what case could a user delete NS_MAIN from $wgContentNamespaces? Only staff can edit this variable and I don't think comsup would ever allow a wiki to not have NS_MAIN as content NS

rybmat, line 81: by user I meant user who can do it ;) and actually there is one wiki which does not have 0 namespace in wgContentNamespaces: http://mum.mpb-test.wikia.com/wiki/AfW8

PR #10885, opened 2016-07-21
luastandalone: bump memory limit because of PLATFORM-2331

macbre:
`Lua error: not enough memory.` - 70% of all errors reported

@Grunny


Grunny:
:+1:

PR #10889, opened 2016-07-21
RECIRC-179: Extract fandomTopic into its own test and add LI tracking

pauloslund:
This also adds hard coded limits for the category type inside the controller


nmonterroso:
LGTM

PR #10892, opened 2016-07-22
luastandalone: bump memory limit because of PLATFORM-2331 - take two

macbre:
Bump to 250 MB - this is a followup of #10885

@Grunny


Grunny:
LGTM.

PR #10896, opened 2016-07-22
WW-80 escape whitelisted message

idradm:


sebastianmarzjan:
LGTM

Grunny:
This is already escaped when it's output: https://github.com/Wikia/app/blob/f25bdd76cfbd8037/extensions/wikia/CommunityPage/templates/pageHeader.mustache#L7 So, this would double-escape it.

sebastianmarzjan:
Thanks @Grunny !

PR #10897, opened 2016-07-22
PLATFORM-2331: Expose PPFrame::getTitle to Lua

macbre:
This field already exists in PHP with exactly the content requested in
bug 47089, so we may as well expose it on the frame object.

Backporting https://github.com/wikimedia/mediawiki-extensions-Scribunto/commit/0763e2229279487bcfd6caefe9c427458fb0a8d5

Fixes the most common Lua error: `attempt to call method 'getTitle' (a nil value).`

``` lua
function cs1.citation(frame)
    local pframe = frame:getParent()
    local validation, utilities, identifiers, metadata;
```

``` lua
    if nil ~= string.find (frame:getTitle(), 'sandbox', 1, true) then           -- did the {{#invoke:}} use sandbox version?
            local parent = frame:getParent()
            if not parent then
                fargs = frame.args
            else
                local title = parent:getTitle():gsub('/sandbox$', '')
```

@Grunny

### DIFF
--- a/extensions/Scribunto/engines/LuaCommon/LuaCommon.php
+++ b/extensions/Scribunto/engines/LuaCommon/LuaCommon.php
@@ -106,6 +106,7 @@ abstract class Scribunto_LuaEngine extends ScribuntoEngineBase {
 			'expandTemplate',
 			'preprocess',
 			'incrementExpensiveFunctionCount',
+			'getFrameTitle',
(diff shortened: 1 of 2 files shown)


Grunny:
:+1:

macbre:
Backported to release branch in e68093b

PR #10901, opened 2016-07-22
ADEN-3584 Remove BLB from non-UAP scenarios

fraszczakszymon:


bognix:
LGTM

extensions/wikia/AdEngine/js/provider/directGptMobile.js
```diff
@@ -2,14 +2,17 @@
...
 	'ext.wikia.adEngine.uapContext',
+	'wikia.document',
 	'wikia.window'
-], function (factory, uapContext, win) {
+], function (factory, uapContext, doc, win) {
 	'use strict';
 
 	// TODO: ADEN-3542
 	function dispatchNoUapEvent(slotName) {
+		var event = doc.createEvent('CustomEvent');
```
nandy-andy: Does it make sense to create `CustomEvent` even though we may not use it if the condition below isn't met?

extensions/wikia/AdEngine/js/template/bfaa.js
```diff
@@ -134,9 +135,8 @@ define('ext.wikia.adEngine.template.bfaa', [
 		uapContext.setUapId(params.uap);
 		unblockedSlots.forEach(btfBlocker.unblock);
 
-		if (skin === 'mercury') {
-			win.dispatchEvent(new Event('wikia.uap'));
-		}
+		event.initCustomEvent('wikia.uap', true, true, {});
```
nandy-andy:
```
var event = doc.createEvent('CustomEvent');
event.initCustomEvent('wikia.uap', true, true, {});
win.dispatchEvent(event);
```

are lines which showed up twice in this PR. Do you think it makes sense extracting them and use one and the same function in those two places:question:

drets:
LGTM :+1:

PR #10910, opened 2016-07-25
WW-17 | Update twitter embed widget

kerriamber:
ping @Wikia/west-wing


extensions/wikia/TwitterTag/TwitterTagController.class.php
```diff
@@ -5,15 +5,14 @@ class TwitterTagController extends WikiaController {
 	const PARSER_TAG_NAME = 'twitter';
 	const TWITTER_NAME = 'Twitter';
 	const TWITTER_BASE_URL = 'https://twitter.com/';
+	const TWITTER_USER_TIMELINE = '/^https:\/\/twitter.com\/[a-z0-9_]{1,15}$/i';
```
sebastianmarzjan: I wonder if we want to be as restrictive as to limit to usernames only, or go with more permissive approach to allow for URLs like "https://twitter.com/hashtag/corgi"?

sebastianmarzjan:
LGTM

sebastianmarzjan:
LGTM

PR #10913, opened 2016-07-25
WW-15 | Better logging

rybmat:
Ticket: https://wikia-inc.atlassian.net/browse/WW-15

Since current logging is not very descriptive, I'd like to add some more information to logs in order to dig deeper into the bug.

@Wikia/west-wing

### DIFF
--- a/extensions/VisualEditor/ApiVisualEditor.php
+++ b/extensions/VisualEditor/ApiVisualEditor.php
@@ -125,7 +125,9 @@ class ApiVisualEditor extends ApiBase {
 			}
 			\Wikia\Logger\WikiaLogger::instance()->error( 'ApiVisualEditor_requestParsoid', [
 				'method' => $method,
-				'error' => 'parsoidserver-' . $code
+				'error' => 'parsoidserver-' . $code,
+				'errors' => json_encode($errors),
+			    'status_code' => $req->getStatus()
 			] );
 			$this->dieUsage( "$message: " . $req->getContent(), 'parsoidserver-' . $code );
 		}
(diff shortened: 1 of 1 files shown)


extensions/VisualEditor/ApiVisualEditor.php
```diff
@@ -125,7 +125,9 @@ protected function requestParsoid( $method, $title, $params ) {
 			}
 			\Wikia\Logger\WikiaLogger::instance()->error( 'ApiVisualEditor_requestParsoid', [
 				'method' => $method,
-				'error' => 'parsoidserver-' . $code
+				'error' => 'parsoidserver-' . $code,
```
sebastianmarzjan: Would it make sense to rename it to error_code and remove the 'parsoidserver-' part?

extensions/VisualEditor/ApiVisualEditor.php
```diff
@@ -125,7 +125,9 @@ protected function requestParsoid( $method, $title, $params ) {
...
+				'errors' => json_encode($errors),
```
sebastianmarzjan: Maybe error_messages ?

extensions/VisualEditor/ApiVisualEditor.php
```diff
@@ -125,7 +125,9 @@ protected function requestParsoid( $method, $title, $params ) {
...
+			    'status_code' => $req->getStatus()
```
sebastianmarzjan: Spacing :)

extensions/VisualEditor/ApiVisualEditor.php
rybmat: error_code may be missleading in my opinion as $code in this place has always value 'http-bad-status'

sebastianmarzjan:
LGTM

extensions/VisualEditor/ApiVisualEditor.php
rybmat: actually we could get rid of this whole line...

sebastianmarzjan: Go for it :)

PR #10914, opened 2016-07-25
SUS-738 fix Broken chat module for logged out users

sqreek:
[SUS-738](https://wikia-inc.atlassian.net/browse/SUS-738)

@mixth-sense


ghost:
Can we `grep` for more `.get(` that should be replaced with `.eq(`?

I guess it might be pretty hard without the right context. I am just asking.

sqreek:
Lets make a ticket for this and do this in separate PR.

ghost:
Tested and accepted!

PR #10915, opened 2016-07-25
WW-15 | Better logging | to preview

rybmat:
Ticket: https://wikia-inc.atlassian.net/browse/WW-15

Since current logging is not very descriptive, I'd like to add some more information to logs in order to dig deeper into the bug.

@Wikia/west-wing


kerriamber:
LGTM

PR #10919, opened 2016-07-26
(SEC-71) Fix support for PHP's taint extension

Grunny:
MediaWiki had some experimental support for the initial taint feature
proposal, which isn't compatible with the newer PHP extension for this
purpose. This fixes the support so there aren't conflicts and makes
WikiaRequest match WebRequest in tainting default values when the
extension is enabled.

/cc @macbre @wladekb


macbre:
:+1:

PR #10923, opened 2016-07-26
SUS-771: Investigate Rail modules requesting CSS already loaded by the skin

mszabo-wikia:
https://wikia-inc.atlassian.net/browse/SUS-771

@macbre @mixth-sense


macbre:
:+1: 

Let's just check this change on preview (when it lands there) whether it does not pollute the ELK with too many log messages.

PR #10924, opened 2016-07-26
Paginator: fix XSS

macbre:
[PLATFORM-2341](https://wikia-inc.atlassian.net/browse/PLATFORM-2341)
- properly escape the HTML in the template
- do not expose `PaginationController` to wikia.php API endpoint

@Grunny / @wladekb

### DIFF
--- a/includes/wikia/services/templates/Pagination_index.php
+++ b/includes/wikia/services/templates/Pagination_index.php
@@ -9,8 +9,8 @@
 		<? $class = ($i == 0) ? ' first ' : ''; ?>
 		<? $class .= ($pages[$i] == $currentPage) ? ' selected ' : ''; ?>
 		
-		<? if( gettype($pages[$i]) == 'string' ): ?>
-			<li class="ellipsis"><?= $pages[$i] ?></li>
+		<? if( is_string($pages[$i]) ): ?>
+			<li class="ellipsis"><?= htmlspecialchars( $pages[$i] ) ?></li>
 		<? else: ?>
 			<li data-page="<?= $pages[$i]; ?>" class="<?= $class ?>">
 				<a href="<?= htmlspecialchars( $url !== '#' ? $url.$pages[$i] : $url ) ?>"><?= $pages[$i] ?></a>
--- a/includes/wikia/services/templates/Pagination_index.php
+++ b/includes/wikia/services/templates/Pagination_index.php
@@ -1,7 +1,7 @@
-<ul class="Pagination"  <? foreach($data as $key => $val ): ?> data-<?php echo $key ?>="<?php echo htmlentities($val); ?>" <? endforeach; ?> >
+<ul class="Pagination"  <? foreach($data as $key => $val ): ?> data-<?= htmlspecialchars( $key ) ?>="<?php echo htmlentities($val); ?>" <? endforeach; ?> >
 	<? if( $prev ): ?>
 		<li data-page="<?= $currentPage - 1; ?>"  class="prev">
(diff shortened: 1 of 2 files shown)


includes/wikia/services/templates/Pagination_index.php
```diff
@@ -1,26 +1,26 @@
-<ul class="Pagination"  <? foreach($data as $key => $val ): ?> data-<?php echo $key ?>="<?php echo htmlentities($val); ?>" <? endforeach; ?> >
+<ul class="Pagination"  <? foreach($data as $key => $val ): ?> data-<?= htmlspecialchars( $key ) ?>="<?php echo htmlentities($val); ?>" <? endforeach; ?> >
 	<? if( $prev ): ?>
 		<li data-page="<?= $currentPage - 1; ?>"  class="prev">
-			<a href="<?= ($url !== '#' ? $url.($currentPage - 1) : $url) ?>"><?= $prevMsg ?></a>
+			<a href="<?= htmlspecialchars( $url !== '#' ? $url.($currentPage - 1) : $url) ?>"><?= htmlspecialchars( $prevMsg ) ?></a>
```
Grunny: For attributes `Sanitizer::encodeAttribute` is preferred over plain `htmlspecialchars` (both are safe in this instance though due to the double quotes).

includes/wikia/services/templates/Pagination_index.php
```diff
@@ -1,26 +1,26 @@
-<ul class="Pagination"  <? foreach($data as $key => $val ): ?> data-<?php echo $key ?>="<?php echo htmlentities($val); ?>" <? endforeach; ?> >
+<ul class="Pagination"  <? foreach($data as $key => $val ): ?> data-<?= htmlspecialchars( $key ) ?>="<?php echo htmlentities($val); ?>" <? endforeach; ?> >
```
Grunny: Is the `$key` in the users of this class influenced by users at all? If so, should we (in addition to escaping) validate that `$key` is a valid attribute name (no spaces or strange characters) since `htmlspecialchars` won't escape spaces and so forth? i.e. something like `Sanitizer::validateTagAttributes`?

includes/wikia/services/templates/Pagination_index.php
macbre: Addressed in 0802f8e

Grunny:
One question above, otherwise LGTM :lock:

includes/wikia/services/templates/Pagination_index.php
macbre: Nope, `Forum_board` is the only caller that passes `data` to `Paginator` controller - the values there are hardcoded.

wladekb:
lgtm

includes/wikia/services/templates/Pagination_index.php
```diff
@@ -1,26 +1,26 @@
...
-			<a href="<?= ($url !== '#' ? $url.($currentPage - 1) : $url) ?>"><?= $prevMsg ?></a>
+			<a href="<?= Sanitizer::encodeAttribute( $url !== '#' ? $url.($currentPage - 1) : $url) ?>"><?= htmlspecialchars( $prevMsg ) ?></a>
 		</li>
 	<? endif; ?>
 	
 	<? for( $i=0; $i<count($pages); $i++ ): ?>
 		<? $class = ($i == 0) ? ' first ' : ''; ?>
 		<? $class .= ($pages[$i] == $currentPage) ? ' selected ' : ''; ?>
 		
-		<? if( gettype($pages[$i]) == 'string' ): ?>
```
wladekb, line 12: `gettype` - wow!

PR #10929, opened 2016-07-26
added  parameter in callApi function SER-343

kpiotrowski:
Added  parameter in callApi function (it was added only in overridden method). Another solution is to remove $endpointPath parameter from parent method (it's unused), but it was added 19 days ago, so I suppose it'll be used it the future.


macbre:
:+1:

macbre:
Thanks :)

PR #10931, opened 2016-07-26
XW-1756 | add wds prefix for new DSv0.0.7

tomasznapieralski:
https://wikia-inc.atlassian.net/browse/XW-1756

@Wikia/x-wing


rogatty:
LGTM

PR #10936, opened 2016-07-27
MWExceptionHandler: log MW exceptions to ELK

macbre:
Preview returns HTTP 500 and Kibanna shows the following:

```
MediaWiki internal error.

Exception caught inside exception handler.

Set $wgShowExceptionDetails = true; at the bottom of LocalSettings.php to show detailed debugging information.
```

We need more context here - log the exception

@ludwikkazmierczak / @Grunny


Grunny:
:+1:

PR #10940, opened 2016-07-27
SUS-493: Sunset AdminDashboard article chrome

mszabo-wikia:
https://wikia-inc.atlassian.net/browse/SUS-493
- Removed `AdminDashboardLogic::displayAdminDashboard` - it's enough to check if we are on Special:AdminDashboard now
- Removed `AdminDashboardSpecialController::chromedArticleHeader` and template - no longer needed
- Removed `AdminDashboardSpecialController::GetSpecialPage` and template - wasn't used anywhere
- Removed `resources/wikia/libraries/excanvas/excanvas.js` - was only used on AdminDashboard to provide `<canvas>` support for legacy IE <8 which are no longer supported by Wikia
- Also simplified controller logic and assets management

@mixth-sense


vforge:
kristen-wiig-kk2s3JrkduMQ8
LGTM

PR #10943, opened 2016-07-27
XW-1574 | Add Design System styles to all Oasis pages

rogatty:
https://wikia-inc.atlassian.net/browse/XW-1574

Add DS styles to all pages. They're currently not used but we want to battle-test if they don't break anything (**shouldn't**).


extensions/wikia/DesignSystem/DesignSystemHooks.class.php
```diff
@@ -0,0 +1,59 @@
...
+	const DESIGN_SYSTEM_DIR = __DIR__ . '/bower_components/design-system/dist';
+
+	/**
+	 * @desc Adds Design System styles to all Oasis pages
+	 *
+	 * @param OutputPage $out
+	 *
+	 * @return bool true
+	 */
+	public static function onBeforePageDisplay( $out, &$skin ) {
```
Grunny: No need for the `&` on `$skin`. Objects in PHP are (effectively) passed by reference anyway.

extensions/wikia/DesignSystem/DesignSystem.setup.php
```diff
@@ -0,0 +1,22 @@
+<?php
+/**
+ * Design System Extension
+ */
+$dir = dirname( __FILE__ ) . '/';
```
Grunny: `__DIR__` is preferred over `dirname( __FILE__ )`.

extensions/wikia/DesignSystem/DesignSystemHelper.class.php
```diff
@@ -0,0 +1,65 @@
...
+			}
+
+			return $xml->asXML();
+		} else {
+			WikiaLogger::instance()->error( 'Design System SVG could not be loaded', [
+				'group' => $group,
+				'name' => $name
+			] );
+
+			return $alt || $name || '';
```
Grunny: This would return a boolean in PHP.

extensions/wikia/DesignSystem/DesignSystemHelper.class.php
```diff
@@ -0,0 +1,65 @@
...
+		$xml = self::getCachedSvg( $group, $name );
+
+		if ( $xml instanceof SimpleXMLElement ) {
+			/* @var $xml SimpleXMLElement */
+
+			if ( !empty( $classNames ) ) {
+				$xml->addAttribute( 'class', $classNames );
+			}
+
+			if ( !empty( $classNames ) ) {
```
Grunny: Should be `$alt`?

extensions/wikia/DesignSystem/DesignSystemHelper.class.php
```diff
@@ -0,0 +1,65 @@
...
+	 * @param string $group
+	 * @param string $name
+	 *
+	 * @return SimpleXMLElement
+	 */
+	private static function getCachedSvg( $group, $name ) {
+		if ( isset( self::$svgCache[ $group ][ $name ] ) ) {
+			$xml = self::$svgCache[ $group ][ $name ];
+		} else {
+			$xml = simplexml_load_file( self::ASSETS_DIR . '/' . $group . '/' . $name . '.svg' );
```
vforge, line 59: Question: Why we read SVG as XML instead of just string (`file_read_contents`)?

extensions/wikia/DesignSystem/DesignSystemHelper.class.php
rogatty: :facepalm2:

extensions/wikia/DesignSystem/DesignSystemHelper.class.php
rogatty, line 59: Because we have to add attributes like `class` and `alt` to the root element. If we loaded it as a string we would have to use RegExp :speak_no_evil:

extensions/wikia/DesignSystem/DesignSystemHelper.class.php
```diff
@@ -0,0 +1,75 @@
...
+				'name' => $name
+			] );
+
+			$fallback = '';
+
+			if ( !empty( $alt ) ) {
+				$fallback = $alt;
+			}
+
+			if ( !empty( $name ) ) {
```
Grunny: Your previous version suggested you wanted to use `$alt` if it was available, and `$name` if it wasn't, but if both are set `$name` will be used as this will overwrite it. Is that the intent, or should this be an `elseif`?

Grunny:
One question, otherwise LGTM.

Warkot:
LGTM

extensions/wikia/DesignSystem/DesignSystemHelper.class.php
vforge, line 59: 👍

PR #10949, opened 2016-07-28
PLATFORM-2360: AsyncTaskList - dispatch staging tasks via HTTP

macbre:
[PLATFORM-2360](https://wikia-inc.atlassian.net/browse/PLATFORM-2360)

Use http://staging.task.service.consul/proxy.php entry point instead of a fallback to SSH execution

@wladekb


wladekb:
lgtm

PR #10952, opened 2016-07-28
ADEN | require optional prebid so it won't throw

bognix:
@nandy-andy


bognix:
Unit tests are failing on dev as well.

Failure not related with the code change

nandy-andy:
LGTM :+1:

PR #10956, opened 2016-07-28
RECIRC 171: Add result dithering to some experiments

pauloslund:
In an effort to make recommendations more _fresh_ and varied, we are adding a certain amount of result dithering on the client to randomize the order a little bit. You can read more about dithering [here](https://buildingrecommenders.wordpress.com/2015/11/11/dithering/)


extensions/wikia/Recirculation/RecirculationApiController.class.php
```diff
@@ -114,4 +111,24 @@ private function getParamType() {
 
 		return $type;
 	}
+
+	private function getParamFill() {
+		$fill = $this->request->getVal( 'fill', false );
```
saipetch: consider using $this->request->getBool( 'fill', false );

saipetch:
LGTM

PR #10957, opened 2016-07-28
RECIRC-189: Switch to using wfShortenText for better multibyte support

pauloslund:
[RECIRC-189](https://wikia-inc.atlassian.net/browse/RECIRC-189)
Multibyte characters are getting corrupted when attempting to truncate them. Fix the issue by switching to wfShortenText which has better multibyte character support.


katanka:
LGTM

PR #10958, opened 2016-07-28
SOC-1408 use batches for migration script

jsutterfield:
As I was preparing to use the migration script which cleans up old values from the `wikicities.user_properties` table, I realized the original approach didn't scale with the amount of data we have in production. I decided to update the script to run in batches and only delete one attribute at a time, rather than trying to do all at once. I also added a sleep between each delete cycle to reduce the amount of replication lag. See [this SO post](http://stackoverflow.com/questions/1318972/deleting-millions-of-rows-in-mysql#answer-1318975).

Can you take another look @armonr?

Also, /cc @drozdo and @frankfarmer. Please let me know if either of you foresee any problems with this new approach. Here's the original PR for additional context: https://github.com/Wikia/app/pull/10875

### DIFF
--- a/lib/Wikia/src/Service/User/Attributes/maintenance/cleanupMigratedAttributes.php
+++ b/lib/Wikia/src/Service/User/Attributes/maintenance/cleanupMigratedAttributes.php
@@ -73,7 +73,6 @@ class CleanupMigratedAttributes extends Maintenance {
                 ->run( $db );
 
             $rowsRemaining = $db->affectedRows() > 0;
-            $db->commit();
 
             sleep( 2 );
         }
--- a/lib/Wikia/src/Service/User/Attributes/maintenance/cleanupMigratedAttributes.php
+++ b/lib/Wikia/src/Service/User/Attributes/maintenance/cleanupMigratedAttributes.php
@@ -19,6 +19,8 @@ require_once( __DIR__ . '/../../../../../../../maintenance/Maintenance.php' );
 
 class CleanupMigratedAttributes extends Maintenance {
 
+    CONST BATCH_SIZE = 1000;
+
     private $dryRun;
 
     public function __construct() {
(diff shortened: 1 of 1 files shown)


lib/Wikia/src/Service/User/Attributes/maintenance/cleanupMigratedAttributes.php
```diff
@@ -42,21 +44,39 @@ private function printAttrsToBeDeleted() {
             $this->output( "NOT DRY RUN, ATTRIBUTES WILL BE DELETED\n" );
         }
         $this->output( "Deletion will begin in 5 seconds...\n" );
-        sleep(5);
+        sleep( 5 );
     }
     
     private function deleteAttributes() {
-        global $wgExternalSharedDB, $wgPublicAttributes;
+        global $wgExternalSharedDB, $wgPublicUserAttributes;
```
jsutterfield, line 51: `$wgPublicUserAttributes` looks like this:

``` php
$wgPublicUserAttributes = [
    "avatar",
    "avatar_rev",
    "bio",
    "coverPhoto",
    "fancysig",
    "fbPage",
    "location",
    "name",
    "nickname",
    "occupation",
    "twitter",
    "website",
    "UserProfilePagesV3_birthday",
    "UserProfilePagesV3_gender"
];
```

lib/Wikia/src/Service/User/Attributes/maintenance/cleanupMigratedAttributes.php
```diff
@@ -42,21 +44,39 @@ private function printAttrsToBeDeleted() {
...
             return;
         }
 
-        $db = wfGetDB( DB_MASTER, [], $wgExternalSharedDB );
-        ( new WikiaSQL() )
-            ->DELETE( 'user_properties' )
-            ->WHERE( 'up_property' )->IN( $wgPublicAttributes )
-            ->run( $db );
+        $rowsRemaining = true;
+        while ( $rowsRemaining ) {
```
armonr, line 67: Should there be a `$db->begin_transaction();` here?

armonr:
Besides one comment, LGTM

lib/Wikia/src/Service/User/Attributes/maintenance/cleanupMigratedAttributes.php
jsutterfield, line 67: Good point. If we were trying to ensure this was being run in a transaction, then yes, I think it should be. In this case though, I was more concerned with transactions being ended after each batch delete. My thinking was that if autocommit was off, then there'd be a chance that these batches wouldn't be commited until the script finished, and then we'd lose the whole advantage of doing it iteratively.

I just double checked though, and it looks like autocommit is on, so we can remove that `commit()` altogether:

```
mysql@master.db-sharedb.service.consul[wikicities]>select @@autocommit;
+--------------+
| @@autocommit |
+--------------+
|            1 |
+--------------+
1 row in set (0.00 sec)
```

PR #10961, opened 2016-07-28
XW-1771 | dried footer API code

Warkot:
@Wikia/x-wing 

I unified the way most of the `href`s are created, so now the `getData()` method looks this simple:

```
public function getData() {
    $this->setHeaderData();

    foreach ( $this->baseData as $sectionName => $sectionBaseData ) {
        $this->setSectionData( $sectionName, $sectionBaseData );
    }

    $this->setLicenseData();

    return $this->data;
}
```

 If you like it, I'll fix the tests.

### DIFF
--- a/includes/wikia/models/DesignSystemGlobalFooterModel.class.php
+++ b/includes/wikia/models/DesignSystemGlobalFooterModel.class.php
@@ -525,6 +525,11 @@ class DesignSystemGlobalFooterModel extends WikiaModel {
 		$this->hrefs = $hrefs;
 	}
 
+	/**
+	 * Get prepared/parsed data.
+	 *
+	 * @return array
+	 */
 	public function getData() {
 		$this->setHeaderData();
 		$this->setSectionsData();
@@ -533,6 +538,9 @@ class DesignSystemGlobalFooterModel extends WikiaModel {
 		return $this->data;
 	}
 
+	/**
+	 * Add footer headers to $this->data property depending on the chosen language.
+	 */
 	private function setHeaderData() {
 		if ( $this->lang === self::DEFAULT_LANG ) {
 			$this->data['fandom'] = [
@@ -574,12 +582,21 @@ class DesignSystemGlobalFooterModel extends WikiaModel {
 		}
 	}
 
+	/**
+	 * Add footer sections to $this->data property by reading $this->baseData property.
+	 */
 	private function setSectionsData() {
 		foreach ( $this->baseData as $sectionName => $sectionBaseData ) {
 			$this->setSectionData( $sectionName, $sectionBaseData );
 		}
 	}
 
+	/**
+	 * Add single footer section to $this->data property.
+	 *
+	 * @param string $sectionName
+	 * @param array $sectionBaseData
+	 */
 	private function setSectionData( $sectionName, $sectionBaseData ) {
 		$sectionData = $this->getSectionData( $sectionBaseData );
 
@@ -588,6 +605,14 @@ class DesignSystemGlobalFooterModel extends WikiaModel {
 		}
 	}
 
+	/**
+	 * Get data for a single footer section.
+	 * Returned data has link href keys parsed to hrefs, making it ready to be added to the response.
+	 *
+	 * @param array $sectionBaseData
+	 *
+	 * @return array
+	 */
 	private function getSectionData( $sectionBaseData ) {
 		if ( !empty( $sectionBaseData['links'] ) ) {
 			$linksData = $this->getLinksData( $sectionBaseData['links'] );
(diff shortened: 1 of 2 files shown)


includes/wikia/models/DesignSystemGlobalFooterModel.class.php
```diff
@@ -188,202 +196,350 @@ class DesignSystemGlobalFooterModel extends WikiaModel {
...
-	public function setHrefs( $hrefs ) {
-		$this->hrefs = $hrefs;
-	}
-
-	public function getData() {
-		$data = [
-			'company_overview' => [
-				'header' => [
-					'type' => 'line-text',
+	private $baseData = [
```
Warkot, line 199: I was thinking about renaming `$baseData` to `$dataTemplate` (and other `base*` variables to `*template` too). Opinions? Maybe something else?

tomasznapieralski, line 199: how about `$schemaData`? you could create method `getSchemaData`

includes/wikia/models/DesignSystemGlobalFooterModel.class.php
```diff
@@ -188,202 +196,350 @@ class DesignSystemGlobalFooterModel extends WikiaModel {
...
+	public function getData() {
+		$this->setHeaderData();
+
+		foreach ( $this->baseData as $sectionName => $sectionBaseData ) {
+			$this->setSectionData( $sectionName, $sectionBaseData );
+		}
+
+		$this->setLicenseData();
+
+		return $this->data;
```
tomasznapieralski, line 542: this could be more readable if you have:
- $this -> setHeaderData();
- $schemaData = $this -> getSchemaData();
- return setOutputData($schemaData);

setOutputData could have:
- setSectionData()
- setLicenseData()

includes/wikia/models/DesignSystemGlobalFooterModel.class.php
tomasznapieralski, line 199: i would move this to separate file. this var takes about 300 lines.

rogatty, line 199: Schema describes the shape of data so `$schemaData` doesn't really fit here. `$baseData` sounds fine to me.

includes/wikia/models/DesignSystemGlobalFooterModel.class.php
```diff
@@ -420,216 +576,80 @@ public function getData() {
...
-				'type' => 'link-image',
-				'image' => 'wds-icons-instagram',
-				'title' => [
-					'type' => 'translatable-text',
-					'key' => 'global-footer-fandom-follow-us-link-instagram'
-				],
-				'href' => $this->getHref( 'social-instagram' )
-			];
+			if ( $href ) {
+				$linkBaseData['href'] = $href;
```
Grunny, line 649: Would we also want to unset `href-key` so it's not in the response?

Warkot, line 649: Good catch!

includes/wikia/models/DesignSystemGlobalFooterModel.class.php
```diff
@@ -420,216 +576,80 @@ public function getData() {
...
-				'type' => 'translatable-text',
-				'key' => 'global-footer-fandom-link-fan-communities'
-			],
-			'href' => $this->getHref( 'fan-communities' )
-		];
-
-		return $out;
-	}
+		if ( !empty( $sectionBaseData['description'] ) ) {
+			$sectionData['description'] = $sectionBaseData['description'];
```
Grunny: Do we need these two parts for the header and description since they don't modify `$sectionBaseData`? This method could in theory just be something like:

``` php
    private function getSectionData( $sectionBaseData ) {
        if ( !empty( $sectionBaseData['links'] ) ) {
            $linksData = $this->getLinksData( $sectionBaseData['links'] );
            if ( !empty( $linksData ) ) {
                $sectionBaseData['links'] = $linksData;
            } else {
                unset( $sectionBaseData['links'] );
            }
        }

        return $sectionBaseData;
    }
```

Warkot: Correct me if I understand wrong, but your suggestion is to just use `$this->baseData` and replace parsed `href`s of the links in sections in place—in the `$baseData` object—and return it. Without copying the contents of `$this->baseData` to `$this->data`. Right?

includes/wikia/models/DesignSystemGlobalFooterModel.class.php
Warkot, line 542: `headerData` is be also part of the output, so it should be in the `setOutputData`. Logically we have:
- the header,
- multiple sections,
- license, which is different then the sections above.

So the only change I would consider is to wrap the `foreach` loop into a `setSectionsData()` method.

Warkot:
Moved to https://github.com/Wikia/app/pull/10973

PR #10965, opened 2016-07-29
SUS-760: Add only valid images to image review queue

lukaszkonieczny:
We are excluding from image review queue pages from File namespace (NS_FILE):
- from newly created wiki
- which are not images
- which don't exists
- which are redirects

@harnash @mixth-sense @wladekb


harnash:
LGTM

lukaszkonieczny:
LGTM

PR #10966, opened 2016-07-29
PLATFORM-2362: Add proper support for incoming X-Trace-Id

wladekb:
So far we have only looked at X-Request-Id to retrieve Trace ID for current transaction. This makes Mediawiki aligned with our standards and looks at X-Trace-Id as well.

/cc @macbre @pwojcik86


macbre:
:+1:

PR #10969, opened 2016-07-29
SUS-126: Fix scrolling in classic editor preview mode

mszabo-wikia:
ticket & QA notes: https://wikia-inc.atlassian.net/browse/SUS-126
@mixth-sense

### DIFF
--- a/extensions/wikia/EditPreview/EditPreview.setup.php
+++ b/extensions/wikia/EditPreview/EditPreview.setup.php
@@ -6,7 +6,6 @@
  * @author Andrzej 'nAndy' Łukaszewski
  * @author Jacek 'mech' Woźniak
  */
-$dir = dirname(__FILE__) . '/';
 
 $wgExtensionCredits['editpreview'][] = [
 	'name' => 'Edit preview',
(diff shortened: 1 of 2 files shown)


extensions/wikia/EditPreview/EditPreview.setup.php
```diff
@@ -21,16 +20,4 @@
 ];
 
 // messages
-$wgExtensionMessagesFiles[ 'EditPreview' ] = $dir . 'EditPreview.i18n.php';
```
vforge, line 24: Since we're removing including this i18n file do we still need the file itself?

extensions/wikia/EditPreview/js/preview.js
```diff
@@ -60,6 +60,11 @@ define('wikia.preview', [
 					'overflow-x': 'hidden'
 				});
 
+				// SUS-126: prevent page scrolling on large pages breaking the classic editor
+				$(document.body).css({
```
vforge: Please use `$('body').css('overflow-y', 'hidden')` <- both lookup and function will be faster

extensions/wikia/EditPreview/EditPreview.setup.php
mszabo-wikia, line 24: The messages are still used by the EditPageLayout JSMessages package (https://github.com/Wikia/app/blob/dev/extensions/wikia/EditPageLayout/EditPageLayout_setup.php#L84 ), only the EditPreview JSMessages package is unused

vforge, line 24: OK

vforge:
LGTM

PR #10970, opened 2016-07-29
ADEN-3662 | Don't use shared memcache key for source point scripts

bognix:


jacekjursza:
LGTM

bognix:
Synthetic tests are passing, merging.

PR #10973, opened 2016-07-29
XW-1779 | dried footer API code

Warkot:
Reissue of https://github.com/Wikia/app/pull/10961

I unified the way most of the hrefs are created, so now the getData() method looks this simple:

```
public function getData() {
    $this->setHeaderData();
    $this->setSectionsData();
    $this->setLicenseData();

    return $this->data;
}
```

I also added doc-blocks. 😁

### DIFF
--- a/includes/wikia/models/DesignSystemGlobalFooterModel.class.php
+++ b/includes/wikia/models/DesignSystemGlobalFooterModel.class.php
@@ -525,6 +525,11 @@ class DesignSystemGlobalFooterModel extends WikiaModel {
 		$this->hrefs = $hrefs;
 	}
 
+	/**
+	 * Get prepared/parsed data.
+	 *
+	 * @return array
+	 */
 	public function getData() {
 		$this->setHeaderData();
 		$this->setSectionsData();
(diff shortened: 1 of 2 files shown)


includes/wikia/models/DesignSystemGlobalFooterModel.class.php
```diff
@@ -420,218 +584,126 @@ public function getData() {
...
-			'type' => 'link-text',
-			'title' => [
-				'type' => 'text',
-				'value' => $licenseText
-			],
-			'href' => $this->getLicenseUrl()
-		];
-	}
-
-	private function getFandomOverview() {
```
kvas-damian, line 440: Probably we should adjust unittests or at least its manes to reflect this change.

includes/wikia/models/DesignSystemGlobalFooterModel.class.php
```diff
@@ -188,202 +196,358 @@ class DesignSystemGlobalFooterModel extends WikiaModel {
...
+	 * Get prepared/parsed data.
+	 *
+	 * @return array
+	 */
+	public function getData() {
+		$this->setHeaderData();
+		$this->setSectionsData();
+		$this->setLicenseData();
+
+		return $this->data;
```
kvas-damian, line 542: I personally don't this kind of code structure. We call few methods that do some magic and don't return any value. Finally we return some attribute. It pretty often leads to some not easy to follow flows.
Personally I think methods returning values are much easier to follow. What's your opinion about it?

PR #10975, opened 2016-08-01
XW-1706 | DS GlobalFooterAPI update link to media-kit contact

kvas-damian:
Link was changed recently so we should update it ;)


rogatty:
LGTM

PR #10982, opened 2016-08-01
SOC-2323 As a user, I want to see my discussion post count in the masthead on my profile

mleczey:
Links:
https://wikia-inc.atlassian.net/browse/SOC-2323

Description:
As a user, I want to see my discussion post count in the masthead on my profile

Reviewers:
@bkoval @rafalkalinski @jsutterfield


extensions/wikia/UserProfilePageV3/UserIdentityBoxDiscussionInfo.class.php
```diff
@@ -0,0 +1,108 @@
...
+	public function getDiscussionPostsCount() {
+		return $this->discussionPostsCount;
+	}
+
+	public function getDiscussionAllPostsByUserLink() {
+		return $this->discussionAllPostsByUserLink;
+	}
+
+	public function fetchDiscussionPostsNumber() {
+		wfProfileIn( __METHOD__ );
```
garthwebb: These wfProfileIn|Out calls are unnecessary.  They come from a time when there was no other way in PHP to profile call times.  Now xdebug supports this and we send this information to influxdb from within the PHP interpreter rather than within our code.

bkoval:
LGTM 🔑

extensions/wikia/UserProfilePageV3/UserIdentityBoxDiscussionInfo.class.php
```diff
@@ -0,0 +1,92 @@
+<?php
+
+use Swagger\Client\ApiException;
+use Swagger\Client\Discussion\Api\ContributionApi;
+use Swagger\Client\Discussion\Api\SitesApi;
+use Wikia\DependencyInjection\Injector;
+use Wikia\Service\Swagger\ApiProvider;
+
+class UserIdentityBoxDiscussion {
```
jsutterfield: Could we rename this to `UserIdentityBoxDiscussionInfo`? Right now, the class and the file have different names.

extensions/wikia/UserProfilePageV3/UserIdentityBoxDiscussionInfo.class.php
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	public function getDiscussionAllPostsByUserLink() {
+		return $this->discussionAllPostsByUserLink;
+	}
+
+	public function fetchDiscussionPostsNumber() {
+		$siteId = F::app()->wg->CityId;
+
+		$this->discussionActive = $this->checkDiscussionActive($siteId);
```
jsutterfield: I'm seeing a number of whitespace violations in this PR. Could we update so it follows [our guidelines](https://github.com/Wikia/guidelines/blob/master/PHP/CodingConventions.md)?

Mostly what I'm seeing is padding need. Eg

``` php
// bad
$this->discussionActive = $this->checkDiscussionActive($siteId);

// good
$this->discussionActive = $this->checkDiscussionActive( $siteId );
```

extensions/wikia/UserProfilePageV3/UserIdentityBoxDiscussionInfo.class.php
```diff
@@ -0,0 +1,92 @@
...
+			$this->discussionPostsCount = $this->fetchDiscussionPostsCount($siteId);
+			$this->discussionAllPostsByUserLink = "/d/u/{$this->userId}";
+		}
+	}
+
+	private function checkDiscussionActive($siteId) {
+		try {
+			$this->getDiscussionSitesApi()->getSite( $siteId );
+			return true;
+		} catch ( ApiException $e ) {
```
jsutterfield, line 52: Just to double check, a 404 results in an ApiException?

extensions/wikia/UserProfilePageV3/UserIdentityBoxDiscussionInfo.class.php
```diff
@@ -0,0 +1,92 @@
...
+					'siteId' => $siteId,
+					'userId' => $this->userId,
+					'error' => $e->getMessage()
+				] );
+		}
+
+		return $postCount;
+	}
+
+	private function getDiscussionContributionApi() {
```
jsutterfield, line 98: Can we add some PHPDocs to help IDEs with reflection?

Eg:

``` php
    /**
     * @return ContributionApi
     */
    private function getDiscussionContributionApi() {
        return $this->getDiscussionApi(ContributionApi::class);
    }
```

For me that helps PHPStorm know what kind of object it's working with. Might be nice to add here, `getDiscussionSitesApi`, and in the constructor.

extensions/wikia/UserProfilePageV3/UserIdentityBoxDiscussionInfo.class.php
```diff
@@ -0,0 +1,92 @@
...
+		$api->getApiClient()->getConfig()->setCurlTimeout(self::TIMEOUT);
+
+		return $api;
+	}
+
+	private function fetchDiscussionPostsCount($siteId) {
+		$postCount = 0;
+
+		try {
+			$postCount = $this->getDiscussionContributionApi()->getPosts($siteId, $this->userId)['post_count'];
```
jsutterfield: `getPosts()` returns a `ForumThreadHalResponse` object which has a `getPostCount()` method. I think we should probably favor that getter here.

extensions/wikia/UserProfilePageV3/UserIdentityBoxDiscussionInfo.class.php
mleczey, line 52: Code was based on `SpecialDiscussionsController.class.php`, which handles errors these way. I could only confirm that 404 responses are caught.

jsutterfield:
LGTM!

PR #10986, opened 2016-08-02
SUS-802: Invalid Sass files should not break the entire group request

mszabo-wikia:
Since the length of URLs is limited, it's possible for certain AssetsManager queries to have their end cut off, which causes an exception that results in no CSS being served.

This PR changes the exception to logging so that we can still serve the valid assets from the group.

Ticket & QA notes: https://wikia-inc.atlassian.net/browse/SUS-802
@vforge


vforge:
LGTM

PR #10987, opened 2016-08-02
SOC-2770 | Pass langCode to New Auth when redirecting from old Signup / Login

bkoval:
@Wikia/social-frontend 

https://wikia-inc.atlassian.net/browse/SOC-2770


rafalkalinski:
LGTM

PR #10989, opened 2016-08-02
PLATFORM-2359: fix sandbox detection when not running using Apache module

macbre:
[PLATFORM-2359](https://wikia-inc.atlassian.net/browse/PLATFORM-2359)

Previously `getallheaders()` function was used that only exists when PHP is run under Apache module. It returns headers names as they arrived to the server. When this function is not present, `$_SERVER` is used which normalizes headers names to uppercase and adds HTTP prefix.

```
staging-fpm:
'HTTP_X_STAGING' => sandbox-fpm

staging-s2
'X-Staging' => sandbox-s2
```

To make it consistent for both Apache module and fpm let's use `$_SERVER` in both cases.

Additionally, remove `Wikia::getAllHeaders()` as it was used only by WikiFactoryLoader.

@artursitarski / @wladekb

### DIFF
--- a/extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
+++ b/extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
@@ -685,13 +685,13 @@ class WikiFactoryLoader {
 				}
 
 				if ($key == 'wgServer') {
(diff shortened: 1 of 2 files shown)


artursitarski:
:+1:

PR #10991, opened 2016-08-02
(XW-1761) Fix notice in MercuryApi when scheme values aren't set

Grunny:
Fixes the PHP notice:

```
PHP Notice: Undefined index: android-scheme in extensions/wikia/
MercuryApi/MercuryApiController.class.php on line 48
```

/cc @Wikia/x-wing


kvas-damian:
LGTM

rogatty:
LGTM

PR #10993, opened 2016-08-03
ADEN-3686 Disable detection if recovery is enabled

drets:
Disable SourcePoint detection when recovery is enabled


fraszczakszymon:
LGTM :+1:

extensions/wikia/AdEngine/js/spec/AdContext.spec.js
```diff
@@ -449,6 +449,27 @@ describe('AdContext', function () {
 		expect(getModule().getContext().opts.sourcePointDetection).toBeFalsy();
 	});
 
+	it('disables Source Point detection when Source Point recovery is enabled', function () {
```
nandy-andy, line 452: :heart:

PR #10995, opened 2016-08-03
WW-89: Checking if data returned from cache is an array

rybmat:
Ticket: https://wikia-inc.atlassian.net/browse/WW-89

Error was occurred because "get" method from cache returns false if there is no data for given key.

@Wikia/west-wing

### DIFF
--- a/extensions/wikia/InsightsV2/helpers/InsightsCache.php
+++ b/extensions/wikia/InsightsV2/helpers/InsightsCache.php
@@ -18,7 +18,7 @@ class InsightsCache {
 
 	public function get( $params ) {
 		$data = $this->memc->get( $this->getMemcKey( $params ) );
-		return is_array($data) ? $data : [];
+		return is_array( $data ) ? $data : [];
 	}
 
 	public function set( $params, $data, $ttl = self::INSIGHTS_MEMC_TTL ) {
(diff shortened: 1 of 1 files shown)


sebastianmarzjan:
LGTM

extensions/wikia/InsightsV2/helpers/InsightsCache.php
```diff
@@ -17,7 +17,8 @@ public function __construct( InsightsConfig $config ) {
 	}
 
 	public function get( $params ) {
-		return $this->memc->get( $this->getMemcKey( $params ) );
+		$data = $this->memc->get( $this->getMemcKey( $params ) );
+		return is_array($data) ? $data : [];
```
lukaszkonieczny: Please add spaces inside brackets per convention

lukaszkonieczny:
I think after this change we should change condition here https://github.com/Wikia/app/blob/e5aa9665b7b2122cbb6c602b8e7dd39869f79653/extensions/wikia/InsightsV2/helpers/InsightsCache.php#L86

sebastianmarzjan:
'change' as in 'remove'?

lukaszkonieczny:
I will check `!empty()`

rybmat:
good catch!

sebastianmarzjan:
👍

PR #10996, opened 2016-08-03
PLATFORM-2367: stable env was unstable

macbre:
- add `X-MediaWiki-Exception: 1` response header when MediaWiki explicitly returns HTTP 500
- try to log an exception in `MWExceptionHandler::escapeEchoAndDie`

@drozdo / @artursitarski / @wladekb


wladekb:
lgtm

PR #10999, opened 2016-08-03
XW-1575 | Introduce Global Footer from Design System

rogatty:
https://wikia-inc.atlassian.net/browse/XW-1575


extensions/wikia/DesignSystem/DesignSystemHelper.class.php
```diff
@@ -47,17 +45,16 @@ public static function getSvg( $group, $name, $classNames = '', $alt = '' ) {
...
 	 *
-	 * @param string $group
 	 * @param string $name
 	 *
 	 * @return SimpleXMLElement
 	 */
-	private static function getCachedSvg( $group, $name ) {
-		if ( isset( self::$svgCache[ $group ][ $name ] ) ) {
-			$xml = self::$svgCache[ $group ][ $name ];
+	public static function getCachedSvg( $name ) {
```
Grunny: Why public?

extensions/wikia/DesignSystem/controllers/DesignSystemGlobalFooterController.class.php
```diff
@@ -0,0 +1,39 @@
...
+	public function linkImage() {
+		$this->setVal( 'model', $this->getVal( 'model' ) );
+	}
+
+	public function linkText() {
+		$this->setVal( 'model', $this->getVal( 'model' ) );
+	}
+
+	private function getData() {
+		global $wgCityId, $wgLang;
```
Grunny: These can both be accessed in a controller via `$this->wg->CityId` and `$this->wg->Lang`.

extensions/wikia/DesignSystem/controllers/DesignSystemGlobalFooterController.class.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+class DesignSystemGlobalFooterController extends WikiaController {
+	public function index() {
+		$this->setVal( 'model', $this->getData() );
+	}
+
+	public function section() {
```
Grunny: As implemented, these are all vulnerable to XSS. Most of the values from the API response should be escaped to be safe. Also, if this class doesn't need to be called outside of PHP, extend `WikiaService` instead of `WikiaController` so they can't be called directly via wikia.php.

extensions/wikia/DesignSystem/DesignSystemHelper.class.php
rogatty: Right. At some point this method was more complicated and I wanted to write a test. Now it makes sense to change it back to private.

extensions/wikia/DesignSystem/bower.json
```diff
@@ -3,6 +3,6 @@
   "private": true,
 
   "dependencies": {
-    "design-system": "Wikia/design-system#0.1.1"
+    "design-system": "Wikia/design-system#XW-1575"
```
rogatty: We should change it after https://github.com/Wikia/design-system/pull/25 is merged and released.

extensions/wikia/DesignSystem/bower_components/design-system/.bower.json
```diff
@@ -11,14 +11,13 @@
...
-    "type": "version",
-    "tag": "0.1.1",
-    "commit": "b9de8e29ade426f7db2bacbb32dbcda9731d4c0f"
+    "type": "branch",
+    "branch": "XW-1575",
+    "commit": "0dceecce3aec087d6cc34853e8487f55841f1ab8"
   },
   "_source": "git://github.com/Wikia/design-system.git",
-  "_target": "0.1.1",
+  "_target": "XW-1575",
```
rogatty: This should change after https://github.com/Wikia/app/pull/10999/files#r73507279 is addressed and new version installed.

extensions/wikia/DesignSystem/controllers/DesignSystemGlobalFooterController.class.php
kvas-damian: We'll handle escaping data in XW-1729. We're going to change templates a little bit and add translations there.

extensions/wikia/DesignSystem/controllers/DesignSystemGlobalFooterController.class.php
```diff
@@ -0,0 +1,37 @@
...
+
+class DesignSystemGlobalFooterController extends WikiaService {
+	public function index() {
+		$this->setVal( 'model', $this->getData() );
+	}
+
+	public function section() {
+		$this->setVal( 'model', $this->getVal( 'model' ) );
+		$this->setVal( 'name', $this->getVal( 'name' ) );
+		$this->setVal( 'parentName', $this->getVal( 'parentName' ) );
```
Grunny: These three can be done in one go with `$this->response->setValues`.

extensions/wikia/DesignSystem/controllers/DesignSystemGlobalFooterController.class.php
```diff
@@ -0,0 +1,37 @@
...
+	}
+
+	public function imageHeader() {
+		$this->setVal( 'model', $this->getVal( 'model' ) );
+		$this->setVal( 'section', $this->getVal( 'section' ) );
```
Grunny: Same here.

extensions/wikia/DesignSystem/styles/design-system.scss
```diff
@@ -0,0 +1,6 @@
+// These are temporary overrides that should be a part of DS resets
```
kvas-damian: After work we did in current sprint I think that we decided to keep those styles here.

tomasznapieralski:
Tested

PR #11004, opened 2016-08-04
WW-114 appreciation feature

idradm:


extensions/wikia/ContributionAppreciation/ContributionAppreciation.i18n.php
```diff
@@ -0,0 +1,10 @@
+<?php
+$messages = [];
+
+$messages['en'] = [
+	'appreciation-text' => 'Me liek it!'
```
sebastianmarzjan: Can we use 'I like it' or 'Send the author your appreciation" even for a temporary copy? :)

extensions/wikia/ContributionAppreciation/ContributionAppreciation.i18n.php
```diff
@@ -0,0 +1,10 @@
...
+];
+
+$messages['qqq'] = [
+	'appreciation-text' => 'Appreciation link value.'
```
sebastianmarzjan: Do we need this for a test targeted at EN (and PL?) only?

extensions/wikia/ContributionAppreciation/ContributionAppreciationController.class.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+class ContributionAppreciationController extends WikiaController {
+
+	public function appreciate() {
+		global $wgUser;
+
+		if ( $this->request->wasPosted() && $wgUser->matchEditToken( $this->getVal( 'token' ) ) ) {
```
sebastianmarzjan: Can isValidWriteRequest be used instead of copying the logic?

extensions/wikia/ContributionAppreciation/js/ContributionAppreciation.js
```diff
@@ -0,0 +1,17 @@
...
+	$(function () {
+		console.info('test');
+
+		$('.like').bind('click', function (e) {
+			e.preventDefault();
+			$.post(
+				'/wikia.php?controller=ContributionAppreciation&method=appreciate',
+				$.extend($(this).data(), {token: mw.user.tokens.get('editToken')})
+			).done(function () {
+				console.info('done');
```
sebastianmarzjan: Isn't the confirmation code missing here? Plus, let's use $().log instead of console.*

extensions/wikia/ContributionAppreciation/js/ContributionAppreciation.js
```diff
@@ -0,0 +1,17 @@
...
+
+	$(function () {
+		console.info('test');
+
+		$('.like').bind('click', function (e) {
+			e.preventDefault();
+			$.post(
+				'/wikia.php?controller=ContributionAppreciation&method=appreciate',
+				$.extend($(this).data(), {token: mw.user.tokens.get('editToken')})
+			).done(function () {
```
sebastianmarzjan: What if not done? :)

extensions/wikia/ContributionAppreciation/ContributionAppreciationController.class.php
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	public static function onPageHistoryBeforeList() {
+		Wikia::addAssetsToOutput( 'contribution_appreciation_js' );
+
+		return true;
+	}
+
+	private static function getAppreciationLink( $revision ) {
+		return Html::element( 'a',
```
dianafa: why link not button?

extensions/wikia/ContributionAppreciation/ContributionAppreciationController.class.php
```diff
@@ -0,0 +1,44 @@
...
+class ContributionAppreciationController extends WikiaController {
+
+	public function appreciate() {
+		global $wgUser;
+
+		if ( $this->request->wasPosted() && $wgUser->matchEditToken( $this->getVal( 'token' ) ) ) {
+			//TODO: do something with apprecation
+			$this->response->setFormat( WikiaResponse::FORMAT_JSON );
+			$this->setVal( 'user', $wgUser->getName() );
+			$this->setVal( 'for', Revision::newFromId( $this->request->getInt( 'revision' ) )->getUserText() );
```
dianafa: can we make this val more descriptive than `for`, for example `revisionId` ? `for` can be both receiver of the appreciation and the thing he made so it can be difficult to figure out outside of this context here

extensions/wikia/ContributionAppreciation/js/ContributionAppreciation.js
```diff
@@ -0,0 +1,17 @@
+(function (window, $) {
+	'use strict';
+
+	$(function () {
+		console.info('test');
```
dianafa: Me no liek it

extensions/wikia/ContributionAppreciation/ContributionAppreciationController.class.php
```diff
@@ -0,0 +1,44 @@
...
+
+		return true;
+	}
+
+	private static function getAppreciationLink( $revision ) {
+		return Html::element( 'a',
+			[
+				'class' => 'like',
+				'href' => '#',
+				'data-revision' => $revision
```
dianafa: could we rename it to `data-new-revision`  and add `data-old-revision`?

extensions/wikia/ContributionAppreciation/ContributionAppreciationController.class.php
```diff
@@ -0,0 +1,44 @@
...
+			//TODO: do something with apprecation
+			$this->response->setFormat( WikiaResponse::FORMAT_JSON );
+			$this->setVal( 'user', $wgUser->getName() );
+			$this->setVal( 'for', Revision::newFromId( $this->request->getInt( 'revision' ) )->getUserText() );
+		}
+	}
+
+	public static function onDiffHeader( DifferenceEngine $diffPage, $oldRev, Revision $newRev ) {
+		Wikia::addAssetsToOutput( 'contribution_appreciation_js' );
+		$diffPage->getOutput()->addHTML( self::getAppreciationLink( $newRev->getId() ) );
```
dianafa: Could we make use of the fact that we have available here both old and new revision and pass to link old revision id as well? Its needed for creating a diff page link ie. in email.

extensions/wikia/ContributionAppreciation/ContributionAppreciation.i18n.php
idradm: no

extensions/wikia/ContributionAppreciation/ContributionAppreciationController.class.php
idradm: This is just a placeholder, change it as you wish. Will remove it to avoid confusion.

idradm:
I addressed whatever was possible. Now I sorely require desgins.

extensions/wikia/AssetsManager/config.php
```diff
@@ -3022,3 +3022,18 @@
...
+	'type' => AssetsManager::TYPE_JS,
+	'skin' => [ 'oasis' ],
+	'assets' => [
+		'//extensions/wikia/ContributionAppreciation/scripts/ContributionAppreciation.js'
+	]
+];
+
+$config['contribution_appreciation_scss'] = array(
+	'type' => AssetsManager::TYPE_SCSS,
+	'assets' => array(
```
sebastianmarzjan: Consistency ;-) 
[ ] literal wanted

extensions/wikia/ContributionAppreciation/templates/ContributionAppreciation_diffModule.php
```diff
@@ -0,0 +1,6 @@
+<div class="appreciation-diff-module">
+	<span class="appreciation-cta"><?= wfMessage( 'appreciation-cta-text', $username )->escaped() ?></span>
+	<span class="appreciation-feedback"><?= wfMessage( 'appreciation-feedback-text', $username )->escaped() ?></span>
+	<button class="appreciation-button"
+			data-revision="<?= $revision ?>"><?= wfMessage( 'appreciation-text' )->escaped() ?></button>
```
Grunny: This data-revision attribute in this and the historyModule template is vulnerable to XSS because the controller methods can be called directly (i.e. try requesting `/wikia.php?controller=ContributionAppreciation&method=diffModule&revision="><script>alert(document.domain);</script>` in Firefox).

idradm: I've added params sanitization. Will that do it?

Grunny: It does, though because the username is also in the request data and is also escaped in the `wfMessage` calls above, it will be double escaped (so `Some "username"` would display as `Some &lt;username&gt;` on viewing the page/module). I'd suggest just escaping in the template where it's needed instead, which is generally a best practice as you should always try to escape as close to the output as possible so it's easier to verify a value is escaped when reviewing.

extensions/wikia/ContributionAppreciation/ContributionAppreciationController.class.php
```diff
@@ -0,0 +1,57 @@
+<?php
+
+class ContributionAppreciationController extends WikiaController {
+
+	public function appreciate() {
+		global $wgUser;
+
+		if ( $this->request->isValidWriteRequest( $wgUser ) ) {
```
lukaszkonieczny: `isValidWriteRequest( $wgUser )` does not return any value

extensions/wikia/ContributionAppreciation/scripts/ContributionAppreciation.js
```diff
@@ -0,0 +1,50 @@
+require(['wikia.window', 'jquery', 'wikia.tracker'],
+	function (window, $, tracker) {
```
lukaszkonieczny: `window` is not used here

extensions/wikia/ContributionAppreciation/scripts/ContributionAppreciation.js
```diff
@@ -0,0 +1,50 @@
...
+		};
+
+		$(function () {
+			$('.appreciation-link').bind('click', function (e) {
+				e.preventDefault();
+				track({
+					action: tracker.ACTIONS.CLICK,
+					label: 'appreciation-history-send'
+				});
+				send($(this).data());
```
lukaszkonieczny: You can save `$(this)` in variable

extensions/wikia/ContributionAppreciation/scripts/ContributionAppreciation.js
```diff
@@ -0,0 +1,50 @@
...
+				$(this).parent('.appreciation-history-link').find('.appreciation-feedback').show();
+			});
+
+			$('.appreciation-button').bind('click', function (e) {
+				e.preventDefault();
+				track({
+					action: tracker.ACTIONS.CLICK,
+					label: 'appreciation-diff-send'
+				});
+				send($(this).data());
```
lukaszkonieczny: You can save $(this) in variable

extensions/wikia/ContributionAppreciation/scripts/ContributionAppreciation.js
```diff
@@ -0,0 +1,50 @@
...
+	function (window, $, tracker) {
+		'use strict';
+
+		var track = tracker.buildTrackingFunction({
+			action: tracker.ACTIONS.IMPRESSION,
+			category: 'contribution-appreciation',
+			trackingMethod: 'analytics'
+		});
+
+		var send = function (data) {
```
lukaszkonieczny: You don't need second `var` here. You can separate variables by comma.

includes/actions/HistoryAction.php
```diff
@@ -630,6 +630,8 @@ function historyLine( $row, $next, $notificationtimestamp = false,
 			}
 		}
 
+		wfRunHooks( 'PageHistoryToolsList', array( $this, &$row , &$tools ) );
```
lukaszkonieczny: Please use `[ ]`

extensions/wikia/ContributionAppreciation/ContributionAppreciationController.class.php
```diff
@@ -0,0 +1,57 @@
...
+			Wikia::addAssetsToOutput( 'contribution_appreciation_scss' );
+		}
+
+		return true;
+	}
+
+	private static function shouldDisplayApprectiation() {
+		global $wgUser, $wgLang;
+
+		return $wgUser->isLoggedIn() && $wgLang->getCode() === 'en';
```
sebastianmarzjan, line 110: Please add a comment WHY :)

extensions/wikia/ContributionAppreciation/ContributionAppreciationController.class.php
sebastianmarzjan: Per offline discussion, let's refactor it to assertValidWriteRequest()

sebastianmarzjan:
LGTM

lukaszkonieczny:
LGTM

PR #11005, opened 2016-08-04
RECIRC 188: Fix issue with in-content module causing large whitespace gaps

pauloslund:
[RECIRC-188](https://wikia-inc.atlassian.net/browse/RECIRC-188)
The incontent module clears floats so it's causing all content below it to get pushed down the page. This is problematic when the Table of Contents is really long as it causes large whitespace issues. The fix is to wait for the Table of Contents to load before looking for a suitable section  to render the module in to


nmonterroso:
LGTM

PR #11009, opened 2016-08-04
ADEN-3704 Add rec src param for recoverable ads

drets:
Set `src` slot param to `rec` for recoverable ad


fraszczakszymon:
LGTM :+1:

PR #11012, opened 2016-08-04
MAIN-7675 Disable MW Group JS because of security risk

mklucsarits:
See a description of MediaWiki Group JS (and CSS) here: https://www.mediawiki.org/wiki/Manual:User_group_CSS_and_Javascript

This should be disabled for now as a security risk. The group JS can possibly be re-enabled later if it can be integrated with our wiki JS review.


kerriamber:
LGTM

lizlux:
LGTM

PR #11021, opened 2016-08-08
SOC-1902 Update Discussion Email controllers

jsutterfield:
Ticket: https://wikia-inc.atlassian.net/browse/SOC-1902

This PR makes some small updates to the discussion email controllers in preparation for sending email notifications for the discussion service. The changes mostly involve renaming fields to align with the payload from the notification service.

cc// @mleczey, @slayful

### DIFF
--- a/extensions/wikia/Email/Controller/DiscussionController.class.php
+++ b/extensions/wikia/Email/Controller/DiscussionController.class.php
@@ -52,7 +52,7 @@ abstract class DiscussionController extends EmailController {
             'details' => $this->getDetails(),
             'buttonText' => $this->getButtonText(),
             'buttonLink' => $this->postUrl,
-            'contentFooterMessages' => $this->getFooterMessages(),
+            'contentFooterMessages' => $this->getContentFooterMessages(),
             'hasContentFooterMessages' => true,
         ] );
     }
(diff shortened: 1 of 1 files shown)


extensions/wikia/Email/Controller/DiscussionController.class.php
```diff
@@ -48,20 +52,22 @@ public function body() {
...
+    protected function getDetails() {
+        return $this->postContent;
+    }
 
     private function getButtonText() {
         return $this->getMessage( 'emailext-discussion-button-text' )->text();
     }
 
-    protected function getFooterMessages() {
+    protected function getContentFooterMessages() {
```
jsutterfield, line 70: There are 2 similar named methods used to populate footer messages in emails. The first, `getFooterMessages`, is used to populate the messages at the far bottom of the email. The second, `getContentFooterMessages`, is used to populate the last messages in the main body of the email. I accidentally had those switched, so I renamed this to use the right method name.

extensions/wikia/Email/Controller/DiscussionController.class.php
```diff
@@ -160,8 +154,8 @@ protected static function getEmailSpecificFormFields() {
 
 class DiscussionUpvoteController extends DiscussionController {
 
-    private $firstPostContent;
-    private $upvoteCount;
+    protected $postTitle;
```
slayful: Does the `$postTitle` need to be `protected`?

slayful:
Besides my question related to the `protected` field - LGTM!

extensions/wikia/Email/Controller/DiscussionController.class.php
jsutterfield: Nope, it certainly doesn't. I'l update.

PR #11030, opened 2016-08-08
PLATFORM-2373: fix bad var reference

owend:
@macbre 

WikiFactoryHub::getInstance()->getCategoryByName() returns null or an array with "id" key.


macbre:
:+1:

PR #11032, opened 2016-08-08
SEO-491 Category exhibition view style is not remembered for anyone

gabrys:
Turned out to be a simple typo. The code uses "type" interchangeably with "option". This causes the typos to happen. While playing around with the file, I also added docs for the class members and moved the list of possible display/sort types to class constants.

Renamed all "options" to "types" in the extension.


saipetch:
LGTM

PR #11034, opened 2016-08-09
PLATFORM-2369: increase luastandalone "cpuLimit" to 15 seconds

macbre:
We already increased the memory limit for luastandalone. Let's do the same with the time limit to allow the most complex articles to be rendered successfully.

```
cpuLimit 
  Specify the CPU time limit in seconds for the standalone interpreter on Linux (enforced using ulimit).
```

@Grunny


Grunny:
:+1:

macbre:
Backported to the upcoming release in 466e8d9

PR #11035, opened 2016-08-09
ADEN-3734 Removed liftium-only needed line

nandy-andy:
- last part in `Wikia/app` connected to Liftium,
- JS style fixed.


fraszczakszymon:
Are we going to remove those lines?
https://github.com/Wikia/app/blob/dev/extensions/wikia/AdEngine/js/SlotTracker.js#L40
https://github.com/Wikia/app/blob/dev/extensions/wikia/AdEngine/js/SlotTracker.js#L43
https://github.com/Wikia/app/blob/dev/extensions/wikia/AdEngine/js/AdLogicPageDimensions.js#L36

Add put here empty div with this id instead of calling adstack?
https://github.com/Wikia/app/blob/dev/extensions/wikia/AdEngine/js/AdLogicPageDimensions.js#L36

nandy-andy:
I think the last link is wrong and after looking more for `TOP_BUTTON_WIDE`, I also found this parts:
https://github.com/Wikia/app/blob/dev/extensions/wikia/AdEngine/js/SlotTweaker.js#L108

And I'm pretty sure this must stay. So, let's not rush and discuss it once again, decide what we want to remove and if it's safe ;)

nandy-andy:
I changed the lines and happy-path tested. The tracker change shouldn't be risky as there are no providers using this slot whereas the ad logic page dimension part also shouldn't break TBW since we hide it with [CSS](https://github.com/Wikia/app/blob/f6feaea9fc1dc8b0f924744e043926ea307c6428/skins/oasis/css/core/ads.scss#L101) on narrow screens. As for containers, I've put links to the [Spotlights ticket](https://wikia-inc.atlassian.net/browse/ADEN-2828) and we can address it there. Ready for the 2nd round of CR :)

fraszczakszymon:
Sorry for the wrong link. I meant this one line: https://github.com/Wikia/app/blob/dev/skins/oasis/modules/templates/Ad_Top.php#L10

There is no provider that can handle this slot, so nothing happens - but is it worth to run AdEngine for that?

nandy-andy:
Not really. But also, if we change it to static `div` we'll lose the "page-type" logic and we might introduce a regression in TBW while cleaning up Liftium. That's why I'd rather re-think at grooming and do it within Spotlight ticket since it's more "Spotlights clean-up" than "Liftium clean-up" ;)

fraszczakszymon:
LGTM :gem:

nandy-andy:
Selenium tests passed and I can't see why PHP tests failed - the link doesn't open. Although, the changes are in JS, so I've merged the changes :japanese_ogre:

PR #11036, opened 2016-08-09
SUS-810 log usage with 1% sampling

ghost:
In general, we aim for removing redundant user name and fetching it from a single place. Before we make actual changes, we need to know how frequently is the code executed.

This change is relevant to `filearchive` table.

We start logging the usage with 1% sampling.

@sqreek


includes/api/ApiQueryFilearchive.php
```diff
@@ -144,6 +144,18 @@ public function execute() {
 				$file['timestamp'] = wfTimestamp( TS_ISO_8601, $row->fa_timestamp );
 			}
 			if ( $fld_user ) {
+				/**
+				 * Check, how often is this code executed. Scope: the following if block.
+				 *
+				 * @author Mix
+				 * @see SUS-810
+				 */
+				if ( ( new Wikia\Util\Statistics\BernoulliTrial( 0.01 ) )->shouldSample() ) {
```
sqreek: Can we extract this to one method?
Maybe `PokazJakCzestoTuJestem( sampling , message)`?

PR #11038, opened 2016-08-09
SEO-456 | Fix php integration tests for robots on Special:AllPages

kvas-damian:
//cc @Wikia/x-wing


nandy-andy:
LGTM :+1:

kvas-damian:
BTW ping @gabrys ;)

gabrys:
That's the one we should take the ownership of...

PR #11041, opened 2016-08-09
SEO-456 Link to either local or fandom sitemap from the footer

gabrys:
Depending on whether the local sitemap extension is enabled or not, link
to either the local sitemap page (/wiki/Local_Sitemap) or Fandom's
sitemap (http://fandom.wikia.com/local-sitemap).

### DIFF
--- a/includes/wikia/models/DesignSystemGlobalFooterModel.class.php
+++ b/includes/wikia/models/DesignSystemGlobalFooterModel.class.php
@@ -289,9 +289,7 @@ class DesignSystemGlobalFooterModel extends WikiaModel {
 							'type' => 'translatable-text',
 							'key' => 'global-footer-site-overview-link-local-sitemap'
 						],
-						'href' => $this->getHref(
-							$this->isLocalSitemapAvailable() ? 'local-sitemap' : 'local-sitemap-fandom'
-						)
+						'href' => $this->getLocalSitemapUrl()
 					],
 					[
 						'type' => 'link-text',
@@ -699,9 +697,18 @@ class DesignSystemGlobalFooterModel extends WikiaModel {
 		return $data;
 	}
 
-	private function isLocalSitemapAvailable() {
+	private function getLocalSitemapUrl() {
 		$default = true; // $wgEnableLocalSitemapPageExt = true; in CommonSettings
-		return WikiFactory::getVarValueByName( 'wgEnableLocalSitemapPageExt' , $this->wikiId, false, $default );
+		$localSitemapAvailable = WikiFactory::getVarValueByName(
+			'wgEnableLocalSitemapPageExt', $this->wikiId, false, $default
+		);
+
+		if ( $localSitemapAvailable ) {
+			return $this->getHref( 'local-sitemap' );
+		}
+
+		// Fall back to fandom sitemap when the local one is unavailable
+		return $this->getHref( 'local-sitemap-fandom' );
 	}
 
 	private function getLicenseUrl() {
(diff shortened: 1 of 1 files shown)


saipetch:
LGTM

gabrys:
Hi guys, can I have a LGTM from you:

@tomasznapieralski 
@kvas-damian
@Warkot
@Grunny
@rogatty

@saipetch and I had a discussion about whether there should be the `isLocalSitemapAvailable` method that is used to choose `local-sitemap` or `local-sitemap-fandom` in the `getHref` or the whole thing should be wrapped in one `getLocalSitemapHref` method that checks the condition and returns one of `getHref( 'local-sitemap' )` or `getHref( 'local-sitemap-fandom' )`. Up to you to decide.

includes/wikia/models/DesignSystemGlobalFooterModel.class.php
```diff
@@ -696,6 +699,11 @@ private function getCommunity() {
 		return $data;
 	}
 
+	private function isLocalSitemapAvailable() {
+		$default = true; // $wgEnableLocalSitemapPageExt = true; in CommonSettings
+		return WikiFactory::getVarValueByName( 'wgEnableLocalSitemapPageExt' , $this->wikiId, false, $default );
```
Warkot: `…PageExt' , $this…` – space before comma. 🚀

Warkot:
One super minor comment, but otherwise it should be fine I suppose. LGTM

Maybe someone who took part in creating the proposition of the API structure (https://wikia-inc.atlassian.net/wiki/display/DESYS/RFC%3A+Design+System+API+URL+structure) could have a look? @dianafa @SebastianMarzjan

includes/wikia/models/DesignSystemGlobalFooterModel.class.php
```diff
@@ -696,6 +699,11 @@ private function getCommunity() {
 		return $data;
 	}
 
+	private function isLocalSitemapAvailable() {
+		$default = true; // $wgEnableLocalSitemapPageExt = true; in CommonSettings
```
tomasznapieralski, line 701: can you take this value from `$wgEnableLocalSitemapPageExt` instead of hardcoding it?

Warkot, line 701: It's taken a line below. This is just the default value if none is taken. https://github.com/Wikia/app/blob/dev/extensions/wikia/WikiFactory/WikiFactory.php#L1010

includes/wikia/models/DesignSystemGlobalFooterModel.class.php
```diff
@@ -288,7 +289,9 @@ public function getData() {
 							'type' => 'translatable-text',
 							'key' => 'global-footer-site-overview-link-local-sitemap'
 						],
-						'href' => $this->getHref( 'local-sitemap' )
+						'href' => $this->getHref(
```
tomasznapieralski: please move condition to function. whole `$data` array is build like that.
you can create method `getLocalSitemapLink()` and call it there.

includes/wikia/models/DesignSystemGlobalFooterModel.class.php
gabrys, line 701: The idea is take the value from CommonSettings and then override from WikiFactory for given wiki. The value of $wgEnableLocalSitemapPageExt would be the value for specifically www wiki, not the one from CommonSettings :-( . That's why I needed to hardcode it. We had a discussion about it on #wikia-design-system

kvas-damian:
LGTM

tomasznapieralski:
LGTM

includes/wikia/models/DesignSystemGlobalFooterModel.class.php
tomasznapieralski, line 701: 👍

PR #11043, opened 2016-08-09
SEO-479 Allow caching of <bloglist> tags if there's no pagination

gabrys:
The only reason to disable caching on blog lists was because there's no
easy way to make it work with pagination.

When there's no pagination, we can let the Parser Cache cache the
rendered pages.


saipetch:
LGTM

PR #11047, opened 2016-08-10
WW-116: Style notification banner

lukaszkonieczny:
cc @Wikia/west-wing


extensions/wikia/ContributionAppreciation/templates/ContributionAppreciation_appreciations.php
```diff
@@ -1,7 +1,20 @@
...
+</svg>
+<div class="banner-body">
+	<h4>Nice work!</h4>
+	<?php foreach( $appreciations as $appreciation ): ?>
+		<p class="message"><?= wfMessage( 'appreciation-user' )
+				->rawParams( $appreciation['diffLink'], implode( ', ', $appreciation['userLinks'] ) )
+				->escaped(); ?>
+		</p>
+	<?php endforeach ?>
+	<?php if ( $numberOfAppreciations > 2 ): ?>
```
sebastianmarzjan: This logic is unnecessarily stored in the template - the deduction could be done in controller and you could use $numberOfHiddenAppreciations as both condition and the message param

sebastianmarzjan:
LG

cormorando:
LGTM

PR #11048, opened 2016-08-10
	BatchRefreshLinksForTemplate: refactored

macbre:
[PLATFORM-2375](https://wikia-inc.atlassian.net/browse/PLATFORM-2375)
### The background

Edit on a template caused lots of `RefreshLinksForTitleTask` tasks to be queued and executed in parallel (performed by `LinksUpdate::queueRecursiveJobs`). Each task renders the page (for instance it can make a huge insert statement when a given article has lots of categories) and these tasks are executed in parallel.

[Core MW task](https://github.com/wikimedia/mediawiki/blob/master/includes/jobqueue/jobs/RefreshLinksJob.php) does not run the refresh in a one task per an article fashion (like we do). Instead titles are batched and processed by a lower number of tasks. That's what this PR introduces...
### Changes applied
- do not enqueue tasks for each title, run them one by one
-   `$wgUpdateRowsPerJob`: update 500 articles in a single batch run

@wladekb / @artursitarski / @mixth-sense


artursitarski:
:+1:

Grunny:
LGTM.

PR #11049, opened 2016-08-10
RECIRC 181: Highlight new stories in recirculation

pauloslund:
For the FANDOM_TOPIC tests, we want stories published within the past 24 hours floated to the top. This helps keep the right rail fresh and also makes sure that time sensitive mosts aren't completely overlooked. In this change we also add extra weight to video posts.

### DIFF
--- a/extensions/wikia/Recirculation/js/experiments/placement/fandomTopic.js
+++ b/extensions/wikia/Recirculation/js/experiments/placement/fandomTopic.js
@@ -2,19 +2,49 @@
 define('ext.wikia.recirculation.experiments.placement.FANDOM_TOPIC', [
 	'jquery',
 	'wikia.window',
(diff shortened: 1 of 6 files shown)


extensions/wikia/Recirculation/js/experiments/placement/fandomTopic.js
```diff
@@ -2,19 +2,50 @@
...
+			}).loadData(),
+			recent = FandomHelper({
+				type: 'latest',
+				limit: 1,
+				ignoreError: true
+			}).loadData();
+
+		return $.when(popular, recent)
+			.then(function(popularData, recentData) {
+				var items = utils.ditherResults(popularData.items, 2).slice(0,5),
```
gabrys: var popularItems

gabrys:
LGTM

PR #11052, opened 2016-08-11
ADEN-3775 Move SP listeners

fraszczakszymon:


extensions/wikia/AdEngine/AdEngine2Hooks.class.php
```diff
@@ -98,7 +98,12 @@ public static function onWikiaSkinTopScripts( &$vars, &$scripts ) {
 		$vars['ads'] = [
 			'context' => $adContext,
 			'runtime' => [
-				'sp' => []
```
fraszczakszymon, line 101: Because json_encode did array (`[]`) instead of object (`{}`) 🙈

bognix:
LGTM

extensions/wikia/AdEngine/js/SourcePointDetection.js
```diff
@@ -62,20 +63,22 @@ define('ext.wikia.adEngine.sourcePointDetection', [
...
-			window.ads.runtime.sp.blocking = false;
-			trackStatusOnce('no');
-		});
-
 		log('Appending detection script to head', 'debug', logGroup);
 		node.parentNode.insertBefore(detectionScript, node);
 		detectionInitialized = true;
 	}
 
+	win.ads.runtime.sp = win.ads.runtime.sp || {};
```
drets: What if `runtime` object doesn't exist?

extensions/wikia/AdEngine/js/SourcePointDetection.js
```diff
@@ -62,20 +63,22 @@ define('ext.wikia.adEngine.sourcePointDetection', [
...
+
+	doc.addEventListener('sp.blocking', function () {
+		win.ads.runtime.sp.blocking = true;
+		trackStatusOnce('yes');
+	});
+	doc.addEventListener('sp.not_blocking', function () {
+		win.ads.runtime.sp.blocking = false;
+		trackStatusOnce('no');
+	});
```
drets: I don't like the thing that the code is executed just during module loading.
What do you think about wrapping this code in the method and call the method outside of this module?

extensions/wikia/AdEngine/AdEngine2Hooks.class.php
```diff
@@ -98,7 +98,12 @@ public static function onWikiaSkinTopScripts( &$vars, &$scripts ) {
...
+				'pf' => [
+					'blocking' => null
+				],
+				'sp' => [
+					'blocking' => null
```
drets: So from now we always will have `blocking` variables inside `runtime` object?

fraszczakszymon: Is it an issue?

drets: Minor one!
I believe `null` values are more error prone and in order to check detection in console I should type more chars :wink:

extensions/wikia/AdEngine/AdEngine2ContextService.class.php
```diff
@@ -50,7 +50,7 @@ public function getContext( Title $title, $skinName ) {
 					'showAds' => $adPageTypeService->areAdsShowableOnPage(),
 					'trackSlotState' => $wg->AdDriverTrackState,
 					'sourcePointDetectionUrl' => $sourcePointDetectionUrl,
-					'sourcePointRecovery' => ARecoveryModule::isEnabled(),
+					'sourcePointRecovery' => $skinName === 'oasis' && ARecoveryModule::isEnabled(),
```
nandy-andy, line 53: Worth updating docs we'll enable it only on desktop/oasis.

extensions/wikia/AdEngine/js/context/adContext.js
```diff
@@ -86,10 +86,6 @@ define('ext.wikia.adEngine.adContext', [
 				geo.isProperGeo(instantGlobals.wgAdDriverSourcePointDetectionMobileCountries));
 		}
 
-		// SourcePoint recovery integration (set in AdEngine2ContextService based on wgEnableUsingSourcePointProxyForCSS)
-		if (isUrlParamSet('sourcepointrecovery')) {
```
nandy-andy, line 90: Does it make sense to be consistent and remove `sourcepointdetection` URL param as well? Or we want to keep both parameters and this logic needs to be moved somewhere else, for example to the [init method](https://github.com/Wikia/app/pull/11052/files#diff-986e3632346ab3e613add8f5a1024b46R55)? Please updated docs too if you decide to remove any of those paramters ;)

drets:
LGTM :+1:

drets:
lgtm

PR #11063, opened 2016-08-11
ADEN-3774 Rmoved unexisting file from AM config

nandy-andy:
[While cleaning up Liftium code](https://github.com/Wikia/app/pull/11010) we removed `AdLogicHighValueCountry.js` file which we missed in AssetManager config. It resulted in PHP warning and the change in this PR fixes it.


fraszczakszymon:
LGTM 👍

PR #11068, opened 2016-08-12
ADEN-3775 Back-ported P2 fix from dev

nandy-andy:
Merge pull request #11052 from Wikia/ADEN-3775-fix-sp-detection

ADEN-3775 Move SP listeners


nandy-andy:
WRONG!

PR #11076, opened 2016-08-12
(MAIN-7726) Only load site RL module in VE if site JS is enabled

Grunny:
This adds a JS variable for wgUseSiteJs which is only set and true if
both wgUseSiteJs and wgEnableContentReviewExt are true. We can then use
this in VE to decide whether or not to also load the site JS RL module
when initialising the editor.


owend:
LGTM

PR #11080, opened 2016-08-15
Cake 400 - update swagger generated code and use new entityUniverse query param

nmonterroso:
@Wikia/cake 
https://wikia-inc.atlassian.net/browse/CAKE-400

Regenerate swagger code for the content entity service and change the cake related content module to use the new entityUniverse query param. 

Also updating README to indicate which version of swagger codegen I used, as well as added a cleanup script to remove generated files that we don't really end up using.


ArturKlajnerok:
Are failing tests related, other than that LGTM 👍

nmonterroso:
failing tests are unrelated

PR #11091, opened 2016-08-16
ADEN-3796 AdEngine runner timeout

fraszczakszymon:


extensions/wikia/ARecoveryEngine/js/sourcePointRecovery.js
```diff
@@ -0,0 +1,33 @@
+/*global define*/
+define('ext.wikia.aRecoveryEngine.recovery.sourcePointRecovery', [
```
nandy-andy, line 2: Don't we need in this module implementation of `getPrices()` as well? 

edit:
It's being used in `ext.wikia.adEngine.lookup.lookupFactory` where the `ext.wikia.aRecoveryEngine.recovery.sourcePointRecovery` module isn't being used as lookup module, so probably we don't but I decided it was worth asking ;)

nandy-andy:
Two things:
- merge `dev` branch, or rebase your branch to see if it helps to fail unit tests (they've been failing in `dev` and it's just got fixed a few hours ago,
- `ext.wikia.aRecoveryEngine.recovery.sourcePointRecovery` is being used like lookup module but it's not and it's not totally clear -- feel free to chat with me offline or let's think if it's the right path and if it doesn't need refactoring (if it does then separate ticket should be fine).

extensions/wikia/ARecoveryEngine/js/sourcePointRecovery.js
fraszczakszymon, line 2: Exactly, it's not used as a bidder :)

fraszczakszymon:
@nandy-andy 
ad. second bullet - it's not used exactly like lookup module. It's used only in AdEngineRunner - that's why I renamed all 'bidders' to 'modules'

nandy-andy:
@fraszczakszymon sure, thanks for the clarification! I think I was just biased because of those lines ;)

```
require.optional('ext.wikia.adEngine.lookup.amazonMatch'),
require.optional('ext.wikia.adEngine.lookup.rubiconFastlane'),
require.optional('ext.wikia.aRecoveryEngine.recovery.sourcePointRecovery')
```

and some similar functions in the new module. Maybe then the delaying part of lookups should be separated to make it all consistent and clean.

Well, that's a longer discussion not for her, I guess...

LGTM :japanese_ogre:

PR #11096, opened 2016-08-16
RECIRC-181: Properly use $type in FandomDataService

pauloslund:
`$type` is required to determine which endpoint to hit and needs to be passed into the apiRequest


katanka:
LGTM

PR #11099, opened 2016-08-17
JPN-628 Implement toggling upvote

perjg:
Allow clicking on the upvote icon several times to upvote / cancel the upvote.

@Grunny @d34th4ck3r @Wikia/spitfires


extensions/wikia/EmbeddableDiscussions/models/DiscussionsThreadModel.class.php
```diff
@@ -43,7 +44,18 @@ private function getUpvoteRequestUrl( $id ) {
 	}
 
 	private function apiRequest( $url ) {
-		$data = Http::get( $url );
+		global $wgUser;
+
+		$request = $wgUser->getRequest();
```
Grunny: If you're just getting the request object, `RequestContext::getMain()->getRequest();` would be better than getting it from the global user object.

Grunny:
Other than the comment above, LGTM.

perjg:
@Grunny I've rewritten to use RequestContext::getMain()->getRequest(); like suggested.

PR #11114, opened 2016-08-17
WW-171 Support blocks in Contribution Appreciation experiment

sebastianmarzjan:
- Check user block status before displaying call to action
- Ensure request to appreciation is made by a non-blocked user


rybmat:
LGTM

PR #11116, opened 2016-08-17
Cake 431 - add tracking on first keyup for source/ck editors

nmonterroso:
@Wikia/cake
https://wikia-inc.atlassian.net/browse/CAKE-431

Adding some tracking for the first keyup event when editing in source mode or wysiwyg mode


resources/wikia/Resources.php
```diff
@@ -293,4 +293,13 @@
 			'resources/wikia/modules/importScriptHelper.js'
 		]
 	],
+
+	'wikia.article.edit' => [
+		'scripts' => [
+			'resources/wikia/modules/article.edit.js',
```
ArturKlajnerok: maybe `article.edit.js` -> `articleEditTracking.js`would be more descriptive ?

ArturKlajnerok:
One small comment for naming but overall LGTM 👍

resources/wikia/Resources.php
nmonterroso: 97cf445

PR #11121, opened 2016-08-18
WW-172 Disable email component of Appreciation feature

dianafa:
WW-172 Disable email component of Appreciation feature

https://wikia-inc.atlassian.net/browse/WW-172


rybmat:
LGTM

PR #11122, opened 2016-08-18
fix annotation from Database to DatabaseBase

sqreek:
eye pain
@macbre


macbre:
`@return LooksGoodToMe`

PR #11123, opened 2016-08-18
In AntiSpoof connect to UserRename::AfterGlobal hook

sqreek:
https://wikia-inc.atlassian.net/browse/SUS-882

@mixth-sense

### DIFF
--- a/extensions/AntiSpoof/AntiSpoofHooks.php
+++ b/extensions/AntiSpoof/AntiSpoofHooks.php
@@ -123,4 +123,21 @@ class AntiSpoofHooks {
 		$spoof->update( $oldName );
 		return true;
 	}
+
+	/**
+	 * Wikia Addition
+	 * (After a successful user rename using Wikia Tool)
+	 *
+	 * @param $dbw DatabaseBase
+	 * @param $uid
+	 * @param $oldusername string
+	 * @param $newusername string
+	 * @param $process
+	 * @param $tasks
+	 */
+	public static function asAfterWikiaRenameUserHook( $dbw, $uid, $oldusername, $newusername, $process, &$tasks ) {
+		$spoof = self::makeSpoofUser( $newusername );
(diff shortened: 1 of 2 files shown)


Grunny:
Does this still block spoofing the old username in addition to the new name? The comment on the method says "On rename, remove the old entry and add the new", so it suggests it does not.

Unlike the MW core rename extension, our user rename tool creates the old user name so it can't be taken, and the old names probably still shouldn't be spoofable (especially for users who were well known under their previous username). The MW core rename tool doesn't create the old name, as renaming frees the old name for reuse by someone else, so it wasn't an issue for them.

sqreek:
That is true, it's for renaming, so old names will be available after this change.
Let me fix that.

Kirkburn:
Though you're already tweaking it: I agree that old usernames should remain covered by antispoof. We wouldn't want a newly renamed user to potentially have their old name start being impersonated around the site.

extensions/AntiSpoof/AntiSpoofHooks.php
```diff
@@ -123,4 +123,22 @@ public static function asAddRenameUserHook( $uid, $oldName, $newName ) {
...
+	 * (After a successful user rename using Wikia Tool)
+	 *
+	 * @param $dbw DatabaseBase
+	 * @param $uid
+	 * @param $oldusername string
+	 * @param $newusername string
+	 * @param $process
+	 * @param $tasks
+	 */
+
```
Grunny: Minor: remove this blank line between the comment and the method.

Grunny:
LGTM.

extensions/AntiSpoof/AntiSpoofHooks.php
sqreek: removed, and rebased

sqreek:
Closing in favor of #11131

PR #11163, opened 2016-08-23
SUS-387 Make Ooyala using HTML player

nandy-andy:
Since March 2016 we've been using HTML player on adtest.wikia.com, since 22nd Aug 2016 on 4 more wikis. No reports of failures - let's make the HTML player preferable site-wide! :japanese_ogre:


Grunny:
LGTM.

PR #11225, opened 2016-08-30
(XW-1906) Make sure when building for Oasis we use the correct ZH codes

Grunny:
Oasis uses the zh lang code for simplified Chinese by falling back to zh-hans,
while the DS and Mercury use zh for traditional Chinese. This makes sure we use
the correct lang code for Oasis so the expected language fallbacks take effect.

/cc @Wikia/x-wing


extensions/wikia/DesignSystem/i18n/build.js
```diff
@@ -37,6 +37,12 @@ var languages = fs.readdirSync(rootDir).filter(function (file) {
 languages.forEach(function (lang) {
 	var i18n = require(rootDir + '/' + lang + '/' + filename);
 
+	// Oasis uses zh for simplified Chinese, while the DS uses zh for traditional
+	// so we need to correct the lang code here so the Oasis fallbacks take effect
+	if (lang === 'zh') {
+		lang = 'zh-hant';
```
rogatty, line 43: Is it `zh-hant` or `zh-hans`?

Grunny: `zh-hant`. `zh-hans` is simplified Chinese, which is the code used in DS. `zh` in DS is traditional Chinese or `zh-hant` for Oasis.

In Oasis, `zh` should fallback to `zh-hans`, but it had it's own definition so was using traditional Chinese. `zh-tw` fallsback to `zh-hant` then to `zh-hans`, but since `zh-hant` was undefined it was showing `zh-hans` or simplified Chinese. If that all makes sense. :)

rogatty:
LGTM

kvas-damian:
LGTM

hakubo:
LGTM

PR #11249, opened 2016-09-01
SUS-969: Fix diff pages on wikis with a dark theme (release)

mszabo-wikia:
ticket: https://wikia-inc.atlassian.net/browse/SUS-969


sqreek:
LGTM, 
but can we merged all darkwiki if's into one?
Maybe as a separate PR?

mszabo-wikia:
Yep, that sounds good, I'll ticket that. :smile:

PR #11259, opened 2016-09-02
Username Service Implementation

sqreek:
SUS want to introduce one source for username, instead keep it in every table.

This implementation give us possibility to fast change between implementation and make experiment about performance

@mixth-sense


macbre:
I'd rather use factory pattern approach. Using five files to call a single static method from PHP class looks a bit overzealous

macbre:
Use factory! :)

PR #11137, opened 2016-08-19
SUS-894 Remove real maps creation

nandy-andy:
Disabled possibility to create real-world maps in MW client of WikiaMaps.

https://wikia-inc.atlassian.net/browse/SUS-894

PR #11272, opened 2016-09-06
SUS-976 Fix custom maps creation

drets:
One required field was accidentally removed in https://github.com/Wikia/app/pull/11137 (adding it back).


ukaka:
Passed QA. LGTM.

PR #11306, opened 2016-09-12
[SUS-991] Downloaded translations from crowdin

jcellary:
https://wikia-inc.atlassian.net/browse/SUS-991

Uploaded local translations to crowdin and then downloaded them back to put everything in sync. Before local translations were applied directly to php files and not translated in crowdin, hence when doing a "download" local translated strings (i.e. in German) were reset back to english.

@mixth-sense @macbre


ghost:
Yeah, that's one of the very nice feature of Crowdin :-(

PR #11343, opened 2016-09-15
ADEN-3943 Add TLB to Vulcan calls

fraszczakszymon:
Configure TOP_LEADERBOARD in Vulcan slot map

### DIFF
--- a/extensions/wikia/AdEngine/js/lookup/rubicon/rubiconVulcan.js
+++ b/extensions/wikia/AdEngine/js/lookup/rubicon/rubiconVulcan.js
@@ -20,6 +20,15 @@ define('ext.wikia.adEngine.lookup.rubicon.rubiconVulcan', [
 						loc: 'hivi'
 					},
 					zoneId: 260296
+				},
+				TOP_LEADERBOARD: {
+					siteId: 55412,
+					size: [640, 480],
+					sizeId: 203,
+					targeting: {
+						loc: 'top'
+					},
+					zoneId: 519058
 				}
 			},
 			mercury: {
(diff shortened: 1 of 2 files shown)


extensions/wikia/AdEngine/js/spec/lookup/rubicon/rubiconVulcan.spec.js
```diff
@@ -221,6 +221,6 @@ describe('ext.wikia.adEngine.lookup.rubicon.rubiconVulcan', function () {
 
 		vulcan.call();
 
-		expect(vulcan.getSlotParams('TOP_LEADERBOARD')).toEqual({});
+		expect(vulcan.getSlotParams('TOP_RIGHT_BOXAD')).toEqual({});
```
nandy-andy, line 224: :trollface: what if we add medrec to Vulcan calls in future? I'm kidding. I think it's fine. Once we add (if we add) there medrec the one working on it will have to update unit tests which is an interesting practice :japanese_ogre:

extensions/wikia/AdEngine/js/lookup/rubicon/rubiconVulcan.js
```diff
@@ -20,6 +20,15 @@ define('ext.wikia.adEngine.lookup.rubicon.rubiconVulcan', [
...
 					zoneId: 260296
+				},
+				TOP_LEADERBOARD: {
+					siteId: 55412,
+					size: [640, 480],
+					sizeId: 203,
+					targeting: {
+						loc: 'top'
+					},
+					zoneId: 519058
```
nandy-andy, line 31: Where do we get all those numbers from?

nandy-andy (approved):
LGTM, I left few comments but they're not blocking the work. Please, just answer them either here or offline :+1:

extensions/wikia/AdEngine/js/lookup/rubicon/rubiconVulcan.js
fraszczakszymon: From the ticket, it's site and zone created in Rubicon dashboard

bulaszenko:
Checked works good :)

PR #11361, opened 2016-09-16
Adjust distance between admin avatars

lukaszkonieczny:
Also `ul` replaced with `div` because there are some wikis which have `margin-left: Zpx !important` set for `ul` element in their custom css.

@Wikia/west-wing


sebastianmarzjan:
LGTM

rybmat:
LGTM

PR #11381, opened 2016-09-19
XW-2003 | update icons in Global Navigation

Warkot:
https://wikia-inc.atlassian.net/browse/XW-2003
https://github.com/Wikia/design-system/pull/65
https://github.com/Wikia/mercury/pull/2798

@Wikia/x-wing


rogatty (approved):

kvas-damian:
I don't like the direction of this PR. In ideal world we should have all icons in the same size and resize them using CSS. Introducing new icon in different size continues trend of not being consistent with what we want.

hakubo:
@kvas-damian I agree and that is why we're meeting with Derek on Wednesday to talk about it.

hakubo (approved):

kvas-damian:
@hakubo Does it make sense to proceed with this PR when we plan to change the direction?

PR #11422, opened 2016-09-22
WW-336 capitalize right rail

cormorando:


sebastianmarzjan:
LG, do you want to change it for other languages too?

idradm:
LG, be careful when uploading to Crowdin

PR #11426, opened 2016-09-23
MAIN-8019 upgrade antispoof maintanance script

sqreek:
https://wikia-inc.atlassian.net/browse/MAIN-8019

use batch instead loading all username at once

@wladekb @macbre

### DIFF
--- a/extensions/AntiSpoof/maintenance/batchAntiSpoof.php
+++ b/extensions/AntiSpoof/maintenance/batchAntiSpoof.php
@@ -25,26 +25,23 @@ class BatchAntiSpoof extends Maintenance {
 		$dbw->bufferResults( false );
 
 		$batchSize = 1000;
-
-		$this->output( "Creating username spoofs...\n" );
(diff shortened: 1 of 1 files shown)


extensions/AntiSpoof/maintenance/batchAntiSpoof.php
```diff
@@ -25,26 +25,23 @@ public function execute() {
...
+				$n++;
 			}
-		}
-
-		$this->batchRecord( $items );
-		$this->output( "$n user(s) done.\n" );
+			$this->batchRecord( $items );
+			$this->output( "$n user(s) done. Current user_id $current\n" );
+			$current += $batchSize;
+		} while ( $result->numRows() > 0 );
```
macbre, line 44: User IDs are not continuous unfortunately. Let's get `MAX(user_id)` and use it here instead.

sqreek:
It seems original maintenance script works does not have problems with memory.

PR #11427, opened 2016-09-23
ADEN-3977 Expose Playwire Bolt API

fraszczakszymon:
Changes:
- possibility to use Bolt API from injected player (inside creative code)


nandy-andy:
LGTM

PR #11432, opened 2016-09-23
JPN-443 Rendering fix for Firefox

perjg:
height: 100% on this element causes trouble in Firefox.

@d34th4ck3r @Grunny @garthwebb


bkoval (approved):

perjg:
Thanks @bkoval !

PR #11477, opened 2016-09-27
RECIRC-213: Remove scroll depth test

pauloslund:
We are no longer running the test so I'm removing the code for it


gabrys (approved):
LGTM

PR #11498, opened 2016-09-29
XW-2133 | Fix anchor links

hakubo:
I used :before here to make target element higher
to accommodate for global navigation

Nice benefit of this is that page won't jump when clicking on TOC

@Wikia/x-wing What do you think?


Warkot:
Are we sure we're not overwriting other `:before` elements in the clients? Because it this PR looks like a Holy Grail. 🍷

Warkot (approved):

kvas-damian:
@Warkot most common use-case for anchor links are section headers and article comments. I believe it's enough to cover them during QA.

kvas-damian (approved):

tomasznapieralski:
We cannot use this solution, because:
- it doesn't work for inline targets (setting `display: block` on `span` target fixes this issue, but it is risky and non-semantic solution)
- it doesn't work on comments when they load after page loads

Fixed in https://github.com/Wikia/app/pull/11529

PR #11437, opened 2016-09-23
RECIRC 200: Replace placement test with new experiment

pauloslund:
## Description

The placement test has run it's course, and is being replaced with a new test! The main goal of the new test is to be able to setup the groups using the AbTesting tool. What this means is that we configure what data we want in what location using the UI. This allows us to iterate on tests quicker, as well as gives us an easier way to mix data between the placements.
## Links
- https://wikia-inc.atlassian.net/browse/RECIRC-200
- https://wikia-inc.atlassian.net/wiki/display/RECIRC/Mixed+placement+test

PR #11502, opened 2016-09-29
refactor proposition

sqreek:
After check https://wikia-inc.atlassian.net/browse/SONAR-16 I spend few minutes to refactor code in question.

### DIFF
--- a/extensions/wikia/Recirculation/services/CakeRelatedContentService.php
+++ b/extensions/wikia/Recirculation/services/CakeRelatedContentService.php
@@ -10,6 +10,10 @@ use Wikia\Service\Swagger\ApiProvider;
 
 class CakeRelatedContentService {
 
+	const FAILED_TO_RETRIEVE_RECOMMENDATIONS = "getRelatedContentFromEntityName failed to retrieve recommendations";
+	const ARTICLES_ARE_NOT_ARRAY = "getRelatedContentFromEntityName expected fandom_articles to be an array";
+	const THREADS_ARE_NOT_ARRAY = "getRelatedContentFromEntityName expected discussion_threads to be an array";
+	const WIKI_ARTICLES_ARE_NOT_ARRAY = "getRelatedContentFromEntityName expected wiki_articles to be an array";
 	use Loggable;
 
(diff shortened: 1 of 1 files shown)


sqreek:
@gabrys

extensions/wikia/Recirculation/services/CakeRelatedContentService.php
```diff
@@ -21,105 +25,80 @@ class CakeRelatedContentService {
...
 
-			return $items;
-		} catch (ApiException $e) {
-			$this->error("error while getting content", [
-				'code' => $e->getCode(),
-				'message' => $e->getMessage(),
-			]);
-			return [];
+		$items = [];
+		$fandomArticles = array_slice( $filteredRelatedContent->getFandomArticles(), 0, $limit - count( $items ), true );
```
sqreek, line 79: Forget about negative length here, but I think it's still better concept than 
`array_map(null, ...)` and `break 2`

gabrys:
Ping @nmonterroso. I wasn't the author of this. I don't have enough context to tell either :+1: or :-1:

nmonterroso:
ping @Wikia/cake, thanks @gabrys for looping us in!

I appreciate trying to clean up the code, but from what I can tell this changes the output to be all the fandom content, then all the discussion content, then all the wiki articles content. But the intention is to be one fandom article, one discussion article, one wiki article, etc until we reach the limit. So this is fundamentally changing what the code does, so it shouldn't be merged.

This code was also recently refactored as part of [RECIRC-200](https://github.com/Wikia/app/pull/11437/files#diff-fc87b012a152561abdce9bd9d87bccdb), so there are some conflicts that would need to be resolved.

As far as the sonar check is concerned, I agree with @gabrys comment on the JIRA:

> I could agree wrapping a whole method in a try/catch is a bad practice. Now, is having 5 return statements is a bad practice? I think it's about what you return and where than how many statements. If we need a number, because that's how SonarQube works, set it to 10 or something like that. That is most probably bad.

As long as the code is clear, I think early returns are a good way to reduce nesting in the current scope.

PR #11500, opened 2016-09-29
ADEN-4053 Fix destroy slots method

fraszczakszymon:

PR #11519, opened 2016-09-30
ADEN-4053 Back-ported P2 fix from dev

nandy-andy:
Merge pull request #11500 from Wikia/ADEN-4053

ADEN-4053 Fix destroy slots method


fraszczakszymon (approved):
LGTM 👍

PR #11531, opened 2016-09-30
Hot fix for SEO-558

gabrys:
The correct translation is already defined in GlobalMessagesW.i18n.php.

I don't know what's the purpose of this file...


macbre:
@gabrys  shouldn't we remove translations for all languages?

macbre:
See #11540

PR #11535, opened 2016-09-30
SOC-3130 Update Wikia to Fandom for the email sender

garthwebb:
https://wikia-inc.atlassian.net/browse/SOC-3130


jsutterfield:
LGTM!

PR #11561, opened 2016-10-04
WW-342: Add flow tracking after VE is loaded

lukaszkonieczny:
@Wikia/west-wing

### DIFF
--- a/extensions/wikia/FlowTracking/scripts/createNewPageFlowsVE.js
+++ b/extensions/wikia/FlowTracking/scripts/createNewPageFlowsVE.js
@@ -1,0 +1,9 @@
+require(['wikia.flowTracking.createPage', 'mw', 'jquery'], function (flowTrackingCreatePage, mw, $) {
+	function init() {
+		mw.hook('ve.activationComplete').add(function () {
+			flowTrackingCreatePage.trackOnEditPageLoad('ve');
+		});
+	}
+
+	$(init);
+});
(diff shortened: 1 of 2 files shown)


extensions/wikia/FlowTracking/scripts/createNewPageFlowsVE.js
```diff
@@ -0,0 +1,9 @@
+require(['wikia.flowTracking.createPage', 'mw', 'jquery'], function (flowTrackingCreatePage, mw, $) {
+	function init() {
+		mw.hook('ve.activationComplete').add(function () {
+			flowTrackingCreatePage.trackOnEditPageLoad('ve');
```
rybmat: As we are using 'sourceedit', 'rte-source' and 'rte-wysiwyg' I'm wondering if we should use 've' or maybe 'visualeditor'. What do you think?

rybmat:
one minor suggestion, rest LGTM

extensions/wikia/FlowTracking/scripts/createNewPageFlowsVE.js
lukaszkonieczny: I don't have strong opinion as long as we are able to distinguish editors

rybmat (approved):

extensions/wikia/FlowTracking/scripts/createNewPageFlowsVE.js
sebastianmarzjan: "visualeditor" is used in current tagging, so I'd use it

sebastianmarzjan:
Please adjust the string literal and then GTG

sebastianmarzjan (approved):

PR #11607, opened 2016-10-11
Toggle 2 more dis vars during CNW

jsutterfield:
Cristina has requested that we toggle 2 more Discussions related wg vars during CNW, `$wgEnableDiscussionPostsCountInUserIdentityBox` and `$wgEnableRecirculationDiscussions`.

`$wgEnableDiscussionPostsCountInUserIdentityBox` should go away following SOC-3082, but we're going to enable it for now until that ticket hits production.

cc @sactage


kerriamber (approved):
LGTM

PR #11609, opened 2016-10-11
RECIRC-219: Video specific analytics

pauloslund:
We want better analytics around when a user clicks to see a video specific recirculation item


gabrys (approved):
LGTM

PR #11622, opened 2016-10-12
RECIRC-MIX: Graceful Fail

pauloslund:
We want to render a default recirculation module if for whatever reason something doesn't work in the right rail.


gabrys (approved):
LGTM

PR #11683, opened 2016-10-21
WW-443 flow tracking improvements

idradm:


rybmat (approved):

lukaszkonieczny:
What do you think about adding additional tracking for example:
- in backend when edit page is loaded
- in front end to check if editor was loaded

extensions/wikia/FlowTracking/scripts/initCreatePageTracking.js
```diff
@@ -22,6 +22,7 @@ require([
 			var qs = QueryString(redlink.href);
 
 			qs.setVal('flow', redLinkFlow);
+			qs.removeVal('tracked');
```
lukaszkonieczny: I think redlinks shouldn't contain `tracked` param in their `href`

lukaszkonieczny (approved):

extensions/wikia/FlowTracking/scripts/initCreatePageTracking.js
```diff
@@ -77,9 +82,9 @@ require([
 			}
 		});
 
-		mw.hook('ve.afterVEInit').add(function(veEditUri) {
-			if (!!mw.config.get( 'wgArticleId' )) {
-				veEditUri.extend( { flow: createButtonFlow } );
+		mw.hook('ve.afterVEInit').add(function (veEditUri) {
+			if (!mw.config.get('wgArticleId')) {
```
cormorando, line 86: to jest castowanie na boola, powinno zostac `!!`

extensions/wikia/FlowTracking/scripts/initCreatePageTracking.js
```diff
@@ -49,12 +49,17 @@ require([
 			flowTracking.beginFlow(inputBoxFlow, {});
 		});
 
-		$( '#ca-ve-edit,  #ca-edit' ).click(function() {
+		$('#ca-edit').on('mousedown', function () {
```
cormorando, line 52: Czemu na VE `click` a tu `mousedown`?

lukaszkonieczny: In VE `click` may stay because page isn't reloaded. Classic editor redirects you to new page, so we should send tracking earlier, before page is reloaded.

extensions/wikia/FlowTracking/scripts/initCreatePageTracking.js
lukaszkonieczny: If we use `!!` this condition is true when article exists and we want to add param only if article doesn't exist

cormorando (approved):

PR #11695, opened 2016-10-24
WW-443: Fixes for flow tracking create button begin step

lukaszkonieczny:


lukaszkonieczny:
Discussed with @idradm offline

PR #11708, opened 2016-10-25
WW-457: Disable redirect for js pages

lukaszkonieczny:
https://wikia-inc.atlassian.net/browse/WW-457

@Wikia/west-wing @Grunny

PR #11716, opened 2016-10-26
WW-457: Disable redirect for js pages

idradm:


idradm:
https://github.com/Wikia/app/pull/11708

PR #11724, opened 2016-10-26
Fix paths to design system styles

lukaszkonieczny:
@Wikia/x-wing


weronika (changes requested):
all imported partial files should be provided without leading _

weronika (approved):

PR #11725, opened 2016-10-26
ADEN-4166 | temporary fix for not supported size by indexExchange

bognix:
This fix is super temporary - here is a link to PR with proper fix, we need to wait for Prebid guys to merge it - https://github.com/prebid/Prebid.js/pull/745


fraszczakszymon:
LGTM 🙈

PR #11755, opened 2016-11-02
XW-1896 | global footer has fixed width on small breakpoint

weronika:
https://wikia-inc.atlassian.net/browse/XW-1896

### DIFF
--- a/skins/oasis/css/core/breakpoints-layout.scss
+++ b/skins/oasis/css/core/breakpoints-layout.scss
@@ -5,17 +5,9 @@
 @if $widthType == 0 {
 	// consider using https://github.com/stowball/mqGenie for release
 	@media #{$breakpoint-small-only} {
-		$content-width: $page-width-small - 2 * $content-border-width;
-
 		.WikiaPage {
 			padding: 0 $padding-width-small;
-			width: $content-width;
-		}
-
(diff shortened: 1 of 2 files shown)


skins/oasis/css/core/breakpoints-layout.scss
```diff
@@ -5,9 +5,17 @@
...
 	@media #{$breakpoint-small-only} {
+		$content-width: $page-width-small - 2 * $content-border-width;
+
 		.WikiaPage {
 			padding: 0 $padding-width-small;
-			width: $page-width-small - 2 * $content-border-width;
+			width: $content-width;
+		}
+
+		.wds-global-footer {
```
rogatty: Shouldn't this be in https://github.com/Wikia/app/blob/dev/extensions/wikia/DesignSystem/styles/design-system.scss?

weronika: I expected this one ;) the reason why I decided to put it here is:
1. we adjust footer appearance to content appearance, so now we have this definitions together
2. as you can see there are 4 variables involved - I didn't see any point in importing 2 files from totally different place to design-system.scss

rogatty (approved):

Warkot (approved):

PR #11760, opened 2016-11-02
XW-2209 | Update to new DS version

kvas-damian:
https://wikia-inc.atlassian.net/browse/XW-2209

Ping @Wikia/x-wing


kvas-damian:
LGTM ;)

PR #11778, opened 2016-11-04
MAIN-8472 Make namespaces aliases mergeable

gabrys:
Fixed in MediaWiki 1.28:
https://github.com/wikimedia/mediawiki/commit/ea9fcc1e4d3b572199d82c426024e3e5efe23879

The problem is:

If you have some namespace aliases defined in "core" and in extensions
MediaWiki (prior to 1.28) will just overwrite the "core" settings with
the extension ones.

Fixed in 1.28 to let both "core" and extensions define the aliases and
handle that by merging the arrays instead of overwriting.


Grunny (approved):
LGTM.

PR #11793, opened 2016-11-04
SUS-1234: Fix PHP fatal in SMW BrowseByProperty API

mszabo-wikia:
ticket: https://wikia-inc.atlassian.net/browse/SUS-1234


sqreek:
This is not koszerne @TK-999.
Semantic Media Wiki use composer and should be loaded from https://github.com/Wikia/SemanticMediaWiki/

PR #11798, opened 2016-11-07
WW-363 typography and spacing changes

idradm:


extensions/wikia/CreateNewWiki/css/CreateNewWiki.scss
```diff
@@ -31,28 +32,40 @@
...
-	font-weight: normal;
+	color: $wds-color-dark-gray;
+	font-size: $wds-typescale-size-base;
+	line-height: $wds-line-height;
 	margin: 4px auto 0;
 	padding: $content-spacing-regular 0 $cnw-bottom-bar-size;
 	position: relative;
 	width: $cnw-size;
 
+	::-webkit-input-placeholder, :-moz-placeholder, ::-moz-placeholder, :-ms-input-placeholder {
```
lukaszkonieczny: You can use placeholder mixin (https://github.com/Wikia/app/blob/dev/skins/shared/mixins/placeholder.scss)

extensions/wikia/CreateNewWiki/css/CreateNewWiki.scss
```diff
@@ -119,24 +119,101 @@
...
+		display: none;
+
+		label {
+			display: inline-block;
+			position: relative;
+			width: 100%;
+		}
+
+		input {
+			position: absolute;
```
lukaszkonieczny, line 164: Why we are using `position: absolute` here?

rybmat (approved):
changes looks good so I'm approving it. However In both designs and this PR there is nothing about styling error messages

extensions/wikia/CreateNewWiki/css/CreateNewWiki.scss
```diff
@@ -119,24 +119,101 @@
...
+		}
+
+		span {
+			display: block;
+			margin-left: 1.75em;
```
lukaszkonieczny, line 169: Why you are using `em` units?

extensions/wikia/CreateNewWiki/templates/CreateNewWiki_Index.php
```diff
@@ -91,30 +82,21 @@
...
-						[
-							'text' => wfMessage( 'cnw-desc-all-ages' )->escaped(),
-							'tooltipIconTitle' => wfMessage( 'cnw-desc-tip-all-ages' )->plain(),
-						]
-					 );
-					?>
+				<div class="desc-block">
+					<textarea id="Description" placeholder="<?= wfMessage( 'cnw-desc-placeholder' )->escaped() ?>"></textarea>
+					<div class="checkbox" id="all-ages-div"
+						<?= (empty( $selectedLang ) || $selectedLang === $params['LangAllAgesOpt']) ? '' : 'style="display: none"' ?>>
```
lukaszkonieczny, line 88: Can we use some class instead of adding inline style?

extensions/wikia/CreateNewWiki/css/CreateNewWiki.scss
```diff
@@ -119,24 +119,101 @@
...
 
+	.language-choice {
+		margin: $content-spacing-xxlarge 0;
+	}
+
+	.select-container {
+		margin: $content-spacing-xxlarge 0;
+	}
+
+	.wiki-domain-container {
```
lukaszkonieczny: What do you think about creating one rule for these 3 classes?

extensions/wikia/CreateNewWiki/templates/CreateNewWiki_Index.php
```diff
@@ -10,37 +10,28 @@
...
 
 			<form name="label-wiki-form">
-				<h3><?= wfMessage( 'cnw-name-wiki-label' )->escaped() ?></h3>
-				<span class="wiki-name-status-icon status-icon"></span>
 				<?=
 					wfMessage('autocreatewiki-title-template')->rawParams(
 						Xml::element('input', [
 							'type' => 'text',
 							'name' => 'wiki-name',
+							'placeholder' => wfMessage( 'cnw-name-wiki-label' )->escaped(),
```
lukaszkonieczny: I think attributes are already encoded inside `Xml:element` method

lukaszkonieczny (changes requested):

extensions/wikia/CreateNewWiki/css/CreateNewWiki.scss
rybmat: I think this out of this ticket's scope. Should be handled in WW-367

extensions/wikia/CreateNewWiki/templates/CreateNewWiki_Index.php
rybmat: should be handled in WW-367 as this is about checkbox. Highlighted because of wrapping in the div.desc-block

extensions/wikia/CreateNewWiki/css/CreateNewWiki.scss
rybmat: this is about input fields, not typography, WW-364 is about that

lukaszkonieczny (approved):

PR #11820, opened 2016-11-08
WW-486 | fix input errors positioning

Warkot:
https://wikia-inc.atlassian.net/browse/WW-486

### DIFF
--- a/extensions/wikia/CreateNewWiki/css/CreateNewWiki.scss
+++ b/extensions/wikia/CreateNewWiki/css/CreateNewWiki.scss
@@ -124,12 +124,13 @@
 		}
 	}
 
-	.wiki-domain-container, .wiki-name-container {
+	.wiki-domain-container,
+	.wiki-name-container {
 		position: relative;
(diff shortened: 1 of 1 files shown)


extensions/wikia/CreateNewWiki/css/CreateNewWiki.scss
```diff
@@ -150,8 +150,9 @@
 
 	.error-msg {
 		color: $wds-color-alert;
-		float: left;
 		font-size: $wds-typescale-size-minus-2;
+		position: absolute;
```
lukaszkonieczny: I'm curious if removing just `float: left` is not enough ;)

Warkot: It's possible, but the whole content beneath the error would be moving down when the error is displayed. For the same reason it was using float.

rybmat (approved):

lukaszkonieczny (approved):

PR #11845, opened 2016-11-11
(PLATFORM-2554) Remove NewWindowLinks extension

Grunny:
https://wikia-inc.atlassian.net/browse/PLATFORM-2554


Grunny:
Note: we need to update the relevant projects on crowdin once this passes CR, but waiting to avoid causing conflicts.

pgrabowski-wikia (approved):
👍

PR #11853, opened 2016-11-14
SUS-1287: add "staff" class to article comments made by staff user

mszabo-wikia:
ticket & QA notes: https://wikia-inc.atlassian.net/browse/SUS-1287


extensions/wikia/ArticleComments/modules/templates/ArticleComments_Comment.php
```diff
@@ -1,5 +1,5 @@
 <? if ( is_array( $comment ) ) :?>
-<li id="comm-<?=$commentId?>" class="comment SpeechBubble <?=$rowClass?>" data-user="<?=$comment['username']?>">
+<li id="comm-<?=$commentId?>" class="comment SpeechBubble <?=$rowClass?><?= $comment['isStaff'] ? 'staff' : '' ?>" data-user="<?=$comment['username']?>">
```
macbre: There should be a space before `"staff"`, otherwise it will be merged with a value of `$rowClass` variable.

macbre (changes requested):
See the inline comment

extensions/wikia/ArticleComments/modules/templates/ArticleComments_Comment.php
macbre: Or actually: can we move `staff` class to `$rowClass` variable - i.e. move the logic from a template to the code.

macbre: And as we're here: let's make sure `$comment['username']` is properly encoded.

macbre (approved):

PR #11878, opened 2016-11-15
WW-5171820 design review fixes

cormorando:
//cc @Wikia/west-wing


extensions/wikia/CreateNewWiki/css/_variables.scss
```diff
@@ -2,6 +2,7 @@ $content-spacing-xxsmall: 6px;
 $content-spacing-xsmall: 10px;
 $content-spacing-small: 12px;
 $content-spacing-small-plus: 14px;
+$content-spacing-regular-minus: 16px;
```
lukaszkonieczny, line 5: This name sounds weird for me ;) I see that `$content-spacing-xsmall` and `$content-spacing-xlarge` are never used so maybe we can use that names?

idradm: Those names are used the same way in CommunityPage. Please add a new variable instead of changing their sizes to avoid confusion.

idradm (approved):

lukaszkonieczny (approved):
After offline discussion we decided to leave the name

PR #11887, opened 2016-11-16
SUS-1275 Add tracking of impressions for custom 404 pages

sebastianmarzjan:
Track impressions too, to be able to determine CTR on the found links as well as how many of those we actually serve.

//cc @mixth-sense @gabrys


gabrys:
LGTM

PR #11940, opened 2016-11-22
ADEN-4270 Hide .hiddenLinks

fraszczakszymon:


skins/oasis/css/core/WikiHeader.scss
```diff
@@ -37,6 +37,7 @@ $WikiNavWidth: $width-outside - 271px;
 	.hiddenLinks {
 		position: absolute;
 		top: -1000px;
+		visibility: hidden;
```
bognix, line 40: why not `display:none`?

bognix: probably because with `display:none` the access keys won't work

bognix (approved):
LGTM

PR #11947, opened 2016-11-23
ADEN-4277 | disable MMS for logged in users

bognix:


v3rth (approved):
LGTM

extensions/wikia/ARecoveryEngine/templates/ARecoveryEngineApiController_getBootstrap.mustache
```diff
@@ -41,7 +41,7 @@
 					helper.verifyContent();
 				});
 			});
-		} else if (context.opts.sourcePointMMS === true) {
+		} else if (context.opts.sourcePointMMS === true && !window.wgUserName) {
```
nandy-andy, line 44: What do you think of using `context.opts.showAds` instead?

bulaszenko:
Checked works good

PR #11962, opened 2016-11-25
XW | Fix messages for Search related wiki

kvas-damian:
Fix Search messages after import from messaging wiki: https://github.com/Wikia/app/commit/6e78371ec17ea505e588313e376dd9f8d8be5af2

CC @Wikia/x-wing


Warkot:
How does it work and how will it after change in a case as shown below?
PL: `123 tys. stron` or `123 tys. strony`
Or this PR doesn't take care of it at all?

kvas-damian:
@Warkot this PR fixes the problem with using the right plural version. The same messages were here some time ago, but got replaced by import from messaging wiki.

Warkot (approved):

PR #12033, opened 2016-12-01
ADEN-4322 Stop indexing part of our content

nandy-andy:
https://perishablepress.com/tell-google-to-not-index-certain-parts-of-your-page/

In order not to display the message addressed to users who use ad blocking software in Google Search results, we added `googleoff` and `googleon` tags.

//cc @gabrys @saipetch

### DIFF
--- a/skins/oasis/modules/templates/Body_Index.php
+++ b/skins/oasis/modules/templates/Body_Index.php
@@ -95,7 +95,8 @@
 						<br />
 						<h3><?= wfMessage('arecovery-blocked-message-part-one')->escaped() ?>
 							<br /><br />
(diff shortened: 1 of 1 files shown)


skins/oasis/modules/templates/Body_Index.php
```diff
@@ -90,13 +90,16 @@
 					<div id="contentSub"><?= $subtitle ?></div>
 				<?php } ?>
 				<?php if ( ARecoveryModule::isLockEnabled() ) { ?>
+					<!--googleoff: all-->
```
bognix, line 93: google isn't the only bot that crawls our site

gabrys:
The official docs linked in the documents are 404 now. I wonder if it still works. Worth a try anyway.

PR #12043, opened 2016-12-01
RECIRC 236: Add Liftigniter as a data provider for RECIRCULATION_MIX test

pauloslund:
Going forward Liftigniter will be our primary data provider for recirculation. Since we want to do a staged roll-out, we are using our RECIRCULATION_MIX test so we can measure performance as we go against our established controls.

Example configuration:

```js
var recircExperiment = [
	// The first two configurations are setting up the actual helpers. `fandom` 
	// loads data and places it directly in the rail and `wiki` just loads data 
	// to be used later. In the context of the code, these first two
	// configurations are what actually call `loadData` on the liftigniter
	// helper and register the API calls
	{
		id: 'fandom',
		placement: 'rail',
		helper: 'liftigniter',
		options: {
			max: 12,
			widget: 'fandom-rec',
			source: 'fandom',
			opts: {
				resultType:'cross-domain',
				domainType:'fandom.wikia.com'
			}
		}
	},
	{
		id: 'wiki',
		helper: 'liftigniter',
		options: {
			max: 17,
			widget: 'in-wiki',
			source: 'wiki'
		}
	},
	// The next three configurations are just placing the data that was already 
	// loaded. Two of them are for placing `wiki` data in the scroller and
	// impactFooter and the other one is just for placing `fandom` data in
	// the impactFooter
	{
		source: 'wiki',
		placement: 'scroller',
		options: {
			limit: 12
		}
	},
	{
		source: 'wiki',
		placement: 'impactFooter',
		options: {
			offset: 12,
		}
	},
	{
		source: 'fandom',
		placement: 'impactFooter',
		options: {
			offset: 5,
		}
	}
];
```

### DIFF
--- a/extensions/wikia/Recirculation/js/helpers/LiftigniterHelper.js
+++ b/extensions/wikia/Recirculation/js/helpers/LiftigniterHelper.js
@@ -1,0 +1,87 @@
+define('ext.wikia.recirculation.helpers.liftigniter', [
+	'jquery',
+	'wikia.window',
+	'wikia.thumbnailer'
+], function ($, w, thumbnailer) {
+	'use strict';
+
+	var helper = function (config) {
+		var defaults = {
+				max: 5,
+				width: 320,
+				height: 180
+			},
+			options = $.extend({}, defaults, config);
+
+		function loadData() {
+			var deferred = $.Deferred(),
+				registerOptions = {
+					max: options.max * 2, // We want to load twice as many because we filter based on thumbnails
+					widget: options.widget,
+					callback: function (response) {
+						deferred.resolve(formatData(response));
+					}
+				};
+
+			if (!w.$p) {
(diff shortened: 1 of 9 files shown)


extensions/wikia/Recirculation/js/helpers/LiftigniterHelper.js
```diff
@@ -0,0 +1,92 @@
+define('ext.wikia.recirculation.helpers.liftigniter', [
+	'jquery',
+	'wikia.window',
+	'wikia.thumbnailer'
+], function ($, w, thumbnailer) {
+	'use strict';
+
+	var helper = function (config) {
+		var defaults = {
```
gabrys, line 9: minor note: maybe move the `defaults` one level up?

gabrys (approved):
LGTM

PR #12064, opened 2016-12-03
Cake 921 - fix "contribute" button styling when using the CP

nmonterroso:
@Wikia/cake 
https://wikia-inc.atlassian.net/browse/CAKE-921

There were some display issues with the contribute button styling when the CP was enabled. Not sure what was causing them, but this fixes them.


ArturKlajnerok:
Lgtm 👍

PR #12101, opened 2016-12-08
SEO-624 Update English noarticletext to work with any wiki language

gabrys:
The "logic" (unfortunately coded within the translation layer) for the `noarticletext` message was as follows:

* Show nothing if `Namespace = Category` (this doesn't work for non-English languages, as they have their own names for the category namespace)
* Show the "Article not found" text otherwise

Now the logic is as follows:

* Show nothing if `Namespace is empty` (this **does** work for non-English languages)
* Show the "Article not found" text otherwise

This message will need to be translated to other languages, so for now to test it you can visit a page like that:

http://de.gta.wikia.com/Category:NoSuchCategory?uselang=en <- shows the message - production
http://stable.de.gta.wikia.com/Category:NoSuchCategory?uselang=en <- doesn't show the message - updated code.

Once the message is translated (and that needs to wait for MAIN-9346 first), the same "logic" should work without `?uselang=en`.


mklucsarits:
LGTM, but I don't know how to test it.

gabrys:
(I updated the description to include the testing notes)

PR #12102, opened 2016-12-08
SEO-337 Don't show MediaWiki redirects on local HTML sitemap

gabrys:
Testing notes:

The local HTML sitemap is available on all wikis under `/wiki/Local_Sitemap`.

The code is based on Special:AllPages with a bunch of modifications. The
code will now make the list skip over redirects (known bug: this can
potentially produce an empty page, if only redirects were about to be
shown).

Original Special:AllPages should be unaffected.


vforge:
LGTM

PR #12116, opened 2016-12-15
CAKE - recirc incontent template fixes

nmonterroso:
@Wikia/cake 

Adding the missing `data-*` tags for the `incontent` template. Without these, we aren't getting correct tracking information in LiftIgniter. ~~With these, however, there seems to be a race-condition that's introduced with the tracking for the second experiment being set up before it's correctly rendered. @gbenson-wikia says that this is OK for now.~~

The race condition mentioned above was an issue with the way we were defining our experiments. From the [docs](https://liftigniter.readme.io/docs/pregister), we can call `register` multiple times, but only call `fetch` once.

### DIFF
--- a/extensions/wikia/Recirculation/templates/client/incontent.mustache
+++ b/extensions/wikia/Recirculation/templates/client/incontent.mustache
@@ -3,7 +3,14 @@
 
 	<div class="items">
 		{{#items}}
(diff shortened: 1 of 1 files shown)


gbenson-wikia:
Assuming all tests pass, LGTM.

nmonterroso:
Cool, I'll open a PR to the release branch and push this out tomorrow morning.

PR #12122, opened 2016-12-17
SOC-3646 Create new endpoint for creating Discussions for a site

jsutterfield:
Ticket: https://wikia-inc.atlassian.net/browse/SOC-3646

This PR adds a new endpoint to create a discussion site for a community. This endpoint doesn't toggle any of the variables to make it available to the outside world, but instead just provisions the new discussion site with the community's language and title, as well as adds default values for the site in the site attribute service.

This was added for the migration script, which needs to create the discussion site before starting the migration. It's an internal endpoint which is protected by the `X-Wiki-Internal-Request` and `X-MW-AUTH-OK` headers.

// cc @garthwebb


garthwebb:
Why doesn't this toggle the any of the variables to make it available to the outside world?

jsutterfield:
My thinking was in case something went wrong with the migration we wouldn't have a broken discussions exposed while we fixed whatever went wrong. With this approach we first create the discussion site, migrate everything over, and when everything is ready we flip the switches to make it available to users.

garthwebb:
@jsutterfield I guess my question is, how will we know if something is wrong if we can't view the discussion site?  Even if we only toggled the wgEnableDiscussions, this means we'll need some other tool to iterate through a set of site IDs and set the wgEnableDiscussionsNavigation variable in a second step.  Plus I'm not sure we have the man power to QM hundreds/thousands of sites.  We'll have to get this right enough the first time, and make any tweaks after.

For larger sites, we'll probably do this type of QM in dev to make sure all content is migrated appropriately.

jsutterfield:
We touched on this offline, but later in the migration process we'll toggle the MW vars to make this available. So that's done for us already in an automated way.

PR #12136, opened 2016-12-21
XW-2394 | VE allow insert tools only in linear selection

kvas-damian:
https://wikia-inc.atlassian.net/browse/XW-2394

Inserting new content in null and table selection isn't very intuitive. For null selection it can replace content that isn't highlighted. For table selection inserting media or templates doesn't have much sense as user can select multiple cells.
Having that said we decided to limit insert tools only to linear selection.

Ping @Wikia/x-wing

### DIFF
--- a/extensions/VisualEditor/wikia/modules/ve/ui/ve.ui.WikiaCommandRegistry.js
+++ b/extensions/VisualEditor/wikia/modules/ve/ui/ve.ui.WikiaCommandRegistry.js
@@ -81,10 +81,7 @@ ve.ui.commandRegistry.register(
 ve.ui.commandRegistry.register(
 	new ve.ui.Command(
 		'wikiaInfobox', 'window', 'open',
-		{
-			args: ['wikiaInfobox'],
-			supportedSelections: ['linear']
-		}
+		{ args: ['wikiaInfobox'] }
 	)
 );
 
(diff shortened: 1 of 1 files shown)


extensions/VisualEditor/wikia/modules/ve/ui/ve.ui.WikiaCommandRegistry.js
```diff
@@ -21,42 +21,60 @@ ve.ui.commandRegistry.register(
...
+		{
+			args: ['wikiaVideoInsert'],
+			supportedSelections: ['linear']
+		}
 	)
 );
 
 ve.ui.commandRegistry.register(
 	new ve.ui.Command(
 		'wikiaSingleMedia', 'window', 'open',
```
kvas-damian, line 53: This command represents gallery tool.

rogatty: Isn't that obvious? :trollface:

rogatty (approved):

idradm (approved):

extensions/VisualEditor/wikia/modules/ve/ui/ve.ui.WikiaCommandRegistry.js
hakubo: :)

hakubo (approved):

PR #12171, opened 2017-01-05
XW-2456 make sure user can create an template

idradm:
https://wikia-inc.atlassian.net/browse/XW-2456

### DIFF
--- a/extensions/wikia/PortableInfoboxBuilder/PortableInfoboxBuilderHooks.class.php
+++ b/extensions/wikia/PortableInfoboxBuilder/PortableInfoboxBuilderHooks.class.php
@@ -65,7 +65,8 @@ class PortableInfoboxBuilderHooks {
 
 		if ( $wgEnableVisualEditorExt && \VisualEditorHooks::isAvailable( \RequestContext::getMain()->getSkin() ) ) {
 			$aVars['wgEnablePortableInfoboxBuilderInVE'] = $wgEnablePortableInfoboxBuilderInVE &&
-				\RequestContext::getMain()->getUser()->isAllowed( 'createpage' );
+				// use dummy title in template ns
+				Title::newFromText( 'ATemplate', NS_TEMPLATE )->userCan( 'createpage' );
 		}
 
 		return true;
--- a/extensions/wikia/PortableInfoboxBuilder/PortableInfoboxBuilderHooks.class.php
+++ b/extensions/wikia/PortableInfoboxBuilder/PortableInfoboxBuilderHooks.class.php
@@ -65,8 +65,9 @@ class PortableInfoboxBuilderHooks {
 
 		if ( $wgEnableVisualEditorExt && \VisualEditorHooks::isAvailable( \RequestContext::getMain()->getSkin() ) ) {
 			$aVars['wgEnablePortableInfoboxBuilderInVE'] = $wgEnablePortableInfoboxBuilderInVE &&
-				// use dummy title in template ns
-				Title::newFromText( 'ATemplate', NS_TEMPLATE )->userCan( 'createpage' );
+				// user needs permissions to create template when using infobox builder, we check those on dummy template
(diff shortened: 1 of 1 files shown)


extensions/wikia/PortableInfoboxBuilder/PortableInfoboxBuilderHooks.class.php
```diff
@@ -65,7 +65,8 @@ public static function onMakeGlobalVariablesScript( array &$aVars ) {
 
 		if ( $wgEnableVisualEditorExt && \VisualEditorHooks::isAvailable( \RequestContext::getMain()->getSkin() ) ) {
 			$aVars['wgEnablePortableInfoboxBuilderInVE'] = $wgEnablePortableInfoboxBuilderInVE &&
-				\RequestContext::getMain()->getUser()->isAllowed( 'createpage' );
+				// use dummy title in template ns
+				Title::newFromText( 'ATemplate', NS_TEMPLATE )->userCan( 'createpage' );
```
macbre: :+1:

rybmat (approved):

extensions/wikia/PortableInfoboxBuilder/PortableInfoboxBuilderHooks.class.php
```diff
@@ -65,7 +65,8 @@ public static function onMakeGlobalVariablesScript( array &$aVars ) {
 
 		if ( $wgEnableVisualEditorExt && \VisualEditorHooks::isAvailable( \RequestContext::getMain()->getSkin() ) ) {
 			$aVars['wgEnablePortableInfoboxBuilderInVE'] = $wgEnablePortableInfoboxBuilderInVE &&
-				\RequestContext::getMain()->getUser()->isAllowed( 'createpage' );
+				// use dummy title in template ns
```
rogatty: Please write an explanation why we do this. Do we care only about [`Title::isNamespaceProtected()`](https://github.com/Wikia/app/blob/dev/includes/Title.php#L2536-L2547) or we want to run all checks it have?

extensions/wikia/PortableInfoboxBuilder/PortableInfoboxBuilderHooks.class.php
rogatty: What will happen if there is already the `Template:ATemplate` page? What if it's protected (`protected_titles` table)? Should we use a longer string that is very unlikely to be an actual title?

rogatty (changes requested):

extensions/wikia/PortableInfoboxBuilder/PortableInfoboxBuilderHooks.class.php
idradm: The whole idea is to check if the user can create any template, this is the easiest way of doing that and it "should" return the correct value. I haven't thought about the case when `ATemplate` is protected to be edited only by admins and moderators. Do you have any proposal on what title could we use? On the other hand, I would recommend to a static name as it would be easier to debug in case of bugs.

extensions/wikia/PortableInfoboxBuilder/PortableInfoboxBuilderHooks.class.php
idradm: We would like to simulate a template creation action.

rogatty: Write it in the code ;) `// use dummy title in template ns` isn't helpful, it should say _why_ we do this, and not duplicate what is already visible.

extensions/wikia/PortableInfoboxBuilder/PortableInfoboxBuilderHooks.class.php
rogatty: How about `PortableInfoboxBuilderHooksDummyTemplateTitle`? Not sure about the length limit, but I doubt any user would use that.

rogatty (approved):

PR #12178, opened 2017-01-05
ST-103 Use New Search API

jogura-zz:
Use the new fandom story Search API for retrieving fandom story search results. The service behavior is identical to the previous API but was changed to align with RESTful practices.

@jimmordino


jimmordino (approved):
LGTM

PR #12189, opened 2017-01-09
ADEN-4427 removes HOME_ prefix from adslots: WITHOUT CSS

dianafa:
Adeng introduces change that removes HOME_ prefix from adslots

https://wikia-inc.atlassian.net/browse/ADEN-4427

PART1: WITHOUT CSS


fraszczakszymon:
LGTM 👍

PR #12190, opened 2017-01-09
ADEN-4427 removes HOME_ prefix from adslots: ONLY CSS

dianafa:
Adeng introduces change that removes HOME_ prefix from adslots

https://wikia-inc.atlassian.net/browse/ADEN-4427

PART2: ONYL CSS


fraszczakszymon:
LGTM 👍

PR #12182, opened 2017-01-06
cake-1080 - add sponsored indicator for impact footer items

nmonterroso:
@Wikia/cake 
https://wikia-inc.atlassian.net/browse/CAKE-1080

Adding "sponsored" indicator for impact footer elements. The output looks like:

image

PR #12181, opened 2017-01-06
cake-1079 - adding "sponsored" indicator on right rail

nmonterroso:
@Wikia/cake 
https://wikia-inc.atlassian.net/browse/CAKE-1079

Adding sponsored indicator on the right rail. Since we send this as `presented_by` I'm assuming it will come back that way. If that's not true we can change this afterwards.

The result is:

image

PR #12192, opened 2017-01-09
CAKE-1080 Explicitly set the request fields for liftigniter 

drsnyder:
Explicitly set the request fields for LI and include the "presented_by" field for the sponsored presentation. See PRs #12181, #12182 for the related UI changes.

https://wikia-inc.atlassian.net/browse/CAKE-1080
https://wikia-inc.atlassian.net/browse/CAKE-1079

@Wikia/cake

### DIFF
--- a/extensions/wikia/Recirculation/js/helpers/LiftigniterHelper.js
+++ b/extensions/wikia/Recirculation/js/helpers/LiftigniterHelper.js
@@ -34,6 +34,7 @@ define('ext.wikia.recirculation.helpers.liftigniter', [
 
 			// Callback renders and injects results into the placeholder.
 			w.$p('register', registerOptions);
+			w.$p("setRequestFields", ["rank", "thumbnail", "title", "url", "presented_by"]);
 
 			if (options.flush) {
 				w.$p('fetch');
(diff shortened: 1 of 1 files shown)


extensions/wikia/Recirculation/js/helpers/LiftigniterHelper.js
```diff
@@ -34,6 +34,7 @@ define('ext.wikia.recirculation.helpers.liftigniter', [
 
 			// Callback renders and injects results into the placeholder.
 			w.$p('register', registerOptions);
+			w.$p("setRequestFields", ["rank", "thumbnail", "title", "url", "presented_by"]);
```
nmonterroso: Hm maybe we should include `author` here, since I think that's part of the defaults. Too bad this completely replaces the options instead of just adding to it.

nmonterroso:
minor comment otw lgtm

extensions/wikia/Recirculation/js/helpers/LiftigniterHelper.js
drsnyder: Ok, I added `author` in 55c5a4d. I'm ok removing the fields as a supported configuration from the options. That's one less thing the user of this undocumented service needs to worry about.

PR #12193, opened 2017-01-10
XW-2351 | change image thumbnail size

latata:
https://wikia-inc.atlassian.net/browse/XW-2351
Change image thumbnail size

@Wikia/x-wing


rybmat (approved):
approved as long as proposed values are accepted

PR #12195, opened 2017-01-10
XW-2569 render smart wrapped items

idradm:


extensions/wikia/PortableInfobox/services/PortableInfoboxRenderService.class.php
```diff
@@ -229,4 +234,44 @@ private function isEuropaTheme() {
...
+					$capacity = 0;
+				}
+				$result[] = $item;
+			}
+		}
+
+		return array_merge( $result, $this->calculateSmartWidths( $smartGroup, $capacity ) );
+	}
+
+	private function calculateSmartWidths( $groupData, $capacity ) {
```
latata: I'd change the name of this function to fit its purpose better.

extensions/wikia/PortableInfobox/services/PortableInfoboxRenderService.class.php
```diff
@@ -229,4 +234,44 @@ private function isEuropaTheme() {
...
+		$capacity = 0;
+		$smartGroup = [ ];
+
+		foreach ( $groupData as $item ) {
+			$data = $item['data'];
+
+			if ( $item['type'] === 'data' && ( !isset( $data['layout'] ) || $data['layout'] !== 'default' ) ) {
+
+				if ( $capacity + $data['span'] > $rowCapacity ) {
+					$result = array_merge( $result, $this->calculateSmartWidths( $smartGroup, $capacity ) );
```
latata: I'd extract these 3 lines to separated function to avoid code duplication as it's the same as few lines below

extensions/wikia/PortableInfobox/styles/PortableInfobox.scss
```diff
@@ -67,6 +71,19 @@
...
 
+// **********  Special styles for SMART WRAP group (row-items attribute provided)  ********** //
+
+.pi-smart-group {
+	display: flex;
+	flex-wrap: wrap;
+}
+
+.pi-smart-data {
+	display: block;
```
latata, line 82: `display: block` is default for divs so I think we can remove this line

extensions/wikia/PortableInfobox/styles/PortableInfobox.scss
```diff
@@ -13,6 +13,10 @@
 	width: $infobox-width;
 }
 
+.pi-item {
```
latata: maybe it'd be safer to have `.pi-smart-group .pi-item` selector?

latata (changes requested):

extensions/wikia/PortableInfobox/services/PortableInfoboxRenderService.class.php
```diff
@@ -113,8 +113,13 @@ protected function renderGroup( $groupData ) {
 		$dataItems = $groupData['value'];
```
rybmat: can we rename this variable to for example 'childItems' or just 'items'? dataItems suggests that only <data> children of <group> are stored there

extensions/wikia/PortableInfobox/services/PortableInfoboxRenderService.class.php
```diff
@@ -229,4 +234,44 @@ private function isEuropaTheme() {
...
+		$smartGroup = [ ];
+
+		foreach ( $groupData as $item ) {
+			$data = $item['data'];
+
+			if ( $item['type'] === 'data' && ( !isset( $data['layout'] ) || $data['layout'] !== 'default' ) ) {
+
+				if ( $capacity + $data['span'] > $rowCapacity ) {
+					$result = array_merge( $result, $this->calculateRowItemsSize( $smartGroup, $capacity ) );
+					$capacity = 0;
```
rybmat: I think 'currentSize' or 'slotsTaken' better fits to what this variable contains than 'capacity'

extensions/wikia/PortableInfobox/services/PortableInfoboxRenderService.class.php
```diff
@@ -229,4 +234,44 @@ private function isEuropaTheme() {
...
+					$smartGroup = [ ];
+				}
+				$result[] = $item;
+			}
+		}
+
+		return array_merge( $result, $this->calculateRowItemsSize( $smartGroup, $capacity ) );
+	}
+
+	private function calculateRowItemsSize( $rowItems, $capacity ) {
```
rybmat: also here 'capacity' is misleading. Maybe 'rowSize'?

extensions/wikia/PortableInfobox/services/PortableInfoboxRenderService.class.php
idradm: This is pretty hard to refactor, as I only reset those variables.

latata (approved):

PR #12196, opened 2017-01-10
ADEN-4481 separate private and open market data for rubicon

dianafa:
https://wikia-inc.atlassian.net/browse/ADEN-4481

### DIFF
--- a/extensions/wikia/AdEngine/js/lookup/rubicon/rubiconFastlane.js
+++ b/extensions/wikia/AdEngine/js/lookup/rubicon/rubiconFastlane.js
@@ -183,20 +183,20 @@ define('ext.wikia.adEngine.lookup.rubicon.rubiconFastlane', [
 	}
 
 	function getBestSlotPrice(slotName) {
-		var price, pricePrivate;
-
-		if (typeof bestPrices[slotName] !== 'undefined') {
-			price = (bestPrices[slotName] / 100).toFixed(2).toString();
-		}
-
-		if (typeof bestPricesPrivate[slotName] !== 'undefined') {
-			pricePrivate = (bestPricesPrivate[slotName] / 100).toFixed(2).toString();
-		}
-
 		return {
-			fastlane: price,
-			fastlane_private: pricePrivate
+			fastlane: getBestPriceString(bestPrices[slotName]),
+			fastlane_private: getBestPriceString(bestPricesPrivate[slotName])
 		};
+	}
+
+	function getBestPriceString(bestPriceForSlot) {
+		var price;
+
+		if (typeof bestPriceForSlot !== 'undefined') {
+			price = (bestPriceForSlot / 100).toFixed(2).toString();
+		}
+
+		return price;
 	}
 
 	function getSlotParams(slotName) {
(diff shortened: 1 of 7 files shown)


extensions/wikia/AdEngine/js/lookup/rubicon/rubiconTier.js
```diff
@@ -38,8 +39,20 @@ define('ext.wikia.adEngine.lookup.rubicon.rubiconTier', [
...
 	}
 
-	function parsePrice(tier) {
-		var matches = /^\d+_tier(\d+)/g.exec(tier);
+	function parseOpenMarketPrice(tier) {
+		if (tier.indexOf(privateTierPostfix) === -1) {
+			var matches = /^\d+_tier(\d+)/g.exec(tier);
+
+			if (matches && matches[1]) {
+				return parseInt(matches[1], 10);
```
bognix, line 47: something I learnt hard way today - if you pass to Math.max something that is not numeric value as one of the arguments the result would be NaN so make sure to always return a number from those functions.

extensions/wikia/AdEngine/js/lookup/rubicon/rubiconTier.js
```diff
@@ -4,7 +4,8 @@ define('ext.wikia.adEngine.lookup.rubicon.rubiconTier', [
 ], function (math) {
 	'use strict';
 
-	var cpmBuckets = [
+	var privateTierPostfix = 'deals',
```
nandy-andy: Nitpicking: it's more like a suffix ;)

extensions/wikia/AdEngine/js/lookup/rubicon/rubiconTier.js
nandy-andy:
```js
parseInt("some-string", 10);
NaN
parseInt("123", 10);
123
```

extensions/wikia/AdEngine/js/lookup/rubicon/rubiconFastlane.js
```diff
@@ -176,20 +177,26 @@ define('ext.wikia.adEngine.lookup.rubicon.rubiconFastlane', [
...
+			fastlane_private: getBestPriceString(bestPricesPrivate[slotName])
+		};
+	}
+
+	function getBestPriceString(bestPriceForSlot) {
 		var price;
 
-		if (typeof bestPrices[slotName] !== 'undefined') {
-			price = (bestPrices[slotName] / 100).toFixed(2).toString();
+		if (typeof bestPriceForSlot !== 'undefined') {
```
nandy-andy, line 195: Isn't `if (bestPriceForSlot) {` enough here?

dianafa: in case of `0` the `(bestPrices[slotName] / 100).toFixed(2).toString();` wont be performed which is bad

extensions/wikia/AdEngine/js/lookup/rubicon/rubiconTier.js
dianafa: hah, thanks!

extensions/wikia/AdEngine/js/lookup/rubicon/rubiconTier.js
dianafa: do you mean something actionable? In case above `matches[1]` is the corresponding group `(\d+)` of regex which will always be a number

nandy-andy (approved):

PR #12199, opened 2017-01-11
XW-2587 | Portable Infoboxes: fixing image tabs

rybmat:
https://wikia-inc.atlassian.net/browse/XW-2587

//cc @Wikia/x-wing

### DIFF
--- a/extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
+++ b/extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
@@ -257,7 +257,7 @@ class ArticleAsJson extends WikiaService {
 		wfProfileIn( __METHOD__ );
 
 		$title = Title::newFromText( $data['name'] );
-		if ( $title && F::app()->checkSkin( 'wikiamobile' ) ) {
+		if ( $title ) {
 			$details = self::getMediaDetailWithSizeFallback( $title, self::$mediaDetailConfig );
 			$details['context'] = $data['context'];
 			self::$media[] = self::createMediaObject( $details, $title->getText(), $data['caption'] );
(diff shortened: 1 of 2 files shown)


kvas-damian (approved):

extensions/wikia/PortableInfobox/services/PortableInfoboxRenderService.class.php
```diff
@@ -145,6 +145,7 @@ protected function renderImage( $data ) {
 		$images = [ ];
 
 		for ( $i = 0; $i < count( $data ); $i++ ) {
+			$data[$i]['context'] = null;
```
latata, line 148: Wouldn't it be better to set it inside `extendImageData` function?

extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
```diff
@@ -257,7 +257,7 @@ public static function onExtendPortableInfoboxImageData( $data, &$ref ) {
 		wfProfileIn( __METHOD__ );
 
 		$title = Title::newFromText( $data['name'] );
-		if ( $title && F::app()->checkSkin( 'wikiamobile' ) ) {
+		if ( $title ) {
```
latata, line 260: I thought that ArticleAsJson class is only for mobile usage isn't it?

latata (approved):
LGTM

extensions/wikia/PortableInfobox/services/PortableInfoboxRenderService.class.php
rybmat: I don't think so, I think each of render service should set its context

PR #12203, opened 2017-01-11
ADEN-4424 Porvata autoplay

fraszczakszymon:


extensions/wikia/AdEngine/js/video/player/porvata/porvataPlayerFactory.js
```diff
@@ -40,6 +41,10 @@ define('ext.wikia.adEngine.video.player.porvata.porvataPlayerFactory', ['wikia.l
 					width = newWidth;
 					height = newHeight;
 				}
+				if (!width || !height) {
+					width = params.container.offsetWidth;
+					height = params.container.offsetHeight;
+				}
```
dianafa, line 51: Why do we want to set height & width based on offset when one of (newWidth, newHeight) will be 0?

extensions/wikia/AdEngine/js/video/player/porvata/googleImaPlayerFactory.js
```diff
@@ -33,6 +33,14 @@ define('ext.wikia.adEngine.video.player.porvata.googleImaPlayerFactory', [
 			}
 		}
 
+		function setAutoplay(value) {
+			// videoAd DOM element is present on mobile only
+			if (videoAd) {
+				videoAd.autoplay = value;
+				videoAd.muted = value;
```
dianafa: nitpick: fact that while setting autoplay we are also setting `muted` property can be counterintuitive

extensions/wikia/AdEngine/js/video/player/porvata/porvataPlayerFactory.js
```diff
@@ -56,6 +61,9 @@ define('ext.wikia.adEngine.video.player.porvata.porvataPlayerFactory', ['wikia.l
 				ima.getAdsManager().resume();
 			},
 			setVolume: function (volume) {
+				if (videoAd) {
+					videoAd.muted = volume === 0;
+				}
 				return ima.getAdsManager().setVolume(volume);
```
dianafa, line 71: why setting volume to `0` here is not enough? and we need to do it also in l49 in `volumeControl.js`?

extensions/wikia/AdEngine/js/video/player/porvata/porvataPlayerFactory.js
fraszczakszymon: Player size cannot be 0/0 - it will break ad load. Play method is being called without arguments here: 
https://github.com/Wikia/app/blob/1afee07902b184ce5b836b2553b974729b869d61/extensions/wikia/AdEngine/js/video/player/porvata/porvata.js#L53
(it wasn't working properly before)

extensions/wikia/AdEngine/js/video/player/porvata/googleImaPlayerFactory.js
fraszczakszymon: It's a requirement to set both flags for autoplay on mobile:
https://developers.google.com/web/updates/2016/07/autoplay
I wasn't sure if it's worth to set those flags in separate methods. Maybe I should change method name?

extensions/wikia/AdEngine/js/video/player/porvata/porvataPlayerFactory.js
fraszczakszymon: Built-in `ima.setVolume(0)` method doesn't set `muted` attribute on video tag by default.
So if we set some volume without disabling mute flag it won't be unmuted automatically.

extensions/wikia/AdEngine/js/video/player/porvata/porvataPlayerFactory.js
```diff
@@ -56,6 +61,9 @@ define('ext.wikia.adEngine.video.player.porvata.porvataPlayerFactory', ['wikia.l
 				ima.getAdsManager().resume();
 			},
 			setVolume: function (volume) {
+				if (videoAd) {
```
bognix: it looks to me like we are using `videoAd` like a distinguish between mobile and desktop. Am I right?
(I'm basing this on this comment -> https://github.com/Wikia/app/pull/12203/files#diff-e77e2c1ef4d0877d80d59d95c1086bc9R37).

bognix:
LGTM

PR #12206, opened 2017-01-11
XW-2586 | styling PI flex (smart) groups

latata:
https://wikia-inc.atlassian.net/browse/XW-2586

### DIFF
--- a/extensions/wikia/PortableInfobox/services/PortableInfoboxRenderService.class.php
+++ b/extensions/wikia/PortableInfobox/services/PortableInfoboxRenderService.class.php
@@ -263,7 +263,12 @@ class PortableInfoboxRenderService extends WikiaService {
 			}
 		}
 
-		return array_merge( $result, $this->applyRowItemsStyles( $rowItems, $rowSpan, true ) );
+		if ( !empty( $rowItems ) ) {
+			$result =
+				array_merge( $result, $this->applyRowItemsStyles( $rowItems, $rowSpan, true ) );
+		}
+
+		return $result;
 	}
 
 	private function applyRowItemsStyles( $rowItems, $capacity, $isLastRow = false ) {
--- a/extensions/wikia/PortableInfobox/services/PortableInfoboxRenderService.class.php
+++ b/extensions/wikia/PortableInfobox/services/PortableInfoboxRenderService.class.php
@@ -263,30 +263,26 @@ class PortableInfoboxRenderService extends WikiaService {
 			}
 		}
 
-		if ( !empty( $rowItems ) ) {
-			$result =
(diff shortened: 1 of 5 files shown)


extensions/wikia/PortableInfobox/services/PortableInfoboxRenderService.class.php
```diff
@@ -263,15 +263,25 @@ private function createSmartGroups( $groupData, $rowCapacity ) {
...
-		return array_map( function ( $item ) use ( $capacity ) {
+	private function applyRowItemsStyles( $rowItems, $capacity, $isLastRow = false ) {
+		$items = array_map( function ( $item, $index ) use ( $capacity, $isLastRow ) {
 			$width = round( ( $item['data']['span'] / $capacity ) * 100 );
 			$item['data']['cssClasses'] = 'pi-smart-data';
+			if ( $isLastRow ) {
+				$item['data']['cssClasses'] .= ' last-row';
+			}
+			if ( $index % 2 === 1 ) {
+				$item['data']['cssClasses'] .= ' even-in-row';
```
rybmat: what about .pi-smart-data__even-in-row? other classes also should be prefixed

extensions/wikia/PortableInfobox/styles/PortableInfobox.scss
```diff
@@ -79,9 +79,31 @@
...
 
 .pi-smart-data {
+	border-left-style: solid;
+	border-left-width: 1px;
 	display: block;
 	flex-direction: column;
 	text-align: center;
+
+	.pi-data-value {
+		padding: 5px 10px 0;
```
rybmat: are these values used only here? Maybe it is worth to extract them to variables?

extensions/wikia/PortableInfobox/services/PortableInfoboxRenderService.class.php
rybmat: or per @adam suggestion just pi-even-in-row as we do not use BEM in portable infoboxes

idradm: Let's keep those names as short as possible, but please keep the `pi-` prefix. As for the BEM notation, it is not used in PIs anywhere, so let's be consistent with that, and remember that users use those class names as well.

extensions/wikia/PortableInfobox/styles/PortableInfobox.scss
idradm: 👍

idradm (changes requested):

rybmat (approved):

idradm (approved):

PR #12213, opened 2017-01-12
SUS-1446 | Search: get rid of old hooks that were used on Wikia One only

macbre:
https://wikia-inc.atlassian.net/browse/SUS-1446

@mixth-sense / @sqreek


sqreek (approved):

sqreek:
LGTM

PR #12218, opened 2017-01-12
SOC-3743 Fix encoding and escaping issues during migration

jsutterfield:
Ticket: https://wikia-inc.atlassian.net/browse/SOC-3743

This PR doesn't address the main purpose of SOC-3743 (adding board descriptions as forum descriptions), but fixes some issues I found when testing the migration script.

Specifically, this PR makes use of the `ArticleComment` class to extract the title, raw_content, and parsed_content for each Forum Thread that we're migrating. Additionally, we're now explicitly setting the connection parameters as UTF-8 when we're loading the data which is dumped from MW into Discussions. Beforehand the `character_set_client` was being set as Latin-1 (at least on my dev-box), which was leading to encoding issues.

The reason for using `ArticleComment` to extract the title and content can be seen from the screenshots below. By manually extracting that data as we were before, some escaped characters (like quotation marks) where being migrating in their escaped form. Using the APIs provided by the `ArticleComment` class fixes that and prevents us from having to duplicate logic it already implements.

**Before the fix**
screen shot 2017-01-11 at 4 00 06 pm
screen shot 2017-01-11 at 3 59 51 pm

**After the fix**:
screen shot 2017-01-12 at 11 39 54 am 1
screen shot 2017-01-12 at 11 40 05 am

###### //cc
@garthwebb

### DIFF
--- a/extensions/wikia/Discussions/maintenance/ForumDumper.php
+++ b/extensions/wikia/Discussions/maintenance/ForumDumper.php
@@ -20,8 +20,6 @@ class ForumDumper {
 	// MySQL to fail the insert.
 	const MAX_CONTENT_SIZE = 65520;
 
-	const REGEXP_MATCH_TITLE = '/<ac_metadata.*title="([^"]*)".*>.*<\/ac_metadata>/';
-
 	// A very loose interpretation of markup favoring false positives for markup.  Match
 	// alphanumerics, anything in a basic URL and punctuation.  If any character in the text
 	// doesn't match, assume there is wiki text and parse it.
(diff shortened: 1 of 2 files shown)


extensions/wikia/Discussions/maintenance/ForumDumper.php
```diff
@@ -74,26 +72,15 @@ class ForumDumper {
...
 	private $revisions = [];
 	private $votes = [];
 
-	private $dummyTitle;
-	private $parserOptions;
-
-	public function __construct() {
-		$this->dummyTitle = \Title::newFromText( "Dummy" );
-		$this->parserOptions = new \ParserOptions();
-		$this->parserOptions->setEditSection( false );
```
jsutterfield, line 87: We're not using the global parse anymore, so all these could be removed.

garthwebb:
LGTM

PR #12225, opened 2017-01-16
XW-2487 | fix escape key press clearing source editor

latata:
https://wikia-inc.atlassian.net/browse/XW-2487

@Wikia/x-wing

### DIFF
--- a/resources/jquery.ui/jquery.ui.autocomplete.js
+++ b/resources/jquery.ui/jquery.ui.autocomplete.js
@@ -88,8 +88,10 @@ $.widget( "ui.autocomplete", {
 					self.menu.select( event );
 					break;
 				case keyCode.ESCAPE:
-					self.element.val( self.term );
-					self.close( event );
+					if ( self.menu.element.is( ":visible" ) ) {
+						self.element.val(self.term);
+						self.close(event);
+					}
 					break;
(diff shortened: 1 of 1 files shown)


resources/jquery.ui/jquery.ui.autocomplete.js
```diff
@@ -88,8 +88,10 @@ $.widget( "ui.autocomplete", {
 					self.menu.select( event );
 					break;
 				case keyCode.ESCAPE:
-					self.element.val( self.term );
-					self.close( event );
+					if ( self.menu.element.is( ":visible" ) ) {
```
kvas-damian, line 91: Is it possible to use `active` flag on a menu instead of checking styles?

latata: I did it the same as it was done in v. 1.9 of jquery ui (https://github.com/jquery/jquery-ui/blob/1-9-stable/ui/jquery.ui.autocomplete.js#L114). I think it's safer to keep it as is.

kvas-damian: Makes sense

kvas-damian (approved):

weronika (approved):

PR #12226, opened 2017-01-16
(PLATFORM-2813) Vary on Cookie for all external Nirvana requests

Grunny:
Vary on Cookie for all external Nirvana requests by default.

/cc @Wikia/core-platform-team


themech (approved):
Looks good as a quick fix, although I would like us to thing whether we should cache those responses at all

PR #12220, opened 2017-01-13
XW-2448 | Portable Infoboxes: fixing bugs with <gallery> tag used as image

rybmat:
https://wikia-inc.atlassian.net/browse/XW-2448
https://wikia-inc.atlassian.net/browse/XW-2491
https://wikia-inc.atlassian.net/browse/XW-2211

Using gallery data from parser instead of parsing <gallery> tag by regexps.

//cc @Wikia/x-wing  @macbre

PR #12232, opened 2017-01-17
Revert "Merge pull request #12220 from Wikia/XW-2448"

macbre:
It causes https://wikia-inc.atlassian.net/browse/SUS-1483

`Notice: Array to string conversion in /usr/wikia/source/app/includes/parser/StripState.php on line 98`

"Array" string is displayed on blog posts listing.

Reverts #12220

@idradm / @rybmat


macbre:
The only failing test is:

```
SamplerProxyTest::testSampling25Percent

Failed asserting that 23 is equal to 20.0 or is less than 20.0.
```

PR #12235, opened 2017-01-17
SOC-3874 address SOC-3874 and SOC-3875 and one other found URL

slayful:


slayful:
Requires config introduced in https://github.com/Wikia/config/pull/2040

tmszdmsk (approved):
seems fine, but I'm not an expert ;)

extensions/wikia/EmbeddableDiscussions/models/DiscussionsThreadModel.class.php
```diff
@@ -22,11 +20,8 @@ private function apiRequest( $url ) {
 	}
 
 	private function getCategoryRequestUrl() {
-		global $wgDevelEnvironment;
-		if ( empty( $wgDevelEnvironment ) ) {
-			return self::DISCUSSIONS_API_BASE . "$this->cityId/forums?responseGroup=small&viewableOnly=true";
-		}
-		return self::DISCUSSIONS_API_BASE_DEV . "$this->cityId/forums?responseGroup=small&viewableOnly=true";
+		global $wgDiscussionsApiUrl;
+		return "$wgDiscussionsApiUrl/forums?responseGroup=small&viewableOnly=true";
```
drsnyder: Should this be `\F::app()->wg->DiscussionsApiUrl`?

drsnyder:
Hmm. Should these be using the consul resolved address instead of the external `services.*`? With the exception of `SiteStatus` it seems suspicious to me that we would be referencing the external hostnames.

extensions/wikia/Discussions/maintenance/SiteStatus.class.php
```diff
@@ -5,7 +5,6 @@
 
 class SiteStatus {
 	const PROD_SERVICE_URL = 'https://serivces.wikia.com';
```
jsutterfield: Looks like there's a typo here in `https://serivces.wikia.com`

extensions/wikia/EmbeddableDiscussions/models/DiscussionsThreadModel.class.php
jsutterfield: I believe this is also is missing the first `cityId` part of the path.

jsutterfield (commented):
+1 for @drsnyder's recommendation. By using consul we can keep the request inside the DC, and helps keep as much of service resolution being done in one way as possible.

[Here's an example](https://github.com/Wikia/app/blob/dev/includes/wikia/services/WikiDetailsService.class.php#L282-L284) of how that's being done for discussions already.

I think `$wgDiscussionsApiUrl` is still needed though for when the FE needs it as part of `ext.embeddableDiscussions.js`

extensions/wikia/Discussions/maintenance/SiteStatus.class.php
slayful: Hah, that's interesting since it was here a while. I wonder if anybody is watching this site status. ;)

slayful:
Thanks a lot @tmszdmsk, @jsutterfield, @drsnyder. Addressing the feedback!

extensions/wikia/Discussions/maintenance/SiteStatus.class.php
slayful: It's obsolete due to the introduction of `$wgDiscussionsApiUrl` now.

drsnyder:
Please add a brief description to the pull request. That way, our future selves will have more context. Otherwise, LGTM.

extensions/wikia/Discussions/maintenance/SiteStatus.class.php
```diff
@@ -4,8 +4,6 @@
 
 
 class SiteStatus {
-	const PROD_SERVICE_URL = 'https://services.wikia.com';
-	const DEV_SERVICE_URL = 'https://services.wikia-dev.com';
```
jsutterfield, line 8: Looks like these removed consts are still being referred to in [getDiscussionUrl](https://github.com/Wikia/app/blob/43a4a2b887946891729f998973b4f0b83c712d17/extensions/wikia/Discussions/maintenance/SiteStatus.class.php#L360-L364).

jsutterfield:
One last comment, apart from that LGTM!

PR #12240, opened 2017-01-17
Cake 1081 - wiki recommendations beneath impactFooter

nmonterroso:
@Wikia/cake 
https://wikia-inc.atlassian.net/browse/CAKE-1081

Moving `footer` beneath `impactFooter` and applying styling. 

image


extensions/wikia/Recirculation/js/trackers/liftigniter.js
```diff
@@ -17,3 +17,4 @@ $p("init", "l9ehhrb6mtv75bp2", {
     }
 });
 $p("send", "pageview");
+$p("setRequestFields", ["rank", "thumbnail", "title", "url", "presented_by", "author"]);
```
drsnyder, line 20: It seems odd to have this call along side the tracking code. Was it not working properly in the helper?

extensions/wikia/Recirculation/js/utils.js
```diff
@@ -65,7 +95,7 @@ define('ext.wikia.recirculation.utils', [
 	 */
 	function loadTemplate(templateName) {
 		var dfd = new $.Deferred(),
-			templateLocation = 'extensions/wikia/Recirculation/templates/client/' + templateName,
+			templateLocation = 'extensions/wikia/Recirculation/templates/' + templateName,
```
drsnyder, line 98: What happened here? Do some of the `templateName` references include `client/`?

drsnyder: Ah. Nvm. I see the call below to `utils.renderTemplate('client/footer.mustache', data)`.

drsnyder:
Minor comments. Otherwise, LGTM. Thanks for doing that!

extensions/wikia/Recirculation/js/trackers/liftigniter.js
nmonterroso: It was working in the helper, but it doesn't really belong there. A helper is created for each experiment. We're planning to have 3 experiments on the page at once (rail, impact footer, footer), so `setRequestFields` will get called three times. 

I see `trackers/liftigniter.js` as more of an initialization step rather than a tracking step. It inserts the liftigniter sdk onto the page, configures our usage of it, and then tracks the pageview. Tracking the pageview is probably better done somewhere else, but `setRequestFields` feels like a setup step to me.

nmonterroso: Ideally each request to `model.json` would have its own values of `setRequestFields`, since now we're going to want to fetch different data for wiki recommendations (wiki name, impressions, etc) than we do for news+stories. As it is now I think we'll have to have a big call `setRequestFields`, something like `$p('setRequestFields', ['rank', 'thumbnail', 'title', 'url', 'presented_by', 'author', 'wikiImpressions', 'wikiDomain', ...])`

PR #12250, opened 2017-01-18
ADEN-4423 New layout for vUAP

v3rth:


extensions/wikia/AdEngine/js/video/uiTemplate.js
```diff
@@ -0,0 +1,24 @@
...
+	'use strict';
+
+	var autoPlay = [
+			'progressBar',
+			'pauseOverlay',
+			'volumeControl',
+			'closeButton',
+			'toggle'
+		],
+		defaultTemplate = [
```
bognix, line 12: rename to `default` to keep it consistent

extensions/wikia/AssetsManager/config.php
```diff
@@ -966,6 +968,7 @@
 		'#group_adengine2_porvata_js',
 		'#group_adengine2_video_ui_js',
 		'//extensions/wikia/AdEngine/js/video/uapVideo.js',
+		'//extensions/wikia/AdEngine/js/video/uiTemplate.js',
```
bognix: should it be a part of `'#group_adengine2_video_ui_js',` - it looks like it's related and should be used together

extensions/wikia/AdEngine/js/video/player/ui/toggle.js
```diff
@@ -0,0 +1,33 @@
+/*global define, setTimeout*/
+define('ext.wikia.adEngine.video.player.ui.toggle', [
```
bognix: rename to `toggleVideo` so it's a bit more descriptive

extensions/wikia/AdEngine/js/video/player/porvata/porvataPlayerFactory.js
```diff
@@ -4,12 +4,24 @@ define('ext.wikia.adEngine.video.player.porvata.porvataPlayerFactory', [
...
+	function prepareVideoAdContainer(videoAdContainer, params) {
 		DOMElementTweaker.hide(videoAdContainer);
-		videoAdContainer.classList.add('video-player');
+		videoAdContainer.classList.add(videoPlayerClassName);
+
+		if (params.autoPlay) {
+			videoAdContainer.classList.add(autoPlayClassName);
+		}
+
+		if (params.splittedLayoutVideoPosition === 'right') {
```
fraszczakszymon: I'm not sure if it should be a part of porvata player. What do you think about moving it to UAP related modules?

extensions/wikia/AdEngine/js/video/player/porvata/porvataPlayerFactory.js
```diff
@@ -4,12 +4,24 @@ define('ext.wikia.adEngine.video.player.porvata.porvataPlayerFactory', [
...
+			videoAdContainer.classList.add(rightVideoPlayerClassName);
+		} else {
+			videoAdContainer.style.position = 'relative';
```
bognix: `.video-player` already has `position:relative` https://github.com/Wikia/app/pull/12250/files#diff-af3e2bcda52bbc6546cf2826cac9335cR181

extensions/wikia/AdEngine/js/video/uapVideo.js
```diff
@@ -29,17 +39,16 @@ define('ext.wikia.adEngine.video.uapVideo', [
...
-					'pauseOverlay',
-					'volumeControl',
-					'closeButton',
-					'toggleAnimation'
-				], {
+				var uiElements = params.autoPlay ? UITemplate.autoPlay : UITemplate.default;
+
+				log(['VUAP UI elements', uiElements], log.levels.debug, logGroup);
+
+				videoInterface.setup(video, uiElements, {
```
bognix, line 47: videoInterface only works with porvata video. How about defining method called `setupUI` on porvata player instance. This instance should have knowledge about itself - maybe we can avoid passing `params` object through 4 files.

However, above change may be tricky because we don't want to set unnecessary params on porvata player - when deciding which params should be defined as part of porvata player instance keep in mind what is the player responsibility.

extensions/wikia/AdEngine/js/video/player/ui/toggle.js
```diff
@@ -0,0 +1,33 @@
...
+], function (DOMElementTweaker) {
+	'use strict';
+
+	function toggle(elementToShow, elementToHide) {
+		DOMElementTweaker.hide(elementToHide);
+		DOMElementTweaker.show(elementToShow);
+	}
+
+	function hideVideo(video, params) {
+		toggle(params.videoPlaceholderElement || params.image, video.container);
```
bognix: we can avoid `params.videoPlaceholderElement || params.image` by doing this or statement when passing params to `videoInterface.setup` - https://github.com/Wikia/app/pull/12250/files#diff-88f79fd88ef24f1113a283bf78cfaf7eR46

skins/oasis/css/core/ads.scss
```diff
@@ -180,6 +180,14 @@ $uap-video-animation-duration: 0.4s;
 .video-player {
 	position: relative;
 
+	&.autoplay {
+		z-index: 1000;
```
bognix: Video is playing above image not instead of image on the latest version of code.

why do we have this here? I removed it and video worked just fine.

extensions/wikia/AdEngine/js/video/player/porvata/porvataPlayerFactory.js
```diff
@@ -4,12 +4,24 @@ define('ext.wikia.adEngine.video.player.porvata.porvataPlayerFactory', [
 	'wikia.log'
 ], function(DOMElementTweaker, log) {
 	'use strict';
-	var logGroup = 'ext.wikia.adEngine.video.player.porvata.porvataPlayerFactory';
+	var logGroup = 'ext.wikia.adEngine.video.player.porvata.porvataPlayerFactory',
+		autoPlayClassName = 'autoplay',
+		videoPlayerClassName = 'video-player',
+		rightVideoPlayerClassName = 'video-player-right';
```
bognix: how about using one convention for class names instead of two?
you have `autoplay` as a modifier without `video-player` prefix but `video-player-right` is also used as modifier but has the `video-player` prefix. Can we use version with or without prefix in both?

extensions/wikia/AdEngine/js/video/uapVideo.js
```diff
@@ -65,35 +74,42 @@ define('ext.wikia.adEngine.video.uapVideo', [
...
+					var size = getVideoSize(adSlot, params);
+					video.play(size.width, size.height);
 				}
 
 				return video;
 			});
 	}
 
+	function getPercentageVideoHeight(size) {
+		return 100 / (size.width / size.height);
```
bognix: isn't `size.width \ size.height` the same as aspect ratio?

extensions/wikia/AdEngine/js/video/uapVideo.js
```diff
@@ -65,35 +74,42 @@ define('ext.wikia.adEngine.video.uapVideo', [
...
+			size;
 
 		log(['loadVideoAd params', params], log.levels.debug, logGroup);
 
-		params.width = videoWidth;
-		params.height = getVideoHeight(videoWidth, params.videoAspectRatio);
+		size = getVideoSize(videoContainer, params);
+		params.width = size.width;
+		params.height = size.height;
+		params.videoSize = getPercentageVideoHeight(size);
```
bognix: it looks like it's not used anywhere - please remove along with `getPercentageVideoHeight`

bognix (changes requested):

extensions/wikia/AdEngine/js/video/uapVideo.js
v3rth: IMO, moving setupUI to porvata player isn't the best idea, because porvata is and should stay standalone player, so UI should be not defined inside. 

Maybe let's try @mleczey's suggestion and hide players configuration behind some abstraction?

extensions/wikia/AdEngine/js/video/uiTemplate.js
v3rth: I can't :/
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/default

bognix:
LGTM

bognix (approved):

PR #12251, opened 2017-01-18
Fix intermittent failures in sampling tests

jimmordino:
The first attempt at sampling tests used a random seed for ```rand()```, which produced a different sequence of pseudo-random numbers for each test.  Occasionally, the distribution of those pseudo-random numbers was so skewed that the sampling results fell outside the allowable deviation of 10%. 

This fix mocks wraps the ```rand()``` call in a method and mocks that method during tests.  The result ensures an even distribution of "random" numbers during the tests this keeps the measured sampling percentages within a small deviation.  

@macbre @kvas-damian


lib/Wikia/tests/Util/SamplerProxyTest.php
```diff
@@ -96,6 +96,115 @@ function testUnsampledMethod() {
...
+		$this->alternateMock->expects( $alternateCallableRecorder )
+			->method( $this->alternateMethod )
+			->with( $testArg1, $testArg2, $testArg3 )
+			->willReturn( $alternateTestResult );
+
+		// Seed the random generator so that we have a consistent dataset of psuedo-random
+		// numbers across the sample set.  We're trying to test the sampling code, not the
+		// random number generator, so this is reasonable and prevents intermittent test failures
+		// caused by a psuedo-random datasets with a highly skewed distribution.
+		srand(123456);
```
kvas-damian: I'm not fully convinced if `srand` is the best option to solve this problem.
* `srand()` is disabled on some php setups
* test is still dependent on result of external code/class results

Did you consider mocking `rand()` results?

jimmordino: Good catch.  I didn't realize ```srand()``` had fallen out of favor.  Looks like I can leverage namespaces to override ```rand()``` in tests, but that seems a bit unsafe as well.  I'll check to see if we're already using a mock package that can safely mock system functions - I'd rather not add a method to production code just for testing if it can be avoided.

kvas-damian: I think it's a common practice to wrap native php function calls in order to mock them later.

If you're not convinced with this solution then there are tricks to mock such code. Wikia already uses test-helpers (https://github.com/Wikia/php-test-helpers) and runkit (http://php.net/manual/en/book.runkit.php). Both should allow you to solve the problem.

jimmordino:
CR addressed, no longer using ```srand```, now mocking ```rand```

lib/Wikia/tests/Util/SamplerProxyTest.php
```diff
@@ -284,6 +288,18 @@ function testExceptionFallbackFromShadow() {
 		$this->assertEquals( $originalTestResult, $result );
 	}
 
+	function getRandomInt() {
```
kvas-damian: Minor: I'd add a comment or rename this method to show it's not real random. What do you think about `getFakeRandomInt` or something that mentions perfect distribution?

lib/Wikia/tests/Util/SamplerProxyTest.php
```diff
@@ -284,6 +288,18 @@ function testExceptionFallbackFromShadow() {
...
+		static $intArray;
+		static $index = 1;
+
+		if ( !isset( $intArray ) ) {
+			$intArray = range( 1, 100 );
```
kvas-damian, line 306: `100` here is closely connected with `callCount`. If we change `100` to `1000` we can again land into not perfect rand distribution problem. Can we make sure other devs know that by using a variable or putting a comment?

kvas-damian (approved):
Left some minor comments. Except that it LGTM

PR #12252, opened 2017-01-18
SOC-3744 add display_order to inserts for Forums

jsutterfield:
Ticket: https://wikia-inc.atlassian.net/browse/SOC-3744

This PR updates the script used when dumping MW Forums/Threads in order to be migrated into Discussions. The dump now includes a `display_order` column, which is to used later to ensure the discussion categories (Discussions analog to a "Forum") are displayed in the same order as the Forums were on `Special:Forum`.

### DIFF
--- a/extensions/wikia/Discussions/maintenance/ForumDumper.php
+++ b/extensions/wikia/Discussions/maintenance/ForumDumper.php
@@ -9,6 +9,7 @@ class ForumDumper {
 	const TABLE_REVISION = 'revision';
 	const TABLE_TEXT = 'text';
 	const TABLE_VOTE = 'page_vote';
+	const TABLE_PAGE_WIKIA_PROPS  = 'page_wikia_props';
 
 	const CONTRIBUTOR_TYPE_USER = "user";
 	const CONTRIBUTOR_TYPE_IP = "ip";
@@ -45,7 +46,8 @@ class ForumDumper {
 		"sticky_ind",
 		"first_revision_id",
 		"last_revision_id",
-		"comment_timestamp"
+		"comment_timestamp",
+		"display_order"
 	];
 
 	const COLUMNS_REVISION = [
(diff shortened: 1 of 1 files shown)


extensions/wikia/Discussions/maintenance/ForumDumper.php
```diff
@@ -100,10 +102,13 @@ public function getPages() {
 
 		$dbh = wfGetDB( DB_SLAVE );
 		( new \WikiaSQL() )
-			->SELECT_ALL()
+			->SELECT( "page.*, comments_index.*, IF(pp.props is NULL, page.page_id, REPLACE(REPLACE(REPLACE(pp.props, ';', ''), 'i', ''), ':', '')) as idx" )
```
jsutterfield: These nested `REPLACE` calls are pretty hideous, but they're the easiest way I found to transform a value like this: `i:123;` into `123`.

garthwebb: Is there a reason `substring` doesn't work, e.g.:
```sql
SELECT SUBSTRING('i:123', 3);
```

garthwebb: Maybe two of them could be replaced with a substring, e.g.:
```sql
> SELECT SUBSTR(REPLACE('i:12773;', ';', ''), 3);
+-----------------------------------------+
| SUBSTR(REPLACE('i:12773;', ';', ''), 3) |
+-----------------------------------------+
| 12773                                   |
+-----------------------------------------+
```

garthwebb: Actually, why bother with this at all?  The values as-is sort correctly and aren't particularly special (in fact as multi-digit page IDs they're kind of gross).  Initialize a PHP var to zero below in the loop, increment it for each row and assign that. (you'll have to change the sort to ascending).

extensions/wikia/Discussions/maintenance/ForumDumper.php
```diff
@@ -125,10 +130,10 @@ public function getPages() {
 						"sticky_ind" => $row->sticky ?: 0,
 						"first_revision_id" => $row->first_rev_id,
 						"last_revision_id" => $row->last_rev_id,
-						"comment_timestamp" => $row->last_touched
+						"comment_timestamp" => $row->last_touched,
+						"display_order" => $row->idx
```
garthwebb: e.g.
```php
"display_order" => $row_idx++,
```

garthwebb (changes requested):

extensions/wikia/Discussions/maintenance/ForumDumper.php
garthwebb: And actually actually, you don't seem to use the display order here (https://github.com/Wikia/pandora/pull/2225/files#diff-e5c435a43d88dbeabadc3185f2abe152R153) but instead sort, then re-number based on the biggest number already there, so you could leave off any special handling or renumbering altogether.

jsutterfield:
Thanks @garthwebb, updated to use the counter var and got rid of those ugly replaces. Verified it works still works as expected.

garthwebb (approved):

PR #12253, opened 2017-01-18
Cake 1141 - embed community promoDetails on main pages

nmonterroso:
@Wikia/cake 
https://wikia-inc.atlassian.net/browse/CAKE-1141

Extend the existing `SiteAttributeService` class to be able to get a specific attribute for a community, and leverage that for embedding promotion details onto community main pages.


extensions/wikia/Recirculation/RecirculationHooks.class.php
```diff
@@ -96,4 +97,20 @@ static public function canShowDiscussions( $cityId ) {
 			return false;
 		}
 	}
+
+	private static function addMainPageMetadata( OutputPage $outputPage ) {
+		if ( F::app()->wg->Title->isMainPage() ) {
+			$promoDetails = WikiaDataAccess::cache(
+					wfMemcKey( "site-attribute-promoDetails" ),
```
drsnyder: I called this `liftigniterMetadata`. I thought we should connect it to LI in some way but I'm not really passionate about it.

drsnyder:
[Here](https://github.com/Wikia/pandora/pull/2227) are the `site-attribute` service changes.

extensions/wikia/Recirculation/RecirculationHooks.class.php
nmonterroso: Yah, I don't care either but let's just settle on one. `superCoolFeatureThatWillMakeUs$$$`?

nmonterroso: I changed this to `liftigniterMetadata` in 96be263

PR #12255, opened 2017-01-19
ADEN-4556 | track 0 for 0 cpm responses and track '' when bidder doesn't respond

bognix:
https://wikia-inc.atlassian.net/browse/ADEN-4556


dianafa:
LGTM

PR #12258, opened 2017-01-19
Cake 1125 - add global context to LI config

nmonterroso:
@Wikia/cake 
https://wikia-inc.atlassian.net/browse/CAKE-1125

Adding global LI context as specified in their email.


extensions/wikia/Recirculation/js/helpers/LiftigniterHelper.js
```diff
@@ -51,13 +51,18 @@ define('ext.wikia.recirculation.helpers.liftigniter', [
...
+				item.isWiki = false;
 
 				if (items.length < options.max && item.thumbnail) {
 					item.source = options.source;
 					item.meta = options.widget;
 					item.index = index;
 
+                    if (item.source === 'wiki') {
+                        item.isWiki = true;
+                        item.thumbnail = thumbnailer.getThumbURL(item.thumbnail, 'image', options.width, options.height);
```
nmonterroso, line 63: I'm piggybacking in this change because when making the slides I noticed that the images listed in the spreadsheet don't fit the right aspect ratio. Rather than having comdev go through and recreate those images I'll just pass them through Vignette

drsnyder:
The ticket referenced `localStorage.kxallsegs` as a variable that would have some of this data. I can't make sense of any of the data in there. Did you do come to the same conclusion?

LGTM

nmonterroso:
Yeah, no clue. Though I don't think we need it since the wiki variables have that stuff too. We'll see.

PR #12261, opened 2017-01-20
SUS-1403 | do not report stalled articles in Solr index

macbre:
https://wikia-inc.atlassian.net/browse/SUS-1403

Given the architecture of our Solr and indexing pipeline (slaves replication taking place every 3 hours) reporting stale article IDs (i.e. deleted articles) in the index does not really make sense.

`WikiaSearchIndexer` triggers  `StaleResultException` exceptions by default when handling requests from Python worker that processes deletes. Only a few exceptions were reported from Special:Search for delete articles (documents were still in the index, but they were already deleted on a wiki).


mszabo-wikia (approved):

macbre:
@TK-999 gave an offline :+1:

PR #12263, opened 2017-01-20
XW-2637 | setting absolute logout url

latata:
https://wikia-inc.atlassian.net/browse/XW-2637

@Wikia/x-wing


includes/wikia/models/DesignSystemGlobalNavigationModel.class.php
```diff
@@ -235,7 +243,7 @@ private function getLoggedInUserData( $user ) {
 
 		$logOutLink = [
 			'type' => 'link-authentication',
-			'href' => '/logout',
+			'href' => $this->getLogoutUrl(),
```
rogatty: Why not use `$this->getHref( 'user-logout' )` and go the same path as `register` and `sign-in`?

weronika (approved):

rybmat (approved):

lizlux:
This seems to do the trick. LGTM.

PR #12264, opened 2017-01-20
cake-1149 - don't show spotlights when on an english community

nmonterroso:
@Wikia/cake 
https://wikia-inc.atlassian.net/browse/CAKE-1149


drsnyder:
:shipit:

PR #12265, opened 2017-01-20
SOC-3891 Added additional logging to this class

garthwebb:
This logging should make tracking why some emails aren't delivered easier.  Ticket here:

https://wikia-inc.atlassian.net/browse/SOC-3891

### DIFF
--- a/includes/EmailNotification.php
+++ b/includes/EmailNotification.php
@@ -58,6 +58,9 @@ class EmailNotification {
 
 	private $composedCommon = false;
 
+	/** @var \Wikia\Logger\WikiaLogger */
+	private $logger;
+
 	/**
 	 * @param User $editor
 	 * @param Title $title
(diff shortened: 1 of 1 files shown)


includes/EmailNotification.php
```diff
@@ -266,43 +314,90 @@ private function actuallyNotifyOnPageChange( $watchers ) {
 
 		$this->setReplyToAndFromAddresses();
 
-		# The following code is only run, if several conditions are met:
-		# 1. EmailNotification for pages (other than user_talk pages) must be enabled
-		# 2. minor edits (changes) are only regarded if the global flag indicates so
-		if ( !$this->isMinorEdit() || ( $this->notifyUsersOnMinorEdits() && $this->editorWantsToNotifyOnMinorEdits() ) ) {
```
garthwebb, line 272: Moved this logic into the method `editIsSignificant` and then negated the if statement so we can exit at the top of the function and not wrap almost all of this function in an IF block.

includes/EmailNotification.php
```diff
@@ -465,29 +566,29 @@ private function getEmailExtensionController() {
 		$controller = false;
```
garthwebb, line 569: Had a hard time finding class usage because all of these were strings; PHPStorm had no idea they were being accessed.  Does not contribute to logging but a good thing to update.

includes/EmailNotification.php
```diff
@@ -465,29 +568,29 @@ private function getEmailExtensionController() {
...
 
 		if ( $this->isArticlePageEditOrCreatedPage() ) {
-			$controller = 'Email\Controller\WatchedPageEditedOrCreated';
+			$controller = Email\Controller\WatchedPageEditedOrCreatedController::class;
```
jsutterfield, line 571: 👍

jsutterfield (approved):
LGTM!

PR #12280, opened 2017-01-24
cake-1148 - adding _kruxTags to global context

nmonterroso:
@Wikia/cake 
https://wikia-inc.atlassian.net/browse/CAKE-1148

Adding krux tags to LI global context. This is assuming the plan we've sent to LI is the path forward. Changes on the `site-attribute` side will come separately.


drsnyder:
LGTM

PR #12285, opened 2017-01-25
SUS-1522: In Phalanx allow internal requests to look up different user from session user

mszabo-wikia:
Allow internal requests to look up user block info in Phalanx for different users

ticket: https://wikia-inc.atlassian.net/browse/SUS-1522

### DIFF
--- a/extensions/wikia/PhalanxII/models/PhalanxContentModel.php
+++ b/extensions/wikia/PhalanxII/models/PhalanxContentModel.php
@@ -18,11 +18,13 @@ class PhalanxContentModel extends PhalanxModel {
 	 *
 	 * @return bool
 	 */
-	public function isOk() {
+	public function isOk(): bool {
+		global $wgUser;
+
 		return (
-			$this->wg->User->isAllowed( 'phalanxexempt' ) ||
 			!( $this->title instanceof Title ) ||
-			$this->isWikiaInternalRequest()
+			$this->isWikiaInternalRequest() ||
+			$wgUser->isAllowed( 'phalanxexempt' )
 		);
 	}
 
--- a/extensions/wikia/PhalanxII/tests/PhalanxModelTest.php
+++ b/extensions/wikia/PhalanxII/tests/PhalanxModelTest.php
@@ -233,11 +233,10 @@ class PhalanxModelTest extends WikiaBaseTest {
 			->with( $this->equalTo( 'phalanxexempt' ) )
 			->willReturn( false );
 
-		if ( $isSelfCheck ) {
-			$userMock->expects( $this->once() )
+
+		$userMock->expects( $this->any() )
 				->method( 'getName' )
-				->willReturn( $this->app->wg->User->getName() );
-		}
+				->willReturn( $isSelfCheck ? $this->app->wg->User->getName() : 'TestDifferentUserName' );
 
 		$requestMock = $this->getMock( WebRequest::class, [ 'isWikiaInternalRequest', 'getIp' ] );
 		$requestMock->expects( $this->exactly( 3 ) )
--- a/extensions/wikia/PhalanxII/tests/PhalanxModelTest.php
+++ b/extensions/wikia/PhalanxII/tests/PhalanxModelTest.php
@@ -193,6 +193,71 @@ class PhalanxModelTest extends WikiaBaseTest {
 		}
 	}
 
+	/**
+	 * @param bool $isPhalanxExempt whether session user is exempt from Phalanx blocks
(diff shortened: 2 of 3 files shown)


extensions/wikia/PhalanxII/tests/PhalanxModelTest.php
```diff
@@ -193,6 +193,71 @@ public function testPhalanxNewFromType( $type, $content, $className, $methodName
...
+		$userMock->expects( $this->any() )
+			->method( 'isAllowed' )
+			->with( $this->equalTo( 'phalanxexempt' ) )
+			->willReturn( false );
+
+		if ( $isSelfCheck ) {
+			$userMock->expects( $this->once() )
+				->method( 'getName' )
+				->willReturn( $this->app->wg->User->getName() );
+		}
```
macbre: What's the `getName` return value when `isSelfCheck` is set to false?

mszabo-wikia: it is simply empty/null

macbre: In real-life examples `getName` returns either "a real" user name (when logged-in) or an IP address (when an anon). Can we change test here so that mocked values will match ones used in production cases?

jcellary (approved):

extensions/wikia/PhalanxII/models/PhalanxContentModel.php
```diff
@@ -18,11 +18,13 @@ public function __construct( Title $title ) {
 	 *
 	 * @return bool
 	 */
-	public function isOk() {
+	public function isOk(): bool {
+		global $wgUser;
+
 		return (
```
ghost, line 24: Here it is worth to mention one thing. Consider the following code:
```php
function one() {
    echo "One!\n";
    return true;
}

function two() {
    echo "Two!\n";
    return false;
}

function three() {
    echo "Three!\n";
    // Some very expensive logic here
    // Which we do not need to execute every time.
    return true;
}

$result = one() || two() || three();
// Will only output:
// One!
```
Getting back to your code: which expression is most likely to be `true` and which expression is the most expensive to evaluate?

Considering https://github.com/Wikia/app/blob/f18c868bd1a0b4a972d30a520172b1e82fcd63da/extensions/wikia/PhalanxII/models/PhalanxContentModel.php#L11 - unless we deliberately modify `$this->title` after `$this` is instantiated, the part `!( $this->title instanceof Title )` will always be `false` as `$this->title` will always be a `Title`.

Then, `$this->isWikiaInternalRequest()` operates on `$this->wg->Request` whereas `global $wgUser` creates a new reference. I suspect the latter is slower and less probable than the former one.

This might not be of top importance here, it's just a nice opportunity for some general considerations about Boolean algebra and the order of execution. :-)

mszabo-wikia: @mixth-sense 

Definitely! My rearranging of the conditions was with this in mind, although with the following reasoning:

* While we do enforce a strict type constraint on `$title` parameter in constructor, it is still theoretically possible (albeit highly unlikely) for it to receive `null` when it is called from hook, and this sanity check accounts for this possibility.
* And then there is the eternal struggle between `global $wgFoo` vs our Nirvana helpers `$this->wg->Foo`. The problem here is that `wg->Foo` is not actually direct property access (if it was, it would be no question it'd be better). `WikiaGlobalRegistry` does not actually store its elements as its own properties - instead, it provides magic `__get` method which then fetches the variable from `$GLOBALS` superglobal while performing sanity checks on the variable name. So using `global $wgFoo` is just creating a new reference, while `$this->wg->Foo` triggers 3 function calls before returning reference from `$GLOBALS`.

This is something I have been contemplating for a while, and I think it'd be actually beneficial in the long run to move away from Nirvana's `WikiaGlobalRegistry`. @macbre probably has more info on this, but I believe one of its main purposes was to ease testing and mocking globals before we had W(U)TF and test helpers. As it stands, the only useful feature it offers is synchronizing values with `RequestContext` main instance, which is of dubious value since most of the time we only read globals with this helper instead of overwriting them.

So in the spirit of #856 I think it is worth giving this a thought. MediaWiki 1.19 also introduced `RequestContext` which provides a way to move away from globals. Many of our common entry points where we hook into MW code already have a way to access local request context (e.g. through `OutputPage` or `Skin` instance), which would allow us to use this local context in our handlers and pass it along as required.

ghost: :+1:

ghost (approved):

PR #12288, opened 2017-01-25
XW-2668 | padding for PI smart groups without labels

latata:
https://wikia-inc.atlassian.net/browse/XW-2668

@Wikia/x-wing

### DIFF
--- a/extensions/wikia/PortableInfobox/styles/PortableInfoboxEuropaTheme.scss
+++ b/extensions/wikia/PortableInfobox/styles/PortableInfoboxEuropaTheme.scss
@@ -190,12 +190,12 @@ $caption-color-text: mix($color-text, $white, $background-color-mix * 7);
 	// need to have both (pi-data-label & pi-smart-data-label) classes in selector to override stacked layout padding (pi-layout-stacked)
 	.pi-data-label {
 		&.pi-smart-data-label {
-			padding: $infobox-data-spacing $infobox-side-spacing 0;
+			padding: $infobox-data-spacing $infobox-side-spacing $horizontal-group-label-spacing;
 		}
 	}
 
 	.pi-smart-group-head + .pi-smart-group-body .pi-smart-data-value {
-		padding-top: $horizontal-group-label-spacing;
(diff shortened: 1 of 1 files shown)


hakubo (approved):

extensions/wikia/PortableInfobox/styles/PortableInfoboxEuropaTheme.scss
```diff
@@ -190,12 +190,12 @@ $caption-color-text: mix($color-text, $white, $background-color-mix * 7);
...
 	.pi-data-label {
 		&.pi-smart-data-label {
-			padding: $infobox-data-spacing $infobox-side-spacing $horizontal-group-label-spacing;
+			padding: $infobox-data-spacing $infobox-side-spacing 0;
 		}
 	}
 
-	.pi-smart-data-value {
-		padding-top: 0;
+	.pi-smart-group-head + .pi-smart-group-body .pi-smart-data-value {
```
hakubo, line 197: that gets pretty deep though.

just to make sure - does the padding have to be set on value? can it be on body maybe?

extensions/wikia/PortableInfobox/styles/PortableInfoboxEuropaTheme.scss
```diff
@@ -190,12 +190,12 @@ $caption-color-text: mix($color-text, $white, $background-color-mix * 7);
...
+		padding-top: $horizontal-group-label-spacing;
```
kvas-damian: What do you think about not changing padding values. If we make selector more specific probably we'll solve the problem without modifying styles that users can rely on.

extensions/wikia/PortableInfobox/styles/PortableInfoboxEuropaTheme.scss
kvas-damian: @hakubo: We need to override padding value that is applied on `.pi-smart-data-value`

kvas-damian (approved):

hakubo (approved):

PR #12290, opened 2017-01-25
XW-2448 make sure adding new param won't break parser tags

idradm:
https://wikia-inc.atlassian.net/browse/XW-2448

@Wikia/x-wing @macbre @Wikia/platform-team

### DIFF
--- a/extensions/CategoryTree/CategoryTree.php
+++ b/extensions/CategoryTree/CategoryTree.php
@@ -222,14 +222,18 @@ function efCategoryTree() {
  * @return bool
  */
 function efCategoryTreeSetHooks( $parser ) {
+	// Wikia change begin
 	$parser->setHook( 'categorytree' , 'efCategoryTreeRender' );
+	// Wikia change end
 	$parser->setFunctionHook( 'categorytree' , 'efCategoryTreeParserFunction' );
 	return true;
 }
 
+// Wikia change begin
 function efCategoryTreeRender( $cat, $argv, $parser = null ) {
 	return efCategoryTreeParserHook( $cat, $argv, $parser = null, false );
 }
+// Wikia change end
 
 /**
  * Entry point for Ajax, registered in $wgAjaxExportList.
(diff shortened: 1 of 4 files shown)


includes/parser/Parser_LinkHooks.php
```diff
@@ -55,7 +55,6 @@ function firstCallInit() {
 
 		wfProfileIn( __METHOD__ );
 
-		$this->setHook( 'pre', array( $this, 'renderPreTag' ) );
```
idradm, line 58: The `renderPreTag` no longer exists in `Parser` and the new method is registered here:
https://github.com/Wikia/app/blob/db60078cc6bb4185814bdd4df8a957be0889cf68/includes/parser/CoreTagHooks.php#L19-L19

kvas-damian (approved):

hakubo (approved):
Magic 🎆

extensions/CategoryTree/CategoryTree.php
```diff
@@ -222,11 +222,15 @@ function efCategoryTree() {
...
-	$parser->setHook( 'categorytree' , 'efCategoryTreeParserHook' );
+	$parser->setHook( 'categorytree' , 'efCategoryTreeRender' );
 	$parser->setFunctionHook( 'categorytree' , 'efCategoryTreeParserFunction' );
 	return true;
 }
 
+function efCategoryTreeRender( $cat, $argv, $parser = null ) {
+	return efCategoryTreeParserHook( $cat, $argv, $parser = null, false );
+}
+
```
macbre, line 237: Please mark this change with `// Wikia change` comment just in case we want to upgrade this extension in the future.

macbre (changes requested):

extensions/CategoryTree/CategoryTree.php
rogatty: Done :sparkles:

macbre (approved):

PR #12293, opened 2017-01-25
ST-122 Look up renamed fandom-search-service

jimmordino:
With the restructuring of the fandom search service pandora directory the service used to find fandom news and stories is now called ```fandom-search-service```.  This PR changes the code that looks up this service to use the new name.

NOTE: This code should not be merged until the service is deployed under the new name in production.

@jogura


jogura-zz (approved):
Tested it in DEV LGTM

PR #12297, opened 2017-01-26
(PLATFORM-2792) Add method for logging Piggyback entries to StaffLog

Grunny:
/cc @Wikia/core-platform-team


includes/wikia/controllers/HeliosHelperController.class.php
```diff
@@ -291,6 +291,38 @@ public function isBlocked() {
...
 	}
 
+	public function logPiggybackAction() {
+		$this->response->setFormat( 'json' );
+		$this->response->setCacheValidity( \WikiaResponse::CACHE_DISABLED );
+		$this->response->setVal( self::FIELD_SUCCESS, false );
+
+		if ( !$this->authenticateViaTheSchwartz() ) {
+			return;
+		}
```
themech: More like a question than comment - we repeat the setFormat/setCacheValidity/authenticateViaTheSchwartz in each public method of this controller. Is there a way to set it for the whole controller in one place?

themech (approved):

PR #12298, opened 2017-01-26
XW-1779 | DRY design system global footer model

rybmat:
https://wikia-inc.atlassian.net/browse/XW-1779

//cc @Wikia/x-wing


includes/wikia/models/DesignSystemGlobalFooterModel.class.php
```diff
@@ -276,230 +149,140 @@ private function getVerticalData() {
...
-			$data['links'][] = [
-				'type' => 'link-text',
-				'title' => [
-					'type' => 'translatable-text',
-					'key' => 'global-footer-community-link-help'
-				],
-				'href' => $this->getHref( 'help' ),
-				'tracking_label' => 'community.help',
-			];
+		foreach ( $links as $link ) {
```
hakubo, line 243: how about combining array_filter and array_map here for a bit more functional style?

includes/wikia/models/DesignSystemGlobalFooterModel.class.php
```diff
@@ -545,4 +332,91 @@ private function getHref( $hrefKey ) {
 	private function getSocialHrefs() {
 		return DesignSystemSharedLinks::getInstance()->getSocialHrefs( $this->lang );
 	}
+
+	private function getLinkTextObjectList( String $section, Array $linkKeys ) {
```
hakubo, line 321: What is our 'policy' around private functions - do we write unit tests for them?

hakubo (commented):
What do you think about some unit tests? 
Looking good!

includes/wikia/models/DesignSystemGlobalFooterModel.class.php
```diff
@@ -545,4 +332,91 @@ private function getHref( $hrefKey ) {
...
+					'global-footer-' . $section . '-link-' . $link,
+					$link,
+					$section . '.' . $link
+				);
+			},
+			$linkKeys
+		);
+	}
+
+	private function getLinkTextTranslatableObject( $titleKey, $hrefKey, $trackingLabel ) {
```
kvas-damian: What I see here are methods that are only here to provide correct API structure. What do you think about creating separate classes (instead of arrays) to reflect API structure? Then instead of calling those methods you could simply instantiate class objects.

kvas-damian:
Did we consider applying similar improvements for `GlobalNavigation` model?

includes/wikia/models/DesignSystemGlobalFooterModel.class.php
rybmat: Yes, I was thinking about that, but firstly I wanted to check if this is good direction. And extracting those methods to separate classes makes even more sense if we'd like to apply those changes also to Global navigation

includes/wikia/models/DesignSystemGlobalFooterModel.class.php
rybmat: it is a '2 iterations over array vs cleaner code' dilema ;) I don't have strong opinion about this so if you have I can rewrite it.

includes/wikia/models/DesignSystemGlobalFooterModel.class.php
rybmat: We'll probably extract those methods to separate classes, then it makes more sense to unittest it.

includes/wikia/models/DesignSystemGlobalFooterModel.class.php
hakubo: 👍

includes/wikia/models/DesignSystemGlobalFooterModel.class.php
hakubo: Hmm, not sure to be honest.

```
$data['links'] = array_map(array_filter($links, function ($link) {
  return $this->getHref( $link['hrefKey'] );
}), function ($link) {
  return $this->getLinkTextTranslatableObject( $link['titleKey'], $link['hrefKey'], $link['trackingLabel'] );
});
```

did I get this right?

Nah, don't think it adds much readability here actually

includes/wikia/models/DesignSystem/WdsExternalImage.class.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+class WdsExternalImage {
```
hakubo, line 3: how about having all data that is to be serialized in public vars?
```
class MyClass {
    public $url = 'my-url';
    public $type = 'image-external';
}

$m = new MyClass();

echo json_encode($m);
```

These classes should be much simplier then and we would not have to have get() - method, which I find ugly. as you have a class to represent some data but then inside that class you have to do it once again in get method.

What do you think?

@kvas-damian you are PHP master here - does that make sense?

rybmat: how can I make field named `image-data` in a response with this approach? it is not allowed name for field

rybmat: done

hakubo (approved):

includes/wikia/models/DesignSystem/WdsLineImage.class.php
```diff
@@ -0,0 +1,30 @@
...
+class WdsLineImage implements JsonSerializable {
+	use WdsImageTrait;
+	use WdsTitleTrait;
+	use WdsTrackingLabelTrait;
+	use WdsSubtitleTrait;
+
+	public $type = 'line-image';
+
+	// need to implement this method in order to return `image-data` as a key
+	function jsonSerialize() {
```
hakubo: an idea that maybe is slightly less verbose and does not need this 'hook'

```php
class myClass {    
    public function __construct( $data ) {
        $this->{'image-data'} = $data;
    }
}

var_dump(json_encode(new myClass('asrt')));
```

includes/wikia/models/DesignSystem/WdsExternalImage.class.php
kvas-damian: It depends. Usually `get()` method gives you more flexibility to develop your classes. It's also easier to override when inheriting. In this case where use classes only for data representation then I think it's ok to skip getters.

includes/wikia/models/DesignSystem/WdsImageTrait.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+trait WdsImageTrait {
+	public $image;
+
+	public function setSvgImageData( $key ) {
+		$this->image = $key;
+		$this->{'image-data'} = new WdsSvg( $key );
```
kvas-damian, line 8: Instead of "hacking" PHP with variable names with dashes what do you think about using [JsonSerializable](http://nl3.php.net/manual/en/jsonserializable.jsonserialize.php) to rename fields?

This solution allows us to make fields private which means that they're not accessible from outside of class code.

kvas-damian: Also we could mark constant as constants.

@rybmat @hakubo what do you think?

includes/wikia/models/DesignSystem/tests/WdsLinkAuthenticationTest.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
+/**
+ * Created by PhpStorm.
+ * User: ryba
+ * Date: 30/01/2017
+ * Time: 12:05
+ */
```
kvas-damian: This docblock is probably not required ;)

kvas-damian:
I see some lack of consistency in how we instantiate classes. Sometimes we use constructor but sometimes I see that we create empty object and use setters. Do you think we can unify it somehow?

includes/wikia/models/DesignSystem/tests/WdsLinkAuthenticationTest.php
rybmat: definitely

rybmat:
constructor is used only for simple types like text or translatable-text, and if you'd like to have setters for these cases too I'm fine with it. But using constructor for more complex object may not work.

includes/wikia/models/DesignSystem/WdsImageTrait.php
rybmat: implementing jsonSerialize was the previous approach, however after @hakubo showed me this curly-magic-thing it seemed to be more consistent with the rest of the classes

kvas-damian:
This change has low RoI, we decided to stop investing our time in it.

PR #12302, opened 2017-01-26
SUS-1530: cleanup Wall and Forum backend controllers

mszabo-wikia:
- internal requests only (all js facing functions are in the `*ExternalController`s
- removed some unused methods

https://wikia-inc.atlassian.net/browse/SUS-1530

### DIFF
--- a/extensions/wikia/Wall/WallController.class.php
+++ b/extensions/wikia/Wall/WallController.class.php
@@ -49,7 +49,7 @@ class WallController extends WallBaseController {
 	 *
 	 * @desc Renders old User_talk:[username] page in new place, using Wall_renderOldUserTalkPage.php template
 	 *
-	 * @author Andrzej 'nAndy' √Ö¬Åukaszewski
+	 * @author Andrzej 'nAndy' Łukaszewski
 	 */
 	public function renderOldUserTalkPage() {
 		$wallUrl = $this->request->getVal( 'wallUrl' );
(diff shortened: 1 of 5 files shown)


extensions/wikia/Wall/WallController.class.php
```diff
@@ -49,7 +49,7 @@ public function messageDeleted() {
 	 *
 	 * @desc Renders old User_talk:[username] page in new place, using Wall_renderOldUserTalkPage.php template
 	 *
-	 * @author Andrzej 'nAndy' √Ö¬Åukaszewski
+	 * @author Andrzej 'nAndy' Łukaszewski
```
ghost, line 52: @nandy-andy :point_up: :smile: 

It deserves a new term to http://dtz.wikia.com/wiki/S%C5%82owniczek - something like _Łukaszewski’s Annotation_ / _Adnotacja Łukaszewskiego_ :smile:

ghost: Why confused? I think it is actually pretty funny to see a P2 fix that corrects a misspelled name :-)

ghost (approved):

macbre (approved):

extensions/wikia/Wall/WallController.class.php
nandy-andy: I agree it's a nice gesture :) Thank you @TK-999 once again! :bowing_man: Wall is confusing me and something deep inside me would like my name to be removed from those files :stuck_out_tongue_closed_eyes: But what's done can't be undone, right? ;)

PR #12304, opened 2017-01-26
MAIN-9594: fixing links in Special:CreateNewWiki

KockaAdmiralac:
In certain languages, on [Special:CreateNewWiki](http://wikia.com/Special:CreateNewWiki), when inserting wiki domain/name that already exists, the link to the wiki in the error popup points to https://wikia.com/"http://wikiname.wikia.com" instead of to http://wikiname.wikia.com. The bug seems to have been introduced in [this commit](https://github.com/Wikia/app/commit/07d421c5b660550b7f97cc7fe83340a4c3c1ede1), but later incompletely removed in [this one](https://github.com/Wikia/app/commit/224e6b26c5890ce1b7490b1afb71b896b5bcc61e).

Support ticket [#315030](https://support.wikia-inc.com/hc/en-us/requests/315030)
Bug ticket [MAIN-9594](https://wikia-inc.atlassian.net/browse/MAIN-9594) (I hope the link is correct)


onei:
I'm not up to date with recent standards, but I think raw html is generally discouraged in messages. Could you simply replace it with `http://$1.wikia.com` and parse the message instead?

mszabo-wikia:
Should be handled on the translators' side

PR #12327, opened 2017-02-02
DeleteBatch: add --by-id switch

macbre:
For Top 10 Lists cleanup task ([SUS-499](https://wikia-inc.atlassian.net/browse/SUS-499)) we would like to use `maintenance/deleteBatch.php` to delete articles by the list of IDs.

@mixth-sense

### DIFF
--- a/maintenance/deleteBatch.php
+++ b/maintenance/deleteBatch.php
@@ -39,6 +39,9 @@ class DeleteBatch extends Maintenance {
 		$this->addOption( 'i', "Interval to sleep between deletions" );
 		$this->addArg( 'listfile', 'File with titles to delete, separated by newlines. ' .
 			'If not given, stdin will be used.', false );
+
+		// Wikia change
+		$this->addOption( 'by-id', 'Provided list contains article IDs instead of titles', false, false /* $withArgs */ );
 	}
 
 	public function execute() {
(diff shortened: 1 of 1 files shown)


maintenance/deleteBatch.php
```diff
@@ -39,6 +39,9 @@ public function __construct() {
 		$this->addOption( 'i', "Interval to sleep between deletions" );
 		$this->addArg( 'listfile', 'File with titles to delete, separated by newlines. ' .
 			'If not given, stdin will be used.', false );
+
+		// Wikia change
+		$this->addOption( 'by-id', 'Provided list contains article IDs instead of titles', false, false /* $withArgs */ );
```
ghost: add `, separated by commas` to document the format of the input data

ghost (changes requested):

maintenance/deleteBatch.php
macbre: It's actually `separated by newlines` and that's what documentation for `listfile` says.

Should I repeat myself here? :smile:

ghost: Should I repeat myself? :-) Yes, please, duplicate the information. Such bits of information should always be self-contained. They should never come off of the previous ones. This pattern is often used in `--help` of many UNIX commands, although our beloved `grep` seems to be an exception, but at least it uses `likewise` to clearly indicate the relation with the preceding option. :-)

macbre: 2c0d416

ghost: The `grep`'s `likewise` indicates that the translators were translating all help at once, which might not always be the case. That obviously does not apply here to any great extent. I am just mentioning my concern and what I find a good practice.

ghost (approved):

PR #12334, opened 2017-02-02
ADEN-4627 | Adjust Prebid price granularity

mleczey:
## Links
https://wikia-inc.atlassian.net/browse/ADEN-4627
## Description
Adjust Prebid price granularity

### DIFF
--- a/extensions/wikia/AdEngine/js/lookup/prebid/prebidSettings.js
+++ b/extensions/wikia/AdEngine/js/lookup/prebid/prebidSettings.js
@@ -1,5 +1,7 @@
 /*global define*/
-define('ext.wikia.adEngine.lookup.prebid.prebidSettings', [], function () {
+define('ext.wikia.adEngine.lookup.prebid.prebidSettings', [
+	'ext.wikia.adEngine.lookup.prebid.priceGranularityHelper'
+], function (helper) {
 	'use strict';
 
 	/*
@@ -8,25 +10,6 @@ define('ext.wikia.adEngine.lookup.prebid.prebidSettings', [], function () {
 	 * - http://prebid.org/dev-docs/publisher-api-reference.html#module_pbjs.bidderSettings
 	 * - http://prebid.org/dev-docs/examples/custom-price-bucket.html
 	 */
-	function transformPrice(bidResponse) {
-		var cpm = bidResponse.cpm,
-			result = '20.00';
-
-		if (cpm === 0) {
-			result = '0.00';
-		} else if (cpm < 0.05) {
-			result = '0.01';
-		} else if (cpm < 5.00) {
-			result = (Math.floor(cpm * 20) / 20).toFixed(2);
-		} else if (cpm < 10.00) {
-			result = (Math.floor(cpm * 10) / 10).toFixed(2);
-		} else if (cpm < 20.00) {
-			result = (Math.floor(cpm * 2) / 2).toFixed(2);
-		}
-
-		return result;
-	}
-
 	function create() {
 		return {
 			standard: {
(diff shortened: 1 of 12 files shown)


extensions/wikia/AdEngine/js/spec/lookup/prebid/prebidSettings.spec.js
```diff
@@ -0,0 +1,105 @@
...
+		{actual: 100.00, expected: '20.00'}
+	];
+
+	cpms.forEach(function (cpm) {
+		it('settings hb_pb function should transform ' + cpm.actual + ' cpm from bidder response to ' + cpm.expected, function () {
+			var settings = prebidSettings.create(),
+				hb_pb = getFunction(settings, 'hb_pb');
+
+			mocks.bidderResponse.cpm = cpm.actual;
+			var actual = hb_pb(mocks.bidderResponse);
```
dianafa: let's declare vars at the beginning of scope

dianafa:
Only a nitpick, rest LGTM

extensions/wikia/AdEngine/js/lookup/prebid/prebidSettings.js
```diff
@@ -0,0 +1,60 @@
...
+
+	/*
+	 * Documentation:
+	 * - http://prebid.org/dev-docs/publisher-api-reference.html#module_pbjs.setPriceGranularity
+	 * - http://prebid.org/dev-docs/publisher-api-reference.html#module_pbjs.bidderSettings
+	 * - http://prebid.org/dev-docs/examples/custom-price-bucket.html
+	 */
+	function transformPrice(bidResponse) {
+		var cpm = bidResponse.cpm,
+			result = '20.00';
```
bognix: why do we want to default to 20.00? maybe null is a better option?

extensions/wikia/AdEngine/js/lookup/prebid/prebidSettings.js
```diff
@@ -0,0 +1,60 @@
...
+	'use strict';
+
+	/*
+	 * Documentation:
+	 * - http://prebid.org/dev-docs/publisher-api-reference.html#module_pbjs.setPriceGranularity
+	 * - http://prebid.org/dev-docs/publisher-api-reference.html#module_pbjs.bidderSettings
+	 * - http://prebid.org/dev-docs/examples/custom-price-bucket.html
+	 */
+	function transformPrice(bidResponse) {
+		var cpm = bidResponse.cpm,
```
bognix: is it possible CPM is not set? Maybe we should proof our solution for such case.

extensions/wikia/AdEngine/js/lookup/prebid/prebidSettings.js
```diff
@@ -0,0 +1,60 @@
+/*global define*/
+define('ext.wikia.adEngine.lookup.prebid.prebidSettings', [], function () {
```
bognix: empty array is redundant.

extensions/wikia/AdEngine/js/lookup/prebid/prebidSettings.js
dianafa: then we don't need to make additional if below I guess

bognix:
and you need to fix all the places in code where we read `pbAg` and send it to tracking. Instead we should track what our custom method is returning.

extensions/wikia/AdEngine/js/lookup/prebid/prebidSettings.js
mleczey: discussed offline, documentation and examples say that cpm is always set and number

extensions/wikia/AdEngine/js/lookup/prebid/prebidSettings.js
mleczey: discussed offline, 20.00 is default value for values >= 20

bognix:
LGTM

nandy-andy (approved):

PR #12335, opened 2017-02-02
Cake 1229 - don't remove footer ad when recirc is present

nmonterroso:
@Wikia/cake 
https://wikia-inc.atlassian.net/browse/CAKE-1229

instead of replacing the contents of `WikiaFooter` with the recirc content, remove the sibling elements of the ad spot `BOTTOM_LEADERBOARD` and then add the recirc content to the end of `WikiaFooter`


drsnyder:
What are the potential siblings? Could one of them be another ad placement?

nmonterroso:
Not from what I can see in the [template](https://github.com/Wikia/app/blob/dev/skins/oasis/modules/templates/Footer_Index.php)

drsnyder:
LGTM

PR #12336, opened 2017-02-02
FAN-1527 | Remove FC permissions

vforge:
- remove fc permissions
- DS api should rely on presence of `wordpressId` user attribute and not permission
- there should be no changes to the output of the DS api if you have the wordpressId (and you had the permission before that):  `/api/v1/design-system/fandoms/1/en/navigation`


gabrys (approved):
LGTM

PR #12338, opened 2017-02-02
PLATFORM-2353: Make FandomBot create Template:Extension DPL

KockaAdmiralac:
[Template:Extension DPL](http://c.wikia.com/wiki/Template:Extension_DPL) often gets created by the first user that refreshed the wiki page. That probably shouldn't be happening so this PR should make FandomBot create it instead.

Support ticket: [#316320](https://support.wikia-inc.com/hc/en-us/requests/316320)
Bug ticket: [PLATFORM-2353](https://wikia-inc.atlassian.net/browse/PLATFORM-2353)


mszabo-wikia (changes requested):
there is now a constant, `Wikia::BOT_USER`

PR #12339, opened 2017-02-03
SUS-1534: Maintenance script for sunsetting video provider

mszabo-wikia:
Goal is to have a maintenance script that is provider-agnostic and can:

- find list of videos by provider on the wiki, and which articles use which of these videos
- delete all videos by this provider
- remove all references to these videos

Please review, especially with regards to:
- string processing magic
- unit test, test cases/coverage
- output format for dry-run (should it be machine-readable)

https://wikia-inc.atlassian.net/browse/SUS-1534


maintenance/Maintenance.php
```diff
@@ -1100,10 +1100,10 @@ protected static function getCoreScripts() {
 	 */
 	protected function &getDB( $db, $groups = array(), $wiki = false ) {
 		if ( is_null( $this->mDb ) ) {
-			return wfGetDB( $db, $groups, $wiki );
-		} else {
-			return $this->mDb;
+			$this->mDb = wfGetDB( $db, $groups, $wiki );
 		}
+
+		return $this->mDb;
```
macbre, line 1106: Are we sure that no script makes a check whether `$this->mDb`  is not set and performs a different action assuming that `wfGetDB` was used?

That would be pretty hacky, but let's double check it.

maintenance/wikia/VideoHandlers/SunsetProvider.php
```diff
@@ -0,0 +1,233 @@
...
+		/** @var object $video */
+		foreach ( $videoList as $video ) {
+			$title = Title::makeTitle( NS_FILE, $video->video_title );
+			$page = WikiPage::factory( $title );
+			$file = wfFindFile( $title );
+
+			// This logic is essentially a simplified form of FileDeleteForm::doDelete or VideoHandlerController::removeVideo
+			// However the order of execution is reversed, since we do not have to update an UI or perform permissions check in this context
+			// This allows for simplified structure
+			$fileDeleteStatus = Status::newGood();
```
macbre: Shouldn't we set this value to an error by default?

macbre (approved):

PR #12342, opened 2017-02-03
VOLDEV-186: Whitelist Gadget-* to make the gadget extension usable

onei:
Also whitelists Gadgets-*, which applies to messages defined by the extension itself, thus allowing users to alter gadgets-definition as well.

@Grunny

### DIFF
--- a/extensions/Gadgets/Gadgets_body.php
+++ b/extensions/Gadgets/Gadgets_body.php
@@ -87,7 +87,10 @@ class GadgetHooks {
 			}
 
 			if ( $section !== '' ) {
-				$section = wfMsgExt( "gadget-section-$section", 'parseinline' );
+				// begin wikia change
+				// wfMsg cleanup
+				$section = wfMessage( "gadget-section-$section" )->parse();
+				// end wikia change
 
 				if ( count ( $available ) ) {
 					$options[$section] = $available;
@@ -103,7 +106,10 @@ class GadgetHooks {
 				'label' => '&#160;',
 				'default' => Xml::tags( 'tr', array(),
 					Xml::tags( 'td', array( 'colspan' => 2 ),
-						wfMsgExt( 'gadgets-prefstext', 'parse' ) ) ),
+						// begin wikia change
+						// wfMsg cleanup
+						wfMessage( 'gadgets-prefstext' )->parse() ) ),
+						// end wikia change
 				'section' => 'gadgets',
 				'raw' => 1,
 				'rawrow' => 1,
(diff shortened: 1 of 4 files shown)


mszabo-wikia:
CR: https://wikia-inc.atlassian.net/browse/SUS-1548

extensions/Gadgets/Gadgets_body.php
```diff
@@ -87,7 +87,10 @@ public static function getPreferences( $user, &$preferences ) {
 			}
 
 			if ( $section !== '' ) {
-				$section = wfMsgExt( "gadget-section-$section", 'parseinline' );
+				// begin wikia change
+				// wfMsg cleanup
+				$section = wfMessage( "gadget-section-$section" )->parse();
```
mszabo-wikia: We should consider either adding separate use for each possible value of `$section`, or at least list the possible values in comments - helps when searching for usage of messages

extensions/Gadgets/SpecialGadgets.php
```diff
@@ -188,7 +208,10 @@ public function showExportForm( $gadget ) {
 			. Html::hidden( 'pages', $exportList )
 			. Html::hidden( 'wpDownload', '1' )
 			. Html::hidden( 'templates', '1' )
-			. Xml::submitButton( wfMsg( 'gadgets-export-download' ) )
+			// begin wikia change
+			// wfMsg cleanup
+			. Xml::submitButton( wfMessage( 'gadgets-export-download' )->escaped() )
```
mszabo-wikia: This function escapes its parameter, so no need for double escaping here

Also this is inside a special page so we can use `$this->msg` to set proper request context

mszabo-wikia (changes requested):

extensions/Gadgets/Gadgets_body.php
onei: In this case, the name of the message is user defined allowing them to group similar gadgets under appropriate heads on gadgets-definition and Special:Gadgets. If you wanted a list of them, you'd have to look them up on Special:PrefixIndex on the wiki the extension is installed on or something similar.

For reference, the messages defined by the extension are prefixed with gadgets-*.

extensions/Gadgets/SpecialGadgets.php
onei: Done in 21954ee

macbre (approved):

macbre:
Ok, let's merge this one into a branch in our upstream repository. Once it passes QA we can merge it into our dev.

PR #12344, opened 2017-02-03
XW-2693 | WikiaRSS - cache feed in memcache

kvas-damian:
https://wikia-inc.atlassian.net/browse/XW-2693

Google RSS has limits on number of requests that are performed by one IP - in our case DC. In order to minimize this problem we introduced memcache per `URL`.

Ping @Wikia/x-wing

### DIFF
--- a/extensions/wikia/WikiaRSS/WikiaRssExternalController.class.php
+++ b/extensions/wikia/WikiaRSS/WikiaRssExternalController.class.php
@@ -63,16 +63,16 @@ class WikiaRssExternalController extends WikiaController {
 
 			if ( !$short && $description ) {
 				$items = $this->getFullItemList( $rss->items, $options );
-				$html = $this->app->renderView( 'WikiaRssExternal', 'fullList', array( 'items' => $items ) );
+				$html = $this->app->renderView( 'WikiaRssExternal', 'fullList', [ 'items' => $items ] );
 			} else {
 				$items = $this->getShortItemList( $rss->items, $options );
-				$html = $this->app->renderView( 'WikiaRssExternal', 'shortList', array( 'items' => $items ) );
+				$html = $this->app->renderView( 'WikiaRssExternal', 'shortList', [ 'items' => $items ] );
 			}
 
 			$this->response->setVal( 'status', true );
 			$this->response->setVal( 'html', $html );
 		} else {
-			$this->response->setVal( 'error', wfMsg( 'wikia-rss-error-invalid-options' ) );
+			$this->response->setVal( 'error', wfMessage( 'wikia-rss-error-invalid-options' )->escaped() );
 		}
 	}
 
--- a/extensions/wikia/WikiaRSS/WikiaRssExternalController.class.php
+++ b/extensions/wikia/WikiaRSS/WikiaRssExternalController.class.php
@@ -28,12 +28,6 @@ class WikiaRssExternalController extends WikiaController {
 				[ 'providerUrl' => $url ]
 			);
 
-			// TODO XW-2693 temporary disable google rss
-			if ( stripos( $url, 'news.google.com') !== false ) {
-				$this->response->setVal('error', wfMsg('wikia-rss-error-wrong-status-503', $url));
-				return;
-			}
-
 			$feedData = WikiaDataAccess::cacheWithLock(
 				wfSharedMemcKey( 'WikiaRss', 'feed', $url ),
 				self::CACHE_TIME,
--- a/extensions/wikia/WikiaRSS/WikiaRssExternalController.class.php
+++ b/extensions/wikia/WikiaRSS/WikiaRssExternalController.class.php
@@ -1,6 +1,8 @@
 <?php
 class WikiaRssExternalController extends WikiaController {
 
+	const CACHE_TIME = 3600;
+
 	/**
 	 * @brief Returns status flag and data in html tags
 	 * @desc Here we use 3rd-part extension magpierss which fetches rss feeds from given url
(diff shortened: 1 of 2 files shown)


extensions/wikia/WikiaRSS/WikiaRssExternalController.class.php
```diff
@@ -1,14 +1,16 @@
 <?php
 class WikiaRssExternalController extends WikiaController {
 
+	const CACHE_TIME = 3600;
```
hakubo, line 4: can we use one of the standard values?

rogatty:
```
const CACHE_SHORT = 10800; // 3 hours
const CACHE_VERY_SHORT = 300; // 5 minutes
```
Neither of them seems fitting, IMO. RSS feed can be updated pretty often.

extensions/wikia/WikiaRSS/WikiaRssExternalController.class.php
```diff
@@ -22,7 +24,7 @@ public function getRssFeeds() {
 			$url = html_entity_decode($options['url']);
 
 			\Wikia\Logger\WikiaLogger::instance()->info(
-				'WikiaRSS request to RSS provider',
+				'WikiaRSS::getRssFeeds',
```
rogatty: `__METHOD__`

extensions/wikia/WikiaRSS/WikiaRssExternalController.class.php
```diff
@@ -32,24 +34,54 @@ public function getRssFeeds() {
...
+						'errorMessageKey' => $errorMessageKey,
+						'error' => $error,
+						'rss' => $rss
+					];
+				}
+			);
 
-			if( $rss->ERROR ) {
-				$this->response->setVal('error', wfMsg('wikia-rss-error', $url, $rss->ERROR));
+			if ( !empty( $feedData['errorMessageKey'] ) ) {
```
hakubo: nitpick:
I'd check if there is an error by checking if error is not null 
and not a message

hakubo (approved):

extensions/wikia/WikiaRSS/WikiaRssExternalController.class.php
```diff
@@ -32,24 +34,54 @@ public function getRssFeeds() {
 				return;
 			}
 
-			$status = null;
-			$rss = @fetch_rss($url, $status);
+			$feedData = WikiaDataAccess::cacheWithLock(
+				wfSharedMemcKey( 'WikiaRss', 'feed', $url ),
+				self::CACHE_TIME,
+				function() use( $url ) {
```
rogatty: How about a separate method? This one is pretty big already.

extensions/wikia/WikiaRSS/WikiaRssExternalController.class.php
hakubo: @kvas-damian what value it had before? 
maybe worth asking if 3 hours is good for that use case?

rogatty: It was always 1 hour. We will now have two caches - Memcache and Varnish. If we use cacheWithLock it doubles the maximum TTL (up to 2 hours). I think this is a sensible maximum for an RSS feed. If we change this value to 3 hours then it would mean that the data can be 6 hours old.

```
Returns cached data if possible
if cached data is older than $cacheTime but fresher than twice that time
- first thread to request it will start getting data
- other threads in the meantime will continue receiving old data
```

extensions/wikia/WikiaRSS/WikiaRssExternalController.class.php
```diff
@@ -32,24 +34,54 @@ public function getRssFeeds() {
 				return;
 			}
 
-			$status = null;
-			$rss = @fetch_rss($url, $status);
+			$feedData = WikiaDataAccess::cacheWithLock(
+				wfSharedMemcKey( 'WikiaRss', 'feed', $url ),
```
rogatty: `__CLASS__, __METHOD__, $url`

extensions/wikia/WikiaRSS/WikiaRssExternalController.class.php
rogatty: `$feedData['error']` is empty for e.g. `$errorMessageKey = 'wikia-rss-empty';`

extensions/wikia/WikiaRSS/WikiaRssExternalController.class.php
kvas-damian: In practice we should never use double cache time. Longer cache time is only used while other process is computing data to be cached. I don't expect it taking more than few seconds. I used this method to limit number of requests to RSS feed on popular page when cache expires. (We can have many concurrent requests fetching the same content in the same time)

extensions/wikia/WikiaRSS/WikiaRssExternalController.class.php
kvas-damian: I don't like using method name for cache key. I don't see a reason to expire cache when we move a method into other place.

extensions/wikia/WikiaRSS/WikiaRssExternalController.class.php
rogatty: I stand corrected :+1:

extensions/wikia/WikiaRSS/WikiaRssExternalController.class.php
rogatty: When we move a method into other place it usually involves changing the logic. `__METHOD__` acts as an automatic cache versioning ;)

rogatty (approved):

PR #12347, opened 2017-02-03
XW-2650 | HTML title for Mercury file pages is consistent with desktop

rogatty:
* https://wikia-inc.atlassian.net/browse/XW-2650
* https://github.com/Wikia/mercury/pull/3062

Mercury will use `Image - Kerm.jpg | Muppet Wiki | Fandom powered by Wikia` instead of `File:Kerm.jpg | Muppet Wiki | Fandom powered by Wikia` for the HTML title. Same for videos.

/cc @gabrys

### DIFF
--- a/extensions/wikia/MercuryApi/MercuryApiController.class.php
+++ b/extensions/wikia/MercuryApi/MercuryApiController.class.php
@@ -357,6 +357,7 @@ class MercuryApiController extends WikiaController {
 
 				if ( $article instanceof Article) {
 					$articleData = MercuryApiArticleHandler::getArticleJson( $this->request, $article );
+					$displayTitle = $articleData['displayTitle'];
 					$data['categories'] = $articleData['categories'];
 					$data['details'] = MercuryApiArticleHandler::getArticleDetails( $article );
 				} else {
@@ -384,10 +385,8 @@ class MercuryApiController extends WikiaController {
 						$data['article'] = $articleData;
 
 						if ( !$title->isContentPage() ) {
-							// This does two things:
-							// - gets displayTitle from parser function {{DISPLAYTITLE:displayTitle}}
-							// - removes the namespace prefix from it
-							$displayTitle = Title::newFromText( $data['article']['displayTitle'] )->getText();
+							// Remove the namespace prefix from display title
+							$displayTitle = Title::newFromText( $displayTitle )->getText();
 							$data['article']['displayTitle'] = $displayTitle;
 						}
 					}
--- a/extensions/wikia/MercuryApi/MercuryApiController.class.php
+++ b/extensions/wikia/MercuryApi/MercuryApiController.class.php
@@ -346,6 +346,7 @@ class MercuryApiController extends WikiaController {
 
 				// getPage is cached (see the bottom of the method body) so there is no need for additional caching here
 				$article = Article::newFromID( $title->getArticleId() );
+				$displayTitle = null;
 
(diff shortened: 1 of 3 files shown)


extensions/wikia/MercuryApi/models/MercuryApi.class.php
```diff
@@ -310,6 +310,37 @@ public function getAdsContext( Title $title ) {
...
+
+		if ( $title->inNamespace( NS_FILE ) ) {
+			$file = WikiaFileHelper::getFileFromTitle( $title );
+			$htmlTitle = SEOTweaksHooksHelper::getTitleForFilePage( $title, $file );
+		} else {
+			$htmlTitle = $displayTitle;
+		}
+
+		if ( empty( $htmlTitle ) ) {
+			$namespace = $namespaces[$title->getNamespace()];
```
kvas-damian: what do you think about using `$title->getPrefixedText()`?

kvas-damian: or at least using `Language::getFormattedNsText()` which replaces underscores?

extensions/wikia/SEOTweaks/SEOTweaksHooksHelper.class.php
```diff
@@ -36,32 +36,39 @@ static function onBeforePageDisplay( $out ) {
...
+	 * @param Title $title
+	 * @param File $file
+	 *
+	 * @return null|string
+	 */
+	public static function getTitleForFilePage( Title $title, File $file ) {
+		$newTitle = null;
+
+		if ( ( new WikiaFileHelper )->isFileTypeVideo( $file ) ) {
+			$newTitle = wfMsg( 'seotweaks-video' ) . ' - ' . $title->getBaseText();
```
kvas-damian: `wfMessage()` ? And escaping?

extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -381,11 +387,9 @@ public function getPage() {
 						$data['article'] = $articleData;
 
 						if ( !$title->isContentPage() ) {
-							// This does two things:
-							// - gets displayTitle from parser function {{DISPLAYTITLE:displayTitle}}
-							// - removes the namespace prefix from it
-							$data['article']['displayTitle'] =
-								Title::newFromText( $data['article']['displayTitle'] )->getText();
+							// Remove the namespace prefix from display title
+							$displayTitle = Title::newFromText( $displayTitle )->getText();
```
hakubo, line 391: This can throw on you
https://github.com/Wikia/app/blob/dev/includes/Title.php#L135

kvas-damian: @hakubo I'm not sure what are you referring to? Can you give us more details?

rogatty: `Title::newFromText` throws when an object is passed. `$displayTitle` comes from `ParserOutput::getTitleText` which is supposed to return only strings.

If it throws, then something bad is happening and we will catch it at the end of the `getPage` method.

kvas-damian: It's not so easy with handling that exception. `newFromText()` throws `MWException` which we don't catch. We catch only `WikiaHttpException` which isn't ancestor for `MWException`.

kvas-damian (approved):

PR #11566, opened 2016-10-04
Wall refactor after merge

sqreek:
Merge conflict after long time
#9988

@macbre

PR #12353, opened 2017-02-06
SUS-1092 | Forum refactor by sqreek

macbre:
See #11566 / https://wikia-inc.atlassian.net/browse/SUS-1092

This reverts commit 3ed0ec9265df4591317632cf83b93c6a17866cc4, reversing
changes made to 898fb96fedc9bd7918b433eb65cfc041e1365e65.

Conflicts:
	extensions/wikia/Forum/ForumController.class.php
	extensions/wikia/Wall/WallBaseController.class.php


extensions/wikia/Forum/ForumBoardInfo.class.php
```diff
@@ -0,0 +1,63 @@
...
+	private $name;
+	private $url;
+	private $description;
+	private $postCount;
+	private $threadCount;
+	/** @var ForumPostInfo $lastPost */
+	private $lastPost;
+
+	/**
+	 * ForumBoardInfo constructor.
```
rybmat: does this comment give us any value?

extensions/wikia/Wall/WallBaseController.class.php
```diff
@@ -635,21 +621,31 @@ protected function checkAndSetUserBlockedStatus( $wallOwner = null ) {
...
+		if ( !empty( $greeting ) && $greeting->exists() ) {
+			$article = new Article( $greeting );
+			$article->getParserOptions();
+			$article->mParserOptions->setIsPreview( true ); // create parser option
+			$article->mParserOptions->setEditSection( false );
+			$greetingText = $article->getParserOutput()->getText();
+		}
 
+		wfRunHooks( 'WallGreetingContent', [ &$greetingText ] );
+		return $greetingText; // used by SWM to add messages to Wall in monobook
```
rybmat: SWM or SMW?

extensions/wikia/Forum/ForumBoard.class.php
```diff
@@ -120,12 +77,41 @@ public function getTotalActiveThreads( $relatedPageId = 0, $db = DB_SLAVE ) {
...
+		return $forumBoardInfo;
+	}
+
 	/**
 	 * clear cache for board info
-	 * @param array $parentPageIds
 	 */
 	public function clearCacheBoardInfo() {
-		$this->getBoardInfo( DB_MASTER );
+		$this->getBoardInfoFromMaster();
```
rybmat: personally I don't like situation when the method name is getSomething and instead of using the returned value we are calling it only because of side effects (cache in this example) But TBH I have no better proposition for name for this method

extensions/wikia/Forum/ForumController.class.php
```diff
@@ -31,30 +31,20 @@ public function board() {
...
-		$this->response->addAsset( 'extensions/wikia/Forum/css/ForumBoard.scss' );
-		$this->response->addAsset( 'extensions/wikia/Wall/css/MessageTopic.scss' );
-
-		// VOLDEV-36: separate monobook styling
-		if ( $this->app->checkSkin( 'monobook' ) ) {
-			$this->response->addAsset( 'extensions/wikia/Forum/css/monobook/ForumMonobook.scss' );
-			$this->response->addAsset( 'extensions/wikia/Forum/css/monobook/ForumBoardMonobook.scss' );
-		}
-
-		$this->addMiniEditorAssets();
```
rybmat: I wonder if this really needs to be separate method

extensions/wikia/Forum/ForumController.class.php
```diff
@@ -113,21 +103,31 @@ public function getWallForIndexPage( $title ) {
...
+					'value' => htmlspecialchars( $value[ 'name' ] ),
+					'content' => htmlspecialchars( $value[ 'name' ] )
+				];
 			}
+
+			$this->response->setVal( 'destinationBoards', $destinationBoards );
 		}
 
-		// MAIN-7865: Don't render MiniEditor on Monobook
-		$this->response->setVal( 'showMiniEditor', $this->wg->EnableMiniEditorExt && $this->app->checkSkin( 'oasis' ) );
```
rybmat: are we sure that this should be removed?

extensions/wikia/Forum/ForumController.class.php
```diff
@@ -356,4 +359,23 @@ public function forumRelatedThreads() {
...
+		$this->response->addAsset( 'extensions/wikia/Forum/css/ForumBoard.scss' );
+		$this->response->addAsset( 'extensions/wikia/Wall/css/MessageTopic.scss' );
+
+		// VOLDEV-36: separate monobook styling
+		if ( $this->app->checkSkin( 'monobook' ) ) {
+			$this->response->addAsset( 'extensions/wikia/Forum/css/monobook/ForumMonobook.scss' );
+			$this->response->addAsset( 'extensions/wikia/Forum/css/monobook/ForumBoardMonobook.scss' );
+		}
+
+		$this->addMiniEditorAssets();
```
rybmat: I wonder if this really needs to be separate method

extensions/wikia/Wall/WallBaseController.class.php
macbre: It's Site Wide Messages, I suppose :) @sqreek ?

extensions/wikia/Forum/ForumBoard.class.php
ghost: I would use `load` instead of `get`. I agree that `get` implies that the method will  go and `get` you something as the returned value. As for better name `load` seems good in this context. We are in the scope of `ForumBoard` and all we need is to `load` the `BoardInfo`.

extensions/wikia/Wall/WallBaseController.class.php
ghost: It is Site Wide Messages, correct.

extensions/wikia/Forum/ForumBoardInfo.class.php
ghost: No. Feel free to remove it. It is ridiculous. :-)

extensions/wikia/Forum/ForumController.class.php
ghost: I think it is an error. MiniEditor is actively used and lack of the value would cause any checks evaluate to `false` unless a default is set.

UPDATE: I don't see any default to be set anywhere. Only `if ($showMiniEditor)` checks.

extensions/wikia/Forum/ForumController.class.php
ghost: It does not. It is used only once and all it does is a side effect. It is not something we can test in a sane way and I do not think a separare `addMiniEditorAssets()` makes `addAssets()` less or more testable. It is simply a piece of code written for IDE rather than for a human. Pointless. :(

extensions/wikia/Wall/WallBaseController.class.php
ghost: Although, it is unclear how exactly Site Wide Messages use this thing.

ghost (dismissed):

extensions/wikia/Forum/ForumBoard.class.php
```diff
@@ -120,12 +77,41 @@ public function getTotalActiveThreads( $relatedPageId = 0, $db = DB_SLAVE ) {
...
 		$this->getTotalActiveThreads( DB_MASTER );
```
macbre: I'd avoid calling these two methods and simply delete memcache keys that are used by these two methods above.

Let's untangle this :spaghetti: :smile:

extensions/wikia/Forum/Forum.class.php
```diff
@@ -286,7 +280,7 @@ protected function createOrEditBoard( $board, $titletext, $body, $bot = false )
 		}
 
 		Forum::$allowToEditBoard = false;
-
+		$board->clearCacheBoardInfo();
```
macbre:
```
Unexpected non-MediaWiki exception encountered, of type "Error"
Error: Call to a member function clearCacheBoardInfo() on null in /usr/wikia/slot1/16124/src/extensions/wikia/Forum/Forum.class.php:283
```

A check similar to the  one in line 242 may be useful here.

extensions/wikia/Forum/ForumBoard.class.php
```diff
@@ -134,4 +120,26 @@ public function clearCacheBoardInfo() {
 		}
 	}
 
+	private function getLastPost( $rev_id ) {
```
macbre: Let's document what kind of values this method returns `ForumPostInfo|null`

extensions/wikia/Forum/ForumBoardInfo.class.php
```diff
@@ -0,0 +1,63 @@
...
+
+	public function setPostCount( $postCount ) {
+		$this->postCount = $postCount;
+	}
+
+	public function setThreadCount( $threadCount ) {
+		$this->threadCount = $threadCount;
+	}
+
+	public function setLastPost( $lastPost ) {
```
macbre: Let's force the class of the argument - `ForumPostInfo`.

extensions/wikia/Forum/ForumController.class.php
```diff
@@ -356,4 +359,23 @@ public function forumRelatedThreads() {
 			$this->response->setVal( 'messages', $messages );
 		}
 	}
+
+	public function messageTopic() {
+		// stub function
+	}
```
macbre: I do not see any usage of this method.

macbre: `Forum_boardNewThread.php` makes the following class:

```php
<?= F::app()->renderPartialCached( 'Wall', 'messageTopic', [ ] ) ?>
```

rybmat (approved):

PR #12356, opened 2017-02-06
XW-2750 | MercuryApi - getWikiVariables start returning Comscore c7Value

kvas-damian:
https://wikia-inc.atlassian.net/browse/XW-2750

MercuryApi: Start returning Comscore `c7Value` as part of targeting section in `getWikiVariables()` method.

Ping @Wikia/x-wing

### DIFF
--- a/extensions/wikia/AnalyticsEngine/AnalyticsProviderComscore.php
+++ b/extensions/wikia/AnalyticsEngine/AnalyticsProviderComscore.php
@@ -12,14 +12,14 @@ class AnalyticsProviderComscore implements iAnalyticsProvider {
 	function trackEvent($event, $eventDetails=array()){
 		switch ($event){
 		  case AnalyticsEngine::EVENT_PAGEVIEW :
-			  if ($this->getC7Value()) {
+			  if (static::getC7Value()) {
 				  return '
 <!-- Begin comScore Tag -->
 <script type="text/javascript">
 	var _comscore = _comscore || [];
-	_comscore.push({ c1: "2", c2: "'.self::$PARTNER_ID.'",
+	_comscore.push({ c1: "2", c2: "'.static::$PARTNER_ID.'",
 	options: {
-		url_append: "'.self::$COMSCORE_KEYWORD_KEYNAME.'='.$this->getC7Value().'"
+		url_append: "'.static::$COMSCORE_KEYWORD_KEYNAME.'='.static::getC7Value().'"
 	}
 });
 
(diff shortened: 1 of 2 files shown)


extensions/wikia/AnalyticsEngine/AnalyticsProviderComscore.php
```diff
@@ -12,14 +12,14 @@ function getSetupHtml($params=array()){
 	function trackEvent($event, $eventDetails=array()){
 		switch ($event){
 		  case AnalyticsEngine::EVENT_PAGEVIEW :
-			  if ($this->getC7Value()) {
+			  if (static::getC7Value()) {
```
rogatty, line 15: Why `static` and not `self`?

rogatty (approved):

extensions/wikia/MercuryApi/models/MercuryApi.class.php
```diff
@@ -146,6 +146,9 @@ public function getWikiVariables() {
 			'theme' => SassUtil::normalizeThemeColors( SassUtil::getOasisSettings() ),
 			'tracking' => [
 				'vertical' => HubService::getVerticalNameForComscore( $wgCityId ),
+				'comscore' => [
+					'c7Value' => AnalyticsProviderComscore::getC7Value(),
```
rogatty, line 150: This can return `false`, I wonder if it wouldn't be better to handle in the API instead of the client? I guess it depends on what we want to do in that case.

extensions/wikia/AnalyticsEngine/AnalyticsProviderComscore.php
kvas-damian: `static` is more flexible and preferred. You can use it in child classes when inheriting methods or attributes.

extensions/wikia/MercuryApi/models/MercuryApi.class.php
kvas-damian: It looks that in fact `false` is not possible.

`WikiFactoryHub::getWikiVertical()` always returns vertical name. I think we can remove this `false'y` handling.

kvas-damian: Just checked kibana and found nothing there for this case.

kvas-damian: For safety: Disabled comscore tracking in mercury if `c7Value` is `false`. The same condition is implemented in: https://github.com/Wikia/app/blob/d676be1e51dd87a5a16bdc4948f09ac4d34ee3fc/extensions/wikia/AnalyticsEngine/AnalyticsProviderComscore.php#L15-L15

kvas-damian:
Ping @nandy-andy @bognix

rogatty (approved):

bognix:
LGTM

PR #12362, opened 2017-02-06
SER-777 Pulled Wall translations from crowdin

jcellary:
https://wikia-inc.atlassian.net/browse/SER-777

Pulled new wall translations from Crowdin, on Keiko's request.

Pinging @macbre and @mixth-sense since Wall is in Sus


extensions/wikia/Wall/Wall.i18n.php
```diff
@@ -4231,14 +4287,29 @@
...
+	'wall-message-edited-summary' => '<a href="$4" >$1</a>（編集：<a class="username" href="$2">$3さん</a>）',
+	'wall-message-limit-reached' => 'スレッドあたりの投稿数の上限は$1件までです。',
+	'wall-action-monobook-login' => 'Please log in to post a message on this wiki',
+	'wall-posting-message-failed-title' => 'メッセージの投稿中にエラーが発生しました',
+	'wall-posting-message-failed-body' => 'メッセージをコピーして、後ほどもう一度お試しください',
+	'tog-enotifmywall-reminder' => '毎日のリマインダーメールも受け取る',
+	'right-walledit' => 'Can edit wall posts',
+	'right-wallarchive' => 'Can archive wall posts',
+	'right-wallremove' => 'Can remove wall threads',
+	'right-wallshowwikiaemblem' => 'Flags user as "Wikia Staff" on wall posts',
```
macbre, line 4303: Shouldn't that say "Fandom Staff"?

macbre (changes requested):

extensions/wikia/Wall/Wall.i18n.php
jcellary: Good catch! I see that the old version also has Wikia used, so I guess we need to adjust the translations in crowdin and repull them. I'll ping Keiko.

mszabo-wikia:
@macbre @jcellary 

We need to verify the state of the translations with LOC and other stakeholders here because Messaging Wiki import left this file out of sync. 😢 

For example, [wall-message-rev-delete](https://github.com/Wikia/app/blob/88eb948e6258a741840ff98b194441785777ff76/extensions/wikia/Wall/Wall.i18n.php#L2305) was changed to ["Borrar revisión"](https://github.com/Wikia/app/blame/dev/extensions/wikia/Wall/Wall.i18n.php#L2157) in 9ee7e97370b11c488e7843a8f5f80d7a0a1c4d8d as part of the import but now it'd be changed back, so we should make sure this is OK to do.

jcellary:
Thanks Mate, I followed up in https://wikia-inc.atlassian.net/browse/MAIN-9694

PR #12364, opened 2017-02-06
ADEN-4640 Handle resolve state ads

v3rth:


extensions/wikia/AdEngine/js/slot/resolveState.js
```diff
@@ -0,0 +1,46 @@
+/* global define */
+define('ext.wikia.adEngine.slot.resolveState', [
```
fraszczakszymon, line 2: What do you think about moving it to some UAP related directory?

extensions/wikia/AdEngine/js/slot/resolveState.js
```diff
@@ -0,0 +1,46 @@
...
+		return params.resolveState.aspectRatio > 0 && params.resolveState.imageSrc !== '';
+	}
+
+	function hasResolvedState(params) {
+		return isForcedByURLParam() || (paramsAreCorrect(params) && !isBlockedByURLParam());
+	}
+
+	function updateAd(params) {
+		log('Resolve state is turned on', logGroup, log.levels.debug);
+		replaceImage(params);
```
fraszczakszymon: Does it mean that ad starts with big image? What do you think about setting urls in both cases and get rid of additional request for image when resolve state is called

extensions/wikia/AdEngine/js/slot/resolveState.js
```diff
@@ -0,0 +1,46 @@
...
+	function replaceImage(params) {
+		params.backgroundImage.src = params.resolveState.imageSrc;
+	}
+
+	function getQueryParam() {
+		return qs.getVal('resolvedState', null);
+	}
+
+	function isForcedByURLParam() {
+		return [true, 'true'].indexOf(getQueryParam()) > -1;
```
fraszczakszymon: Could you add `'1'` value?

extensions/wikia/AdEngine/js/slot/resolveState.js
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	function isBlockedByURLParam() {
+		return [false, 'blocked', 'false'].indexOf(getQueryParam()) > -1;
```
fraszczakszymon: Could you add `'0'` value?

extensions/wikia/AdEngine/js/slot/resolveState.js
```diff
@@ -0,0 +1,46 @@
...
+
+	var logGroup = 'ext.wikia.adEngine.slot.resolveState',
+		qs = new QueryString();
+
+	function replaceImage(params) {
+		params.backgroundImage.src = params.resolveState.imageSrc;
+	}
+
+	function getQueryParam() {
+		return qs.getVal('resolvedState', null);
```
fraszczakszymon: Can we use underscore?

extensions/wikia/AdEngine/js/slot/resolveState.js
v3rth: discussed offline, there is no good place for that function :(

nandy-andy (approved):

PR #12365, opened 2017-02-06
XW-2653 | Link video thumbnails to file pages on Mercury

rogatty:
## Links

* https://wikia-inc.atlassian.net/browse/XW-2653
* https://github.com/Wikia/mercury/pull/3066

## Description

Every thumbnail is now linked to something:
- if it's an image and user provided a link, then use it and flag the thumbnail with `isLinkedByUser`
- if above is not true:
  - if it's a video link to a file page
  - if it's not a video link to the Vignette URL (yes, it supports audio etc.)

## Reviewers

@Wikia/x-wing

### DIFF
--- a/extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
+++ b/extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
@@ -45,7 +45,7 @@ class ArticleAsJson extends WikiaService {
 					'height' => $scaledSize['height'],
 					'width' => $scaledSize['width'],
 					'title' => $media['title'],
-					'link' => $media['link'] ?? '',
+					'href' => $media['href'],
 					'caption' => $media['caption'] ?? ''
 				]
 			)
@@ -65,7 +65,7 @@ class ArticleAsJson extends WikiaService {
 					'title' => $media['title'],
 					'fileUrl' => $media['fileUrl'],
 					'caption' => $media['caption'] ?? '',
-					'link' => $media['link'],
+					'href' => $media['href'],
 					'isLinkedByUser' => $media['isLinkedByUser'],
 					/**
 					 * data-ref has to be set for now because it's read in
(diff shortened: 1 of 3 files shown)


extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
```diff
@@ -65,7 +65,8 @@ private static function renderImage( $media, $id ) {
 					'title' => $media['title'],
 					'fileUrl' => $media['fileUrl'],
 					'caption' => $media['caption'] ?? '',
-					'link' => $media['link'] ?? '',
+					'href' => $media['href'],
+					'isLinkedByUser' => $media['isLinkedByUser'],
```
kvas-damian, line 69: I know it's minor but I'm not sure if the name is ok here. What do you think about simple `isLink`?

rogatty: The issue is that now all the thumbnails are technically speaking links. That's why I wanted to explicitly state that user added a link and not us. I'm not a huge fan of `isLinkedByUser` neither but I don't have a better name that would tell the full story.

kvas-damian: Maybe `isDefaultLink` and negate it's value?

rogatty: That would make sense but I don't agree it'd be easier to understand the logic behind it. "Default" is a pretty generic word and if you don't dive deep into the code you won't know what that means. `isLinkedByUser` is self-explanatory, which is more important than being pretty :wink:

kvas-damian: ok

kvas-damian (approved):

PR #12366, opened 2017-02-07
SUS-868 | set max-width for no-title-warning

rybmat:
https://wikia-inc.atlassian.net/browse/SUS-868

//cc @mixth-sense @macbre


macbre:
```css
lgtm: true;
```

macbre (approved):

PR #12367, opened 2017-02-07
XW-2747 | WikiaRSS fixes

kvas-damian:
https://wikia-inc.atlassian.net/browse/XW-2747

Ping @Wikia/x-wing @Grunny

### DIFF
--- a/extensions/wikia/WikiaRSS/WikiaRssExternalController.class.php
+++ b/extensions/wikia/WikiaRSS/WikiaRssExternalController.class.php
@@ -131,6 +131,7 @@ class WikiaRssExternalController extends WikiaController {
 			if( $item['description'] ) {
 				$text = trim(mb_convert_encoding($item['description'],$outputEncoding,$charset));
 				$text = str_replace( array("\r", "\n", "\t", '<br>'), ' ', $text );
+				$text = Sanitizer::removeHTMLtags( $text, null, [], [ 'img' ] );
 
 				$d_text = $this->doRssFilter($text, $rssFilter );
 				$d_text = $this->doRssFilterOut($text, $rssFilterout);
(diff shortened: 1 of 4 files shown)


extensions/wikia/WikiaRSS/WikiaRssExternalController.class.php
```diff
@@ -127,6 +133,12 @@ private function getFullItemList($items, $options) {
 			if( $item['description'] ) {
 				$text = trim(mb_convert_encoding($item['description'],$outputEncoding,$charset));
 				$text = str_replace( array("\r", "\n", "\t", '<br>'), ' ', $text );
+				$text = strip_tags( $text, '<img><a><br>' );
```
rogatty, line 136: Why do we strip tags twice? First here and then `Sanitizer::removeHTMLtags()`. Is it *to be sure*?

extensions/wikia/WikiaRSS/WikiaRssExternalController.class.php
```diff
@@ -127,6 +133,12 @@ private function getFullItemList($items, $options) {
...
+
+				$globalStateWrppaer = new \Wikia\Util\GlobalStateWrapper( [ 'wgAllowImageTag' => true ] );
```
rogatty: Typo in the var name.

extensions/wikia/WikiaRSS/WikiaRssExternalController.class.php
kvas-damian: `removeHTMLtags()` is more complex method. By defaults it strips tags that are not whitelisted. The same rule is applied to attributes. Other tags are kept in place. Unfortunately this method can't handle more complex HTML (or not fully valid HTML) that's why we limit allowed tags using `strip_tags()`.

rogatty: Interesting, sounds like `removeHTMLtags()` should call `strip_tags()` internally. Not in this PR, obviously :slightly_smiling_face:

kvas-damian: Name for `removeHTMLtags()` is very misleading. In fact it's trying to modify HTML in a way that it's safe.

rogatty (approved):

PR #12373, opened 2017-02-07
SUS-262 | improve Wall performance

macbre:
https://wikia-inc.atlassian.net/browse/SUS-262

### Improvements implemented

* batch calls to `CommentsIndex::selectRow` to avoid multiple queries getting just a single row - introduce `CommentsIndex::newFromIds` (that is called just once per page request, instead of multiple times per a single message wall entry)
* cache the result of a first `ForumHelper::isForum` call - it's executed 30+ times for `$wgTitle`, hence the result is not likely to change :smile: 
* added in-memory cache for `WallMessage::newFromId` method that was called multiple times for the ID of the parent message wall posts
* improved the caching in `WallThread` (as suggested by @TK-999) and saved two additional DB queries (`CommentsIndex::newFromIds` and `Title::newFromIDs`)

### Results

I've tested my changes on a rather small message wall page (with just six messages). Number of queries made dropped from 233 to 77 (when visiting the page as an anon via `curl`).

### Comparing the performance

Let's observe the graphs of select queries performed on both `comments_index` and `page` table when handling Message Wall requests.


extensions/wikia/Wall/WallMessage.class.php
```diff
@@ -39,24 +42,41 @@ function __construct( Title $title, $articleComment = null ) {
...
+		$title = Title::newFromID( $id, $master == true ? Title::GAID_FOR_UPDATE : 0 );
+
 		if ( $title instanceof Title && $title->exists() ) {
 			wfProfileOut( __METHOD__ );
-			return WallMessage::newFromTitle( $title );
+			return self::$wallMessageCache[$id] = WallMessage::newFromTitle( $title );
 		}
 
 		if ( $master == false ) {
+			// TODO: instead of relying on fallback to master let's implement a proper wfWaitForSlaves() use
```
rybmat, line 72: what about this todo?

rybmat (commented):
- do we need setDB method in CommentsIndex.class.php? looks like its unused

extensions/wikia/Wall/WallMessage.class.php
macbre: Let's wait for the logs to go live, take a look at them and decide then. Maybe nothing will be logged (fingers crossed ;)).

macbre:
@rybmat, `setDB` is called in `newFromId` method on a cloned instance of `CommentsIndex` object. Let's take a closer look...

macbre:
I'd rather not touch it at this point.

extensions/wikia/Forum/ForumHelper.class.php
```diff
@@ -2,15 +2,39 @@
...
+	 * @return bool
+	 */
+	private static function isTitleForum( $title ) {
+		if ( $title instanceof Title ) {
+			if ( $title->isSpecial( 'Forum' ) || in_array( $title->getNamespace(), self::$forumNamespaces ) ) {
 				return true;
-			} else if ( $wgTitle->getNamespace() === NS_USER_WALL_MESSAGE ) {
-				$mainTitle = Title::newFromId( $wgTitle->getText() );
+			} else if ( $title->getNamespace() === NS_USER_WALL_MESSAGE ) {
+				$mainTitle = Title::newFromID( $title->getText() );
```
rybmat, line 35: is it correct?

rybmat: per comment bellow it is...

rybmat (approved):

extensions/wikia/Forum/ForumHelper.class.php
macbre: It's it, unfortunately :)

mszabo-wikia: It's by design

TK-999:
Should we consider caching full `WallMessage` objects instead of just ids to spare lookup on `WikiActivity (as experimented in 1020474)?

mszabo-wikia (commented):
Should we consider caching full WallMessage objects instead of just ids to spare lookup on WikiActivity (as experimented in 1020474)?

macbre:
@TK-999 - good point! I'm on it...

rybmat (approved):

mszabo-wikia (approved):
LGTM

rybmat (approved):

macbre:
Travis fails with `npm ERR! shasum check failed for /tmp/npm-2801-302c8201/registry.npmjs.org/sauce-connect-launcher/-/sauce-connect-launcher-0.6.1.tgz` - does not seem to be related.

PR #12374, opened 2017-02-07
MAIN-9681: Don't send popular wikis if hidden

KockaAdmiralac:
**Warning:** This change is untested. Please test it, and sorry if it breaks

The API returns information about wikis the user recently edited on, even if the user has hidden them from their masthead. This change makes a check if that preference is set and returns an empty array if so.

Support ticket: [#316786](http://support.wikia-inc.com/hc/requests/316786)
Bug ticket: [MAIN-9681](https://wikia-inc.atlassian.net/browse/MAIN-9681)
Examples: [API call](http://c.wikia.com/wikia.php?controller=UserProfilePage&method=renderUserIdentityBox&title=User:Count_of_Howard&format=json), [script for unhiding user wikis](http://kocka.wikia.com/wiki/User:KockaAdmiralac/UnhideUserWikis.javascript?action=raw)

### DIFF
--- a/extensions/wikia/UserProfilePageV3/UserIdentityBox.class.php
+++ b/extensions/wikia/UserProfilePageV3/UserIdentityBox.class.php
@@ -585,7 +585,11 @@ class UserIdentityBox extends WikiaObject {
 	 * @return array
 	 */
 	public function getTopWikis( $refreshHidden = false ) {
-		return $this->getFavoriteWikisModel()->getTopWikis( $refreshHidden );
+		if ( $this->user->getGlobalPreference( 'hideEditsWikis' ) ) {
+			return [];
+		} else {
+			return $this->getFavoriteWikisModel()->getTopWikis( $refreshHidden );
+		}
 	}
 
 	/**


mszabo-wikia:
Merged to local feature branch for testing.

mszabo-wikia:
Unit tests are failing:
```
There was 1 error:

1) UserIdentityBoxTest::testGetTopWikis
Error: Call to a member function getGlobalPreference() on null
```
Please update the tests so they pass 🙂

KockaAdmiralac:
Can I push commits to merged branches? If not, the issue seems to be happening because [the user mock](https://github.com/Wikia/app/blob/dev/extensions/wikia/UserProfilePageV3/tests/UserIdentityBoxTest.php#L185) isn't created with `[ 'getOption' ]` (or possibly `[ 'getGlobalPreference' ]`) as the second parameter to `getMock`, and also because [the user identity box mock's constructor isn't being executed](https://github.com/Wikia/app/blob/dev/extensions/wikia/UserProfilePageV3/tests/UserIdentityBoxTest.php#L200), therefore `$this->user` is null.

mszabo-wikia:
Feel free to open a pull request against the feature branch with the fix ^^

PR #12378, opened 2017-02-08
SUS-1556 | do not check tables existence on each request

macbre:
https://wikia-inc.atlassian.net/browse/SUS-1556

Instead rely on CreateNewWiki creating these tables (Wall is enabled site-wide because Forum extension is).

* `comments_index` existence was checked ~80k times a day
* `wall_related_pages` existence was checked ~10mm times a day

### DIFF
--- a/extensions/wikia/Wall/WallRelatedPages.class.php
+++ b/extensions/wikia/Wall/WallRelatedPages.class.php
@@ -47,8 +47,6 @@ class WallRelatedPages extends WikiaModel {
 		wfProfileIn( __METHOD__ );
 		$db = wfGetDB( DB_MASTER );
 
-		$this->createTable();
-
 		$db->begin();
 		$db->delete( 'wall_related_pages', [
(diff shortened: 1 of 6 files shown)


extensions/wikia/Wall/WallRelatedPages.class.php
```diff
@@ -47,8 +47,6 @@ function set( $messageId, $pages = [ ] ) {
 		wfProfileIn( __METHOD__ );
 		$db = wfGetDB( DB_MASTER );
 
-		$this->createTable();
```
mszabo-wikia, line 50: Should we remove the method entirely?

mszabo-wikia (dismissed):

extensions/wikia/Wall/WallRelatedPages.class.php
macbre: We should - eace18b

rybmat (approved):

PR #12382, opened 2017-02-08
SUS-1542 | Fixing caching issues on closing/reopening threads on wall

rybmat:
https://wikia-inc.atlassian.net/browse/SUS-1542

Invalidating cache for wall after closing/reopening thread, so the message about closing the thread will be visible right after closing the thread (or disapear when reopening).

### DIFF
--- a/extensions/wikia/Wall/WallMessage.class.php
+++ b/extensions/wikia/Wall/WallMessage.class.php
@@ -1683,9 +1683,8 @@ class WallMessage {
 	 * The flow then goes to TitleGetSquidURLs hook which cleans the list of URLs in Wall and Forum
 	 */
 	public function purgeSquid() {
-		$title = Title::newFromID( $this->getId() );
-		if ( $title instanceof Title ) {
-			$title->purgeSquid();
+		if ( $this->title instanceof Title ) {
+			$this->title->purgeSquid();
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


extensions/wikia/Wall/WallMessage.class.php
```diff
@@ -1679,13 +1679,12 @@ public function canMove( User $user, $shouldLogBlockInStats = true ) {
...
-	 * @desc Creates wall message title (a board, a thread, a message) instance and calls purgeSquid() on it
+	 * @desc calls purgeSquid() on $title instance
 	 * The flow then goes to TitleGetSquidURLs hook which cleans the list of URLs in Wall and Forum
 	 */
 	public function purgeSquid() {
-		$title = Title::newFromID( $this->getId() );
-		if ( $title instanceof Title ) {
-			$title->purgeSquid();
+		if ( $this->title instanceof Title ) {
+			$this->title->purgeSquid();
```
macbre: `WallMessage::purgeSquid` is called when thread state is changed / message post is deleted, restored or undo is performed.

In all such cases invalidation of CDN cache and bump of `ETag` value (via `Title::invalidateCache`) would be nice. What if we move these two calls here?

```php
$mw->getWall()->getTitle()->invalidateCache();
wfWaitForSlaves();
```

macbre (changes requested):

extensions/wikia/Wall/WallMessage.class.php
rybmat: :+1:

macbre: And maybe rename the method to `invalidateCache`

macbre (approved):

PR #12388, opened 2017-02-09
Update UserProfilePage.scss

LunarTwilight:
Makes the avatar a little bit bigger and remove that little arrow/pointer.
Before:
349-032-before
After:
349-032-after


LunarTwilight:
https://support.wikia-inc.com/hc/en-us/requests/317196

PR #12409, opened 2017-02-14
SUS-1642 | Improve schema of wall_notification table

macbre:
https://wikia-inc.atlassian.net/browse/SUS-1642

* remove redundant `user` index (as it's already covered by `user_wiki_unique`)
* change `int(11)` column what have either 0 or 1 as a value to `tinyint`

Performed the table cleanup, removed 61 out of 67 mm of rows. Its size dropped from 14 to 0,7 GB (**95% size reduction**).

The production database has been altered.


drozdo:
👍

PR #12410, opened 2017-02-14
SUS-1648 | Category pages are displaying broken in some wikis - CSS missing

macbre:
As `toplists-loading-screen` ID is no longer used in the code, we can simply remove mixin import and ID styling.


themech (approved):

macbre:
Backported to dev in 474795c4cf3a389b9a70af736072342c0209c9cd

PR #12412, opened 2017-02-14
(MAIN-9750) Handle email changes when confirming an email

Grunny:
When changing an email, we also need to set the new email as the main
email when confirming the address.

Also, remove the duplicate ConfirmEmailComplete hook as it is called
within User::confirmEmail.


jsutterfield:
LGTM!

PR #12414, opened 2017-02-14
MAIN-9729 add protocol to discussions internal URL

jsutterfield:
Ticket: https://wikia-inc.atlassian.net/browse/MAIN-9729

This PR fixes a bug that was leading to the Discussion Recirculation module not being rendered on wikis where it's enabled.

The issue stemmed from us changing the way we were generating the URL to get content. Beforehand, we were using an external URL (eg, "http://services.wikia.com/discussion"). As part of SOC-3874, we updated to using the Consul library to get an internal address (eg, "10.1.0.2:3153"). The problem is that the Consul library doesn't prepend the protocol, it just gives us the IP and port leading to URLs like this , "10.1.0.2:3153", which were failing when we tried to make the request.

The fix is to prepend the host and port we get from the Consul library with the protocol ("http://").

### DIFF
--- a/extensions/wikia/Recirculation/services/DiscussionsDataService.class.php
+++ b/extensions/wikia/Recirculation/services/DiscussionsDataService.class.php
@@ -83,7 +83,7 @@ class DiscussionsDataService {
 		$endpoint = $this->cityId . '/threads';
 
 		$url = $this->buildUrl( $endpoint, $options );
-		$data = Http::get( $url );
+		$data = Http::get( $url, 'default', [ 'noProxy' => true ] );
 
 		return json_decode( $data, true );
 	}
@@ -108,7 +108,7 @@ class DiscussionsDataService {
 	private function getDiscussionsApiUrl() {
 		global $wgConsulServiceTag, $wgConsulUrl;
 
-		return ( new ConsulUrlProvider( $wgConsulUrl,
+		return "http://" . ( new ConsulUrlProvider( $wgConsulUrl,
 			$wgConsulServiceTag ) )->getUrl( 'discussion' );
 	}
 


WikiaTech:
SonarQube analysis reported 2 issues

1. MINOR [DiscussionsDataService.class.php#L37](https://github.com/Wikia/app/blob/ac78a3d8ec9a40e841985f11536734b17726b047/extensions/wikia/Recirculation/services/DiscussionsDataService.class.php#L37): Define a constant instead of duplicating this literal "posts" 5 times. [rule](https://sonarqube.wikia-inc.com/coding_rules#rule_key=php%3AS1192)
1. MINOR [DiscussionsDataService.class.php#L70](https://github.com/Wikia/app/blob/ac78a3d8ec9a40e841985f11536734b17726b047/extensions/wikia/Recirculation/services/DiscussionsDataService.class.php#L70): Define a constant instead of duplicating this literal "postCount" 3 times. [rule](https://sonarqube.wikia-inc.com/coding_rules#rule_key=php%3AS1192)

extensions/wikia/Recirculation/services/DiscussionsDataService.class.php
```diff
@@ -83,7 +83,7 @@ private function apiRequest() {
 		$endpoint = $this->cityId . '/threads';
 
 		$url = $this->buildUrl( $endpoint, $options );
-		$data = Http::get( $url );
+		$data = Http::get( $url, 'default', [ 'noProxy' => true ] );
```
jsutterfield, line 86: `noProxy` defaults to false, which causes the request to be proxied through `prod.border.service.consul:80`, which would then fail. So we need to set `noProxy` to `true` here.

slayful (approved):
LGTM, with a question:
We can't use https:// here, right?

jsutterfield:
That's correct, the connection is refused if we use https:
```bash
james@sandbox-s1:/usr/wikia/slot1/current/src/extensions/wikia/Recirculation/services$ curl -Isv 'https://10.8.42.49:31205/916/threads?limit=5&sortKey=trending&sortDirection=descending'
* Hostname was NOT found in DNS cache
*   Trying 10.8.42.49...
* Connected to 10.8.42.49 (10.8.42.49) port 31205 (#0)
* successfully set certificate verify locations:
*   CAfile: none
  CApath: /etc/ssl/certs
* SSLv3, TLS handshake, Client hello (1):
* error:140770FC:SSL routines:SSL23_GET_SERVER_HELLO:unknown protocol
* Closing connection 0
```

PR #12417, opened 2017-02-15
FAN-1646 Be able to distinguish /wp-content/ requests coming from Recirculation

gabrys:
We're in the middle of moving all WordPress images to Vignette. All articles now have updated image URLs, but the rate at which images are requested from WordPress dropped by only about 10%.

To confirm the hypothesis most of the remaining images are coming from Recirculation module, I want to add a query param to the requested URL so that we can filter them in Kibana.

(I tested this on sandbox-s1)


nmonterroso:
LGTM. ping @Wikia/cake

PR #12435, opened 2017-02-17
IRIS-3987 Fix logic error

garthwebb:
A clean up in SUS-1557 resulted conditional logic being inverted.  This fixes the logic.


jsutterfield:
LGTM

PR #12436, opened 2017-02-17
FAN-1665 Resize Fandom images in recirculation modules

gabrys:
Now that the Fandom images are served by Vignette we can request resized version of them instead of requesting super huge images and rescaling in the client (like it used to be).

To do so I needed to update `isThumbUrl` function to recognize the UUID Vignette URLs.

Also we don't need the `?from=recirc` URL params added to the images requested from the recirculation module so I'm removing them now.

### DIFF
--- a/resources/wikia/modules/thumbnailer.js
+++ b/resources/wikia/modules/thumbnailer.js
@@ -17,7 +17,9 @@
 		var extRegExp = /\.(jpg|jpeg|gif|bmp|png|svg)$/i,
 			imagePath = '/images/',
 			legacyThumbnailerPath = '/images/thumb/',
-			thumbnailerBaseURLRegex = /(.*\/revision\/\w+).*/;
+			// [0-9a-f-]{36} is to match UUIDs like in
+			// https://vignette.wikia.nocookie.net/ff8b7617-46fa-4efb-ac2f-ff98edf04bcf
+			thumbnailerBaseURLRegex = new RegExp('(.*/revision/\\w+|.*/[0-9a-f-]{36}).*');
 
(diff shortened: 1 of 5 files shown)


resources/wikia/modules/thumbnailer.js
```diff
@@ -17,7 +17,9 @@
 		var extRegExp = /\.(jpg|jpeg|gif|bmp|png|svg)$/i,
 			imagePath = '/images/',
 			legacyThumbnailerPath = '/images/thumb/',
-			thumbnailerBaseURLRegex = /(.*\/revision\/\w+).*/;
+			// [0-9a-f-]{36} is to match UUIDs like in
+			// https://vignette.wikia.nocookie.net/ff8b7617-46fa-4efb-ac2f-ff98edf04bcf
+			thumbnailerBaseURLRegex = new RegExp('(.*/revision/\\w+|.*/[0-9a-f-]{36}).*');
```
mklucsarits, line 22: There's a regex for valid UUID here, if you want to make this even more strict:
http://stackoverflow.com/questions/7905929/how-to-test-valid-uuid-guid#answer-13653180

mklucsarits (approved):

gbenson-wikia:
With the caveat that I haven't run the code under review, LGTM.

My only concern - and it's a very small one - is whether the change to `resources/wikia/modules/thumbnailer.js` will have unexpected side effects elsewhere by changing what gets matched by the thumbnailer to be more restrictive.  One of those "in theory it's fine, but we occasionally get bit by some other component relying on the broken behavior" things.

PR #12438, opened 2017-02-18
SUS-1649: In ApiQueryUsers exclusively use new fields for Phalanx block info

mszabo-wikia:
https://wikia-inc.atlassian.net/browse/SUS-1649


macbre (approved):

jcellary:
Per offline conversation 
`curl --proxy icache.service.poz-dev.consul:80 -H "X-Wikia-Internal-Request:true" "http://fallout.jcellary.wikia-dev.pl/api.php?action=query&format=json&list=users&ususers=SUSpect&usprop=blockinfo&cb=4ewtaasdf"`
returns both phalanx and non phalanx fields, so most likely still needs some tweaking

includes/User.php
```diff
@@ -1373,6 +1373,19 @@ private function getBlockedStatus( $bFromSlave = true, $shouldLogBlockInStats =
 	}
 
 	/**
+	 * SUS-1649: Clear all block-related info from this instance of User class
+	 * This allows us to re-run block check functions with different parameters (e.g. checking for only local blocks)
+	 * @see User::getBlockedStatus()
+	 */
+	public function clearBlockInfo() {
```
macbre: Let's add an information here that this is a custom method added by us - a simple `Wikia change` will do :)

macbre (approved):

jcellary:
Tested and confirmed working with user-permissions service!

jcellary (approved):

PR #12443, opened 2017-02-20
SUS-1521 | report a proper DB query caller in WallThread::getLastMessage

macbre:
I'll move `ForumController::boardThread` (see inline code comment) as a part of SUS-1521 refactoring. Here I just want to improve reporting of database queries.

### DIFF
--- a/extensions/wikia/Wall/WallThread.class.php
+++ b/extensions/wikia/Wall/WallThread.class.php
@@ -191,10 +191,16 @@ class WallThread {
 		WikiaDataAccess::cachePurge( $key );
 	}
 
+	/**
+	 * TODO: used by ForumController::boardThread() method only. Move it there?
+	 *
+	 * @return null|WallMessage
+	 */
 	public function getLastMessage() {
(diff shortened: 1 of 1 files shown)


extensions/wikia/Wall/WallThread.class.php
```diff
@@ -191,10 +191,16 @@ public function purgeLastMessage() {
 		WikiaDataAccess::cachePurge( $key );
 	}
 
+	/**
+	 * TODO: used by ForumController::boardThread() method only. Move it there?
```
rybmat, line 195: why not to do it now?

rybmat (approved):

extensions/wikia/Wall/WallThread.class.php
macbre: I want to avoid conflicts with big refactoring changes from @TK-999 that are in QA column.

mszabo-wikia (approved):

extensions/wikia/Wall/WallThread.class.php
rybmat: :+1:

PR #12455, opened 2017-02-22
SUS-1698: In Chat fetch user name from right rail

mszabo-wikia:
https://wikia-inc.atlassian.net/browse/SUS-1698

### DIFF
--- a/extensions/wikia/Chat2/js/views/views.js
+++ b/extensions/wikia/Chat2/js/views/views.js
@@ -424,8 +424,10 @@ var NodeChatUsers = Backbone.View.extend({
 
 		this.delegateEventsToTrigger(this.triggerEvents, function(e) {
     		e.preventDefault();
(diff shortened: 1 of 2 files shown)


extensions/wikia/Chat2/js/views/views.js
```diff
@@ -426,7 +426,7 @@ var NodeChatUsers = Backbone.View.extend({
     		e.preventDefault();
     		var name = $(e.target).closest('.UserStatsMenu').find('.username').text();
     		if(!(name.length > 0)) {
```
rybmat: can we simplify this condition? i.e  `if (name.length === 0)` or even `if (name)` should also work

rybmat (changes requested):

rybmat (approved):

macbre (approved):

PR #12471, opened 2017-02-23
SUS-1189 | Wall: error message when phalanx content filters blocked attempt to save wall message

rybmat:
https://wikia-inc.atlassian.net/browse/SUS-1189


extensions/wikia/Wall/Wall.i18n.php
```diff
@@ -107,6 +107,8 @@
 	'wall-action-monobook-login' => 'Please log in to post a message on this wiki',
 	'wall-posting-message-failed-title' => 'An error occurred while posting a message',
 	'wall-posting-message-failed-body' => 'Copy your work and try again later',
+	'wall-posting-message-failed-filter-title' => 'Inapropriate content detected', //TODO: set proper message
```
rybmat: What should be the message (and the one below)? Or who should be pinged about this?

rybmat: discussion in https://wikia-inc.atlassian.net/browse/SUS-1718

extensions/wikia/Wall/builders/WallMessageBuilder.class.php
```diff
@@ -94,11 +94,19 @@ public function postNewMessage(): WallMessageBuilder {
 			// Permissions check are performed on article comment level by EditPage class of Mediawiki
 			// Text is matched there against Phalanx filters and all user blocks (global and local) are also checked
 			// This can cause edit to fail
-			$this->throwException( 'Failed to create article comment' );
+			$reason =  $result && (
```
mszabo-wikia: Let's consider creating a different exception class for edits that were prevented by Phalanx content block / user block and handling it in controller. This would allow us to separate handling of failed edit caused by block from failed edit caused by DB issues or invalid input, and this way we won't log every blocked edit attempt to Kibana as error.

mszabo-wikia (changes requested):

extensions/wikia/Wall/builders/WallMessageBuilder.class.php
rybmat: yeah, that is actually good idea. Thanks!

macbre:
LGTM - the exception comment from @TK-999

extensions/wikia/Wall/WallExternalController.class.php
```diff
@@ -278,6 +277,13 @@ public function postNewMessage() {
 		} catch ( WallBuilderException $builderException ) {
 			\Wikia\Logger\WikiaLogger::instance()->error( $builderException->getMessage(), $builderException->getContext() );
 			$this->response->setVal( 'status', false );
+			$this->response->setVal( 'reason', 'other' );
+			$this->response->setCode( WikiaResponse::RESPONSE_CODE_INTERNAL_SERVER_ERROR );
+			return;
+		} catch ( InappropriateContentException $exception) {
+			\Wikia\Logger\WikiaLogger::instance()->error( $exception->getMessage(), $exception->getContext() );
```
mszabo-wikia: Let's not log this as error in Kibana, since it'd be picked up by ER

extensions/wikia/Wall/WallExternalController.class.php
```diff
@@ -278,6 +277,13 @@ public function postNewMessage() {
...
+			$this->response->setVal( 'status', false );
+			$this->response->setVal( 'reason', 'badcontent' );
 			$this->response->setCode( WikiaResponse::RESPONSE_CODE_INTERNAL_SERVER_ERROR );
```
mszabo-wikia: Let's use a different response code here, to indicate user was prevented from performing the action deliberately

mszabo-wikia (changes requested):

extensions/wikia/Wall/WallExternalController.class.php
rybmat: Which one would be appropriate in your opinion? 403?

mszabo-wikia (approved):

macbre (approved):

PR #12424, opened 2017-02-16
SUS-1681: Wall create/edit flow refactor

mszabo-wikia:
## Rationale
Wall has several monolithic classes with a multitude of responsibilities, such as `CommentsIndex`, which manages the internal state of `comments_index` table while representing also individual rows of the same table, or `WallMessage`, which handles translation between ArticleComment layer and incoming Nirvana request, does comment index updates and miscellaneous tasks. It may be worth considering extracting some of this functionality elsewhere from base classes.

## Changes implemented
- Introduced class `CommentsIndexEntry` whose instances represent individual rows in `CommentsIndex` table.
- `CommentsIndex` class is now only responsible for managing internal state of `comments_index` table. It can produce `CommentsIndexEntry` instances or insert/update them and maintains in-object cache for entries.
- Moved `CommentsIndex` from ArticleComments extension to Wall because it is only used by Wall/Forum.
- Introduced `WallMessageBuilder` class to handle creating new thread or reply. Class uses builder pattern to produce instance of `WallMessage` that belongs to the newly created message. It replaces `WallMessage::buildNewMessageAndPost`.
- Introduced `WallEditBuilder` class to handle editing thread or reply. Class uses builder pattern and produces parsed wikitext as `string`.

## Moving on
- extract all functionality which manages internal state of article comment from `WallMessage` to builders or elsewhere (ie handle move/delete updates)
- review Wall Permissions checking and simplify
- extract move/delete/remove functionality (possibly as part of https://wikia-inc.atlassian.net/browse/SUS-1521)

Báthory

PR #12472, opened 2017-02-23
SUS-1719 | bring back the call to ArticleComment::newFromId that adds small delay allowing slaves to catch up

macbre:
https://wikia-inc.atlassian.net/browse/SUS-1719

Yes, it is an ugly hack (backport of code removed as a part of #12424) and we first want to check whether it solves the issue on sandbox.

**TODO**: consider improving handling of slave lag - `wfWaitForSlaves` / update the cache instead of invalidating it...

### DIFF
--- a/extensions/wikia/Wall/WallMessage.class.php
+++ b/extensions/wikia/Wall/WallMessage.class.php
@@ -1511,7 +1511,7 @@ class WallMessage {
 	}
 
 	protected function getPropCacheKey() {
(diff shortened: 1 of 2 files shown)


TK-999:
@macbre I think we should call `WallMessage::invalidateCache` on thread post or reply post which calls `wfWaitForSlaves` and invalidates thread url, wall url etc

rybmat:
see: https://github.com/Wikia/app/pull/12474

PR #12478, opened 2017-02-24
SUS-1706 | override float for .mw-collapsible-toggle on wall

rybmat:
https://wikia-inc.atlassian.net/browse/SUS-1706

### DIFF
--- a/extensions/wikia/Wall/css/Wall.scss
+++ b/extensions/wikia/Wall/css/Wall.scss
@@ -25,6 +25,12 @@ html>body #Wall .new-reply .speech-bubble-avatar { *position: absolute; *left:-9
 }
 
 .Wall {
+
+	/* overrides /resources/jquery/jquery.makeCollapsible.css */
+	li .mw-collapsible-toggle {
+		float: right;
+	}
+
 	&.Board {
 		.deleteorremove-infobox {
 			margin-left: 75px;


extensions/wikia/Wall/css/Wall.scss
```diff
@@ -25,6 +25,12 @@ html>body #Wall .new-reply .speech-bubble-avatar { *position: absolute; *left:-9
 }
 
 .Wall {
+
+	/* overrides /resources/jquery/jquery.makeCollapsible.css */
+	li .mw-collapsible-toggle {
+		float: right;
```
mszabo-wikia, line 31: Is this setting correct for RTL languages as well?

E.g: http://ko.pokemon.wikia.com/wiki/%EA%B2%8C%EC%8B%9C%EA%B8%80:69751?usesitecss=0&uselang=ar

mszabo-wikia (changes requested):

extensions/wikia/Wall/css/Wall.scss
rybmat: I didn't find anything in the app repo that would set it to other value than right or none, so I guess it is correct

mszabo-wikia (approved):

macbre (approved):

PR #12501, opened 2017-02-28
MAIN-9881: Upgraded PHPUnit to 6.0.*

mszabo-wikia:
https://wikia-inc.atlassian.net/browse/MAIN-9881


extensions/wikia/ScavengerHunt/tests/ScavengerHuntTest.php
```diff
@@ -1,5 +1,5 @@
 <?php
-class ScavengerHuntTest extends WikiaBaseTest {
+abstract class ScavengerHuntTest extends WikiaBaseTest {
```
macbre, line 2: Worth checking if we actually have ` ScavengerHunt` enabled anywhere.

lib/Wikia/tests/Logger/WikiaLoggerTest.php
```diff
@@ -7,7 +7,7 @@
 use Wikia\Logger\SyslogHandler;
 use Wikia\Logger\WebProcessor;
 
-class WikiaLoggerTest extends PHPUnit_Framework_TestCase {
+class WikiaLoggerTest extends \PHPUnit\Framework\TestCase {
```
macbre: Let's be consistent when extending `TestCase` class, either provide the full "path" with namespaces or use `use`.

extensions/wikia/ScavengerHunt/tests/ScavengerHuntTest.php
mszabo-wikia: Yeah. All tests were marked as `Broken` 4 years ago and the extension might be disabled now.

macbre (changes requested):

extensions/wikia/ScavengerHunt/tests/ScavengerHuntTest.php
macbre: https://wikia-inc.atlassian.net/browse/SUS-1730

macbre:
:+1: let's make the `TestCase` use consistent and we're good :)

macbre (approved):

macbre:
@TK-999, worth checking why `unitTests_php_branches` Jenkins task did not run (or at least did not report the results here).

macbre:
:+1:  if tests do pass now, let's simply merge it and announce the good news

Thanks, @TK-999  :)

PR #12504, opened 2017-03-01
SUS-723 | comments index cleanup

rybmat:
Script for updating "deleted" flag in comments_index and also for removing entries from this table that has other namespace than supported by Wall/Forum associated in page table

### DIFF
--- a/extensions/wikia/Wall/maintenance/CleanupCommentsIndex.php
+++ b/extensions/wikia/Wall/maintenance/CleanupCommentsIndex.php
@@ -18,7 +18,7 @@ class CleanupCommentsIndex extends Maintenance {
 		$updated = $this->updateDeletedFlag( $dbMaster, $dbSlave );
 		$deleted = $this->removeBrokenRows( $dbMaster, $dbSlave );
 
-		$this->output( "db:{$wgDBname} before:{$deleted['count']} updated:{$updated} deleted:{$deleted['deleted']}\n" );
+		$this->output( "db: {$wgDBname} before: {$deleted['count']} updated: {$updated} deleted: {$deleted['deleted']}\n" );
 	}
 
 	public function updateDeletedFlag( DatabaseMysqli $dbMaster, DatabaseMysqli $dbSlave ) {
--- a/extensions/wikia/Wall/maintenance/CleanupCommentsIndex.php
+++ b/extensions/wikia/Wall/maintenance/CleanupCommentsIndex.php
@@ -32,7 +32,7 @@ class CleanupCommentsIndex extends Maintenance {
 		if ( is_array( $ids ) ) {
 			$ids = array_map( function( $item ) {
 				return intval( $item );
-			}, $ids);
+			}, $ids );
 
 			$this->output( "{$wgDBname}: " . count( $ids ) . " rows to update\n" );
 
@@ -60,13 +60,7 @@ class CleanupCommentsIndex extends Maintenance {
 		if ( is_array( $ids ) ) {
 			$ids = array_map( function( $item ) {
 				return intval( $item );
-			}, $ids);
-
-			//TODO: remove it
-			\Wikia\Logger\WikiaLogger::instance()->info(
-				"{$wgDBname}: " . count( $ids ) . " rows to remove",
-				[ "rows_count" => count( $ids ) ]
-			);
+			}, $ids );
 
 			$this->output( "{$wgDBname}: " . count( $ids ) . " rows to remove\n" );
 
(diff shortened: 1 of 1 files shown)


extensions/wikia/Wall/maintenance/CleanupCommentsIndex.php
```diff
@@ -10,31 +10,66 @@ public function __construct() {
...
+	}
+
+	public function removeBrokenRows( DatabaseMysqli $dbMaster, DatabaseMysqli $dbSlave ) {
+		global $wgDBname;
+
+		$ids = $dbSlave->selectFieldValues(
 			[ 'page', 'comments_index' ],
 			'comment_id',
-			[ 'page_id=comment_id', 'page_namespace not in (500,501,1200,1201,2000,2001)' ],
+			[ 'page_id=comment_id', 'archived=0', 'deleted=0', 'removed=0', 'page_namespace not in (500,501,1200,1201,2000,2001)' ],
```
ghost: I would suggest using named constants for namespaces and `makeList()` to create a list for SQL `IN()` clause?

extensions/wikia/Wall/maintenance/CleanupCommentsIndex.php
```diff
@@ -10,31 +10,66 @@ public function __construct() {
...
+			}, $ids );
 
-			$this->output( "{$wgDBname}: " . count( $ids ) . " rows affected\n" );
+			$this->output( "{$wgDBname}: " . count( $ids ) . " rows to remove\n" );
 
 			if ( !$this->hasOption('dry-run') && !empty( $ids ) ) {
-				$db->delete(
+				$dbMaster->delete(
 					'comments_index',
 					[ "comment_id" => $ids ]
```
ghost: Please, add `__METHOD__` here. This deletes data which is logged in a more detailed manner. The source of the query is an important bit of information and helps to distinguish controlled deletion from unwanted incidents.

extensions/wikia/Wall/maintenance/CleanupCommentsIndex.php
mszabo-wikia: Is there a scenario where a page not in allowed namespace might be legitimately in `comments_index` table as archived/removed/deleted item?

ghost (changes requested):

extensions/wikia/Wall/maintenance/CleanupCommentsIndex.php
```diff
@@ -10,31 +10,66 @@ public function __construct() {
...
+	}
+
+	public function updateDeletedFlag( DatabaseMysqli $dbMaster, DatabaseMysqli $dbSlave ) {
 		global $wgDBname;
 
-		$db = wfGetDB( DB_MASTER );
+		$ids = $dbSlave->selectFieldValues(
+			[ 'archive', 'comments_index' ],
+			'comment_id',
+			[ 'archived=0', 'deleted=0', 'removed=0', 'comment_id=ar_page_id', 'ar_namespace in (500,501,1200,1201,2000,2001)', 'comment_id not in (SELECT page_id from page)' ],
```
mszabo-wikia: I think it's enough to check `deleted` flag here. The other two aren't really delete options (thread closing/removal).

mszabo-wikia (changes requested):

extensions/wikia/Wall/maintenance/CleanupCommentsIndex.php
rybmat: I didn't want to update by mistake rows that has one of those flags already set, so if it does not break anything I'd like to leave these checks as they are now

extensions/wikia/Wall/maintenance/CleanupCommentsIndex.php
rybmat: Yes it is. Just remove wall message with the delete flow that sets one of these flags and moves entry from page table to archive table (don't remember which flow it is) right after this create new article/upload file and Boom! New entry in page table have the same id as just removed wall message, but different namespace

rybmat: (auto increment on page_id works like max(page_id)+1 )

macbre (approved):

extensions/wikia/Wall/maintenance/CleanupCommentsIndex.php
```diff
@@ -18,7 +18,7 @@ public function execute() {
 		$updated = $this->updateDeletedFlag( $dbMaster, $dbSlave );
 		$deleted = $this->removeBrokenRows( $dbMaster, $dbSlave );
 
-		$this->output( "db:{$wgDBname} before:{$deleted['count']} updated:{$updated} deleted:{$deleted['deleted']}\n" );
+		$this->output( "db: {$wgDBname} before: {$deleted['count']} updated: {$updated} deleted: {$deleted['deleted']}\n" );
```
macbre: :+1:

PR #12508, opened 2017-03-01
SUS-1507: Use "A Fandom User" as image alt text for anon avatars

mszabo-wikia:
https://wikia-inc.atlassian.net/browse/SUS-1507

### DIFF
--- a/includes/wikia/services/AvatarService.class.php
+++ b/includes/wikia/services/AvatarService.class.php
@@ -181,7 +181,7 @@ class AvatarService extends Service {
 			'width' => $avatarSize,
 			'height' => $avatarSize,
 			'class' => 'avatar',
-			'alt' => IP::isIPAddress( $userName ) ? wfMessage( 'oasis-anon-user' )->inContentLanguage()->text() : $userName,
+			'alt' => IP::isIPAddress( $userName ) ? wfMessage( 'oasis-anon-user' )->text() : $userName,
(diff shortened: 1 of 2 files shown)


includes/wikia/services/AvatarService.class.php
```diff
@@ -178,7 +181,7 @@ static function renderAvatar( $userName, $avatarSize = 20 ) {
 			'width' => $avatarSize,
 			'height' => $avatarSize,
 			'class' => 'avatar',
-			'alt' => $userName,
+			'alt' => IP::isIPAddress( $userName ) ? wfMessage( 'oasis-anon-user' )->inContentLanguage()->text() : $userName,
```
rybmat: I'm not sure if it should be content language or maybe user language?

TK-999: For anonymous users user language = content language anyways, so this way we spare `wgLang` calculation since content language is already known.

mszabo-wikia: Yeah, interesting question. I think it's better to go with user language.

rybmat (approved):

macbre (approved):

ghost (approved):

ghost:
Tested and accepted.

mixth-sense:
Tested and accepted.

PR #12517, opened 2017-03-03
XW-2864 abtest autoplay

idradm:
https://wikia-inc.atlassian.net/browse/XW-2864


idradm:
Merge after https://github.com/Wikia/app/pull/12518 is done.

extensions/wikia/ArticleVideo/scripts/article-video.js
```diff
@@ -166,6 +169,27 @@ require(['wikia.window', 'wikia.onScroll', 'wikia.tracker', 'ooyala-player'], fu
 				}
 			});
 
+			player.mb.subscribe( OO.EVENTS.PLAYHEAD_TIME_CHANGED, 'featured-video', function(eventName, time, totalTime) {
+				var secondsPlayed = Math.floor(time),
+					percentage = Math.round(time / totalTime * 100);
+
+				if ( secondsPlayed % 5 === 0 && secondsPlayed !== playTime ) {
```
latata, line 186: Are we sure we will get at least one event every second?

latata, line 186: In the proposal of drop off tracking I proposed to stop sending tracking events when user uses seek bar (look at examples here https://wikia-inc.atlassian.net/wiki/display/PREMIUM/Videos+drop-offs+tracking)

idradm: Under development the event occurred 3 to 5 times per second, that's why it had to be throttled.

idradm: I'd say the more data we have the better. I don't see any value in stopping the events flow.

latata (approved):

PR #12527, opened 2017-03-06
IRIS-4090 | Escape characters for a tag parameter

slayful:
https://wikia-inc.atlassian.net/browse/IRIS-4090

### DIFF
--- a/extensions/wikia/MyHome/ActivityFeedTag.php
+++ b/extensions/wikia/MyHome/ActivityFeedTag.php
@@ -25,20 +25,19 @@ function ActivityFeedTag_render($content, $attributes, $parser, $frame) {
 	$parameters = ActivityFeedHelper::parseParameters($attributes);
 
 	$element =
-		renderElement( ActivityFeedHelper::getList( $parameters ), getStyle( $parameters ),
+		renderElement( ActivityFeedHelper::getList( $parameters ), getAttributes( $parameters ),
 			getSnippets( $parameters ) );
 	wfProfileOut( __METHOD__ );
 
 	return $element;
 }
 
-function renderElement($content, $style, $jsSnippets) {
-	return Html::element( 'div', [ 'style' => $style ], $content ) . $jsSnippets;
+function renderElement($content, $attributes, $jsSnippets) {
+	return Html::element( 'div', $attributes, $content ) . $jsSnippets;
 }
 
-function getStyle( $parameters ) {
-	$style = empty($parameters['style']) ? '' : $parameters['style'];
-	return sprintf(' style="%s"',  $style);
+function getAttributes( $parameters ) {
+	return empty($parameters['style']) ? [] : ['style' => $parameters['style']];
 }
 
 function getSnippets( &$parameters ) {
--- a/extensions/wikia/MyHome/ActivityFeedTag.php
+++ b/extensions/wikia/MyHome/ActivityFeedTag.php
@@ -23,13 +23,17 @@ function ActivityFeedTag_render($content, $attributes, $parser, $frame) {
 	wfProfileIn(__METHOD__);
 
 	$parameters = ActivityFeedHelper::parseParameters($attributes);
-	$feedHTML = ActivityFeedHelper::getList($parameters);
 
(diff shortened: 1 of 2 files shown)


extensions/wikia/MyHome/ActivityFeedTag.php
```diff
@@ -53,6 +51,13 @@ function ActivityFeedTag_render($content, $attributes, $parser, $frame) {
...
 
+	$style = getStyle($parameters);
+
 	wfProfileOut(__METHOD__);
 	return "<div$style>$feedHTML</div>$snippets";
 }
+
+function getStyle( $parameters ) {
+	$style = $parameters['style'];
+	return empty($style) ? '' : ' style="' . $style . '"';
```
ghost: You have changed `empty($parameters['style'])` to less safe `$style = $parameters['style'];` and `empty($style)`. The key `style` is not something mandatory. The `$parameters` array might simply not have that key in which case the assignment will result with a PHP Warning. And in this case, the warning is not something that needs consideration.

extensions/wikia/MyHome/ActivityFeedTag.php
```diff
@@ -53,6 +51,13 @@ function ActivityFeedTag_render($content, $attributes, $parser, $frame) {
 		)
 	);
 
+	$style = getStyle($parameters);
+
 	wfProfileOut(__METHOD__);
 	return "<div$style>$feedHTML</div>$snippets";
```
ghost: Personally, I am not a fan of such interpolations. I would sacrifice performance a bit and use `sprintf()` and use it consistently elsewhere (e.g. in the `getStyle()` method, where string concatenation with `.` is used.

ghost (changes requested):
I imagine, using `htmlspecialchars` changes the behaviour and addresses some issue. Perhaps it's worth to add a test to make sure the contract is kept and special characters are escaped properly after future changes?

mszabo-wikia (changes requested):
@mixth-sense @slayful Let's use `Html::element` helper function here. It guarantees escaping of all attributes provided as array - we can pass empty array, or `[ 'style' => $parameters['style'] ]` depending on whether style attribute was provided.

`htmlspecialchars` is not safe for attribute escaping unless used with `ENT_QUOTES`.

extensions/wikia/MyHome/ActivityFeedTag.php
```diff
@@ -25,34 +24,48 @@ function ActivityFeedTag_render($content, $attributes, $parser, $frame) {
...
+	$element =
+		renderElement( ActivityFeedHelper::getList( $parameters ), getStyle( $parameters ),
+			getSnippets( $parameters ) );
+	wfProfileOut( __METHOD__ );
+
+	return $element;
+}
+
+function renderElement($content, $style, $jsSnippets) {
+	return Html::element( 'div', [ 'style' => $style ], $content ) . $jsSnippets;
```
slayful: I don't think it's appropriate to use sprintf here (it would look like `sprintf('%s%s', $first, $second);`).

slayful:
Thanks for the feedback @TK-999, @mixth-sense. Could you review once again?
I've extracted the code into separate methods, to diff will probably look weird, sorry about that.

extensions/wikia/MyHome/ActivityFeedTag.php
ghost: `Html::element()` is surely the best approach

extensions/wikia/MyHome/ActivityFeedTag.php
```diff
@@ -25,34 +24,48 @@ function ActivityFeedTag_render($content, $attributes, $parser, $frame) {
...
+}
+
+function getStyle( $parameters ) {
+	$style = empty($parameters['style']) ? '' : $parameters['style'];
+	return sprintf(' style="%s"',  $style);
```
mszabo-wikia: Since we're using a helper, we no longer need to `sprintf` the value

mszabo-wikia (changes requested):

mszabo-wikia (approved):

ghost (approved):

mszabo-wikia:
Merged to dev per @tortila QA on #12531

PR #12535, opened 2017-03-07
IRIS-4091

slayful:
* Addressed Igor's feedback
* Added mustache
* Fetched translations from Design System I18N
* Re-ordered the icons
* Added the mark-all-as-read button


extensions/wikia/DesignSystem/i18n/de.json
```diff
@@ -72,5 +72,22 @@
   "global-navigation-wikis-community-central": "Community Deutschland",
   "global-navigation-wikis-explore": "Wikis erforschen",
   "global-navigation-wikis-fandom-university": "Fandom Universität",
-  "global-navigation-wikis-header": "Wikis"
+  "global-navigation-wikis-header": "Wikis",
+  "notifications-no-notifications-message": "Noch keine Benachrichtigungen.",
+  "notifications-mark-all-as-read": "Alle als gelesen markieren",
+  "notifications-replied-by-multiple-users-with-title": "__mostRecentUser__ und __number__ weitere Benutzer haben auf __postTitle__ <b>geantwortet</b>",
```
rogatty, line 78: You need to update https://github.com/Wikia/app/blob/dev/extensions/wikia/DesignSystem/i18n/build.js#L6-L19 as this format won't work in MediaWiki. It should be:
```
$1 und $2 weitere Benutzer haben auf $3 <b>geantwortet</b>
```
See https://wikia-inc.atlassian.net/wiki/display/DESYS/RFC%3A+Storing+translations for the background.

extensions/wikia/AssetsManager/config.php
```diff
@@ -2211,14 +2211,20 @@
...
 	'skin' => [ 'oasis' ],
 	'assets' => [
 		'//extensions/wikia/DesignSystem/scripts/DesignSystemGlobalNavigationOnSiteNotifications.js'
 	],
 ];
 
+$config['design_system_on_site_notifications_mustache_js'] = [
+	'type' => AssetsManager::TYPE_JS,
+	'skin' => [ 'oasis' ],
+	'assets' => [],
```
rogatty: If you're going to implement this incrementally IMO it would be better if you skipped the placeholders "for the future". I guess you will start using it at some point but if I don't know for sure, then it's hard to do a proper CR.

It makes me want to say "remove it, it's not being used" :wink:

extensions/wikia/DesignSystem/i18n/ko.json
```diff
@@ -0,0 +1,19 @@
+{
+  "notifications-no-notifications-message": "No notifications yet.",
```
rogatty, line 2: Seems like we shouldn't support a few of these languages if they don't have any existing translations, right?

extensions/wikia/DesignSystem/scripts/DesignSystemGlobalNavigationOnSiteNotifications.js
```diff
@@ -1,42 +1,141 @@
...
-				this.$notificationsCount = $('.on-site-notifications-count');
+				this.$notificationsCount = $('#on-site-notifications-count');
+				this.$container = $('#on-site-notifications');
+
+				this.addDropdownLoadingEvent();
+				this.addMarkAllAsReadEvent();
+			},
+
+			addMarkAllAsReadEvent: function () {
+				this.$markAllAsReadButton = $('#mark-all-as-read-button');
```
rogatty: It's used in a few places so I think it'd be safer to define it in `init()`. It should be fine now but if someone starts moving things around, it'll be easy to introduce a bug.

extensions/wikia/DesignSystem/scripts/DesignSystemGlobalNavigationOnSiteNotifications.js
```diff
@@ -1,42 +1,141 @@
...
+				this.$markAllAsReadButton.click(function () {
+					OnSiteNotifications.markAllAsRead();
```
rogatty: `this.$markAllAsReadButton.click(OnSiteNotifications.markAllAsRead)` would work, right?

extensions/wikia/DesignSystem/scripts/DesignSystemGlobalNavigationOnSiteNotifications.js
```diff
@@ -1,42 +1,141 @@
...
+			addDropdownLoadingEvent: function () {
+				var $dropdown = $('#on-site-notifications-dropdown');
+				$dropdown.click(function () {
+					OnSiteNotifications.loadFirstPage();
+				});
 			},
 
-			updateCounts: function () {
+			shouldLoadFirstPage: function () {
+				return !this.updateInProgress && this.nextPage && this.allPagesLoaded !== true;
```
rogatty: Both `updateInProgress` and `allPagesLoaded` are boolean. Let's stick to a single convention for evaluating them.

extensions/wikia/DesignSystem/scripts/DesignSystemGlobalNavigationOnSiteNotifications.js
```diff
@@ -1,42 +1,141 @@
...
+			},
+
+			shouldLoadNextPage: function () {
```
rogatty: It's not used anywhere.

extensions/wikia/DesignSystem/scripts/DesignSystemGlobalNavigationOnSiteNotifications.js
```diff
@@ -1,42 +1,141 @@
...
+					this.renderUnreadCount(0);
+					//TODO clear read status from all loaded notifications
+				})).always(this.proxy(function () {
+					this.bucky.timer.stop('markAllAsRead');
+				}));
+			},
+
+			loadFirstPage: function () {
+				if (this.shouldLoadFirstPage()) {
+					return;
```
rogatty: If it should load the first page then we don't load the first page? Doesn't seem right.

extensions/wikia/DesignSystem/scripts/DesignSystemGlobalNavigationOnSiteNotifications.js
```diff
@@ -1,42 +1,141 @@
...
+						// latestActors: NotificationModel.createActors(x.events.latestActors')),
+						// type: NotificationModel.getTypeFromApiData(notificationData)
+					};
+				});
+			},
+
+			renderNotifications: function (notifications) {
+				notifications.forEach(this.proxy(function (notification) {
+					var html = mustache.render(this.template, notification);
+					this.$container.append(html);
```
rogatty: I think it would be better to render all notifications and then append them to DOM at once. Also, what do you think about putting this loop inside the template and passing the array there? Then, the Mustache engine wouldn't need to be run multiple times.

extensions/wikia/DesignSystem/services/templates/DesignSystemGlobalNavigationOnSiteNotifications_index.php
```diff
@@ -1,19 +1,25 @@
-<div id="onSiteNotificationsEntryPoint"
-     class="wds-global-navigation__notifications-menu wds-global-navigation__notifications-menu wds-dropdown notifications-container">
-	<div class="wds-dropdown__toggle wds-global-navigation__dropdown-toggle"
+<div class="wds-global-navigation__notifications-menu wds-dropdown notifications-container">
+	<div id="on-site-notifications-dropdown"
+	     class="wds-dropdown__toggle wds-global-navigation__dropdown-toggle"
 	     title="<?= wfMessage( 'global-navigation-notifications-title' )->escaped() ?>">
 		<div class="bubbles">
```
rogatty, line 9: Do you plan to get rid of this class and `notifications-container` before finishing the work? I don't think we should stick to the legacy code if not necessary.

extensions/wikia/DesignSystem/scripts/DesignSystemGlobalNavigationOnSiteNotifications.js
```diff
@@ -45,14 +144,33 @@ require(
...
+				}
+			}).done(function (assets) {
+				loader.processScript(assets.scripts);
+				OnSiteNotifications.init(assets.mustache[0]);
+			});
+		}
+
+		function getSafely(obj, path) {
+			return path.split(".").reduce(function (acc, key) {
+				return (typeof acc == "undefined" || acc === null) ? acc : acc[key];
```
rogatty: What if `typeof acc === 'string'`? Shouldn't we return the `acc` too, then? How about checking if it's an array or object?

extensions/wikia/DesignSystem/services/templates/DesignSystemGlobalNavigationOnSiteNotifications_index.php
```diff
@@ -1,19 +1,25 @@
...
-		<ul id="GlobalNavigationOnSiteNotifications"
-		    class="OnSiteNotifications global-nav-dropdown"></ul>
+	<div
+		class="wds-dropdown__content wds-is-right-aligned wds-global-navigation__dropdown-content wds-notifications__dropdown-content">
+		<div id="mark-all-as-read-button" class="wds-notifications__mark-all-as-read-button">
+			<a class="wds-notifications__mark-all-as-read">
+				<?= wfMessage( 'notifications-mark-all-as-read' )->escaped() ?>
+			</a>
+		</div>
+		<div class="notification-list" id="on-site-notifications">
```
rogatty: Let's stick to the `wds-` class naming convention.

extensions/wikia/DesignSystem/styles/design-system.scss
```diff
@@ -91,3 +91,53 @@ $global-navigation-z-index: 5000101;
...
+.wds-notifications {
+  &__mark-all-as-read-button {
+	box-shadow: 0 -1px 10px 0 rgba(0, 0, 0, 0.1);
+	display: none;
+	justify-content: flex-start;
+	min-height: 44px;
+	padding: 0;
+	width: 100%;
+
+	&.show {
```
rogatty: How about `.wds-is-visible`?

extensions/wikia/DesignSystem/styles/design-system.scss
```diff
@@ -91,3 +91,53 @@ $global-navigation-z-index: 5000101;
 		}
 	}
 }
+
+
+
+// on-site-notifications to be moved to Design System
```
rogatty: Until it's moved, I'd put it in a separate file and import.

extensions/wikia/AssetsManager/config.php
slayful: So this is not a placeholder, rather than that, the JS gets loaded by the mustache loader. I would have included this in the `design_system_on_site_notifications` but it causes it to be reloaded and therefore re-initialised again and again (because this in turn loads mustache again).

extensions/wikia/DesignSystem/services/templates/DesignSystemGlobalNavigationOnSiteNotifications_index.php
slayful: https://wikia-inc.atlassian.net/browse/IRIS-4117

extensions/wikia/DesignSystem/i18n/de.json
slayful: Oh my this is going to be painful. Will address this as I move on to the notification titles.

extensions/wikia/DesignSystem/i18n/ko.json
slayful: https://wikia-inc.atlassian.net/browse/IRIS-4118

extensions/wikia/DesignSystem/scripts/DesignSystemGlobalNavigationOnSiteNotifications.js
slayful: Oh it's going to be, trust me I'm doing this the second time, just differently. =)

extensions/wikia/DesignSystem/i18n/de.json
slayful: https://wikia-inc.atlassian.net/browse/IRIS-4107

extensions/wikia/AssetsManager/config.php
rogatty: But there are no scripts here to load (so far?), so what is it for?

extensions/wikia/DesignSystem/scripts/DesignSystemGlobalNavigationOnSiteNotifications.js
slayful: This is far from a perfect solution but it works for chains where at some point the object may be either undefined or null.
The last object in the chain does not have to be an object, so ignoring everything but objects will fail the last step.

rogatty (approved):

slayful:
Closing because https://github.com/Wikia/app/pull/12573/files is more up to date.

PR #12536, opened 2017-03-07
SUS-1768 | report exceptions in MWExceptionHandler as the first thing

macbre:
https://wikia-inc.atlassian.net/browse/SUS-1768

Let's move ELK logging of exceptions handled by `MWExceptionHandler::report` to the top of this method. Jira Reporter change will follow to improve normalization of such exceptions so that we can report all of them properly grouped.

@TK-999

### DIFF
--- a/includes/Exception.php
+++ b/includes/Exception.php
@@ -562,13 +562,14 @@ class MWExceptionHandler {
 
 		$cmdLine = MWException::isCommandLine();
 
+		# Wikia change - begin
+		# @see PLATFORM-2008 - report non-MediawWiki exceptions to ELK
+		Wikia\Logger\WikiaLogger::instance()->error( __METHOD__, [
+			'exception' => $e,
+		] );
+		# Wikia change - end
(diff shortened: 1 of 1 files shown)


includes/Exception.php
```diff
@@ -609,13 +610,6 @@ protected static function report( $e ) {
 			} else {
 				self::escapeEchoAndDie( $message );
 			}
-
-			# Wikia change - begin
-			# @see PLATFORM-2008 - report non-MediawWiki exceptions to ELK
-			Wikia\Logger\WikiaLogger::instance()->error( __METHOD__ . ' - unexpected non-MediaWiki exception encountered', [
-				'exception' => $e,
-			] );
-			# Wikia change - end
```
macbre, line 618: `self::escapeEchoAndDie` dies internally when run on production - hence we got no logs for non-MediaWiki exceptions...

mszabo-wikia (approved):

PR #12537, opened 2017-03-07
XW-2868 | MercuryApi - handling interwiki links

rybmat:
https://wikia-inc.atlassian.net/browse/XW-2868

@Wikia/x-wing

### DIFF
--- a/extensions/wikia/InterwikiDispatcher/SpecialInterwikiDispatcher_body.php
+++ b/extensions/wikia/InterwikiDispatcher/SpecialInterwikiDispatcher_body.php
@@ -23,6 +23,8 @@ if (!defined('MEDIAWIKI')) {
 
 class InterwikiDispatcher extends UnlistedSpecialPage {
 	const IS_WIKI_EXISTS_CACHE_TTL = 10800;
+	const SUPPORTED_IW_PREFIXES = [ 'w', 'wikia', 'wikicities' ];
+
 	/**
 	 * contructor
 	 */
@@ -102,7 +104,7 @@ class InterwikiDispatcher extends UnlistedSpecialPage {
 	public static function getInterWikiaURL(Title &$title, &$url, $query) {
 		global $wgArticlePath, $wgScriptPath;
 
-		if (in_array($title->mInterwiki, array('w', 'wikia', 'wikicities'))) {
+		if (in_array($title->mInterwiki, self::SUPPORTED_IW_PREFIXES)) {
 			$aLinkParts = explode(':', $title->getFullText());
 			if ($aLinkParts[1] == 'c') {
 				$iCityId = self::isWikiExists($aLinkParts[2]);
--- a/extensions/wikia/InterwikiDispatcher/SpecialInterwikiDispatcher_body.php
+++ b/extensions/wikia/InterwikiDispatcher/SpecialInterwikiDispatcher_body.php
@@ -183,7 +183,7 @@ class InterwikiDispatcher extends UnlistedSpecialPage {
 	 *
 	 * @return bool
 	 */
-	public static function getInterWikiaURLhook( Title &$title, &$url, $query ) {
+	public static function getInterWikiaURLHook( Title &$title, &$url, $query ) {
 		$interwikiUrl = self::getInterWikiaURL( $title );
 		$url = empty( $interwikiUrl ) ? $url : $interwikiUrl;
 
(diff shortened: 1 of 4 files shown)


extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -324,12 +324,23 @@ public function getPage() {
...
+			if ( empty( $interWikiUrl ) && $this->isSupportedByMercury( $title ) ) {
+				// Empty category pages are not known but contain article list;
 				if ( !$title->isKnown() && $title->getNamespace() !== NS_CATEGORY ) {
 					throw new NotFoundApiException( 'Page doesn\'t exist' );
 				}
 
+				// InterwikiDispatcher::getInterWikiaURL does not support other prefixes than InterwikiDispatcher::SUPPORTED_IW_PREFIXES
+				// but other prefixes may be defined in `interwiki` table for given wiki - in such case $title->isKnown()
+				// returns true in previous `if` statement
+				if ( !empty( $title->mInterwiki ) && !in_array($title->mInterwiki, InterwikiDispatcher::SUPPORTED_IW_PREFIXES ) ) {
```
hakubo: I think on $title there is a method isExternal()

extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -424,6 +435,13 @@ public function getPage() {
 			);
 		}
 
+		// if $interwikiUrl is not empty it means that we should redirect to other wiki (follow interwiki link)
+		if ( !empty( $interWikiUrl ) ) {
+			$data = [
+				'redirectTo' => $interWikiUrl,
```
hakubo, line 440: shouldn't that be a redirect-to?

extensions/wikia/MercuryApi/MercuryApiController.class.php
hakubo: What do you think about InterwikiDispatcher::isSupportedPrefix($prefix)?
and adding a test for it?

extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -324,12 +324,23 @@ public function getPage() {
 				'ns' => $title->getNamespace()
 			];
 
-			if ( $this->isSupportedByMercury( $title ) ) {
-				// Empty category pages are not known but contain article list
+			// handle cases like starwars.wikia.com/wiki/w:c:clashroyale:Tesla (interwiki links)
+			$interWikiUrl = '';
+			InterwikiDispatcher::getInterWikiaURL( $title, $interWikiUrl, '' );
```
hakubo: mutating params is yuck, I know this returns true; is this true used somewhere?
or maybe we can actually return the value?

extensions/wikia/MercuryApi/MercuryApiController.class.php
hakubo: ok I see other properties are also camel case.

extensions/wikia/MercuryApi/MercuryApiController.class.php
rybmat: yeah, makes sense

extensions/wikia/MercuryApi/MercuryApiController.class.php
rybmat: this returns true because this method is used as a hook, and it always returns true

extensions/wikia/InterwikiDispatcher/SpecialInterwikiDispatcher_body.php
```diff
@@ -16,98 +16,126 @@
...
-	public static function getInterWikiaURL(Title &$title, &$url, $query) {
-		global $wgArticlePath, $wgScriptPath;
+	/**
+	 * checks if prefix given as parameter is supported by InterwikiDispatcher
+	 *
+	 * @param string $prefix
+	 *
+	 * @return bool
+	 */
+	public static function isSupportedPrefix( string $prefix ) {
```
hakubo: public static function isSupportedPrefix( string $prefix ): bool {}

hakubo (approved):

extensions/wikia/InterwikiDispatcher/SpecialInterwikiDispatcher_body.php
```diff
@@ -116,32 +144,49 @@ public static function getInterWikiaURL(Title &$title, &$url, $query) {
...
+	/**
+	 * hook run by Title::getFullURL()
+	 *
+	 * @param Title $title
+	 * @param $url
+	 * @param $query
+	 *
+	 * @return bool
+	 */
+	public static function getInterWikiaURLhook( Title &$title, &$url, $query ) {
```
rogatty: Hook should be capitalized.

rybmat: True

rogatty (approved):

PR #12544, opened 2017-03-08
XW-2904 | ArticleAsJson: check if 'caption' field is set

rybmat:
https://wikia-inc.atlassian.net/browse/XW-2904

check if 'caption' field is set in ArticleAsJson::onImageBeforeProduceHtml

@Wikia/x-wing

### DIFF
--- a/extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
+++ b/extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
@@ -306,7 +306,7 @@ class ArticleAsJson extends WikiaService {
 			$details['context'] = self::isIconImage( $details, $handlerParams ) ? self::MEDIA_CONTEXT_ICON :
 				self::MEDIA_CONTEXT_ARTICLE_IMAGE;
 
-			$caption = isset( $frameParams['caption'] ) ? $frameParams['caption'] : null;
+			$caption = $frameParams['caption'] ?? null;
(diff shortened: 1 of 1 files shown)


extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
```diff
@@ -306,7 +306,8 @@ public static function onImageBeforeProduceHTML(
 			$details['context'] = self::isIconImage( $details, $handlerParams ) ? self::MEDIA_CONTEXT_ICON :
 				self::MEDIA_CONTEXT_ARTICLE_IMAGE;
 
-			$media = self::createMediaObject( $details, $title->getText(), $frameParams['caption'], $linkHref );
+			$caption = isset( $frameParams['caption'] ) ? $frameParams['caption'] : null;
```
rogatty: Would `$caption = $frameParams['caption'] ?? null` work?

rogatty (approved):

extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
rybmat: ?? checks if variable is null, not sure how this will perform on missing key in array

rybmat: but I'll check

rybmat: ok, it should work

PR #12551, opened 2017-03-09
XW-2928 | Wiki header - wordmark + tally + add new page

latata:
https://wikia-inc.atlassian.net/browse/XW-2928

@idradm @weronika


idradm (approved):

extensions/wikia/PremiumPageHeader/scripts/wiki-header.js
```diff
@@ -0,0 +1,41 @@
...
+		});
+
+		$wikiHeader.find('.pph-wordmark-text').on('click', function () {
+			track({
+				action: tracker.ACTIONS.CLICK,
+				label: 'wordmark-text'
+			});
+		});
+
+		$wikiHeader.find('.pph-tally').on('click', function () {
```
weronika, line 25: does it make any sense to track tally if there is no link and even a cursor does not change to pointer on hover? absolutely nothing indicates that this is a clickable element, so I would say that results of tracking will give us no useful information

PR #12553, opened 2017-03-09
SUS-1771: Use local edit count instead of per-cluster edit count

mszabo-wikia:
Refactored edit count handling - rely on local edit count everywhere.

- New class `UserStats` encapsulates the properties of user (edit count, edit count this week etc.) we're interested in and handles loading and storing them into DB.
- `UserStatsService` allows accessing `UserStats` data, handles caching, cache invalidation and updates.
- Updates are performed outside of main request using background task.

https://wikia-inc.atlassian.net/browse/SUS-1771


includes/db/DatabaseUtility.php
```diff
@@ -202,6 +202,10 @@ class FakeResultWrapper extends ResultWrapper {
 	var $pos        = 0;
 	var $currentRow = null;
 
+	/**
+	 * FakeResultWrapper constructor.
```
macbre, line 206: `FakeResultWrapper constructor.` - sounds a bit too obvious :)

includes/wikia/services/persistence/UserStats.php
```diff
@@ -0,0 +1,105 @@
+<?php
+
+/**
+ * Class UserStats represents a collection of user properties
+ * that provide basic user statistics, such as edit count etc.
+ */
+class UserStats implements ArrayAccess {
```
macbre: Nice idea :+1:

includes/wikia/services/persistence/UserStats.php
```diff
@@ -0,0 +1,105 @@
...
+			$sqlStatName = $db->addQuotes( $statName );
+			$sqlStatValue = $db->addQuotes( $statValue );
+
+			$sql .= " WHEN $sqlStatName THEN $sqlStatValue";
+		}
+
+		$sql .= ' ELSE wup_value END';
+		$sql .= ' WHERE `wup_property` IN ' . $db->makeList( $statNames );
+		$sql .= ' AND `wup_user` = ' . $db->addQuotes( $this->userId );
+		$sql .= ';';
```
macbre: I'm wondering if we need this complex query here. Do we always set all four possible properties for `wikia_user_properties` table? Won't (at most) four separate queries by better here? @drozdo, any hints here?

includes/User.php
```diff
@@ -4048,35 +3978,15 @@ public function getFirstEditTimestamp() {
...
-			else {
-				$dbw = wfGetDB( DB_MASTER );
-			}
-
-			$dbw->update( '`user`',
-				array( 'user_editcount=user_editcount+1' ),
-				array( 'user_id' => $this->getId() ),
-				__METHOD__ );
-			$dbw->commit();
-
```
macbre, line 4017: :+1:

includes/wikia/services/persistence/UserStats.php
```diff
@@ -70,23 +70,17 @@ public function load( DatabaseBase $db ) {
...
+			];
 		}
 
-		$sql .= ' ELSE wup_value END';
-		$sql .= ' WHERE wup_property IN ' . $db->makeList( $statNames );
-		$sql .= ' AND wup_user = ' . $db->addQuotes( $this->userId );
-		$sql .= ';';
-
-		$db->query( $sql, __METHOD__ );
+		$db->replace( 'wikia_user_properties', [], $statInsert, __METHOD__ );
```
macbre: :+1:

macbre (approved):

includes/db/DatabaseUtility.php
mszabo-wikia: Better to make sure people don't think the constructor is fake too 😛

includes/wikia/services/UserStatsService.class.php
```diff
@@ -1,15 +1,9 @@
...
 
 	const CACHE_TTL = 86400;
-	const CACHE_VERSION = 'v1.2';
-	const USER_STATS_PROPERTIES = [
-		'editcount',
-		'editcountThisWeek',
-		'firstContributionTimestamp',
-		'lastContributionTimestamp'
-	];
+	const CACHE_VERSION = 'v2';
```
ghost, line 6: @TK-999 — per our conversation about "cache versioning", how will recaching affect our performance and stability?

includes/wikia/services/UserStatsService.class.php
```diff
@@ -55,23 +49,23 @@ public function increaseEditsCount() {
 		global $wgMemc;
 
 		if ( !$this->validateUser() ) {
-			return false;
+			return;
```
ghost: This is a general remark: for the sake of better code, we should avoid `null` as much as possible: nullchecks, returning nulls, etc. This method has side-effect, therefore it should return the status of it - whether it worked or not. Or, it should generally return the new number of edits and throw and exception on failure.

includes/wikia/services/UserStatsService.class.php
```diff
@@ -134,30 +131,44 @@ function () use ( $flags ) {
 					$stats['lastContributionTimestamp'] = $this->initLastContributionTimestamp();
 				}
 
+				if ( $stats->needsUpdate() ) {
+					$this->scheduleStatsUpdateTask( $stats );
```
ghost, line 136: Hmm, `getStats()` method scheduling a task? :-)

includes/wikia/services/UserStatsService.class.php
```diff
@@ -134,30 +131,44 @@ function () use ( $flags ) {
...
 			}
 		);
 		return $stats;
 	}
 
+	/**
+	 * Reset edit count of this user - force a recalculation and trigger update
+	 */
+	public function resetEditCount() {
+		$stats = $this->getStats();
```
ghost: `getStats()` schedules a task when needed, and here the same task is scheduled again?

includes/wikia/services/UserStatsService.class.php
mszabo-wikia: It is an unfortunate side-effect of how we store edit count information. When loading edit count for user from DB, some of the data may not yet be initialized, so they have to be calculated on-demand and persisted to DB.

includes/wikia/services/UserStatsService.class.php
mszabo-wikia: In this case we are safe - in https://wikia-inc.atlassian.net/browse/JPN-561 we also updated cache versionining of this component.

ghost (approved):

mszabo-wikia:
I'll be opening a new PR for this.

PR #12569, opened 2017-03-10
XW-2855 | CMP: filter out categories without members from MercuryApi response

rybmat:
https://wikia-inc.atlassian.net/browse/XW-2855

@Wikia/x-wing

### DIFF
--- a/extensions/wikia/MercuryApi/models/MercuryApi.class.php
+++ b/extensions/wikia/MercuryApi/models/MercuryApi.class.php
@@ -491,7 +491,8 @@ class MercuryApi {
 			}
 		} elseif ( $item['article_id'] === 0 ) {
 			$title =  Title::newFromText( $item['title'] );
-			$category = Category::newFromTitle( $title );
+
+			$category = empty( $title ) ? null : Category::newFromTitle( $title );
 
 			if ( !empty( $category ) && $category->getPageCount() ) {
 				$result['url'] = $title->getLocalURL();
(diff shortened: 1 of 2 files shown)


extensions/wikia/MercuryApi/models/MercuryApi.class.php
```diff
@@ -490,9 +490,14 @@ public function processCuratedContentItem( $item ) {
 				return $result;
 			}
 		} elseif ( $item['article_id'] === 0 ) {
-			$result['url'] = Title::newFromText( $item['title'] )->getLocalURL();
+			$title =  Title::newFromText( $item['title'] );
```
hakubo, line 493: newFromText may return null
and then Category::newFromTitle won't like it

extensions/wikia/MercuryApi/tests/MercuryApiModelTest.php
```diff
@@ -409,11 +409,10 @@ public function getCuratedContentItemsDataProvider() {
 	 * @param $wgArticlePath
 	 * @param $getLocalURL
 	 */
-	public function testProcessCuratedContentItem( $expected, $item, $wgArticlePath, $getLocalURL ) {
+	public function testProcessCuratedContentItem( $expected, $item, $wgArticlePath, $getLocalURL, $categoryMembersCount ) {
```
hakubo, line 412: Tests!
🙇

extensions/wikia/MercuryApi/models/MercuryApi.class.php
rybmat: checked line below

kvas-damian (approved):

PR #12570, opened 2017-03-10
(PLATFORM-2934) Cleanup of references to now unused columns

Grunny:
/cc @Wikia/core-platform-team


extensions/wikia/UserLogin/js/UserBaseAjaxForm.js
```diff
@@ -44,14 +43,6 @@
 	};
 
 	/**
-	 * Bind form events
-	 */
-	UserBaseAjaxForm.prototype.bindEvents = function () {
-		this.form.on('submit', this.submitLogin.bind(this));
```
themech: Hmm, does it mean we're disabling the ajax login form here?

Grunny: Yeah, good catch. Was a left over from the larger cleanup that I undid, I missed restoring this. This extension is only used on devboxes and would still work (you'd just go to Special:UserLogin on an error rather than it being displayed in the form) but I'll restore it.

Grunny: ...and done. :)

themech (approved):

PR #12572, opened 2017-03-12
PLATFORM-2351: Make Special:Wanted* not register redlinks on JS/CSS/Lua pages

KockaAdmiralac:
[WantedPages](http://c.wikia.com/wiki/Special:WantedPages), [WantedTemplates](http://c.wikia.com/wiki/Special:WantedTemplates), [WantedFiles](http://c.wikia.com/wiki/Special:WantedFiles) and [WantedCategories](http://c.wikia.com/wiki/Special:WantedCategories) are registering a redlink even when it appears on pages whose content isn't wikitext. This happens on newer MediaWiki versions too, but since it's not supposed to be working like that on Wikia, this pull request fixes that. It also prevents non-wikitext pages from being added into [Pages with broken file links](http://c.wikia.com/wiki/Category:Pages_with_broken_file_links) category.

Changes were tested on MediaWiki 1.28, so if something breaks, it's most likely due to incompatibilities between versions.

Support ticket: [#320236](https://support.wikia-inc.com/hc/en-us/requests/320236)
Bug ticket: [PLATFORM-2351](https://wikia-inc.atlassian.net/browse/PLATFORM-2351)

### DIFF
--- a/includes/parser/Parser.php
+++ b/includes/parser/Parser.php
@@ -5443,7 +5443,7 @@ class Parser {
 		$shouldAddTrackingCategory = true;
 		wfRunHooks( 'ParserShouldAddTrackingCategory', array( $this, $title, $file, &$shouldAddTrackingCategory ) );
 
-		if ( !$file && $shouldAddTrackingCategory ) {
+		if ( !$file && $this->mTitle->isWikitextPage() && $shouldAddTrackingCategory ) {
 		# wikia end
 
 			$this->addTrackingCategory( 'broken-file-category' );
--- a/includes/specials/SpecialWantedcategories.php
+++ b/includes/specials/SpecialWantedcategories.php
@@ -35,17 +35,30 @@ class WantedCategoriesPage extends WantedQueryPage {
 	}
 
 	function getQueryInfo() {
-		return array (
-			'tables' => array ( 'categorylinks', 'page' ),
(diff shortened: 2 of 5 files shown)


mszabo-wikia:
This is fixed by wrapping the JS:
```
// <nowiki>
...
// </nowiki>
```

KockaAdmiralac:
I'm aware, but the ticket was opened after Staff refused to remove redlinks from user JS on Dev Wiki ([ZD#317255](https://support.wikia-inc.com/hc/en-us/requests/317255)) and because registering redlinks in those pages is certainly not intended. But if the community team decided there's no need for a change there now, sure.

shaedrich:
Using `<nowiki></nowiki>` tags is inconvenient when using json or a similar content. On most JS/CSS/Lua pages we newer need redlinks. I agree with @KockaAdmiralac.

mszabo-wikia:
There is quite a discussion in [T18683](https://phabricator.wikimedia.org/T18683) around the issue. Given that the expected behavior is not clear, and the current behavior matches that of upstream, there is little benefit in implementing hacky fix, especially when there is an easy workaround available.

shaedrich:
A workaround is hacky, too: JSON does not support wikicode tags.

Another possibility would be to implement a namespace filter on these pages.

PR #12617, opened 2017-03-16
XW-2915 | PPH: wiki header background image

rybmat:
https://wikia-inc.atlassian.net/browse/XW-2915

@Wikia/x-wing

### DIFF
--- a/extensions/wikia/PremiumPageHeader/styles/_wiki-header.scss
+++ b/extensions/wikia/PremiumPageHeader/styles/_wiki-header.scss
@@ -65,15 +65,27 @@
 }
 
 .pph-wiki-header {
-	background-color: $color-buttons;
+	background: {
+		color: $color-buttons;
+		position: 600px 0;
+		repeat: no-repeat;
+		size: cover;
+	}
 	color: $color-button-text;
-	display: flex;
 	margin: -1px -31px 20px;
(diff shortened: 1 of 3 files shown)


weronika (approved):

hakubo (approved):

extensions/wikia/PremiumPageHeader/styles/_wiki-header.scss
```diff
@@ -74,7 +83,20 @@
...
+		display: block;
+		height: 100%;
+		position: absolute;
+		width: 100%;
+		z-index: -1;
+	}
+
 	.wordmark {
+		// !important to avoid showing bg image from parent when wordmark has opacity
+		background: $color-buttons !important;
```
weronika, line 109: doesn't that affect the gradient? maybe `background-image: none !important;` would be enough.

rybmat: no, it does not affect the gradient, at least I don't see the difference in gradient. And taking into account that we need to use !important here, I'd prefer to set this colour explicitly to make sure that it is the same as colour of the rest of header

idradm (approved):

idradm (approved):

PR #12621, opened 2017-03-17
ADEN-4872 | Add price for AdSence/AdX in wg variable

mleczey:
## Links
https://wikia-inc.atlassian.net/browse/ADEN-4872
## Description
Add price for AdSence/AdX in wg variable


dianafa:
to be changed anyways in the scope of https://wikia-inc.atlassian.net/browse/ADEN-4857

so in that case LGTM

PR #12623, opened 2017-03-20
XW-2851 | AdBlockEnabled flag in Qualaroo

latata:
https://wikia-inc.atlassian.net/browse/XW-2851

@Wikia/x-wing

### DIFF
--- a/extensions/wikia/Qualaroo/scripts/Qualaroo.js
+++ b/extensions/wikia/Qualaroo/scripts/Qualaroo.js
@@ -1,4 +1,4 @@
-( function ( window ) {
+( function ( window, document ) {
 	'use strict';
 	var _kiq = [],
 		createCookie;
@@ -15,10 +15,20 @@
 		document.cookie = cookieValue;
 	};
 
-	setTimeout(function(){
-		var d = document, f = d.getElementsByTagName('script')[0], s = d.createElement('script'); s.type = 'text/javascript';
-		s.async = true; s.src = window.wgQualarooUrl; f.parentNode.insertBefore(s, f);
-	}, 1);
+	function loadQualaroo () {
+		setTimeout(function(){
+			var d = document, f = d.getElementsByTagName('script')[0], s = d.createElement('script'); s.type = 'text/javascript';
+			s.async = true; s.src = window.wgQualarooUrl; f.parentNode.insertBefore(s, f);
+		}, 1);
+	}
+
+	function setAdBlockEnabledAndLoadQualaroo(enabled) {
+		_kiq.push(['set', {
+			'adBlockedEnabled': enabled
+		}]);
+		loadQualaroo();
+	}
+
 
 	if (window.wgUser) {
 		_kiq.push(['identify', window.wgUser]);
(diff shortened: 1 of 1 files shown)


extensions/wikia/Qualaroo/scripts/Qualaroo.js
```diff
@@ -70,5 +80,17 @@
 		createCookie('qualaroo_survey_submission');
 	});
 
+	if(ads.context.opts.sourcePointDetection) {
+		document.addEventListener('sp.blocking', function () {
```
kvas-damian: Is it possible that this event handler will be registered after event was fired? Can we mitigate that problem?

extensions/wikia/Qualaroo/scripts/Qualaroo.js
```diff
@@ -15,10 +15,20 @@
...
+	function loadQualaroo () {
+		setTimeout(function(){
+			var d = document, f = d.getElementsByTagName('script')[0], s = d.createElement('script'); s.type = 'text/javascript';
+			s.async = true; s.src = window.wgQualarooUrl; f.parentNode.insertBefore(s, f);
+		}, 1);
+	}
+
+	function setAdBlockEnabledAndLoadQualaroo(enabled) {
+		_kiq.push(['set', {
+			'adBlockedEnabled': enabled
```
rogatty: `adBlocked` or `adBlock`?

extensions/wikia/Qualaroo/scripts/Qualaroo.js
rogatty: I wonder if this piece of code would fit better in the file that handles SP integration. We could require this file in there or something like that.

extensions/wikia/Qualaroo/scripts/Qualaroo.js
```diff
@@ -70,5 +80,21 @@
 		createCookie('qualaroo_survey_submission');
 	});
 
+	if (ads.context.opts.sourcePointDetection) {
```
kvas-damian: Please decide `ads` vs `window.ads`.

extensions/wikia/Qualaroo/scripts/Qualaroo.js
```diff
@@ -70,5 +80,21 @@
...
+		if (!window.ads || !window.ads.runtime || !window.ads.runtime.sp || window.ads.runtime.sp.blocking === undefined) {
```
rogatty: Seems brittle, would be nice to have an integration test.

rogatty (approved):

extensions/wikia/Qualaroo/scripts/Qualaroo.js
```diff
@@ -70,5 +80,21 @@
 		createCookie('qualaroo_survey_submission');
 	});
 
+	if (window.ads.context.opts.sourcePointDetection) {
+		if (!window.ads.runtime || !window.ads.runtime.sp || window.ads.runtime.sp.blocking === undefined) {
```
dianafa, line 84: per discussion in the corridor, we can add a comment that this is the same logic as in: https://github.com/Wikia/app/blob/dev/extensions/wikia/ARecoveryEngine/js/sourcePointHelper.js#L54

latata: Actually it's not the same; firstly we check if sourcePointDetection is enabled not sourcePointRecovery, secondly we check if blocking is undefined (so not set to false or true).

hakubo (approved):

PR #12627, opened 2017-03-20
SUS-1508: Introduce limit parameter for domains query

mszabo-wikia:
https://wikia-inc.atlassian.net/browse/SUS-1508

### DIFF
--- a/extensions/wikia/WikiaApi/WikiaApiQueryDomains.php
+++ b/extensions/wikia/WikiaApi/WikiaApiQueryDomains.php
@@ -59,6 +59,23 @@ class WikiaApiQueryDomains extends ApiQueryBase {
 			if ( !empty($from) ) {
 				if ($from && is_int($from)) $this->addWhere('city_id >= '.intval($from));
 			}
+
+			if ( !empty( $from ) && !empty ( $to ) ) {
+				$settings = $this->getAllowedParams();
+
+				if ( $this->getMain()->canApiHighLimits() ) {
+					$maxLimit = $settings['limit'][ApiBase::PARAM_MAX2];
+				} else {
+					$maxLimit = $settings['limit'][ApiBase::PARAM_MAX];
+				}
+
+				$requested = 1 + $to - $from;
+				if ( $requested > $maxLimit ) {
+					$this->setWarning( "Count of results must not be over $maxLimit (set to $requested)" );
+				}
+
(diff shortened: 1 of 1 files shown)


macbre (approved):
`city_id` are obviously not continuous values (as we remove unactive wikis), but still :+1:  :)

extensions/wikia/WikiaApi/WikiaApiQueryDomains.php
```diff
@@ -59,6 +59,23 @@ public function execute() {
...
+				} else {
+					$maxLimit = $settings['limit'][ApiBase::PARAM_MAX];
+				}
+
+				$requested = 1 + $to - $from;
+				if ( $requested > $maxLimit ) {
+					$this->setWarning( "Count of results must not be over $maxLimit (set to $requested)" );
+				}
+
+				$limit = min( $requested, $maxLimit );
```
ghost, line 77: Do we want `$limit` to be set only when `$from` and `$to` are not empty? What if any of those is not set? No limit at all?

extensions/wikia/WikiaApi/WikiaApiQueryDomains.php
```diff
@@ -85,6 +102,7 @@ public function execute() {
 		} else {
 			$this->addFields(array('city_id', 'city_url', 'city_lang'));
 			$this->addOption( "ORDER BY ", "city_id" );
+			$this->addOption( 'LIMIT', $limit );
```
ghost, line 105: Possible use of undeclared variable here and MySQL error.

ghost (changes requested):

extensions/wikia/WikiaApi/WikiaApiQueryDomains.php
mszabo-wikia: If the variable is not provided by user in the URL, MW API will set it to the default value declared on https://github.com/Wikia/app/pull/12627/files#diff-7f7f6f25d0e2d3ff0f4770cea8e296dbR165 . So without a provided parameter, limit will be 10.

ghost: Thanks!

ghost (approved):

ghost:
Tested and accepted.

PR #12632, opened 2017-03-20
ADEN-4886 Fixed XSS vulnerability in Krux integration

nandy-andy:
`kxsrc` most likely is an old way of testing integration which we don't need anymore.


dianafa:
LGTM

PR #12637, opened 2017-03-20
CAKE-1460: Removes MW-specific styling for MW+CP integration

gbenson-wikia:
Layout styling was left alone - this intentionally only addressed the colors.  There's some additional styling work that'll be required to get this polished up, but that's outside the scope of this ticket.

@Wikia/cake

### DIFF
--- a/extensions/wikia/ContributionPrototype/styles/ContributionPrototype.scss
+++ b/extensions/wikia/ContributionPrototype/styles/ContributionPrototype.scss
@@ -27,7 +27,6 @@ body {
   display: flex;
   align-items: center;
   margin-bottom: 10px;
-  background: #030f1d;
 
   h2 {
     flex-grow: 1;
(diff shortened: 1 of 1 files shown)


drsnyder:
Was there a commit that added these that you referenced to unwind them?

nmonterroso:
LGTM

gbenson-wikia:
@drsnyder The bake-off specific changes were scattered across a large enough number of commits that it was faster to do it this way than to try to unwind them individually.  Knowing that we're going to need to make a fit-and-finish pass anyway, I elected to skip anything structural and just strip out anything having to do with colors.  This let me test each color in small batches as I went, permitting me to ensure that the look-and-feel was still acceptable.

drsnyder:
@gbenson-wikia ok, sounds good. I was cautious about giving a blanket "LGTM" since I wasn't familiar with the changes that were added previously.

PR #12640, opened 2017-03-21
ADEN-4853 | make porvata beautiful

mleczey:
## Links
https://wikia-inc.atlassian.net/browse/ADEN-4853
## Description
make porvata beautiful


extensions/wikia/AdEngine/js/video/player/porvata/floater.js
```diff
@@ -40,9 +43,7 @@ define('ext.wikia.adEngine.video.player.porvata.floater', [
 				elements.video.addEventListener('start', listeners.start);
 			}
 
-			if (elements.video.isPlaying()) {
-				elements.video.resize(width, height);
-			}
+			elements.video.resize(width, height);
```
dianafa, line 46: why we don't want to resize video only when it's playing?

extensions/wikia/AdEngine/js/video/player/porvata/porvata.js
```diff
@@ -64,6 +64,15 @@ define('ext.wikia.adEngine.video.player.porvata', [
...
 				tracker.register(video, params);
 
+				function shouldResume(isVisible) {
+					// Don't resume when video was paused manually
+					return isVisible && autoPaused
+						// Do not resume automatically when video has controls
+						&& !videoSettings.hasUiControls()
+						// Do not resume when video floating is active
+						&& !isFloatingEnabled(params);
+				}
```
dianafa: per coding conventions (https://github.com/Wikia/guidelines/blob/master/PHP/CodingConventions.md#conditional-logic) please move `&&`s to the EOLs

extensions/wikia/AdEngine/js/video/player/ui/playPauseButton.js
```diff
@@ -0,0 +1,70 @@
...
+define('ext.wikia.adEngine.video.player.ui.playPauseButton', [
+	'wikia.document',
+	'wikia.log'
+], function (doc, log) {
+	'use strict';
+
+	var logGroup = 'ext.wikia.adEngine.video.player.ui.playPauseButton',
+		animateCssClass = 'animate',
+		playSvg = '<svg width="12" height="16" viewBox="0 0 12 16" xmlns="http://www.w3.org/2000/svg"><path d="M11.767 8.437L.857 15.903a.553.553 0 0 1-.564.037.53.53 0 0 1-.293-.473V.533c0-.2.113-.38.293-.473a.557.557 0 0 1 .565.036l10.91 7.467A.53.53 0 0 1 12 8a.53.53 0 0 1-.233.437z"/></svg>',
+		pauseSvg = '<svg width="14" height="16" viewBox="0 0 14 16" xmlns="http://www.w3.org/2000/svg"><g><rect width="5" height="16" rx="1"/><rect x="9" width="5" height="16" rx="1"/></g></svg>',
```
dianafa, line 11: couldn't we reuse buttons used in pfff video?

extensions/wikia/AdEngine/js/video/videoSettings.js
```diff
@@ -20,6 +21,7 @@ define('ext.wikia.adEngine.video.videoSettings', [
 			state.autoPlay = isAutoPlay(params);
 			state.splitLayout = Boolean(params.splitLayoutVideoPosition);
 			state.moatTracking = Boolean(params.moatTracking);
+			state.withUiControls = Boolean(params.hasUiControls);
```
dianafa, line 24: I really don't like this typecasting. The double negation`!!` does the exact same thing, but is shorter. It seems to have better performance (https://www.sitepoint.com/javascript-double-negation-trick-trouble/) and from what I see is much more widely used in our repo.

What would need to happen if we stick to one, `!!` convention? @v3rth @fraszczakszymon @nandy-andy what do you think guys?

skins/oasis/css/core/ads.scss
```diff
@@ -646,13 +658,151 @@ body > a > img[src$="noad.gif"] {
...
+	width: 100%;
+
+	.control-bar {
+		padding: 0 $video-control-bar-padding;
+
+		.progress-bar {
+			background-color: rgba(175, 175, 175, 0.5);
+			border-radius: 3px;
+			bottom: initial;
+			height: 4.2px;
```
dianafa: 4px wouldn't be enough?

skins/oasis/css/core/ads.scss
```diff
@@ -646,13 +658,151 @@ body > a > img[src$="noad.gif"] {
...
+	bottom: 0;
+	left: 0;
+	position: absolute;
+	width: 100%;
+
+	.control-bar {
+		padding: 0 $video-control-bar-padding;
+
+		.progress-bar {
+			background-color: rgba(175, 175, 175, 0.5);
```
dianafa: can we move this color to a const with descriptive name?

extensions/wikia/AdEngine/js/video/player/porvata/floater.js
mleczey: Because we want to resize video even when it's paused. Not doing that ends up with not resized floating video.

extensions/wikia/AdEngine/js/video/videoSettings.js
mleczey: Doing that only once has very low impact on overall performance, it's not done in loop. For me static constructor is better, because this not stops from further optimisation done by browser JS engines that might be done. `!!` on the other hand does not have such guarantee. Either way I'm ok with both solutions, just please decide which one to chose.

skins/oasis/css/core/ads.scss
mleczey: This value was taken from pfff video player.

extensions/wikia/AdEngine/js/video/player/ui/playPauseButton.js
mleczey: Unfortunately pfff video uses font as buttons. I've used icons that were attached to the ticket. After CR I'll create another ticket to introduce mechanism, that will allow inlining svg in ads.

dianafa:
LGTM, let's discuss `!!` on daily! :)

PR #12646, opened 2017-03-21
fixing css class names

nmonterroso:
@Wikia/cake 
fixing the styles to conform to the new style guidelines. I have an open branch where I'm doing this on the CMS-side as well


drsnyder:
LGTM

PR #12654, opened 2017-03-22
 IRIS-4182 | Display the exact number of up voters

slayful:
Since we're not displaying the username of the first up voter, then we do not want to decrement the number of up votes.

https://wikia-inc.atlassian.net/browse/IRIS-4182


tortila (approved):
:+1:

PR #12659, opened 2017-03-23
ADEN-4890 | Additional logging to kikimora

mleczey:
##Links
https://wikia-inc.atlassian.net/browse/ADEN-4890
## Description
Adding additional logging to kikimora to check how cpm is mapping to price.


mleczey:
This change is not necessary.

PR #12666, opened 2017-03-23
ADEN-4888 Update MR slot if is loaded from BTF

v3rth:


extensions/wikia/AdEngine/js/provider/gpt/googleSlots.js
```diff
@@ -28,9 +28,19 @@ define('ext.wikia.adEngine.provider.gpt.googleSlots', [
 		return slots[slotId];
 	}
 
+	function refreshSlot(slots, updateCorrelator) {
```
dianafa: if slots can be an array, I'd rename this function to `refreshSlots`

extensions/wikia/AdEngine/js/provider/gpt/googleSlots.js
```diff
@@ -28,9 +28,19 @@ define('ext.wikia.adEngine.provider.gpt.googleSlots', [
...
+		slots = slots.isArray() ? slots : [slots];
+		window.googletag.pubads().clear(slots);
+		if (updateCorrelator) {
+			window.googletag.pubads().updateCorrelator();
+		}
+		window.googletag.pubads().refresh(slots);
```
dianafa: shouldn't we use `win` instead of `window` ?

extensions/wikia/AdEngine/js/template/bfaaDesktop.js
```diff
@@ -114,6 +121,11 @@ define('ext.wikia.adEngine.template.bfaaDesktop', [
 		});
 
 		unblockedSlots.forEach(btfBlocker.unblock);
+
+		if (params.loadMedrecFromBTF) {
```
dianafa: can we add a comment here why it's done exactly in this place, not earlier?

nandy-andy (approved):

extensions/wikia/AdEngine/js/provider/gpt/googleTag.js
```diff
@@ -152,6 +152,10 @@ define('ext.wikia.adEngine.provider.gpt.googleTag', [
 		return slot;
 	}
 
+	function refreshSlot(slot) {
+		window.googletag.pubads().refresh([slot]);
```
nandy-andy, line 156: Think of creating a follow-up to check the data we're afraid will stop working without using here AdEngine's `push` or maybe add it to AC of already existing follow-up ticket.

PR #12671, opened 2017-03-23
MAIN-10068

timmyquivy:
More often than not, for user block it makes sense to use the EXACT attribute as opposed to not, so let's default check it to avoid user error.


timmyquivy:
Approved in SUS chan

timmyquivy:
And I'm terrible at this .... *sigh*

PR #12672, opened 2017-03-24
IRIS-4148 Make sure lazy loaded images appear in discussions

garthwebb:
Our approach removes the lazy loading aspects and makes these normal images.

Ticket:

https://wikia-inc.atlassian.net/browse/IRIS-4148

### DIFF
--- a/extensions/wikia/Discussions/maintenance/ForumDumper.php
+++ b/extensions/wikia/Discussions/maintenance/ForumDumper.php
@@ -194,6 +194,15 @@ class ForumDumper {
 		$title = $articleComment->getMetadata( 'title', '' );
 		$parsedText = $this->getParsedText( $articleComment );
 
+		if ( empty( $parsedText ) ) {
+			// If there's nothing to parse, use rawText as the default
+			$parsedText = $rawText;
+		} else {
+			// If there is parsed text, use the tag stripped version as rawText so it can
+			// be the plaintext version (otherwise its full of wikitext)
+			$rawText = strip_tags( $parsedText );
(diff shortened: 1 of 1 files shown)


extensions/wikia/Discussions/maintenance/ForumDumper.php
```diff
@@ -111,6 +111,13 @@ public function getPages() {
 			->WHERE( 'page_namespace' )->IN( self::FORUM_NAMEPSACES )
 			->ORDER_BY( 'idx' )
 			->runLoop( $dbh, function ( &$pages, $row ) use ( &$display_order ) {
+				// A few of these properties were removed and do not appear on some wikis
+				foreach ( [ 'sticky', 'locked', 'protected' ] as $prop ) {
+					if ( !property_exists( $row, $prop ) ) {
+						$row->$prop = 0;
+					}
+				}
+
```
garthwebb, line 120: This code eliminates some noisy warnings.

mszabo-wikia: AFAIK we performed a site-wide schema update in  [SUS-1690](https://wikia-inc.atlassian.net/browse/SUS-1690), so these columns should no longer appear anywhere (and they were never used)

garthwebb: @TK-999, thanks for the update!  I've seen some still in DEV so its good to know I don't need to worry about them in production.

jsutterfield:
LGTM

PR #12678, opened 2017-03-24
XW-3022 | replacing document.write() with DOM api

rybmat:
https://wikia-inc.atlassian.net/browse/XW-3022

@Wikia/x-wing

### DIFF
--- a/extensions/wikia/AbTesting/js/AbTest.js
+++ b/extensions/wikia/AbTesting/js/AbTest.js
@@ -135,19 +135,21 @@
 	};
 
 	AbTest.loadExternalData = function( data ) {
-		var index, groupData, html = '';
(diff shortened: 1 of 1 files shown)


extensions/wikia/AbTesting/js/AbTest.js
```diff
@@ -135,20 +135,22 @@
 	};
 
 	AbTest.loadExternalData = function( data ) {
-		var index, groupData, html = '';
+		var index, groupData,
+			head = document.getElementsByTagName('head')[0];
```
latata: you can change it to simpler document.head

rybmat: didn't know, thanks!

PR #12685, opened 2017-03-27
SUS-1839 | remove UserChangesHistory::SavePreferencesHook method

macbre:
https://wikia-inc.atlassian.net/browse/SUS-1839

`specials.user_history` table is (or used to be) populated via method bound to `SavePreferencesHook` hook. Nothing really reads this data, there used to be INSERTs only (8,94 GB stored). Let's sunset it.


mszabo-wikia (commented):
What about the extension itself? Does the login hook still function properly in Helios era? Could we possibly get rid of the whole extension?

ghost (approved):

macbre:
@TK-999 , it does - user_login_history and user_login_history_summary tables are still updated.

mszabo-wikia (approved):

PR #12691, opened 2017-03-27
cake-1498 - only overwrite for NS_MAIN pages

nmonterroso:
@Wikia/cake 
https://wikia-inc.atlassian.net/browse/CAKE-1498

Explicitly check that we're on the `NS_MAIN` namespace before embedding the CP.


drsnyder:
LGTM

PR #12694, opened 2017-03-28
SUS-1898 Remove avatars from old Signup view

fraszczakszymon:
`user_login_history` stores around 33GB of data and it seems that only random avatars on Special:UserSignup view use whole table. Signup flow seems to be deprecated and it's enabled on one wiki only.

Second place where the table is used is FounderEmailsCompleteDigestEvent (to calculate wiki new users count) but it seems that variable `newUsers` and translations `emailext-founder-digest-users-*` are not included in e-mail message.

After this cleanup we can get rid of whole `user_login_history`.


extensions/wikia/UserChangesHistory/UserChangesHistory.class.php
```diff
@@ -79,13 +79,6 @@ static public function LoginHistoryHook( $from, $user, $type = false ) {
 						// @see PLATFORM-1309
 						$dbw_specials = wfGetDB( DB_MASTER, array(), $wgSpecialsDB ) ;
 
-						$dbw_specials->insert(
-							"user_login_history",
-							$params,
-							__METHOD__,
-							array('IGNORE')
-						);
-
```
macbre: :+1: 

The same needs to be made in the Perl backend repo that handles Scribe events.

macbre (approved):

ghost (approved):

PR #12703, opened 2017-03-29
XW-3075 | disable inspectlet on harry potter articles

idradm:
https://wikia-inc.atlassian.net/browse/XW-3075


includes/wikia/services/InspectletService.class.php
```diff
@@ -21,7 +21,6 @@ class InspectletService extends WikiaService {
 		 * - overwatch.wikia.com
 		 */
 		self::MAIN_PAGE => 1469642463,
-		self::HARRY_POTTER_ARTICLES => 904885521
```
latata: Didn't we decide that we want to disable it on main pages as well?

latata (approved):

weronika (approved):

PR #12707, opened 2017-03-29
(MAIN-10111) Fix JS error in Safari 9

Grunny:
Don't use ES6 syntax in order to fix the error:

    SyntaxError: Unexpected keyword 'const'. Const declarations are
    not supported in strict mode. VMundefined
    DesignSystemGlobalNavigationOnSiteNotifications.view.js:79

/cc @rogatty @slayful


slayful:
Seems PHP failed on GIT error.
`hudson.plugins.git.GitException: Failed to fetch from git://github.com/Wikia/app.git`

slayful (approved):
LGTM

PR #12723, opened 2017-03-31
ADEN-4903 Fix SP MMS

v3rth:


extensions/wikia/ARecoveryEngine/ARecoveryEngineApiController.class.php
```diff
@@ -8,11 +8,11 @@ public function getDelivery() {
 		$this->loadScript(new ResourceLoaderAdEngineSourcePointCSDelivery());
 	}
 
-	public function getSourcePointMessaging() {
+	public function getspmsg() {
```
nandy-andy, line 11: Keep in mind cached pages which inline HTML will have:
```js
window._sp_.bootstrap('{{domain}}/api/v1/ARecoveryEngine/SourcePointMessaging');
```
Should we keep the old methods as aliases till 1-2 releases after releasing these changes?

v3rth: Currently, SP MMS is turned off if [wgVariable is set to false](https://github.com/Wikia/app/blob/5339152470b1d9a68f66428e27765ed49ce93c60/extensions/wikia/ARecoveryEngine/templates/ARecoveryEngineApiController_getBootstrap.mustache#L44) we are safe :) But I will write it down in ticket

PR #12725, opened 2017-03-31
SUS-1919: Fix parameter handling in WikiaMapsPoiController

mszabo-wikia:
https://wikia-inc.atlassian.net/browse/SUS-1919


extensions/wikia/WikiaMaps/controllers/WikiaMapsPoiController.class.php
```diff
@@ -264,7 +264,7 @@ public function isValidUrl( WikiaValidator $validator ) {
 	 */
 	private function getSanitizedData() {
 		$poiData = [
-			'name' => $this->getData( 'name' ),
+			'name' => htmlspecialchars( $this->getData( 'name' ) ),
```
Grunny: Looks like some of these are included in attributes in the templates, so using `ENT_QUOTES` might be a best practice (even though it looks like all attributes are currently using double quotes).

Grunny (approved):
Minor comment, but looks good.

ghost (approved):

PR #12732, opened 2017-04-03
SUS-1923 | get rid of redundant pages.page_title_lower column

macbre:
Thanks to the case insensitive collation we can perform queries using `page_title` column and `page_title_lower` can be safely removed when this PR is merged.

```sql
mysql@geo-db-archive-slave.query.consul[dataware]>select page_id, page_title, page_title_lower from pages where page_wikia_id = 5915 and page_title = 'SzneKA_z_Glancem';
+---------+------------------+------------------+
| page_id | page_title       | page_title_lower |
+---------+------------------+------------------+
|    1464 | Szneka_z_glancem | szneka_z_glancem |
+---------+------------------+------------------+
1 row in set (0.01 sec)
```

`dataware.pages` table weights 217 GB (with 47 GB of data) and stores ~300 mm rows. Let's make it a bit lighter.

Scribe events-powered table has been introduced in 2013 - https://github.com/Wikia/backend/commit/cecc0f41f574caee8c5d67b0dfa5417110c5e19f

@drozdo


extensions/wikia/ImageReview/ImageReviewDatabaseHelper.class.php
```diff
@@ -23,15 +23,16 @@ public function selectImagesForList( $sOrder,
 		$oDB = $this->getDatawareDB( DB_SLAVE );
 
 		$oResults = $oDB->query('
-			SELECT pages.page_title_lower, image_review.wiki_id, image_review.page_id, image_review.state, image_review.flags, image_review.priority, image_review.last_edited
```
macbre, line 26: The only caller of this method does not make any use of `page_title_lower` field.

maintenance/wikia/updateDatawarePages.php
```diff
@@ -207,7 +206,6 @@ private function updateDatawarePage( $dbw, $localPage, $datawarePage ) {
 				'page_wikia_id' => $wgCityId,
 				'page_id' => $localPage->page_id,
 				'page_namespace' => $localPage->page_namespace,
-				'page_title_lower' => mb_strtolower( $localPage->page_title ),
```
macbre, line 210: We may get rid of this script completely (being considered in https://wikia-inc.atlassian.net/browse/SUS-1789).

mszabo-wikia (approved):

PR #12741, opened 2017-04-04
SUS-1923: ImageReview does not need to join with dataware.pages

macbre:
https://wikia-inc.atlassian.net/browse/SUS-1923


mszabo-wikia (approved):

macbre:
This join is used to include images that still exist only (`dataware.pages` gets DELETE queries on file deletes, `image_review` tables does not).

PR #12745, opened 2017-04-05
ADEN-4908 New way of building remnant ad units

v3rth:


extensions/wikia/AdEngine/js/AdSlotTracker.js
```diff
@@ -0,0 +1,47 @@
+/*global define*/
+define('ext.wikia.adEngine.adSlotTracker',  [
```
fraszczakszymon: Could you please define this module in some subdirectory (new one?) and name file with lowercase? I'm not sure about `tracker` in name. I expected here something a little bit different :P

extensions/wikia/AdEngine/js/slot/adUnitBuilder.js
```diff
@@ -18,7 +19,61 @@ define('ext.wikia.adEngine.slot.adUnitBuilder', [
...
+		if (params.skin === 'oasis') {
+			result = browserDetect.isMobile() ? 'tablet' : 'desktop';
+		} else if (params.skin === 'mercury' || params.skin === 'mobile-wiki') {
+			result = 'smartphone';
+		}
+
+		return result;
+	}
+
+	function buildNew(src, slotName, passback) {
```
fraszczakszymon: What do you think about moving this to new module and keep exactly the same API for both adUnitBuilders? Then we can pass new module in:
https://github.com/Wikia/app/pull/12745/files#diff-31d08cf2d202c80ee219ba9cc3e46e03R43
And keep 'decision' logic in one place: 
https://github.com/Wikia/app/pull/12745/files#diff-4addc132482932ea704402dd5c7d7746R67

fraszczakszymon (approved):

fraszczakszymon (approved):

extensions/wikia/AdEngine/js/context/adContext.js
```diff
@@ -151,6 +151,8 @@ define('ext.wikia.adEngine.adContext', [
 				geo.isProperGeo(instantGlobals.wgAdDriverRubiconFastlaneProviderCountries);
 		}
 
+		context.opts.enableRemnantNewAdUnit = geo.isProperGeo(instantGlobals.wgAdDriverMEGACountries);
```
nandy-andy, line 154: What do you think of setting a default value of it in PHP as well?

PR #12756, opened 2017-04-06
SUS-1945 | get rid of dataware.photo_gallery_feeds table

macbre:
https://wikia-inc.atlassian.net/browse/SUS-1945

Not used since early 2014. The maintenance script that referred to this table is registered in chef for devboxes only.


maintenance/wikia/WikiaPhotoGalleryFeedUpdate.php
```diff
@@ -1,45 +0,0 @@
...
-	array("timestamp < '$timestemp'"),
-	__METHOD__
-);
-
-$count = 0;
-while ($oRow = $dbr->fetchObject($oRes)) {
-	if (!isset($options['quiet'])) {
-		echo "Updating feed: {$oRow->url}\n";
-	}
-	$images = WikiaPhotoGalleryRSS::parseFeed($oRow->url, true);
```
macbre, line 39: `WikiaPhotoGalleryRSS` class does not exist.

ghost (approved):

PR #12758, opened 2017-04-06
XW-3127 | new right rail recirculation

latata:
For the test I just copied rail recirc module code. We gonna clean it up when we will productize it.

https://wikia-inc.atlassian.net/browse/XW-3127

@Wikia/x-wing


extensions/wikia/Recirculation/js/views/premiumRail.js
```diff
@@ -0,0 +1,59 @@
...
+
+		return curated.injectContent(data)
+			.then(renderTemplate('client/premiumRail.mustache'))
+			.then(utils.waitForRail)
+			.then(function ($html) {
+				if (options.before) {
+					$html = options.before($html);
+				}
+
+				$('#RECIRCULATION_RAIL').html($html);
```
rybmat, line 24: why this id is written with ALL_CAPS?

extensions/wikia/Recirculation/styles/premium-rail.scss
```diff
@@ -0,0 +1,64 @@
+@import 'skins/shared/color';
+@import 'skins/shared/mixins/transform';
+@import 'skins/shared/mixins/flexbox';
+
+.WikiaRail .premium-recirculation-rail {
+	border: 0;
```
rybmat, line 6: is there any difference between border: 0; and border: none; ?

rybmat (approved):
two minor comments but lgtm

extensions/wikia/Recirculation/styles/premium-rail.scss
latata: It's shorter ;)

extensions/wikia/Recirculation/js/views/premiumRail.js
latata: let's leave it for now

extensions/wikia/Recirculation/styles/premium-rail.scss
weronika: and we should use border: 0; according to our css coding standards

extensions/wikia/Recirculation/templates/client/premiumRail.mustache
```diff
@@ -0,0 +1,35 @@
...
+						{{#thumbnail}}
+							<img src="{{thumbnail}}" alt="{{title}}" itemprop="thumbnail">
+						{{/thumbnail}}
+					</div>
+					<div class="content">
+						<div class="title" title="{{title}}">{{title}}</div>
+
+							<div class="recirculation-sponsored-footer">
+								Sponsored by {{presented_by}}
+							</div>
```
weronika: Do we alway have `presented_by` var set? And why every second item has this div hidden? I would also recommend to use i18n string here (even if not translated).

weronika (changes requested):

extensions/wikia/Recirculation/templates/client/premiumRail.mustache
latata: That was just a temporary change for design review. Sorry I though that no one will review the code at the night. ;) I'd extract strings to i18n in recirc productization task.

weronika (approved):

PR #12765, opened 2017-04-07
ADEN-5009 Fix vpaid styles

fraszczakszymon:
It turns out that we should not override IMA's inline styles because some vpaids are styled base on those defaults. Instead of overriding IMA properties everywhere we can adjust styles for UAP units only and keep defaults for all standalone players.


dianafa (approved):

extensions/wikia/AdEngine/js/template/porvata.js
```diff
@@ -76,12 +76,17 @@ define('ext.wikia.adEngine.template.porvata', [
...
 		var container = doc.createElement('div'),
+			displayWrapper = doc.createElement('div'),
 			providerContainer = doc.querySelector('#' + slotName + ' > .provider-container');
 
-		container.classList.add('vpaid-container');
+		container.classList.add('video-overlay');
+		displayWrapper.classList.add('video-display-wrapper');
+
+		container.appendChild(displayWrapper);
+
```
dianafa: unnecessary \n here :p

dianafa (approved):
LGTM

PR #12767, opened 2017-04-07
ADEN-5004 New key-val for PageFair recovery

nandy-andy:
PageFair clean-up:
* added new key-val, so we can have separate collapse lines for PageFair recovery and SourcePoint,
* set default value for local variables.


nandy-andy:
Failing PHP unit tests are not connected to the changes and will get fixed once this PR is merged to `dev`: https://github.com/Wikia/app/pull/12764

fraszczakszymon (approved):

PR #12771, opened 2017-04-10
cake-546 - overtake the history action as well

nmonterroso:
@Wikia/cake 
https://wikia-inc.atlassian.net/browse/CAKE-546

See Wikia/sd-poc#314 for the changes on the CMS side to display history when embedded in MW


drsnyder:
LGTM

PR #12772, opened 2017-04-11
ADEN-4866 PageFair & SourcePoint small refactor & old dependency update

dianafa:
https://wikia-inc.atlassian.net/browse/ADEN-4866


skins/oasis/modules/templates/Body_Index.php
```diff
@@ -137,7 +137,7 @@
 					<?= $bodytext ?>
 
 				</div>
-				<?php if ( !(new ARecoveryModule)->isSourcePointRecoveryDisabled() ) { ?>
+				<?php if ( (new ARecoveryModule)->isSourcePointRecoveryEnabled() ) { ?>
```
nandy-andy: What do you think of making it consistent and using "PHP template conditions" instead of brackets everywhere in the file? :japanese_ogre:

nandy-andy:
I don't think we should revert now the changes in logic but it's definitely something I haven't seen and we can discuss it on parking lot tomorrow: we add assets and change response size sitewide if we set local variable by default to `true`.

dianafa:
@nandy-andy what do you mean by setting local var by default to `true`?

nandy-andy:
@dianafa `wgAdDriverEnableSourcePointRecovery` and `wgAdDriverEnablePageFairRecovery` are the local vars for me, `wgAdDriverSourcePointDetectionCountries` and `wgAdDriverPageFairRecoveryCountries` are (instant) global vars in my mind. More context: https://github.com/Wikia/app/pull/12767/files#diff-013f1f772182745f48c5f1429d0cdd35L1580

PR #12776, opened 2017-04-11
XW-3139 | Disable importing scripts when user JS is not allowed

rogatty:
## Links
* https://wikia-inc.atlassian.net/browse/XW-3139

## Description
Disallow use of ImportJS on pages like Special:Preferences.
It's a quick fix, I'll leave the refactoring suggested by @TK-999 to SUS.

## Reviewers
@Grunny @TK-999 @kvas-damian

### DIFF
--- a/extensions/wikia/ContentReview/ContentReview.hooks.php
+++ b/extensions/wikia/ContentReview/ContentReview.hooks.php
@@ -123,9 +123,9 @@ class Hooks {
 	 * @throws \MWException
 	 */
 	public function onSkinAfterBottomScripts( $skin, &$bottomScripts ) {
-		global $wgUseSiteJs;
+		global $wgOut, $wgUseSiteJs;
 
-		if ( !empty( $wgUseSiteJs ) ) {
+		if ( !empty( $wgUseSiteJs ) && $wgOut->isUserJsAllowed() ) {
 			$bottomScripts .= ( new ImportJS() )->getImportScripts();
 		}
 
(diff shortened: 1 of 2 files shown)


extensions/wikia/ContentReview/ContentReview.hooks.php
```diff
@@ -123,9 +123,9 @@ public function onBeforePageDisplay( \OutputPage $out, \Skin $skin ) {
 	 * @throws \MWException
 	 */
 	public function onSkinAfterBottomScripts( $skin, &$bottomScripts ) {
-		global $wgUseSiteJs;
+		global $wgOut, $wgUseSiteJs;
 
-		if ( !empty( $wgUseSiteJs ) ) {
+		if ( !empty( $wgUseSiteJs ) && $wgOut->isUserJsAllowed() ) {
```
mszabo-wikia: Here we can use the `RequestContext` provided by the `Skin` instance passed to the hook handler instead of relying on global state: `$skin->getOutput()`

mszabo-wikia (changes requested):

extensions/wikia/ContentReview/ContentReview.hooks.php
Grunny: You can avoid the global since `$skin` is a `ContextSource`, so this could be `$skin->getOutput()->isUserJsAllowed()`.

Grunny (commented):
Ideally we would move to use an RL module here as TK suggested, but ImportJS was a hacky feature to begin with and this looks like a nice and simple fix for now. :+1:

mszabo-wikia (approved):

Grunny (approved):

PR #12778, opened 2017-04-12
IRIS-4266 | Redirect to Mercury's Piggyback from Special:Piggyback

slayful:
https://wikia-inc.atlassian.net/browse/IRIS-4266

### DIFF
--- a/extensions/wikia/Piggyback/Piggyback_body.php
+++ b/extensions/wikia/Piggyback/Piggyback_body.php
@@ -17,7 +17,6 @@ class Piggyback extends SpecialPage {
 	function execute( $par ) {
 		global $wgRequest, $wgOut, $wgUser, $wgEnableMercuryPiggyback;
 
-
 		$this->logger->info( 'IRIS-4219 Piggyback has been rendered' );
 
 		if ( !empty( $par ) ) {
--- a/extensions/wikia/Piggyback/Piggyback_body.php
+++ b/extensions/wikia/Piggyback/Piggyback_body.php
@@ -17,14 +17,16 @@ class Piggyback extends SpecialPage {
 	function execute( $par ) {
 		global $wgRequest, $wgOut, $wgUser, $wgEnableMercuryPiggyback;
 
+
 		$this->logger->info( 'IRIS-4219 Piggyback has been rendered' );
 
 		if ( !empty( $par ) ) {
-			$wgRequest->setVal( 'target', $par );
+			$this->getRequest()->setVal( 'target', $par );
 		}
 
 		if ( $wgEnableMercuryPiggyback ) {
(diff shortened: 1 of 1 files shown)


extensions/wikia/Piggyback/Piggyback_body.php
```diff
@@ -12,12 +11,23 @@ function __construct() {
...
+		global $wgRequest, $wgOut, $wgUser, $wgEnableMercuryPiggyback;
+
+		$this->logger->info( 'IRIS-4219 Piggyback has been rendered' );
+
+		if ( !empty( $par ) ) {
+			$wgRequest->setVal( 'target', $par );
+		}
+
+		if ( $wgEnableMercuryPiggyback ) {
+			$this->redirectToMercuryPiggyback( $wgOut, $wgRequest->getVal( 'target' ) );
```
mszabo-wikia: You can use the local `RequestContext` here instead of globals, i.e. `$this->getRequest()` and `$this->getOutput()`

slayful:
Please don't merge until Sign out is fully deployed.

extensions/wikia/Piggyback/Piggyback_body.php
```diff
@@ -43,10 +49,27 @@ function execute( $par ) {
...
 
 		$LoginForm->render();
 	}
+
+	private function redirectToMercuryPiggyback( OutputPage $out, $target ) {
+		$redirectUrl = '/piggyback';
+		if ( !empty( $target ) ) {
+			$redirectUrl .= sprintf( "?target=%s", $target );
+		}
+		$out->redirect( $redirectUrl );
```
mszabo-wikia, line 62: Does Mercury piggyback handle authentication of user?

slayful: Indeed it does. Unauthorised users get redirected to sign in and upon signing in back to piggyback.

ghost (changes requested):
I would like the following questions to be answered:

1. Does Mercury's Piggyback rely on the legacy Piggyback code or is it a completely new and independent implementation?

2. I assume Mercury's Piggyback is supposed to replace the legacy Piggyback completely or partially (this will be clarified in the answer to the previous question). I do not see any legacy code removed. Why do we need to keep it?

slayful:
1. Mercury's Piggyback is a proxy on Helios' Piggyback and it is a new implementation.
2. If the variable is on, then it will redirect from the Piggyback form to Mercury, so it's effectively disabled (unless users do POST calls manually). If there are no issues reported with the new Piggyback, then old Piggyback will be up for removal. I will be creating a follow up ticket for that.

ghost (approved):

mszabo-wikia (approved):

PR #12793, opened 2017-04-14
XW-3077 | Purge mcache for premium page header on MediaWiki:Wikia-navigation edit

rybmat:
https://wikia-inc.atlassian.net/browse/XW-3077

@Wikia/x-wing

### DIFF
--- a/includes/wikia/models/NavigationModel.class.php
+++ b/includes/wikia/models/NavigationModel.class.php
@@ -82,6 +82,15 @@ class NavigationModel extends WikiaModel {
 		WikiaDataAccess::cachePurge(
 			$this->getMemcKey( $key )
 		);
+
+		// Purging mcache set by invocation of NavigationModel::getTree() by NavigationModel::getLocalNavigationTree();
+		WikiaDataAccess::cachePurge(
+			$this->getTreeMemcKey( self::TYPE_MESSAGE, self::WIKI_LOCAL_MESSAGE, implode( [
+				self::LOCALNAV_LEVEL_1_ITEMS_COUNT,
+				self::LOCALNAV_LEVEL_2_ITEMS_COUNT,
+				self::LOCALNAV_LEVEL_3_ITEMS_COUNT
+			], '-' ) )
+		);
 	}
 
 	private function setShouldTranslateContent( $shouldTranslateContent ) {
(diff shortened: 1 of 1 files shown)


weronika (approved):

includes/wikia/models/NavigationModel.class.php
```diff
@@ -82,6 +82,15 @@ public function clearMemc( $key = self::WIKIA_GLOBAL_VARIABLE ) {
...
 			$this->getMemcKey( $key )
 		);
+
+		// Purging mcache set by invocation of NavigationModel::getTree() by NavigationModel::getLocalNavigationTree();
+		WikiaDataAccess::cachePurge(
+			$this->getTreeMemcKey( self::TYPE_MESSAGE, self::WIKI_LOCAL_MESSAGE, implode( [
+				self::LOCALNAV_LEVEL_1_ITEMS_COUNT,
+				self::LOCALNAV_LEVEL_2_ITEMS_COUNT,
+				self::LOCALNAV_LEVEL_3_ITEMS_COUNT
+			], '-' ) )
```
rogatty: http://php.net/manual/en/function.implode.php
`glue` should be the first parameter. I got confused when PhpStorm showed me `pieces: '-'` :smile:

rogatty: Also, `getTreeMemcKey` already implodes the method arguments. Just pass in all the ingredients.

includes/wikia/models/NavigationModel.class.php
```diff
@@ -82,6 +82,15 @@ public function clearMemc( $key = self::WIKIA_GLOBAL_VARIABLE ) {
 		WikiaDataAccess::cachePurge(
 			$this->getMemcKey( $key )
 		);
+
+		// Purging mcache set by invocation of NavigationModel::getTree() by NavigationModel::getLocalNavigationTree();
```
rogatty, line 86: What about `NavigationModel::getOnTheWikiNavigationTree`? Should we purge that cache too?

includes/wikia/models/NavigationModel.class.php
rybmat: according to the doc you linked `implode() can, for historical reasons, accept its parameters in either order. For consistency with explode(), however, it may be less confusing to use the documented order of arguments.`

includes/wikia/models/NavigationModel.class.php
rybmat: this method is not used at all, and I wonder if we should still keep it or just remove

rogatty: It's not used for a long time, see https://github.com/Wikia/app/commit/2f0c83fa704a81c9f2880fd8591b84c8fa540016. Let's remove it.

rogatty (approved):

PR #12794, opened 2017-04-14
ST-273 - Fix Search Suggest to handle Diacritical Marks 

jogura-zz:
* Convert both the query string and the page titles to a common binary format for the `LIKE` comparison
* Same results now are returned when the query uses the diacritical marks or when it doesn't (i.e., "pokémon trading" or "pokemon trading")


garthwebb:
@mixth-sense Can you take a look?

ghost (approved):

PR #12798, opened 2017-04-14
cake-1585 - rewrite Title urls with the text rather than the db key

nmonterroso:
@Wikia/cake 
https://wikia-inc.atlassian.net/browse/CAKE-1585

This fixes both `Special:Random` redirects, navigation links, and possibly other places that generate links to pages from the `Title` objects.

There's a 3 hour cache on the wiki nav, so we'd have to wait for that once this code is pushed out.


drsnyder:
LGTM

PR #12804, opened 2017-04-18
ADEN-5042

v3rth:


fraszczakszymon (approved):

extensions/wikia/AdEngine/js/lookup/rubicon/rubiconFastlane.js
```diff
@@ -87,7 +87,7 @@ define('ext.wikia.adEngine.lookup.rubicon.rubiconFastlane', [
 			'970x250': 57,
 			'320x480': 67
 		},
-		slots;
+		slots = [];
```
fraszczakszymon: Why is it needed? Shouldn't it be an object?

dianafa (approved):

PR #12807, opened 2017-04-18
Cake 1586 - new page header support

nmonterroso:
@Wikia/cake 
https://wikia-inc.atlassian.net/browse/CAKE-1586

The MW changes for Wikia/sd-poc#322.

The new page header needs the title set in the `OutputPage`, so setting that in `CPArticleRenderer`. Also sending over the enabled status of the new page header to the node server via header so the initial state is correct.


drsnyder:
LGTM

PR #12808, opened 2017-04-18
ADEN-5041 Fix PF so it doesn't fire when a user does not have an ad blocking software

dianafa:
https://wikia-inc.atlassian.net/browse/ADEN-5041


extensions/wikia/ARecoveryEngine/js/adBlockDetection.js
```diff
@@ -1,38 +1,32 @@
 /*global define*/
-define('ext.wikia.aRecoveryEngine.recovery.sourcePoint', [
+define('ext.wikia.aRecoveryEngine.adBlockDetection', [
```
fraszczakszymon: Just wonder if it's safe to use this name :>

extensions/wikia/ARecoveryEngine/js/SourcePoint/sourcePoint.js
```diff
@@ -0,0 +1,52 @@
+/*global define*/
+define('ext.wikia.aRecoveryEngine.recovery.sourcePoint', [
```
fraszczakszymon: Can we keep the same directory structure?

extensions/wikia/ARecoveryEngine/js/adBlockDetection.js
fraszczakszymon: IMO this module has too many responsibilities. I'd move `isEnabled`=`isRecoveryEnabled`, `getSafeUri`, `verifyContent`, `track` to another, second abstraction - "recovery". Then AdEngineRunner can really depnend on recovery (which depends on detection) instead of detection directly. Let's keep detection abstraction away from user facing changes

extensions/wikia/AdEngine/js/provider/gpt/helper.js
```diff
@@ -55,48 +57,47 @@ define('ext.wikia.adEngine.provider.gpt.helper', [
...
 
 		setAdditionalTargeting(slotTargetingData);
 
-		element = new AdElement(slot.name, slotPath, slotTargetingData);
+		element = new AdElement(slotName, slotPath, slotTargetingData);
+
+		// add adonis marker needed for PF recovery
+		// basing on extra.isPageFairRecoverable param set in factoryWikiaGpt
+		if (isRecoveryEnabled && isBlocking && extra.isPageFairRecoverable) {
+			pageFair.addMarker(element.node);
```
fraszczakszymon: Consider moving it to `factoryWikiaGpt`

extensions/wikia/AdEngine/js/provider/yavliTag.js
```diff
@@ -1,18 +1,18 @@
 /*global define*/
 define('ext.wikia.adEngine.provider.yavliTag', [
```
fraszczakszymon: What do you think about removing whole yavli? Worth to ask Nef/Nandy and make some small clean up :)

extensions/wikia/AdEngine/js/slot/revcontentSlots.js
```diff
@@ -1,10 +1,10 @@
 /*global define*/
 define('ext.wikia.adEngine.slot.revcontentSlots', [
```
fraszczakszymon: What do you think about removing whole revcontent? Worth to ask Nef/Nandy and make some small clean up :)

extensions/wikia/AssetsManager/config.php
```diff
@@ -114,10 +114,10 @@
...
 		'//extensions/wikia/AdEngine/js/video/uapVideo.js',
 		'//extensions/wikia/AdEngine/js/video/videoSettings.js',
-		'//extensions/wikia/ARecoveryEngine/js/sourcePoint.js',
-		'//extensions/wikia/ARecoveryEngine/js/slotFinder.js',
-		'//extensions/wikia/ARecoveryEngine/js/recovery/tweaker.js',
-		'//extensions/wikia/ARecoveryEngine/js/sourcePointRecovery.js',
+		'//extensions/wikia/ARecoveryEngine/js/SourcePoint/sourcePoint.js',
+		'//extensions/wikia/ARecoveryEngine/js/SourcePoint/slotFinder.js',
+		'//extensions/wikia/ARecoveryEngine/js/tweaker.js',
+		'//extensions/wikia/ARecoveryEngine/js/adBlockDetection.js',
```
fraszczakszymon: Keep alphabetical order please

extensions/wikia/ARecoveryEngine/js/SourcePoint/slotFinder.js
```diff
@@ -1,5 +1,5 @@
 /*global define*/
-define('ext.wikia.aRecoveryEngine.recovery.slotFinder', [
+define('ext.wikia.aRecoveryEngine.recovery.sourcePoint.sourcePoint.slotFinder', [
```
fraszczakszymon: Why double `sourcePoint`?

extensions/wikia/ARecoveryEngine/js/adBlockDetection.js
```diff
@@ -0,0 +1,62 @@
...
+		onBlockingEventsQueue.push(callback);
+	}
+
+	function addOnNotBlockingCallback(callback) {
+		onNotBlockingEventsQueue.push(callback);
+	}
+
+	function isBlocking() {
+		return isDetectionEnabled() &&
+			(sourcePoint.isBlocking() || (pageFair && pageFair.isBlocking()));
```
fraszczakszymon: Do we want to rely on pageFair here?

extensions/wikia/ARecoveryEngine/js/adBlockDetection.js
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	function isDetectionEnabled() {
```
fraszczakszymon: s/isDetectionEnabled/isEnabled

extensions/wikia/ARecoveryEngine/js/adBlockRecovery.js
```diff
@@ -0,0 +1,95 @@
...
+
+		return url;
+	}
+
+	function isEnabled() {
+		var context = adContext.getContext(),
+			enabled = context.opts.sourcePointRecovery || context.opts.pageFairRecovery;
+
+		log(['isRecoveryEnabled', enabled], log.levels.debug, logGroup);
+		return enabled;
```
fraszczakszymon: `return adBlockDetection.isEnabled() && enabled`

extensions/wikia/ARecoveryEngine/templates/ARecoveryEngineApiController_getPageFairBootstrap.mustache
```diff
@@ -1,13 +1,10 @@
...
-		if (context.opts.pageFairRecovery === true && !window.wgUserName) {
-			{{{code}}}
-		}
-	});
+	if (!window.wgUserName &&
+			window.ads &&
+			window.ads.context &&
+			window.ads.context.opts.pageFairRecovery === true) {
+		{{{code}}}
+	}
```
fraszczakszymon: IMO it does not solve the issue, it hides the race condition. Instead of throwing an error it won't enable pagefair recovery which will be harder to notice. What do you think about switching this line:
https://github.com/Wikia/app/blob/dev/skins/oasis/modules/templates/Oasis_Index.php#L37
(with 38)

extensions/wikia/AdEngine/AdEngine2.setup.php
```diff
@@ -51,7 +50,6 @@
 
 AdEngine2Resource::register('wikia.ext.adengine.pf.detection', 'ResourceLoaderAdEnginePageFairDetectionModule');
 AdEngine2Resource::register('wikia.ext.adengine.sp.detection', 'ResourceLoaderAdEngineSourcePointDetectionModule');
-AdEngine2Resource::register('wikia.ext.adengine.yavli', 'ResourceLoaderAdEngineYavliModule');
```
fraszczakszymon, line 54: :gem:

extensions/wikia/ARecoveryEngine/js/PageFair/pageFair.js
```diff
@@ -33,10 +33,10 @@ define('ext.wikia.aRecoveryEngine.recovery.pageFair', [
 		return isBlocking;
 	}
 
-	function addMarker(slotId) {
-		var slot = document.getElementById(slotId);
-
-		slot.setAttribute('adonis-marker', '');
+	function addMarker(slotElement) {
+		if (slotElement) {
```
nandy-andy: Is the `if` really needed? We always [create an element](https://github.com/Wikia/app/pull/12808/files#diff-d0456a56fa8978685b777a3c10aaf5abR88) before we call `addMarker()` with it, don't we?

extensions/wikia/ARecoveryEngine/js/PageFair/pageFair.js
```diff
@@ -1,11 +1,11 @@
-define('ext.wikia.aRecoveryEngine.recovery.pageFair', [
+define('ext.wikia.aRecoveryEngine.pageFair.pageFair', [
```
nandy-andy: What are the reasons to change the module name? Oh, probably it is supposed to be mapped to the path, isn't it? [It looks the same with SourcePoint](https://github.com/Wikia/app/pull/12808/files#diff-c55e0a194e1a5eb559d4321c8e034aeeR2). I'm not a big fan but it's not a big issue, I guess. @fraszczakszymon, @v3rth what do you think?

extensions/wikia/ARecoveryEngine/js/SourcePoint/sourcePoint.js
```diff
@@ -0,0 +1,52 @@
...
+			'GPT_FLUSH'
+		];
+
+	/**
+	 * SourcePoint can be enabled only if PF recovery is disabled
+	 *
+	 * @returns {boolean}
+	 */
+	function isEnabled() {
+		var enabled = !!context.opts.sourcePointRecovery && !context.opts.pageFairRecovery;
```
nandy-andy: Let's make sure we document this logic outside the code: it's possible to overwrite SourcePoint recovery by enabling PageFair recovery :smiley_cat:

extensions/wikia/ARecoveryEngine/js/SourcePoint/slotFinder.js
```diff
@@ -1,5 +1,5 @@
 /*global define*/
-define('ext.wikia.aRecoveryEngine.recovery.slotFinder', [
+define('ext.wikia.aRecoveryEngine.sourcePoint.slotFinder', [
```
nandy-andy, line 2: Shouldn't we keep an old version until cache invalidates?

extensions/wikia/ARecoveryEngine/js/tweaker.js
```diff
@@ -1,22 +1,22 @@
 /*global define*/
-define('ext.wikia.aRecoveryEngine.recovery.tweaker', [
+define('ext.wikia.aRecoveryEngine.tweaker', [
```
nandy-andy, line 2: What about old version cached?

extensions/wikia/ARecoveryEngine/js/PageFair/pageFair.js
dianafa: yes, we should, but my thinking was not to throw an error in case we'll call addMarker and `doc.getElementById()` which will return null. But this may be overprotective and totally unneeded here. No strong feelings about that from my side.

extensions/wikia/ARecoveryEngine/js/SourcePoint/sourcePoint.js
dianafa: https://wikia-inc.atlassian.net/wiki/pages/diffpagesbyversion.action?pageId=66551887&selectedPageVersions=33&selectedPageVersions=32

extensions/wikia/ARecoveryEngine/js/SourcePoint/slotFinder.js
dianafa: I'd add fallback only for `ext.wikia.aRecoveryEngine.recovery.sourcePoint` due to this change: https://github.com/Wikia/app/pull/12808/files#diff-939a1bcfebcc0e82ae00ac908560e39cR22

extensions/wikia/ARecoveryEngine/js/PageFair/pageFair.js
nandy-andy: Let's remove it then and ask @bulaszenko to do his best to find a way for throwing the error ;) A weak compromise is to add `else` with logging inside. So, once we notice no markers added when all conditions are met, we'll add logging params and see if a lack of element is an issue.

extensions/wikia/ARecoveryEngine/js/SourcePoint/slotFinder.js
nandy-andy: Right, I alarmed "automatically" because I saw the rename. Sorry about that and thanks! :)

nandy-andy (approved):

extensions/wikia/ARecoveryEngine/js/PageFair/pageFair.js
v3rth: IMO pageFair.pageFair looks terrible :D even if it's based on the directory structure, but I don't have a strong argument against than "I don't like it".

Maybe the problem is that last "pageFair" isn't the best and most described name. What do you thing about rename it? If there is no good enough name, maybe it's too big (it has too many responsibilities)

extensions/wikia/ARecoveryEngine/js/adBlockDetection.js
```diff
@@ -0,0 +1,57 @@
+/*global define*/
+define('ext.wikia.aRecoveryEngine.adBlockDetection', [
```
v3rth, line 2: isn't it only SP detection? Let's move it to SP directory

extensions/wikia/ARecoveryEngine/js/adBlockRecovery.old.js
```diff
@@ -0,0 +1,100 @@
...
+					action: Wikia.Tracker.ACTIONS.IMPRESSION,
+					label: type,
+					trackingMethod: 'analytics'
+				});
+			}
+			if (instantGlobals.wgARecoveryEngineCustomLog) {
+				try {
+					var xmlHttp = new XMLHttpRequest();
+					xmlHttp.open('GET', customLogEndpoint + type, true);
+					xmlHttp.send();
```
v3rth, line 71: Please, 'pack it' minimum in some function (even whole ```try ... catch```) it will be much cleaner

extensions/wikia/ARecoveryEngine/js/spec/adBlockDetection.spec.js
```diff
@@ -0,0 +1,70 @@
...
+			mocks.sourcePoint,
+			null,
+			null,
+			mocks.lazyQueue,
+			mocks.log,
+			null
+		);
+	}
+
+	it('isBlocking is true when sourcePointDetection is enabled and SP is blocking', function () {
```
v3rth: Shouldn't "and SP is blocking" be replaced with "SP detects blocking user" or something similar?

extensions/wikia/AdEngine/js/provider/gpt/googleTag.js
```diff
@@ -35,40 +35,52 @@ define('ext.wikia.adEngine.provider.gpt.googleTag', [
...
-			pageFairRecoveryEnabled = pageFair && pageFair.isEnabled(),
-			// load GPT when API not loaded yet, ads are not blocked(SP detection) or recovery is PageFair
-			// as it needs gpt to work
-			gptCanBeLoaded = !window.googletag.apiReady &&
-				(!sourcePoint.isBlocking() || pageFairRecoveryEnabled);
-
-		if (gptCanBeLoaded) {
+			node = doc.getElementsByTagName('script')[0];
+
+		if (canGptBeLoaded()) {
```
v3rth, line 83: <3 <3 <3

extensions/wikia/VideoHandlers/js/handlers/Ooyala.js
```diff
@@ -165,9 +166,9 @@ define('wikia.videohandler.ooyala', [
 				log('All Ooyala assets loaded', log.levels.info, logGroup);
 
 				win.OO.ready(function () {
-					if (sourcePoint.isEnabled()) {
-						sourcePoint.addOnBlockingCallback(initRecoveredPlayer);
-						sourcePoint.addOnNotBlockingCallback(initRegularPlayer);
+					if (adBlockDetection.isEnabled()) {
+						adBlockDetection.addOnBlockingCallback(initRecoveredPlayer);
```
v3rth, line 170: IMO it's still only SP not detection in general,  so the name of module isn't adequate

extensions/wikia/ARecoveryEngine/js/adBlockDetection.js
dianafa: the point was to create an interface for all detection no matter what's underneath. For now it's SP only, but it can be anything and the point of this module is just to deliver information if user is blocking ads, no matter who is detecting that

v3rth (approved):

PR #12817, opened 2017-04-19
IRIS-4298 | Remove piggyback code

slayful:
https://wikia-inc.atlassian.net/browse/IRIS-4298

After the introduction of /piggyback in Mercury's auth pages we can now remove Piggyback code in Media Wiki.

We'll not removing the Special Page itself, because it's referenced in multiple places. Instead this page redirects to our new piggyback page.


extensions/wikia/Piggyback/Piggyback.i18n.php
```diff
@@ -3,516 +3,126 @@
...
-	'piggyback-nosuchuser' => 'There is no user by the name "$1".
-User names are case sensitive.
-Check your spelling.',
-	'piggyback-wronguser' => '$1 is not your user name',
-	'piggyback-loginprompt' => 'Please re-enter your username and password. Warning: This will run ALL custom user-account CSS and JS on your machine. Be careful!',
-	'piggyback-firstlogout' => 'Please first logout back to your account.',
-	'piggyback-itisyou' => 'You cannot piggyback yourself.',
-	'piggyback-issysopto' => 'You cannot piggyback a user with the right to piggyback.',
-	'action-piggyback' => 'log on as another user',
-	'log-name-piggyback' => 'Piggyback log',
```
Grunny, line 18: The log seems to be left in the setup file (for historical reference?), so should these messages stay as well for the log page?

Grunny (commented):
One question, looks OK otherwise.

extensions/wikia/Piggyback/Piggyback_body.php
```diff
@@ -4,54 +4,23 @@
 use Wikia\Service\User\Auth\CookieHelper;
```
ghost: These imports/aliases are no longer needed as the code using them has been removed. I have checked it in the previous revision of this file, but a second pair of eyes will not hurt.

ghost (approved):
I have made a comment about removing imports/aliases.

Other than that, if there is more special pages to be handled in a similar way - redirected to their Mercury counterparts - it might be a good idea to abstract it instead of repeating similar logic, e.g. ` class MercurySpecialPage extends SpecialPage` and `new MercurySpecialPage('Piggyback')` or `new MercurySpecialPage('ListUsers')`, etc.

If there is 100 special pages and there is just one "more special", a simple class member, such as `$isMercury = true;` works fine. But if there is fifty-fifty, we are dealing with a whole "class" of special pages, literally. So introducing a separate abstract class makes much more sense, I think.

drsnyder:
Please provide a [description](https://github.com/Wikia/guidelines/blob/master/Process/PullRequests.md).

extensions/wikia/Piggyback/Piggyback.i18n.php
slayful: To be on the safe side let's not remove it. Thank you!

extensions/wikia/Piggyback/Piggyback_body.php
slayful: IntelliJ says they're not used. Thanks for spotting that!

slayful:
@mixth-sense I like your suggestion. I've been working on disabling sign in and sign out endpoints, and the code is similar. It's not exactly the same so I don't believe extending and abstract class would be appropriate. I'll extract common code to a separate class though.

ghost (approved):

ghost (approved):

PR #12829, opened 2017-04-21
ADEN-5054 check isBlocking afer we have info about blocking

dianafa:
https://wikia-inc.atlassian.net/browse/ADEN-5054


v3rth (approved):

extensions/wikia/AdEngine/js/AdInfoTracker.js
```diff
@@ -30,19 +30,23 @@ define('ext.wikia.adEngine.adInfoTracker',  [
 		};
 
 	function shouldHandleSlot(slot) {
```
nandy-andy: What do you think of making it public and unit testing it?

nandy-andy (approved):

PR #12830, opened 2017-04-21
XW-3137 | ArticleAsJson skip not existing media objects

kvas-damian:
https://wikia-inc.atlassian.net/browse/XW-3137

In articleAsJson not existing media doesn't provide any value. We just skip them in output `context` and `media` array.

Ping @Wikia/x-wing


idradm (approved):

rogatty:
Should we do the same for [`onGalleryBeforeProduceHTML`](https://github.com/Wikia/app/blob/aa9ca3fe21831daaa342167c96f146fb612d9006/extensions/wikia/ArticleAsJson/ArticleAsJson.class.php#L215)?

rogatty (approved):

PR #12831, opened 2017-04-21
ADEN-5069 Disable PF if SP MMS is enabled

v3rth:


dianafa (approved):

v3rth:
Won't do, follow up ticket created

PR #12863, opened 2017-04-30
Fixing I18N (es)

xaeky:


mszabo-wikia:
Thanks for the PR! The translators have been notified about the errors.

PR #12890, opened 2017-05-11
ADEN-5107 Limit providers queue on premium-only pages

fraszczakszymon:


extensions/wikia/AdEngine/js/context/adContext.js
```diff
@@ -79,6 +79,9 @@ define('ext.wikia.adEngine.adContext', [
 			}
 		}
 
+		context.opts.premiumOnly = context.targeting.hasFeaturedVideo &&
+			geo.isProperGeo(instantGlobals.wgAdDriverSrcPremiumCountries);
```
nandy-andy, line 83: Do we have the variable created in WikiFactory (dev, prod)?

nandy-andy (approved):

extensions/wikia/AdEngine/js/context/adContext.js
nandy-andy: It was for prod, I've added it for PL devboxes.

PR #12902, opened 2017-05-12
ST-311 - Add v1 template to ESFandomSearchService Request

jogura-zz:
Pin the version of the right rails search to the v1 version of the search
template.

### DIFF
--- a/extensions/wikia/Search/classes/Services/ESFandomSearchService.php
+++ b/extensions/wikia/Search/classes/Services/ESFandomSearchService.php
@@ -2,8 +2,8 @@
 
 namespace Wikia\Search\Services;
 
-use \Wikia\Logger\WikiaLogger;
-use \Wikia\Service\Gateway\ConsulUrlProvider;
+use Wikia\Logger\WikiaLogger;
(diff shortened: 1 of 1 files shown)


extensions/wikia/Search/classes/Services/ESFandomSearchService.php
```diff
@@ -17,6 +17,7 @@ class ESFandomSearchService extends AbstractSearchService {
 	const STORIES_IMAGE_URL_KEY = 'image';
 	const STORIES_URL_KEY = 'url';
 	const STORIES_COUNT_MAX = 6;
+	const STORIES_SEARCH_TEMPLATE = 'fandom_stories';
```
jimmordino, line 20: Suggestion: Can we make this a WikiFactory variable?  It would make it easier to change this next time.

jogura-zz: I wonder if there is going to be a lot of work around this module. I would say let's add the WF variable if there is future work involved.

jimmordino (approved):
LGTM

PR #13035, opened 2017-06-01
Added missing crowdin projects and removed unsupported variants

jcellary:
Added missing crowdin projects and removed unsupported variants. Converted pt-br to pt.


jcellary:
Reviewed and accepted by Keiko

PR #13061, opened 2017-06-06
Chat.i18n.php: add chat-blocklist-* messages for Special:ChatBanList

macbre:
* introduce `chat-blocklist-*` messages to be used on Special:ChatBanList to be able to tweak them without altering core MW messages used on Special:BlockList
* HTML / UI improvements - add a label of user name field and auto focus on it when the page is loaded

https://wikia-inc.atlassian.net/browse/SUS-468


sqreek (dismissed):
Are added messages should be added to `qqq` section also?

macbre:
@sqreek - `qqq` docs added in 4e604f4

sqreek (approved):

PR #13074, opened 2017-06-07
SUS-1594: AsyncTaskList - $wgUser might not yet be initialized, fallback to IP

mszabo-wikia:
Background task added in https://github.com/Wikia/app/pull/12983 sometimes gets called before `$wgUser` global is initialized which breaks the task queue - provide a sane default for this scenario.

https://wikia-inc.atlassian.net/browse/SUS-1594

PR #13077, opened 2017-06-08
Revert "SUS-1594: AsyncTaskList - $wgUser might not yet be initialized, fallback to IP"

sqreek:
Reverts Wikia/app#13074


macbre:
Was causing:

```
Unexpected non-MediaWiki exception encountered, of type "Error"
Error: Call to a member function getCanonicalName() on null in /usr/wikia/slot1/17923/src/includes/User.php:915
```

macbre (approved):

PR #13087, opened 2017-06-08
XW-3550 | css alignment adjustments for Page Header

weronika:
https://wikia-inc.atlassian.net/browse/XW-3550


latata (approved):

extensions/wikia/PageHeader/styles/index.scss
```diff
@@ -49,14 +48,15 @@
 	}
 
 	&__languages {
-		padding-bottom: $text-dropdown-padding-bottom;
+		margin-bottom: 12px;
```
rybmat, line 51: @weronika why did you change it from 10px back to 12px? it is not consistent with categories dropdown

PR #13099, opened 2017-06-09
cake-1794 - fix tracking for recirc footer items

nmonterroso:
@Wikia/cake 
https://wikia-inc.atlassian.net/browse/CAKE-1794

looks like tracking uses the `data-meta` attribute in recirc items to set up its tracking. This wasn't set in the footer template, so tracking wasn't working for items in that module.


gbenson-wikia:
Verified on my devbox.  Once tests pass, LGTM.

PR #13132, opened 2017-06-14
Re-branding (fixing i18n blogs)

xaeky:


mszabo-wikia:
Thank you for your contribution! The Localization team is busy with the rebranding effort, and we would like to make sure the changes are tracked and centralized.

PR #13136, opened 2017-06-16
XW-3597 | support for italic in page title

latata:
https://wikia-inc.atlassian.net/browse/XW-3597

@Wikia/x-wing

### DIFF
--- a/extensions/wikia/PageHeader/styles/index.scss
+++ b/extensions/wikia/PageHeader/styles/index.scss
@@ -119,6 +119,10 @@
 		line-height: 1;
 		max-width: 100%;
 		word-wrap: break-word;
+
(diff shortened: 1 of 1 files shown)


JamesHawking (approved):

extensions/wikia/PageHeader/styles/index.scss
```diff
@@ -119,6 +119,10 @@
 		line-height: 1;
 		max-width: 100%;
 		word-wrap: break-word;
+
+		i {
```
rybmat, line 123: is the italic style only one that we should support? did you ask stanley/tor/julia about this?

rybmat (approved):

PR #13139, opened 2017-06-16
XW-3612 | Language selector visible on all pages that have interlanguage links

latata:
https://wikia-inc.atlassian.net/browse/XW-3612
@Wikia/x-wing

### DIFF
--- a/extensions/wikia/LocalSitemapPage/LocalSitemapPage.setup.php
+++ b/extensions/wikia/LocalSitemapPage/LocalSitemapPage.setup.php
@@ -12,4 +12,3 @@ $wgExtensionMessagesFiles['LocalSitemapPage'] = __DIR__ . '/LocalSitemapPage.i18
 // hooks
 $wgHooks['ArticleFromTitle'][] = 'LocalSitemapPageHooks::onArticleFromTitle';
 $wgHooks['PageHeaderActionButtonShouldDisplay'][] = 'LocalSitemapPageHooks::onPageHeaderActionButtonShouldDisplay';
-$wgHooks['PageHeaderLanguageSelectorShouldDisplay'][] = 'LocalSitemapPageHooks::onPageHeaderLanguageSelectorShouldDisplay';
--- a/extensions/wikia/PageHeader/Languages.class.php
+++ b/extensions/wikia/PageHeader/Languages.class.php
@@ -10,7 +10,6 @@ class Languages {
 	public $languageList;
 
 	private $title;
-	private $shouldDisplay;
 
 	/**
 	 * Language constructor.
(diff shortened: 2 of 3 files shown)


extensions/wikia/LocalSitemapPage/LocalSitemapPage.setup.php
```diff
@@ -12,4 +12,3 @@
 // hooks
 $wgHooks['ArticleFromTitle'][] = 'LocalSitemapPageHooks::onArticleFromTitle';
 $wgHooks['PageHeaderActionButtonShouldDisplay'][] = 'LocalSitemapPageHooks::onPageHeaderActionButtonShouldDisplay';
-$wgHooks['PageHeaderLanguageSelectorShouldDisplay'][] = 'LocalSitemapPageHooks::onPageHeaderLanguageSelectorShouldDisplay';
```
rybmat, line 15: why did you removed this hook? it was on purpose to not display the language on local sitemap page

rybmat (changes requested):

extensions/wikia/LocalSitemapPage/LocalSitemapPage.setup.php
latata: Currently it's not necessary as we only show language selector for pages where `count( $this->languageList ) > 0` is `true`. On local sitemap it's true.

rybmat: 👍

rybmat (approved):

extensions/wikia/PageHeader/Languages.class.php
```diff
@@ -21,10 +21,7 @@ public function __construct( WikiaApp $app ) {
 		$this->title = \RequestContext::getMain()->getTitle();
 		$this->currentLangName = \Language::getLanguageName( $this->title->getPageLanguage()->getCode() );
 		$this->languageList = $this->handleLanguages( $app );
-
-		$shouldDisplay = $this->title->isContentPage();
-		wfRunHooks( 'PageHeaderLanguageSelectorShouldDisplay', [$this->title, &$shouldDisplay] );
-		$this->shouldDisplay = $shouldDisplay;
+		$this->shouldDisplay = count( $this->languageList ) > 0;
```
rogatty: This plus:
```
 	public function shouldDisplay(): bool {
 		return $this->shouldDisplay && count( $this->languageList ) > 0;
 	}
```
don't make sense together.

rybmat: heh, true, didn't notice that. @rogatty it is fixed now

weronika (approved):

PR #13165, opened 2017-06-22
CAKE-1657 Set the title using X-Page-Title from the FC

drsnyder:
https://wikia-inc.atlassian.net/browse/CAKE-1657

This PR changes the CP article rendering to use the title supplied by the FC to set the title. This prevents the brief display of the title with the id prefixed before it's replaced during the FC rendering.

Here we: 
  * change getArticleContent to getArticle
  * change the return type of getArticle from a string to a CPArticle object
  * use the CPArticle object to set the title and content

The changes in the FC to make this possible are here https://github.com/Wikia/sd-poc/pull/413.

@Wikia/cake

### DIFF
--- a/extensions/wikia/ContributionPrototype/CPArticle.php
+++ b/extensions/wikia/ContributionPrototype/CPArticle.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace ContributionPrototype;
+
+class CPArticle {
+
+	private $content;
+	private $entityName;
+
+	function __construct($content, $entityName) {
+		$this->content = $content;
+		$this->entityName = $entityName;
+	}
+
+	public function getContent() {
+		return $this->content;
+	}
+
+	public function getEntityName() {
+		return $this->entityName;
+	}
+}
--- a/extensions/wikia/ContributionPrototype/CPArticle.php
+++ b/extensions/wikia/ContributionPrototype/CPArticle.php
@@ -5,18 +5,18 @@ namespace ContributionPrototype;
 class CPArticle {
 
 	private $content;
-	private $entityName;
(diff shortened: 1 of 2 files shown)


extensions/wikia/ContributionPrototype/CPArticleRenderer.php
```diff
@@ -131,6 +135,6 @@ private function getArticleContent($title, $action) {
 			return false;
 		}
 
-		return $response->getContent();
+		return new CPArticle($response->getContent(), $response->getResponseHeader('X-Page-Title'));
```
nmonterroso: What does `getResponseHeader` if the header isn't present?

extensions/wikia/ContributionPrototype/CPArticle.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+namespace ContributionPrototype;
+
+class CPArticle {
+
+	private $content;
+	private $entityName;
```
nmonterroso: nitpicky - can we name this `title`?

nmonterroso (approved):
minor comments lgtm

extensions/wikia/ContributionPrototype/CPArticleRenderer.php
drsnyder: It returns `null` which works as expected as input to `empty()`.

extensions/wikia/ContributionPrototype/CPArticle.php
drsnyder: Sure, changed in ee6144f39eb.

PR #13233, opened 2017-07-04
SUS-2321: Sunset LyricsMinimal skin

mszabo-wikia:
https://wikia-inc.atlassian.net/browse/SUS-2321


macbre:
Let's double check with the mobile apps team if this skin is not used by them.

mszabo-wikia:
@macbre Mobile team does not know this skin, so probably safe to delete :D

PR #13249, opened 2017-07-05
DE-2724 Removed unsupported languages from global messages

jcellary:
part of https://wikia-inc.atlassian.net/browse/DE-2724 

Removed unsupported language varaints from global messages.

Pinging Keiko for review.


jcellary:
Reviewed and accepted by Keiko

PR #13266, opened 2017-07-07
DE-2724 Replaced Fandom with FANDOM in Global Messages

jcellary:
https://wikia-inc.atlassian.net/browse/DE-2724

Pinging Keiko for review


jcellary:
accepted by Keiko

PR #13274, opened 2017-07-07
XW-3650 | Featured Video on scroll unit should have priority over other page elements

JamesHawking:
[Ticket](https://wikia-inc.atlassian.net/browse/XW-3650)

We decided to hide site wide messages on Featured Video pages so they do not interfere with video experience.

@Wikia/x-wing

### DIFF
--- a/extensions/wikia/SiteWideMessages/SiteWideMessagesController.class.php
+++ b/extensions/wikia/SiteWideMessages/SiteWideMessagesController.class.php
@@ -8,10 +8,10 @@ class SiteWideMessagesController extends WikiaController {
 	private function shouldShowSiteWideMessage() {
 		global $wgEnableArticleFeaturedVideo;
 
-		$articleTitle = Title::newFromID( $this->getVal( 'articleId' ) );
-		if ( !is_null( $articleTitle ) ) {
+		$title = Title::newFromID( $this->getVal( 'articleId' ) );
+		if ( !is_null( $title ) ) {
 			$isFeaturedVideoEmbedded = !empty( $wgEnableArticleFeaturedVideo ) &&
-				ArticleVideoContext::isFeaturedVideoEmbedded( $articleTitle->getPrefixedDBkey() );
+				ArticleVideoContext::isFeaturedVideoEmbedded( $title->getPrefixedDBkey() );
 		} else {
 			$isFeaturedVideoEmbedded = false;
 		}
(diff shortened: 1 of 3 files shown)


extensions/wikia/SiteWideMessages/SiteWideMessagesController.class.php
```diff
@@ -1,14 +1,30 @@
 <?php
 
-class SiteWideMessagesController extends WikiaController  {
+class SiteWideMessagesController extends WikiaController {
 
 	const CACHE_VALIDITY_VARNISH = 10800; // 3 hours
 	const CACHE_VALIDITY_BROWSER = 3600; // 1 hour
 
+	private function shouldShowSiteWideMessage() {
```
hakubo: shouldSkip

extensions/wikia/SiteWideMessages/SiteWideMessagesController.class.php
```diff
@@ -1,14 +1,30 @@
...
+			$isFeaturedVideoEmbedded = false;
+		}
+
+		return $this->wg->User->isLoggedIn() || $isFeaturedVideoEmbedded;
+	}
+
 	public function getAnonMessages() {
-		if ( $this->wg->User->isLoggedIn() ) {
+
+		if ( $this->shouldShowSiteWideMessage() ) {
```
rybmat: the method name says something opposite than it does ;) please change it to shouldNotShowSiteWideMessages()

extensions/wikia/SiteWideMessages/SiteWideMessagesController.class.php
```diff
@@ -1,14 +1,30 @@
...
 
 	const CACHE_VALIDITY_VARNISH = 10800; // 3 hours
 	const CACHE_VALIDITY_BROWSER = 3600; // 1 hour
 
+	private function shouldShowSiteWideMessage() {
+		global $wgEnableArticleFeaturedVideo;
+
+		$title = Title::newFromID( $this->getVal( 'articleId' ) );
+		if ( !is_null( $title ) ) {
+			$isFeaturedVideoEmbedded = !empty( $wgEnableArticleFeaturedVideo ) &&
```
hakubo, line 13: actually there is no need to call newFromId is that is false

hakubo (changes requested):

rybmat (changes requested):

extensions/wikia/SiteWideMessages/SiteWideMessagesController.class.php
JamesHawking: but we need title to check it if i understand correctly

hakubo (approved):

rybmat (approved):

PR #13275, opened 2017-07-07
IRIS-4645 - fix OG title on message walls

katanka:


extensions/OpenGraphMeta/OpenGraphMeta.php
```diff
@@ -78,8 +78,10 @@ function efOpenGraphMetaPageHook( OutputPage &$out, &$sk ) {
 		$meta["og:site_name"] = $wgSitename;
 		// Try to chose the most appropriate title for showing in news feeds.
 		if ( ( defined('NS_BLOG_ARTICLE') && $title->getNamespace() == NS_BLOG_ARTICLE ) ||
-			( defined('NS_BLOG_ARTICLE_TALK') && $title->getNamespace() == NS_BLOG_ARTICLE_TALK ) ){
+			( defined('NS_BLOG_ARTICLE_TALK') && $title->getNamespace() == NS_BLOG_ARTICLE_TALK ) ) {
 			$meta["og:title"] = $title->getSubpageText();
+		} elseif (defined('NS_USER_WALL_MESSAGE') && $title->getNamespace() == NS_USER_WALL_MESSAGE) {
+			$meta["og:title"] = $out->mHTMLtitle;
```
jsutterfield: So this doesn't do exactly what the ticket asks. It includes the title, but it also includes the wiki name, and the "Fandom powered by Wikia".

Eg, for this post: http://thelastofus.james.wikia-dev.us/wiki/Thread:31880 the og:title is "Saying hi | The Last of Us Wiki | Fandom powered by Wikia". Can we find a way to just have this include the title of the post? The quick way would be to split on "|" and only include the first part, but I think that's a bit fragile. Ideally we'd find some API to access the title directly.

PR #13294, opened 2017-07-11
Remove legacy script

mszabo-wikia:
These legacy svn utils are long unused and are missing from upstream repo too.


themech (approved):

bin/ulimit4.sh
```diff
@@ -1,4 +0,0 @@
-#!/bin/bash
-
-ulimit -t $1 -v $2 -f $3
-eval "$4"
```
macbre: Please keep it - it's [used by `wfShellExec`](https://github.com/Wikia/app/blob/100a2b7f92c7fd83f601b85897b8d319c8b9f1d8/includes/GlobalFunctions.php#L2946-L2951)

macbre (changes requested):

macbre (approved):

PR #13301, opened 2017-07-12
ADEN-5518 Fix localStorage on Safari private mode

fraszczakszymon:


resources/wikia/modules/localStorage.js
```diff
@@ -5,6 +5,9 @@ define('wikia.localStorage', ['wikia.window'], function(window) {
 	'use strict';
 
 	try {
+		window.localStorage.setItem('localStorageTestItem', 'testValue');
+		window.localStorage.getItem('localStorageTestItem');
```
themech: Just wondering: how does this module differ from https://github.com/Wikia/app/blob/b03df0a89ed672697e9c130d529bf1eb25f49cda/resources/wikia/modules/cache.js#L39, which makes basically the same check?

fraszczakszymon: Fair point. It's used by one module, so I'm going to refactor it a little bit and remove `wikia.localStorage`. Thanks!

themech: Thank you @fraszczakszymon

dianafa:
+4 −28 and a bug fix, love that! <3

PR #13304, opened 2017-07-12
XW-3734 | MediaWiki:Edit is not escaped, allowing JS injection

latata:
https://wikia-inc.atlassian.net/browse/XW-3734
@Wikia/x-wing @Grunny

### DIFF
--- a/extensions/wikia/PageHeader/templates/PageHeaderController_actionButton.php
+++ b/extensions/wikia/PageHeader/templates/PageHeaderController_actionButton.php
@@ -1,25 +1,29 @@
 <div class="wds-button-group">
 	<a href="<?= $buttonAction['href'] ?>"
 	   class="wds-is-squished wds-button"
-	   id="<?= $buttonAction['id'] ?>"
+	   id="<?= Sanitizer::encodeAttribute( $buttonAction['id'] ) ?>"
 	   data-tracking="<?= $buttonAction['data-tracking'] ?>"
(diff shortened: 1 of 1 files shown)


hakubo (approved):

extensions/wikia/PageHeader/templates/PageHeaderController_actionButton.php
```diff
@@ -20,7 +20,7 @@ class="wds-is-squished wds-button"
 						   href="<?= $action['href'] ?>"
 						   data-tracking="<?= $action['data-tracking'] ?>"
 						   <?= empty( $action['accesskey'] ) ? '' : "accesskey=\"{$action['accesskey']}\"" ?>>
```
Grunny: The `accesskey` comes from a message as plain text as well and should be escaped just in case of previous customisations.

Grunny (changes requested):

mszabo-wikia (changes requested):
What do you think about making `PageHeaderController` extends `WikiaService` instead of `WikiaController`? Are any of its methods meant to be called from JS via `wikia.php`?

rogatty:
@TK-999 that's something we wanted to do when creating PageHeader extension and then... forgot 😛 
We'll handle it in https://wikia-inc.atlassian.net/browse/XW-3737 to not make this vulnerability wait any longer.

mszabo-wikia (approved):

rogatty (approved):

Grunny (approved):

PR #13328, opened 2017-07-14
SUS-2431: Cleanup TOC extensions

themech:
Table of contents is full handled by TOC extension, so I'm disabling the TOC. More details in the ticket: https://wikia-inc.atlassian.net/browse/SUS-2431. Config change: https://github.com/Wikia/config/pull/2211/


mszabo-wikia (approved):

includes/DefaultSettings.php
```diff
@@ -5273,6 +5273,12 @@
 $wgAjaxLicensePreview = true;
 
 /**
+ * @name $wgEnableWikiaTOC
+ * enable Wikia JS driven TOC
+ */
+$wgEnableWikiaTOC = true;
+
+/**
```
macbre: Why not move the include of the WikiaTOC extension here, if it's enabled sitewide?

macbre (changes requested):

macbre (approved):

includes/DefaultSettings.php
themech: Done, thanks

PR #13357, opened 2017-07-18
IRIS-3678 - don't include deleted discussions posts in recirc footer

katanka:
https://wikia-inc.atlassian.net/browse/IRIS-3678


extensions/wikia/Recirculation/services/DiscussionsDataService.class.php
```diff
@@ -99,6 +99,7 @@ private function buildUrl( $endpoint, $options ) {
 			'limit' => self::DISCUSSIONS_API_LIMIT,
 			'sortKey' => self::DISCUSSIONS_API_SORT_KEY,
 			'sortDirection' => self::DISCUSSIONS_API_SORT_DIRECTION,
+			'viewableOnly' => true,
```
jsutterfield: Minor: looks like the pattern here is to define the value as a constant of this class. Could we do that here as well to keep things consistent?

jsutterfield (approved):

PR #13369, opened 2017-07-19
SUS-2445: Upgrade abandoned videlalvaro/php-amqplib dependency

mszabo-wikia:
Use `php-amqplib/php-amqplib` instead.

https://wikia-inc.atlassian.net/browse/SUS-2445


macbre:
Let's check if `IndexingPipeline` still works after these changes.

macbre (approved):

macbre:
Let's keep dependencies up to date :+1:

PR #13393, opened 2017-07-24
SUS-2120: Undeleting does not imply editing

mszabo-wikia:
These lines were added during last Core MW security backport. This assumption is not true with many of our features such as blogs, comments, Wall etc. I'm removing it so undeleting works properly.

https://wikia-inc.atlassian.net/browse/SUS-2120


Grunny (approved):
Sure, I don't think we have many valid cases where someone can undelete a page they can't edit, but shouldn't be able to undelete it. Technically in the case of blogs, comments, and Wall, nearly all users who can undelete can also edit the comment pages, they just rely on a different edit permission, right?

We can also probably remove the two i18n messages that were added with this.

macbre (approved):

macbre (approved):

ludwikkazmierczak:
Tested and Accpeted

PR #13403, opened 2017-07-25
add page number to url in MessageWall Pagination

shaedrich:


shaedrich:
Is fixed/improved by #13436

PR #13434, opened 2017-07-31
IRIS-4770 | Updated the staff welcome post for migrated Discussions

slayful:
https://wikia-inc.atlassian.net/browse/IRIS-4770


shaedrich:
The other language messages have to be updated as well, right?

slayful:
@shaedrich, @jsutterfield sure, I'll be pushing it to i18n shortly, which should create a ticket for LOC team. :)

PR #13436, opened 2017-07-31
Wall page num url param

shaedrich:
url param (instead of hash) for consistency and initial loading for permalink

__Fixes/improves__ #13403


shaedrich:
#13403 was buggy, but you approved that, @TK-999. This one should fix it, but again, you close it without any comment on what's wrong with it (why don't you tell me the reason when closing a pr?).

mszabo-wikia:
After internal discussion it was determined we cannot accept such feature requests for now to maintain compliance with the Pull Request Policy.

shaedrich:
Okay, then I have to report a bug I implemented by myself to get it solved?

PR #13443, opened 2017-07-31
ADEN-5624 Add fallbacks before cache invalidates

nandy-andy:
Follow-up after https://github.com/Wikia/app/pull/13440


dianafa:
LGTM

PR #13444, opened 2017-07-31
SUS-2178: &$this should not be passed by reference

mszabo-wikia:
https://wikia-inc.atlassian.net/browse/SUS-2178

Passing `&$this` as a reference will cause issues with PHP 7.1. It is also not needed as objects are passed by reference automatically since PHP 5:
* https://bugs.php.net/bug.php?id=73751
* https://3v4l.org/vMc1N
* https://wiki.php.net/rfc/this_var


extensions/wikia/Answers/Answers.php
```diff
@@ -638,16 +617,3 @@ function wfAnswersGetEditPointsAjax() {
...
 }
-
-$wgHooks["UserToggles"][] = "wfHideFromAttribution";
-function wfHideFromAttribution(&$toggles) {
-	$toggles[] = "hidefromattribution";
-	return true;
-}
-
-
-$wgHooks['CustomArticleFooter'][] = "wfAnswersHideFooter";
```
mszabo-wikia, line 649: I could not find any invocations of hooks `CustomArticleFooter` and `UserToggles` in our codebase.

extensions/wikia/Answers/EditSimilarAnswers.php
```diff
@@ -1,13 +0,0 @@
-<?php
-
-$wgHooks["EditSimilar::getSimilarArticles"][] = "wfGetSimilarArticlesOnAnswers";
-
-function wfGetSimilarArticlesOnAnswers(&$data) {
```
mszabo-wikia, line 5: Hook `EditSimilar::getSimilarArticles` is never invoked in our codebase.

extensions/wikia/WikiAnswers/WikiAnswers.php
```diff
@@ -11,10 +11,6 @@
 
 // add CSS
 $wgHooks['BeforePageDisplay'][] = 'wfWikiAnswersAddStyle';
-// remove Follow link on the main page
-$wgHooks['FooterMenuAfterExecute'][] = 'wfWikiAnswersFooterMenu';
-// replace create-a-wiki link with a create-answers-wiki link
-$wgHooks['GlobalHeaderIndexAfterExecute'][] = 'wfWikiAnswersGlobalHeaderIndex';
```
mszabo-wikia, line 17: These controllers have been replaced by Design System controllers, so their `AfterExecute` hooks are never called.

WikiaTech:
* ![MAJOR][MAJOR] 1 major

extensions/AbuseFilter/AbuseFilter.hooks.php
```diff
@@ -100,7 +100,9 @@ public static function onAbortMove( $oldTitle, $newTitle, $user, &$error, $reaso
 		return $filter_result == '' || $filter_result === true;
 	}
 
-	public static function onArticleDelete( &$article, &$user, &$reason, &$error ) {
+	public static function onArticleDelete(
+		WikiPage $article, User $user, $reason, &$error
```
slayful, line 104: I can see that $user is not used here, so it's possible that some people pass `null` here. If that's the case this method will crash if you make the param typed.

extensions/Cite/Cite.php
```diff
@@ -87,10 +87,10 @@ function wfCite( $parser ) {
 
 /**
  * @param $out OutputPage
- * @param $sk Skin
+ * @param Skin $skin
  * @return bool
  */
-function wfCiteBeforePageDisplay( $out, &$sk ) {
+function wfCiteBeforePageDisplay( OutputPage $out, Skin $skin ): bool {
```
slayful, line 93: Same deal, some code may pass in `null` instead of skin and this will crash.

extensions/AbuseFilter/AbuseFilter.hooks.php
mszabo-wikia: If `$user` is null `WikiPage::doDeleteArticleReal` will crash anyways when it reaches the call to `LogEntry::setPerformer`, so it does not make much of a difference 😉

extensions/Cite/Cite.php
mszabo-wikia: Hook `BeforePageDisplay` is only invoked from `OutputPage::output` where it receives `$this` as first parameter

extensions/OpenGraphMeta/OpenGraphMeta.php
```diff
@@ -1,132 +1,132 @@
...
-
-	return true;
-}
-
-$egFacebookAppId = null;
-$egFacebookAdmins = null;
-
+<?php
+/**
+ * OpenGraphMeta
```
slayful, line 3: It might be worth checking if we can kill this extension and use our Open Graph service.

extensions/VisualEditor/VisualEditor.hooks.php
```diff
@@ -44,121 +33,12 @@ public static function onSetup() {
 	 * @param Skin $skin
 	 * @return boolean
 	 */
-	public static function onBeforePageDisplay( OutputPage &$output, Skin &$skin ) {
+	public static function onBeforePageDisplay( OutputPage $output, Skin $skin ): bool {
 		// Wikia change
 		if ( self::isAvailable( $skin ) ) {
 			$output->addModules( array( 'ext.visualEditor.wikia.viewPageTarget.init' ) );
 		}
```
slayful, line 40: unused methods below?

mszabo-wikia: yeah, the hook registrations for them were commented out in setup file

extensions/wikia/AchievementsII/Ach_setup.php
```diff
@@ -118,36 +118,33 @@ function Ach_Setup() {
...
-					$summary, $minoredit, $watchthis,
-					$sectionanchor, &$flags, $revision,
-					&$status, $baseRevId) {
+function Ach_ArticleSaveComplete(
+	WikiPage $article, User $user, $text, $summary, $minoredit, $watchthis, $sectionanchor, $flags,
+	$revision, Status &$status, $baseRevId
+): bool {
 	wfProfileIn(__METHOD__);
 
-	if($status instanceof Status && $status->ok == true && count($status->errors) == 0) {
```
slayful, line 150: I believe instanceof may have been used to check if status is `null`.

extensions/wikia/AchievementsII/Ach_setup.php
```diff
@@ -158,25 +155,25 @@ function Ach_ArticleSaveComplete(	&$article, &$user, $text,
...
-	if($wgUser->isLoggedIn() && !($wgUser->getGlobalPreference( 'hidepersonalachievements' ))) {
+	if ($user->isLoggedIn() && !( $user->getGlobalPreference( 'hidepersonalachievements' ) ) ) {
 		if( $skin->getTitle()->isSpecial( 'MyHome' ) ) {
 			$awardingService = new AchAwardingService();
-			$awardingService->awardCustomNotInTrackBadge($wgUser, BADGE_WELCOME);
+			$awardingService->awardCustomNotInTrackBadge( $user, BADGE_WELCOME );
 		}
 
-		if((!empty($_SESSION['achievementsNewBadges']) || 5 == rand(1, 20)) && get_class(RequestContext::getMain()->getSkin()) != 'SkinMonobook') {
+		if ( ( !empty( $_SESSION['achievementsNewBadges'] ) || 5 == rand( 1, 20 ) ) && $skin->getSkinName() !== 'monobook' ) {
```
slayful, line 169: Perhaps there's a convenience method for that or we could extract one? For checking if skin is monobook.

extensions/wikia/AchievementsII/Ach_setup.php
```diff
@@ -253,79 +240,6 @@ function Ach_TakeRankingSnapshot($force = false) {
...
-			$title = Title::newFromText('CArticle_'.$i);
-			$article = new Article($title);
-			$article->doEdit($content, 'test summary');
-		}
-
-		$dbw = wfGetDB(DB_MASTER, array(), $wgExternalSharedDB);
-		$dbw->commit();
-
-		$dbw = wfGetDB(DB_MASTER);
-		$dbw->commit();
```
slayful, line 292: Why is this being removed?

extensions/wikia/AchievementsII/Ach_setup.php
mszabo-wikia: In `WikiPage::doEdit` `$status` is created via `Status::newGood` call which always returns `Status` instance - and a property is accessed before it is passed on:

```		// Return the new revision (or null) to the caller
		$status->value['revision'] = $revision;

		Hooks::run( 'ArticleSaveComplete', [
			$this,
			$user,
			$text,
			$summary,
			$flags & EDIT_MINOR,
			null,
			null,
			$flags,
			$revision,
			&$status,
			$baseRevId,
		] );
```

extensions/wikia/AchievementsII/Ach_setup.php
mszabo-wikia: Hmm, the closest I could think of would be to introduce some `Skin::is( string $skinName ): boolean;` method, which would essentially call this under the hood, but I'm not sure if it'd be worth it.

extensions/wikia/ArticleComments/classes/ArticleCommentInit.class.php
```diff
@@ -105,17 +105,16 @@ static public function ArticleCommentEnable( &$data ) {
...
 	 * @return bool
 	 */
-	static public function ArticleCommentAddJS( OutputPage &$out, Skin &$sk ) {
+	static public function ArticleCommentAddJS( OutputPage $out, Skin $skin ): bool {
 		global $wgExtensionsPath;
 
 		if ( self::ArticleCommentCheck() ) {
 			// FB#21244 this should run only for MonoBook, Oasis and WikiaMobile have their own SASS-based styling
-			if ( $sk instanceof SkinMonoBook ) {
+			if ( $skin->getSkinName() === 'monobook' ) {
```
slayful, line 117: Convenience method?

extensions/wikia/AchievementsII/Ach_setup.php
mszabo-wikia: It appears to have been only used for testing years ago and its `$wgAjaxExportList` entry is commented out: https://github.com/Wikia/app/pull/13444/files/3d502da2f23fceed6ea1eb05f5ed72a62edb0639#diff-f1fb8c62f74f81312278d6454394bcceL259

extensions/wikia/ArticleComments/classes/ArticleCommentList.class.php
```diff
@@ -760,36 +757,34 @@ static public function articleDeleteComplete( WikiPage &$wikiPage, &$user, $reas
 	 *
 	 * @return boolean -- because it's a hook
```
slayful, line 758: I've stopped CRing for today.

extensions/OpenGraphMeta/OpenGraphMeta.php
mszabo-wikia: AFAIK the service can only fetch OG metadata for a given URL, so it could not serve as a replacement for this functionality

extensions/wikia/MyHome/MyHome.class.php
```diff
@@ -111,9 +111,6 @@ public static function storeInRecentChanges(RecentChange $rc, $data = array()) {
 				break;
 		}
 
-		//allow to alter $data by other extensions (eg. Article Comments)
-		Hooks::run('MyHome:BeforeStoreInRC', array(&$rc, &$data));
```
mszabo-wikia, line 115: This hook was not used by any extensions.

extensions/wikia/Qualaroo/QualarooHooks.class.php
```diff
@@ -7,40 +7,40 @@ class QualarooHooks {
 	*
 	* @return bool
 	*/
-	static public function onOasisSkinAssetGroups( &$assetsArray ) {
-		global $wgNoExternals, $wgOut;
+	static public function onOasisSkinAssetGroups( array &$assetsArray ): bool {
+		global $wgNoExternals;
 
-		if ( empty( $wgNoExternals ) && $wgOut->getSkin()->getSkinName() == 'oasis' ) {
```
mszabo-wikia, line 13: This check is not needed because this hook is run by Oasis itself.

extensions/wikia/RTE/RTE.class.php
```diff
@@ -264,9 +266,6 @@ public static function makeGlobalVariablesScript( &$vars ) {
 		$vars['RTECookieDomain'] = $wgCookieDomain;
 		$vars['RTECookiePath'] = $wgCookiePath;
 
-		// allow other extensions to add extra global JS variables to edit form
-		Hooks::run('RTEAddGlobalVariablesScript', array(&$vars));
```
mszabo-wikia, line 268: These RTE hooks were not used by any other extensions.

includes/EditPage.php
```diff
@@ -2079,13 +2079,8 @@ protected function showHeader() {
 					}
 
 					if ( !$revision->isCurrent() ) {
-						/* Wikia change begin - @author: Marooned */
-						/* Add new hook and condition to allow alternate message to be displayed when editing old revision */
-						if ( Hooks::run( 'EditPage::showEditForm:oldRevisionNotice', array( &$this ) ) ) {
```
mszabo-wikia, line 2084: This custom hook was not used anywhere.

macbre:
Tests suite fails:

```
Warning: Parameter 1 to MapsHooks::onParserClearState() expected to be a reference, value given in /var/lib/jenkins/workspace/integrationTests_php_branches/app/includes/Hooks.php on line 216
```

Is this work in progress? :smile:

extensions/Cite/Cite.php
ghost: @slayful — a crippled program does more damage than a dead one. It is a good practice to crash early.  If we pay for `Skin` byt get `null`, we should crash and investigate rather than desperately try to handle the situation. :-)  A massive PHP fatal error will be automatically reported and JIRA case will be created. In my opinion this is a good change.

extensions/wikia/ArticleComments/classes/ArticleCommentList.class.php
ghost: Yeah, having reached here, I stopped CRing and started CRying.

ghost (changes requested):
There isn't much to review.  We need to do proper testing which in our case means "run all tests we have". So let's have all the tests passing for this branch.

extensions/ConfirmEdit/Captcha.php
```diff
@@ -604,11 +605,11 @@ public function APIGetAllowedParams( &$module, &$params ) {
 	}
 
 	/**
-	 * @param $module ApiBae
```
mszabo-wikia, line 607: https://www.theguardian.com/culture/shortcuts/2017/aug/02/from-bae-to-millennial-11-words-that-make-the-heart-sink
>*Bae* Definition of the sort of neologism you probably should have looked up by now, but somehow could never be arsed. You secretly suspect that even if you did, the meaning would simply glide past the eyes in a seizure of inanity.

ghost (approved):

extensions/AbuseFilter/AbuseFilter.hooks.php
```diff
@@ -100,7 +100,9 @@ public static function onAbortMove( $oldTitle, $newTitle, $user, &$error, $reaso
...
+	): bool {
```
Krinkle, line 105: I'm not sure of the wider context here, but note that mandatory boolean returns not been present in MediaWiki for some time. This requirement was lifted in 2013 (release in MW 1.22) due to the common mistake where a hook is forgetting to return true/false (resulting in the implicit null return being casted to return false, which means to abort the action, and not allow any other hooks to run, either).

Given many hooks are just to allow modification to an array of some sorts (which cannot be aborted and has its return value ignored), this was removed to make the usage of those hooks more idiomatic.

This return type is probably safe to add here, but might conflict in the future or produce false negatives with regards to core not strictly expecting it (and not supposed to).

Good day :)

mszabo-wikia: Thanks @Krinkle ! Yeah, my rationale was that adding the explicit return type hint can be useful if a code path in the handler can result in early abort, while it does not do any harm at all if we always allow processing to continue.

includes/wikia/nirvana/WikiaParserTagController.class.php
```diff
@@ -69,7 +69,7 @@ protected function registerResourceLoaderModules( Parser $parser ) {
 		return $markerId;
 	}
 
-	public final function onParserAfterTidy( Parser &$parser, &$text ) {
+	public final function onParserAfterTidy( Parser $parser, &$text ) {
```
macbre, line 72: Let's try to find such places automatically, `grep` to the rescue? Or maybe update the `stylize.php` from the upstream to clean-up such cases?

mszabo-wikia: Replaced the few final occurrences of those. I have also executed complete SUS Selenium suite + X-Wing Selenium test suite against the changeset - time to merge.

PR #13448, opened 2017-07-31
MAIN-10932 - LyricWiki: Support Cyrillic, Greek, and more Latin letters in fletter magic words

xofe:
Currently, the `artistfletter`/`albumfletter`/`songfletter` magic words support the English alphabet only. Anything else defaults to 'Symbol' and needs to be set manually in a template on the page. This change would reduce the number of cases where that'd be needed.

----
Note that I've kept the existing code style for consistency with the rest of the file and simplicity of the changes - if you'd like me to follow code conventions instead, just let me know!

Ticket: MAIN-10932

### DIFF
--- a/extensions/3rdparty/LyricWiki/Parser_LWMagicWords.php
+++ b/extensions/3rdparty/LyricWiki/Parser_LWMagicWords.php
@@ -25,14 +25,13 @@ function findFirstLetterOf($fLetterOf)
 	{
 		return "0-9";
 	}
-	else if(($fLetter < "A") || ($fLetter > "Z"))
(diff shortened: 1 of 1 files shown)


xofe:
Thanks for merging these, @TK-999! I've noticed now that the `findFirstLetterOf()` function is not using the multibyte versions of `strtoupper` and `substr`, meaning non-ASCII characters aren't handled correctly even after this change.

Should I request another ticket to be opened regarding this, or would it be okay to open a follow-up pull request to change these to `mb_strtoupper` and `mb_substr` without a dedicated ticket?

mszabo-wikia:
@xofe Sure, feel free to open a PR with fix 🙂

xofe:
Thanks! I've opened the PR at #13797

PR #13451, opened 2017-08-01
SUS-2521: Ensure we always use valid thread ID when rendering Wall

mszabo-wikia:
`WallHooksHelper::onArticleViewHeader` assumes that if `intval( $title->getText() ) > 0` then `$title->getText()` is a positive integer. This assumption is false:
```
echo intval('2123_karamba');
2123
```
The fix is to pass through the value through `intval` before using it, not just use it for checking.
https://wikia-inc.atlassian.net/browse/SUS-2521


extensions/wikia/Wall/WallHooksHelper.class.php
```diff
@@ -70,7 +72,15 @@ static public function onArticleViewHeader( &$article, &$outputDone, &$useParser
 				// try master
 				$mainTitle = Title::newFromId( $threadId, Title::GAID_FOR_UPDATE );
 				if ( !empty( $mainTitle ) ) {
+					WikiaLogger::instance()->info( 'Wall thread master fallback - found', [
+						'threadId' => $threadId
+					] );
```
macbre: A `warning` perhaps?

PR #13466, opened 2017-08-02
Show latest poll results

shaedrich:
AjaxPoll have to be purged so templates containing the poll have to show the latest results when transcluded on any page


mszabo-wikia:
Thanks for your contribution! Using polls via template transclusion is a very rare use case, so an doing an expensive purge operation on each vote is not worth it.

shaedrich:
So what do you think should be done instead to achieve this?

PR #13471, opened 2017-08-03
SUS-2528: Move legacy default styles from MediaWiki:Common.css to core

mszabo-wikia:
These styles are included in Common.css on newly created English wikis, which results in confusing `@import` behavior. In newer versions of MW, they have been moved to `mediawiki.special` module - let's do so.

https://wikia-inc.atlassian.net/browse/SUS-2528

### DIFF
--- a/extensions/GlobalMessages/GlobalMessagesC.i18n.php
+++ b/extensions/GlobalMessages/GlobalMessagesC.i18n.php
@@ -103,18 +103,7 @@ Since you do not have the hideuser right, you cannot see or edit the user's bloc
 	'cant-move-user-page' => 'You do not have permission to move user pages (apart from subpages).',
 	'cant-move-to-user-page' => 'You do not have permission to move a page to a user page (except to a user subpage).',
 	'cantmove-titleprotected' => 'You cannot move a page to this location, because the new title has been protected from creation',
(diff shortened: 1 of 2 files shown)


extensions/GlobalMessages/GlobalMessagesC.i18n.php
```diff
@@ -103,18 +103,7 @@
...
 	'cant-move-to-user-page' => 'You do not have permission to move a page to a user page (except to a user subpage).',
 	'cantmove-titleprotected' => 'You cannot move a page to this location, because the new title has been protected from creation',
-	'common.css' => '/***** CSS placed here will be applied to all skins on the entire site. *****/
-
-/* Mark redirects in Special:Allpages and Special:Watchlist */
-.allpagesredirect {
-   font-style: italic;
-}
-.allpagesredirect:after {
-   color: #808080; content: " (redirect)"
```
mszabo-wikia, line 113: Not localizable, not included in newer core, pretty much useless.

macbre (approved):

extensions/GlobalMessages/GlobalMessagesC.i18n.php
HM100: Hmmm. So you had to do this by yourself.

PR #13477, opened 2017-08-03
CAKE-2028 Add CORS support to the DesignSystemApiController

drsnyder:
https://wikia-inc.atlassian.net/browse/CAKE-2028
https://wikia-inc.atlassian.net/browse/CAKE-2033

Add `Access-Control-Allow-Origin:*` to all design system API requests. We are adding requests to these APIs in the header and footer web components in the [`fandom-global-elements`](https://github.com/Wikia/fandom-global-elements). Some of these requests will be coming from *.wikia.com and *.wikia-dev.us as we roll out support for the header and footer. Without the `Access-Control-Allow-Origin` header the browser won't propagate the response up and the request will error out.

Since this is public information and there are no state changes taking place I think this should be ok. If we are concerned about abuse we can follow this up with domain whitelisting.

@Wikia/cake @vforge

### DIFF
--- a/includes/wikia/api/DesignSystemApiController.class.php
+++ b/includes/wikia/api/DesignSystemApiController.class.php
@@ -11,7 +11,8 @@ class DesignSystemApiController extends WikiaApiController {
 	public function __construct() {
 		parent::__construct();
 		$this->cors = new CrossOriginResourceSharingHeaderHelper();
-		$this->cors->setAllowOrigin( [ '*' ] );
+		$this->cors->readConfig();
+		$this->cors->setAllowCredentials( true );
 	}
 
 	public function getFooter() {
--- a/includes/wikia/api/DesignSystemApiController.class.php
+++ b/includes/wikia/api/DesignSystemApiController.class.php
@@ -6,6 +6,14 @@ class DesignSystemApiController extends WikiaApiController {
 	const PARAM_LANG = 'lang';
 	const PRODUCT_WIKIS = 'wikis';
 
+	protected $cors;
+
+	public function __construct() {
+		parent::__construct();
+		$this->cors = new CrossOriginResourceSharingHeaderHelper();
+		$this->cors->setAllowOrigin( [ '*' ] );
+	}
+
 	public function getFooter() {
 		$params = $this->getRequestParameters();
 		$footerModel = new DesignSystemGlobalFooterModel(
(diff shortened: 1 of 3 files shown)


includes/wikia/api/DesignSystemApiController.class.php
```diff
@@ -6,6 +6,14 @@ class DesignSystemApiController extends WikiaApiController {
 	const PARAM_LANG = 'lang';
 	const PRODUCT_WIKIS = 'wikis';
 
+	protected $cors;
+
+	public function __construct() {
+		parent::__construct();
+		$this->cors = new CrossOriginResourceSharingHeaderHelper();
+		$this->cors->setAllowOrigin( [ '*' ] );
```
nmonterroso: I wonder if this should be more specific. 

There's probably no harm in having this be open to everyone, but maybe we should be a little paranoid with this sort of thing and restrict it to `*.wikia.com` and `*.wikia-dev.pl`, `*.wikia-dev.us` for now. What do you think?

nmonterroso (commented):
minor comment otw lgtm

includes/wikia/helpers/CrossOriginResourceSharingHeaderHelper.php
```diff
@@ -50,11 +61,17 @@ public function setHeaders( WikiaResponse $response, $mergeExisting = true ) {
 					$response->removeHeader( $headerName );
 
 					foreach ( $headers as $header ) {
-						$valuesToSet = array_merge( $valuesToSet, explode( self::HEADER_DELIMETER, $header['value'] ) );
+						if ( strpos( $header['value'], self::HEADER_DELIMETER ) !== false ) {
```
WikiaTech: MAJOR Refactor this code to not nest more than 4 "if", "for", "while", "switch" and "try" statements. [rule](https://sonarqube.wikia-inc.com/coding_rules#rule_key=php%3AS134)

WikiaTech:
* ![MAJOR][MAJOR] 1 major

drsnyder:
Update: I added support for `Access-Control-Allow-Credentials` so that cookies can be used. I also discovered the `wgCORSAllowOrigin` variable which allows you to configure the set of origins that are allowed. We can set that value on our dev boxes or anywhere else that needs to allow extra-domain requests.

Here's an example setting for that variable:
```
$wgCORSAllowOrigin = [ 'http://local-cms.wikia-dev.us:8083' ];
```

This allows the `fandom-global-elements` components to work properly with a dev box.

drsnyder:
As an aside to this pull request, the support for `Access-Control-Allow-Origin` is busted in that `Access-Control-Allow-Origin` is not a multi-value header. It only [accepts either a URL including the scheme or '*'](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin). `CrossOriginResourceSharingHeaderHelper` should be updated to have `setAllowOrigin` take a scalar as a parameter instead of an array.

nmonterroso:
LGTM

rogatty (approved):

PR #13488, opened 2017-08-07
SUS-1594: Use shared DB for initializing edit count and defer it to an offline task

mszabo-wikia:
https://wikia-inc.atlassian.net/browse/SUS-1594

### DIFF
--- a/includes/User.php
+++ b/includes/User.php
@@ -947,45 +947,35 @@ class User implements JsonSerializable {
 	 */
 	public static function edits( $uid ) {
 		wfProfileIn( __METHOD__ );
-		$dbr = wfGetDB( DB_SLAVE );
-		// check if the user_editcount field has been initialized
-		$field = $dbr->selectField(
-			'user', 'user_editcount',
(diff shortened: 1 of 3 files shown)


includes/User.php
```diff
@@ -947,45 +947,35 @@ public static function getCanonicalName( $name, $validate = 'valid' ) {
...
-		$field = $dbr->selectField(
-			'user', 'user_editcount',
-			array( 'user_id' => $uid ),
+		global $wgExternalSharedDB;
+
+		// check if the user_editcount field has been initialized on shared DB
+		$wikiCitiesDb = wfGetDB( DB_SLAVE, [], $wgExternalSharedDB );
+		$count = $wikiCitiesDb->selectField(
+			'`user`', 'user_editcount',
+			[ 'user_id' => $uid ],
```
macbre, line 956: Why do we check a shared edit count when handling a per-wiki count?

macbre (changes requested):

PR #13493, opened 2017-08-07
SUS-2566: Avoid falling back to master in case of no Community Messages

mszabo-wikia:
This fallback logic is not needed here. If there is no cached community messages timestamp, it simply means the message was not yet edited so there is no use querying - we just trigger useless fallback to master. If someone has edited the message, we already warm the cache on edit: https://github.com/Wikia/app/blob/312c0aaeb5a19a6707a8212f2f30dab47ef62f88/extensions/wikia/CommunityMessages/CommunityMessages.class.php#L96

https://wikia-inc.atlassian.net/browse/SUS-2566

### DIFF
--- a/extensions/wikia/CommunityMessages/CommunityMessages.class.php
+++ b/extensions/wikia/CommunityMessages/CommunityMessages.class.php
@@ -37,26 +37,12 @@ class CommunityMessages {
 		}
 
 		//get timestamp of message
-		$communityMessagesTimestamp = $wgMemc->get(wfMemcKey('CommunityMessagesTimestamp'));
+		$cacheKey = static::getCommunityMessagesCacheKey();
+		$communityMessagesTimestamp = $wgMemc->get( $cacheKey );
 
-		if (!$communityMessagesTimestamp) {
-			$msgTitle = Title::newFromText('community-corner', NS_MEDIAWIKI);
-			if ($msgTitle) {
-				$msgRev = Revision::newFromTitle($msgTitle);
-				if ($msgRev) {
-					$communityMessagesTimestamp = wfTimestamp(TS_UNIX, $msgRev->getTimestamp());
-					$wgMemc->set(wfMemcKey('CommunityMessagesTimestamp'), $communityMessagesTimestamp, 86400 /*24h*/);
-				}
-			}
-		}
-
-		if (!$communityMessagesTimestamp) {
-			//no message?
-			return true;
-		}
-
-		if ($communityMessagesTimestamp < (time() - 86400 /*24h*/)) {
-			//message older than 24h - do not inform user about it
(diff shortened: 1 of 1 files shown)


extensions/wikia/CommunityMessages/CommunityMessages.class.php
```diff
@@ -37,26 +37,12 @@ static function onSkinTemplatePageBeforeUserMsg(&$msgs) {
...
-			if ($msgTitle) {
-				$msgRev = Revision::newFromTitle($msgTitle);
-				if ($msgRev) {
-					$communityMessagesTimestamp = wfTimestamp(TS_UNIX, $msgRev->getTimestamp());
-					$wgMemc->set(wfMemcKey('CommunityMessagesTimestamp'), $communityMessagesTimestamp, 86400 /*24h*/);
-				}
-			}
-		}
+		$cacheKey = static::getCommunityMessagesCacheKey();
+		$communityMessagesTimestamp = $wgMemc->get( $cacheKey );
```
macbre: Please remember that cache eviction can take place. I do not have a full context here, maybe it's not a problem in this particular piece of code.

extensions/wikia/CommunityMessages/CommunityMessages.class.php
```diff
@@ -194,4 +180,8 @@ private static function getUserTimestamp(User $user) {
 		);
 		return $userTimestamp ? wfTimestamp(TS_UNIX, $userTimestamp) : false;
 	}
+
+	private static function getCommunityMessagesCacheKey(): string {
+		return wfMemcKey( 'CommunityMessagesTimestamp' );
```
macbre, line 183: Let's use this helper where the cache key is set.

macbre (changes requested):
Two minor things noted

extensions/wikia/CommunityMessages/CommunityMessages.class.php
mszabo-wikia: In below logic we already return if the message timestamp is older than 24 hours so I think this is safe this way

macbre: True, but eviction can take place before TTL expiration would happen.

WikiaTech:

1. ![MAJOR][MAJOR] [CommunityMessages.class.php#L89](https://github.com/Wikia/app/blob/6d65981cd5933fcb0a9ee55d8ca0218d0a37bc87/extensions/wikia/CommunityMessages/CommunityMessages.class.php#L89): Remove the unused function parameter "$baseRevId". [rule](https://sonarqube.wikia-inc.com/coding_rules#rule_key=php%3AS1172)

macbre (approved):

PR #13502, opened 2017-08-08
Identify new message hook to use elsewhere

shaedrich:


KockaAdmiralac:
I think it would be helpful to note in which cases would that hook be useful. `wikipage.content` is already firing, why not bind to it instead? Is [this thread](http://dev.wikia.com/wiki/Thread:13503) the context?

shaedrich:
Let's say, it gave me an idea. My problem in general ist, that `wikipage.content` hook is way too generic to identifiy what happened so some scripts could handle several cases (of course cases like that one you linked as well but it's just an example).

PR #13503, opened 2017-08-08
SUS-2564 | cache local user groups in PermissionsServiceImpl

macbre:
https://wikia-inc.atlassian.net/browse/SUS-2564

## Query

```sql
SELECT /* Wikia\\Service\\User\\Permissions\\PermissionsServiceImpl::loadLocalGroups Tzaflidisp - 97ca8f71-d483-4d98-bb95-16883634b3fc */  ug_group  FROM `user_groups`  WHERE ug_user = '123'
```

## The numbers

This query was explicitly made on master database node. Let's keep it that way, but add a cache around it (as we do for shared user groups).

This method is responsible for 21% of master node queries when GET requests is handled / over **780 k** in the last 24h.

### DIFF
--- a/lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
+++ b/lib/Wikia/src/Service/User/Permissions/PermissionsServiceImpl.php
@@ -51,13 +51,23 @@ class PermissionsServiceImpl implements PermissionsService {
 	}
 
 	/**
-	 * Get memcached key used for storing groups for a given user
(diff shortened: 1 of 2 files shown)


macbre:
Unit tests:

```
There was 1 failure:

1) RobotsIntegrationTest::testRobotsDeniedOnShowcasePage
```

Not related

mszabo-wikia (approved):

PR #13519, opened 2017-08-09
SUS-2359 | Guard against unknown remote database error

rybmat:
https://wikia-inc.atlassian.net/browse/SUS-2359

Guard against unknown remote database error


includes/filerepo/LocalRepo.php
```diff
@@ -194,16 +194,25 @@ protected function getArticleID( $title ) {
...
+			$id = $dbr->selectField(
+				'page', // Table
+				'page_id',  //Field
+				[  //Conditions
+					'page_namespace' => $title->getNamespace(),
+					'page_title' => $title->getDBkey(),
+				],
+				__METHOD__  //Function name
+			);
+		} catch ( DBError $e ) {
```
mszabo-wikia: Let's check for only `DBConnectionError` here. We probably want other errors to be passed on so that they show up in logs, which means Error Reporter is engaged and JIRA ticket is created.

We should also log this.

mszabo-wikia (changes requested):

includes/filerepo/LocalRepo.php
rybmat: @TK-999, DBError has logging implemented in constructor, do you think additional logging is needed here?

mszabo-wikia (approved):

PR #13538, opened 2017-08-10
SUS-2580: Do not require hook handlers to always return boolean

mszabo-wikia:
See [feedback](https://github.com/Wikia/app/pull/13444#discussion_r131309376) by @Krinkle in #13444 
>This requirement was lifted in 2013 (release in MW 1.22) due to the common mistake where a hook is forgetting to return true/false (resulting in the implicit null return being casted to return false, which means to abort the action, and not allow any other hooks to run, either).
Given many hooks are just to allow modification to an array of some sorts (which cannot be aborted and has its return value ignored), this was removed to make the usage of those hooks more idiomatic.

This lessens the chance of some rogue hook handler causing whitepages.

https://wikia-inc.atlassian.net/browse/SUS-2580

### DIFF
--- a/includes/Hooks.php
+++ b/includes/Hooks.php
@@ -87,19 +87,17 @@ class Hooks {
 	 * in here than would normally be necessary.
 	 *
 	 * @param $event String: event name
-	 * @param $args Array: parameters passed to hook functions
-	 * @return Boolean True if no handler aborted the hook
+	 * @param $args array: parameters passed to hook functions
+	 * @return bool True if no handler aborted the hook
+	 * @throws FatalError if a hook handler returns string
+	 * @throws MWException if the hook handler setup for this hook is invalid
 	 */
 	public static function run( $event, $args = array() ) {
-		wfProfileIn(__METHOD__.'-hook-'.$event);
-		wfProfileIn(__METHOD__);
 		// Wikia change - begin - @author: wladek
 		// optimized hooks execution
 
 		// Return quickly in the most common case
 		if ( !isset( self::$handlers[$event] ) ) {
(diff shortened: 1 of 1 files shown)


WikiaTech:

1. ![MAJOR][MAJOR] [Hooks.php#L120](https://github.com/Wikia/app/blob/3cd3bd77f6253e45adee23ec3e5e33bdddf16c89/includes/Hooks.php#L120): Remove this unused "$badhookmsg" local variable. [rule](https://sonarqube.wikia-inc.com/coding_rules#rule_key=php%3AS1481)

includes/Hooks.php
```diff
@@ -212,46 +209,19 @@ public static function run( $event, $args = array() ) {
...
-						"Hook $prettyFunc failed to return a value; " .
-						'should return true to continue hook processing or false to abort.'
-					);
-				}
-			} elseif ( !$retval ) {
-				wfProfileOut(__METHOD__);
-				wfProfileOut(__METHOD__.'-hook-'.$event);
+			} elseif ( $retval === false ) {
+				// False was returned. Stop processing, but no error.
 				return false;
```
Krinkle, line 214: Upstream MediaWiki no longer does manual function-level profiling (we encourage using XHProf instead), but given it still exists here, you'll want to make sure the relevant `wfProfileOut()` are called before the early return here. Otherwise there's a profile in, without the matching profile out.

mszabo-wikia: True :) This is something we are trying to get rid of anyways

Krinkle (approved):

macbre (approved):

PR #13544, opened 2017-08-10
SUS-2471: Fix Special:WithoutImages

rybmat:
https://wikia-inc.atlassian.net/browse/SUS-2471
https://wikia-inc.atlassian.net/browse/PLATFORM-2133

Currently Special:WithoutImages shows only pages that do not have image included (except images that are included on more than 20 pages - to avoid images like template icons) and additionally are linked on other page(s). While filtering the 'template icons' makes sense, the second condition does not, so in this PR it is fixed. Additionally general performance of db query is improved.


macbre (approved):

mszabo-wikia (approved):

drozdo:
LGTM

mszabo-wikia (approved):

PR #13550, opened 2017-08-11
SUS-2584: Use slave DB to fetch block info in WikiAnswers AnswerBox

mszabo-wikia:
As this is merely an UI item, we can afford to use slave instead of master here and save 3,000+ SELECT queries per day that are directly targeted at master nodes.

https://wikia-inc.atlassian.net/browse/SUS-2584

### DIFF
--- a/extensions/wikia/WikiAnswers/templates/WikiAnswers_AnswerBox.php
+++ b/extensions/wikia/WikiAnswers/templates/WikiAnswers_AnswerBox.php
@@ -7,12 +7,7 @@ global $wgBlankImgUrl, $wgTitle, $wgStylePath, $wgUser, $wgScriptPath;
 			<?php
 			$title = $wgTitle;
 		// check whether current user is blocked (RT #48058)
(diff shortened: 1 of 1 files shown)


extensions/wikia/WikiAnswers/templates/WikiAnswers_AnswerBox.php
```diff
@@ -3,37 +3,32 @@
...
 			<?php
 			$title = $wgTitle;
 		// check whether current user is blocked (RT #48058)
-		$isUserBlocked = $wgUser->isBlockedFrom($title, false);
-
-		if ($isUserBlocked) {
-			//FIXME port the code from Answer::getBlockedInfo()
-			//echo $this->getBlockedInfo();
-		} else {
+		if ( !$wgUser->isBlocked( true ) ) {
```
mszabo-wikia, line 10: No need to use `User::isBlockedFrom` here as that one is only really relevant for user talk pages, so this way we skip hook execution.

macbre (approved):

rybmat (approved):

PR #13553, opened 2017-08-11
SUS-2585: Use background task to dismiss Community Messages when Special:WikiActivity is viewed

mszabo-wikia:
Instead of performing an SQL REPLACE on page view, use a background task instead and save 4,000+ queries to master DB per day.

https://wikia-inc.atlassian.net/browse/SUS-2585

### DIFF
--- a/extensions/wikia/CommunityMessages/CommunityMessages.class.php
+++ b/extensions/wikia/CommunityMessages/CommunityMessages.class.php
@@ -114,9 +114,32 @@ class CommunityMessages {
 	static function onBeforePageDisplay( OutputPage $out, Skin $skin ): bool {
 		$title = $out->getTitle();
 
-		if ( $title->isSpecial('ActivityFeed') || $title->isSpecial('MyHome') || $title->isSpecial('WikiActivity') ) {
-			self::dismissMessage();
+		if ( !$title->isSpecial( 'ActivityFeed' ) && !$title->isSpecial( 'MyHome' ) &&
+			 !$title->isSpecial( 'WikiActivity' ) ) {
+			return true;
 		}
+
+		global $wgMemc;
+		$communityMessagesTimestamp = $wgMemc->get( static::getCommunityMessagesCacheKey() ) ?? time();
+
+		// SUS-2585: Update table for logged in users
+		$user = $out->getUser();
+		if ( $user->isLoggedIn() && $communityMessagesTimestamp > (int) static::getUserTimestamp( $user ) ) {
+			global $wgCityId;
+
+			$task = ( new \Wikia\Tasks\Tasks\DismissCommunityMessageTask() )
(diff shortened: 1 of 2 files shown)


extensions/wikia/CommunityMessages/CommunityMessages.class.php
```diff
@@ -114,10 +114,33 @@ static function onArticleSaveComplete( WikiPage $article, User $user, $text, $su
...
+
+			$task->call( 'dismissCommunityMessage', $communityMessagesTimestamp );
+			$task->queue();
+
+			return true;
 		}
 
+		// set cookie for anons
+		$out->getRequest()->response()
+			->setcookie('CommunityMessages', $communityMessagesTimestamp, time() + 86400 /*24h*/);
```
macbre, line 142: Shouldn't it be wrapped in an `if` statement that checks `$user->isAnon()`

includes/wikia/tasks/Tasks/DismissCommunityMessageTask.class.php
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * @see https://wikia-inc.atlassian.net/browse/SUS-2585
+ * @package Wikia\Tasks\Tasks
+ */
+class DismissCommunityMessageTask extends BaseTask {
+	public function dismissCommunityMessage( $communityMessagesTimestamp ) {
+		global $wgExternalDatawareDB;
+
+		$wikiId = $this->getWikiId();
+		$userId = $this->createdBy;
```
macbre, line 16: Let's use `createdBy()` accessor here.

macbre: And do we actually need these variables? We can assign `$row` fields directly.

mszabo-wikia: `createdBy` will return `BaseTask` instead of `User` :(

extensions/wikia/CommunityMessages/CommunityMessages.class.php
mszabo-wikia: We do early return before

includes/wikia/tasks/Tasks/DismissCommunityMessageTask.class.php
macbre: Ah, true. It's used to chain methods. Scratch that :)

macbre (approved):

extensions/wikia/CommunityMessages/CommunityMessages.class.php
macbre: But logged-in user can still reach this line (when the timestamps do not match in the previous if statement).

PR #13570, opened 2017-08-15
Removed duplicated user name in blog subtitle

shaedrich:
Is already displayed in the first item quite after the avatar as user page link text.


shaedrich:
Do you just need the ticket id [#332034](http://support.wikia-inc.com/hc/requests/332034)?

mszabo-wikia:
ZenDesk support ticket IDs do not map to an actual bug ticket. I'm not aware of one being opened for this case.

shaedrich:
Me neither.

PR #13572, opened 2017-08-16
ADEN-5694 add scrollY to kikimora

dianafa:
https://wikia-inc.atlassian.net/browse/ADEN-5694

### DIFF
--- a/extensions/wikia/AdEngine/js/lookup/services.js
+++ b/extensions/wikia/AdEngine/js/lookup/services.js
@@ -11,10 +11,11 @@
 /*global define, require*/
 define('ext.wikia.adEngine.lookup.services', [
 	'wikia.log',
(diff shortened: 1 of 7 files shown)


extensions/wikia/AdEngine/js/lookup/services.js
```diff
@@ -37,7 +38,8 @@ define('ext.wikia.adEngine.lookup.services', [
 			}
 		},
 		bidMarker = ['x', 'x', 'x', 'x', 'x'],
-		realSlotPrices = {};
+		realSlotPrices = {},
+		scrollY = {};
```
fraszczakszymon, line 42: Can we find better module for this information? Like `slotRegistry` which is already holding the refresh count (`rv` parameter)

PR #13580, opened 2017-08-17
SUS-1393 | Increase wgMaxImageArea and display error message when uploading bigger image

rybmat:
https://wikia-inc.atlassian.net/browse/SUS-1393

* increase wgMaxImageArea from 12.5MP to 60MP
* display error when trying to upload image with bigger resolution than wgMaxImageArea

DO NOT MERGE - waiting for translations


includes/upload/UploadBase.php
```diff
@@ -249,6 +251,25 @@ public function verifyUpload() {
 			);
 		}
 
+		// Wikia change - check if image resolution does not exceed wgMaxImageArea - @author: ryba
+		$imgInfo = getimagesize($this->mUpload->getTempName());
+		$imgWidth = $imgInfo[0];
+		$imgHeight = $imgInfo[1];
+		$mime = $imgInfo['mime'];
+		$imageResolution = $imgHeight * $imgWidth;
+		if ( $imageResolution > $wgMaxImageArea && $mime !== 'image/jpeg' ) {
```
macbre: I'm wondering if we should be checking JPEG files as well

macbre (approved):

includes/upload/UploadBase.php
rybmat: It would be cool to have consistent rules across all file types, however if we introduce it now, thumbnails of jpeg images larger than wgMaxImageArea (that are already uploaded) will stop rendering on a page

macbre (approved):

macbre (approved):

extensions/GlobalMessages/GlobalMessagesF.i18n.php
```diff
@@ -215,6 +216,7 @@
 	'feed-unavailable' => 'This message is displayed when a user tries to use an RSS or Atom feed on a wiki where such feeds have been disabled.',
 	'feed-atom' => '{{optional}}',
 	'feed-rss' => '{{optional}}',
+	'file-resolution-exceeded' => 'Error dispayed when user tries to upload image with too high resolution',
```
KockaAdmiralac: dispayed -> disp**l**ayed? (I don't know if it's engineers or the localization team that are supposed to correct this but leaving a comment just in case)

rybmat: @KockaAdmiralac thanks for noticing this, I'll fix it in a moment

ludwikkazmierczak:
Tested and Accepted

PR #13597, opened 2017-08-21
(PLATFORM-3106) Remove domain sharding from Vignette URLs

Grunny:
Remove domain sharding from Vignette URL generation.

/cc @Wikia/core-platform-team


harnash (approved):
I would recommend one more person to check this PR though.

PR #13606, opened 2017-08-21
ADEN-5663 Added missing variable

nandy-andy:


nandy-andy:
Nope. False alarm. We don't need it, we've got: [wgAdDriverPlayAdsOnNextFVCountries](https://github.com/Wikia/app/blob/519231f65f6cb8ba89d74f4f9c8d9102310e3f41/extensions/wikia/AdEngine/AdEngine2Hooks.class.php#L67).

PR #13608, opened 2017-08-21
ADEN-5632 Requests ABCD with MEGA adUnit builder

leshniak:


extensions/wikia/AdEngine/js/video/uapVideo.js
```diff
@@ -35,6 +35,14 @@ define('ext.wikia.adEngine.video.uapVideo', [
 		};
 	}
 
+	function shouldUseMegaAdUnitBuilder(type) {
+		var uapTypesMap = {
```
kvas-damian: You can define this map outside of function body. This way you avoid creating it on every function call.

extensions/wikia/AdEngine/js/video/vastUrlBuilder.js
```diff
@@ -56,21 +56,30 @@ define('ext.wikia.adEngine.video.vastUrlBuilder', [
...
-				'description_url=' + encodeURIComponent(loc.href),
-				'correlator=' + correlator,
-				'cust_params=' + getCustomParameters(slotParams)
-			],
+			useMegaAdUnitBuilder = options.useMegaAdUnitBuilder,
+			params,
 			url;
 
+		// fallback to adContext flag
+		if (useMegaAdUnitBuilder === undefined) {
```
kvas-damian: please use `typeof` instead of comparing values.

extensions/wikia/AdEngine/js/video/vastUrlBuilder.js
```diff
@@ -56,21 +56,30 @@ define('ext.wikia.adEngine.video.vastUrlBuilder', [
...
+			useMegaAdUnitBuilder = adContext.getContext().opts.megaAdUnitBuilderEnabled;
```
kvas-damian: Can we add a little jsdoc telling that by default we use this global flag?

extensions/wikia/AdEngine/js/video/vastUrlBuilder.js
```diff
@@ -87,5 +96,6 @@ define('ext.wikia.adEngine.video.vastUrlBuilder', [
 
 	return {
 		build: build
+
```
kvas-damian: Not required ;)

extensions/wikia/AdEngine/js/spec/video/player/porvata/googleImaSetup.spec.js
```diff
@@ -65,12 +65,13 @@ describe('ext.wikia.adEngine.video.player.porvata.googleImaSetup', function () {
 			request = imaSetup.createRequest({
 				width: slotWidth,
 				height: slotHeight,
+				useMegaAdUnitBuilder: true,
```
fraszczakszymon: We shouldn't change behaviour of all IMA instances. It should affect only ABCD ad product

extensions/wikia/AdEngine/js/video/uapVideo.js
leshniak: Just wanted to keep things more readable instead of taking care about of performance in this case ;)

extensions/wikia/AdEngine/js/video/vastUrlBuilder.js
leshniak: What's the benefit in this case?

kvas-damian: It doesn't change anything in this case but it's a good practice to not forget about it when using on variables that can be not defined.

extensions/wikia/AdEngine/js/video/uapVideo.js
leshniak: Done.

kvas-damian (approved):

extensions/wikia/AdEngine/js/video/vastUrlBuilder.js
```diff
@@ -46,32 +46,48 @@ define('ext.wikia.adEngine.video.vastUrlBuilder', [
 		return aspectRatio >= 1 || !isNumeric(aspectRatio) ? adSizes.horizontal : adSizes.vertical;
 	}
 
-	function buildAdUnit(slotParams) {
-		var adUnitBuilder = adContext.getContext().opts.megaAdUnitBuilderEnabled ? megaAdUnitBuilder : regularAdUnitBuilder;
```
v3rth, line 50: Right now it will be totally outside Disaster recovery, are we ok with that? Please note that in ticket

v3rth (changes requested):

extensions/wikia/AdEngine/js/video/vastUrlBuilder.js
v3rth: There is no way to turn it off by InstantGlobal

leshniak: As I understand, MEGA should be used for ABCD every time.

v3rth (approved):

PR #13609, opened 2017-08-22
cake-2132/2133 - add api call to receive email definition and wrapper task to send email

nmonterroso:
@Wikia/cake 
https://wikia-inc.atlassian.net/browse/CAKE-2132
https://wikia-inc.atlassian.net/browse/CAKE-2133

This changeset covers adding an internal-only API that will be called from the CGS with inputs required to send the content updated email as well as a wrapper task that MW will use to send the emails through the async queue.

Some notes:
- I _could_ have made `SendEmailController` receive the controller class as an input rather than having a specific `contentUpdated` method - think something like `SendEmailController.sendEmail` where `emailController` is another input param. This would have made this API more general, but I didn't want to have the CGS know the internal class names for the email controller. The CGS already has enough of the internals with having to know the controller name for the api including the namespace. I could be easily swayed on this if someone else thinks it should be made more generic so we hopefully don't have to touch this code if we want to add more emails.
- `SendEmailTask` _is_ generic, so we shouldn't have to add any more functionality in the future if we want to send other emails from the FC/CGS.
- Batching is explicitly not included in the api/task. I could see the argument being made here so that the CGS doesn't manage the size of an email (it should maybe be handled by the "email service"?) but from a practical perspective it's easier for us to change/deploy code from the CGS side, so I think it's better to have the batching logic there.

### DIFF
--- a/extensions/wikia/FandomCreatorEmail/Api/SendEmailController.php
+++ b/extensions/wikia/FandomCreatorEmail/Api/SendEmailController.php
@@ -1,0 +1,20 @@
+<?php
+
+namespace FandomCreatorEmail\Api;
+
+use FandomCreatorEmail\Controller\ContentUpdatedController;
+use FandomCreatorEmail\SendEmailTask;
+use WikiaApiController;
+
+class SendEmailController extends WikiaApiController {
+
+	public function allowsExternalRequests() {
+		return false;
+	}
+
+	public function contentUpdated() {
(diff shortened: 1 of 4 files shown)


drsnyder:
I think it's ok as-is for now. If we need to add a different kind of email we can figure out how we want the CGS to toggle that when that time comes.

LGTM

PR #13618, opened 2017-08-24
Ajax HTTP Method not needed

shaedrich:
Results in wrong url


KockaAdmiralac:
You could make it look cleaner with
```javascript
$.get(wgScriptPath + '/index.php', {
    action: 'ajax',
    rs: 'WMU',
    method: 'clean',
    mwname: $('#ImageUploadMWname').val(),
    tempid: $( '#ImageUploadTempid' ).val()
});
```

shaedrich:
Oh, that I should have seen :D Thank's for that hint :+1:

PR #13624, opened 2017-08-25
Implement upvote in embeddable discussions

shaedrich:
It wasn't able to upvote by clicking on the link with wrong href


shaedrich:
It's still buggy.

PR #13660, opened 2017-09-04
XW-3875 | support non-main namespaces in mainpage in GlobalTitle

latata:
`newMainPage` method in `GlobalTitle` class doesn't support namespaces other than main. I've added support for them.

@Wikia/x-wing @macbre


shaedrich:
Couldn't you do something like

```PHP
$mainPage = self::newFromText( 'Mainpage', NS_MEDIAWIKI, $city_id );
$mainPageName = $mainPage->getContent();
$mainPageNamespaceId = $mainPage->getNamespace();
$mainPageNamespace = MWNamespace::getCanonicalName($mainPageNamespaceId);
$title = self::newFromText( $mainPageName, $mainPageNamespace, $city_id );
```

includes/wikia/GlobalTitle.php
```diff
@@ -100,8 +100,21 @@ public static function newMainPage( $city_id = null ) {
...
 		}
 
-		$mainPageName = self::newFromText( 'Mainpage', NS_MEDIAWIKI, $city_id )->getContent();
-		$title = self::newFromText( $mainPageName, NS_MAIN, $city_id );
+		$mainPageTitle = self::newFromText( 'Mainpage', NS_MEDIAWIKI, $city_id );
+		$mainPageName = str_replace( ' ', '_', $mainPageTitle->getContent() );
+		$namespace = NS_MAIN;
+		// support for non-MAIN namespace pages, based on Title::secureAndSplit method
+		$prefixRegexp = "/^(.+?)_*:_*(.*)$/S";
+		if ( preg_match( $prefixRegexp, $mainPageName, $m ) ) {
```
hakubo: $m?

hakubo: $m?

hakubo (changes requested):

includes/wikia/GlobalTitle.php
```diff
@@ -100,8 +100,21 @@ public static function newMainPage( $city_id = null ) {
...
+			$mainPageTitle->loadContLang();
+			$namespaceIndex = $mainPageTitle->mContLang->getNsIndex( $m[1] );
+			if ( $namespaceIndex !== false ) {
+				$namespace = $namespaceIndex;
+				$mainPageName = $m[2];
+			}
+		}
+
+		$title = self::newFromText( $mainPageName, $namespace, $city_id );
```
macbre: Can we, instead of all this magic above, make an internal HTTP call to MediaWiki API to get the main page title?

Example: http://poznan.wikia.com/api.php?action=query&meta=siteinfo

```xml
<general mainpage="Wszystko o Poznaniu" base="http://poznan.wikia.com/wiki/Wszystko_o_Poznaniu" sitename="Poznańska Wiki" generator="MediaWiki 1.19.24" phpversion="7.0.7-4+deb.sury.org~trusty+1" phpsapi="apache2handler" dbtype="mysql" dbversion="5.6.24-72.2-log" case="first-letter" rights="CC-BY-SA" lang="pl" fallback8bitEncoding="iso-8859-2" writeapi="" timezone="UTC" timeoffset="0" articlepath="/wiki/$1" scriptpath="" script="/index.php" variantarticlepath="" server="http://poznan.wikia.com" wikiid="plpoznan" time="2017-09-04T11:34:35Z" misermode="">
```

macbre (changes requested):

includes/wikia/GlobalTitle.php
macbre: Actually, why do we use `GlobalTitle` when all we need is a local wiki mainpage that can be taken from `Title::newMainPage` ?

latata: Cause we need to run it out of wiki context (API request from discussions)

shaedrich (commented):
Couldn't you do something like

```PHP
$mainPage = self::newFromText( 'Mainpage', NS_MEDIAWIKI, $city_id );
$mainPageName = $mainPage->getContent();
$mainPageNamespaceId = $mainPage->getNamespace();
$mainPageNamespace = MWNamespace::getCanonicalName($mainPageNamespaceId);
$title = self::newFromText( $mainPageName, $mainPageNamespace, $city_id );
```

MediaWiki supports mainspaces on its own. So it seems the wrong way to me to try it by parsing a string using regular expressions by yourself.

mszabo-wikia:
@shaedrich no, because `MWNamespaces::getCanonicalNamespaces()` would be in wrong wiki context, as mentioned above

includes/wikia/GlobalTitle.php
latata: I'd rather fix GlobalTitle::newMainPage method cause it's used in a few more places.

shaedrich:
Okay. But the regex to get interwiki should be moved to a separate function.

includes/wikia/GlobalTitle.php
macbre: That's fine.

But please use `Title::newMainPage` in this API, because `GlobalTitle` is not really needed (as we discussed it yesterday).

latata (commented):
I reopened this PR cause we believe that it's the best solution for this issue for now. We can't use Title instead of GlobalTitle because DesignSystem API is prepared for requests without context of wiki (request to DS API are sent via www.wikia.com domain) and we can't change that easily.

shaedrich:
Does that mean that DesignSystem API has a design fault?

includes/wikia/tests/GlobalTitleTest.php
```diff
@@ -125,6 +125,31 @@ function testUrlsWithUtf8() {
 	}
 
 	/**
+	 * @dataProvider mainPageDataProvider
+	 */
+	function testMainPage( $mediaWikiMainpageContent, $exists, $expectedUrl ) {
```
xkxd: could you rename func name or comment about it to explain what is the subject of these tests? (i assume it's about title or URL)

includes/wikia/tests/GlobalTitleTest.php
```diff
@@ -165,4 +190,14 @@ public function stripArticlePathDataProvider() {
...
 		];
 	}
+
+	public function mainPageDataProvider() {
+		return [
+			['Whatever', true, 'http://community.wikia.com/wiki/Whatever'],
+			['Namespace:Test', true, 'http://community.wikia.com/wiki/Namespace:Test'],
+			['Namespace with spaces:Test', true, 'http://community.wikia.com/wiki/Namespace_with_spaces:Test'],
+			['Namespaces_with_underlines:Test', true, 'http://community.wikia.com/wiki/Namespaces_with_underlines:Test'],
+			['Notexists', false, 'http://community.wikia.com/wiki/Main_Page']
```
xkxd, line 200: why there is a "notexists" here? how it differs from the previous cases?

latata: it has `false` as a second attribute which mocks return value for `exists` method.

xkxd (approved):

hakubo (approved):

PR #13666, opened 2017-09-04
SUS-2628: Allow Phalanx to be synchronised using RabbitMQ queue

harnash:
This will send notification to both clusters - Mesos (old) via plain HTTP request and Kubernetes (new) via RabbitMQ queue.

There is companion PR in config repo: https://github.com/Wikia/config/pull/2260

ping: @Wikia/sus

### DIFF
--- a/extensions/wikia/PhalanxII/services/PhalanxService.class.php
+++ b/extensions/wikia/PhalanxII/services/PhalanxService.class.php
@@ -131,9 +131,9 @@ class PhalanxService {
 			: [];
 
 		$result = $this->sendToPhalanxDaemon( "reload", $params );
-		$queue_result = $this->sendToPhalanxQueue( $changed );
+		$this->sendToPhalanxQueue( $changed );
 		wfProfileOut( __METHOD__ );
-		return $result && $queue_result;
+		return $result;
 	}
 
 	private function sendToPhalanxQueue( $changed ) {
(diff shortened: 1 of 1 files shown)


extensions/wikia/PhalanxII/services/PhalanxService.class.php
```diff
@@ -129,8 +131,19 @@ public function reload( $changed = [] ) {
...
 
 		$result = $this->sendToPhalanxDaemon( "reload", $params );
+		$queue_result = $this->sendToPhalanxQueue( $changed );
 		wfProfileOut( __METHOD__ );
-		return $result;
+		return $result && $queue_result;
+	}
+
+	private function sendToPhalanxQueue( $changed ) {
+		wfProfileIn( __METHOD__ );
```
mszabo-wikia: Explicit profiler calls should no longer be needed as we now use Tideways profiler.

mszabo-wikia (changes requested):

macbre (approved):

mszabo-wikia (approved):

extensions/wikia/PhalanxII/services/PhalanxService.class.php
harnash: Fixed in e9262c9aeb9879ecce2904e4fb4a5bddd7a1bc3b

mszabo-wikia (approved):

PR #13667, opened 2017-09-04
Fix Link to "Begrüßungshilfe" help page

shaedrich:


mszabo-wikia:
i18n is handled externally and issues with it should be reported via the official channels

PR #13671, opened 2017-09-05
ADEN-5784 | fix outstream keyval on pvs without outstream slot

kvas-damian:
https://wikia-inc.atlassian.net/browse/ADEN-5784

@Wikia/adeng


fraszczakszymon:
@kvas-damian what do you think about clearing prebid state somewhere here:
https://github.com/Wikia/app/blob/dev/extensions/wikia/AdEngine/js/lookup/prebid/prebid.js#L34
Wouldn't it be better to solve the issue earlier? The same issue may influence kikimora?

kvas-damian:
The only problem with this solution is lack of public method for clearing bids. If you think it's ok to override `_bidsReceived` directly then we should be good to go.

BTW prebid keeps those bids for some reason: https://github.com/prebid/Prebid.js/blob/master/src/prebid.js#L97-L99

PR #13675, opened 2017-09-06
Improve VE handling of permissions & read-only mode

mszabo-wikia:
* Port https://github.com/wikimedia/mediawiki-extensions-VisualEditor/commit/d2a9aefba707544779be5dc209efd0417b60ea34 from upstream - `ApiVisualEditor` is not write mode
* make `ApiVisualEditorEdit` act like everyone was allowed `writeapi` (hack)

### DIFF
--- a/extensions/VisualEditor/ApiVisualEditorEdit.php
+++ b/extensions/VisualEditor/ApiVisualEditorEdit.php
@@ -10,8 +10,9 @@
 
 class ApiVisualEditorEdit extends ApiVisualEditor {
 
-	public function __construct( ApiMain $main, $name /*, Config $config */ ) {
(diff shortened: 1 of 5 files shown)


extensions/VisualEditor/ApiVisualEditorEdit.php
```diff
@@ -47,6 +48,7 @@ protected function saveWikitext( $title, $wikitext, $params ) {
 			true // enable write
 		);
 
+		$api->setShouldCheckWriteApiPermission( false );
```
macbre: Do we need to call this method twice?

macbre (approved):

PR #13686, opened 2017-09-07
Fix base text logic in Wall

mszabo-wikia:
Reality check

```
	/**
	 * Get the base page name, i.e. the leftmost part before any slashes
	 *
	 * @return String Base name
	 */
```
versus
```
$parts = explode( '/', $this->getText() );
# Don't discard the real title if there's no subpage involved
if ( count( $parts ) > 1 ) {
	unset( $parts[count( $parts ) - 1] );
}
return implode( '/', $parts );
```

### DIFF
--- a/extensions/wikia/Wall/WallMessage.class.php
+++ b/extensions/wikia/Wall/WallMessage.class.php
@@ -546,14 +546,25 @@ class WallMessage {
 		return $wallOwnerName;
 	}
 
+	private function getTitleBaseText( Title $title ): string {
+		$titleText = $title->getText();
+		$slashPosition = strpos( $titleText, '/' );
+
+		if ( $slashPosition === false ) {
+			return $titleText;
+		}
+
+		return substr( $titleText, 0, $slashPosition );
+	}
+
 	public function getWallOwner() {
 		if ( $this->title->inNamespace( NS_USER_WALL_MESSAGE ) ) {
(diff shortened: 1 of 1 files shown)


macbre (approved):

extensions/wikia/Wall/WallMessage.class.php
```diff
@@ -546,14 +546,25 @@ public function getMainPageText() {
...
+	}
+
 	public function getWallOwner() {
 		if ( $this->title->inNamespace( NS_USER_WALL_MESSAGE ) ) {
-			$ownerUserName = $this->title->getBaseText();
+			$ownerUserName = $this->getTitleBaseText( $this->title );
 			return User::newFromName( $ownerUserName, false );
 		}
 
 		// horribile visu: Forum threads treat board name as user!
```
shaedrich, line 566: > Forum threads treat board name as user

Shouldn't this be fixed?

mszabo-wikia: No.

shaedrich: But the comment shouldn't be removed either?

mszabo-wikia: It helps explain the situation to other developers who have not worked with this code before.

PR #13692, opened 2017-09-07
FEG JS cleanup PoC

mszabo-wikia:


extensions/wikia/PageHeader/PageHeader.setup.php
```diff
@@ -21,3 +21,24 @@
 
 // i18n
 $wgExtensionMessagesFiles[ 'PageHeader' ] = $dir . 'PageHeader.i18n.php';
+
+$wgResourceModules['ext.wikia.pageHeader.styles'] = [
+	'styles' => [ 'styles/index.scss' ],
+
+	// use domain sharding
+	'source' => 'common',
```
rogatty, line 29: Page Header styles use Theme Designer values, so we shouldn't cache the final CSS between wikis. Is that what this line does?

mszabo-wikia: It just makes sure we use `nocookie.net` domain to load this CSS - like AM does. The Sass variables still get passed via the URL. However maybe we could just use the local wiki domain.

rogatty (approved):

PR #13707, opened 2017-09-08
ADEN-5846 Separate variable to block on scroll pausing for fallback bid video

nikodamn:
@wwarren @TK-999 @Wikia/adeng


nikodamn:
@TK-999 handled it already. It's pushed to verify so I'm closing this ticket

PR #13721, opened 2017-09-12
XW-3875 | use Title instead of GlobalTitle to get main page url

latata:
@Wikia/x-wing


hakubo (approved):

xkxd (approved):

rogatty (changes requested):
It won't work when accessed from `www.wikia.com/api/v1`

PR #13724, opened 2017-09-12
SUS-2741: Improve performance of LogPager queries

mszabo-wikia:
Port of upstream bug [T123225](https://phabricator.wikimedia.org/T123225) - add auto increment primary keys to `change_tag` and `tag_summary` tables.

https://wikia-inc.atlassian.net/browse/SUS-2741


drozdo (approved):
LGTM

PR #13729, opened 2017-09-13
Fix german interlanguage prefix

shaedrich:
Remove not needed underscores in links


mszabo-wikia:
i18n is handled externally and issues with it should be reported via the official channels

PR #13744, opened 2017-09-15
MAIN-11309: fix for tabbers in preview

mszabo-wikia:
Legacy editor wasn't prepared for parser tag extensions with RL modules.

https://wikia-inc.atlassian.net/browse/MAIN-11309

### DIFF
--- a/extensions/3rdparty/tabber/tabber.js
+++ b/extensions/3rdparty/tabber/tabber.js
@@ -537,7 +537,7 @@ function tabberAutomaticOnLoad(tabberArgs)
 	if (!tabberArgs) { tabberArgs = {}; }
 	/* Taken from: http://simon.incutio.com/archive/2004/05/26/addLoadEvent */
 
-	$(function() {
+	mw.hook('wikipage.content').add(function () {
 		tabberAutomatic(tabberArgs);
 	});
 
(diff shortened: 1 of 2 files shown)


extensions/3rdparty/tabber/tabber.js
```diff
@@ -537,7 +537,7 @@ function tabberAutomaticOnLoad(tabberArgs)
 	if (!tabberArgs) { tabberArgs = {}; }
 	/* Taken from: http://simon.incutio.com/archive/2004/05/26/addLoadEvent */
```
rogatty: These comments seem outdated.

extensions/3rdparty/tabber/tabber.js
```diff
@@ -537,7 +537,7 @@ function tabberAutomaticOnLoad(tabberArgs)
...
 
-	$(function() {
+	mw.hook('wikipage.content').add(function () {
```
rogatty: I see we have multiple `wikia change` comments. Is that something we still do?

extensions/3rdparty/tabber/tabber.js
```diff
@@ -550,7 +550,7 @@ function tabberAutomaticOnLoad(tabberArgs)
 /* Run tabberAutomaticOnload() unless the "manualStartup" option was specified */
 if (typeof tabberOptions == 'undefined') {
 
-	tabberAutomaticOnLoad();
+	$(tabberAutomaticOnLoad);
```
rogatty, line 551: What about `tabberAutomaticOnLoad(tabberOptions)` below? It should be wrapped the same way or move `$()` back to `tabberAutomaticOnLoad`

rogatty: Or is `$()` obsolete if we use `mw.hook`?

rogatty (approved):

PR #13747, opened 2017-09-18
(PLATFORM-3181) Load some external assets over HTTPS

Grunny:
/cc @Wikia/core-platform-team

### DIFF
--- a/extensions/wikia/AssetsManager/AssetsConfig.class.php
+++ b/extensions/wikia/AssetsManager/AssetsConfig.class.php
@@ -62,8 +62,8 @@ class AssetsConfig {
 
 		if ( !empty( $wgUseJQueryFromCDN ) && empty( $params['noexternals'] ) ) {
 			$url = $minify
-				? '#external_http://ajax.googleapis.com/ajax/libs/jquery/' . static::JQUERY_VERSION . '/jquery.min.js'
-				: '#external_http://ajax.googleapis.com/ajax/libs/jquery/' . static::JQUERY_VERSION . '/jquery.js';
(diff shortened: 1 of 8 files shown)


extensions/wikia/AssetsManager/AssetsConfig.class.php
```diff
@@ -62,8 +62,8 @@ public static function getJQueryUrl( $combine, $minify, $params ) {
 
 		if ( !empty( $wgUseJQueryFromCDN ) && empty( $params['noexternals'] ) ) {
 			$url = $minify
-				? '#external_http://ajax.googleapis.com/ajax/libs/jquery/' . static::JQUERY_VERSION . '/jquery.min.js'
-				: '#external_http://ajax.googleapis.com/ajax/libs/jquery/' . static::JQUERY_VERSION . '/jquery.js';
+				? '#external_https://ajax.googleapis.com/ajax/libs/jquery/' . static::JQUERY_VERSION . '/jquery.min.js'
```
harnash, line 65: Where it is expanded/evaluated? I can't find any other reference in the code

harnash (approved):

extensions/wikia/AssetsManager/AssetsConfig.class.php
Grunny: They're handled here:
* https://github.com/Wikia/app/blob/4803384c48be7/extensions/wikia/AssetsManager/AssetsManager.class.php#L465-L466
* https://github.com/Wikia/app/blob/4803384c48be7/extensions/wikia/AssetsManager/AssetsManager.class.php#L489-L490

Hard to grep for since it only uses the prefix in the code. :)

PR #13760, opened 2017-09-19
SUS-1801: remove MASTER_POS_WAIT from updateCollation.php

mszabo-wikia:
In accordance with the principle of asynchronous replication there is no need to further stress the poor databases when this ghastly update crawls out from somewhere.

https://wikia-inc.atlassian.net/browse/SUS-1801

Example

### DIFF
--- a/maintenance/updateCollation.php
+++ b/maintenance/updateCollation.php
@@ -30,7 +30,6 @@ require_once( dirname( __FILE__ ) . '/Maintenance.php' );
 
 class UpdateCollation extends Maintenance {
 	const BATCH_SIZE = 50; // Number of rows to process in one batch
-	const SYNC_INTERVAL = 20; // Wait for slaves after this many batches
 
 	public function __construct() {
 		parent::__construct();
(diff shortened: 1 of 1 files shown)


frankfarmer:
The history of this code block in upstream is kind of funny

https://github.com/wikimedia/mediawiki/commit/a47f2dcb2d660fc53b49e0eec78f13010c180cc6

> Was used so Tim could sleep while a schema change was going on, but this is the kind of live hack that doesn't belong in core.

I'm not totally familiar with the context here -- is there a risk that this change might cause slaves to fall behind and stop receiving queries?

drozdo (commented):
batching such jobs is a good idea, here wfWaitForSlaves makes sense (if it talks to real slaves instead of consul address), it could also be replaced with simple sleep but basically batching and waiting a bit just to be more gentle on the db load is a good thing here

mszabo-wikia:
Keeping `MASTER_POS_WAIT` as per above.

PR #13774, opened 2017-09-20
Cake 2300 - overwrite DS api outputs for FC communities

nmonterroso:
@Wikia/cake 
https://wikia-inc.atlassian.net/browse/CAKE-2300

Overwriting DS API values for FC enabled communities. This is a little different from the previous PR because the DS Api is _always_ run on www.wikia.com. This means that the FandomCreator extension needs to be enabled there, and only when asking for data for FC-enabled communities from the DS api do we need to overwrite the data. To accomplish this we need to set the value of the `$wgFandomCreatorCommunityId` to `-1` for `www.wikia.com`. There's a check in `Hooks` now to make sure the community id we're trying to act on is valid, and the DS hook implementation looks at the wiki id for the external wiki that's being requested. I created https://wikia-inc.atlassian.net/browse/CAKE-2309 to make sure we properly enable the wiki factory variables.


drsnyder:
LGTM

lizlux:
Is all this for the sake of the global header and footer in discussions?

nmonterroso:
@lizlux It's so the community header in discussions has the correct theme and nav information. It also looks like it applies some styles to the page (links, button colors)

PR #13794, opened 2017-09-21
MAIN-9681: Fix unit test

KockaAdmiralac:
Attempt to fix a unit test broken in #12374.

The change is still untested as I don't have any kind of a development environment but @WikiaJenkins will hopefully complain if the tests still fail.


KockaAdmiralac:
@TK-999 is there any way of getting Jenkins to run unit tests on this branch?

PR #13796, opened 2017-09-21
IRIS-4908 | Stop parsing thread title for email subject

rogatty:
https://wikia-inc.atlassian.net/browse/IRIS-4908

We don't want thread titles which use `{{TemplateName}}` to be parsed. There is no need to escape HTML entities in an email subject, so we can use `rawParams`.

/cc @slayful @Grunny

### DIFF
--- a/extensions/wikia/Email/Controller/ForumController.class.php
+++ b/extensions/wikia/Email/Controller/ForumController.class.php
@@ -154,7 +154,7 @@ class ReplyForumController extends ForumController {
 
 	public function getSubject() {
 		return $this->getMessage( $this->getSubjectKey() )
-			// No need to encode HTML entities in subject
+			// We should not encode HTML entities in the subject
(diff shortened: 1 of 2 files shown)


extensions/wikia/Email/Controller/ForumController.class.php
```diff
@@ -153,7 +153,10 @@ protected function assertValidReplyParams() {
 	}
 
 	public function getSubject() {
-		return $this->getMessage( $this->getSubjectKey(), $this->titleText )->text();
+		return $this->getMessage( $this->getSubjectKey() )
+			// No need to encode HTML entities in subject
```
slayful: Wouldn't "We should not encode HTML entities in the subject" be more appropriate in this case?

slayful (approved):
LGTM with a suggestion

PR #13800, opened 2017-09-21
SUS-1442: Skip preferences and Phalanx check if VE is not available on the page

mszabo-wikia:
https://wikia-inc.atlassian.net/browse/SUS-1442


rogatty:
LGTM

PR #13805, opened 2017-09-22
XW-3875 | fix GlobaTitle::newMainPage

latata:
@Wikia/x-wing


hakubo (approved):

JamesHawking (approved):

includes/wikia/GlobalTitle.php
```diff
@@ -205,7 +205,7 @@ protected static function getWgArticlePath( $wikiId ) {
 	 *  constructor doesnt load anything from database. This is the place
 	 *  for that kind of things
 	 */
-	private function loadAll() {
+	protected function loadAll() {
```
rybmat, line 208: @latata why is it changed to protected?

latata: Can't we couldn't mock private method.

PR #13832, opened 2017-09-26
Do not render zero states if user wikis are hidden & "important" fields left blank

ghost:
As per MAIN-9681, if user wikis are hidden, the array is returned as empty. In cases wherein the user has hidden user wikis and left all the important fields blank, <tt>checkIfDisplayZeroStates</tt> now returns true and renders zero states, whereas before, even when hidden, user wikis were still queried and the array populated with entries, thus not rendering zero states. This fix should ensure that zero states are not rendered in cases wherein the user wikis are hidden and important fields left blank.

### DIFF
--- a/extensions/wikia/UserProfilePageV3/UserIdentityBox.class.php
+++ b/extensions/wikia/UserProfilePageV3/UserIdentityBox.class.php
@@ -115,11 +115,17 @@ class UserIdentityBox extends WikiaObject {
 			if ( !empty( $data['edits'] ) ) {
 				$data['edits'] = $this->wg->Lang->formatNum( $data['edits'] );
 			}
+			
+			// Zero states should not be shown if wikis are hidden
+			if ( $this->user->getGlobalPreference( 'hideEditsWikis' ) ) {
+				$data['showZeroStates'] = false;
+			} else {
+				$data['showZeroStates'] = $this->checkIfDisplayZeroStates( $data );
+			}
 
 			// other data operations
 			$this->getUserTags( $data );
 			$data = $this->extractBirthDate( $data );
-			$data['showZeroStates'] = $this->checkIfDisplayZeroStates( $data );
 		}
 
 		// Sanitize data to prevent XSS (VE-720)


mszabo-wikia (changes requested):
Consider the case where an user has the preference set to `false` but also does not have any favorite wikis set. In such a case, the zero-state should be shown, but it would not.

In general I'm not sure if there is a benefit to not showing zero state here. If the preference is `false` and all other fields are empty, then there is nothing useful to display in the masthead.

ghost:
If the <tt>hideEditsWikis</tt> preference is set to <tt>false</tt>, then <tt>checkIfDisplayZeroStates</tt> is called in the body of the <tt>else</tt> statement. In that case, if the favorite wikis array is empty, then zero states would be shown, as a value of <tt>true</tt> would be returned from the function. Or am I misunderstanding what you mean?

KockaAdmiralac:
You may want to update [this line](https://github.com/Wikia/app/blob/dev/extensions/wikia/UserProfilePageV3/tests/UserIdentityBoxTest.php#L29) in the unit test class to make the user mock have the correct method the test will be attempting to call like I did in [here](https://github.com/Wikia/app/commit/d86bb9dbc242bc1f7f430ee80376430e3f801b03#diff-a19a8297b14dba6d6450a2a2d5a2d342R185). Also, if you sent in a S:C before creating this pull request you could probably ask them for a ticket ID so engineers know which ticket to associate this pull request with (unless it's just a follow-up to MAIN-9681).

**Edit:** I wrote this before TK requested changes.

mszabo-wikia:
@usereizen hah yeah, what I meant is when `hideEditsWikis` is `true`, but there aren't actually any wikis that could be hidden to begin with, and all the other fields are empty as well. In that case, we should be showing a zero-state because there is no data at all.

ghost:
On second thought, perhaps you're right. I was seeking to replicate the behavior of the masthead prior to MAIN-9681, but I suppose in practice my changes add nothing of substance. Perhaps an adjustment to the masthead CSS when viewing other users' profiles would be a better alternative? To my eyes, other users' profiles that are rendered with zero-state styling appear a bit "vertically squished," for lack of a better term, specifically in the region of <tt>.masthead-info-lower</tt>.

mszabo-wikia:
The expected behavior of the masthead is to render the zero-state when there is no useful data to display. I don't think it makes a difference if this is because the top wikis are hidden, or if there are none at all.

PR #13834, opened 2017-09-26
XW-3821 | Add new URLs to app buttons in global footer + update translations

xkxd:
https://wikia-inc.atlassian.net/browse/XW-3821

@Wikia/x-wing 

## NOTE!

Regarding multiple changes in i18n jsons - it happened because of running `npm run setup` in DesignSystem's directory and it related to changing the order of keys to mirror the one from design-system-i18n repo - I can revert that if it's necessary.


rybmat (approved):

includes/wikia/models/DesignSystemGlobalFooterModel.class.php
```diff
@@ -159,12 +159,12 @@ public function getData() {
 					'type' => 'line-text',
 					'title' => [
 						'type' => 'translatable-text',
-						'key' => 'global-footer-community-apps-header'
+						'key' => 'global-footer-fandom-app-header'
```
latata: What about the case that `The FANDOM App` string is only for english?

xkxd: I'm aware of that, this was submitted before we agreed to change the ACs

extensions/wikia/DesignSystem/package.json
```diff
@@ -5,7 +5,7 @@
     "setup-template": "npm install && grunt"
   },
   "dependencies": {
-    "design-system-i18n": "Wikia/design-system-i18n#1.0.0",
+    "design-system-i18n": "Wikia/design-system-i18n#XW-4018",
```
latata: remember to replace that after merging translations

latata (approved):

latata (changes requested):
Update ds-i18n version

latata (approved):

PR #13839, opened 2017-09-27
SUS-1276 | refactor search code for an old corporate wiki

macbre:
https://wikia-inc.atlassian.net/browse/SUS-1276

Do not rely on `wgEnableWikiaHomePageExt` WikiFactory variable.

### DIFF
--- a/extensions/wikia/Search/WikiaSearchAjaxController.class.php
+++ b/extensions/wikia/Search/WikiaSearchAjaxController.class.php
@@ -27,7 +27,7 @@ class WikiaSearchAjaxController extends WikiaController {
 		$crossWikia = $this->request->getBool( 'crossWikia' );
 		$hub = false;
 
-		$isInterWiki = $crossWikia ? true : WikiaPageType::isCorporatePage();
+		$isInterWiki = $crossWikia || WikiaPageType::isCorporatePage();
 
 		$cityId = $isInterWiki ? 0 : $this->wg->CityId;
 
--- a/extensions/wikia/Search/WikiaSearchAjaxController.class.php
+++ b/extensions/wikia/Search/WikiaSearchAjaxController.class.php
@@ -27,10 +27,7 @@ class WikiaSearchAjaxController extends WikiaController {
 		$crossWikia = $this->request->getBool( 'crossWikia' );
 		$hub = false;
 
-		// TODO: refactor to avoid using wgEnableWikiaHomePageExt
-		$isCorporateWiki = !empty( $this->wg->EnableWikiaHomePageExt );
-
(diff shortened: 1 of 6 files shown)


includes/wikia/Wikia.php
```diff
@@ -85,6 +85,7 @@ class Wikia {
 
 	const COMMUNITY_WIKI_ID = 177; // community.wikia.com
 	const NEWSLETTER_WIKI_ID = 223496; // wikianewsletter.wikia.com
```
shaedrich, line 87: This wiki has one page only oO

macbre: But way more [images](http://wikianewsletter.wikia.com/wiki/Special:Images)

extensions/wikia/Search/WikiaSearchAjaxController.class.php
```diff
@@ -27,10 +27,7 @@ public function getNextResults() {
 		$crossWikia = $this->request->getBool( 'crossWikia' );
 		$hub = false;
 
-		// TODO: refactor to avoid using wgEnableWikiaHomePageExt
-		$isCorporateWiki = !empty( $this->wg->EnableWikiaHomePageExt );
-
-		$isInterWiki = $crossWikia ? true : $isCorporateWiki;
+		$isInterWiki = $crossWikia ? true : WikiaPageType::isCorporatePage();
```
mszabo-wikia: `$crossWikia || WikiaPageType::isCorporatePage();` ?

extensions/wikia/Search/WikiaSearchController.class.php
```diff
@@ -881,8 +881,7 @@ protected function templateExists( $name ) {
 	 * @see SearchControllerTest::testIsCorporateWiki
 	 */
 	protected function isCorporateWiki() {
-		// TODO: refactor to avoid using wgEnableWikiaHomePageExt
-		return !empty( $this->wg->EnableWikiaHomePageExt );
+		return WikiaPageType::isCorporatePage();
```
mszabo-wikia, line 884: we could kill this method wrapper here and use `WikiaPageType::isCorporatePage` directly

mszabo-wikia (changes requested):

extensions/wikia/Search/WikiaSearchController.class.php
macbre: I'm going to review all uses of `WikiaPageType::isCorporatePage` and eventually get rid of it. Search is the only place that has a different functionality on www.wikia.com.

includes/wikia/Wikia.php
shaedrich: So it's just a image storage with no other use? They could be stored elsewhere.

PR #13847, opened 2017-09-28
SUS-1276 | CityVisualization refactored

macbre:
Removed unused methods (including a few in `PromoImage` class) + make those used internally private methods

https://wikia-inc.atlassian.net/browse/SUS-1276


mszabo-wikia (approved):
Let's fix the tests too :)

PR #13891, opened 2017-10-03
ADEM-5982 Fix calculating preferred video size and prevent content jumping

fraszczakszymon:


extensions/wikia/AdEngine/js/video/player/ui/dynamicReveal.js
```diff
@@ -20,7 +20,9 @@ define('ext.wikia.adEngine.video.player.ui.dynamicReveal', [
 				if (!slotExpanded) {
 					slotTweaker.expand(params.slotName);
 					slotExpanded = true;
-					video.ima.dispatchEvent('wikiaSlotExpanded');
+					setTimeout(function () {
```
kvas-damian: Can we add few words why do we need to delay this operation?

extensions/wikia/AdEngine/js/video/player/porvata/floaterConfiguration.js
```diff
@@ -43,7 +40,9 @@ define('ext.wikia.adEngine.video.player.porvata.floaterConfiguration', [
 			onBeforeDetach: function (floatingContext) {
 				var viewport = floatingContext.elements.viewport;
 
-				viewport.style.height = viewport.offsetHeight + 'px';
+				// Magic 14 is related to #INCONTENT_WRAPPER > #INCONTENT_PLAYER margins
+				// and prevents jumping when video is going to float
+				viewport.style.height = (viewport.offsetHeight - 14) + 'px';
```
kvas-damian, line 45: Maybe we could calculate `14` based on DOM node properties (margins in this case)?

fraszczakszymon: It would require using `getComputedStyle` - is it worth?

kvas-damian: Nein! Let's leave it as it is. Comment is enough.

kvas-damian (approved):

nandy-andy (approved):

PR #13901, opened 2017-10-04
XW-3280 | check if "submit" element is set in array on Special:JSPages

xkxd:
https://wikia-inc.atlassian.net/browse/XW-3280

@Wikia/x-wing

### DIFF
--- a/extensions/wikia/ContentReview/controllers/templates/JSPagesSpecialController_index.php
+++ b/extensions/wikia/ContentReview/controllers/templates/JSPagesSpecialController_index.php
@@ -31,7 +31,7 @@
 				<td><?= $page['latestReviewed']['statusMessage'] ?></td>
 				<td><?= $page['liveRevision']['statusMessage'] ?></td>
 				<td>
-					<?php if ( $page['submit'] ): ?>
+					<?php if ( isset( $page['submit'] ) ): ?>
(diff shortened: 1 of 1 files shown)


extensions/wikia/ContentReview/controllers/templates/JSPagesSpecialController_index.php
```diff
@@ -31,7 +31,7 @@
 				<td><?= $page['latestReviewed']['statusMessage'] ?></td>
 				<td><?= $page['liveRevision']['statusMessage'] ?></td>
 				<td>
-					<?php if ( $page['submit'] ): ?>
+					<?php if ( isset( $page['submit'] ) ): ?>
```
rybmat: I would use !empty($page[''submit']) since $page['submit'] = false would resolve this if to false and isset would return true in that case

rybmat (changes requested):

extensions/wikia/ContentReview/controllers/templates/JSPagesSpecialController_index.php
xkxd: thx, this'll be more accurate!

hakubo (approved):

rybmat (approved):

PR #13908, opened 2017-10-04
XW-4049 | update liftigniter-metadata client, handle NO CONTENT response

rybmat:
https://wikia-inc.atlassian.net/browse/XW-4049

Followup to changes in liftigniter-metadata service: https://github.com/Wikia/pandora/pull/3297
When there is no metadata associated with given article, service will return 204 instead of 404. This will prevent throwing errors by ApiClient.

@Wikia/x-wing


latata (approved):

lib/Swagger/src/Liftigniter/Metadata/Api/GeosApi.php
```diff
@@ -171,15 +171,15 @@ public function getGeosWithHttpInfo()
             );
 
             return array($this->apiClient->getSerializer()->deserialize($response, '\Swagger\Client\Liftigniter\Metadata\Models\Geo[]', $httpHeader), $statusCode, $httpHeader);
-        } catch (ApiException $apiException) {
-            switch ($apiException->getCode()) {
+        } catch (ApiException $e) {
```
xkxd, line 174: can you rename this to `$error` / `$response` or thx?

xkxd (approved):

lib/Swagger/src/Liftigniter/Metadata/Api/GeosApi.php
rybmat: @xkxd  I can but I don't want to :p This is autogenerated code

PR #13913, opened 2017-10-04
SUS-2766: Cleanup Phalanx after Kubernetes migration

mszabo-wikia:
* no need to call `/reload` anymore - rule synchronization is handled by 🐇 
* use k8s url instead of consul

https://wikia-inc.atlassian.net/browse/SUS-2766

### DIFF
--- a/extensions/wikia/PhalanxII/services/PhalanxService.class.php
+++ b/extensions/wikia/PhalanxII/services/PhalanxService.class.php
@@ -2,6 +2,8 @@
 
 use Wikia\Service\Gateway\ConsulUrlProvider;
 use Wikia\Rabbit\ConnectionBase;
+use Wikia\Service\Gateway\KubernetesUrlProvider;
+use Wikia\Service\Gateway\UrlProvider;
 
 /**
  * @method PhalanxService setLimit( int $limit )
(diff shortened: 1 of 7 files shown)


extensions/wikia/PhalanxII/services/PhalanxService.class.php
```diff
@@ -233,22 +206,14 @@ private function sendToPhalanxDaemon( $action, $parameters ) {
 				}
 			}
 
-			// SUS-964: Give reload requests more time to succeed
-			// Reload requests are only sent upon saving/modifying a block,
-			// so using a higher value here won't affect site performance
-			if ( $action === 'reload' ) {
-				$options['timeout'] = static::PHALANX_SERVICE_RELOAD_TIMEOUT;
-			}
```
macbre, line 241: Are we sure that the `reload` requests are now faster than they used to be?

macbre (approved):

extensions/wikia/PhalanxII/services/PhalanxService.class.php
mszabo-wikia: We now use RabbitMQ instead, so no need to set special timeout 😉

macbre: Ah, true :)

macbre (approved):

PR #13920, opened 2017-10-05
PLATFORM-3218: don't prepend RL urls with server address

themech:
RL for local assets should load over https when a page was open using https. Unfortunately `wfExpandUrl` doesn't really produce relative addresses - in order for it to do so, we would need to remove protocol from `wgServer`, but this may open a pandora box. We also shouldn't output http or https links, as we have article cache shared between http and https. Http links will break when used on https pages and https links will break when used on language wikis. So it looks like we should try local urls.

The change is hidden behind `wgEnableLocalResourceLoaderLinks` so we can test it on devboxes and few wikis before enabling this globally.


extensions/wikia/AdEngine/tests/AdEngine2ContextServiceTest.php
```diff
@@ -230,6 +230,7 @@ public function testGetContext(
 		$customDartKvs = 'a=b;c=d';
 		$catId = WikiFactoryHub::CATEGORY_ID_LIFESTYLE;
 		$shortCat = 'shortcat';
+		// @todo(mech); change this once $wgEnableLocalResourceLoaderLinks is enabled globally
```
WikiaTech, line 233: INFO Complete the task associated to this "TODO" comment. [rule](https://sonarqube.wikia-inc.com/coding_rules#rule_key=php%3AS1135)

WikiaTech:
* ![INFO][INFO] 1 info

[INFO]: https://sonarsource.github.io/sonar-github/severity-info.png 'Severity: INFO'

harnash (approved):

PR #13926, opened 2017-10-06
Load amazon ads over https

themech:
Switching the protocol, as I want it to work on https pages.


kvas-damian:
This code is already disabled on production. We're going to remove it in next couple weeks.

kvas-damian (approved):

PR #13952, opened 2017-10-10
XW-4089 | Connect GA and Internal Warehouse to JW tracking

bkoval:
@Wikia/x-wing @jcellary

### DIFF
--- a/extensions/wikia/ArticleVideo/scripts/featured-video.jwplayer.tracking.js
+++ b/extensions/wikia/ArticleVideo/scripts/featured-video.jwplayer.tracking.js
@@ -1,4 +1,4 @@
-define('wikia.articleVideo.featuredVideo.tracking', [], function () {
+define('wikia.articleVideo.featuredVideo.tracking', ['wikia.tracker'], function (tracker) {
 	var state = getDefaultState(),
 		defaultGACategory = 'featured-video',
 		wasAlreadyUnmuted = false,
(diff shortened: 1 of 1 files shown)


extensions/wikia/ArticleVideo/scripts/featured-video.jwplayer.tracking.js
```diff
@@ -28,16 +28,21 @@ define('wikia.articleVideo.featuredVideo.tracking', [], function () {
...
 
-		var finalGAData = {
+		var trackingData = {
 			action: gaData.action || 'click',
 			category: gaCategory,
 			label: gaData.label,
 			//value tracks sound state: 1 for muted, 0 for unmuted
-			value: Number(playerInstance.getMute())
+			value: Number(playerInstance.getMute()),
+
```
bkoval, line 37: @jcellary We would propose to use (default) trackingevent as an eventName for tracking all video-related events, is that fine with you?

jcellary: I think that's fine, but could you paste me the full url that you'd be sending? (e.g. what you see in chrome on your devbox under networking) I just need to make sure we can easily filter them out by a short set of criteria.

latata (approved):

extensions/wikia/ArticleVideo/scripts/featured-video.jwplayer.tracking.js
bkoval:
```
https://beacon.wikia-services.com/__track/special/trackingevent?c=509&x=harrypotter&a=60661&lc=en&n=0&u=0&s=oasis&beacon=ThisIsFake&cb=24728&videoId=FYykS9se&player=jwplayer&ga_action=impression&ga_category=featured-video&ga_label=load&ga_value=1
```
would be an exemplary request

hakubo (approved):

PR #13970, opened 2017-10-11
ADEN-6020 | Fix TOP_BUTTON_WIDE rade condition

kvas-damian:
https://wikia-inc.atlassian.net/browse/ADEN-6020

Important changes in `adEngine`:
* `customAdsLoader` is executed synchronously and returns output from a template
* providers hook on slots after the event happens to make sure that code was already executed


dianafa:
LGTM

PR #13973, opened 2017-10-12
XW-4114 | Updating tracking for JW and Ooyala

bkoval:
- Both JW and Ooyala are sending events in internal DW under eventName **videoplayerevent**
- Ooyala sends videoId and player fields as JW is

@Wikia/x-wing @jcellary

### DIFF
--- a/extensions/wikia/ArticleVideo/scripts/featured-video.jwplayer.tracking.js
+++ b/extensions/wikia/ArticleVideo/scripts/featured-video.jwplayer.tracking.js
@@ -2,7 +2,7 @@ define('wikia.articleVideo.featuredVideo.tracking', ['wikia.tracker'], function 
 	var state = getDefaultState(),
 		defaultGACategory = 'featured-video',
 		//This will replace 'trackingevent' in internal tracker url path
-		eventName: 'videoplayerevent',
+		eventName = 'videoplayerevent',
 		wasAlreadyUnmuted = false,
 		depth = 0,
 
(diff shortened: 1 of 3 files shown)


jcellary:
Don't have full code context, but the direction is good ;)

rybmat (approved):

extensions/wikia/ArticleVideo/scripts/featured-video.jwplayer.tracking.js
```diff
@@ -1,6 +1,8 @@
 define('wikia.articleVideo.featuredVideo.tracking', ['wikia.tracker'], function (tracker) {
 	var state = getDefaultState(),
 		defaultGACategory = 'featured-video',
+		//This will replace 'trackingevent' in internal tracker url path
+		eventName = 'videoplayerevent',
```
xkxd, line 5: maybe we should store such things in different "constants" module / config module?

xkxd (approved):

extensions/wikia/ArticleVideo/scripts/featured-video.jwplayer.tracking.js
bkoval: We could, but we use these values literally - storing them separately would make sense if we'd use the entire config objects somewhere but we just use the 2 variables in the tracker for now, so this seems easier and more readable for me. What do you think?

JamesHawking (approved):

PR #13998, opened 2017-10-16
Fixes to WAM UI

jcellary:
Yet another attempt to improve WAMs UI. Few fixes:
- Currently when entering a different date and clicking somewhere the UI used timestamp in miliseconds instead of expected one in seconds which resulted in redirect to page with default date. Fixed that to properly interpret the timestamp
- Added setting of current date in the selector to what we actually got from the Api. This should help knowing what one is looking at in case of selecting a value out of range or when backend is missing newest data.
- Removed fallback to data from two days before, as after last fixes when there is no date or it is out of range we simply take the most recent value.

@macbre @Grunny @TK-999


macbre (approved):

mszabo-wikia (changes requested):
Let's remove the unused private methods from `WAMPageController`: `redirectIfFirstTab`, `redirectIfUnknownTab`.

jcellary:
Removed!

mszabo-wikia (approved):

PR #13999, opened 2017-10-16
SUS-2956: Migrate MultiLookup updates to MediaWiki task

mszabo-wikia:
https://wikia-inc.atlassian.net/browse/SUS-2956

### DIFF
--- a/extensions/wikia/SpecialMultipleLookup/MultiLookupHooks.php
+++ b/extensions/wikia/SpecialMultipleLookup/MultiLookupHooks.php
@@ -1,0 +1,35 @@
+<?php
+
+class MultiLookupHooks {
+	static public function onContributionsToolLinks( $id, $nt, &$links ) {
+		$user = RequestContext::getMain()->getUser();
+
+		if ( $id == 0 && $user->isAllowed( 'multilookup' ) ) {
+			$attribs = [
+				'href' => 'http://community.wikia.com/wiki/Special:MultiLookup?target=' . urlencode( $nt->getText() ),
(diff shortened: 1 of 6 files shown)


extensions/wikia/SpecialMultipleLookup/MultiLookupHooks.php
```diff
@@ -0,0 +1,35 @@
...
+		$ip = $recentChange->getAttribute( 'rc_ip' );
+
+		if ( IP::isIPAddress( $ip ) ) {
+			global $wgCityId;
+
+			$task = ( new \Wikia\Tasks\Tasks\MultiLookupTask() )
+				->wikiId( $wgCityId );
+			$task->call( 'updateMultiLookup', $ip );
+			$task->queue();
+		}
```
macbre, line 33: Does it make sense to log cases when `$ip` is not a proper IPv4/IPv6?

macbre (approved):
As discussed offline, let's make sure that `SpecialMultipleLookup.php` is enabled site-wide and move the `include` to app's `DefaultSettings.php` file.

extensions/wikia/SpecialMultipleLookup/MultiLookupHooks.php
TK-999: If IP is not valid, we do not queue task here

mszabo-wikia: It can happen if some wiki has `wgPutIPinRC = false;`, unlikely, but possible.

PR #14000, opened 2017-10-16
SUS-2990 | Remove join, query wikicities for user_name

JamesHawking:
[Ticket](https://wikia-inc.atlassian.net/browse/SUS-2990)

This PR removes remaining queries to local db's.

### DIFF
--- a/extensions/wikia/SpecialNewWikis/SpecialNewWikis_body.php
+++ b/extensions/wikia/SpecialNewWikis/SpecialNewWikis_body.php
@@ -169,9 +169,11 @@ class NewWikisPage extends AlphabeticPager {
 	}
 	/* need to be overwrite because of not unique col name (city_id) */
 
-	function reallyDoQuery( $offset, $limit, $descending ) {
+	function reallyDoQuery( $offset, $limit, $descending, $info = [] ) {
 		$fname = __METHOD__ . ' (' . get_class( $this ) . ')';
-		$info = $this->getQueryInfo();
+		if ( !isset( $info ) ) {
+			$info = $this->getQueryInfo();
+		}
 		$tables = $info['tables'];
 		$fields = $info['fields'];
 		$conds = isset( $info['conds'] ) ? $info['conds'] : array();
--- a/extensions/wikia/SpecialNewWikis/SpecialNewWikis_body.php
+++ b/extensions/wikia/SpecialNewWikis/SpecialNewWikis_body.php
@@ -169,11 +169,10 @@ class NewWikisPage extends AlphabeticPager {
 	}
 	/* need to be overwrite because of not unique col name (city_id) */
 
-	function reallyDoQuery( $offset, $limit, $descending, $info = [] ) {
+	function reallyDoQuery( $offset, $limit, $descending ) {
 		$fname = __METHOD__ . ' (' . get_class( $this ) . ')';
-		if ( !isset( $info ) ) {
-			$info = $this->getQueryInfo();
-		}
+		$info = $this->getQueryInfo();
+
 		$tables = $info['tables'];
 		$fields = $info['fields'];
 		$conds = isset( $info['conds'] ) ? $info['conds'] : array();
--- a/includes/Pager.php
+++ b/includes/Pager.php
@@ -97,6 +97,9 @@ abstract class IndexPager extends ContextSource implements Pager {
 	public $mIsFirst;
 	public $mIsLast;
 
+	/** User names associated with the result */
+	private $mUsers;
+
 	protected $mLastShown, $mFirstShown, $mPastTheEndIndex, $mDefaultQuery, $mNavigationBar;
 
 	/**
@@ -177,6 +180,7 @@ abstract class IndexPager extends ContextSource implements Pager {
 			$this->getQueryInfo()
 		);
 		$this->extractResultInfo( $this->mOffset, $queryLimit, $this->mResult );
+		$this->mUsers = $this->fetchUsersForHistoryPage();
 		$this->mQueryDone = true;
 
 		$this->preprocessResults( $this->mResult );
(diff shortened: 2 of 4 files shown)


extensions/wikia/SpecialNewWikis/SpecialNewWikis_body.php
```diff
@@ -169,9 +169,11 @@ function getQueryInfo() {
 	}
 	/* need to be overwrite because of not unique col name (city_id) */
 
-	function reallyDoQuery( $offset, $limit, $descending ) {
+	function reallyDoQuery( $offset, $limit, $descending, $info = [] ) {
```
mszabo-wikia: This will cause `PHP Warning` because the method declaration will not be compatible with the super method declaration.

includes/specials/SpecialAllmessages.php
```diff
@@ -291,7 +291,7 @@ public static function getCustomisedStatuses( $messageNames, $langcode = 'en', $
 	 *  This function normally does a database query to get the results; we need
 	 * to make a pretend result using a FakeResultWrapper.
 	 */
-	function reallyDoQuery( $offset, $limit, $descending ) {
+	function reallyDoQuery( $offset, $limit, $descending, $info = [] ) {
```
mszabo-wikia: Is this needed here?

mszabo-wikia (changes requested):

extensions/wikia/SpecialNewWikis/SpecialNewWikis_body.php
JamesHawking: We've added $info = [] to other methods, so it won't warn.

includes/specials/SpecialAllmessages.php
JamesHawking: Same as above basically.

extensions/wikia/SpecialNewWikis/SpecialNewWikis_body.php
TK-999: It will - parent method in `IndexPager` superclass does not take `$info` parameter.

mszabo-wikia (approved):

includes/specials/SpecialAllmessages.php
ghost: @TK-999 — yes, as child and parent methods declarations have to be identical.

includes/Pager.php
```diff
@@ -284,20 +289,22 @@ function getSqlComment() {
 	 * @param $descending Boolean: query direction, false for ascending, true for descending
 	 * @return ResultWrapper
 	 */
-	function reallyDoQuery( $offset, $limit, $descending ) {
+	function reallyDoQuery( $offset, $limit, $descending, $info = [] ) {
 		$fname = __METHOD__ . ' (' . $this->getSqlComment() . ')';
-		$info = $this->getQueryInfo();
+		if ( !isset( $info ) ) {
+			$info = $this->getQueryInfo();
```
mszabo-wikia: Why is standard subclassing & implementing of `abstract` function `getQueryInfo` not viable here?

includes/Pager.php
```diff
@@ -309,10 +316,48 @@ function reallyDoQuery( $offset, $limit, $descending ) {
...
+
 		return new ResultWrapper( $this->mDb, $res );
 	}
 
 	/**
+	 * Fetch user names from ExternalSharedDB.
+	 * @see SUS-2990
+	 * @return array
+	 */
+	protected function fetchUsersForHistoryPage() {
```
mszabo-wikia: This shouldn't be in `IndexPager` class.

mszabo-wikia (changes requested):

includes/Pager.php
JamesHawking: In order to add $info as a parameter we need this if to override only default value of $info which is an empty array, otherwise, it should use $info from the argument.

includes/Pager.php
JamesHawking: Any proposals? :)

mszabo-wikia: We could move this logic to `HistoryPager` class.

mszabo-wikia (approved):

includes/actions/HistoryAction.php
```diff
@@ -372,21 +371,59 @@ function getQueryInfo() {
...
+		$ids = [];
+		foreach ( $res as $row ) {
+			if ( $row->rev_user ) {
+				$ids[] = $row->rev_user;
+			}
+		}
+
+		$dbr = wfGetDB( DB_SLAVE, [], $wgExternalSharedDB );
+		$users =
+			$dbr->select( '`user`', [ 'user_id', 'user_name' ], [ 'user_id' => array_unique( $ids ) ], __METHOD__ );
```
macbre: Why new line here? :)

includes/actions/HistoryAction.php
```diff
@@ -395,7 +432,7 @@ function doBatchLookups() {
 		$this->mResult->seek( 0 );
 		$batch = new LinkBatch();
 		foreach ( $this->mResult as $row ) {
-			if( !is_null( $row->user_name ) ) {
+			if( !empty( $row->user_name ) ) {
```
macbre: What if user name is `0`? :)

```
=== Found user ===
	ID: 174051
	Username: 0
	Name: 
	Email:
```

`isset` should be fine.

includes/actions/HistoryAction.php
JamesHawking: Shit, autoformatting, probably i have bad rules.

macbre (approved):

includes/actions/HistoryAction.php
ghost: Good catch, @macbre

ghost (approved):

includes/actions/HistoryAction.php
```diff
@@ -432,7 +436,7 @@ function doBatchLookups() {
 		$this->mResult->seek( 0 );
 		$batch = new LinkBatch();
 		foreach ( $this->mResult as $row ) {
-			if( !empty( $row->user_name ) ) {
+			if( !isset( $row->user_name ) ) {
```
macbre: I believe it should be `isset` - without the negation.

macbre (changes requested):
See above

includes/actions/HistoryAction.php
ghost: Aye, goes without saying…

macbre (approved):

PR #14019, opened 2017-10-17
SUS-2999: Use jQuery to notify Image Lazy Loading about scroll

mszabo-wikia:
This way we can ensure IE11 compatibility.
https://wikia-inc.atlassian.net/browse/SUS-2999


extensions/wikia/ImageLazyLoad/js/ImgLzy.module.js
```diff
@@ -27,12 +23,10 @@ define('wikia.ImgLzy', ['jquery', 'wikia.log'], function ($, log) {
 			this.onScroll();
 
 			// Scan & load images every 250 ms if the user is scrolling
-			window.addEventListener('scroll', throttled);
+			$(window).on('scroll', throttled);
```
macbre, line 26: :+1:

macbre (approved):

PR #14034, opened 2017-10-18
introduce User::whoAre() for convenience

ghost:
This is a sister method to `User::whoIs()`.  It takes an array of `user_id` and returns an array mapping each `user_id` to the corresponding `user_name`.

I did not want to extend `User::whoIs()` and make it returning a single `user_name` as a string or an array of those when provided with a single `user_id` as an integer or an array of those, respectively.  It is a core MW method.  And functions taking input of different types and producing output of different types are not exactly to my taste.

Example usage:

```
michal@dev-michal:/usr/wikia/source/app$ SERVER_ID=177 php maintenance/eval.php 
> var_dump(User::whoAre([-3, 0, 1, 8, 8, 8]));
/usr/wikia/source/app/maintenance/eval.php(88) : eval()'d code:1:
array(4) {
  [-3] =>
  string(0) ""
  [0] =>
  string(25) "An unknown anonymous user"
  [1] =>
  string(8) "W-Jasonr"
  [8] =>
  string(6) "Sannse"
}
```

### DIFF
--- a/includes/User.php
+++ b/includes/User.php
@@ -592,6 +592,40 @@ class User implements JsonSerializable {
 	}
 
 	/**
+	 *
+	 * @param $ids Array User IDs
+	 * @return Array User ID to User name mapping
+	 */
+	public static function whoAre( Array $ids, $source = DB_SLAVE ): Array {
+		global $wgExternalSharedDB;
+		$sdb = wfGetDB( $source, [], $wgExternalSharedDB );
+		$res = $sdb->select(
+			'`user`',
+			[ 'DISTINCT user_id', 'user_name' ],
+			[ 'user_id' => $ids ],
+			__METHOD__
+		);
+
+		// Pre-fill the returned array with empty strings
+		// so that missing users are not skipped.
+		// It makes further iterating over the array
+		// and handling anons and missing users a little
+		// bit easier.
+		$users = array_fill_keys( $ids, '' );
+
+		// This is optional, let's decide whether it is actually
+		// convenient to handle it here or programmers using
+		// this method should handle it themselves.
+		$users[0] = wfMessage( 'oasis-anon-user' )->escaped();
+
+		foreach ( $res as $row ) {
+			$users[ $row->user_id ] = (string) $row->user_name;
+		}
+
+		return $users;
+	}
+
+	/**
 	 * Get the real name of a user given their user ID
 	 *
 	 * @param $id Int User ID


WikiaTech:

1. ![MAJOR][MAJOR] [User.php#L644](https://github.com/Wikia/app/blob/c1ba24b90b667fd0849d066ec6d547f8c6f8d11a/includes/User.php#L644): Reduce the number of returns of this function 5, down to the maximum allowed 3. [rule](https://sonarqube.wikia-inc.com/coding_rules#rule_key=php%3AS1142)

includes/User.php
```diff
@@ -592,6 +592,40 @@ public static function whoIs( $id ) {
...
+			[ 'user_id' => $ids ],
+			__METHOD__
+		);
+
+		// Pre-fill the returned array with empty strings
+		// so that missing users are not skipped.
+		// It makes further iterating over the array
+		// and handling anons and missing users a little
+		// bit easier.
+		$users = array_fill_keys( $ids, '' );
```
mszabo-wikia, line 614: Should we log missing user names?

includes/User.php
```diff
@@ -592,6 +592,40 @@ public static function whoIs( $id ) {
...
+	 */
+	public static function whoAre( Array $ids, $source = DB_SLAVE ): Array {
+		global $wgExternalSharedDB;
+		$sdb = wfGetDB( $source, [], $wgExternalSharedDB );
+		$res = $sdb->select(
+			'`user`',
+			[ 'DISTINCT user_id', 'user_name' ],
+			[ 'user_id' => $ids ],
+			__METHOD__
+		);
```
mszabo-wikia, line 607: Should we cache the results in some way?

mszabo-wikia (changes requested):

includes/User.php
ghost: I thought about it.  We can order and serialize the input array, `sha1` it, include in the memcache key, and cache the result for some short period of time.  Caching for long period of time will result with a large number of keys and pose a few questions such as:

* how to selectively invalidate cache after a user rename?
* what is the chance that we will query for the exact same set of names more than once?

includes/User.php
ghost: I don't think so.  I would leave it to the programmer using this method to deal with the output properly.  The sole purpose of this mapping is to provide a convenient mapping.  We can call this method within some other method created for data integrity checks.

mszabo-wikia (approved):

includes/User.php
```diff
@@ -592,6 +592,40 @@ public static function whoIs( $id ) {
...
+	 * @param $ids Array User IDs
+	 * @return Array User ID to User name mapping
+	 */
+	public static function whoAre( Array $ids, $source = DB_SLAVE ): Array {
+		global $wgExternalSharedDB;
+		$sdb = wfGetDB( $source, [], $wgExternalSharedDB );
+		$res = $sdb->select(
+			'`user`',
+			[ 'DISTINCT user_id', 'user_name' ],
+			[ 'user_id' => $ids ],
```
macbre, line 605: Shouldn't we call `array_unique` on `$ids` instead of relying on MySQL doing `DISTINCT`?

includes/User.php
```diff
@@ -592,6 +592,40 @@ public static function whoIs( $id ) {
...
+		// so that missing users are not skipped.
+		// It makes further iterating over the array
+		// and handling anons and missing users a little
+		// bit easier.
+		$users = array_fill_keys( $ids, '' );
+
+		// This is optional, let's decide whether it is actually
+		// convenient to handle it here or programmers using
+		// this method should handle it themselves.
+		$users[0] = wfMessage( 'oasis-anon-user' )->escaped();
```
macbre, line 619: I'd rather return `false` to make the behaviour here similar to the one of `User::whoIs`:

```
@return String|false The corresponding username
```

includes/User.php
ghost: Why?

includes/User.php
ghost: Consistency is a value, but being consistently wrong?  Considering PHP 7 returning type declarations in function signatures, how would you declare `User::whoIs()`?  User names cannot be empty as in `''`.  That's why it is a sign of a problem when they are.  There is no need to return `false`, or even worse `null`.

PR #14035, opened 2017-10-18
SUS-3068 query wikicities rather than wikicities_c1

ghost:
https://wikia-inc.atlassian.net/browse/SUS-3068

Get rid of
```
SELECT 
log_id,user_name  FROM `logging`,`wikicities_c1`.`user`  WHERE log_id BETWEEN 242701 AND 242692 AND log_user = user_id
```

in favour of fetching users from `wikicities` with `User::whoAre() method introduced in https://github.com/Wikia/app/pull/14034

https://github.com/Wikia/app/pull/14034 is a dependency for this one.

### DIFF
--- a/maintenance/populateLogUsertext.php
+++ b/maintenance/populateLogUsertext.php
@@ -69,6 +69,8 @@ class PopulateLogUsertext extends LoggedUpdateMaintenance {
 				$ids[] = $row->log_user;
 			}
 
+			$res->rewind();
+
 			$users = User::whoAre( $ids );
 
 			$db->begin();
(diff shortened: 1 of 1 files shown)


mszabo-wikia (approved):
Are we sure this script is still needed?

ghost:
@TK-999 — I've asked that question in the ticket.  It might not be needed once we remove `log_user_text` column.

maintenance/populateLogUsertext.php
```diff
@@ -53,16 +53,34 @@ protected function doDBUpdates() {
...
+
+			$ids = [];
+
+			foreach ( $res as $row ) {
+				$ids[] = $row->log_user;
+			}
+
+			$res->rewind();
+
+			$users = User::whoAre( $ids );
```
macbre, line 74: Will `$ids` have unique values or should we first call `array_unique` over them?

ghost: `User::whoAre()` uses MySQL's `DISTINCT`.  We can consider using `array_unique` inside the method.  We are discussing that in the relevant pull request.

PR #14045, opened 2017-10-20
ADEN-6077 Kill BTF slots from creative even for PAL pages

v3rth:


extensions/wikia/AdEngine/js/provider/btfBlocker.js
```diff
@@ -71,16 +75,15 @@ define('ext.wikia.adEngine.provider.btfBlocker', [
...
+			var context = adContext.getContext();
 
 			log('startBtfQueue', log.levels.info.debug, logGroup);
 
 			if (btfQueueStarted) {
 				return;
 			}
 
-			if (context.opts.premiumAdLayoutEnabled && !roadblockBlocksBtf) {
+			if (context.opts.premiumAdLayoutEnabled && !isBTFDisabledByCreative()) {
```
nandy-andy, line 86: Can we, therefore, remove: https://github.com/Wikia/app/blob/9621f1a95c03f77d3910f6cd054351d0384220c5/extensions/wikia/AdEngine/js/context/uapContext.js#L39?

nandy-andy: And was this the actual bug? Because everything else looks cosmetic + unit tests :+1: :)

nandy-andy (approved):

extensions/wikia/AdEngine/js/provider/btfBlocker.js
kvas-damian: I believe this change breaks kill code in roadblock. By design it should block all slots except `TOP_RIGHT_BOXAD`. In this case all premium slots will be unblocked.

kvas-damian: Ups, I was wrong :)

PR #14049, opened 2017-10-23
PLATFORM-3183: fix DesignSystemApi links to use relative protocol

harnash:
Ensure that Design System links are protocol relative (where it makes sense).

### DIFF
--- a/includes/GlobalFunctions.php
+++ b/includes/GlobalFunctions.php
@@ -4084,16 +4084,3 @@ function wfRandomString( $length = 32 ) {
 	}
 	return substr( $str, 0, $length );
 }
-
-/**
- * @param string $url the url to convert to protocol relative
- * @return string
- */
-function wfProtocolUrlToRelative( $url ) {
-	$pos = stripos( $url, "://" );
-	if ( $pos > 0 ) {
-		$url = substr_replace( $url, "", 0, $pos+1 );
-	}
-
-	return $url;
-}
--- a/includes/GlobalFunctions.php
+++ b/includes/GlobalFunctions.php
@@ -4084,3 +4084,16 @@ function wfRandomString( $length = 32 ) {
 	}
 	return substr( $str, 0, $length );
 }
+
+/**
+ * @param string $url the url to convert to protocol relative
+ * @return string
+ */
+function wfProtocolUrlToRelative( $url ) {
+	$pos = stripos( $url, "://" );
+	if ( $pos > 0 ) {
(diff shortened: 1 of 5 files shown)


includes/GlobalFunctions.php
```diff
@@ -4084,3 +4084,16 @@ function wfRandomString( $length = 32 ) {
 	}
 	return substr( $str, 0, $length );
 }
+
+/**
+ * @param string $url the url to convert to protocol relative
+ * @return string
+ */
+function wfProtocolUrlToRelative( $url ) {
+	$pos = stripos( $url, "://" );
```
Grunny: Minor: Should use single quotes per conventions.

includes/GlobalFunctions.php
```diff
@@ -4084,3 +4084,16 @@ function wfRandomString( $length = 32 ) {
...
+	if ( $pos > 0 ) {
+		$url = substr_replace( $url, "", 0, $pos+1 );
```
Grunny: Minor: Should use single quotes per conventions.

Grunny (changes requested):
Minor coding conventions comments, otherwise LGTM.

includes/GlobalFunctions.php
```diff
@@ -4084,3 +4084,16 @@ function wfRandomString( $length = 32 ) {
 	}
 	return substr( $str, 0, $length );
 }
+
+/**
+ * @param string $url the url to convert to protocol relative
+ * @return string
+ */
+function wfProtocolUrlToRelative( $url ) {
```
Grunny: Also, this method should probably go to `includes/wikia/GlobalFunctions.php` since one of ours rather than Core MediaWiki's.

includes/GlobalFunctions.php
themech: why do we use case-insensitive search if we're just looking for `://`?

includes/GlobalFunctions.php
harnash: fix: 6d46c8630b

includes/GlobalFunctions.php
harnash: fix: 6d46c8630b

harnash: fix: 6d46c8630b4

includes/GlobalFunctions.php
harnash: fix: 6d46c8630b4

Grunny (approved):

PR #14069, opened 2017-10-24
ADEN-6063 Fix prerolls on tablets

fraszczakszymon:
It turns out we have to play preroll on `beforePlay` event (as [documentation](https://developer.jwplayer.com/jw-player/docs/javascript-api-reference/#jwplayerplayadtag) says 🙄 ). Sometimes it doesn't work when it's fired on manually triggered event (i.e. `videoStart`).

Caveat: `beforePlay` is fired on each play click (even if video is going to be resumed), so we need to check whether preroll position has been reached before and reset state after video complete.


latata (approved):

kvas-damian (approved):
Please resolve conflicts

kvas-damian (approved):

PR #14135, opened 2017-11-03
XW-4222 | fix comscore tracking

latata:
https://wikia-inc.atlassian.net/browse/XW-4222
@Wikia/x-wing


extensions/wikia/ArticleVideo/node_modules/jwplayer-fandom/src/sound.js
```diff
@@ -0,0 +1,8 @@
+function wikiaJWPlayerRelatedVideoSound(playerInstance) {
```
latata, line 1: This file should be committed here: https://github.com/Wikia/app/pull/14128

hakubo (approved):

PR #14142, opened 2017-11-03
StaffLogger::log - make sure int value is inserted into database

macbre:
https://wikia-inc.atlassian.net/browse/SUS-3212

Avoid the following:

```
A database query syntax error has occurred. This may indicate a bug in the software. The last attempted database query was:

    (SQL query hidden)

from within function "StaffLogger::log". Database returned error "1366: Incorrect integer value: '' for column 'slog_userdst' at row 1 (geo-db-archive-master.query.consul)".
```

### DIFF
--- a/extensions/wikia/StaffLog/StaffLog.events.php
+++ b/extensions/wikia/StaffLog/StaffLog.events.php
@@ -1,6 +1,6 @@
 <?php
 class StaffLogger {
-	static public function log($type, $action, $userId, $userName, $userdstId, $userNamedst, $comment = "") {
+	static public function log(string $type, string $action, int $userId, string $userName, int $userdstId, string $userNamedst, $comment = "") {
 		global $wgSitename,$wgCityId,$wgExternalDatawareDB;
 		$dbw =  wfGetDB( DB_MASTER, array(), $wgExternalDatawareDB );
 		$now = wfTimestampNow();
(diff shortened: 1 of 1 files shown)


extensions/wikia/StaffLog/StaffLog.events.php
```diff
@@ -1,6 +1,6 @@
 <?php
 class StaffLogger {
-	static public function log($type, $action, $userId, $userName, $userdstId, $userNamedst, $comment = "") {
+	static public function log(string $type, string $action, int $userId, string $userName, int $userdstId, string $userNamedst, $comment = "") {
```
themech, line 3: 👍

themech (approved):

extensions/wikia/StaffLog/StaffLog.events.php
macbre: MySQL 5.7 stopped casting strings to int. Let's move it to PHP then :)

Grunny (approved):

macbre:
Backported to a production branch in de60f35

macbre:
https://www.percona.com/blog/2016/10/18/upgrading-to-mysql-5-7-beware-of-the-new-strict-mode/ says:

> A value can be invalid for several reasons. For example, it might have the wrong data type for the column, or it might be out of range.

PR #14150, opened 2017-11-06
SUS-3139: Avoid fatal error in DPL extension

mszabo-wikia:
do not try to call nonexistent method
https://wikia-inc.atlassian.net/browse/SUS-3139


shaedrich:
Do both `creation_rev.rev_user` table and column still exist? I'm asking because I'm getting error message "Unknown column 'creation_rev.rev_user' in 'where clause' (geo-db-a-slave.query.consul)" when trying to use it in combination with `category` and `modifiedby` param.

PR #14159, opened 2017-11-07
SUS-2977 | remove wgSharedDB checks

macbre:
https://wikia-inc.atlassian.net/browse/SUS-2977

* remove `wgSharedDB` checks, it now default to null
* `wgDontWantShared` removed as well - there was no default value set nor in the code nor in config, WikiFactory has this variable defined, but it's not set on any wiki (variable has been removed from a database)
* CreateNewWikis' `CreateTables` class no longer needs to "mock" `wgSharedDB` to force creation of previously shared table (`user`) on a local wiki database (we're considering removing per-wiki `user` tables)

### DIFF
--- a/extensions/wikia/CreateNewWiki/tasks/CreateTables.php
+++ b/extensions/wikia/CreateNewWiki/tasks/CreateTables.php
@@ -52,12 +52,6 @@ class CreateTables extends Task {
 	}
 
 	public function run() {
-		global $wgSharedDB;
-		$tmpSharedDB = $wgSharedDB;
-		//This is needed because otherwise the underlying tableName() function treats the "user" table as a shared table
-		//and adds a DB prefix which causes the table to be created in a wikicities_x DB instead of the newly created one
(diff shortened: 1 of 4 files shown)


ghost (approved):

extensions/wikia/CreateNewWiki/tasks/CreateTables.php
```diff
@@ -52,12 +52,6 @@ public function prepare() {
 	}
 
 	public function run() {
-		global $wgSharedDB;
-		$tmpSharedDB = $wgSharedDB;
-		//This is needed because otherwise the underlying tableName() function treats the "user" table as a shared table
-		//and adds a DB prefix which causes the table to be created in a wikicities_x DB instead of the newly created one
-		$wgSharedDB = $this->taskContext->getDBname();
-
```
macbre, line 60:
```
> echo wfGetDB(DB_SLAVE)->tableName('user')
...
`user`
```

PR #14164, opened 2017-11-07
SUS-2977 | do not call ExternalUser_Wikia::initFromCond

macbre:
Internally it queries `wikicities.user`, it's already performed by `User` class.

This code is no longer called after https://github.com/Wikia/config/pull/2314 (`$wgExternalAuthType` is now set to null).

**Save ~5.8 mm queries a day to shared DB:**

```sql
SELECT /* ExternalUser_Wikia::initFromCond Redride' s OG - b1fcbaca-e625-4c41-b382-a32235cb81c4 */  *  FROM `user`  WHERE user_id = '1121346'  LIMIT 1  
```

### DIFF
--- a/includes/wikia/Wikia.php
+++ b/includes/wikia/Wikia.php
@@ -35,10 +35,6 @@ $wgHooks['BeforeInitialize']         [] = "Wikia::onBeforeInitializeMemcachePurg
 $wgHooks['SkinTemplateOutputPageBeforeExec'][] = "Wikia::onSkinTemplateOutputPageBeforeExec";
 $wgHooks['UploadVerifyFile']         [] = 'Wikia::onUploadVerifyFile';
 
-# User hooks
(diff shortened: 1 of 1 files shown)


mszabo-wikia (approved):

includes/wikia/Wikia.php
```diff
@@ -35,10 +35,6 @@
 $wgHooks['SkinTemplateOutputPageBeforeExec'][] = "Wikia::onSkinTemplateOutputPageBeforeExec";
 $wgHooks['UploadVerifyFile']         [] = 'Wikia::onUploadVerifyFile';
 
-# User hooks
-$wgHooks['UserNameLoadFromId']       [] = "Wikia::onUserNameLoadFromId";
-$wgHooks['UserLoadFromDatabase']     [] = "Wikia::onUserLoadFromDatabase";
```
macbre, line 40: I kept the invocation of `UserLoadFromDatabase` and `UserNameLoadFromId` hooks - they come from MediaWiki upstream.

PR #14168, opened 2017-11-08
XW-4192 | subtitles/captions in featured videos

latata:
@Wikia/x-wing


extensions/wikia/ArticleVideo/scripts/featured-video.jwplayer.instant.js
```diff
@@ -80,9 +84,13 @@ require([
...
 			},
 			autoplay: willAutoplay,
+			captions: {
+				enabled: featuredVideoCookieService.areCaptionsEnabled()
+			},
 			settings: {
 				showAutoplayToggle: true,
-				showQuality: true
+				showQuality: true,
+				showCaptionsToggle: true
```
hakubo, line 93: it won't be a toogle soon, should we rename it already?

extensions/wikia/ArticleVideo/scripts/featured-video.cookies.js
```diff
@@ -0,0 +1,36 @@
...
+		return function () {
+			return cookies.get(cookieName) !== '0';
+		};
+	}
+
+	return {
+		isAutoplayEnabled: isCookieEnabled(autoplayCookieName),
+		areCaptionsEnabled: isCookieEnabled(captionsCookieName),
+		toggleAutoplay: toggleCookie(autoplayCookieName),
+		toggleCaptions: toggleCookie(captionsCookieName)
```
hakubo, line 34: it supports only on/off, should it support, what language was chosen?

latata (commented):
LGTM

extensions/wikia/ArticleVideo/scripts/featured-video.jwplayer.instant.js
rybmat: I would rename it when we'll be doing the other solution

rybmat (approved):

extensions/wikia/ArticleVideo/scripts/featured-video.cookies.js
xkxd: I thought about taking that into account, but I'd agree with @rybmat to not include this here but in specific feature / PR

PR #14235, opened 2017-11-16
SUS-3261 DPL syntax error

mszabo-wikia:
* use `addIdentifierQuotes` to produce backticks
* add comma

https://wikia-inc.atlassian.net/browse/SUS-3261


shaedrich:
I'm getting the following exeption:
You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near "```creation_rev' `page` INNER JOIN```".

PR #14272, opened 2017-11-22
DE-3223 Added dimension data import api

jcellary:
https://wikia-inc.atlassian.net/browse/DE-3223

Working on new data import mechanism for wiki dimension data. First step is to create MW Api that we can use from qubole.

This is a WIP which will be extended with remaining dimension data APIs

### DIFF
--- a/includes/wikia/api/DWDimensionApiController.class.php
+++ b/includes/wikia/api/DWDimensionApiController.class.php
@@ -1,42 +1,42 @@
 <?php
 
 class DWDimensionApiController extends WikiaApiController {
-    const LIMIT = 100;
-    const AFTER_DOMAIN = null;
+	const LIMIT = 100;
+	const AFTER_DOMAIN = null;
 
-    private function getSharedDbSlave() {
-        global $wgExternalSharedDB;
-        return wfGetDB( DB_SLAVE, array(), $wgExternalSharedDB );
-    }
+	private function getSharedDbSlave() {
+		global $wgExternalSharedDB;
+		return wfGetDB( DB_SLAVE, array(), $wgExternalSharedDB );
+	}
 
 	public function getWikiDomains()
-    {
-        $limit = $this->getRequest()->getVal( 'limit', static::LIMIT );
-        $afterDomain = $this->getRequest()->getVal( 'after_domain', static::AFTER_DOMAIN );
+	{
+		$limit = $this->getRequest()->getVal( 'limit', static::LIMIT );
+		$afterDomain = $this->getRequest()->getVal( 'after_domain', static::AFTER_DOMAIN );
 
-        $db = $this->getSharedDbSlave();
-        $dbResult = $db->select(
-            ["city_domains"],
-            ["city_id", "city_domain"],
-            isset( $afterDomain ) ? [ "city_domain > ".$afterDomain ] : [],
-            __METHOD__,
-            [
-                "ORDER BY" => "city_domain",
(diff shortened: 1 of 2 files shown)


includes/wikia/api/DWDimensionApiController.class.php
```diff
@@ -0,0 +1,41 @@
...
+
+	public function getWikiDomains() {
+		$limit = $this->getRequest()->getVal( 'limit', static::LIMIT );
+		$afterDomain = $this->getRequest()->getVal( 'after_domain', static::AFTER_DOMAIN );
+
+		$db = $this->getSharedDbSlave();
+		$dbResult = $db->select(
+			[ 'city_domains' ],
+			[ 'city_id', 'city_domain' ],
+			isset( $afterDomain ) ? [ 'city_domain > '.$afterDomain ] : [ ],
```
jcellary: @macbre is this ok or do I need to add some explicit escaping to avoid sql injection? Or is it handled by getRequest()->getVal?

macbre: Nope, you need to call `$db->strencode`

includes/wikia/api/DWDimensionApiController.class.php
```diff
@@ -0,0 +1,41 @@
...
+	const AFTER_DOMAIN = null;
+
+	private function getSharedDbSlave() {
+		global $wgExternalSharedDB;
+		return wfGetDB( DB_SLAVE, array(), $wgExternalSharedDB );
+	}
+
+	public function getWikiDomains() {
+		$limit = $this->getRequest()->getVal( 'limit', static::LIMIT );
+		$afterDomain = $this->getRequest()->getVal( 'after_domain', static::AFTER_DOMAIN );
```
macbre: `after_domain`?

How will this API be called?

jcellary: first call you don't specify anything and then you take the last value of domain you received and set it in the after_domain params - you'll get everything than comes after.

I was researching how to best approach pagination and I saw limit supports setting two parameters, but read that for large data sets a where clause is more efficient as it doesn't need to get all the entries and then filter. Also I saw this table has an index on domain column and city_id unfortunatelly is not unique, so cannot be used.

Does that make sense or do you have a better idea?

includes/wikia/api/DWDimensionApiController.class.php
jcellary: Ok, added it

includes/wikia/api/DWDimensionApiController.class.php
```diff
@@ -0,0 +1,42 @@
...
+	public function getWikiDomains() {
+		$db = $this->getSharedDbSlave();
+
+		$limit = $db->strencode( $this->getRequest()->getVal( 'limit', static::LIMIT ) );
+		$afterDomain = $db->strencode( $this->getRequest()->getVal( 'after_domain', static::AFTER_DOMAIN ) );
+
+		$dbResult = $db->select(
+			[ 'city_domains' ],
+			[ 'city_id', 'city_domain' ],
+			isset( $afterDomain ) ? [ 'city_domain > "'.$afterDomain.'"' ] : [ ],
```
macbre: So that we do not forget - properly encode `$afterDomain` value :)

macbre (changes requested):
Looks good, one comment left regarding SQL injection

macbre (approved):

PR #14279, opened 2017-11-23
SUS-805: Use user ID in ipblocks when possible

mszabo-wikia:
* `ipb_by` is always a registered user ID - `ipb_by_text` is redundant and can be dropped
* query based on user ID (`ipb_user`) when checking blocked status of registered user

https://wikia-inc.atlassian.net/browse/SUS-805


includes/Block.php
```diff
@@ -223,8 +229,8 @@ protected function newLoad( $vagueTarget = null ) {
 			list( $target, $type ) = self::parseTarget( $vagueTarget );
 			switch( $type ) {
 				case self::TYPE_USER:
-					# Slightly wierd, but who are we to argue?
```
macbre, line 226: :smile:

includes/User.php
```diff
@@ -1350,7 +1350,7 @@ private function getBlockedStatus( $bFromSlave = true, $shouldLogBlockInStats =
 		}
 
 		# User/IP blocking
-		$block = Block::newFromTarget( $this->getName(), $ip, !$bFromSlave );
+		$block = Block::newFromTarget( $this, $ip, !$bFromSlave );
```
macbre, line 1353: :+1:

tests/unit/includes/BlockTest.php
```diff
@@ -0,0 +1,100 @@
...
+	const NOT_BLOCKED_ANON = '1.1.1.1';
+
+	private $wgMemc;
+
+	protected function setUp() {
+		parent::setUp();
+
+		$this->wgMemc = $GLOBALS['wgMemc'];
+		$GLOBALS['wgMemc'] = new EmptyBagOStuff();
+	}
```
macbre, line 34: We have `WikiaDatabaseTest::disableMemCache` helper for, well, disabling memcache...

macbre (approved):
Let's run Selenium tests suite for handling blocked users before merging this.

tests/unit/includes/BlockTest.php
mszabo-wikia: That's only available in `WikiaBaseTest`, and trying to make `WikiaDatabaseTest` subclass `WikiaBaseTest` breaks `WikiaMockProxy` and associated hacks 😞

PR #14281, opened 2017-11-24
DE-3223 Added get wikis api

jcellary:
Followup of https://github.com/Wikia/app/pull/14272

Added api for importing list of all wikis and tags for a wiki

### DIFF
--- a/extensions/wikia/WikiFactory/WikiFactory.php
+++ b/extensions/wikia/WikiFactory/WikiFactory.php
@@ -3435,15 +3435,16 @@ class WikiFactory {
 	/**
 	 * Returns value of a given variable for all wikis
 	 *
-	 * @param $variableId String Id of variable to get
+	 * @param $variableName String Name of variable to get
 	 * @param $limit String Limit of rows
 	 * @param $afterWikiId String get value for wikis with id greater than this value. Optional, used for pagination.
 	 * @return array list of wiki ids and variable values
 	 */
-	static public function getVariableForAllWikis( $variableId, $limit, $afterWikiId = null ) {
+	static public function getVariableForAllWikis( $variableName, $limit, $afterWikiId = null ) {
 
 		$db = static::db( DB_SLAVE );
 
+		$variableId = static::getVarIdByName( $variableName );
 		$where = [ 'cv_variable_id = '.$variableId ];
 		if ( isset( $afterWikiId ) ) {
 			array_push( $where, 'cv_city_id > '.$afterWikiId );
--- a/extensions/wikia/WikiFactory/WikiFactory.php
+++ b/extensions/wikia/WikiFactory/WikiFactory.php
@@ -3436,11 +3436,11 @@ class WikiFactory {
 	 * Returns value of a given variable for all wikis
 	 *
 	 * @param $variableId String Id of variable to get
-	 * @param $afterWikiId String get value for wikis with id greater than this value. Used for pagination.
 	 * @param $limit String Limit of rows
+	 * @param $afterWikiId String get value for wikis with id greater than this value. Optional, used for pagination.
 	 * @return array list of wiki ids and variable values
 	 */
-	static public function getVariableForAllWikis( $variableId, $afterWikiId, $limit ) {
+	static public function getVariableForAllWikis( $variableId, $limit, $afterWikiId = null ) {
 
 		$db = static::db( DB_SLAVE );
 
--- a/extensions/wikia/WikiFactory/WikiFactory.php
+++ b/extensions/wikia/WikiFactory/WikiFactory.php
@@ -3436,8 +3436,8 @@ class WikiFactory {
 	 * Returns value of a given variable for all wikis
 	 *
 	 * @param $variableName String Name of variable to get
-	 * @param $limit String Limit of rows
-	 * @param $afterWikiId String get value for wikis with id greater than this value. Optional, used for pagination.
+	 * @param $limit Int Limit of rows
+	 * @param $afterWikiId Int get value for wikis with id greater than this value. Optional, used for pagination.
 	 * @return array list of wiki ids and variable values
 	 */
 	static public function getVariableForAllWikis( $variableName, $limit, $afterWikiId = null ) {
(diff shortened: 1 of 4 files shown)


includes/wikia/api/DWDimensionApiControllerSQL.class.php
```diff
@@ -0,0 +1,58 @@
+<?php
+
+class DWDimensionApiControllerSQL {
```
jcellary, line 3: @macbre Is this a reasonable way to include SQL or is there some other preferred way?

includes/wikia/api/DWDimensionApiControllerSQL.class.php
```diff
@@ -0,0 +1,58 @@
...
+
+	const DIMENSION_WIKIS_QUERY = '
+      SELECT 
+           city.city_id            AS wiki_id,
+           city.city_dbname        AS dbname,
+           city.city_sitename      AS sitename,
+           REPLACE(REPLACE(city.city_url, \'http://\', \'\'), \'/\', \'\') AS url,
+           REPLACE(REPLACE(city.city_url, \'http://\', \'\'), \'/\', \'\') AS domain,
```
macbre: Can't we perform this logic in PHP?

includes/wikia/api/DWDimensionApiControllerSQL.class.php
```diff
@@ -0,0 +1,58 @@
...
+                ELSE 9 END            AS vertical_id,
+           CASE WHEN vert.vertical_name IN (\'Music\', \'Comics\', \'TV\', \'Movies\', \'Books\')               THEN \'Entertainment\'
+                WHEN vert.vertical_name = \'Video Games\'                                               THEN \'Gaming\'
+                WHEN city.city_id IN (SELECT city_id FROM city_cat_mapping WHERE cat_id = 4)             THEN \'Wikia\'
+                WHEN vert.vertical_name = \'Lifestyle\'                                                 THEN \'Lifestyle\'
+                WHEN cat.cat_name IN (\'Entertainment\', \'Music\') THEN \'Entertainment\'
+                WHEN cat.cat_name = \'Gaming\'                    THEN \'Gaming\'
+                WHEN cat.cat_name = \'Wikia\'                     THEN \'Wikia\'
+                WHEN cat.cat_name = \'Lifestyle\'                 THEN \'Lifestyle\'
+                ELSE \'Lifestyle\' END        AS vertical_name,
```
macbre: `WikiFactory` class has a few, well cached methods for getting these data, such as category or vertical names.

jcellary:
@macbre I replaced use cases with WikiFactory. Will still need to confirm with analytics if the default vertical mapping is ok. I also added a new api for dart and removed one that won't be used. Please take a look if it's good and if yes we could merge this part.

includes/wikia/api/DWDimensionApiController.class.php
```diff
@@ -52,11 +52,16 @@ public function getWikiDartTags() {
...
+			#extract from list like "s:199:\"sex=m;sex=f;age=under18;age=13-17;age=18-24;age=25-34;age=18-34;\";"
+			preg_match_all("/([^;= ]+)=([^;= ]+)/", unserialize( $row->cv_value ), $r);
+
+			for ($i = 0; $i < count( $r[1] ); $i++) {
+				$result[] = [
+					'wiki_id' => $row->cv_city_id,
+					'tag' => $r[ 1 ][ $i ],
+					'value' => $r[ 2 ][ $i ]
+				];
+			}
```
macbre: Don't we want to move WikiFactory to a service? This adds a dependency on a database. Let's use `WikiFactory` class helpers to get a variable value. It's pretty well cached.

jcellary: WikiFactory in it's current form doesn't have a method to fetch values for all wikis, only a single one (at least from what I have seen). I was thinking about adding one, but wasn't sure if it's worth it. Do you think I should do that? Or is there some existing mechanism that I missed?

jcellary: Ok moved to WikiFactory

extensions/wikia/WikiFactory/WikiFactory.php
```diff
@@ -3432,4 +3432,40 @@ static private function parseValue( $value, $type ) {
 		return var_export( $value, true );
 	}
 
+	/**
+	 * Returns value of a given variable for all wikis
+	 *
+	 * @param $variableId String Id of variable to get
```
macbre: That's actually `int`, right?

extensions/wikia/WikiFactory/WikiFactory.php
```diff
@@ -3432,4 +3432,40 @@ static private function parseValue( $value, $type ) {
...
+	 * @param $limit String Limit of rows
```
macbre: The same here

includes/wikia/api/DWDimensionApiController.class.php
```diff
@@ -2,36 +2,101 @@
...
-			isset( $afterDomain ) ? [ 'city_domain > "'.$afterDomain.'"' ] : [ ],
-			__METHOD__,
-			[
-				'ORDER BY' => 'city_domain',
-				'LIMIT' => $limit
-			]
+		$limit = min($db->strencode( $this->getRequest()->getVal( 'wiki_limit', static::LIMIT ) ), static::LIMIT_MAX);
+		$afterWikiId = $db->strencode( $this->getRequest()->getVal( 'after_wiki_id', static::WIKIS_AFTER_WIKI_ID ) );
+
+		$variables = WikiFactory::getVariableForAllWikis( static::DART_TAG_VARIABLE_ID, $limit, $afterWikiId );
```
macbre: Let's use `WikiFactory::getVarIdByName` instead of hard-coding variable ID.

extensions/wikia/WikiFactory/WikiFactory.php
jcellary: true

extensions/wikia/WikiFactory/WikiFactory.php
jcellary: Not sure, because we get this from query param, so I assume everything is a string (although obviously it contains a number). Should I do parsing in such case or just pass it along?

includes/wikia/api/DWDimensionApiController.class.php
jcellary: done

jcellary:
Merging per offline conversation with Macbre

PR #14288, opened 2017-11-27
XW-4281 | Reduce amount of tracking events sent to GA

rybmat:
https://wikia-inc.atlassian.net/browse/XW-4281

* don't track connection type
* don't track embeddable-discussions impressions
* don't track premium rail impressions
* don't send events to special and loggedin_users GA accounts

@Wikia/x-wing


extensions/wikia/Track/Track.php
```diff
@@ -99,19 +99,13 @@ private static function getGAURL( $type, $category, $action, $label, $value, $ti
 	}
 
 	private static function getGATrackingIds() {
-		global $wgDevelEnvironment, $wgStagingEnvironment, $wgIsGASpecialWiki, $wgUser;
+		global $wgDevelEnvironment, $wgStagingEnvironment;
 
 		$tids = [ ];
 		// 10% sampling for general account
 		if ( mt_rand( 0, 9 ) === 0 ) {
```
hakubo: hmm, should we remove this? 
does that mean we don't track at 100% on general account?

extensions/wikia/AnalyticsEngine/js/universal_analytics.js
```diff
@@ -121,66 +121,6 @@
...
-					'name': 'special',
-					'sampleRate': 100,
-					'allowLinker': true,
-					'userId': window.wgGAUserIdHash
-				}
-			);
-		}
-
-		// Enable Demographics and Interests Reports
-		window.ga('special.require', 'displayfeatures');
```
hakubo, line 151: should we enable that on general account?

extensions/wikia/Track/Track.php
rybmat: @hakubo this is tracking on backend side, on frontend it is 100%, I'll change it to 100%

extensions/wikia/AnalyticsEngine/js/universal_analytics.js
rybmat: @hakubo makes sense to me (since we have it enabled already for ads and ve accounts), thanks for noticing!

rybmat: actually it is enabled via Admin UI, so I won't change it here.

extensions/wikia/AnalyticsEngine/js/universal_analytics.js
```diff
@@ -230,20 +170,6 @@
...
-
-				// If user is logged in we send to Logged-In Users Account
-				if (window.wgGAUserIdHash) {
-					spec = args[i].slice();
-					spec[0] = 'loggedin_users.' + spec[0];
-					window.ga.apply(window, spec);
-				}
-
 				// If category is editor-ve, track for VE account
 				if (args[i][1] && args[i][1] === 'editor-ve') {
```
hakubo, line 174: another interesting account, would be good to know if we should remove as well

extensions/wikia/Track/Track.php
```diff
@@ -99,19 +99,10 @@ private static function getGAURL( $type, $category, $action, $label, $value, $ti
...
-		}
-		if ( $wgIsGASpecialWiki ) {
-			$tids[] = $wgDevelEnvironment || $wgStagingEnvironment ? 'UA-32132943-2' : 'UA-32132943-1';
-		}
-		if ( !$wgUser->isAnon() ) {
-			$tids[] = $wgDevelEnvironment || $wgStagingEnvironment ? 'UA-32132943-8' : 'UA-32132943-7';
-		}
+		$tids[] = $wgDevelEnvironment || $wgStagingEnvironment ? 'UA-32129070-2' : 'UA-32129070-1';
 
 		return $tids;
```
hakubo: return  [
$wgDevelEnvironment || $wgStagingEnvironment ? 'UA-32129070-2' : 'UA-32129070-1'
]

hakubo (approved):

latata (approved):

PR #14297, opened 2017-11-27
SRE-36 → sanitize usage of $wgDBserver

ghost:
At this point `$wgDBserver` is not yet set properly which renders the fallback
futile.  The variable is updated later.  It depends on the mode (rw/ro) and
the cluster (a, b, …) therefore cannot be preset in the config.

https://wikia-inc.atlassian.net/browse/SRE-36

As a side note, the current config value is an IP of a machine that no longer exists.

### DIFF
--- a/extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
+++ b/extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
@@ -177,8 +177,7 @@ class WikiFactoryLoader {
 	 * getDB
 	 *
 	 * Method for getting database handler. It checks if $wgDBservers is
-	 * available, if yes it will take one of active slaves. If not it fallbacks
-	 * to $wgDBserver
+	 * available, if yes it will take one of active slaves.
 	 *
 	 * @author Krzysztof Krzyżaniak <eloy@wikia-inc.com>
 	 * @access public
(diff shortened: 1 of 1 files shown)


extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
```diff
@@ -202,15 +200,6 @@ public function getDB( $type = DB_SLAVE ) {
...
 
-		/**
-		 * if something goes wrong just fallback to $wgDBserver
-		 */
-		if( !$this->mDBhandler || !$this->mDBhandler->isOpen() ) {
-			error_log( "WikiFactoryLoader[{$this->mCityID}]: fallback to {$wgDBserver}" );
-			$this->mDBhandler = new DatabaseMysqli( $wgDBserver, $wgDBuser, $wgDBpassword, $this->mDBname );
-			$this->debug( "fallback to wgDBserver {$wgDBserver}" );
-		}
-
```
macbre: Can we leave an `assert` here and make sure this edgecase does not happen?

ghost: Good evening, Sir. :-)

If it happened, we would end up with a fatal error, as the initial value was a not-responding IP of a machine that has been already decommissioned. Although assert is not a bad idea.

macbre (approved):

PR #14302, opened 2017-11-28
SUS-3079

mszabo-wikia:
https://wikia-inc.atlassian.net/browse/SUS-3079

### DIFF
--- a/extensions/DynamicPageList/DPLMain.php
+++ b/extensions/DynamicPageList/DPLMain.php
@@ -2165,7 +2165,7 @@ class DPLMain {
 
 			if ( $bAddContribution ) {
 				$sSqlRCTable = $sRCTable . ' AS rc, ';
-				$sSqlSelPage .= ', SUM( ABS( rc.rc_new_len - rc.rc_old_len ) ) AS contribution, rc.rc_user_text AS contributor, rc.rc_user AS contributor_id'; // SUS-812
+				$sSqlSelPage .= ', SUM( ABS( rc.rc_new_len - rc.rc_old_len ) ) AS contribution, rc.rc_ip_bin AS contributor, rc.rc_user AS contributor_id'; // SUS-812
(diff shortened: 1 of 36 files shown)


maintenance/tables.sql
```diff
@@ -967,7 +967,8 @@ CREATE TABLE /*_*/recentchanges (
 
   -- As in revision
   rc_user int unsigned NOT NULL default 0,
-  rc_user_text varchar(255) binary NOT NULL,
+  -- SUS-3079: Placeholder empty value until migration
+  rc_user_text VARCHAR(255) BINARY NOT NULL DEFAULT '',
```
macbre, line 971: Do we need to update the schema site-wide before we move to the next step?

mszabo-wikia: It'll be handled in `update.php`. Post-merge, it will migrate data to `rc_ip_bin` column, drop indexes, and add this default value to already created `rc_user_text` columns. Then we can remove explicit value setting for `rc_user_text` in inserts, and drop the column entirely.

extensions/DynamicPageList/DPLMain.php
```diff
@@ -2165,7 +2165,7 @@ public static function dynamicPageList( $input, $params, $parser, &$bReset, $cal
 
 			if ( $bAddContribution ) {
 				$sSqlRCTable = $sRCTable . ' AS rc, ';
-				$sSqlSelPage .= ', SUM( ABS( rc.rc_new_len - rc.rc_old_len ) ) AS contribution, rc.rc_user_text AS contributor, rc.rc_user AS contributor_id'; // SUS-812
+				$sSqlSelPage .= ', SUM( ABS( rc.rc_new_len - rc.rc_old_len ) ) AS contribution, rc.rc_ip_bin AS contributor, rc.rc_user AS contributor_id'; // SUS-812
```
macbre, line 2168: Should we decode the IP from the binary format?

PR #14305, opened 2017-11-28
ADEN-6235 Adds stickiness to BFAA on Oasis

leshniak:


fraszczakszymon:
Shouldn't we use the template from ad-products?

fraszczakszymon (changes requested):
Please use BFAA from ad-products repository

PR #14314, opened 2017-11-29
SUS-3120 Rename local user related pages

mszabo-wikia:
On user rename completion, trigger background tasks to move user pages and similar content to the new user name.

https://wikia-inc.atlassian.net/browse/SUS-3120


includes/wikia/tasks/Tasks/RenameUserPagesTask.class.php
```diff
@@ -0,0 +1,64 @@
...
+	 * @return int[] array of wiki IDs
+	 */
+	public static function getTargetCommunities( string $oldUserName ) {
+		global $wgExternalDatawareDB;
+
+		$dbr = wfGetDB( DB_SLAVE, [], $wgExternalDatawareDB );
+
+		return $dbr->selectFieldValues( 'pages', 'page_wikia_id', [
+			'page_namespace' => static::NAMESPACES,
+			'page_title' => $oldUserName
```
macbre, line 28: Should we change it to `LIKE` condition and move user page sub-pages as well?

macbre: Oh, wait, it's covered in `renameLocalPages`

mszabo-wikia: There shouldn't be a need to do that when checking for where to run task:
* if Wall root page does not exist, Wall is broken
* if blog listing page does not exist, Blogs are also broken
* if user does not have user page, user talk page, Message Wall, or blog, but somehow has user or user talk subpages... - do we want to check for such situation?

includes/wikia/tests/core/WikiaBaseTest.class.php
```diff
@@ -453,26 +427,11 @@ public function __retrieveMessageMock( $key ) {
...
-
-		return $wgMemcMock;
-	}
-
 	/**
 	 * Mocks global $wgMemc->get() so it always returns null
 	 */
 	protected function disableMemCache() {
-		$this->mockGlobalVariable( 'wgMemc', $this->getMemCacheMock() );
+		$this->mockGlobalVariable( 'wgMemc', new EmptyBagOStuff() );
```
macbre, line 434: :+1:

extensions/wikia/UserRenameTool/tests/fixtures/dataware-pages.sql
```diff
@@ -0,0 +1,13 @@
...
+  `page_namespace` int unsigned NOT NULL DEFAULT '0',
+  `page_title` varchar(255) NOT NULL,
+  `page_is_content` tinyint unsigned NOT NULL DEFAULT '0',
+  `page_is_redirect` tinyint unsigned NOT NULL DEFAULT '0',
+  `page_latest` int unsigned NOT NULL DEFAULT '0',
+  `page_last_edited` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  PRIMARY KEY (`page_wikia_id`,`page_id`)
+);
+
+CREATE INDEX `page_title_namespace_latest_idx` ON `pages` (`page_title`,`page_namespace`,`page_latest`);
```
macbre, line 13: :+1:

macbre (approved):

PR #14316, opened 2017-11-29
SUS-3443 | PhalanxStatsPager - resolve user name using user ID

macbre:
`ps_blocked_user` will store IP addresses for anons.

https://wikia-inc.atlassian.net/browse/SUS-3443

This change can go live before the data migration, as the code will fallback to a full user name kept in `ps_blocked_user` when `ps_blocked_user_id` is zero (or `NULL` as we cast the data to `int` before passing it to `User::getUsername`).

### DIFF
--- a/extensions/wikia/PhalanxII/classes/PhalanxStatsPager.class.php
+++ b/extensions/wikia/PhalanxII/classes/PhalanxStatsPager.class.php
@@ -9,7 +9,7 @@ class PhalanxStatsPager extends PhalanxPager {
 	public function __construct( int $id ) {
 		parent::__construct();
 
-		$this->id = (int) $id;
+		$this->id = $id;
 		$this->mDb = $this->getDatabase( DB_SLAVE );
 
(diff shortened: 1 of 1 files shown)


extensions/wikia/PhalanxII/classes/PhalanxStatsPager.class.php
```diff
@@ -3,7 +3,10 @@ class PhalanxStatsPager extends PhalanxPager {
...
 	public $pInx = 'blockId';
 
-	public function __construct( $id ) {
+	protected $id;
+	public $mDb;
+
+	public function __construct( int $id ) {
 		parent::__construct();
 
 		$this->id = (int) $id;
```
mszabo-wikia: Do we need this cast if we enforce strict types in signature?

mszabo-wikia (changes requested):

mszabo-wikia (approved):

PR #14318, opened 2017-11-29
SUS-3446 Fix QuickTools rollback feature

mszabo-wikia:
Same as https://github.com/Wikia/app/pull/14312

https://wikia-inc.atlassian.net/browse/SUS-3446

Same in English

### DIFF
--- a/extensions/wikia/SyntaxHighlight/tests/SyntaxHighlightTest.php
+++ b/extensions/wikia/SyntaxHighlight/tests/SyntaxHighlightTest.php
@@ -17,6 +17,8 @@ class SyntaxHighlightTest extends WikiaBaseTest {
 		$this->setupFile = __DIR__ . '/../SyntaxHighlight.setup.php';
 		parent::setUp();
 
+		Hooks::register( 'ParserFirstCallInit', 'SyntaxHighlightHooks::onParserFirstCallInit' );
+
 		$this->title = new Title();
 		$this->opts = new ParserOptions();
 		$this->parser = new Parser();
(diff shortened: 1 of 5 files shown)


extensions/wikia/SyntaxHighlight/tests/SyntaxHighlightTest.php
```diff
@@ -17,6 +17,8 @@ protected function setUp() {
 		$this->setupFile = __DIR__ . '/../SyntaxHighlight.setup.php';
 		parent::setUp();
 
+		Hooks::register( 'ParserFirstCallInit', 'SyntaxHighlightHooks::onParserFirstCallInit' );
+
```
macbre, line 21: ?

includes/db/Database.php
```diff
@@ -3606,6 +3606,18 @@ protected function indexNameCallback( $matches ) {
...
+	 *
+	 * @param string $lockName name of lock to poll
+	 * @param string $method name of method calling us
+	 * @return Boolean
+	 * @since 1.20
+	 */
+	public function lockIsFree( $lockName, $method ) {
+		return true;
+	}
+
```
macbre, line 3620: ?

extensions/wikia/SyntaxHighlight/tests/SyntaxHighlightTest.php
mszabo-wikia: We've unset the hooks in other test, so it's better to register it here

includes/db/Database.php
mszabo-wikia: This method was backported from 1.20 but only to `DatabaseMysqlBase` class. It is thus missing from the parent class and causes an error if something tries to call it in SQLite integration tests.

macbre (approved):
Noticed two "strange" changes

PR #14326, opened 2017-12-01
SUS-3459 | Revision - rev_user_text can now be an empty string

macbre:
That means that we should use `user_id` from `rev_user` column to properly resolve a user name.

https://wikia-inc.atlassian.net/browse/SUS-3459

This method is used by `Linker::revUserTools` that is used on Special:NewPages to render contributor link.

### DIFF
--- a/includes/Revision.php
+++ b/includes/Revision.php
@@ -664,7 +664,7 @@ class Revision implements IDBAccessObject {
 	 * @return String
 	 */
 	public function getRawUserText() {
-		if ( $this->mUserText === null ) {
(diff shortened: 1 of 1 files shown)


mszabo-wikia (changes requested):
`if ( empty( $this->mUserText ) )` ?

macbre:
```
=== Found user ===
	ID: 174051
	Username: 0
	Name: 
	Email: 
```

:smile:

mszabo-wikia (approved):

PR #14329, opened 2017-12-01
DE-3199 Added api for users

jcellary:
Added API used by DW in Qubole to import list of users


includes/wikia/api/DWDimensionApiControllerSQL.class.php
```diff
@@ -32,6 +32,28 @@ class DWDimensionApiControllerSQL {
...
+			user_name,
+			user_real_name,
+			user_email_authenticated,
+			user_editcount,
+			STR_TO_DATE(user_registration, \'%Y%m%d%H%i%s\') AS user_registration,
+			CASE WHEN b.ug_user  IS NULL THEN false ELSE true END AS is_bot,
+			CASE WHEN bg.ug_user IS NULL THEN false ELSE true END AS is_bot_global,
+			CASE WHEN IFNULL(up.up_value,0) = 0 THEN false ELSE true END AS user_marketingallowed,
+			CASE WHEN up_skin.up_value IS NULL OR up_skin.up_value = \'\' THEN \'oasis\'    ELSE up_skin.up_value END AS user_skin,
+			CASE WHEN up_rte.up_value = 0 THEN \'source\' ELSE \'richtext\' END AS user_editor
```
macbre: I don't really like this logic embedded in SQL queries.

includes/wikia/api/DWDimensionApiControllerSQL.class.php
```diff
@@ -32,6 +32,28 @@ class DWDimensionApiControllerSQL {
...
+		FROM user u LEFT JOIN user_groups b ON b.ug_user  = u.user_id AND b.ug_group = \'bot\'
+		LEFT JOIN user_groups bg ON bg.ug_user  = u.user_id AND bg.ug_group = \'bot-global\'
+		LEFT JOIN user_properties up ON up.up_user = u.user_id AND up.up_property = \'marketingallowed\'
+		LEFT JOIN user_properties up_skin ON up_skin.up_user = u.user_id AND up_skin.up_property = \'skin\'
+		LEFT JOIN user_properties up_rte ON up_rte.up_user = u.user_id AND up_rte.up_property = \'enablerichtext\'
```
macbre: Groups can be checked using `User` class helpers.

includes/wikia/api/DWDimensionApiController.class.php
```diff
@@ -111,6 +117,43 @@ public function getWikis() {
...
 
+	public function getUsers() {
+		$db = $this->getSharedDbSlave();
+
+		$limit = min( $db->strencode( $this->getRequest()->getVal(
+			'limit', static::LIMIT ) ), static::LIMIT_MAX );
+		$afterUserId = $db->strencode( $this->getRequest()->getVal(
+			'after_user_id', static::DEFAULT_AFTER_USER_ID ) );
+
+		$query = str_replace( '$user_id', $afterUserId, DWDimensionApiControllerSQL::DIMENSION_USERS );
```
macbre, line 136: Query param encoding is missing

macbre (changes requested):
See above

includes/wikia/api/DWDimensionApiController.class.php
jcellary: What do you mean? $db->strencode is called above or is there some other encoding you have in mind? aa or do you mean using getInt instead of getVal?

includes/wikia/api/DWDimensionApiControllerSQL.class.php
jcellary: You're right, unfortunately that would lead to us making a get user groups query for every single row we get back from this query, so making joins is much lighter on the DB/performance. Is there some other reasonable compromise? Maybe passing the specific groups into this templated query from code consts?

includes/wikia/api/DWDimensionApiControllerSQL.class.php
jcellary: Ok, I can move it to code mapping!

includes/wikia/api/DWDimensionApiControllerSQL.class.php
jcellary: Ok I found a nice solution - turns out Permission Service has a method get users in groups and since only a few users are bots we can get a list of those and then simply check if user is in it, without making additional queries

includes/wikia/api/DWDimensionApiControllerSQL.class.php
```diff
@@ -39,17 +39,8 @@ class DWDimensionApiControllerSQL {
...
-			CASE WHEN b.ug_user  IS NULL THEN false ELSE true END AS is_bot,
-			CASE WHEN bg.ug_user IS NULL THEN false ELSE true END AS is_bot_global,
-			CASE WHEN IFNULL(up.up_value,0) = 0 THEN false ELSE true END AS user_marketingallowed,
-			CASE WHEN up_skin.up_value IS NULL OR up_skin.up_value = \'\' THEN \'oasis\'    ELSE up_skin.up_value END AS user_skin,
-			CASE WHEN up_rte.up_value = 0 THEN \'source\' ELSE \'richtext\' END AS user_editor
-		FROM user u LEFT JOIN user_groups b ON b.ug_user  = u.user_id AND b.ug_group = \'bot\'
-		LEFT JOIN user_groups bg ON bg.ug_user  = u.user_id AND bg.ug_group = \'bot-global\'
-		LEFT JOIN user_properties up ON up.up_user = u.user_id AND up.up_property = \'marketingallowed\'
-		LEFT JOIN user_properties up_skin ON up_skin.up_user = u.user_id AND up_skin.up_property = \'skin\'
-		LEFT JOIN user_properties up_rte ON up_rte.up_user = u.user_id AND up_rte.up_property = \'enablerichtext\'
```
macbre: :+1:

macbre (approved):

PR #14332, opened 2017-12-04
SUS-3071 | stop creating local user and user_properties tables for new wikis

macbre:
https://wikia-inc.atlassian.net/browse/SUS-3071

This is a first step. Let's start with new wikis and make sure that this change does not introduce any regressions. Then we can proceed with dropping these tables on existing wikis.


maintenance/tables.sql
```diff
@@ -33,84 +33,6 @@
...
---
--- Some multi-wiki sites may share a single central user table
--- between separate wikis using the $wgSharedDB setting.
---
--- Note that when a external authentication plugin is used,
--- user table entries still need to be created to store
--- preferences and to key tracking information in the other
--- tables.
---
-CREATE TABLE /*_*/user (
```
mszabo-wikia, line 49: Let's move these table definitions to a separate file and make sure they are included by https://github.com/Wikia/app/blob/dev/includes/wikia/tests/core/WikiaDatabaseTest.php initialization

mszabo-wikia (changes requested):

maintenance/tables.sql
macbre: Done in 4748d6d

mszabo-wikia (approved):

PR #14343, opened 2017-12-04
SUS-3469 | let's automate detection of outdated video providers

macbre:
Search Solr for the latest video from each provider and check when was it uploaded.

https://wikia-inc.atlassian.net/browse/SUS-3469


mszabo-wikia (approved):
Let's unify spacing

PR #14344, opened 2017-12-04
DE-3199 Added try catch

jcellary:
Added try catch to the DE api, as some wikis have invalid schemas and we want to ommit them.


mszabo-wikia (commented):
You can use `WikiaDatabaseTest` to write integration test for this database access code if you feel it'd be useful 🙂

jcellary:
That's a good idea, I'll give it a try, however I'd like for this simple fix to go live tomorrow, so that I can use the API in our qubole task - I'll experiment with the integration tests and see if I can create a separate PR for them, does that sound reasonable?

mszabo-wikia:
Sure, as you prefer :) This change looks good 👍

PR #14351, opened 2017-12-05
DWDimensionApiController added try catch for connection building

jcellary:
Added try catch for connecting to particular wiki db. Some entries are invalid in wikicities and we need to skip those rather than fail whole api call

### DIFF
--- a/includes/wikia/api/DWDimensionApiController.class.php
+++ b/includes/wikia/api/DWDimensionApiController.class.php
@@ -233,7 +233,7 @@ class DWDimensionApiController extends WikiaApiController {
 				];
 			}
 			$db->freeResult( $rows );
-		} catch (DBQueryError $e) {
+		} catch ( DBQueryError $e ) {
 			Wikia\Logger\WikiaLogger::instance()->error(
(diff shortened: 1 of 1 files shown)


includes/wikia/api/DWDimensionApiController.class.php
```diff
@@ -336,7 +336,16 @@ private function getWikiConnection( $cluster, $dbname ) {
 		}
 		$connection = $connections[ $cluster ];
 		$dbname = $db->strencode( $dbname );
-		$connection->query( "USE `".$dbname."`", __METHOD__ );
+		try {
+			$connection->query( "USE `".$dbname."`", __METHOD__ );
+		} catch (DBQueryError $e) {
```
KrystianMistrzak: formatting?

KrystianMistrzak (approved):
LGTM

includes/wikia/api/DWDimensionApiController.class.php
KrystianMistrzak: overall looks good to me

KrystianMistrzak (approved):
LGTM

PR #14357, opened 2017-12-06
DO NOT MERGE | XW-4280 | Migration of user attribution to the JW custom params

bkoval:
@Wikia/x-wing 

THIS WILL NOT BE MERGED


maintenance/wikia/jwplayerattribution/migrateJWAttributionToDashboard.php
```diff
@@ -0,0 +1,77 @@
...
+	protected $dryRun = true;
+
+	protected $map = [];
+
+	public function __construct() {
+		parent::__construct();
+		$this->mDescription = "Migrate JW videos attribution from WF to JW Dashboard";
+		$this->addOption( 'dry-run', 'Dry run mode', false, false, 'd' );
+
+		$this->jwApi = new BotrAPI('key', 'secret');
```
bkoval, line 21: This will be swapped for the real key and secret during migration.

maintenance/wikia/jwplayerattribution/migrateJWAttributionToDashboard.php
```diff
@@ -0,0 +1,77 @@
...
+
+	private function doUpdate( $varId, $varName ) {
+		// Get all wikis with var set
+		$wikisWithVar = ( WikiFactory::getListOfWikisWithVar( $varId, 'array', '!=', [] ) );
+
+		$this->output( 'Found ' . count( $wikisWithVar ) . "wikis with " . $varName . " config\n" );
+
+
+		// Iterate over all wikis that have the var set
+		foreach ( $wikisWithVar as $wikiId => $wiki ) {
```
rybmat, line 40: I know this way is easier now since it is copy-paste, but in the future can we just run maintenance script like that: https://wikia-inc.atlassian.net/wiki/spaces/ONE/pages/35724059/Engineering+Onboarding+How+to+run+maintenance+script
which by default iterates on each wiki, so the  only thing that we'd need to do in this script would be checking if wiki has FV variable and migrate if yes

rybmat (approved):

PR #14368, opened 2017-12-07
(PLATFORM-2851) Make Nirvana cache private for logged-in users

Grunny:
This makes the default cache behaviour for Nirvana controllers to use
public cache for anon users and private cache for logged-in users,
rather than relying on varying on cookie. This matches the behaviour
elsewhere in MediaWiki. If the controller doesn't need to be private
for logged-in users, it can be overridden by explicitly setting the
cache as public.

The Cache-Control header sent to clients in this case will remain public
so that browser caching works as expected even for logged-in users.

### DIFF
--- a/includes/wikia/nirvana/WikiaResponse.class.php
+++ b/includes/wikia/nirvana/WikiaResponse.class.php
@@ -302,15 +302,21 @@ class WikiaResponse {
 	 *
 	 * @param string $policy caching policy (either private or public)
 	 */
-	public function setCachePolicy($policy) {
-		if ( in_array( $policy, [
+	public function setCachePolicy( $policy ) {
+		if ( !in_array( $policy, [
 				self::CACHE_PRIVATE,
 				self::CACHE_PUBLIC,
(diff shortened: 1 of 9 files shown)


includes/wikia/nirvana/WikiaResponse.class.php
```diff
@@ -304,7 +303,14 @@ public function setHeader( $name, $value, $replace = true ) {
...
 	public function setCachePolicy($policy) {
-		$this->cachingPolicy = $policy === self::CACHE_PRIVATE ? self::CACHE_PRIVATE : self::CACHE_PUBLIC;
+		if ( in_array( $policy, [
+				self::CACHE_PRIVATE,
+				self::CACHE_PUBLIC,
+				self::CACHE_ANON_PUBLIC_USER_PRIVATE,
+			] )
+		) {
+			$this->cachingPolicy = $policy;
+		}
```
themech: what about error handling in case `$policy` is not valid? I would log an error/exception at least.

themech (changes requested):

themech (approved):

PR #14373, opened 2017-12-08
PLATFORM-3302: allow loading WikiFactory variables for non-public wikis

harnash:
ping: @Wikia/core-platform-team @macbre

### DIFF
--- a/extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
+++ b/extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
@@ -395,7 +395,7 @@ class WikiFactoryLoader {
 		/**
 		 * redirection to another url
 		 */
-		if( $this->mIsWikiaActive == 2 ) {
+		if( $this->mIsWikiaActive == 2 && !$this->mCommandLine ) {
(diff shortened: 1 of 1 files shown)


themech (approved):

extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
```diff
@@ -395,7 +395,7 @@ public function execute() {
 		/**
 		 * redirection to another url
 		 */
-		if( $this->mIsWikiaActive == 2 ) {
+		if( $this->mIsWikiaActive == 2 && !$this->mCommandLine ) {
```
themech: I would add a small comment explaining why we need it in command line mode. it may save some WTFs in the future.

themech (changes requested):

extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
harnash: I'll add but this is happening in a couple of places in this file so somebody was taking that into account someday... but not this time.

harnash: added in: f46127d8cc0

PR #14379, opened 2017-12-08
use https to load analytics

hakubo:


extensions/wikia/AnalyticsEngine/js/universal_analytics.js
```diff
@@ -39,7 +39,7 @@
 			a.async = 1;
 			a.src = g;
 			m.parentNode.insertBefore(a, m);
-		})(window, document, 'script', '//www.google-analytics.com/analytics.js', 'ga');
+		})(window, document, 'script', 'https//www.google-analytics.com/analytics.js', 'ga');
```
shaedrich, line 42: I guess that won't work because there might miss a colon after `https` and before `//`.

PR #14382, opened 2017-12-11
XW-4284 | featured video migration script

latata:
@Wikia/x-wing

### DIFF
--- a/maintenance/wikia/jwplayermigration/migrateWikiFactoryVariableToService.php
+++ b/maintenance/wikia/jwplayermigration/migrateWikiFactoryVariableToService.php
@@ -1,0 +1,112 @@
+<?php
+
+/**
+ * Add JWPlayer mediaId to wgFeaturedVideo
+ *
+ */
+require_once( dirname( __FILE__ ) . '/../../Maintenance.php' );
+
+class MigrateOoyalaVideos extends Maintenance {
+
+	protected $dryRun = true;
+
+	protected $map = [];
+
+	// pass access token
+	const ACCESS_TOKEN = '';
+
+	public function __construct() {
+		parent::__construct();
+		$this->mDescription = "Migrate videos";
+		$this->addOption( 'dry-run', 'Dry run mode', false, false, 'd' );
+
+		$videos = array_map( 'str_getcsv', file( 'output.csv' ) );
+		$map = [];
+
+		foreach ( $videos as $video ) {
+			$map[$video[1]] = $video[0];
+		}
+
+		$this->map = $map;
+	}
+
+
+	public function execute() {
+		$this->dryRun = $this->hasOption( 'dry-run' );
+		$this->output( "Start \n" );
+
+		$this->doUpdate( 1689, 'wgArticleVideoFeaturedVideos' );
+		$this->doUpdate( 1749, 'wgArticleVideoFeaturedVideos2' );
+	}
+
+	private function toDevURL($url) {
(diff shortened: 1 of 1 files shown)


maintenance/wikia/jwplayermigration/migrateWikiFactoryVariableToService.php
```diff
@@ -0,0 +1,112 @@
...
+		$data = [
+			'url' => $url,
+			'media_id' => $mediaId,
+		];
+		$options = [
+			'postData' => json_encode( $data ),
+			'headers' => [
+				'Content-Type' => 'application/json',
+				'Accept' => 'application/json',
+				'X-Wikia-AccessToken' => self::ACCESS_TOKEN,
```
rybmat, line 58: if you'd like I can easily create internal endpoint for creating mapping in the service so there is no need to use access token

maintenance/wikia/jwplayermigration/migrateWikiFactoryVariableToService.php
```diff
@@ -0,0 +1,112 @@
...
+			],
+			'returnInstance' => true,
+		];
+
+		$endpointUrl = 'https://services.wikia.com/article-video/mappings';
```
rybmat, line 63: I'm going to create swagger autogenerated api client for this service. If we are going to migrate data once and forget about this script then this way is ok, however if we'd like to merge this script to dev I'd like to use the swagger client here instead of hardcoded urls

maintenance/wikia/jwplayermigration/migrateWikiFactoryVariableToService.php
```diff
@@ -0,0 +1,112 @@
...
+	private function doUpdate( $varId, $varName ) {
+		global $wgDevelEnvironment;
+		// Get all wikis with var set
+		$wikisWithVar = ( WikiFactory::getListOfWikisWithVar( $varId, 'array', '!=', [] ) );
+
+		$this->output( 'Found ' . count( $wikisWithVar ) . "wikis with " . $varName . " config\n" );
+
+
+		// Iterate over all wikis that have the var set
+		foreach ( $wikisWithVar as $wikiId => $wiki ) {
```
rybmat, line 81: I know this way is easier now since it is copy-paste, but in the future can we just run maintenance script like that: https://wikia-inc.atlassian.net/wiki/spaces/ONE/pages/35724059/Engineering+Onboarding+How+to+run+maintenance+script
which by default iterates on each wiki, so the only thing that we'd need to do in this script would be checking if wiki has FV variable and migrate if yes

maintenance/wikia/jwplayermigration/migrateWikiFactoryVariableToService.php
```diff
@@ -0,0 +1,112 @@
...
+				$this->output( 'Url: ' . $url . "\n" );
+				if ( !empty( $config['mediaId'] ) ) {
+					$this->output( 'MediaId: ' . $config['mediaId'] . "\n" );
+					$result = $this->addMapping( $url, $config['mediaId'] );
+					$status = $result->getStatus();
+					if ( $status === 201 ) {
+						$this->output( 'Mapping has been saved.' . "\n"  );
+					} else {
+						$this->output( 'ERROR ' . $status . '. Mapping has not been saved.' . "\n" );
+						$this->output( $result->getContent() . "\n"  );
```
rybmat: how about logging status code here?

rybmat (approved):

PR #14384, opened 2017-12-11
SUS-3532 | ImageServing - add a check to prevent fatal errors

macbre:
https://wikia-inc.atlassian.net/browse/SUS-3532

Prevent `Error: Call to a member function getRevIdFetched() on null` for not existing articles.

### DIFF
--- a/extensions/wikia/ImageServing/imageServing.class.php
+++ b/extensions/wikia/ImageServing/imageServing.class.php
@@ -125,10 +125,10 @@ class ImageServing {
 
 			// fetch articles metadata (try from cache)
 			foreach ( $articles as $key => $value ) {
-				$article = Article::newFromID( $key );
+				$title = Title::newFromID( $key );
 
-				if ( $article instanceof  Article ) {
-					$mcValue = $this->memc->get( $this->makeKey( $key, $article->getRevIdFetched() ) );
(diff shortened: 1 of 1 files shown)


extensions/wikia/ImageServing/imageServing.class.php
```diff
@@ -125,11 +125,15 @@ public function getImages( $limit = 5, $driverName = null ) {
...
 			// fetch articles metadata (try from cache)
 			foreach ( $articles as $key => $value ) {
-				$mcValue = $this->memc->get( $this->makeKey( $key, Article::newFromID( $key )->getRevIdFetched() ) );
+				$article = Article::newFromID( $key );
 
-				if ( !empty( $mcValue ) ) {
-					unset( $articles[$key] );
-					$this->addArticleToList( $mcValue );
+				if ( $article instanceof  Article ) {
+					$mcValue = $this->memc->get( $this->makeKey( $key, $article->getRevIdFetched() ) );
```
mszabo-wikia: It seems to me we can bypass instantiating `Article` here and just use `Title::newFromID` + `Title::getLatest`

mszabo-wikia (changes requested):

mszabo-wikia (approved):

PR #14390, opened 2017-12-11
Manualy translations download after SER-1333

nikodamn:
SER-1333 & SER-1476
@Wikia/services-team


keikolaitong:
The translation update looks good.

PR #14394, opened 2017-12-11
SUS-3543: Script to automatically prune old rows from specials.multilookup

mszabo-wikia:
https://wikia-inc.atlassian.net/browse/SUS-3543


macbre:
I don't think we can remove old entries. CoppaTool uses this table to get the list of all wikis given IP made an edit.

PR #14396, opened 2017-12-12
XW-4337 | Integrate mediawiki with article-video service

rybmat:
https://wikia-inc.atlassian.net/browse/XW-4337

Use article-video service to get what featured video should be displayed on a page instead of wikifactory variable

@Wikia/x-wing 

#14416


extensions/wikia/ArticleVideo/ArticleVideo.hooks.php
```diff
@@ -2,9 +2,9 @@
 
 class ArticleVideoHooks {
 	public static function onBeforePageDisplay( \OutputPage $out/*, \Skin $skin*/ ) {
-		$title = RequestContext::getMain()->getTitle()->getPrefixedDBkey();
+		$pageId = RequestContext::getMain()->getTitle()->getArticleID();
```
mszabo-wikia: You can use `$out->getTitle()`

extensions/wikia/ArticleVideo/ArticleVideoContext.class.php
```diff
@@ -5,53 +5,40 @@ class ArticleVideoContext {
...
 	 * Checks if featured video is embedded on given article
 	 *
-	 * @param  string $title Prefixed article title (see: Title::getPrefixedDBkey)
+	 * @param $pageId
 	 *
 	 * @return bool
+	 *
 	 */
-	public static function isFeaturedVideoEmbedded( string $prefixedDbKey ) {
+	public static function isFeaturedVideoEmbedded( $pageId ) {
```
mszabo-wikia: should we add type hint?

mszabo-wikia (changes requested):

hakubo (approved):

mszabo-wikia (approved):

extensions/wikia/AdEngine/AdEngine2ContextService.class.php
```diff
@@ -17,7 +17,9 @@ public function getContext( Title $title, $skinName ) {
 			$adPageTypeService = new AdEngine2PageTypeService();
 			$wikiaPageType = new WikiaPageType();
 			$pageType = $wikiaPageType->getPageType();
-			$hasFeaturedVideo = !empty( $wg->EnableArticleFeaturedVideo ) && ArticleVideoContext::isFeaturedVideoEmbedded( $title->getPrefixedDBkey() );
+			$hasFeaturedVideo = !empty( $wg->EnableArticleFeaturedVideo )
+				&& !empty( $title->getArticleID() )
```
latata: I'd check it inside isFeaturedVideoEmbedded method

extensions/wikia/ArticleVideo/ArticleVideo.hooks.php
```diff
@@ -2,9 +2,9 @@
 
 class ArticleVideoHooks {
 	public static function onBeforePageDisplay( \OutputPage $out/*, \Skin $skin*/ ) {
-		$title = RequestContext::getMain()->getTitle()->getPrefixedDBkey();
+		$pageId = $out->getTitle()->getArticleID();
```
latata: `$articleId`?

extensions/wikia/ArticleVideo/ArticleVideoContext.class.php
```diff
@@ -5,53 +5,39 @@ class ArticleVideoContext {
...
 	 */
-	public static function getFeaturedVideoData( string $prefixedDbKey ) {
+	public static function getFeaturedVideoData( string $pageId ) {
 		$wg = F::app()->wg;
 
-		if ( self::isFeaturedVideoEmbedded( $prefixedDbKey ) ) {
-			$videoData = self::getFeaturedVideos()[$prefixedDbKey];
+		if ( self::isFeaturedVideoEmbedded( $pageId ) ) {
+			$videoData = [];
+			$videoData['mediaId'] = ArticleVideoService::getFeatureVideoForArticle($wg->cityId, $pageId);
```
latata: spaces

includes/wikia/services/ArticleVideoService.class.php
```diff
@@ -0,0 +1,119 @@
...
+			array_filter(
+				self::getFeaturedVideosForWiki( $cityId ),
+				function ( Mapping $mapping ) use ( $pageId ) {
+					return $mapping->getId() === (string) $pageId;
+				}
+			)
+		);
+
+		if ( !empty( $forArticle ) ) {
+			$mediaId = array_pop($forArticle);
```
latata: spaces

includes/wikia/services/ArticleVideoService.class.php
```diff
@@ -0,0 +1,119 @@
...
+	 * @return Mapping[]
+	 * @internal param $pageId
+	 *
+	 */
+	public static function getFeaturedVideosForWiki( string $cityId ): array {
+		$key = wfMemcKey( 'article-video', 'get-for-product', $cityId );
+
+		return WikiaDataAccess::cache(
+			$key,
+			WikiaResponse::CACHE_SHORT,
```
latata: Do we want to give possiblity to purge this cache? For example when adding mapping.

includes/wikia/services/ArticleVideoService.class.php
```diff
@@ -0,0 +1,119 @@
...
+					WikiaLogger::instance()->debug(
+						'could not fetch data from article-video service',
+						[
+							'exception' => $apiException,
+							'status_code' => intval( $apiException->getCode() )
+						]
+					);
+				}
+
+				return $mappings;
```
latata, line 55: is there possibility to return current cached value instead of empty array when api call fails?

extensions/wikia/ArticleVideo/ArticleVideo.hooks.php
```diff
@@ -2,9 +2,9 @@
...
 
-		$featuredVideoData = ArticleVideoContext::getFeaturedVideoData( $title );
+		$featuredVideoData = ArticleVideoContext::getFeaturedVideoData( $pageId );
 
 		if ( !empty( $featuredVideoData ) ) {
 			self::addJWPlayerAssets( $out, $featuredVideoData );
```
latata: Looks like we can remove these arguments.

extensions/wikia/ArticleVideo/ArticleVideo.hooks.php
```diff
@@ -2,9 +2,9 @@
 
 class ArticleVideoHooks {
 	public static function onBeforePageDisplay( \OutputPage $out/*, \Skin $skin*/ ) {
-		$title = RequestContext::getMain()->getTitle()->getPrefixedDBkey();
+		$pageId = $out->getTitle()->getArticleID();
 
-		$featuredVideoData = ArticleVideoContext::getFeaturedVideoData( $title );
+		$featuredVideoData = ArticleVideoContext::getFeaturedVideoData( $pageId );
```
latata: I think we can use `isFeaturedVideoEmbedded` instead of `getFeaturedVideoData`.

includes/wikia/services/ArticleVideoService.class.php
rybmat: I had an idea to do that, but it would require to use second layer of caching and as I discussed it with @hakubo it could be painful when we extend API of the service somewhere in the future.

includes/wikia/services/ArticleVideoService.class.php
rybmat: Staff users will use this tool, they have access to ?mcache=writeonly to check if video works. If we receive feedback that 3h is too long, we'll try to figure it out

extensions/wikia/AdEngine/AdEngine2ContextService.class.php
rybmat: I'd like to leave it here. It should be responsibility of client of given method to provide correct data to the method rather than doing checks within the method. This way logic within the method is clearer and focused on problem it solves. Also we use type hints for parameters, so if null is provided instead of articleId, then php will throw warning

latata (approved):

PR #14398, opened 2017-12-12
ADEN-6383 Fix JW Player tags on featured video pages

nandy-andy:


fraszczakszymon (approved):

extensions/wikia/JWPlayerTag/JWPlayerTagController.class.php
```diff
@@ -38,6 +38,7 @@ public function renderTag( $input, array $args, Parser $parser, PPFrame $frame )
 			RequestContext::getMain()->getTitle()->getPrefixedDBkey()
 		) ) {
 			$script = JSSnippets::addToStack( [
+				'jwplayer_tag_ads_js',
```
latata, line 41: I guess it will cause loading some javascript twice for articles with featured videos. But I'm not sure if there are any articles with both featured video and jwplayer tag video:
`'//extensions/wikia/AdEngine/js/video/player/jwplayer/jwplayerAdsTracking.js',
'//extensions/wikia/AdEngine/js/video/player/jwplayer/jwplayerTracker.js',`

nandy-andy: [The bug](https://wikia-inc.atlassian.net/browse/ADEN-6383) was exactly opposite: JW Player instances generated by `<jwplayer>` tag don't work on featured video pages because there are no ads assets loaded (featured video JW Player instance works). By adding this group here we fix it. If it's not a featured video page it goes to the `else` block and loads two packages. The JS package includes `jwplayer_tag_ads_js` in that case. We still can think of a cleaner way for loading those assets, though.

PR #14407, opened 2017-12-13
SUS-3547 | SpecialRenameuser - preserve Special:UserRenameTool subpage parameter in URL and do not ask for a password when staff renames a different account

macbre:
https://wikia-inc.atlassian.net/browse/SUS-3547

* preserve Special:UserRenameTool subpage parameter in URL

### DIFF
--- a/includes/User.php
+++ b/includes/User.php
@@ -3499,7 +3499,7 @@ class User implements JsonSerializable {
 	/**
 	 * Check to see if the given clear-text password is one of the accepted passwords
 	 * @param $password String: user password.
-	 * @return AuthResult: True if the given password is correct, otherwise False.
+	 * @return AuthResult
 	 */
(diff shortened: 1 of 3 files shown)


includes/User.php
```diff
@@ -3499,7 +3499,7 @@ public function isNewbie() {
 	/**
 	 * Check to see if the given clear-text password is one of the accepted passwords
 	 * @param $password String: user password.
-	 * @return Boolean: True if the given password is correct, otherwise False.
+	 * @return AuthResult: True if the given password is correct, otherwise False.
```
mszabo-wikia: `AuthResult` cannot be `true`

mszabo-wikia (changes requested):

mszabo-wikia (approved):

PR #14422, opened 2017-12-14
ADEN-6198 Fix slot targeting for test pages

nandy-andy:
https://wikia-inc.atlassian.net/browse/ADEN-6198


nandy-andy:
Nope. It needs more time.

PR #14426, opened 2017-12-15
Sus 3498 special edit count: events -> rollup

jcellary:
https://wikia-inc.atlassian.net/browse/SUS-3498

Replaced usage of events table with rollup

### DIFF
--- a/extensions/wikia/Editcount/SpecialEditcount_body.php
+++ b/extensions/wikia/Editcount/SpecialEditcount_body.php
@@ -140,7 +140,7 @@ class Editcount extends SpecialPage {
 					array( 'GROUP BY' => 'page_namespace' )
 				);
 
-				while( $row = $dbr->fetchObject( $res ) ) {
+				foreach( $res as $row ) {
 					$nscount[$row->page_namespace] = $row->count;
 				}
 			} else {
@@ -198,7 +198,7 @@ class Editcount extends SpecialPage {
 
 		if ( empty($nscount) ) {
 			$nscount = array();
-			$dbs = wfGetDB(DB_SLAVE, array(), $wgDWStatsDB);
+			$dbs = wfGetDB( DB_SLAVE, array(), $wgDWStatsDB );
 			$res = $dbs->select(
 				array( 'rollup_wiki_namespace_user_events' ),
 				array( 'namespace_id as namespace', 'sum(creates + edits) as count' ),
@@ -210,7 +210,6 @@ class Editcount extends SpecialPage {
 			while( $row = $dbs->fetchObject( $res ) ) {
 				$nscount[$row->namespace] = $row->count;
 			}
-			$dbs->freeResult( $res );
 			$wgMemc->set( $key, $nscount, self::CACHE_TIME );
 			$this->refreshTimestamps['allWikias'] = time()+self::CACHE_TIME;
 			$wgMemc->set( $keyTimestamp, $this->refreshTimestamps['allWikias'], self::CACHE_TIME );
--- a/extensions/wikia/Editcount/SpecialEditcount_body.php
+++ b/extensions/wikia/Editcount/SpecialEditcount_body.php
@@ -189,7 +189,7 @@ class Editcount extends SpecialPage {
 	}
 
 	function editsByNsAll( $uid ) {
-		global $wgStatsDB, $wgStatsDBEnabled, $wgMemc;
+		global $wgDWStatsDB, $wgMemc;
 
 		$key = wfSharedMemcKey( 'namespaceCountAllWikis', $uid );
 		$keyTimestamp = wfMemcKey( 'namespaceCountTimestamp', $uid );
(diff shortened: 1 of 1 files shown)


extensions/wikia/Editcount/SpecialEditcount_body.php
```diff
@@ -198,27 +198,19 @@ function editsByNsAll( $uid ) {
...
+				array( 'user_id' => $uid, 'period_id' => DataMartService::PERIOD_ID_DAILY ),
+				__METHOD__,
+				array ( 'GROUP BY' => 'namespace_id' )
+			);
 
-				while( $row = $dbs->fetchObject( $res ) ) {
-					$nscount[$row->namespace] = $row->count;
-				}
-				$dbs->freeResult( $res );
+			while( $row = $dbs->fetchObject( $res ) ) {
```
mszabo-wikia, line 210: We can use `foreach` loop here

extensions/wikia/Editcount/SpecialEditcount_body.php
```diff
@@ -198,27 +198,19 @@ function editsByNsAll( $uid ) {
...
+				$nscount[$row->namespace] = $row->count;
 			}
+			$dbs->freeResult( $res );
```
mszabo-wikia: Shouldn't be needed

extensions/wikia/Editcount/SpecialEditcount_body.php
```diff
@@ -198,27 +198,19 @@ function editsByNsAll( $uid ) {
...
-						'user_id' => $uid,
-						' ( event_type = 1 ) or ( event_type = 2 ) '
-					),
-					__METHOD__,
-					array (
-						'GROUP BY' => 'page_ns',
-						'ORDER BY' => 'null'
-					)
-				);
+			$dbs = wfGetDB(DB_SLAVE, array(), $wgDWStatsDB);
```
mszabo-wikia: Spaces

mszabo-wikia (changes requested):

extensions/wikia/Editcount/SpecialEditcount_body.php
jcellary: Why is there some mechanism that does that or we just don't care because PHP restarts the whole process for every request?

extensions/wikia/Editcount/SpecialEditcount_body.php
```diff
@@ -198,27 +198,19 @@ function editsByNsAll( $uid ) {
...
+			$res = $dbs->select(
+				array( 'rollup_wiki_namespace_user_events' ),
+				array( 'namespace_id as namespace', 'sum(creates + edits) as count' ),
+				array( 'user_id' => $uid, 'period_id' => DataMartService::PERIOD_ID_DAILY ),
+				__METHOD__,
+				array ( 'GROUP BY' => 'namespace_id' )
+			);
```
macbre: Let's make sure that indices are used by this query

jcellary:
```
describe select sum(creates + edits), namespace_id from statsdb.rollup_wiki_namespace_user_events where user_id = 455627 and period_id = 1 group by namespace_id;

+----+-------------+-----------------------------------+------+----------------------------------+------------------+---------+-------------+------+----------------------------------------------+
| id | select_type | table                             | type | possible_keys                    | key              | key_len | ref         | rows | Extra                                        |
+----+-------------+-----------------------------------+------+----------------------------------+------------------+---------+-------------+------+----------------------------------------------+
|  1 | SIMPLE      | rollup_wiki_namespace_user_events | ref  | PRIMARY,time_id,user_period_time | user_period_time | 6       | const,const |  617 | Using where; Using temporary; Using filesort |
+----+-------------+-----------------------------------+------+----------------------------------+------------------+---------+-------------+------+----------------------------------------------+
```

REF - I'm assuming that's good?

extensions/wikia/Editcount/SpecialEditcount_body.php
mszabo-wikia: According to method docs most of the time this method does not need to be called explicitly as standard GC is enough:

```
	/**
	 * Free a result object returned by query() or select(). It's usually not
	 * necessary to call this, just use unset() or let the variable holding
	 * the result object go out of scope.
	 *
	 * @param $res Mixed: A SQL result
	 */
	function freeResult( $res ) {
	}
```

extensions/wikia/Editcount/SpecialEditcount_body.php
macbre: `Using temporary; Using filesort` is fine as long as it's not performed on thousands of rows.

macbre (approved):

mszabo-wikia (approved):

extensions/wikia/Editcount/SpecialEditcount_body.php
```diff
@@ -198,26 +198,17 @@ function editsByNsAll( $uid ) {
...
-					__METHOD__,
-					array (
-						'GROUP BY' => 'page_ns',
-						'ORDER BY' => 'null'
-					)
-				);
+			$dbs = wfGetDB( DB_SLAVE, array(), $wgDWStatsDB );
+			$res = $dbs->select(
+				array( 'rollup_wiki_namespace_user_events' ),
+				array( 'namespace_id as namespace', 'sum(creates + edits) as count' ),
```
chessky: Let's remove the sum of creates and edits and leave just edits.

macbre (approved):

PR #14487, opened 2018-01-10
ADEN-6452 Introduces new AdEngine 3 API

leshniak:


fraszczakszymon:
Is it necessary to keep bridge in ES6 here? Can we use built ES5 ad-engine/ad-products, `require` needed modules and get rid of build step of ad-products in app repository?

extensions/wikia/AdEngine/.babelrc
```diff
@@ -0,0 +1,16 @@
...
+      }]
+    ],
+    "plugins": [
+      "transform-class-properties",
+      "transform-object-rest-spread",
+      "transform-decorators-legacy",
+      "lodash"
+    ]
+  }
+
```
fraszczakszymon: Something is wrong with indents here ;)

leshniak: fixed

leshniak:
We decided to consider bridge in ES5 in a follow up ticket, because it will change the scope of the current ticket. (discussed offline).

PR #14341, opened 2017-12-04
SUS-3475 Use background task to update city_list table

mszabo-wikia:
Replace legacy script with background task that will bump `wikicities`.`city_list`.`city_last_timestamp` when page edit/delete/undelete/move occurs, and also verify adoptability status.

https://wikia-inc.atlassian.net/browse/SUS-3475

PR #14500, opened 2018-01-11
SUS-3472: Update Special:ListUsers with MW background task

mszabo-wikia:
On edit and user rights change, queue job to update `specials.events_local_users` table.

https://wikia-inc.atlassian.net/browse/SUS-3472

### DIFF
--- a/extensions/wikia/Listusers/SpecialListusers.php
+++ b/extensions/wikia/Listusers/SpecialListusers.php
@@ -12,54 +12,60 @@ if ( !defined( 'MEDIAWIKI' ) ) {
 	exit( 1 ) ;
 }
 
-$wgExtensionCredits['specialpage'][] = array(
+$GLOBALS['wgExtensionCredits']['specialpage'][] = [
 	"name" => "Local users",
 	"description-msg" => "listusers-desc",
 	"author" => "Piotr Molski",
-	'url' => 'https://github.com/Wikia/app/tree/dev/extensions/wikia/Listusers'
-);
+	'url' => 'https://github.com/Wikia/app/tree/dev/extensions/wikia/Listusers',
+];
 
 /**
  * Messages
  */
-$wgExtensionMessagesFiles['Listusers'] = __DIR__ . '/SpecialListusers.i18n.php';
-$wgExtensionMessagesFiles['ListusersAlias'] = __DIR__ . '/SpecialListusers.alias.php';
+$GLOBALS['wgExtensionMessagesFiles']['Listusers'] = __DIR__ . '/SpecialListusers.i18n.php';
+$GLOBALS['wgExtensionMessagesFiles']['ListusersAlias'] = __DIR__ . '/SpecialListusers.alias.php';
+
+$GLOBALS['wgAutoloadClasses']['EditCountService'] = __DIR__ . '/update/EditCountService.php';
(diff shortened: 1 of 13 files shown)


extensions/wikia/Listusers/SpecialListusers_helper.php
```diff
@@ -185,13 +176,12 @@ public function loadData() {
...
 				/* select records */
 				$oRes = $dbs->select(
-					array( self::TABLE . ( ($this->mUseKey) ? ' use key('.$this->mUseKey.')' : '' ) ),
+					array( self::TABLE ),
 					array(
 						'user_id',
 						'cnt_groups',
 						'all_groups',
 						'edits',
-						'user_is_blocked',
```
macbre, line 194: Does that mean that we can drop this column?

mszabo-wikia: @macbre After this goes live and backend scripts stop inserting to this column, yes. IMO it makes no sense to keep this column, since you can't listen to the event of a block expiring - you'd have to maintain cron job just for that. Better to check if the user is blocked when rendering ListUsers.

extensions/wikia/Listusers/SpecialListusers_hooks.php
```diff
@@ -24,23 +24,32 @@ public static function Activeusers( &$list ) {
...
+			$task = UpdateListUsersTask::newLocalTask();
+			$task->call( 'updateUserGroups', $listUsersUpdate );
+			$task->queue();
+		}
+	}
+
+	public static function doEditUpdate( WikiPage $wikiPage, Revision $revision, $baseRevId, User $user ) {
+		if ( $user->isAnon() || $user->isAllowed( 'bot' ) ) {
+			return;
+		}
```
macbre: Are you sure that we want to skip bot accounts? http://poznan.wikia.com/wiki/Specjalna:U%C5%BCytkownicy lists bots...

macbre (changes requested):

extensions/wikia/Listusers/SpecialListusers_hooks.php
mszabo-wikia: seems like legacy script had this option for no reason - I'll remove it

macbre (approved):

extensions/wikia/Listusers/SpecialListusers.php
```diff
@@ -12,54 +12,60 @@
 	exit( 1 ) ;
 }
 
-$wgExtensionCredits['specialpage'][] = array(
+$GLOBALS['wgExtensionCredits']['specialpage'][] = [
```
ghost, line 15: Is this some new MW convention? If so, how should we refer to global variables in methods? Should we keep using `global` keyword or `$GLOBALS`? The latter one looks cleaner and self explanatory, especially in long methods.

extensions/wikia/Listusers/SpecialListusers_helper.php
ghost: @macbre — you can drop whatever you wish!

extensions/wikia/Listusers/SpecialListusers.php
KockaAdmiralac: Maybe [this](https://github.com/Wikia/app/pull/14341#discussion_r154668912) is related.

PR #14504, opened 2018-01-13
SRE-76 move PHP classes from config to app

ghost:
https://wikia-inc.atlassian.net/browse/SRE-76

move PHP classes from config to app

* class FullRequestLogger from config/Logging.php
* class AuditLog from config/AuditLog.php
* $wgHooks assignments from CommonExtensions

https://github.com/Wikia/config/pull/2355

@drozdo @ljagiello @TK-999 @macbre


includes/wikia/FullRequestLogger.class.php
```diff
@@ -0,0 +1,166 @@
+<?php
+
+class FullRequestLogger {
```
mszabo-wikia: Let's verify if this is still needed. In the era of global auth pages, `UserLoginSpecialController` is no longer used, and the Digital Ocean IPs may well be out of date. We might also have other sources for such access logs.

includes/wikia/AuditLog.class.php
```diff
@@ -0,0 +1,216 @@
...
+					'new_email_old' => $oldUser->getNewEmail(),
+					'new_email_new' => $user->getNewEmail(),
+				] );
+		}
+
+		return true;
+	}
+
+	static public function onAddNewAccount() {
+		return true;
```
mszabo-wikia: redundant hook handler

mszabo-wikia (changes requested):

includes/wikia/FullRequestLogger.class.php
ljagiello: Why do we even have hardcoded IPs there? It sounds wrong.

ljagiello: It looks like it was introduced back in security incident days I think we should do at CDN level not inside application.

ghost: Okay, I will review it and remove if no longer needed.

ghost:
I have removed `FullRequestLogger` completely, we are going to have ca. 3 million entries per day less in our logs. :)

ljagiello (approved):

ghost:
Removing obsolete hooks from `CheckUser` is a bonus.

mszabo-wikia (approved):

PR #14514, opened 2018-01-15
PUR-96: Fixed talkpage/reason link in Content Review module

KockaAdmiralac:
In the Content Review module, the rejection reason link is pointing to the talkpage incorrectly because the `Title` object passed to `Html::element` is unexpectedly converted to a string instead of its local URL.

Support ticket: [#364236](https://support.wikia-inc.com/hc/en-us/requests/364236)
Bug ticket: [PUR-96](https://wikia-inc.atlassian.net/browse/PUR-96)

### DIFF
--- a/extensions/wikia/ContentReview/services/ContentReviewStatusesService.class.php
+++ b/extensions/wikia/ContentReview/services/ContentReviewStatusesService.class.php
@@ -313,7 +313,7 @@ class ContentReviewStatusesService {
 	private function createRevisionTalkpageLink( \Title $title ) {
 		return \Html::element(
 			'a',
-			[ 'href' => $title->getTalkPage() ],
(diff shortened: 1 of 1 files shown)


extensions/wikia/ContentReview/services/ContentReviewStatusesService.class.php
```diff
@@ -313,7 +313,7 @@ private function createRevisionLink( \Title $title, $revisionId, $liveRevisionId
 	private function createRevisionTalkpageLink( \Title $title ) {
 		return \Html::element(
 			'a',
-			[ 'href' => $title->getTalkPage() ],
+			[ 'href' => $title->getTalkPage()->getLocalURL() ],
```
HM100, line 316: Thanks for the fix. I just wanted to see rejection message more easily so that fix is the best!!!

PR #14518, opened 2018-01-16
XW-4516 | don't treat not parsed nowiki tag as an edgecase

rybmat:
https://wikia-inc.atlassian.net/browse/XW-4516

@Wikia/x-wing

### DIFF
--- a/extensions/wikia/RTE/RTEData.class.php
+++ b/extensions/wikia/RTE/RTEData.class.php
@@ -48,7 +48,13 @@ class RTEData {
 
 				unset($data['wikitextIdx']);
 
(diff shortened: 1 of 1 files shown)


extensions/wikia/RTE/RTEData.class.php
```diff
@@ -48,7 +48,7 @@ public static function replaceIdxByData($var) {
 
 				unset($data['wikitextIdx']);
 
-				if(strpos($data['wikitext'], '_rte_wikitextidx') !== false) {
+				if(strpos($data['wikitext'], '_rte_wikitextidx') !== false && !preg_match('/<ext _rte_wikitextidx="[^"]*"><name>nowiki/', $data['wikitext'])) {
```
hakubo: please add a comment why the edge case to the edge case  :)

hakubo (changes requested):

bkoval (approved):

hakubo (approved):

PR #14520, opened 2018-01-16
SUS-3471 | remove old jobs system maintenance scripts

macbre:
https://wikia-inc.atlassian.net/browse/SUS-3471

And `MediaWiki::doJobs` which was never executed because `wgJobRunRate` was set to `0` in `CommonSettings.php`.


includes/DefaultSettings.php
```diff
@@ -5356,14 +5356,6 @@
...
  */
 
-/**
- * Number of jobs to perform per request. May be less than one in which case
- * jobs are performed probabalistically. If this is zero, jobs will not be done
- * during ordinary apache requests. In this case, maintenance/runJobs.php should
- * be run periodically.
- */
-$wgJobRunRate = 1;
-
```
macbre, line 5366: See https://github.com/Wikia/config/pull/2364

mszabo-wikia (approved):

PR #14524, opened 2018-01-16
XW-4520 Increase request timeout + retry logic for FV JWPlayer requests

mszabo-wikia:
https://wikia-inc.atlassian.net/browse/XW-4520


JamesHawking (approved):

bkoval:
Thanks for a quick reaction @TK-999 !

I would wait with merging this though - we need to figure out if it's a good idea that we postpone page rendering by 6 seconds if the JW api is not responding.

We already have different ideas on how to make video more independent from JW API. Let's meet and discuss what can be done.

PR #14529, opened 2018-01-17
XW-4524 | Do not replace placeholder when editing

bkoval:
@Wikia/x-wing


hakubo (approved):

extensions/wikia/RTE/js/plugins/template/plugin.js
```diff
@@ -537,8 +537,8 @@ RTE.templateEditor = {
 			if (this.placeholder.parent().exists()) {
 				// If placeholder has a parent = it's already inserted in the article
 				// So we are editing, not creating. Thus, replacing the existing article element
-				// With it's edited version
-				this.placeholder.replaceWith(html)
+				// With it's edited version (but keeping the placeholder)
+				this.placeholder.html($(html).children());
 			} else {
 				RTE.tools.insertElement(html);
```
rybmat, line 549: does this popup on hover work for this case?

rybmat (approved):

extensions/wikia/RTE/js/plugins/template/plugin.js
bkoval: Yeah :D

rybmat (approved):

PR #14532, opened 2018-01-17
DE-3242 Removed categories and other unused fields from api

jcellary:
https://wikia-inc.atlassian.net/browse/DE-3242

Removed fields from data warehouse api


KrystianMistrzak (approved):
LGTM

PR #14533, opened 2018-01-17
MAIN-12711 | sanitize website and fbPage

xkxd:
https://wikia-inc.atlassian.net/browse/MAIN-12711

@Wikia/iris

### DIFF
--- a/extensions/wikia/UserProfilePageV3/templates/UserProfilePage_renderUserIdentityBox.php
+++ b/extensions/wikia/UserProfilePageV3/templates/UserProfilePage_renderUserIdentityBox.php
@@ -109,7 +109,7 @@
 
 					<? if ( !empty( $user['website'] ) ): ?>
 						<li class="website">
-							<a href="<?= $user['website'] ?>" rel="nofollow">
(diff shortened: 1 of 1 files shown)


extensions/wikia/UserProfilePageV3/templates/UserProfilePage_renderUserIdentityBox.php
```diff
@@ -109,7 +109,7 @@
 
 					<? if ( !empty( $user['website'] ) ): ?>
 						<li class="website">
-							<a href="<?= $user['website'] ?>" rel="nofollow">
+							<a href="<?= Sanitizer::encodeAttribute( $user['website'] ) ?>" rel="nofollow">
```
KockaAdmiralac, line 112: Is the user prevented from inserting `javascript:` links into their website in the user-attribute service code now?

rogatty:
These attributes are already sanitized in https://github.com/Wikia/app/blob/bd0958bd993315499d6b672dfdc6d7c42afb0d82/extensions/wikia/UserProfilePageV3/UserIdentityBox.class.php#L125-L131

Doesn't make sense to do it twice as it doesn't work anyway. Let's focus on https://github.com/Wikia/pandora/pull/3799 instead.

PR #14534, opened 2018-01-17
(PLATFORM-3335) Add preference and permission for opting in to HTTPS

Grunny:
Add a user preference to allow users to opt in to HTTPS and a permission
for restricting access to it.

/cc @Wikia/core-platform-team

### DIFF
--- a/extensions/wikia/HTTPSOptIn/HTTPSOptIn.i18n.php
+++ b/extensions/wikia/HTTPSOptIn/HTTPSOptIn.i18n.php
@@ -1,0 +1,13 @@
+<?php
+
+$messages = [];
+
+$messages['en'] = [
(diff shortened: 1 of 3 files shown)


KockaAdmiralac:
Since this is a global preference, does the team that handles the user-preference service need to be contacted so they make it check for user's permissions before setting the HTTPS opt-in preference? Otherwise we would easily be able to set the preference by sending a PUT request to https://services.wikia.com/user-preference/{user_id}, bypassing the permission check.

Grunny:
@KockaAdmiralac The permissions check here only exposes the preference on the page, but we still check the permission before doing anything: https://github.com/Wikia/app/pull/14535 It will also be only enabled per wiki initially, so it doesn't matter if people who don't have the permission set the preference via the service, as the preference by itself won't have any effect. :)

harnash (approved):

PR #14536, opened 2018-01-18
Cleanup useless binaries

ljagiello:
Every release we deploy our code with Lua binaries for Win32/Win64/Darwin/Linux32. We run code only at x86_64, so there is no point to keep other architectures.


mszabo-wikia (approved):

ghost (approved):

ghost:
That was _portability according to imperative programming junkies_ — _There is no replacement for `if` statement_ :-)

Grunny (approved):

macbre (approved):

PR #14540, opened 2018-01-18
XW-4523 | improved detection of block vs inline templates, fix new line issues

rybmat:
https://wikia-inc.atlassian.net/browse/XW-4523

@Wikia/x-wing


extensions/wikia/RTE/RTEReverseParser.class.php
```diff
@@ -414,19 +414,21 @@ private function handlePlaceholder($node, $textContent) {
...
-			switch($data['type']) {
+		if ( isset( $data['type'] ) ) {
+			switch ( $data['type'] ) {
 				// templates and extension tags which are wrapped in <div> needs to have additional newline at the end
 				case 'double-brackets':
 				case 'ext':
-					if ($node->nodeName === 'div') {
+					if ( $node->nodeName === 'div' &&
+						!self::isFirstChild( $node ) &&
+						!self::previousSiblingIs( $node, [ 'h1', 'h2', 'h3', 'h4', 'h5', 'h6' ] ) ) {
```
bkoval, line 424: What about ```<header>``` - do we allow it? In theory, this can contain h1-h6 + something inside.

bkoval (approved):
LGTM, left one question

hakubo (approved):

extensions/wikia/RTE/RTEReverseParser.class.php
rybmat: I think wikitext markup does not insert <header> but we can add it here if we spot it breaks something

PR #14564, opened 2018-01-23
HOTFIX | IRIS-5482 | Re-add SpecialDiscussionsLog i18n

rogatty:
https://wikia-inc.atlassian.net/browse/IRIS-5482
https://github.com/Wikia/app/pull/14563

### DIFF
--- a/extensions/wikia/SpecialDiscussionsLog/crowdin.conf
+++ b/extensions/wikia/SpecialDiscussionsLog/crowdin.conf
@@ -1,0 +1,5 @@
+[crowdin.project]
+projectid = "ext-SpecialDiscussionsLog"
+product_person = "Kornelia"
+format = "mediawiki"
+source.file = "SpecialDiscussionsLog.i18n.php"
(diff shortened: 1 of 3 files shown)


extensions/wikia/SpecialDiscussionsLog/crowdin.conf
```diff
@@ -0,0 +1,5 @@
+[crowdin.project]
+projectid = "ext-SpecialDiscussionsLog"
+product_person = "Kornelia"
```
xkxd, line 3: put `kkydon` here because in other hand script will break while uploading

xkxd (approved):

extensions/wikia/SpecialDiscussionsLog/crowdin.conf
rogatty: Interesting, it didn't break for me. I'll change it in https://github.com/Wikia/app/pull/14563 anyway, thanks.

PR #14587, opened 2018-01-25
SUS-3846: in UserProfilePageV3 clean up

mszabo-wikia:
https://wikia-inc.atlassian.net/browse/SUS-3846

cc @Wikia/iris @Wikia/sus

### DIFF
--- a/extensions/wikia/UserProfilePageV3/templates/UserProfilePage_WikiaMobileRenderUserIdentityBox.php
+++ b/extensions/wikia/UserProfilePageV3/templates/UserProfilePage_WikiaMobileRenderUserIdentityBox.php
@@ -16,10 +16,10 @@
 				<span class=mstGr><?= wfMessage( 'user-identity-box-banned-from-chat' )->escaped(); ?></span>
 			<? endif; ?>
 			<? if( !empty($user['realName']) ): ?>
-				<h2 class=mstRn><?= wfMessage( 'user-identity-box-aka-label', htmlspecialchars( $user['realName'] ) )->plain(); ?></h2>
+				<h2 class=mstRn><?= wfMessage( 'user-identity-box-aka-label' )->rawParams( htmlspecialchars( $user['realName'] ) )->parse(); ?></h2>
 			<? endif; ?>
 			<?php if( $user['edits'] >= 0 ): ?>
-				<p class=mstEd><?= wfMessage( 'user-identity-box-edits', htmlspecialchars( $user['edits'] ) )->plain(); ?></p>
+				<p class=mstEd><?= wfMessage( 'user-identity-box-edits' )->rawParams( htmlspecialchars( $user['edits'] ) )->parse(); ?></p>
 			<?php endif; ?>
 		</hgroup>
 	</div>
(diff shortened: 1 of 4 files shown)


extensions/wikia/UserProfilePageV3/templates/UserProfilePage_WikiaMobileRenderUserIdentityBox.php
```diff
@@ -1,47 +1,46 @@
...
 					<span class=mstGr><?= $tag; ?></span>
 				<?php endforeach; ?>
 			<? endif; ?>
 			<? if(!empty($user['chatBanned'])): ?>
-				<span class=mstGr><?= wfMsg('user-identity-box-banned-from-chat'); ?></span>
+				<span class=mstGr><?= wfMessage( 'user-identity-box-banned-from-chat' )->escaped(); ?></span>
 			<? endif; ?>
 			<? if( !empty($user['realName']) ): ?>
-				<h2 class=mstRn><?= wfMsg('user-identity-box-aka-label', array('$1' => $user['realName']) ); ?></h2>
+				<h2 class=mstRn><?= wfMessage( 'user-identity-box-aka-label', htmlspecialchars( $user['realName'] ) )->plain(); ?></h2>
```
Grunny: `->plain()` should be avoided where possible, so `->parse()` or `->escaped()` should be used. Since these have `<span>` in them, I'd use `->parse()` and pass the param as a `rawParam` (with `htmlspecialchars` still there) since we probably don't want the param parsed as wikitext. The same throughout for all the `->plain()` instances.

extensions/wikia/UserProfilePageV3/templates/UserProfilePage_WikiaMobileRenderUserIdentityBox.php
```diff
@@ -1,47 +1,46 @@
...
 
 	<div class=mstInf>
 		<ul class=mstUl>
 			<? if( !empty($user['location']) ): ?>
-				<li class=mstSectLi itemprop=address><?= wfMsg('user-identity-box-location', array( '$1' => $user['location'] )); ?></li>
+				<li class=mstSectLi itemprop=address><?= wfMessage( 'user-identity-box-location', htmlspecialchars( $user['location'] ) )->plain(); ?></li>
 			<? endif; ?>
 			<? if( !empty($user['birthday']) && intval( $user['birthday']['month'] ) > 0 && intval( $user['birthday']['month'] ) < 13 ): ?>
-				<li class=mstSectLi><?= wfMsg('user-identity-box-was-born-on', array( '$1' => F::app()->wg->Lang->getMonthName( intval($user['birthday']['month']) ), '$2' => $user['birthday']['day'] )); ?></li>
+				<li class=mstSectLi><?= wfMessage( 'user-identity-box-was-born-on', F::app()->wg->Lang->getMonthName( intval( $user['birthday']['month'] ) ), htmlspecialchars( $user['birthday']['day'] ) )->plain();
```
Grunny: Since we're expecting `day` to be a number (and are enforcing that on the month), we could also wrap `$user['birthday']['day']` in `intval` so we know it's a number on output. What do you think? Same below.

extensions/wikia/UserProfilePageV3/templates/UserProfilePage_WikiaMobileRenderUserIdentityBox.php
```diff
@@ -1,47 +1,46 @@
...
-		<!-- <img src="<?= $wg->BlankImgURL ;?>" data-src="<?= $user['avatar']; ?>" itemprop=image class="lazy imgPlcHld"> --!>
-		<!-- <noscript><img src="<?= $user['avatar']; ?>" itemprop=image></noscript> -->
 		<div class=mstAva>
-			<img class=mstAvaImg src="<?= $user['avatar']; ?>" itemprop=image>
+			<img class=mstAvaImg src="<?= Sanitizer::encodeAttribute( $user['avatar'] ); ?>" itemprop=image>
 		</div>
 		<hgroup class=mstHgr>
 			<? if( !empty($user['name']) ): ?>
-				<h1 class=mstNm itemprop=name><?= $user['name']; ?></h1>
+				<h1 class=mstNm itemprop=name><?= htmlspecialchars( $user['name'] ); ?></h1>
```
xkxd, line 8: why not Sanitizer::encodeAttribure?

extensions/wikia/UserProfilePageV3/templates/UserProfilePage_WikiaMobileRenderUserIdentityBox.php
xkxd: why plain istead of escaped?

xkxd (approved):

Grunny (changes requested):

extensions/wikia/UserProfilePageV3/templates/UserProfilePage_WikiaMobileRenderUserIdentityBox.php
mszabo-wikia: For normal tag content this is OK

extensions/wikia/UserProfilePageV3/templates/UserProfilePage_WikiaMobileRenderUserIdentityBox.php
mszabo-wikia: Message contains raw HTML 😞

Grunny:
An additional thing worth considering, is ensuring the URL attributes are actually URLs, rather than trusting that the service-side validation always worked/wasn't changed/can't be bypassed. We could abstract out the URL validation logic from `Sanitizer::validateTagAttributes` (this could be a hardening step as a follow up to this PR though), i.e.
```
> print_r( Sanitizer::validateTagAttributes( ['href' => 'http://starwars.wikia.com' ], 'a' ) );
Array
(
    [href] => http://starwars.wikia.com
)

> print_r( Sanitizer::validateTagAttributes( ['href' => 'javascript:alert(0);' ], 'a' ) );
Array
(
)
```

Since HTML escaping won't help when putting a `javascript:` URL in a `href` attribute. What do you think?

extensions/wikia/UserProfilePageV3/templates/UserProfilePage_renderUserIdentityBox.php
```diff
@@ -59,16 +59,16 @@
 		<div class="masthead-info-lower">
 			<div class="contributions-details tally">
 				<? if ( !empty( $user['registration'] ) ): ?>
-					<a href="<?= Sanitizer::encodeAttribute( $user['contributionsURL'] ) ?>">
-						<em><?= $user['edits'] ?></em>
+					<a href="<?= Sanitizer::encodeAttribute( $user['contributionsURL'] ); ?>">
+						<em><?= htmlspecialchars( $user['edits'] ); ?></em>
 						<span>
-							<?= wfMessage( 'user-identity-box-edits-since-joining', $user['registration'] )->plain() ?>
+							<?= wfMessage( 'user-identity-box-edits-since-joining' )->rawParams( $user['registration'] )->parse(); ?>
```
Grunny: Registration doesn't need to be in `rawParams` (and probably shouldn't since it has no pre-escaped HTML in it).

Grunny (approved):
One minor issue left, otherwise looks good.

PR #14590, opened 2018-01-25
A small fix in the group register Russian localization

estintax:
A small fix in the group register for messages with -member in Russian localization


KockaAdmiralac:
Translation fixes should generally go through the translation team so, according to [Pull Request Policy](http://dev.wikia.com/wiki/Volunteer_Developers/Pull_Request_Policy), you should probably send your fixes through [Special:Contact](http://ru.c.wikia.com/wiki/Special:Contact/general).

PR #14592, opened 2018-01-29
ADEN-6493 Build optimizations

leshniak:


extensions/wikia/AdEngine/.npmrc
```diff
@@ -1,3 +1,3 @@
 loglevel="warn"
 save-exact=true
-registry=https://artifactory.wikia-inc.com/artifactory/api/npm/wikia-npm/
+@wikia:registry=https://artifactory.wikia-inc.com/artifactory/api/npm/wikia-npm/
```
leshniak, line 3: Currently npm freezes during package resolving when all packages are fetched from artifactory. Let's minimize this number to necessary packages.

extensions/wikia/AdEngine/package.json
```diff
@@ -19,7 +19,8 @@
   },
   "scripts": {
     "build": "webpack -p",
-    "watch": "webpack -p -w"
+    "watch": "webpack -p -w",
+    "postinstall": "npm dedupe; npm prune"
```
leshniak: Automaticaly dedupe dependencies. Unnecessary packages in nested node_modules causes redundancy in a target build because of resolving hierarchy (same thing happens if packages are linked). We can use webpack aliases, but it isn't a nice solution.

kvas-damian (approved):

PR #14597, opened 2018-01-30
Wikia/config revamp, stage 1

ghost:
Sister Wikia/config PR: https://github.com/Wikia/config/pull/2374

# Wikia/config revamp, stage 1

This is the first stage of Wikia/config revamp.  The ultimate goal is to have a clean logicless, declarative configuration that allows portability.  This is a part of SRE's _Active-Active_ project.

## Changes

- remove unneeded imports and declaration (variables are declared and files are included earlier in the code flow)
- bring back vanilla MW's `includes/DefaultSettings.php` inclusion
- bring back vanilla MW's `includes/DefaultSettings.php`
- apply FANDOM-specific settings in `includes/wikia/DefaultSettings.php` so that we know what defaults we have changed
- let core MW flow to include `includes/Hooks.php`
- load `includes/GlobalFunctions.php` earlier (normally they are loaded after the configuration, but we have WikiFactory which relies on MW code)
- load FANDOM-specific configuration in `app/LocalSettings.php` (this file is normally meant to contain custom configuration in regular MediaWiki installations)
- let class autoloader handle core MW classes loading
- load WikiFactory in `includes/wikia/DefaultSettings.php`
- let class autoloader load `WikiFactoryLoader`
- general cleanup where needed 

I have executed my config dumper against `SERVER_ID=177` and have not observed any differences between final configuration between `config-revamp-stage-1` and `dev`, so I am confident about merging and releasing this one.


maintenance/Maintenance.php
```diff
@@ -132,6 +132,10 @@ public function __construct() {
 			? getenv( 'MW_INSTALL_PATH' )
 			: realpath( dirname( __FILE__ ) . '/..' );
 
+		# FANDOM change: add common MW directories to include_path to
+		# speed up possible require/include with relative paths.
+		ini_set( "include_path", "{$IP}:{$IP}/includes:{$IP}/languages:{$IP}/lib/vendor:.:" );
```
macbre, line 137: `/lib/vendor` can highly likely go away as we now use Composer's autoloader for them.

mszabo-wikia: nah, it contains dependencies such as Solarium which are not managed by composer

includes/DefaultSettings.php
```diff
@@ -512,7 +526,7 @@
  *
  * Please specify the namespace, as in the example below.
  */
-$wgRepositoryBaseUrl = "https://commons.wikimedia.org/wiki/File:";
+$wgRepositoryBaseUrl = "http://commons.wikimedia.org/wiki/File:";
```
macbre: wikimedia.org switched to HTTPS some time ago. Let's keep using HTTPS here.

ghost: I kept 1.19.24 `DefaultSettings.php` untouched and added an override to `includes/wikia/DefaultSettings.php` https://github.com/Wikia/app/pull/14597/files#diff-013f1f772182745f48c5f1429d0cdd35R27

mszabo-wikia (approved):

macbre (approved):

includes/DefaultSettings.php
```diff
@@ -3373,6 +3490,150 @@
...
+$wgGroupPermissions['sysop']['apihighlimits']    = true;
+$wgGroupPermissions['sysop']['browsearchive']    = true;
+$wgGroupPermissions['sysop']['noratelimit']      = true;
+$wgGroupPermissions['sysop']['movefile']         = true;
+$wgGroupPermissions['sysop']['unblockself']      = true;
+$wgGroupPermissions['sysop']['suppressredirect'] = true;
+#$wgGroupPermissions['sysop']['mergehistory']     = true;
+
+// Permission to change users' group assignments
+$wgGroupPermissions['bureaucrat']['userrights']  = true;
```
Grunny: Do this get overridden to false anywhere? We don't want bureaucrats to have the `userrights` permission as it allows changing all groups, not just the ones we want to limit them to (we don't even give `userrights` to the `util` group).

I'd also want to check these match up and don't give extra permissions than those defined in `PermissionsDefinesBeforeWikiFactory.php` in config. I know this is resetting the file to the default MW 1.19, but since all the permissions config was pulled out to `PermissionsDefines*WikiFactory.php`, I'd consider just removing all these from the file again, especially since the plan is for all the permission config to be pulled out of MW and into the user-permissions service. What do you think?

Grunny (dismissed):

includes/DefaultSettings.php
```diff
@@ -1066,6 +1072,27 @@
...
+/**
+ * Minimum time, in hours, which must elapse between password reminder
+ * emails for a given account. This is to prevent abuse by mail flooding.
+ */
+$wgPasswordReminderResendTime = 24;
+
+/**
+ * The time, in seconds, when an emailed temporary password expires.
+ */
+$wgNewPasswordExpiry = 3600 * 24 * 7;
```
ljagiello: Do we need that?

ljagiello:
vanilla MW `includes/DefaultSettings.php` - is there any benefit of storing there variables unused in the code?

includes/DefaultSettings.php
ghost: @Grunny - it is discarded completely in `includes/wikia/DefaultSettings.php`: https://github.com/Wikia/app/pull/14597/files#diff-013f1f772182745f48c5f1429d0cdd35R99 - all FANDOM-specific overrides to `includes/DefaultSettings.php` are in `includes/wikia/DefaultSettings.php`, the section starts here: https://github.com/Wikia/app/pull/14597/files#diff-013f1f772182745f48c5f1429d0cdd35R23

I am hesitant about chaning `includes/DefaultSettings.php`.  Keeping it vanilla and having all overrides explicitly stated can help to understand what we changed and why.  We also randomly port pieces of code from newer versions of MediaWiki, so the ability to `diff` our file with the one from any newer version of MediaWiki might be helpful.

`DefaultSettings.php` clearly states _do not edit_, so changing it might be confusing to our younger developers, as they would expect it to be vanilla MediaWiki, untouched by us.

On the other hand, we may make a slight change in the core logic and remove `includes/DefaultSettings.php` completely in favour of `includes/wikia/DefaultSettings.php` and set all required defaults there.  It would address your remarks, as well as force developers who import fresh MW stuff (essentially @TK-999 , @macbre and myself) to consult the original `DefaultSettings.php` and port certain defaults to our `includes/wikia/DefaultSettings.php`.

Also, in the upcoming stages, I will be moving stuff from `CommonExtensions.php` to Wikia/app - as it not configuration, just importing libraries (after some cleanup).  In addition I will be merging `CommonSettings.php` and `config/LocalSettings.php` into some base config, on top of which we'll have environment-specific overrides as well as `PermissionDefines*WikiFactory.php`, and WikiFactory injected in the right place: all that in `app/LocalSettings.php`. There will be some overrides after WikiFactory, although we should aim for WikiFactory being the final layer.

ghost:
@ljagiello — my original intent to keep the vanilla `includes/DefaultSettings.php` was only for reference, so that it is clear what we changed, compared to the defaults.  But I am now considering removing `includes/DefaultSettings.php` completely in favour of `includes/wikia/DefaultSettings.php`, which would clearly state that there are no defaults other than ours, nothing "more default".

Now, when I see all the changes in Stage 1 together, I think the less files/layers the better. There is no point to keep multiple files named `DefaultSettings.php`.  :-)  An override is no longer a default.

ghost:
As for unused config variables, there will be more of them.  With @macbre we're planning to clean up WikiFactory database, too.  And also remove all entries with `$x = Y` if `$x` is actually set to `Y` in config files, globally.  It will be easier for me to remove unused variables in a single, separate run.

LocalSettings.php
```diff
@@ -1,4 +1,6 @@
 <?php
+require_once "$IP/includes/wikia/Defines.php";
+require_once "$IP/includes/wikia/DefaultSettings.php";
+require_once "$IP/includes/wikia/GlobalFunctions.php";
```
kvas-damian: I'm wondering if it's possible to include these files only in one place. Currently, I'm seeing them in multiple places which makes it harder to follow.

includes/DefaultSettings.php
Grunny: Ah, I missed that reset of the array.

Yeah, removing `includes/DefaultSettings.php` completely might be preferable. We're so far away from core MW (at least the latest versions) now so diffing is often a mess anyway, and it removes the confusion of having yet another set of default values to refer to, limiting the number of config files and overrides.

Aiming for WF being the final config layer makes sense, though the post-WF overrides may need to be there so that they explicitly cannot be overridden in WikiFactory (which is preferable in some cases).

PR #14613, opened 2018-01-31
SUS-3589 | sitemap_blobs tables is created across all wikis, but never used

macbre:
https://wikia-inc.atlassian.net/browse/SUS-3589

```php
		// Sitemaps are ONLY saved to database for NS_6 and only if wgEnableVideoSitemaps is true.
		// If you turn that var ON you're putting the pre-cached sitemaps into the db.
		// If you then turn the var back OFF you're serving the outdated sitemaps.
		// SEO-557
```

Once this is merged, we can drop all per-wiki `sitemap_blobs` tables that are empty and not queried.

`wgEnableVideoSitemaps` has no default value in config. It's set on 14 wikis in WikiFactory, but to `false` -
 http://community.wikia.com/wiki/Special:WikiFactoryReporter?varid=1240


macbre:
As for the broken unit tests, it seems that we have problems with i18n cache:

```
6) FileNamespaceSanitizeHelperTest::testSanitizeImageFilename with data set #29 ('[[файл:Luke+1.jpg]]', 'ru', null, 'Luke+1.jpg', 'Link to filename with lower c...espace')
Link to filename with lower case multibyte letter in namespace
--- Expected
+++ Actual
@@ @@
-'Luke+1.jpg'
+'файл:Luke+1.jpg'
```

macbre:
Unit tests suite does pass on my devbox.

xkxd (approved):

extensions/wikia/Sitemap/maintenance.php
```diff
@@ -34,17 +34,6 @@
 	$indexes[ $namespace ] = $sitemap->cachePages( $namespace );
 	echo " pages " . count( $indexes[ $namespace ] );
 	echo " done\n";
-
-	$includeVideo = (bool) F::app()->wg->EnableVideoSitemaps;
```
macbre, line 38: This variable has no default in the config and is set on only 14 wikis - in all case to `false`.

PR #14617, opened 2018-01-31
XW-4579 | Prevent CKE from sanitising <br style="clear:both" />

rybmat:
https://wikia-inc.atlassian.net/browse/XW-4579

@Wikia/x-wing


rybmat:
jenkins says that PHP unit tests fail but on devbox running php-all results in all tests passed

hakubo (approved):

PR #14623, opened 2018-01-31
 SUS-3859 | remove anyclip-related code

xkxd:
@Wikia/sus 

https://wikia-inc.atlassian.net/browse/SUS-3859

### CHANGE BRANCH TO DEV BEFORE MERCHING


skins/oasis/modules/OasisController.class.php
```diff
@@ -87,8 +87,6 @@ public static function onMakeGlobalVariablesScript(Array &$vars) {
 	 * at the bottom for performance reasons, but there are some exceptions for engineering reasons).
 	 *
 	 * TODO: make sure JavaScripts can be always loaded on bottom
-	 *
-	 * Note: NS_FILE pages need JS at top because AnyClips relies on jQuery.
```
mszabo-wikia, line 91: if this is no longer true, let's get rid of the exception for `NS_FILE`

mszabo-wikia (changes requested):

skins/oasis/modules/OasisController.class.php
xkxd: git history for this fragment shows that this comment is not relevant since the code related to NS_FILE check was removed in 2012 🙈

mszabo-wikia (approved):

PR #14627, opened 2018-01-31
MAIN-12799

timmyquivy:
Enabling URT on Special:Contact

### DIFF
--- a/extensions/wikia/SpecialContact2/SpecialContact.body.php
+++ b/extensions/wikia/SpecialContact2/SpecialContact.body.php
@@ -781,7 +781,7 @@ class ContactForm extends SpecialPage {
 	}
 	
 	private function isRenameAccountSupported() {
-		global $wgContLang, $wgEnableUserRenameToolExt;
+		global $wgEnableUserRenameToolExt;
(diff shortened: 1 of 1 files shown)


extensions/wikia/SpecialContact2/SpecialContact.body.php
```diff
@@ -774,4 +779,9 @@ private function isCloseMyAccountSupported() {
 		return !empty( $wgEnableCloseMyAccountExt )
 				&& in_array( $wgContLang->getCode(), $wgSupportedCloseMyAccountLang );
 	}
+	
+	private function isRenameAccountSupported() {
+		global $wgContLang, $wgEnableUserRenameToolExt;
```
Grunny: `$wgContLang` is not used.

Grunny (changes requested):

timmyquivy:
@Grunny - Change made.

extensions/wikia/SpecialContact2/SpecialContact.body.php
timmyquivy: Fixed.

macbre (approved):

PR #14647, opened 2018-02-02
(PLATFORM-3292) Support HTTPS in GlobalTitle

Grunny:
When a wiki supports HTTPS and the current request is made over HTTPS,
load the URL as HTTPS in GlobalTitle. Also, vary the cache key if it
is using HTTPS so that we don't pollute plain HTTP requests.

/cc @Wikia/core-platform-team

### DIFF
--- a/includes/wikia/GlobalTitle.php
+++ b/includes/wikia/GlobalTitle.php
@@ -719,8 +719,8 @@ class GlobalTitle extends Title {
 		 */
 		$server = WikiFactory::getVarValueByName( "wgServer", $this->mCityId );
 		if ( $server ) {
-			$this->mServer = \WikiFactory::getLocalEnvURL( $server );
-			return $server;
+			$this->mServer = $this->formatServer( $server );
+			return $this->mServer;
 		}
 
 		/**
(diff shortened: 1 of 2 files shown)


includes/wikia/GlobalTitle.php
```diff
@@ -730,13 +730,26 @@ private function loadServer() {
...
+		if ( $this->usingHTTPS() ) {
+			$server = preg_replace( '/^http:\/\//', 'https://', $server );
+		}
+
+		return \WikiFactory::getLocalEnvURL( $server );
+	}
+
+	private function usingHTTPS(): bool {
+		return WebRequest::detectProtocol() === 'https' &&
+			WikiFactory::getVarValueByName( 'wgAllowHTTPS', $this->mCityId );
```
themech, line 750: I wonder check the `wgAllowHTTPS` value here - maybe if the request was made over https, we should assume https was allowed.
I know that we want to make extra checks if https was allowed, but on the other hand we're not checking if the current user opted-in to https.

themech (changes requested):
Just one question regarding the https check

includes/wikia/GlobalTitle.php
Grunny: We're checking if the target wiki has HTTPS enabled, not the wiki of the current request. So, for example, if the request was made to https://30rock.wikia.com has `wgAllowHTTPS` set to true (and the user has opted in, etc.) so the request is allowed over HTTPS, and we were trying to use GlobalTitle to link to a page on de.30rock.wikia.com or any wiki with `wgAllowHTTPS` set to false, we will correctly link to the plain HTTP version of the page. If we didn't have this check, we'd always generate HTTPS links even if HTTPS isn't allowed on the target wiki.

themech: oh, my bad, I missed the fact that we may be generating a link to another wiki, sorry.

themech (approved):

PR #14672, opened 2018-02-06
(PLATFORM-3357) Don't send dev CSP reports to prod service

Grunny:
/cc @Wikia/core-platform-team

### DIFF
--- a/includes/wikia/Wikia.php
+++ b/includes/wikia/Wikia.php
@@ -97,8 +97,6 @@ class Wikia {
 	const DEFAULT_FAVICON_FILE = '/skins/common/images/favicon.ico';
 	const DEFAULT_WIKI_LOGO_FILE = '/skins/common/images/wiki.png';
 
-	const CSP_ENDPOINT = 'https://services.wikia.com/csp-logger/csp';
-
 	private static $vars = [];
 	private static $cachedLinker;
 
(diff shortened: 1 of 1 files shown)


includes/wikia/Wikia.php
```diff
@@ -1252,8 +1252,14 @@ static public function onWebRequestInitialized( WebRequest $request ) {
 	 * @return bool true, it's a hook
 	 */
 	static public function outputHTTPSHeaders( WebRequest $request ) {
+		global $wgDevelEnvironment, $wgDevDomain;
 		if ( WebRequest::detectProtocol() === 'https' ) {
-			$request->response()->header( "Content-Security-Policy-Report-Only: default-src https:; script-src https: 'unsafe-inline' 'unsafe-eval'; style-src https: 'unsafe-inline'; img-src https: data:; report-uri " . self::CSP_ENDPOINT );
+			if ( !empty( $wgDevelEnvironment ) ) {
```
themech: Cannot we simply use `"services.{$wgWikiaBaseDomain}"` here?

Grunny: `wgWikiaBaseDomain` is always `wikia.com` even on devboxes. We could make a services domain config var that all code can use though, and move this to config. What do you think?

themech: Yes, I would prefer having services domain in config. Can be similar to how `wgDiscussionsApiUrl` is implemented.

Grunny: And done: https://github.com/Wikia/config/pull/2385

themech (approved):

PR #14676, opened 2018-02-06
SUS 4042 | check for token in Wall/Forum where applicable

mszabo-wikia:
https://wikia-inc.atlassian.net/browse/SUS-4042

### DIFF
--- a/extensions/wikia/Forum/ForumExternalController.class.php
+++ b/extensions/wikia/Forum/ForumExternalController.class.php
@@ -1,10 +1,6 @@
 <?php
 
 class ForumExternalController extends WallExternalController {
-	public function __construct() {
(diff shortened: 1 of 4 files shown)


macbre (approved):

extensions/wikia/Forum/ForumExternalController.class.php
```diff
@@ -32,6 +28,14 @@ public function policies() {
 	 */
 
 	public function swapOrder() {
+		try {
+			// SUS-4042: Validate edit token
```
pwojcik86, line 32: Just a nitpick, but could we not have this snippet copied 4 times. Perhaps make a function that takes a function which will be invoked when the token is valid?

PR #14688, opened 2018-02-07
SUS-3927 Remove moveOn.php maintenance script

v3rth:


mszabo-wikia (changes requested):
used by MultiTasks extension

PR #14689, opened 2018-02-07
SUS-3928 Remove editOn.php maintenance script

v3rth:


mszabo-wikia (approved):

mszabo-wikia (changes requested):
used by MultiTasks extension

v3rth:
Thanks!

PR #14698, opened 2018-02-07
SUS-3934 Remove unused checkTableExists.php script

v3rth:


macbre (approved):
We can query `INFORMATION_SCHEMA` instead.

PR #14724, opened 2018-02-08
SUS-3966 Remove unused pushUploadsToImageReview.php script

v3rth:


macbre (approved):
It was a one-time script used to fix an issue with ImageReview queue.

PR #14734, opened 2018-02-09
ADEN-6676 Inlining InstantGlobals

dmnsobczak:


extensions/wikia/JSVariables/JSVariables.php
```diff
@@ -78,7 +78,12 @@ function wfJSVariablesTopScripts(Array &$vars, &$scripts) {
 
 	$vars['wgTransactionContext'] = Transaction::getAttributes();
 
-	$scripts .= Html::inlineScript("var wgNow = new Date();") .	"\n";
+	$scripts .= Html::inlineScript("var wgNow = new Date();") . "\n";
+
+	// ADEN-6676: Inline InstantGlobals to prevent AdBlock block
+	$instantGlobalsModule = new InstantGlobalsModule();
+	$instantGlobals = $instantGlobalsModule->getScript(new ResourceLoaderContext(new ResourceLoader(), $wg->Request));
```
fraszczakszymon: Could you please refactor a bit this method? (extract?) https://github.com/Wikia/app/blob/dev/extensions/wikia/InstantGlobals/InstantGlobalsModule.class.php#L48
So we don't need to pass ResourceLoaderContext. It seems not needed in this case.

extensions/wikia/JSVariables/JSVariables.php
```diff
@@ -78,7 +78,12 @@ function wfJSVariablesTopScripts(Array &$vars, &$scripts) {
...
+	$scripts .= Html::inlineScript($instantGlobals) . "\n";
```
kvas-damian: You can minify js before output: `\JavaScriptMinifier::minify( $instantGlobals )`

BTW. You're missing a lot of spaces: https://www.mediawiki.org/wiki/Manual:Coding_conventions/PHP

extensions/wikia/JSVariables/JSVariables.php
```diff
@@ -78,7 +78,12 @@ function wfJSVariablesTopScripts(Array &$vars, &$scripts) {
 
 	$vars['wgTransactionContext'] = Transaction::getAttributes();
 
-	$scripts .= Html::inlineScript("var wgNow = new Date();") .	"\n";
+	$scripts .= Html::inlineScript("var wgNow = new Date();") . "\n";
+
+	// ADEN-6676: Inline InstantGlobals to prevent AdBlock block
```
kvas-damian: It's an open source. I'd suggest removing we do that ;) Just in case. Ticket number should be enough

extensions/wikia/JSVariables/JSVariables.php
kvas-damian: `InstantGlobalsModule::getVariablesValues` isn't cached. It means ~50 memcache calls for separate variables on every pageview. Maybe we could apply very short cache on top of it?

extensions/wikia/JSVariables/JSVariables.php
dmnsobczak: Right, removed.

extensions/wikia/JSVariables/JSVariables.php
dmnsobczak: Thanks, minified. And these coding conventions are straight from hell, but reformatted.

extensions/wikia/JSVariables/JSVariables.php
dmnsobczak: @fraszczakszymon Indeed. Added default null parameter.
@kvas-damian The whole page is cached for 24 hours, do we need extra cache only for Instant Globals? But on the other hand why not, added Memcached usage.

extensions/wikia/InstantGlobals/InstantGlobalsModule.class.php
```diff
@@ -45,8 +45,16 @@ private function getVariablesValues() {
...
 	}
 
-	public function getScript( ResourceLoaderContext $context ) {
-		$variables = $this->getVariablesValues();
+	public function getScript( ResourceLoaderContext $context = null ) {
+		global $wgMemc;
+		$variables = null;
+		$cacheKey = wfMemcKey( 'instantGlobalsVariablesValues' );
+
+		if ( !( $variables = $wgMemc->get( $cacheKey ) ) ) {
```
kvas-damian: You can use `WikiaDataAccess:cache` instead: https://github.com/Wikia/app/blob/538336fcb489015585fdde30e1d19a11eca1b127/includes/wikia/WikiaDataAccess.class.php#L52-L52

extensions/wikia/InstantGlobals/InstantGlobalsModule.class.php
```diff
@@ -45,8 +45,16 @@ private function getVariablesValues() {
 		return (object) $ret;
 	}
 
-	public function getScript( ResourceLoaderContext $context ) {
-		$variables = $this->getVariablesValues();
+	public function getScript( ResourceLoaderContext $context = null ) {
+		global $wgMemc;
+		$variables = null;
+		$cacheKey = wfMemcKey( 'instantGlobalsVariablesValues' );
```
kvas-damian: As these values are wiki independent, I'd recommend using  `wfSharedMemcKey`

extensions/wikia/InstantGlobals/InstantGlobalsModule.class.php
dmnsobczak: Thanks, changed.

extensions/wikia/InstantGlobals/InstantGlobalsModule.class.php
dmnsobczak: Ok, switched.

extensions/wikia/InstantGlobals/InstantGlobalsModule.class.php
fraszczakszymon: Don't assign variables in the if statement please. It's much less readable

extensions/wikia/InstantGlobals/InstantGlobalsModule.class.php
```diff
@@ -46,15 +47,16 @@ private function getVariablesValues() {
...
+		$cacheKey = wfSharedMemcKey( 'instantGlobalsVariablesValues' );
 
-		if ( !( $variables = $wgMemc->get( $cacheKey ) ) ) {
-			$variables = $this->getVariablesValues();
-
-			$wgMemc->set( $cacheKey, $variables, 300 );
-		}
+		$variables = WikiaDataAccess::cache(
+			$cacheKey,
+			self::IG_CACHE_TIME,
```
kvas-damian: You can use `WikiaDataAccess::DEFAULT_TTL`

kvas-damian (approved):

extensions/wikia/InstantGlobals/InstantGlobalsModule.class.php
dmnsobczak: Ok, changed.

fraszczakszymon (approved):

PR #14741, opened 2018-02-09
SUS-3904 Remove unused renameUserRelatedPages.php script

v3rth:


mszabo-wikia:
Used by rename user process

PR #14772, opened 2018-02-13
ADEN-6692 Add audio tracking to kikimora

dmnsobczak:


extensions/wikia/AdEngine/js/video/player/jwplayer/jwplayerTracker.js
```diff
@@ -30,6 +30,8 @@ define('ext.wikia.adEngine.video.player.jwplayer.jwplayerTracker', [
 			return;
 		}
 
+		params.withAudio = !player.getConfig().mute;
```
kvas-damian, line 33: can we call `player.getMute()` here instead? I see we're doing it on f2.

dmnsobczak: No, we can't. Check link bellow, on app we check recirculation parameter so this variables sometimes have different values:
https://github.com/Wikia/app/blob/a6fed43bae2b32c472aeabcc932171b7bab6a398/extensions/wikia/ArticleVideo/scripts/featured-video.jwplayer.instant.js#L100

PR #14780, opened 2018-02-13
SUS-3867 | Cleanup after old spotlights stuff

mszabo-wikia:
https://wikia-inc.atlassian.net/browse/SUS-3867


macbre (approved):

nandy-andy (changes requested):
Let's wait for @nefretete if we remove this stuff (and I'm fan of this removal) we'll remove in-house wikia bar ad as well: https://github.com/Wikia/app/blob/b03df0a89ed672697e9c130d529bf1eb25f49cda/extensions/wikia/WikiaBar/js/WikiaBar.js#L73

PR #14781, opened 2018-02-13
(PLATFORM-3360) Force protocol-relative URLs for JS imports

Grunny:
Force protocol-relative URLs for JS imports when they are importing from
our domain and for a single subdomain.

This will let us automatically fix most imports for use with both HTTP and HTTPS.

/cc @Wikia/core-platform-team


skins/common/wikibits.js
```diff
@@ -720,7 +728,7 @@ if (skin != 'monaco' && skin != 'oasis') {
 window.importScriptPage = function( page, server ) {
 	var url = '/index.php?title=' + encodeURIComponent(page.replace(/ /g,'_')).replace('%2F','/').replace('%3A',':') + '&action=raw&ctype=text/javascript';
 	if( typeof server == "string" ) {
-		if( server.indexOf( '://' ) == -1  && server.substring( 0, 2 ) !== '//' ) url = 'http://' + server + '.wikia.com' + url;
+		if( server.indexOf( '://' ) == -1  && server.substring( 0, 2 ) !== '//' ) url = 'http://' + server + '.' + mw.config.get('wgWikiaBaseDomain') + url;
```
harnash, line 731: Do we want it to be `http` instead of protocol relative?

harnash (changes requested):

skins/common/wikibits.js
Grunny: We only want to make them protocol-relative if they are a single subdomain, so we don't want language subdomain wikis to be protocol-relative. This is being converted for all the methods here: https://github.com/Wikia/app/blob/0044c1603/skins/common/wikibits.js#L99 and here: https://github.com/Wikia/app/blob/0044c1603/skins/common/wikibits.js#L120 Since `server` can also be a full URL or to a language subdomain wiki, so we just handle it in one place and I didn't also do a check here for whether or not server was a single subdomain. We could also just do `server.indexOf('.') === -1 ? '//' : 'http://'` here as well for example, but will be converted later if needed as well. What do you think?

harnash: If this is handled elsewhere then let's leave it as it is and keep logic in one place.

harnash (approved):

PR #14784, opened 2018-02-14
SUS-4074 | sunset extensions/wikia/VideosModule  …

macbre:
https://wikia-inc.atlassian.net/browse/SUS-4074

This was "removed" in January 2016 as a part of an experiment with Taboola. Now do it fully :)

See e91d6f9


extensions/wikia/VideosModule/Modules/VideosModuleBase.class.php
```diff
@@ -1,281 +0,0 @@
...
-	/**
-	 * Get a list of videos for the source defined by this class
-	 *
-	 * @return array
-	 */
-	public function getVideos() {
-		$cacheKey = $this->getCacheKey();
-
-		$asyncCache = $this->getCache();
-		$asyncCache
```
mszabo-wikia, line 94: let's check if we can kill off async-cache related stuff too

mszabo-wikia (changes requested):

extensions/wikia/VideosModule/Modules/VideosModuleBase.class.php
macbre: Still used by `MediaQueryService::getTotalVideoViewsByTitle` method

macbre: Actually, a good point :) https://wikia-inc.atlassian.net/browse/SUS-4170

mszabo-wikia (approved):

PR #14789, opened 2018-02-14
SUS-4172 | Delete Phalanx Stats entries when filter is deleted

mszabo-wikia:
https://wikia-inc.atlassian.net/browse/SUS-4172


macbre (approved):
Yay for tests!

PR #14790, opened 2018-02-14
XW-4631 | Error handling for getFeaturedVideoData

bkoval:
What it does:
* Returns empty array if call to JW API returns a response that is either empty or does not include any playlist items
* In the case above, code will hide the featured video.

@Wikia/x-wing

### DIFF
--- a/extensions/wikia/ArticleVideo/ArticleVideoContext.class.php
+++ b/extensions/wikia/ArticleVideo/ArticleVideoContext.class.php
@@ -55,8 +55,6 @@ class ArticleVideoContext {
 
 			if ( empty( $details ) || empty( $details['playlist'] ) || empty( $details['playlist'][0] ) ) {
 				$logger->error( self::JWPLAYER_API_ERROR_MESSAGE );
-
-				return [];
 			} else {
 				$videoData = array_merge( $videoData, $details );
 
--- a/extensions/wikia/ArticleVideo/ArticleVideoContext.class.php
+++ b/extensions/wikia/ArticleVideo/ArticleVideoContext.class.php
@@ -2,6 +2,9 @@
 
 class ArticleVideoContext {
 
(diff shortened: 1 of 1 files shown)


extensions/wikia/ArticleVideo/ArticleVideoContext.class.php
```diff
@@ -45,28 +45,26 @@ public static function getFeaturedVideoData( string $pageId ) {
 				true
 			);
 
-			if ( !empty( $details ) ) {
+			if ( !empty( $details ) && !empty( $videoData['playlist'] ) && empty( $videoData['playlist'][0] ) ) {
```
mszabo-wikia: [want to test it? 😉](https://wikia-inc.atlassian.net/wiki/spaces/~mszabo/blog/2018/02/02/226721977/Testing+external+HTTP+clients+in+MediaWiki)

latata (commented):
How about adding logging to have data how often we do not display video and which http request caused it (to article-video service or to jwplayer api)?

extensions/wikia/ArticleVideo/ArticleVideoContext.class.php
bkoval: Sure - I'll run that by the team and we'll figure out if it can help us. Definitely worth considering!

extensions/wikia/ArticleVideo/ArticleVideoContext.class.php
```diff
@@ -45,28 +53,28 @@ public static function getFeaturedVideoData( string $pageId ) {
 				true
 			);
 
-			if ( !empty( $details ) ) {
+			if ( empty( $details ) || empty( $details['playlist'] ) || empty( $details['playlist'][0] ) ) {
+				$logger->error( self::JWPLAYER_API_ERROR_MESSAGE );
+
+				return [];
```
latata: `return` is redundant

latata (approved):

PR #14792, opened 2018-02-14
cake-2713 - adding api for linking a fandom creator community to a mw community

nmonterroso:
@Wikia/cake 
https://wikia-inc.atlassian.net/browse/CAKE-2713

Adds API that can be used to link an FC community to a MW community.


drsnyder:
LGTM

PR #14804, opened 2018-02-15
XW-4649 | [CKE] Removing all newlines in span placeholders, handle templates inside list items correctly

kolek7:
https://wikia-inc.atlassian.net/browse/XW-4649

We are removing all newlines from span placeholders.
This fixes the XW-4649 bug.
We're are making such decision because templates that are wrapped in span are inline, and shouldn't have any newlines inside.(This would be a non-valid HTML.)

@Wikia/x-wing


includes/parser/Preprocessor_DOM.php
```diff
@@ -1223,7 +1223,7 @@ function expand( $root, $flags = 0 ) {
 								if ($placeholderTag === 'span') {
 									// wrap content of inline template with non-width spaces to prevent CKE from modifying
 									// dom structure
-									$content = "&#x0200B;" . $ret['text'] . "&#x0200B;";
+									$content = "&#x0200B;" . trim($ret['text'],'\n') . "&#x0200B;";
```
latata: spaces

latata (approved):

includes/parser/Parser.php
```diff
@@ -1274,12 +1274,26 @@ function internalParse( $text, $isMain = true, $frame = false ) {
...
-			$text = preg_replace_callback('/^([\*#;:][^\n]*<div class="placeholder placeholder-double-brackets"[^>]+>&#x0200B;)\n?(.*?)(&#x0200B;<\/div>)/ms', function($matches) {
-				return preg_replace('/\\n/', '', $matches[0]);
-			}, str_replace("\r\n", "\n", $text));
-			$text = preg_replace_callback('/^([\*#;:][^\n]*<span class="placeholder placeholder-double-brackets"[^>]+>&#x0200B;)\n?(.*?)(&#x0200B;<\/span>)/ms', function($matches) {
-				return preg_replace('/\\n/', '', $matches[0]);
-			}, str_replace("\r\n", "\n", $text));
+
+			do {
+				$before = $text;
+				$text = preg_replace_callback('/^([\*#;:][^\n]*<div class="placeholder placeholder-double-brackets"[^>]+>&#x0200B;)(.*?)(&#x0200B;<\/div>)/ms',
```
bkoval, line 1280: Could we run these checks inside preprocessor_dom, straight after a decision about which placeholder (div or span) should be displayed? This way we could lower the complexity of these regexes :)

rybmat: @bart unfortunately not, in preprocessor_dom we have access only to template's wikitext so we don't know if it is used inside list or not. Here we have access to whole page wikitext and these regexes checks if given line starts with *, #, ; or : which are used to create list

bkoval: OK, I get it.

bkoval (approved):

PR #14805, opened 2018-02-15
PLATFORM-3369: send beacon errors to event-logger

themech:
We don't use loggly, so switching to event-logger service.
Also services urls are generated when needed based on a shared `$wgServicesExternalDomain` variable introduced in https://github.com/Wikia/config/pull/2403


extensions/wikia/Track/templates/track.mustache
```diff
@@ -29,13 +29,17 @@
...
-			document.body.appendChild(image);
-		});
+		try {
+			var request = new XMLHttpRequest(),
+				data = {
+					name: 'MediaWiki.track',
+					description: 'wikia_beacon_id is empty',
+					client: 'oasis'
+				};
+			request.open('POST', '{{{event-logger-url}}}/error', true);
```
Grunny: We should default to HTML escaping on output, even if it's a hard coded string, so `{{event-logger-url}}` would be preferred.

extensions/wikia/Discussions/Discussions.setup.php
```diff
@@ -66,3 +66,5 @@
 $wgGroupPermissions['staff']['specialdiscussions'] = true;
 
 $wgHooks['WikiaSkinTopScripts'][] = 'DiscussionsHooksHelper::addDiscussionJsVariable';
+
+$wgDiscussionsApiUrl = $wgServicesExternalDomain . 'discussion';
```
Grunny: Should this one use the URL provider wherever it is used as well?

Grunny (changes requested):

extensions/wikia/Discussions/Discussions.setup.php
harnash: fix: 3c073840caa

extensions/wikia/Track/templates/track.mustache
harnash: fix: 7f0c497dfd4

Grunny (approved):

PR #14807, opened 2018-02-15
SUS | remove Java injectors from PHP 🐘

mszabo-wikia:
Java DI style does not play well with PHP 🐘 — time to sunset it
No Java

using a simpler approach here results in
* no useless dependencies
* less overhead (no dependency container, no reflection, no "annotation parsing" via regex etc.)
* type safety instead of relying on inline comments
* ability to stub/mock dependencies selectively instead of having to build full container

tests are failing b/c of circular dependency on config PR in CI env—when executed locally with proper config they pass. 🍏


composer.json
```diff
@@ -18,6 +18,12 @@
     "prepend-autoloader": false,
     "optimize-autoloader": true
   },
+  "autoload": {
+    "psr-4" : {
+      "Wikia\\": "lib/Wikia/src/",
+      "Swagger\\Client\\": "lib/Swagger/src/"
+    }
+  },
```
macbre, line 26: :+1:

extensions/wikia/FacebookPreferences/tests/FacebookPreferencesControllerTest.php
```diff
@@ -34,14 +35,12 @@ protected function setUp() {
 		$context = new RequestContext();
 		$context->setUser( $this->userMock );
 
+		ServiceFactory::instance()->externalAuthFactory()->setFacebookService( $this->facebookServiceMock );
+
```
macbre, line 39: :+1:

macbre (approved):

PR #14816, opened 2018-02-16
SUS | Optimize Search autoloader

mszabo-wikia:
Since Search was included after the composer autoloader, any lookup of classes in the `Wikia\Search` namespace would fail over to a PSR-4 scan of `lib/Wikia/src` before invoking Search's custom autoloader function. Search is enabled globally, so we can simply register its classes with the composer autoloader for better performance.


macbre (approved):
This required good QA-ing

PR #14828, opened 2018-02-20
(PLATFORM-3365) Use HTTPS in Search result URLs

Grunny:
When a search page is requested over HTTPS, the URLs created should be
HTTPS links.

This also adds methods for converting a URL to HTTPS or HTTP to make
grepping for cleanup easier in the future.

/cc @Wikia/core-platform-team


mszabo-wikia (approved):
LGTM, we could consider using substr instead of regex in case we end up performing a lot of these replacement somewhere.

macbre (approved):
I second Mate's hint.

ludwikkazmierczak:
TESTED AND ACCEPTED

PR #14830, opened 2018-02-20
SUS-3851 | Support for loading different wikis from the same domain based on the language code in the path

mszabo-wikia:
Have WikiFactory take optional language code from the path and use it when distinguishing between wikis. This allows several wikis to share a root domain while remaining fully compatible with our current setup.

This change is dependent on config changes which causes test failures in CI - when run locally with correct config the tests pass. 🍏

https://wikia-inc.atlassian.net/browse/SUS-3851


skins/oasis/modules/BlogListingController.class.php
```diff
@@ -33,9 +33,6 @@ static function renderBlogListing(&$html, $posts, $aOptions, $sPager = null) {
 
 		// macbre: prevent PHP warnings and try to find the reason of them
 		if (!is_array($posts)) {
-			$url = wfGetCurrentUrl();
```
ghost, line 36: Is this a part of this changeset? :-)

mszabo-wikia: Yeah, the function was only used in WikiFactoryLoader and here so I removed it.

extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
```diff
@@ -395,82 +414,37 @@ public function execute() {
...
 		 */
 		if( $this->mIsWikiaActive == 2 && !$this->mCommandLine ) {
-			$this->debug( "city_id={$this->mWikiID};city_public={$this->mIsWikiaActive}), redirected to {$this->mCityHost}" );
+			$this->debug( "city_id={$this->mWikiID};city_public={$this->mIsWikiaActive}), redirected to {$this->mCityUrl}" );
 			header( "X-Redirected-By-WF: 2" );
-			header( "Location: http://{$this->mCityHost}/", true, 301 );
+			header( "Location: {$this->mCityUrl}", true, 301 );
 			wfProfileOut( __METHOD__ );
-			exit(0);
+			return false;
```
macbre, line 421: :heart:

extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
```diff
@@ -254,7 +252,7 @@ public function execute() {
 			 * (domain), $this->mCityId is unknown (set to false in constructor)
 			 */
 			wfProfileIn( __METHOD__."-domaincache" );
-			$key = WikiFactory::getDomainKey( $this->mServerName );
+			$key = WikiFactory::getDomainKey( $this->mServerName . '/' . $this->langCode );
```
Grunny, line 255: For existing wikis without the path, this would invalidate the cache for all wikis when released because of the added `/`, right? I assume that probably won't be a problem (I don't think there's anything heavy that's cached in it), but wanted to check.

extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
```diff
@@ -395,82 +414,37 @@ public function execute() {
...
-			// Special treatment for paragon.wiki
-			// For URLs starting with upper-case letter (/Something) redirect to /wiki/Something
-			$wikiPrefix = '';
-			if ( $host == 'paragon.wikia.com' ) {
-				$firstLetter = substr( $url['path'], 1, 1 );
-				if ( is_string( $firstLetter ) && ctype_upper( $firstLetter ) ) {
-					$wikiPrefix = '/wiki';
-				}
-			}
+			$url = rtrim( $this->mCityUrl, '/' );
```
Grunny: Minor suggestion: Since `$url` before and after this block is a parsed URL array, I'd avoid overwriting it to a string in this block just for code clarity (for example, while reviewing the part below setting the script path, I looked back up for the last reference to `$url` and found this, and needed to double check it would be an array then). We could probably even just have `$target = rtrim( $this->mCityUrl, '/' ) . '/' . $this->pathParams;` below. What do you think?

extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
mszabo-wikia: Yes, we'll have to bump cache version inside this method anyways since we will need full URL instead of just host. Since it's just a simple query I think this is fine.

extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
mszabo-wikia: good point, this was not intentional :)

macbre (approved):

fraszczakszymon (approved):

extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
```diff
@@ -146,19 +147,18 @@ public function  __construct( $id = null, $server_name = false ) {
...
+		$name = preg_replace( "/^www\./", "", $this->mServerName );
+
+		foreach ( $wikiFactoryDomains as $domain ) {
+			$tldLength = strlen( $this->mServerName ) - strlen( $domain );
+
+			if ( $domain !== "wikia.com" && strpos( $this->mServerName, $domain ) === $tldLength ) {
+				$this->mOldServerName = $this->mServerName;
+				$this->mServerName = str_replace( $domain, "wikia.com", $name );
+				$this->mAlternativeDomainUsed = true;
+				break;
```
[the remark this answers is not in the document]
TK-999: I think we don't have enough information yet about what that will entail and how.

Grunny (approved):

PR #14837, opened 2018-02-21
De 3343 Added api for wiki articles

jcellary:
https://wikia-inc.atlassian.net/browse/DE-3343

Added new api for wiki articles so that it includes "created_at" column which was missing in the old api which queried the global pages table.

### DIFF
--- a/includes/wikia/api/DWDimensionApiController.class.php
+++ b/includes/wikia/api/DWDimensionApiController.class.php
@@ -237,7 +237,7 @@ class DWDimensionApiController extends WikiaApiController {
 						'article_id' => $row->article_id,
 						'title' => $row->title,
 						'is_redirect' => $row->is_redirect,
-						'created_at' => (new DateTime($row->created_at))->format('Y-m-d H:i:s')
+						'created_at' => ( new DateTime( $row->created_at ) )->format( 'Y-m-d H:i:s' )
 					];
 				}
 				$db->freeResult( $rows );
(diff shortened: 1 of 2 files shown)


includes/wikia/api/DWDimensionApiControllerSQL.class.php
```diff
@@ -42,6 +42,15 @@ class DWDimensionApiControllerSQL {
             page_wikia_id, page_id
         LIMIT $limit';
 
+	const DIMENSION_WIKI_ARTICLES = '
+		SELECT
+            page_namespace AS namespace_id,
+            page_id AS article_id,
+            page_title AS title,
+            page_is_redirect AS is_redirect,
+            (SELECT MIN(rev_timestamp) from revision r where r.rev_page = p.page_id) as created_at
```
mszabo-wikia, line 51: I believe a join instead of subquery can work here since joining on `rev_page = page_id` should return the oldest revision (needs verification though)

includes/wikia/api/DWDimensionApiController.class.php
```diff
@@ -222,6 +222,37 @@ public function getWikiCategories() {
...
 	}
 
+	public function getWikiArticles() {
+		$this->getDataPerWiki( array( $this, 'getWikiArticlesData' ) );
+	}
+
+	private function getWikiArticlesData( DatabaseMysqli $db ) {
+		$result = [];
+		try {
+			$rows = $db->query( DWDimensionApiControllerSQL::DIMENSION_WIKI_ARTICLES, __METHOD__ );
```
mszabo-wikia, line 232: We could build query using MW `select` helper instead of storing it as string

mszabo-wikia (changes requested):

mszabo-wikia (approved):

PR #14860, opened 2018-02-23
XW-4654 | Send media data as data attrs and remove relatedPages from MercuryApi::getPage response

rybmat:
https://wikia-inc.atlassian.net/browse/XW-4654

@Wikia/x-wing


extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
```diff
@@ -382,7 +386,8 @@ public static function onParserAfterTidy( Parser $parser, &$text ): bool {
 				[
 					'content' => $text,
 					'media' => self::$media,
-					'users' => self::$users
+					'users' => self::$users,
+					'heroImage' => empty( self::$heroImage ) ? null : self::$heroImage
```
hakubo: what do you think about setting heroImage as null by default
this way we would not have to have this ternary here

hakubo (approved):

extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
rybmat: 👍

JamesHawking (approved):

hakubo (approved):

bkoval (approved):

JamesHawking (approved):

PR #14873, opened 2018-02-26
SUS | Make Nirvana local URLs respect wgScriptPath

mszabo-wikia:


macbre (approved):
Let's add test cases in `WikiaDispatchableObjectTest`

mszabo-wikia:
There are already tests there with custom path

PR #14878, opened 2018-02-26
ADEN-6759 Adding Wikia key-vals to Rubicon Video

dmnsobczak:


kvas-damian (dismissed):
Please make sure rubicon isn't returning 400s before merging.

extensions/wikia/AdEngine/js/lookup/prebid/adapters/rubicon.js
```diff
@@ -27,12 +28,48 @@ define('ext.wikia.adEngine.lookup.prebid.adapters.rubicon', [
...
+			adUnit,
+			bidParams;
+
+		bidParams = {
+			accountId: rubiconAccountId,
+			siteId: rubiconSiteId,
+			zoneId: config.zoneId,
+			name: slotName,
+			position: config.position,
+			inventory: targeting,
```
kvas-damian: In display we're using `config.targeting` by default. Do you think we should use it here as well?

extensions/wikia/AdEngine/js/lookup/prebid/adapters/rubicon.js
```diff
@@ -27,12 +28,48 @@ define('ext.wikia.adEngine.lookup.prebid.adapters.rubicon', [
...
+
+		return {
+			pos: slotName,
+			src: provider,
+			s0: adLogicZoneParams.getSite(),
+			s1: s1,
+			s2: adLogicZoneParams.getPageType(),
+			lang: adLogicZoneParams.getLanguage()
+		};
+	}
```
kvas-damian: This method is a copy&past from `rubiconDisplay.js`. Can we extract it?

kvas-damian:
Maybe we could try pass arrays as a value. See an example here: https://github.com/prebid/Prebid.js/pull/950

extensions/wikia/AdEngine/js/lookup/prebid/adapters/rubicon.js
dmnsobczak: Good idea. Refactored.

kvas-damian (approved):

PR #14883, opened 2018-02-26
Fix: use $wgDisableHTTPSDowngrade in HTTPSOptInHooks

harnash:
There was one function call left after refactoring.

### DIFF
--- a/extensions/wikia/HTTPSOptIn/HTTPSOptInHooks.class.php
+++ b/extensions/wikia/HTTPSOptIn/HTTPSOptInHooks.class.php
@@ -26,13 +26,14 @@ class HTTPSOptInHooks {
 	}
 
 	public static function onMercuryWikiVariables( array &$wikiVariables ): bool {
+		global $wgDisableHTTPSDowngrade;
 		$basePath = $wikiVariables['basePath'];
 		$user = RequestContext::getMain()->getUser();
 		if ( startsWith( $basePath, 'http://' ) && self::userAllowedHTTPS( $user ) ) {
 			$wikiVariables['basePath'] = preg_replace( '/^http:\/\//', 'https://', $basePath );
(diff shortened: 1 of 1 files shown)


extensions/wikia/HTTPSOptIn/HTTPSOptInHooks.class.php
```diff
@@ -26,13 +26,14 @@ public static function onGetPreferences( User $user, array &$preferences ): bool
...
 	public static function onMercuryWikiVariables( array &$wikiVariables ): bool {
+		global $wgDisableHTTPSDowngrade;
 		$basePath = $wikiVariables['basePath'];
 		$user = RequestContext::getMain()->getUser();
 		if ( startsWith( $basePath, 'http://' ) && self::userAllowedHTTPS( $user ) ) {
 			$wikiVariables['basePath'] = preg_replace( '/^http:\/\//', 'https://', $basePath );
 		} elseif ( startsWith( $basePath, 'https://' ) &&
 			!self::userAllowedHTTPS( $user ) &&
-			!self::disableHTTPSDowngrade()
+			!$wgDisableHTTPSDowngrade
```
Grunny: We should either check `empty( $wgDisableHTTPSDowngrade )` here, since there isn't a default value for `wgDisableHTTPSDowngrade`, or set `$wgDisableHTTPSDowngrade = false;` as the default value in our config.

Grunny (changes requested):

extensions/wikia/HTTPSOptIn/HTTPSOptInHooks.class.php
harnash: @Grunny fix in 53af388e60f

Grunny (approved):

PR #14884, opened 2018-02-27
Update links in global footer pointing to contact pages

mklucsarits:
Instead of relying on VCL to redirect from www.wikia.com/mediakit to fandom.wikia.com, let's just update the URLs directly.

The Media Kit page isn't translated (yet), so I've removed the uselang parameter for now, as it won't do anything. Additionally, the /contact page is outdated; the general Contact link should instead point to /about, which is translated.

//cc @frankfarmer


vforge (approved):
Let's just replace `'contact' => '//fandom.wikia.com/about'` with `'contact' => '//fandom.wikia.com/about#contact'` - there's already PR for that feature on the about page: https://github.com/Wikia/f2/pull/1614

PR #14896, opened 2018-02-27
ADEN-6544 Bump ad-products version

dmnsobczak:


extensions/wikia/AdEngine/js/run/mercury.run.js
```diff
@@ -81,9 +81,17 @@ require([
...
+		window.addEventListener('adengine.emitter', function (event) {
+			adEngineBridge.passSlotEvent(event.detail.slotName, event.detail.eventName);
+		});
 	});
 
 	mercuryListener.afterPageWithAdsRender(function () {
 		callBiddersOnConsecutivePageView();
 	});
+
+	mercuryListener.onMenuOpen(function () {
```
kvas-damian: In https://github.com/Wikia/app/blob/327eda7f4904975cc9db1484e900ddee30077264/extensions/wikia/AdEngine/src/templates/big-fancy-ad-above-mobile-config.js we have access to `adSlot` and `mercuryListener` maybe we could define it there?

It's more obvious for me to have all BFAA configuration in one place.

kvas-damian: Did you think about removing the callback after the page change?

dmnsobczak: Moved and added. Thanks!

extensions/wikia/AdEngine/src/templates/big-fancy-ad-above-mobile-config.js
```diff
@@ -36,15 +36,19 @@ export const getConfig = mercuryListener => ({
...
+		if (mercuryListener) {
+			mercuryListener.onPageChange(() => {
 				page.classList.remove('bfaa-template');
 				document.body.classList.remove('vuap-loaded');
 				document.body.classList.remove('has-bfaa');
 				document.body.style.paddingTop = '';
 				this.adsModule.setSiteHeadOffset(0);
 				window.removeEventListener('resize', onResize);
 			});
+			mercuryListener.onMenuOpenClear();
```
kvas-damian: We can move this method call into `mercuryListener`.

extensions/wikia/AdEngine/src/ad-engine.bridge.js
```diff
@@ -136,10 +136,15 @@ function checkAdBlocking(detection) {
 	);
 }
 
+function passSlotEvent(slotName, eventName) {
```
kvas-damian, line 139: We don't use this method. Do you see a value in keeping it in code?

extensions/wikia/AdEngine/src/templates/big-fancy-ad-above-mobile-config.js
dmnsobczak: Yes, we can! Moved.

extensions/wikia/AdEngine/src/ad-engine.bridge.js
dmnsobczak: It's not used now, but on mobile-wiki someone added this method:
https://github.com/Wikia/mobile-wiki/blob/c12cf9a36abd9f842a3fe94542f70d4961658a98/app/modules/ads.js#L192
So in the future it might be good to have working function to pass events to AdEngine 3, when he will be back to this task.

kvas-damian (approved):

PR #14897, opened 2018-02-27
DE-3343 Added created_at to pages

jcellary:
Added page_created_at to pages table. I haven't tested these changes yet, as I need to extend the schema first. Let me know if the approach makes sense and if so, I'll run an alter table (on dev first) and test the changes. The value is needed for API that imports articles to qubole

Also I will created some script to update existing entries after rolling this out.

### DIFF
--- a/extensions/wikia/Pages/PagesEntry.php
+++ b/extensions/wikia/Pages/PagesEntry.php
@@ -24,9 +24,6 @@ class PagesEntry implements JsonSerializable {
 	/** @var string $latestRevisionTimestamp */
 	private $latestRevisionTimestamp;
 
-	/** @var string $createdAtTimestamp */
-	private $createdAtTimestamp;
-
 	public static function newFromPageAndRevision( WikiPage $wikiPage, Revision $revision ): PagesEntry {
 		$pagesEntry = new PagesEntry();
 
@@ -42,8 +39,6 @@ class PagesEntry implements JsonSerializable {
 
 		$pagesEntry->latestRevisionId = $revision->getId();
 		$pagesEntry->latestRevisionTimestamp = $revision->getTimestamp();
-
-		$pagesEntry->createdAtTimestamp = Revision::getOldestRevisionTimestamp( $revision->getPage() );
 
 		return $pagesEntry;
 	}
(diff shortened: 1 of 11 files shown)


includes/Revision.php
```diff
@@ -258,6 +258,24 @@ public static function loadFromTimestamp( $db, $title, $timestamp ) {
...
+	 */
+	public static function getOldestRevisionTimestamp( $pageid ) {
+		//use master, as this function can be called during article creation hook handling
+		$db = wfGetDB( DB_MASTER );
+		return $db->selectField(
+			'revision',
+			'rev_timestamp',
+			[ 'rev_page' => intval( $pageid ) ],
+			__METHOD__,
+			[ 'ORDER BY' => 'rev_id', 'LIMIT' => 1 ]
```
jcellary: Is this assumption fine? That revision with lowest ID will be the oldest one?

extensions/wikia/Pages/PagesEntry.php
```diff
@@ -40,6 +43,8 @@ public static function newFromPageAndRevision( WikiPage $wikiPage, Revision $rev
 		$pagesEntry->latestRevisionId = $revision->getId();
 		$pagesEntry->latestRevisionTimestamp = $revision->getTimestamp();
 
+		$pagesEntry->createdAtTimestamp = Revision::getOldestRevisionTimestamp( $revision->getPage() );
```
mszabo-wikia: We could avoid a database query here: `UpdatePagesTask` will perform an `INSERT ... ON DUPLICATE KEY UPDATE` into `dataware.pages`. If it's an `INSERT` then the page was just created, and we can insert the current revision timestamp as `dataware.pages.page_created_at`. Otherwise, we do not specify `page_created_at` for `UPDATE` so changes to existing rows will not change it.

mszabo-wikia (changes requested):

extensions/wikia/Pages/PagesEntry.php
sqreek: as far as I know those data could be deleted (or don't exists atm), that's why it should be possible to retrieve created_at at any time

sqreek:
PagesEntry are used only by schedule task, and nowhere else created_at is needed.
getOldestRevisionTimestamp needs only page_id.
Because of above I prefer to put getOldestRevisionTimestamp out of revision class (to keep revision as close as possible to mw) and use it at class where you needed (so it will be PagesEntry)

extensions/wikia/Pages/PagesEntry.php
mszabo-wikia: I think it's better to skip reading from master here and if a maintenance script or other thing needs this info, they can make a SQL query

jcellary:
Thanks for the tips - I removed the stuff from revision and used Mates suggestion to use revision timestamp but only on insert

PR #14910, opened 2018-03-01
Revert "SUS-3851 | Support for loading different wikis from the same domain based on the language code in the path"

ludwikkazmierczak:
Reverts Wikia/app#14830


ludwikkazmierczak:
MIx reviewed

PR #14920, opened 2018-03-05
PLATFORM-3367: use non-protocol relative urls in emails

harnash:
Updating links in emails to use HTTP protocol.


Grunny (approved):

harnash:
@Grunny does this affect somehow crowdin and translations? Do we need to do some extra steps to make it work on prod?

Grunny:
@harnash we'd only need extra steps if we wanted to localise the magic word, which I don't think we would in this case, otherwise it should work fine with the definition in MessagesEn.php for all languages.

harnash (commented):
I've grepped the sources and there are only correct one used: `{{SERVERCANONICAL}}`

PR #14922, opened 2018-03-05
PLATFORM-3403: script for making image urls in custom JS files HTTPS-compatible

themech:


maintenance/wikia/HttpsMigration/migrateCustomJs.php
```diff
@@ -0,0 +1,350 @@
...
+			if ($text !== $updatedText) {
+				if ( $this->saveChanges ) {
+					// pay attention to JS-review process. if the latest revision is the same as
+					// the one returned from ContentReviewHelper's getReviewedRevisionId, just autoapprove
+					// (which requires FANDOMBot to be added to the content review group.
+					// autoapprove can be set by using wpApprove parameter.
+					// if the current JS revision is not approved yet, do not autoapprove JS changes!
+					$latestReviewedID = $this->contentReview->getReviewedRevisionId( $title->getArticleID(), $wgCityId );
+					if ( $revId === $latestReviewedID ) {
+						$this->output( "Auto-approving change\n" );
```
harnash: As discussed we should autoapprove images urls and go through review flow for other changes

harnash (approved):
For dry-run it looks ok but we should probably add some more logic (autoapprove) before running it in "normal" mode.

PR #14924, opened 2018-03-05
SUS-4291 | video_info - avoid SELECT * query

macbre:
https://wikia-inc.atlassian.net/browse/SUS-4291

### DIFF
--- a/extensions/wikia/VideoHandlers/videoInfo/VideoInfo.class.php
+++ b/extensions/wikia/VideoHandlers/videoInfo/VideoInfo.class.php
@@ -253,7 +253,15 @@ class VideoInfo extends WikiaModel {
 
 			$row = $db->selectRow(
 				'video_info',
-				'*',
+				[
+					"video_title",
+					"video_id",
+					"provider",
+					"added_at",
+					"added_by",
+					"duration",
+					"removed",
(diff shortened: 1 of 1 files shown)


extensions/wikia/VideoHandlers/videoInfo/VideoInfo.class.php
```diff
@@ -253,7 +253,15 @@ public static function newFromTitle( $videoTitle ) {
...
-				'*',
+				[
+					"video_title",
+					"video_id",
+					"provider",
+					"added_at",
+					"added_by",
+					"duration",
+					"removed",
+				],
```
macbre, line 264: List taken from `self::newFromRow` method below.

sqreek (approved):

extensions/wikia/VideoHandlers/videoInfo/VideoInfo.class.php
shaedrich: Wouldn't `self::$fields` suffice?

macbre: Nope, they have capitalised letters instead of underscores.

PR #14926, opened 2018-03-05
SUS-4279 | Refactor WikiFactory variable manager to not use eval()

mszabo-wikia:
Refactor WikiFactory variable manager to not use `eval()` - JSON format is used instead for arrays. Additionally tests were added for validation logic and AJAX endpoints.

https://wikia-inc.atlassian.net/browse/SUS-4279


extensions/wikia/WikiFactory/api/ApiWikiFactorySaveVariable.php
```diff
@@ -0,0 +1,70 @@
...
+			$this->dieUsageMsg( 'badaccess-groups' );
+		}
+
+		$params = $this->extractRequestParams();
+
+		try {
+			$result = WikiFactory::validateAndSetVarById(
+				$params['variable_id'],
+				$params['wiki_id'],
+				$params['variable_value'],
```
macbre, line 19: Won't the variable value be always passed as a string (even if a user provides an array or an integer)?

extensions/wikia/WikiFactory/WikiFactoryVariableParser.php
```diff
@@ -0,0 +1,80 @@
...
+	 * @param $value
+	 * @return array
+	 * @throws WikiFactoryVariableParseException
+	 */
+	private function handleArrayType( $value ): array {
+		if ( is_array( $value ) ) {
+			return $value;
+		}
+
+		$decoded = json_decode( (string)$value, true );
```
macbre, line 72: Aren't arrays kept in `array('foo' => 123)` notation?

mszabo-wikia: This method receives user input from the API which will be a JSON string. We'll decode it to a PHP array which will then be serialized.

Previously the AJAX endpoint would indeed receive a PHP array in `var_export` format (i.e. `array('foo' => 123)`), which would be passed to `eval` and then serialized.

extensions/wikia/WikiFactory/api/ApiWikiFactorySaveVariable.php
mszabo-wikia: Yep it will be - the variable parser will take care of formatting based on the variable type as set in database.

macbre (approved):

PR #14950, opened 2018-03-08
Sus-2047 update timeago

sqreek:
https://wikia-inc.atlassian.net/browse/SUS-2047
upgrade timeago to newer version.
because logic is behind many languages (like pl, ar etc) this logic should be in js.
What that means is translations are also in js.

Also question is this way of loading script is appropriate


resources/wikia/libraries/jquery/timeago/locales/jquery.timeago.pl.js
```diff
@@ -0,0 +1,34 @@
+console.log("zupa");
```
kvas-damian, line 1: Please don't use console.log in production code ;)

PR #14973, opened 2018-03-12
ADEN-6826 Fix global nav user logged with UAP hidding

dmnsobczak:


extensions/wikia/DesignSystem/scripts/DesignSystemGlobalNavigationHeadroom.js
```diff
@@ -31,7 +31,7 @@ $(function() {
 					mutations.forEach(function(mutation) {
 						if (
 							mutation.type === 'attributes' && mutation.attributeName === 'class' &&
-							mutation.target.classList.contains('bfaa-pinned')
+							mutation.target.classList.contains('bfaa-pinned') && globalNavigation.get(0).offsetTop
```
kvas-damian, line 34: This change suggests me that we aren't fixing the root cause of this bug. 
I believe that `bfaa-pinned` class name isn't correctly applied (or we remove it too early) in BFAA configuration: https://github.com/Wikia/app/blob/dev/extensions/wikia/AdEngine/src/templates/big-fancy-ad-above-desktop-config.js#L75-L75

Please check if we can fix it there before we decide to add more rules here.

kvas-damian: After further analysis we want to block global nav hiding in 2 scenarios:
1. ad is sticky and global nav is just below the ad
2. ad is unsticked - global nav should stop under global navigation.

`bfaa-pinned` covers the 2nd scenario. We can't easily use the same class for 1. because we use different positioning method. This makes me feel we need a second flag that covers 1.

Having that said, I don't see better solution for now.

kvas-damian (approved):

PR #14979, opened 2018-03-12
SUS-4308 | Sunset AppPromoLanding extension

mszabo-wikia:
@aniaMu confirms it is safe to remove

https://wikia-inc.atlassian.net/browse/SUS-4308


extensions/wikia/AppPromoLanding/crowdin.conf
```diff
@@ -1,5 +0,0 @@
-[crowdin.project]
-projectid = "ext-wikia-AppPromoLanding"
```
mrmike, line 2: Are we using this crowding project somewhere else? Should we remove it from crowdin as well?

mszabo-wikia: yes it's safe to be removed from crowdin, I will notify Keiko about it

macbre (approved):

PR #14987, opened 2018-03-13
SUS-4250 | Sunset DevBoxPanel extension

mszabo-wikia:
https://wikia-inc.atlassian.net/browse/SUS-4250


macbre (approved):
We finally made it!

PR #14988, opened 2018-03-14
Quick typo fixes

penguinstyles:


KockaAdmiralac:
I think i18n changes should go through the translation team so they can be properly updated in Crowdin

mszabo-wikia:
I believe it should be fine for EN source strings (but not their translations), as the source strings can easily be synced. This is not the case for translations though, translation errors should indeed be reported.

mszabo-wikia:
...and this extension does not even have a Crowdin project.

PR #14998, opened 2018-03-15
SUS-4325 | CloseWikiMaintenance should remove directories with images after tar file is created

macbre:
Otherwise we end up with 200+ directories in `/tmp/images` on `cron-s1` and disk space alerts when a large wiki is closed and images are dumped to s3.

Or we even may end up with such directory in /tmp/images:

```
11G    ./teentitansgo
```

tar file with compressed content of this directory weights ~10 GiB. This pair consumes 50% of cron-s1 disk space...

### DIFF
--- a/extensions/wikia/WikiFactory/Close/maintenance.php
+++ b/extensions/wikia/WikiFactory/Close/maintenance.php
@@ -169,7 +169,7 @@ class CloseWikiMaintenance {
 				if( $dbname && $folder ) {
 					$this->log( "Dumping images on remote host" );
 					try {
-						$source = $this->tarFiles( $folder, $dbname, $cityid );
+						$source = $this->tarFiles( $dbname, $cityid );
 
 						if( is_string( $source ) ) {
 							$retval = DumpsOnDemand::putToAmazonS3( $source, !$hide,  MimeMagic::singleton()->guessMimeType( $source ) );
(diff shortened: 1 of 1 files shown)


drozdo:
LGTM

extensions/wikia/WikiFactory/Close/maintenance.php
```diff
@@ -378,6 +378,11 @@ private function tarFiles( $directory, $dbname, $cityid ) {
 			$this->log( "List of files in {$directory} is empty" );
 			throw new WikiaException( "List of files in {$directory} is empty" );
 		}
+
+		// SUS-4325 | CloseWikiMaintenance should remove directories with images after tar file is created
+		$this->log( "Removing '{$directory}' directory" );
```
sqreek, line 387: from `Tar->create()` `The method return false and a PEAR error text`
please Justin Case™  check is $result false

macbre: Good point - see fb286a6

extensions/wikia/WikiFactory/Close/maintenance.php
```diff
@@ -310,14 +310,13 @@ public function execute() {
...
 	 * @access public
 	 *
-	 * @param string $uploadDirectory path to images
 	 * @param string $dbname database name
 	 * @param int $cityid city ID
 	 *
 	 * @return string path to created archive or false if there are no files to backup (S3 bucket does not exist / is empty)
 	 * @throws WikiaException thrown on failed backups
 	 */
-	private function tarFiles( $directory, $dbname, $cityid ) {
```
macbre, line 320: This value is set in line 332

sqreek (approved):

PR #15001, opened 2018-03-15
SUS-4320 Removes INSERT IGNORE option in Block::insert()

leshniak:
https://wikia-inc.atlassian.net/browse/SUS-4320

### DIFF
--- a/includes/Block.php
+++ b/includes/Block.php
@@ -445,8 +445,7 @@ class Block {
 		$dbw->insert(
 			'ipblocks',
 			$row,
-			__METHOD__,
-			array( 'IGNORE' )
+			__METHOD__
 		);
 		$affected = $dbw->affectedRows();
 		$this->mId = $dbw->insertId();
--- a/includes/Block.php
+++ b/includes/Block.php
@@ -442,17 +442,22 @@ class Block {
 		$row = $this->getDatabaseArray();
 		$row['ipb_id'] = $dbw->nextSequenceValue("ipblocks_ipb_id_seq");
 
-		$dbw->insert(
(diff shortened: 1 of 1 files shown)


includes/Block.php
```diff
@@ -445,8 +445,7 @@ public function insert( $dbw = null ) {
 		$dbw->insert(
 			'ipblocks',
 			$row,
-			__METHOD__,
-			array( 'IGNORE' )
```
macbre: :+1:

macbre (approved):

mszabo-wikia (approved):
Callers expect this method to not throw an exception in case of conflict—let's preserve this guarantee by catching query errors and returning false. The `DBError` constructor will log it for us.

mszabo-wikia (changes requested):
☝️ above was meant to be changes requested

includes/Block.php
```diff
@@ -442,17 +442,22 @@ public function insert( $dbw = null ) {
...
 
-		if ( $affected ) {
-			$auto_ipd_ids = $this->doRetroactiveAutoblock();
-			return array( 'id' => $this->mId, 'autoIds' => $auto_ipd_ids );
+			if ( $affected ) {
+				$auto_ipd_ids = $this->doRetroactiveAutoblock();
+				return array( 'id' => $this->mId, 'autoIds' => $auto_ipd_ids );
+			}
+		}
+		catch ( DBError $ex ) {
```
mszabo-wikia: let's align the braces otherwise 👍

mszabo-wikia (approved):

PR #15015, opened 2018-03-19
(PLATFORM-3324) Make wgWikiaBarConfig use protocol-relative links

Grunny:
Make wgWikiaBarConfig use protocol-relative links when possible.

/cc @Wikia/core-platform-team


Grunny:
You can ignore whitespace changes to see the fix properly: https://github.com/Wikia/app/pull/15015/files?w=1

extensions/wikia/WikiaBar/models/WikiaBarDataFailsafeModel.class.php
```diff
@@ -30,71 +30,71 @@ class WikiaBarDataFailsafeModel extends WikiaBarModelBase {
...
 		),
 		'de' => array(
 			WikiFactoryHub::CATEGORY_ID_GAMING =>
 			'button-1-class=cup
 button-1-text=Videospiele
-button-1-href=http://de.wikia.com/Videospiele
+button-1-href=//de.wikia.com/Videospiele
 button-2-class=cup
 button-2-text=Herbstauswahl
 button-2-href=http://de.community.wikia.com/wiki/Benutzer_Blog:Foppes/Herbstauswahl_2012
```
harnash, line 100: We leave this our because of the language prefix. Right?

harnash (approved):

extensions/wikia/WikiaBar/models/WikiaBarDataFailsafeModel.class.php
Grunny: Yep, these can't support HTTPS yet, so making them relative would break the links if you saw them on say de.wikia.com which would support HTTPS.

PR #15017, opened 2018-03-20
SUS-4274 | WikiMetrics class is gone since #14906

macbre:
See #14906


kvas-damian (changes requested):
Please remove the entry point as well: https://github.com/Wikia/app/blob/b1e7e9cb035b64db4c55b7ecad0d731da08a92d8/extensions/wikia/WikiFactory/templates/selector.tmpl.php#L53-L53

kvas-damian (approved):

PR #15022, opened 2018-03-20
Dummy commit to trigger tests

ludwikkazmierczak:


hakubo:
FYI if you go to
https://github.com/Wikia/app/settings/hooks/1769902
you can retrigger webhooks so you don't have to do empty commits to test stuff :)
might come handy some day

PR #15028, opened 2018-03-20
[WiP] SUS-1941 | Allow using UTF-8 connection for MySQl shared DBs

kvas-damian:


mszabo-wikia (changes requested):
this will use `utf8`but we want `utf8mb4`.

PR #15030, opened 2018-03-20
Sus 4292 https fix

sqreek:
https://wikia-inc.atlassian.net/browse/SUS-4292

@Wikia/sus


extensions/wikia/Search/classes/MediaWikiService.php
```diff
@@ -950,7 +950,9 @@ public function isPageIdMainPage( $pageId ) {
 	 * @return string
 	 */
 	public function getHostName() {
-		return substr( $this->getGlobal( 'Server' ), 7 );
+		preg_match( "^https?:\/\/", $this->getGlobal( 'Server' ), $matches );
+
+		return $matches[1];
```
kvas-damian: I don't see any grouping in the pattern. How do you get any value in `$matches`?
Please add some unit tests to cover such issues.

extensions/wikia/WikiFactory/Close/SpecialCloseWiki_body.php
```diff
@@ -230,8 +230,10 @@ private function doProcess() {
 	}
 
 	private function prefixMainDomain( $cityId ) {
-		$mainDomain = substr( WikiFactory::getVarValueByName("wgServer", $cityId), 7 );
-		if(!empty($mainDomain)) {
+		//local prefix to not accidentally mix it with global wgServer
+		$localWgServer = WikiFactory::getVarValueByName( "wgServer", $cityId );
+		if ( preg_match( "^https?:\/\/", $localWgServer, $matches ) ) {
+			$mainDomain = $matches[1];
```
kvas-damian: The same problem with regex. Please consider having that logic in one place instead of duplicating problems ;)

extensions/wikia/WikiFactory/WikiFactory.php
```diff
@@ -401,18 +402,22 @@ static public function setmainDomain ( $city_id, $domain = null, $reason = null
...
+		$newWgServer = WikiFactory::addProtocolIfNotExists( $domain );
+		$retVal = WikiFactory::setVarByName( "wgServer", $city_id, $newWgServer, $reason );
 
 		static::clearDomainCache( $city_id );
 
 		return $retVal;
 	}
 
+	private static function addProtocolIfNotExists( $domain ) {
+		if ( !preg_match( "^https?:\/\/", $domain, $matches ) ) {
```
kvas-damian: `$matches` is optional. You don't need to pass it when not using.

kvas-damian (changes requested):
Please, correct issues.

macbre (changes requested):
I'd add a helper (or maybe one already exists) that extracts the domain out of the full URL. And unit test it.

sqreek:
fixed issues
@macbre I will left helper for @Wikia/core-platform-team

extensions/wikia/WikiFactory/WikiFactory.php
```diff
@@ -401,18 +402,22 @@ static public function setmainDomain ( $city_id, $domain = null, $reason = null
...
+		$retVal = WikiFactory::setVarByName( "wgServer", $city_id, $newWgServer, $reason );
 
 		static::clearDomainCache( $city_id );
 
 		return $retVal;
 	}
 
+	private static function addProtocolIfNotExists( $domain ) {
+		if ( !preg_match( "^https?:\/\/", $domain ) ) {
+			$domain = static::DOMAIN_PROTOCOL . '://' . $domain;
```
harnash: @Wikia/core-platform-team we need to add logic here: `wfHttpsAllowedForURL()`

harnash: It's late. I will add this first thing in the morning tomorrow.

extensions/wikia/Search/classes/MediaWikiService.php
```diff
@@ -950,7 +950,9 @@ public function isPageIdMainPage( $pageId ) {
 	 * @return string
 	 */
 	public function getHostName() {
-		return substr( $this->getGlobal( 'Server' ), 7 );
+		preg_match( "/^https?:\/\/(.*)/", $this->getGlobal( 'Server' ), $matches );
```
harnash: I'm not 100% sure but we really need this? How about http://php.net/manual/en/function.parse-url.php ?

harnash: Additionally this greedy capture will catch everything after protocol - even trailing `\` if there's one.

harnash: OK. Regex seems a bit faster if we do hostname resolution only :(

```
harnas@dev-harnas:~$ time for a in {1..100}; do php regex.php; done

real	0m5.539s
user	0m1.692s
sys	0m3.192s
harnas@dev-harnas:~$ time for a in {1..100}; do php parse_url.php; done

real	0m8.348s
user	0m1.732s
sys	0m5.988s
```

extensions/wikia/WikiFactory/WikiFactory.php
```diff
@@ -401,18 +402,22 @@ static public function setmainDomain ( $city_id, $domain = null, $reason = null
 			Wikia::log( __METHOD__, "", "WikiFactory is not used." );
 			return false;
 		}
-
-		if ( 'http://' != strpos($domain, 0, 7) ) {
-			$domain = 'http://' . $domain;
-		}
-
-		$retVal = WikiFactory::setVarByName("wgServer", $city_id, $domain, $reason);
+		$newWgServer = WikiFactory::addProtocolIfNotExists( $domain );
```
kvas-damian: Can we use `wfExpandUrl()` https://github.com/Wikia/app/blob/a24cfb9e4e2219a510cdc22a50872a157827a1a3/includes/GlobalFunctions.php#L507-L507 instead?

Also, I'd suggest pinging Platform Team to make sure we have shared vision on how we deal with protocols.

extensions/wikia/WikiFactory/WikiFactory.php
harnash: This function would look something like this:
```php
private static function addProtocolIfNotExists( $domain ) {
	if ( !preg_match( "^https?:\/\/", $domain ) ) {
		if ( wfHttpsAllowedForURL($domain) ) {
			$proto = PROTO_HTTPS;
		} else {
			$proto = PROTO_HTTP;
		}
		$domain = $proto . $domain;
	}

	return $domain;
}
```
ping: @Grunny

harnash (changes requested):

extensions/wikia/WikiFactory/WikiFactory.php
Grunny: I don't think we'd want to have the main domain set as HTTPS for any wiki right now, as that would affect it for anons. So, I think it always being HTTP, unless we explicitly state HTTPS is probably better at the moment. We already handle overriding the server value to HTTP when the request is allowed to be HTTPS elsewhere. We could do something similar to the above once we were ready to allow anons to see HTTPS, but not yet.

extensions/wikia/Search/classes/MediaWikiService.php
```diff
@@ -950,7 +950,8 @@ public function isPageIdMainPage( $pageId ) {
 	 * @return string
 	 */
 	public function getHostName() {
-		return substr( $this->getGlobal( 'Server' ), 7 );
+		preg_match( "/^https?:\/\/([\w\-\.]+)/", $this->getGlobal( 'Server' ), $matches );
```
kvas-damian: I believe the regex would be easier if we use `preg_replace`.
Also I remember some talks about support for language path for https migration. Is it still valid?

extensions/wikia/WikiFactory/Close/SpecialCloseWiki_body.php
```diff
@@ -230,8 +230,10 @@ private function doProcess() {
 	}
 
 	private function prefixMainDomain( $cityId ) {
-		$mainDomain = substr( WikiFactory::getVarValueByName("wgServer", $cityId), 7 );
-		if(!empty($mainDomain)) {
+		//local prefix to not accidentally mix it with global wgServer
+		$localWgServer = WikiFactory::getVarValueByName( "wgServer", $cityId );
+		if ( preg_match( "/^https?:\/\/([\w\-\.]+)/", $localWgServer, $matches ) ) {
```
kvas-damian: I thought we already talked about DRY ;)

extensions/wikia/Search/classes/MediaWikiService.php
sqreek: wgServer will not contain language path

extensions/wikia/WikiFactory/Close/SpecialCloseWiki_body.php
sqreek: preg_match is faster

harnash (approved):

macbre (approved):

PR #15033, opened 2018-03-20
XF-248 | Cleanup after XW-4654

rybmat:
https://wikia-inc.atlassian.net/browse/XF-248

@Wikia/x-wing


latata (approved):

extensions/wikia/PortableInfobox/services/Helpers/PortableInfoboxImagesHelper.php
```diff
@@ -63,20 +63,16 @@ public function extendImageData( $data, $thumbnailFileWidth, $thumbnailImgTagWid
 
 		return array_merge( $data, [
 			'ref' => $ref,
-			'height' => $imgTagDimensions['height'],
-			'width' => $imgTagDimensions['width'],
+			'height' => intval( $imgTagDimensions['height'] ),
```
hakubo, line 66: are we doing some math operations on it?
if not, I'd send it as string

hakubo (approved):

extensions/wikia/PortableInfobox/services/Helpers/PortableInfoboxImagesHelper.php
rybmat: I did it because it started to return float values. It is used only as width and height attributes in html

PR #15035, opened 2018-03-21
SUS-4181 | CreateNewWiki - do not set wgDBname and wgDBcluster WikiFactory variables

macbre:
Values from `city_list` table row are now used instead.

https://wikia-inc.atlassian.net/browse/SUS-4181

http://pl.sus4181.sandbox-s6.wikia.com/wiki/Specjalna:Wersja has been properly created with this code.

### Cleanup

Remove `WikiFactory::getFileCachePath` method

### DIFF
--- a/extensions/wikia/WikiFactory/WikiFactory.php
+++ b/extensions/wikia/WikiFactory/WikiFactory.php
@@ -798,29 +798,6 @@ class WikiFactory {
 						__METHOD__ );
 					break;
 
-				case "wgDBname":
-					#--- city_dbname
-					$dbw->update(
-						static::table("city_list"),
-						[ "city_dbname" => $value ],
-						[ "city_id" => $city_id ],
-						__METHOD__ );
-					break;
-
-				case "wgDBcluster":
-					/**
-					 * city_cluster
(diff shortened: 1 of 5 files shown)


extensions/wikia/WikiFactory/WikiFactory.php
```diff
@@ -798,29 +798,6 @@ static public function setVarById( $cv_variable_id, $city_id, $value, $reason=nu
...
-					 * city_cluster = null for first cluster
-					 * @todo handle deleting values of this variable
-					 */
-					$dbw->update(
-						static::table("city_list"),
-						[ "city_cluster" => $value ],
-						[ "city_id" => $city_id ],
-						__METHOD__ );
-					break;
-
```
macbre, line 823: These WikiFactory variables will be removed from `city_variables_pool` table when this goes live.

These `city_list` fields will be used instead:

```sql
  `city_dbname` varchar(64) NOT NULL DEFAULT 'notreal',
  `city_cluster` varchar(255) DEFAULT NULL,
```

kvas-damian (approved):

PR #15037, opened 2018-03-21
SUS-4341 | AvatarService - handle "avatar" URLs with scaling parameters stored

macbre:
https://wikia-inc.atlassian.net/browse/SUS-4341

### DIFF
--- a/includes/wikia/services/AvatarService.class.php
+++ b/includes/wikia/services/AvatarService.class.php
@@ -245,7 +245,7 @@ class AvatarService {
 	 * @param Masthead $masthead
 	 * @param int $width
 	 * @param $timestamp
-	 * @return \Wikia\Vignette\UrlGenerator
+	 * @return string
 	 */
 	public static function getVignetteUrl( Masthead $masthead, $width, $timestamp ) {
 		$relativePath = $masthead->mUser->getGlobalAttribute( AVATAR_USER_OPTION_NAME );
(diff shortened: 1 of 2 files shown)


idradm (approved):

includes/wikia/services/AvatarService.class.php
```diff
@@ -267,16 +267,18 @@ public static function getVignetteUrl( Masthead $masthead, $width, $timestamp )
...
 
-		$url = $relativePath;
 		if ( !preg_match( '/^https?:\/\//', $relativePath ) ) {
 			$bucket = VignetteRequest::parseBucket( "{$wgVignetteUrl}/common/avatars" );
 			$relativePath = ltrim( $relativePath, '/' );
 			$url = self::buildVignetteUrl( $width, $bucket, $relativePath, $timestamp );
 		}
 		else {
+			// SUS-4341 | something is setting "avatar" URLs with scaling parameters, remove them
+			$relativePath = preg_replace( '#(/revision/latest)?/scale-to-width-down/\d+$#', '', $relativePath );
```
kvas-damian, line 280: Did you consider using `VignetteUrlToUrlGenerator` https://github.com/Wikia/app/blob/730c673fc590952adba60f0b0d56f2e2c6cdd6a4/includes/wikia/vignette/VignetteUrlToUrlGenerator.php#L9-L9 instead of regexing? I'm not sure if we are able to handle this case, but I'd prefer to not introduce logic related to vignette URL parsing in the other places.

macbre: It's a "hack" for broken avatar URLs. That's why I wanted to keep handling of this specific case here. PLATFORM or SERVICES will take care of migrating the data, so that we will have canonical avatar URLs only.

mszabo-wikia (approved):

PR #15048, opened 2018-03-22
PLATFORM-3401: fix test for GlobalTitle

harnash:
This should fix the tests and make it more reliable on DEV.

ping: @Wikia/core-platform-team, @mszabo-wikia

### DIFF
--- a/includes/wikia/GlobalFunctions.php
+++ b/includes/wikia/GlobalFunctions.php
@@ -1562,14 +1562,13 @@ function wfHttpsToHttp( $url ) {
 }
 
 function wfHttpsAllowedForURL( $url ): bool {
-	global $wgWikiaBaseDomain, $wgDevDomain, $wgWikiaEnvironment;
-	$parsedURL = parse_url( $url );
-	if ( $parsedURL === false ) {
+	global $wgWikiaBaseDomain, $wgDevDomain, $wgWikiaEnvironment, $wgDevelEnvironment;
+	$host = parse_url( $url, PHP_URL_HOST );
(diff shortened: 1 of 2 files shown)


includes/wikia/GlobalFunctions.php
```diff
@@ -1562,14 +1562,13 @@ function wfHttpsToHttp( $url ) {
...
-	if ( $parsedURL === false ) {
+	global $wgWikiaBaseDomain, $wgDevDomain, $wgWikiaEnvironment, $wgDevelEnvironment;
+	$host = parse_url( $url, PHP_URL_HOST );
+	if ( $host === false ) {
 		return false;
 	}
 
-	$host = $parsedURL['host'];
-	if ( !empty( $wgDevDomain ) ) {
+	if ( $wgDevelEnvironment && !empty( $wgDevDomain ) ) {
```
Grunny, line 1571: Do we need to check both?

Grunny (approved):
One question, otherwise looks good.

includes/wikia/GlobalFunctions.php
harnash: Probably don't but I'm not 100% sure that those two are set always

Grunny: Both are set at the same time: https://github.com/Wikia/config/blob/dev/LocalSettings.php#L19-L21 but doesn't hurt I suppose. :)

harnash: Especially on tests when they are being mocked out

PR #15059, opened 2018-03-23
SUS-4352: MWF - don't query dataware.pages for a lower case title

mszabo-wikia:
The column contains titles in standard MediaWiki format (first letter capitalized).

https://wikia-inc.atlassian.net/browse/SUS-4352

### DIFF
--- a/extensions/wikia/MultiTasks/SpecialMultiWikiFinder_body.php
+++ b/extensions/wikia/MultiTasks/SpecialMultiWikiFinder_body.php
@@ -23,6 +23,7 @@ class MultiwikifinderSpecialPage extends SpecialPage {
             list( $limit, $offset ) = wfCheckLimits();
         }
         $this->mfp = new MultiwikifinderPage($this->mName, $page, $limit, $offset);
+		$this->mfp->setContext( $this->getContext() );
 
 		if (!empty($show)) {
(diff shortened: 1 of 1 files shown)


extensions/wikia/MultiTasks/SpecialMultiWikiFinder_body.php
```diff
@@ -122,7 +122,7 @@ function showResults( ) {
 				'pages',
 				array( 'page_latest, page_wikia_id, page_id, page_is_redirect' ),
 				array(
-					'page_title'		=> mb_strtolower($this->mPageTitle),
+					'page_title'		=> mb_strtoupper( mb_substr( $this->mPageTitle, 0, 1 ) ) . mb_substr( $this->mPageTitle, 1 ),
```
macbre: Maybe let's use `Language::ucfirst` instead?

```
> echo $wgLang->ucfirst('foo')
Foo
```

macbre (changes requested):

extensions/wikia/MultiTasks/SpecialMultiWikiFinder_body.php
mszabo-wikia: We don't know what language the input string is in.

PR #15081, opened 2018-03-27
XW-4312 | Don't send request to lyricfind api for mw bots and search crawlers

rybmat:
https://wikia-inc.atlassian.net/browse/XW-4312

@Wikia/x-wing


extensions/3rdparty/LyricWiki/LyricFind/LyricFindHooks.class.php
```diff
@@ -54,7 +58,7 @@ static public function onAlternateEdit(EditPage $editPage) {
 					$blockEdit = true;
 				}
 			}
-		} else {
+		} elseif ( !$isCrawler ) {
```
rybmat: TBH I'm not sure if we should go with this check, it will allow bots to create pages that are prohibited by LyricFind (whatever that means). @hakubo what do you think?

bkoval (approved):

extensions/3rdparty/LyricWiki/LyricFind/LyricFindTrackingService.class.php
```diff
@@ -251,6 +251,29 @@ public static function isPageBlockedViaApi($amgId="", $gracenoteId="", $pageTitl
...
+		$crawlers = [
+			'bot',
+			'slurp',
+			'crawler',
+			'spider',
+			'curl',
+			'facebook',
+			'fetch',
+			'ruby',
+			'python'
```
frankfarmer: The top UA's we're being hit by are things like:

```
perl/lyrics::fetcher::lyricwiki 0.10
mozilla/5.0 (windows; u; windows nt 5.1; en-us; rv:1.8.1.6) gecko/20070725 firefox/2.0.0.6
mozilla/5.0 (x11; linux x86_64) applewebkit/535.19 (khtml, like gecko) chrome/18.0.1025.142 safari/535.19
```

(note that the last two appear to be some sort of scraper attempting to disguise itself as ancient versions of FF/chrome)

Trying to blacklist these by UA seems fragile

frankfarmer:
Most of the problematic requests we're seeing are in `isPageBlockedViaApi()`, not `track()`.

extensions/3rdparty/LyricWiki/LyricFind/LyricFindTrackingService.class.php
rybmat: 👍

rybmat:
@frankfarmer `isPageBlockedViaApi()` sends request to lyricfind api in order to check if a new page can be created (I'm not sure why lyricfind would block us from creating a page on lyric wiki, but still it was done this way for some reason). If we don't send this request for bots, we'll need to either block creating pages on this wiki for bots or allow users / bots to create pages that would be prohibited by lyricfind if they set user agent to value we recognise as bot's ua. Our proposal is to go with this change, check how much it will help and then make a decision if we should change the logic in `isPageBlockedViaApi()`.

latata (approved):

frankfarmer:
I spent quite a bit of time analyzing our data.  We're not sending a significant number of track() requests for non-human traffic -- javascript initiates those requests, so there's no reason for most "bots" to hit the url that eventually calls track().  But we are sending hundreds of thousands of isPageBlockedViaApi() requests for non-human traffic, because those are generated by requests for urls like `lyrics.wikia.com/index.php?action=edit&title=Paul_Young:The_Tracks_Of_My_Tears` which non-javascript clients request frequently.

Note, I think there's some ambiguity regarding the term "bot" here.  The non-human traffic does not appear to be search engine spiders nor mediawiki bots.  It's dubious scrapers operating out of AWS with spoofed user agents, and tools like `perl/lyrics::fetcher::lyricwiki`, some of which are fetching lyric data for display in MP3 players and comparable software.

rybmat:
@frankfarmer I added back check for `?action=edit`, but out of curiosity, does these bots operating from AWS create new pages on lyric wiki? Because this is the only way to invoke `isPageBlockedByApi()`. For `?action=edit` on existing article `isPageBlockedViaApi` is not called and request to LyricFind is not sent

frankfarmer:
They don't actually create pages.  In some cases, AWS spiders may follow red links.   In the case of `perl/lyrics::fetcher::lyricwiki`, I suspect they use ?action=edit as a clumsy way to extract article source markup.  Additionally, `perl/lyrics::fetcher::lyricwiki` may speculatively fetch these urls based on mp3 id3 tags in users' media players, and things like that -- so if a user plays a song, their media player might request `/index.php?action=edit&title=Paul_Young:The_Tracks_Of_My_Tears` as an attempt to get wikitext lyrics for the currently playing song

extensions/3rdparty/LyricWiki/LyricFind/LyricFindHooks.class.php
```diff
@@ -111,4 +114,29 @@ static public function onParserBeforeStrip(Parser $parser, &$text, &$strip_state
...
+			'crawler',
+			'spider',
+			'curl',
+			'facebook',
+			'fetch',
+			'ruby',
+			'python',
+			'perl/lyrics::fetcher::lyricwiki',
+			'mozilla/5.0 (windows; u; windows nt 5.1; en-us; rv:1.8.1.6) gecko/20070725 firefox/2.0.0.6',
+			'mozilla/5.0 (x11; linux x86_64) applewebkit/535.19 (khtml, like gecko) chrome/18.0.1025.142 safari/535.19'
```
frankfarmer, line 131: Who will monitor and maintain this blacklist? 😕 

Ultimately, I think the long term solution is to do something like move the isPageBlockedViaApi() call to a javascript-triggered beacon (or something along those lines), and/or when an edit form is actually submitted with new markup, so only users who actually attempt an edit end up triggering the request.

rybmat: I agree, however before we do any change to this logic I'd like to know what are the reasons of asking LyricFind if we can create page with given title. Do you know somebody who could remember this? Or maybe some articles in Confluence? Maybe we could get rid of this call to api from `isPageBlockedViaApi()`.

frankfarmer: `isPageBlockedViaApi` checks for a 206 response code.  According to [the docs](http://www.lyricfind.com/documentation/?key=eyJkb2N1bWVudCI6IndlYl9zZXJ2aWNlIiwib3B0aW9ucyI6eyJwcmVtaXVtIjoiZmFsc2UiLCJvbGRrZXl0eXBlcyI6ImZhbHNlIn19#the-lyric-display-api) that means:

> Lyric is not available because of a publisher specific block. No data is returned. 

If a user attempts to create a new lyric page, and the API returns this code, we have no license to host those lyrics on our site, and thus there's a potential for copyright infringement.

There's still an open question as to whether or not we can perform this check without it being tracked as a "view" which we must pay for.

PR #15093, opened 2018-03-28
PLATFORM-3429: fix protocol for redirects to main wiki domain

harnash:
Keep protocol the same when redirecting to main Wiki domain.


extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
```diff
@@ -427,6 +427,9 @@ public function execute() {
 
 		if( ( $cond1 || $cond2 ) && empty( $wgDevelEnvironment ) ) {
 			$redirectUrl = WikiFactory::getLocalEnvURL( $this->mCityUrl );
+			if ( !empty($_COOKIE[\Wikia\Service\User\Auth\CookieHelper::ACCESS_TOKEN_COOKIE_NAME]) && $_SERVER['HTTP_FASTLY_SSL'] && wfHttpsAllowedForURL( $redirectUrl ) ) {
```
Grunny: Minor: This should be spacey and split with one condition per line per our coding conventions.

extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
```diff
@@ -443,6 +446,7 @@ public function execute() {
 			}
 
 			header( "X-Redirected-By-WF: NotPrimary" );
+			header( "Vary: Cookie");
```
Grunny: One last question here is how long Fastly is caching this? I'm mostly thinking if we should make sure the cache is truly private for logged-in users so we don't cache a lot of versions for each access_token. If it's a very short cache, it's probably not a big concern at this point given the redirect URLs are probably relatively rare to be clicked by many logged-in users.

Grunny (commented):
Looks good in general, just one question about the cache.

extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
harnash: Fix in: cec76f99c01

extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
harnash: ping: @ljagiello could you check this one?

harnash: After offsite talk we decided to vary on source protocol (bda16cdff15) instead of `Cookie` (which is a bad idea). It should cover most of our cases.

Grunny: We already vary on source protocol on the Fastly level: https://wikia-inc.atlassian.net/browse/OPS-11119 so this should be unnecessary in theory.

The reason to vary on cookie is because the behaviour is different depending on if you're logged in or anonymous. Anons should go HTTPS -> HTTP, and logged in users should go HTTPS -> HTTPS (but they currently have two redirects for aliases: HTTPS -> HTTP -> HTTPS). If this redirect is cached and we don't vary on cookie/logged-in, we may move the double redirect to anons in some cases, i.e. if they request the alias over HTTPS and the redirect is currently cached, they will be redirected HTTPS -> HTTPS -> HTTP. It also means [this check](https://github.com/Wikia/app/blob/bda16cdff1580/extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php#L430) wouldn't work very well.

Providing we make sure the logged-in request/redirect is not cached, there should only be two variations of the request for Fastly to handle (anons who have the cookie header stripped because they don't have any of our auth-related cookies set, and logged-in users for whom the cache should be private (so no versions should be stored in Fastly)).

If this is not possible, and the redirect is being cached on Fastly, I think we should consider abandoning this, as it's probably better to have a slightly broken redirect experience for logged-in users (and get them to update their imports and other things that can break because of this right now), than potentially start giving double redirects to any HTTPS URLs anons (as it includes Google and other crawlers) may encounter, even if it's relatively unlikely right now (though it will become more likely as logged-in users opt-in and start pasting HTTPS URLs in various places).

What do you think?

ljagiello: Just to keep a record, Fastly makes an exception for us and allow more Vary compare to regular clients but at the same time they ask us multiple times to reduce number of variations. It's a problem on their side when we have many variations of the same object and cause instability of cache. In past we deployed few times Vary based on cookie and we end up with unstable Fastly cache (it cause locks in Fastly).

harnash:
@Grunny hopefully everything should be addressed now.

extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
```diff
@@ -427,6 +427,14 @@ public function execute() {
 
 		if( ( $cond1 || $cond2 ) && empty( $wgDevelEnvironment ) ) {
 			$redirectUrl = WikiFactory::getLocalEnvURL( $this->mCityUrl );
+			$hasAuthCookie = !empty( $_COOKIE[\Wikia\Service\User\Auth\CookieHelper::ACCESS_TOKEN_COOKIE_NAME] );
```
Grunny: Just incase, I'd maybe include a check here for any of the cookies we check for on VCL: `(wikicities_session|wikicitiesUserID|UserName|Token|LoggedOut|access_token)`, which is also similar to what we check for elsewhere in MW, i.e. https://github.com/Wikia/app/blob/da75702/includes/OutputPage.php#L1766-L1773

Grunny (changes requested):
Last comment, otherwise looks good.

extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
harnash: commit: dbe66c04017

Grunny (approved):

PR #15099, opened 2018-03-29
ADEN-6957 No call for BLB if MR is in the same viewport

dmnsobczak:


resources/wikia/modules/viewportObserver.js
```diff
@@ -15,6 +16,34 @@ define('wikia.viewportObserver', [
...
+					found = true;
+				}
+			} else {
+				if (elementOffset - otherOffset - otherHeight < windowHeight) {
+					found = true;
+				}
+			}
+		});
+
+		return found;
```
v3rth: I think you need [some](https://www.w3schools.com/jsref/jsref_some.asp) function :)

extensions/wikia/AdEngine/js/slot/bottomLeaderboard.js
```diff
@@ -26,8 +39,22 @@ define('ext.wikia.adEngine.slot.bottomLeaderboard', [
 			if (!pushed && pushPos < scrollPosition) {
 				pushed = true;
 				doc.removeEventListener('scroll', pushSlot);
-				win.adslots2.push(slotName);
-				log(['pushSlot', 'Pushed ' + slotName], 'debug', logGroup);
+
+				if (adContext.get('opts.isBLBViewportEnabled') &&
+					doc.getElementsByTagName('body')[0].className.indexOf('uap-skin') === -1 &&
```
v3rth: I think that part could be behind some "fancy" function

extensions/wikia/AdEngine/js/slot/bottomLeaderboard.js
```diff
@@ -26,8 +39,22 @@ define('ext.wikia.adEngine.slot.bottomLeaderboard', [
...
+					viewportObserver.sameViewport(slotName, breakSlots)) {
```
v3rth: ... and this whole condition also could be extracted, right now it is hard to fast "guess" whats going on here

resources/wikia/modules/viewportObserver.js
```diff
@@ -15,6 +16,34 @@ define('wikia.viewportObserver', [
...
+			elementHeight = element.offsetHeight,
+			found = false;
+
+		otherElements.forEach(function (other) {
+			other = doc.getElementById(other);
+
+			var otherOffset = domCalculator.getTopOffset(other),
+				otherHeight = other.offsetHeight;
+
+			if (elementOffset < otherOffset) {
```
v3rth: I think you can easily reduce the number of ifs by extracting ```otherOffset - elementOffset - elementHeigh```
and ``` elementOffset - otherOffset - otherHeight ``` to variable (even into to ternary operator, but then it might be a little bit messy).

v3rth (changes requested):

extensions/wikia/AdEngine/js/slot/bottomLeaderboard.js
dmnsobczak: Good idea, refactored.

extensions/wikia/AdEngine/js/slot/bottomLeaderboard.js
dmnsobczak: Yep, refactored too.

resources/wikia/modules/viewportObserver.js
dmnsobczak: I made some cleanup :)

resources/wikia/modules/viewportObserver.js
dmnsobczak: I did this different way :)

extensions/wikia/AdEngine/js/slot/bottomLeaderboard.js
```diff
@@ -26,11 +39,35 @@ define('ext.wikia.adEngine.slot.bottomLeaderboard', [
...
+			log(['pushSlot', 'Same viewport ' + slotName], 'debug', logGroup);
+
+			return true;
+		}
+
+		return false;
+	}
+
+	function isUap() {
+		return doc.getElementsByTagName('body')[0].className.indexOf('uap-skin') === -1;
```
fraszczakszymon: Use uapContext please

extensions/wikia/AdEngine/js/tracking/adInfoListener.js
```diff
@@ -105,13 +105,11 @@ define('ext.wikia.adEngine.tracking.adInfoListener',  [
...
 
-	function shouldHandleSlot(slot) {
+	function shouldHandleSlot(slot, status) {
 		var dataGptDiv = slot.container && slot.container.firstChild;
 
 		return (
-			enabledSlots[slot.name] &&
-			dataGptDiv &&
-			dataGptDiv.dataset.gptPageParams
+			enabledSlots[slot.name] && (status === 'viewport' || (dataGptDiv && dataGptDiv.dataset.gptPageParams))
```
fraszczakszymon: It won't be needed if we decide to address other comments

extensions/wikia/AdEngine/js/slot/bottomLeaderboard.js
```diff
@@ -1,14 +1,27 @@
...
+	dom,
+	log,
+	throttle,
+	viewportObserver,
+	win
+) {
 	'use strict';
 
 	var slotName = 'BOTTOM_LEADERBOARD',
+		breakSlots = ['TOP_RIGHT_BOXAD'],
```
fraszczakszymon: What about creating some module that would have some kind of definition of slots that conflicts with each other. In case of some conflict it may update information in the slotContext so slot would be disabled in all features automatically.

extensions/wikia/AdEngine/js/slot/bottomLeaderboard.js
```diff
@@ -26,11 +39,35 @@ define('ext.wikia.adEngine.slot.bottomLeaderboard', [
 			if (!pushed && pushPos < scrollPosition) {
 				pushed = true;
 				doc.removeEventListener('scroll', pushSlot);
-				win.adslots2.push(slotName);
-				log(['pushSlot', 'Pushed ' + slotName], 'debug', logGroup);
+
+				if (!collapseOnSameViewport()) {
+					win.adslots2.push(slotName);
```
fraszczakszymon: What about refactoring AdEngine queue a bit, so you can pass all needed informations (maybe full AdSlot object)?

extensions/wikia/AdEngine/js/slot/bottomLeaderboard.js
```diff
@@ -26,11 +39,35 @@ define('ext.wikia.adEngine.slot.bottomLeaderboard', [
...
+					log(['pushSlot', 'Pushed ' + slotName], 'debug', logGroup);
+				}
 			}
 		});
 
+	function collapseOnSameViewport() {
+		if (adContext.get('opts.isBLBViewportEnabled') && isUap() && viewportObserver.sameViewport(slotName, breakSlots)) {
+			eventDispatcher.dispatch('adengine.slot.status', {
+				slot: {
```
fraszczakszymon: I believe it will hurt us that we're dispatching this or similar events with objects that have different interfaces

extensions/wikia/AdEngine/js/slot/bottomLeaderboard.js
```diff
@@ -26,11 +39,35 @@ define('ext.wikia.adEngine.slot.bottomLeaderboard', [
...
+					name: slotName
+				},
+				status: 'viewport',
```
fraszczakszymon: We already have a solution to collapse slots with specific type:
https://github.com/Wikia/app/blob/dev/extensions/wikia/AdEngine/js/provider/btfBlocker.js#L73
and this flow makes clear tracking (without calling DFP ofc)

extensions/wikia/AdEngine/js/tracking/adInfoListener.js
```diff
@@ -51,7 +51,7 @@ define('ext.wikia.adEngine.tracking.adInfoListener',  [
 	}
 
 	function trackSlot(slot, status, adInfo) {
-		var slotFirstChildData = slot.container.firstChild.dataset,
+		var slotFirstChildData = slot.container ? slot.container.firstChild.dataset : {},
```
fraszczakszymon: This is the weak side of this solution. We need to hack all places in adengine component where adSlot object may be used and be prepared for a case when adSlot is notReallyAdSlot. And in future we'll have to be careful about this case.

v3rth (approved):

v3rth (approved):

PR #15102, opened 2018-04-02
XF-100 add error handling to dumpForumData.php

jsutterfield:
Ticket: https://wikia-inc.atlassian.net/browse/XF-100

We've received a number of error reporter tickets with the following error:
```
PHP Warning: fwrite() expects parameter 1 to be resource, boolean given in /usr/wikia/slot1/21916/src/extensions/wikia/Discussions/maintenance/dumpForumData.php on line 79
```
The problem is that inside our `dumpForumData.php` script we're opening a file without ensuring we did so successfully. The php function we're using, `fopen` returns `false` on error and we're then passing `false` to our `fwrite` calls, rather than a file handler. This PR updates that script to add a check the file was opened successfully, failing if not.

### DIFF
--- a/extensions/wikia/Discussions/maintenance/dumpForumData.php
+++ b/extensions/wikia/Discussions/maintenance/dumpForumData.php
@@ -25,11 +25,12 @@ class DumpForumData extends Maintenance {
 	}
 
 	public function execute() {
-		if ( $this->hasOption( 'out' ) ) {
-			$this->fh = fopen( $this->getArg(), 'w' );
-		} else {
-			$this->fh = STDOUT;
+		$outputName = $this->hasOption( 'out' ) ? $this->getArg() : "php://stdout";
+		$this->fh = fopen( $outputName, 'w' );
+		if ( $this->fh === false ) {
+			$this->error( "Unable to open file " . $outputName, 1 );
 		}
+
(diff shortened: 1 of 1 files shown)


extensions/wikia/Discussions/maintenance/dumpForumData.php
```diff
@@ -25,11 +25,12 @@ public function __construct() {
...
 
 	public function execute() {
-		if ( $this->hasOption( 'out' ) ) {
-			$this->fh = fopen( $this->getArg(), 'w' );
-		} else {
-			$this->fh = STDOUT;
+		$outputName = $this->hasOption( 'out' ) ? $this->getArg() : "php://stdout";
+		$this->fh = fopen( $outputName, 'w' );
+		if ( $this->fh === false ) {
+			$this->error( "Unable to open file " . $outputName, 1 );
```
jsutterfield, line 31: The `1` we're passing to the `error` method has the script die with a return value of 1. That return value is checked [here](https://github.com/Wikia/pandora/blob/master/service/discussion/src/main/java/com/wikia/discussionservice/tools/export/batchmigration/mediawiki/WikiaExporter.java#L28-L31), which halts the migration process.

rogatty (approved):

PR #15104, opened 2018-04-03
SUS-4506 | File page with video not handling t URL param

macbre:
https://wikia-inc.atlassian.net/browse/SUS-4506

Don't be so strict with type hints - prevent `Unexpected non-MediaWiki exception encountered, of type "TypeError"
TypeError: Argument 1 passed to WikiaFilePage::getVideoInfoLine() must be an instance of WikiaLocalFile, instance of OldWikiaLocalFile given`.

### DIFF
--- a/extensions/wikia/FilePage/WikiaFilePage.php
+++ b/extensions/wikia/FilePage/WikiaFilePage.php
@@ -55,6 +55,7 @@ class WikiaFilePage extends ImagePage {
 
 		JSMessages::enqueuePackage('VideoPage', JSMessages::EXTERNAL);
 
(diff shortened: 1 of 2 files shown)


extensions/wikia/FilePage/WikiaFilePage.php
```diff
@@ -94,10 +95,10 @@ protected function openShowVideo() {
 
 	/**
 	 * Display info about the video below the video player: provider, views, expiration date (if any)
-	 * @param WikiaLocalFileShared $file
+	 * @param WikiaLocalFile|OldWikiaLocalFile $file
```
pwojcik86, line 98: Can these classes implement a common interface?

PR #15105, opened 2018-04-03
SUS-4512 | Dockerfile for MediaWiki PHP app container

macbre:
https://wikia-inc.atlassian.net/browse/SUS-4512

This is an initial version of `Dockerfile` to run Wikia app and make all tests pass. Once we complete this step, we can create a base PHP image out of this one and then build prod and dev images.

### Unit tests pass

```
# sh php-all
Time: 1.02 minutes, Memory: 164.00MB

OK, but incomplete, skipped, or risky tests!
Tests: 4344, Assertions: 6740, Skipped: 19.

--

Time: 51.08 seconds, Memory: 108.00MB

There was 1 failure:

1) UploadVerifyFile::testUploadVerifyFile with data set #1 ('image/gif', false, 'foobar')
Failed asserting that true matches expected false.

/usr/wikia/slot1/current/src/includes/wikia/tests/UploadVerifyFileTest.php:38

FAILURES!
Tests: 201, Assertions: 769, Failures: 1.
```

There's just a single failing integration test (it checks the code that calls `identify` binary).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,11 @@
-FROM php:7.0.28-cli-alpine
+FROM php:7.0.28-cli
 
-# install PHP extensions required by MediaWiki
-RUN docker-php-ext-install mysqli
+# install make (reuired by unit tests)
+# and install PHP extensions required by MediaWiki
+RUN apt-get update && apt-get install -y \
+    make \
+    && docker-php-ext-install \
+    mysqli
 
 # expose volumes for app and config repositories clones
 ENV WIKIA_DOCROOT=/usr/wikia/slot1/current/src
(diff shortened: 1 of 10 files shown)


pwojcik86 (dismissed):
We also need a container that has a fixed version of the code inside, which we could deploy on a k8s cluster. The one you've created is suited for development only.

Dockerfile
```diff
@@ -0,0 +1,85 @@
...
+    && mkdir -p /tmp/mustache \
+    && tar -xf mustache.tar.gz -C /tmp/mustache --strip-components=1 \
+    && docker-php-ext-configure /tmp/mustache \
+    && docker-php-ext-install /tmp/mustache \
+    && rm -r /tmp/mustache
+
+
+# uopz extension / @see https://pecl.php.net/package/uopz
+RUN pecl install uopz-5.0.2 \
+    && docker-php-ext-enable uopz
```
mszabo-wikia: should we install uopz in production?

pwojcik86 (approved):

PR #15110, opened 2018-04-04
ADEN-6939 Collapse TLB and BLB for UAP:JWP

nandy-andy:


extensions/wikia/AdEngine/js/provider/btfBlocker.js
```diff
@@ -30,6 +30,16 @@ define('ext.wikia.adEngine.provider.btfBlocker', [
 		return win.ads.runtime.disableBtf;
 	}
 
+	function isFanTakeoverFromPlayerLoaded() {
+		return bridge.universalAdPackage.getType() === 'jwp';
+	}
+
+	function shouldCollapseSlotForFanTakeover(slotName) {
+		var uapJwpSlotsToBlock = ['TOP_LEADERBOARD', 'BOTTOM_LEADERBOARD'];
```
fraszczakszymon: What about mobile?

fraszczakszymon:
What do you think about moving fan-takeover-related logic out from btfBlocker? We can disable specific slots in https://github.com/Wikia/app/blob/dev/extensions/wikia/AdEngine/js/run/desktop.run.js#L82 (by using `slotRegistry.get('TOP_LEADERBOARD').disable()`). Then, in the btfBlocker, we check whether it `isEnabled()` and collapse it in case of false.

Mobile can work in the same way.

extensions/wikia/AdEngine/js/provider/directGpt.js
```diff
@@ -26,6 +26,12 @@ define('ext.wikia.adEngine.provider.directGpt', [
 			'GPT_FLUSH'
 		];
 
+	if (context.opts.isFVUapKeyValueEnabled) {
```
fraszczakszymon: It won't be needed if we move fan-takeover-related logic from btf blocker

v3rth (approved):

PR #15111, opened 2018-04-04
ADEN-6972 Add geo with sampling

v3rth:


extensions/wikia/AdEngine/js/context/adContext.js
```diff
@@ -215,6 +216,10 @@ define('ext.wikia.adEngine.adContext', [
 		context.opts.disableSra = geo.isProperGeo(instantGlobals.wgAdDriverDisableSraCountries);
 		context.opts.isBLBMegaEnabled = geo.isProperGeo(instantGlobals.wgAdDriverBottomLeaderBoardMegaCountries);
 
+		context.opts.labradorTestGroup =
```
fraszczakszymon: Where is it used?

v3rth: It isn't, it is defined only for testing purposes

extensions/wikia/AssetsManager/config.php
```diff
@@ -219,6 +219,7 @@
 		'//extensions/wikia/AdEngine/js/utils/eventDispatcher.js',
 		'//extensions/wikia/AdEngine/js/utils/sampler.js',
 		'//extensions/wikia/AdEngine/js/utils/timeBuckets.js',
+		'//extensions/wikia/AdEngine/js/build/bridge.js',
```
nandy-andy: Yet another reason to move to AE3 ASAP :japanese_ogre:

fraszczakszymon: 😱 is it really really needed here? It will affect all bidders performance (?)

v3rth:
Tests will pass after merge dev https://github.com/Wikia/app/commit/9c0c5d230755051c1ce76547d81609b5b1b389c8 . I can't do this right now, because it will break versions of dependencies

PR #15127, opened 2018-04-06
SUS-4523 | Double check hardcoded php binary path

dmnsobczak:


pwojcik86 (changes requested):
I think it's enough to assign `$wgPhpCli` to `phpBin`

dmnsobczak:
I think you're right. Changed.

pwojcik86 (changes requested):
`phpBin` is used in other places as well, we should update all usages in this class

pwojcik86 (approved):

PR #15138, opened 2018-04-09
XW-4851 | [MercuryApi] Cleanup snippets for file usage list from broken html tags

rybmat:
https://wikia-inc.atlassian.net/browse/XW-4851

@Wikia/x-wing

### DIFF
--- a/extensions/wikia/MercuryApi/handlers/MercuryApiFilePageHandler.class.php
+++ b/extensions/wikia/MercuryApi/handlers/MercuryApiFilePageHandler.class.php
@@ -12,8 +12,25 @@ class MercuryApiFilePageHandler {
 		$details = WikiaFileHelper::getMediaDetail( $title );
 		$mediaObject = ArticleAsJson::createMediaObject( $details, $title->getText() );
 
+		// if article contains user provided HTML which is invalid (e.g. to many </div>), snippetter treat it as a text
+		// and it is not removed as other tags. In mobile-wiki we do not escape snippets, therefore invalid html may
+		// cause errors on front-end side
+		$fileUsageList = array_map( function( $item ) {
+			$tidy = new tidy();
+			$tidy->parseString( $item['snippet'] );
+			$tidy->cleanRepair();
+			$item['snippet'] = array_reduce(
+				$tidy->body()->child,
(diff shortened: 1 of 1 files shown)


extensions/wikia/MercuryApi/handlers/MercuryApiFilePageHandler.class.php
```diff
@@ -12,8 +12,25 @@ public static function getFileContent( Title $title ) {
 		$details = WikiaFileHelper::getMediaDetail( $title );
 		$mediaObject = ArticleAsJson::createMediaObject( $details, $title->getText() );
 
+		// if article contains user provided HTML which is invalid (e.g. to many </div>), snippetter treat it as a text
```
hakubo: 'too'

extensions/wikia/MercuryApi/handlers/MercuryApiFilePageHandler.class.php
```diff
@@ -12,8 +12,25 @@ public static function getFileContent( Title $title ) {
...
+		// and it is not removed as other tags. In mobile-wiki we do not escape snippets, therefore invalid html may
+		// cause errors on front-end side
+		$fileUsageList = array_map( function( $item ) {
```
hakubo: should we check if fileUsageList is an array?

rybmat: we don't have to, $fileUsageData['fileList'] is always an array and array_map returns array if array provided as source

extensions/wikia/MercuryApi/handlers/MercuryApiFilePageHandler.class.php
rybmat: done line below

hakubo (approved):

extensions/wikia/MercuryApi/handlers/MercuryApiFilePageHandler.class.php
KockaAdmiralac: I think they meant `e.g. to many </div>` -> `e.g. too many </div>`

PR #15150, opened 2018-04-09
XW-4645 | Implement Mobile Site link on Oasis for mobile

kolek7:
@Wikia/x-wing 

https://wikia-inc.atlassian.net/browse/XW-4645

### DIFF
--- a/extensions/wikia/DesignSystem/scripts/DesignSystemGlobalFooter.js
+++ b/extensions/wikia/DesignSystem/scripts/DesignSystemGlobalFooter.js
@@ -1,3 +1,9 @@
-function footerClick(){
-	document.cookie = 'useskin=; expires=Thu, 01 Jan 1970 00:00:01 GMT; path=/; domain=.wikia-dev.pl';
-}
+
+require(['jquery'],
+	function ($) {
+		$('.mobile-site-link').click(mobileSiteLinkClicked);
+		function mobileSiteLinkClicked(){
+			document.cookie = 'useskin=; expires=Thu, 01 Jan 1970 00:00:01 GMT; path=/; domain=' + wgCookieDomain;
+		}
+	}
+);
--- a/extensions/wikia/DesignSystem/scripts/DesignSystemGlobalFooter.js
+++ b/extensions/wikia/DesignSystem/scripts/DesignSystemGlobalFooter.js
@@ -1,12 +1,10 @@
 
 require(['jquery'],
 	function ($) {
-
 		if ( $.cookie('useskin') === 'oasis' ) {
 			$('.mobile-site-link').click(mobileSiteLinkClicked);
 			$('.mobile-site-link').css("display", "inline-block");
 		}
-
 		function mobileSiteLinkClicked(){
 			$.cookie("useskin", null, {path: '/', domain: wgCookieDomain});
 		}
--- a/extensions/wikia/DesignSystem/scripts/DesignSystemGlobalFooter.js
+++ b/extensions/wikia/DesignSystem/scripts/DesignSystemGlobalFooter.js
@@ -1,9 +1,10 @@
 require(['jquery'],
 	function ($) {
-		if ( $.cookie('useskin') === 'oasis' ) {
+		if ($.cookie('useskin') === 'oasis')  {
 			$('.mobile-site-link').click(mobileSiteLinkClicked);
 			$('.mobile-site-link').css("display", "inline-block");
 		}
+
 		function mobileSiteLinkClicked(){
 			$.cookie("useskin", null, {path: '/', domain: wgCookieDomain});
 		}
--- a/extensions/wikia/DesignSystem/scripts/DesignSystemGlobalFooter.js
+++ b/extensions/wikia/DesignSystem/scripts/DesignSystemGlobalFooter.js
@@ -1,8 +1,9 @@
 require(['jquery'],
 	function ($) {
 		if ($.cookie('useskin') === 'oasis')  {
-			$('.mobile-site-link').click(mobileSiteLinkClicked);
-			$('.mobile-site-link').css("display", "inline-block");
+			$('.mobile-site-link')
+				.click(mobileSiteLinkClicked)
+				.addClass("mobile-site-link--active");
 		}
 
 		function mobileSiteLinkClicked(){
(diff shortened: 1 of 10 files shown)


extensions/wikia/DesignSystem/scripts/DesignSystemGlobalFooter.js
```diff
@@ -0,0 +1,12 @@
+require(['jquery'],
+	function ($) {
+		if ($.cookie('useskin') === 'oasis')  {
+			$('.mobile-site-link').click(mobileSiteLinkClicked);
+			$('.mobile-site-link').css("display", "inline-block");
```
rybmat: do we need to do it via JS? what about simple css like this:
`.WikiaSiteWrapper .wds-global-footer .mobile-site-link {display: block;}` ?

includes/wikia/models/DesignSystemGlobalFooterModel.class.php
```diff
@@ -212,6 +212,13 @@ public function getData() {
 					]
 				],
 			],
+			'mobile_site_button' => [
```
rybmat, line 215: during QA it is worth to test if it won't break mobile-wiki, discussions-frontend or fandom

extensions/wikia/DesignSystem/scripts/DesignSystemGlobalFooter.js
```diff
@@ -0,0 +1,12 @@
+require(['jquery'],
+	function ($) {
+		if ($.cookie('useskin') === 'oasis')  {
```
rybmat, line 3: is this mobile-link button intended to show only if user clicked the full site link on mobile wiki? AFAIK this cookie is set only in this case. Maybe we could show it always on desktop?

rybmat (changes requested):

extensions/wikia/DesignSystem/scripts/DesignSystemGlobalFooter.js
bkoval: Nah, I think that's what we want in the ticket - to only show it when you switch to oasis on a mobile device.

extensions/wikia/DesignSystem/scripts/DesignSystemGlobalFooter.js
```diff
@@ -0,0 +1,12 @@
...
+			$('.mobile-site-link').click(mobileSiteLinkClicked);
```
bkoval: You can chain these, e.g.
```
$('.mobile-site-link')
    .click(mobileSiteLinkClicked)
    .css("display", "inline-block");
```
This way you do not evaluate the same selector twice (makes performance better, all the DOM operations are performance-costly).

extensions/wikia/DesignSystem/scripts/DesignSystemGlobalFooter.js
bkoval: We need to set it dynamically as we only show it when the cookie is set. That set, we can prepare a class in .scss, e.g. "mobile-site-link--active" that will set display: inline-block and we assign this class here, e.g.
```
.addClass ('mobile-site-link--active');
```

This way we keep all the presentation (styling) layer in the SCSS file :)

extensions/wikia/DesignSystem/services/templates/DesignSystemGlobalFooter_licensingAndVertical.php
```diff
@@ -1,3 +1 @@
-<div class="wds-global-footer__bottom-bar">
-	<div class="wds-global-footer__bottom-bar-row wds-has-padding"><?= DesignSystemHelper::renderText( $model['description'] ) ?></div>
-</div>
+<div class="wds-global-footer__bottom-bar-row wds-has-padding"><?= DesignSystemHelper::renderText( $model['description'] ) ?></div>
```
bkoval, line 1: Why did you remove the wrapper div?

extensions/wikia/DesignSystem/styles/DesignSystemGlobalFooter.scss
```diff
@@ -0,0 +1,3 @@
+.mobile-site-link {
+  display: none;
```
bkoval: Indentation should use one tab, not two spaces :)

includes/wikia/models/DesignSystemGlobalFooterModel.class.php
```diff
@@ -212,6 +212,13 @@ public function getData() {
...
+				'type' => 'link-text',
+				'title' => [
+					'type' => 'translatable-text',
+					'key' => 'global-footer-mobile-site-link'
```
bkoval, line 219: Was that already translated?

bkoval (changes requested):
Good job! Left a couple of comments.

includes/wikia/models/DesignSystemGlobalFooterModel.class.php
kolek7: Fortunately, yes 🙂

extensions/wikia/DesignSystem/styles/DesignSystemGlobalFooter.scss
kolek7: Gotta configure my IDE config then.

extensions/wikia/DesignSystem/services/templates/DesignSystemGlobalFooter_licensingAndVertical.php
kolek7: I moved this wrapper to `DesignSystemGlobalFooter_index.php`, because I wanted to nest both licensingAndVertical template and mobileSiteButton template inside it.

extensions/wikia/DesignSystem/services/DesignSystemGlobalFooterService.class.php
```diff
@@ -29,6 +29,10 @@ public function licensingAndVertical() {
 		$this->setVal( 'model', $this->getVal( 'model' ) );
 	}
 
+	public function mobileSiteButton() {
+		$this->setVal( 'model', $this->getVal( 'model' ) );
```
mszabo-wikia: you could use `renderPartial` instead when rendering this template, it wont invoke the full dispatcher process and doesn't require having such a boilerplate method defined

rybmat (approved):

JamesHawking (approved):

bkoval (approved):

PR #15152, opened 2018-04-10
SUS-4544 allow rabbit publishes to fail fast

pwojcik86:
https://wikia-inc.atlassian.net/browse/SUS-4544

This change puts all rabbit channels in **publish confirm** mode, which allows publishes to fail when rabbit is blocking the connection, due to insufficient resources - https://www.rabbitmq.com/alarms.html.

**IMPORTANT**: Because timeout exceptions are caught, this change may cause an inconsistent state when a request is successful, but a publish fails (e. g. updating Solr, tasks, etc.). Keep this in mind while reviewing this.

### DIFF
--- a/extensions/wikia/PhalanxII/hooks/PhalanxHooks.class.php
+++ b/extensions/wikia/PhalanxII/hooks/PhalanxHooks.class.php
@@ -6,6 +6,8 @@ use Wikia\Rabbit\ConnectionBase;
 class PhalanxHooks {
 	const ROUTING_KEY = 'onUpdate';
 
+	protected static $rabbitConnection;
+
 	/**
 	 * Add a link to central:Special:Phalanx from Special:Contributions/USERNAME
 	 * if the user has 'phalanx' permission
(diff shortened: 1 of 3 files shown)


mszabo-wikia (approved):

extensions/wikia/PhalanxII/hooks/PhalanxHooks.class.php
```diff
@@ -267,4 +266,15 @@ static public function onContributionsLogEventsList( OutputPage $out, User $user
...
+	/** @return \Wikia\Rabbit\ConnectionBase */
+	protected static function getRabbitConnection() {
+		global $wgPhalanxQueue;
+
+		if ( !isset( self::$rabbitConnection ) ) {
+			self::$rabbitConnection = new ConnectionBase( $wgPhalanxQueue );
+		}
+
+		return self::$rabbitConnection;
+	}
```
macbre, line 279: :+1:

includes/wikia/rabbitmq/ConnectionBase.class.php
```diff
@@ -56,6 +57,8 @@ public function publish( $routingKey, $body ) {
 				$this->exchange,
 				$routingKey
 			);
+
+			$channel->wait_for_pending_acks(ACK_WAIT_TIMEOUT_SECONDS);
```
macbre: Shouldn't it be `self::ACK_WAIT_TIMEOUT_SECONDS`?

macbre (changes requested):
One note left

macbre (approved):

PR #15166, opened 2018-04-12
SUS-4549: Generate article snippets inside MW

mszabo-wikia:
This gets rid of Solr replication lag, allowing for e.g. search result descriptions to be updated immediately after an article is edited, thus preventing search engines from caching outdated description snippets.

Also:
* changed TTL to 14 days (same as parser cache)
* snippet is recached on article save instead of busting the cache

https://wikia-inc.atlassian.net/browse/SUS-4549


includes/wikia/services/ArticleService.class.php
```diff
@@ -10,36 +10,11 @@
 class ArticleService extends WikiaObject {
 	const MAX_LENGTH = 500;
 	const CACHE_VERSION = 9;
-	const SOLR_SNIPPETS_FIELD = 'snippet_s';
```
macbre, line 13: Can we remove this field from Solr documents?

mszabo-wikia: Sure, definitely. I'd wait with that some time after release to ensure everything is working smoothly with new snippets generation.

includes/wikia/services/ArticleService.class.php
```diff
@@ -177,19 +152,9 @@ private function getTextSnippetSource( $articleId ) {
...
-						$content = $service->getTextFromSolr();
-					}
-
-					if ( $content === '' ) {
-						// back-off is to use mediawiki
-						$content = $service->getUncachedSnippetFromArticle();
-					}
-
-					return $content;
+				86400 * 14 /* 14 days, same as parser cache */,
```
macbre, line 154:
```
includes/DefaultSettings.php
1531:$wgParserCacheExpireTime = 86400;
```

Parser cache expiry is set to 24h

includes/wikia/services/ArticleService.class.php
```diff
@@ -177,19 +152,9 @@ private function getTextSnippetSource( $articleId ) {
...
+				function () use ( $service ) {
```
macbre: Let's avoid assigning `$this` to `$service` and use `$this` in this callback - `use` will be redundant.

macbre (changes requested):
Left few notes

macbre (approved):

includes/wikia/services/ArticleService.class.php
mszabo-wikia: its overridden in our config, `$wgParserCacheExpireTime = 86400 * 14;`

PR #15179, opened 2018-04-16
IRIS-5811 | add discussions' sitemap path to sitemap-index

xkxd:
https://wikia-inc.atlassian.net/browse/IRIS-5811

## DON'T MERGE UNTIL `discussions-sitemap` SERVICE WILL BE DEPLOYED ON PROD


extensions/wikia/SitemapXml/SitemapXmlController.class.php
```diff
@@ -163,6 +163,9 @@ private function generateSitemapPage( array $namespaces, $page ) {
 	}
 
 	private function generateSitemapIndex() {
+		global $wgCityId, $wgEnableDiscussions;
+
+		$discussionsSitemapUrl = 'http://services.wikia.com/discussions-sitemap/sitemap/' . $wgCityId;
```
rogatty: This can be inside the `if`

rogatty: And I think we should always use https for services.wikia.com

rogatty (approved):

rogatty (approved):

PR #15187, opened 2018-04-17
SUS-4593 | Use shared DB for localisation cache

mszabo-wikia:
Replace serialized CDB files with shared DB + memcached combination, so we don't need to store 700MiB+ of cached messages on every MediaWiki server.

The `LCStore` database implementation is pulled from upstream but a prefix field was added to allow to vary the cache between different clients (devboxes and sandboxes each running different versions of code and messages etc.).

**TODO:**
- [x] update config
- [x] create table


pwojcik86 (approved):

drozdo:
Awesome 👍

macbre (approved):

macbre:
```
Notice: unserialize(): Error at offset 0 of 16 bytes in Localisation/LCStoreDB.php on line 61
```

PR #15198, opened 2018-04-18
SUS-4605: Improve edit count caching

mszabo-wikia:
* Don't try to calculate edit count if the value is absent from database—since it's set on first edit, we'll always get 0 and just waste our time.
* Always increment and set edit count and weekly edit count values on edit
* Use current timestamp to set last contribution timestamp (and also first, if this is their first edit), instead of wasting time on an useless lookup

https://wikia-inc.atlassian.net/browse/SUS-4605

### DIFF
--- a/includes/wikia/services/UserStatsService.class.php
+++ b/includes/wikia/services/UserStatsService.class.php
@@ -22,7 +22,7 @@ class UserStatsService extends WikiaModel {
 	function __construct( $userId, $wikiId = 0 ) {
 		$this->userId = intval( $userId );
 		parent::__construct();
-		$this->initWikiId( $wikiId );
+		$this->wikiId = ( $wikiId === 0 ) ? $this->wg->CityId : $wikiId;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


includes/wikia/services/UserStatsService.class.php
```diff
@@ -103,41 +95,25 @@ public function increaseEditsCount() {
...
-					// editcount not set yet, calculate it
-					$stats['editcount'] = $this->calculateEditCountWiki( $flags );
-				}
-				else {
-					// make sure this value is an integer
-					$stats['editcount'] = (int) $stats['editcount'];
-				}
-
-				if ( $stats['editcount'] === 0 ) {
+					// editcount not set yet, so user has no edits
```
macbre, line 106: Are we sure that's always true? Did we initialise values in the table with proper user edit counters on each wiki?

mszabo-wikia: SQL logs indicate this gets called for users with 0 edits on the wiki

macbre (approved):

macbre (approved):

PR #15164, opened 2018-04-12
SUS-4573 | Run MediaWiki in Docker locally

dmnsobczak:
* use `php-fpm` and `nginx` duo
* use a local instance of `memcached` to greatly speed up the Docker instance running on 127.0.0.1

PR #15203, opened 2018-04-19
SUS-4593 | Use shared DB for localisation cache

macbre:
See #15187 // bring back changes reverted in #15164

Replace serialized CDB files with shared DB + memcached combination, so we don't need to store 700MiB+ of cached messages on every MediaWiki server.

The `LCStore` database implementation is pulled from upstream but a prefix field was added to allow to vary the cache between different clients (devboxes and sandboxes each running different versions of code and messages etc.).

**TODO:**
- [x] update config
- [x] create table

**Config**: https://github.com/Wikia/config/pull/2459


macbre:
`LocalisationCache` is in many cases called for not existing messages via `wfEmptyMsg()` and `Message::exists()` methods. It makes more sense to fetch all messages for a given language code in a single query and cache it under a single memcache key. This will avoid database queries (more than 200 on a single page view) that always return no results.

```
 51.38% 6.493350    483 - Wikia\Localisation\CachedLCStoreDB::get
 12.28% 1.551955    225 - Wikia\Localisation\LCStoreDB::get
 11.65% 1.472562    225 - query: SELECT lc_value FROM `lNn_cache` WHERE lc_prefix = 'X' LIMIT N 
```

mszabo-wikia (approved):

lib/Wikia/src/Localisation/CachedLCStoreDB.php
```diff
@@ -43,9 +60,33 @@ function get( $code, $key ) {
...
 		}
 
+		wfProfileOut( __METHOD__ );
 		return $value;
 	}
 
+	/**
+	 * Return the list of all messages that are kept in localisation cache.
+	 *
+	 * Avoid DB queries for messages that we now they do not exist.
```
bkoval, line 70: *know

lib/Wikia/src/Localisation/CachedLCStoreDB.php
```diff
@@ -43,9 +60,33 @@ function get( $code, $key ) {
...
+	 * via wfEmptyMsg() and Message::exists() methods
+	 *
+	 * @param string $code
+	 * @return string[]
+	 */
+	private function getAllKeys( $code ) {
+		$cacheKey = $this->getCacheKeyForMessage( $code, __METHOD__ );
+		$keys = $this->cacheService->get( $cacheKey );
+
+		if ( !is_array( $keys ) ) {
```
bkoval, line 82: Can $keys ever be an empty array? I suppose it cannot, since something either is cached or not, just asking :)

lib/Wikia/tests/Localisation/LCStoreDBIntegrationTest.php
```diff
@@ -1,48 +0,0 @@
-<?php
-namespace Wikia\Localisation;
-
-/**
- * @group Integration
- */
-class LCStoreDBIntegrationTest extends \WikiaDatabaseTest {
```
bkoval, line 7: Why do we remove the test class?

mszabo-wikia: SQLite unfortunately has trouble with `INSERT ... ON DUPLICATE KEY UPDATE SET foo = VALUES(foo)`syntax, so it's not really possible to write an isolated integration test now—if we could use mysql in docker or similar solution, it'd be fine.

bkoval (approved):

macbre:
Let's wait for #15226 - new memcache client works great with a new localisation cache implementation.

PR #15226, opened 2018-04-23
SUS-4611 | backport MemcachedPeclBagOStuff from MW 1.20

macbre:
https://wikia-inc.atlassian.net/browse/SUS-4611

This pull request includes a [backport from MW 1.20 that introduces `memcached` extension based client](https://github.com/wikimedia/mediawiki/commit/3c62077fe28778eb41bde7549fbae402c8be0ef7#diff-86946fda95cb93f7a1deb64d78e6c661). This will replace an old, Perl-based memcache client that is written in PHP.

I've also improved caching and errors handling in `VideoHandlers` classes that control video uploads from 3rd party providers. 

## Performance

In terms of performance there's no difference (tested on sandbox) in overall response time (median value for 100 requests difference by less than 1%). Profiling shows that the old client uses ~17% of the response time, while the new one - ~13%.

The new client is way more reliable when using a new localisation cache - see #15203. The code is much simpler (there's no sockets communication handling and results parsing in PHP code).

## Deployment

This PR **does not switch the memcache client on production nor on preview / verify**. For now it will only be enabled on sandboxes and devboxes.

Due to incompatibility of serialization of non-scalar types between these two clients we'd need to introduce a key prefix for a new client (that's a part of this PR). In order to avoid a flood of cache misses after production deploy we need to start enabling the new client on a small fraction of production requests first.


includes/objectcache/MemcachedPeclBagOStuff.php
```diff
@@ -0,0 +1,200 @@
...
+			default:
+				$msg = $this->client->getResultMessage();
+				if ( $key !== false ) {
+					$server = $this->client->getServerByKey( $key );
+					$serverName = "{$server['host']}:{$server['port']}";
+					$msg = "Memcached error for key \"$key\" on server \"$serverName\": $msg";
+				} else {
+					$msg = "Memcached error: $msg";
+				}
+				wfDebugLog( 'memcached-serious', $msg );
```
mszabo-wikia: Let's update logging statements as needed

mszabo-wikia (changes requested):

ghost (approved):

includes/objectcache/MemcachedBagOStuff.php
```diff
@@ -0,0 +1,176 @@
...
+
+	protected function encodeKeyCallback( $m ) {
+		return rawurlencode( $m[0] );
+	}
+
+	/**
+	 * TTLs higher than 30 days will be detected as absolute TTLs
+	 * (UNIX timestamps), and will result in the cache entry being
+	 * discarded immediately because the expiry is in the past.
+	 * Clamp expiries >30d at 30d, unless they're >=1e9 in which
```
bkoval: *expires?

mszabo-wikia (approved):

bkoval (approved):

PR #15237, opened 2018-04-25
ADEN-6976 Add device info to kikimora tracking

dmnsobczak:


extensions/wikia/AdEngine/js/tracking/adInfoTracker.js
```diff
@@ -45,6 +46,7 @@ define('ext.wikia.adEngine.tracking.adInfoTracker',  [
 			'pv': pageParams.pv || '',
 			'pv_unique_id': win.pvUID,
 			'browser': [ browserDetect.getOS(), browserDetect.getBrowser() ].join(' '),
+			'device': megaAdUnitBuilder.getDevice(pageParams),
```
fraszczakszymon: I believe we don't want to have results like `unknown-specialpage` :P. Can we avoid using megaAdUnitBuilder for that? What about extracting separate module for device and use it here and in megaAdUnitBuilder?

dmnsobczak: Done!

extensions/wikia/AdEngine/js/utils/device.js
```diff
@@ -0,0 +1,33 @@
+/* global define */
+define('ext.wikia.adEngine.utils.device', [
+	'wikia.browserDetect'
+], function (browserDetect) {
+	'use strict';
+
+	var result = '';
```
fraszczakszymon: We can set 'unknown' by default

extensions/wikia/AdEngine/js/utils/device.js
```diff
@@ -0,0 +1,33 @@
...
+
+	function getDevice(params) {
+		if (params.skin === 'oasis') {
+			result = browserDetect.isMobile() ? 'tablet' : 'desktop';
+		} else if (params.skin === 'mercury' || params.skin === 'mobile-wiki') {
+			result = 'smartphone';
+		}
+
+		return result;
```
fraszczakszymon: return 'unknown' please

dmnsobczak: Done.

fraszczakszymon (approved):

PR #15239, opened 2018-04-25
PLATFORM-3431: send Lyric Find API usage logs to s3

themech:
Logging entrypoint set up in https://github.com/Wikia/chef-repo/pull/16486 and https://github.com/Wikia/wikia-vcl/pull/1743

### DIFF
--- a/extensions/3rdparty/LyricWiki/LyricFind/LyricFindTrackingService.class.php
+++ b/extensions/3rdparty/LyricWiki/LyricFind/LyricFindTrackingService.class.php
@@ -12,6 +12,7 @@ class LyricFindTrackingService extends WikiaObject {
 	const CODE_LYRIC_IS_INSTRUMENTAL = 102;
 	const CODE_LRC_IS_AVAILABLE  = 111;
 	const CODE_LYRIC_IS_BLOCKED  = 206;
+	const USAGE_TRACKING_URL = 'https://lyricsfind.wikia-services.com/__track/';
 
 	// Not documented. The response body says "SUCCESS: NO LYRICS" which I assume means that they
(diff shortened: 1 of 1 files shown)


frankfarmer:
Using fastly as a logging service seems potentially fragile.  What would be the next easiest way to log to s3 without using fastly as a middleman?

harnash (approved):

themech:
@frankfarmer  I went with Fastly-based logging due to fairly low volume of the lyricswiki logs and because that's how PHP-side event logging has been working for years https://github.com/Wikia/app/blob/2e71444e0462cccb72cbaee67b25f6717eebdb7e/extensions/wikia/Track/Track.php#L14 - sending requests to Fastly and storing them on S3.

I still think s3 is the best place to store this kind of logs for longer term (definitely better than ELK and MySQL db). Ideally I think we could do it in kubernetes somewhere on syslog/nsq level as it would work for every app. We could send some of the application the logs messages to places like ELK/S3 or even both. But I don't think we have such a solution now and MediaWiki has not been migrated to k8s yet, so I think sending the logs through Fastly is the most straightforward solution for now. Especially that we do like to start gathering the LyricsFind usage logs.

frankfarmer (approved):

frankfarmer:
I'm doing some work to try to set up a direct route for logging from PHP to S3 without needing fastly as a middleman.  This should make logging to s3 more reusable for any other log producers that might benefit.

Unfortunately the packages I started my proof of concept with proved to have several deficiencies so this is taking longer than I'd hoped.

themech:
replaced with https://github.com/Wikia/app/pull/15299

PR #15243, opened 2018-04-26
SUS-4493 | populateEventsLocalUsers - process members of local groups as well

macbre:
https://wikia-inc.atlassian.net/browse/SUS-4493

Some accounts are missing from `events_local_users` table. This causes some admins not being listed on Special:ListUsers.

Let's update `ListusersData::populateEventsLocalUsers` method to process accounts that made an edit or are of the members of a local wiki user group. Then we will run this method site-wide to make `events_local_users` have correct users data.

### DIFF
--- a/extensions/wikia/Listusers/SpecialListusers_helper.php
+++ b/extensions/wikia/Listusers/SpecialListusers_helper.php
@@ -463,11 +463,16 @@ class ListusersData {
 	 * Fills specials.events_local_users table with entries for a given wiki.
 	 * Used by CreateNewWikiTask class
 	 *
+	 * Get the list of all users that made an edit or a members of local groups
+	 *
 	 * @see SUS-3264
+	 * @see SUS-4493
 	 * @param int $cityId
 	 */
 	public static function populateEventsLocalUsers( int $cityId ) {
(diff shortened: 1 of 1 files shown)


extensions/wikia/Listusers/SpecialListusers_helper.php
```diff
@@ -463,20 +463,44 @@ public function updateUserGroups( User $user, $addgroup = array(), $removegroup
...
+		);
+
+		foreach( $res as $row ) {
+			$ids[] = $row->user_id;
+		}
+
+		// process all accounts
+		$listUsers = new \ListusersData( $cityId );
+
+		foreach( array_unique( $ids ) as $id ) {
```
mszabo-wikia, line 502: let's use either array_unique or DISTINCT but not both :)

mszabo-wikia (changes requested):

extensions/wikia/Listusers/SpecialListusers_helper.php
macbre: Why? These `DISTINCT`s come from two different sets of values.

mszabo-wikia (approved):

PR #15245, opened 2018-04-27
allow configurable hostaname

pwojcik86:
We don't want to use actual hostnames when running sandboxes on k8s. This change allows us to configure a synthetic hostname in env.

### DIFF
--- a/lib/Wikia/src/Tasks/AsyncTaskList.php
+++ b/lib/Wikia/src/Tasks/AsyncTaskList.php
@@ -249,7 +249,7 @@ class AsyncTaskList {
 		];
 
 		if ( $wgWikiaEnvironment != WIKIA_ENV_PROD ) {
-			$host = gethostname();
+			$host = wfGetEffectiveHostname();
 			$executionMethod = 'http';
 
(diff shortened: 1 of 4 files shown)


lib/Wikia/src/Tasks/AsyncTaskList.php
```diff
@@ -249,7 +249,7 @@ protected function getExecutor() {
 		];
 
 		if ( $wgWikiaEnvironment != WIKIA_ENV_PROD ) {
-			$host = gethostname();
+			$host = wfGetEffectiveHostname();
 			$executionMethod = 'http';
 
 			if ( $wgWikiaEnvironment == WIKIA_ENV_DEV && preg_match( '/^dev-(.*?)$/', $host ) ) {
```
macbre, line 255: I'm wondering if this `preg_match` is actually needed.

macbre (approved):

lib/Wikia/src/Tasks/AsyncTaskList.php
pwojcik86: probably not

mszabo-wikia: Jenkins counts as dev env but doesn't have a `WIKIA_DEV_DOMAIN` so I think this is meant to cover that :(

PR #15246, opened 2018-04-30
(PLATFORM-3469) Fix parsing language path on thread ID pages

Grunny:
When accessing a thread page on a language path wiki, a redirect to
the main page of the root (non-language path) wiki occurred, i.e.
`https://some-example.wikia.com/szl/wiki/Thread:24` would redirect
to `https://some-example.wikia.com/wiki/Main_Page`.

This was caused by using PHP's `parse_url` on a relative URL to get the
URL path. Most of the time this works fine, but if the relative URL
resembles the use of a port in a URL, it fails. For example:

    > var_dump( parse_url( '/szl/wiki/Thread:Foo', PHP_URL_PATH ) );
    string(20) "/szl/wiki/Thread:Foo"

    > var_dump( parse_url( '/szl/wiki/Thread:24', PHP_URL_PATH ) );
    bool(false)

As a result, the language code in the path was not parsed and the code
behaved as if it was on the root wiki. Putting anything after the path
allows this to work as expected:

    > var_dump( parse_url( '/szl/wiki/Thread:24?', PHP_URL_PATH ) );
    string(19) "/szl/wiki/Thread:24"

Similarly, changing the URL to the thread page to add anything after the
number, i.e. `https://some-example.wikia.com/szl/wiki/Thread:24?`, would
prevent the redirect bug.

This is due to a bug in the PHP `parse_url` function when handling relative
URLs with a port-like string: https://bugs.php.net/bug.php?id=71646 In order
to work around the bug, we can simply provide a full URL to the parse_url
method, rather than the relative URL.

### DIFF
--- a/extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
+++ b/extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
@@ -71,7 +71,7 @@ class WikiFactoryLoader {
 			// normal HTTP request
 			$this->mServerName = strtolower( $server['SERVER_NAME'] );
 
-			$path = parse_url( $server['REQUEST_URI'], PHP_URL_PATH );
+			$path = parse_url( $server['REQUEST_SCHEME'] . '://' . $server['SERVER_NAME'] . $server['REQUEST_URI'], PHP_URL_PATH );
 
 			$slash = strpos( $path, '/', 1 ) ?: strlen( $path );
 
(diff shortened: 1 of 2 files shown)


mszabo-wikia (approved):

mszabo-wikia:
Let's add a test case for this

harnash (approved):

harnash:
Great find! I think it calls for a engineering blog article or at least an e-mail so nobody runs into this again.

PR #15255, opened 2018-05-07
XF-522 | [MercuryApi] do not try to generate srcset and thumbnailUrl for ogg file pages

rybmat:
https://wikia-inc.atlassian.net/browse/XF-522

@Wikia/x-wing

### DIFF
--- a/extensions/wikia/MercuryApi/handlers/MercuryApiFilePageHandler.class.php
+++ b/extensions/wikia/MercuryApi/handlers/MercuryApiFilePageHandler.class.php
@@ -11,8 +11,10 @@ class MercuryApiFilePageHandler {
 
 		$details = WikiaFileHelper::getMediaDetail( $title );
 		$mediaObject = ArticleAsJson::createMediaObject( $details, $title->getText() );
-		$mediaObject['srcset'] = ArticleAsJson::getSrcset( $mediaObject['url'], $mediaObject['width'] );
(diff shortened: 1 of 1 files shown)


extensions/wikia/MercuryApi/handlers/MercuryApiFilePageHandler.class.php
```diff
@@ -11,8 +11,10 @@ public static function getFileContent( Title $title ) {
 
 		$details = WikiaFileHelper::getMediaDetail( $title );
 		$mediaObject = ArticleAsJson::createMediaObject( $details, $title->getText() );
-		$mediaObject['srcset'] = ArticleAsJson::getSrcset( $mediaObject['url'], $mediaObject['width'] );
-		$mediaObject['thumbnailUrl'] = ArticleAsJson::getThumbnailUrlForWidth( $mediaObject['url'], 340 );
+		if ( $mediaObject['isOgg'] === false ) {
```
latata, line 14: Are you sure it's the only type we should handle this way?

rybmat: @latata It is the only type we can currently differentiate from others. We'll probably need to do followup ticket to handle other file types like .pdf .odf .woff and so on, but it is more work to do and AFAIK ogg is the most common after image / video

latata: 👍

latata (approved):

PR #15257, opened 2018-05-07
SUS-4751: Add functionality to permanently delete files & articles

mszabo-wikia:
https://wikia-inc.atlassian.net/browse/SUS-4751


includes/wikia/WikiaUpdater.php
```diff
@@ -52,6 +52,8 @@ public static function update( DatabaseUpdater $updater ) {
 			array( 'dropIndex', 'wall_related_pages', 'comment_id_idx',  $dir . 'patch-wall_related_pages-drop-comment_id_idx.sql', true ), // SUS-3096
 			array( 'dropIndex', 'wall_related_pages', 'page_id_idx_2',  $dir . 'patch-wall_related_pages-drop-page_id_idx_2.sql', true ), // SUS-3096
 			array( 'dropIndex', 'video_info', 'added_at',  $dir . 'patch-video_info-drop-added_at_idx.sql', true ), // SUS-4297
+			array( 'dropIndex', 'image', 'img_usertext_timestamp', $dir . 'patch-image-drop-usertext-index.sql', true ),
```
rybmat, line 55: @mszabo-wikia why are we changing these indices?

rybmat (approved):

includes/wikia/WikiaUpdater.php
mszabo-wikia: they are not used according to query-digest, so we should replace them with one that can be used for the new queries (to delete entries by user ID)

macbre (approved):

PR #15258, opened 2018-05-07
ADEN-7095 Add GDPR tracking disabling

dmnsobczak:


fraszczakszymon (changes requested):
We didn't want to change real InstantGlobals state :(. What about implementing optOut module that would have method like `isBlacklistedByOptOut('krux')`. That may be used somewhere here:
https://github.com/Wikia/app/blob/dev/extensions/wikia/AdEngine/js/context/adContext.js#L198

nandy-andy (approved):

extensions/wikia/InstantGlobals/InstantGlobalsApiController.class.php
```diff
@@ -12,6 +12,21 @@ public function getNewsAndStoriesVariables() {
...
 
+	/**
+	 * Get news&stories tracking opt-outs values
+	 *
+	 * @return array
+	 */
+	public function getNewsAndStoriesTrackingOptOut() {
+		$trackingOptOut = [];
+
+		Hooks::run( 'InstantGlobalsGetTrackingOptOut', [&$trackingOptOut] );
```
fraszczakszymon: Why do we need instant globals hook here? Can't we return simple array (better: object `{"blacklist":[...]}`?

dmnsobczak: Good suggestion. Changed.

PR #15263, opened 2018-05-08
SUS-4753 | clear user identity box

rybmat:
https://wikia-inc.atlassian.net/browse/SUS-4753

@Wikia/sus

### DIFF
--- a/extensions/wikia/GDPR/UserGlobalProfileDataDelete.php
+++ b/extensions/wikia/GDPR/UserGlobalProfileDataDelete.php
@@ -2,7 +2,7 @@
 
 class UserGlobalProfileDataDelete {
 	public static function deleteGlobalProfileData( User $user ): bool {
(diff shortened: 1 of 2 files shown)


mszabo-wikia (approved):

extensions/wikia/GDPR/UserGlobalProfileDataDelete.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+class UserGlobalProfileDataDelete {
+	public static function deleteGlobalProfileData( User $user ): bool {
+		if ( !( $user instanceof User ) || $user->getId() === 0 ) {
```
macbre: * the class of `$user` argument is forced by the type hint, no need to check instance later on
* let's use `isAnon` instead of checking if user id is zero.

macbre (changes requested):

macbre (approved):

rybmat:
merged with SUS-4752

PR #15281, opened 2018-05-09
SUS-4749 | enable a new memcache client consistently on 10% of wikis

macbre:
https://wikia-inc.atlassian.net/browse/SUS-4749

Let's enable a new memcache client (introduced in #15226) on 10% of production wikis to keep the cache client consistent within a context of a specific wiki. Use a hash of domain name - this unfortunately will force an old client for CLI scripts.

```php
> echo hexdec(substr(md5('de.gta.wikia.com'), 0, 7)) % 100
9
```

Modify `WikiFactory::clearCache` method to purge all memcache keys using both clients, so that WikiFactory changes will be visible on wikis using either client.

Set up new client:

* to prefix all keys with `lm` string (**l**ib**m**emcached) to avoid issue with serialisation incompatibility with an old client
* use [`igbinary` serializer](https://phpolyk.wordpress.com/2011/08/28/igbinary-the-new-php-serializer/) to make memcache have more space (and decrease cache eviction)

> `igbinary` is installed site-wide as it's a dependency of [`php-memcached` package](https://packages.debian.org/source/stretch/php-memcached)


includes/objectcache/MemcachedPeclBagOStuff.php
```diff
@@ -100,7 +104,10 @@ protected function applyDefaultParams( $params ) {
 		}
 
 		if ( !isset( $params['serializer'] ) ) {
-			$params['serializer'] = 'php';
+			// SUS-4749 | use igbinary serializer when available
+			// @see https://github.com/igbinary/igbinary#igbinary
+			// @see https://phpolyk.wordpress.com/2011/08/28/igbinary-the-new-php-serializer/
+			$params['serializer'] = extension_loaded( 'igbinary' ) ? 'igbinary' : 'php';
```
mszabo-wikia: let's move it to config file

themech: are there any machines where we don't have `igbinary` and we fallback to `php`. I wouldn't like to end up in a situation when two machines connecting to the same memcache use different serializers :)

macbre: Done in 914c6c8

ljagiello (approved):
Do you want to remove prefix after migration?

macbre:
@ljagiello - this will require the second round of gradual deploys to production. I'm not quite sure is it worth it. However, feel free to prove me wrong ;)

mszabo-wikia (approved):

ljagiello:
@macbre obviously it's not mandatory, I was more worried if everyone in future will remember about an extra prefix.

Paraphrase famous quote:
For use only in during limited "migration" window, and then maybe again twenty years later.

Rick

PR #15285, opened 2018-05-10
SUS-4111 | Do not render maps for articleAsJson output

rybmat:
https://wikia-inc.atlassian.net/browse/SUS-4111

Since Maps extension was never supported by mobile-wiki, lets not render output for this extension at all there


mszabo-wikia (changes requested):
These changes should be made in Wikia/Maps and updated version pulled in via composer.

rybmat:
ok, did't know, thanks!

mszabo-wikia (approved):

PR #15288, opened 2018-05-11
PLATFORM-3442: include language wikis in robots.txt

themech:
When generating robots.txt, call language wikis hosted under the same domain and include their rules.


extensions/wikia/WikiFactory/WikiFactory.php
```diff
@@ -1200,6 +1242,23 @@ static public function IDtoDB( $city_id, $master = false ) {
...
+	 * "Unlocalizes" the host replaces env-specific domains with "wikia.com", for example
+	 * 'muppet.preview.wikia.com' -> 'muppet.wikia.com'
+	 *
+	 * @param $host
+	 * @return string normalized host name
+	 */
+	protected static function normalizeHost( $host ) {
+		global $wgDevDomain;
+		// strip env-specific pre- and suffixes for staging environment
+		$host = preg_replace( '/\.(stable|preview|verify|sandbox-[a-z0-9]+)\.wikia\.com/', static::WIKIA_TOP_DOMAIN, $host );
```
Grunny: Can we use `$wgWikiaBaseDomain` in place of the hardcoded `wikia\.com` while we're touching this code? i.e. `preg_quote( $wgWikiaBaseDomain )`.

Grunny (changes requested):

extensions/wikia/WikiFactory/WikiFactory.php
themech: good point, done

Grunny (approved):

PR #15289, opened 2018-05-11
ADEN-7109 Disable Prebid.js on ?trackingoptout=1

dmnsobczak:


nandy-andy (approved):
Minor note: remove `$wgCountriesVarName` from `AnalyticsProviderPrebid.php` as it's not used anymore and we're good to go.

PR #15290, opened 2018-05-11
SUS-4752 | remove user data process

pwojcik86:
https://wikia-inc.atlassian.net/browse/SUS-4752

An extension that will allow us to remove all user related data, when we need to process "right to be forgotten" requests.


extensions/AbuseFilter/abusefilter.tables.sqlite.sql
```diff
@@ -28,20 +28,20 @@ CREATE TABLE /*$wgDBprefix*/abuse_filter_action (
...
 	afl_var_dump BLOB NOT NULL,
 	afl_timestamp binary(14) NOT NULL,
 	afl_namespace int NOT NULL,
-	afl_title varchar(255) binary NOT NULL,
-	afl_wiki varchar(64) binary NULL,
+	afl_title varchar(255) NOT NULL,
+	afl_wiki varchar(64),
 	afl_deleted tinyint(1) NOT NULL DEFAULT 0,
-	afl_patrolled_by int unsigned NULL
+	afl_patrolled_by int
```
macbre, line 44: Should we add these schema changes to `WikiaUpdated` script and apply them site-wide?

extensions/wikia/Privacy/RemoveUserDataController.php
```diff
@@ -0,0 +1,85 @@
...
+	use Loggable;
+
+	const ACCEPTED = 202;
+	const METHOD_NOT_ALLOWED = 405;
+
+	public function allowsExternalRequests() {
+		return false;
+	}
+
+	public function removeUserData() {
```
macbre, line 16: Should we guard access to this method via internal request check or a token?

extensions/AbuseFilter/abusefilter.tables.sqlite.sql
mszabo-wikia: This SQLite schema is only used in tests

extensions/wikia/Privacy/RemoveUserDataController.php
mszabo-wikia: `allowsExternalRequests` guard should prevent it

extensions/wikia/Privacy/UserDataRemover.php
```diff
@@ -31,6 +31,10 @@ public function removeGlobalData( int $userId ) {
 			$dbMaster->delete( 'user_email_log', ['user_id' => $userId] );
 			$dbMaster->delete( 'user_properties', ['up_user' => $userId] );
 
+			// commit early so that cache is properly invalidated
+			$dbMaster->commit();
+			wfWaitForSlaves( $dbMaster->getDBname() );
+
```
macbre: :+1:

macbre (approved):

extensions/wikia/Privacy/RemoveUserDataController.php
```diff
@@ -0,0 +1,85 @@
...
+	public function getRemoveDetails() {
+		$this->response->setFormat( WikiaResponse::FORMAT_JSON );
+
+		$userId = $this->getVal( 'userId' );
+		if ( empty( $userId ) ) {
+			$this->response->setCode( WikiaResponse::RESPONSE_CODE_BAD_REQUEST );
+			return;
+		}
+
+		$this->response->setCode( self::ACCEPTED );
```
macbre: HTTP 200 fits better here :)

pwojcik86: I disagree. The queued tasks may still be in progress.

macbre: But this request just gets some stats. However, it's just a matter of being kosher here.

pwojcik86: Ah right, my bad

extensions/wikia/Privacy/UserDataRemover.php
```diff
@@ -0,0 +1,51 @@
...
+	public function removeGlobalData( int $userId ) {
+		try {
+			$user = User::newFromId( $userId );
+			if( $user->isAnon() ) {
+				return false;
+			}
+
+			$userIdentityBox = new UserIdentityBox( $user );
+			$userIdentityBox->clearMastheadContents();
+			Wikia::invalidateUser( $user, true, false );
```
macbre: Let's move this cache invalidation section after the update + commit lines. Additionally, before the DB query let's just call `$user->invalidateCache` to make sure to invalidate old user name -> user ID mapping cache.

macbre: Or even further: let's do what user rename process performs:

```php
		// SUS-3523: invalidate the destination user (by both name and ID, user name to ID
		// mapping is cached as well)
		$this->invalidateUser( User::newFromId( $this->mUserId ) );
		$this->invalidateUser( $this->mNewUsername );
```

Before and after an update query.

extensions/wikia/Privacy/RemoveUserDataOnWikiTask.php
```diff
@@ -0,0 +1,130 @@
...
+	 * @param int $userId
+	 * @return true if operation was successful
+	 */
+	private function removeAbuseFilterData( int $userId ) {
+		global $wgEnableAbuseFilterExtension;
+		if( $wgEnableAbuseFilterExtension ) {
+			try {
+				$db = wfGetDB( DB_MASTER );
+				$db->update( 'abuse_filter', ['af_user_text' => ''], ['af_user' => $userId], __METHOD__ );
+				$db->update( 'abuse_filter_history', ['afh_user_text' => ''], ['afh_user' => $userId], __METHOD__ );
```
mszabo-wikia, line 68: the _user_text columns are always empty

extensions/wikia/Privacy/RemoveUserDataOnWikiTask.php
```diff
@@ -0,0 +1,130 @@
...
+		global $wgEnableBlogArticles, $wgEnableWallExt;
+		try {
+			$namespaces = [NS_USER, NS_USER_TALK];
+			if( $wgEnableBlogArticles ) {
+				$namespaces[] = NS_BLOG_ARTICLE;
+			}
+			if( $wgEnableWallExt ) {
+				$namespaces[] = NS_USER_WALL_MESSAGE_GREETING;
+				$namespaces[] = NS_USER_WALL_MESSAGE;
+				$namespaces[] = NS_USER_WALL;
```
mszabo-wikia: let's remove the wg conditions and use literals instead of constants so we cover also wikis where wall or blogs used to be enabled then got turned off

mszabo-wikia (changes requested):

extensions/wikia/Privacy/RemoveUserDataOnWikiTask.php
pwojcik86: They're not, but we can remove them separately.

extensions/wikia/Privacy/RemoveUserDataOnWikiTask.php
```diff
@@ -0,0 +1,159 @@
+<?php
+
+use Wikia\Logger\Loggable;
+use Wikia\Tasks\Tasks\BaseTask;
+
+/**
+ * Class RemoveUserDataOnWikiTask
+ *
+ * Removes or anonimizes all PII for a specific user on the current community
```
macbre, line 9: `PII` ?

pwojcik86: Personally Identifiable Information

PR #15292, opened 2018-05-11
SUS-4752 | special page for submitting RTBF

rybmat:
https://wikia-inc.atlassian.net/browse/SUS-4752 // will be merged into #15290

@Wikia/sus 

Introduce `request-to-be-forgotten-admin` global users group and expose Special:RequestToBeForgottenInternal for them - [`$wgEnableRequestToBeForgottenInternalSpecialPage`](https://community.wikia.com/wiki/Special:WikiFactory/1474483/variables/wgEnableRequestToBeForgottenInternalSpecialPage) needs to  be set to true.


extensions/wikia/Privacy/Privacy.setup.php
```diff
@@ -5,3 +5,9 @@
 $GLOBALS['wgAutoloadClasses']['RemoveUserDataOnWikiTask'] = __DIR__ . '/RemoveUserDataOnWikiTask.php';
 $GLOBALS['wgAutoloadClasses']['PermanentArticleDelete'] = __DIR__ . '/PermanentArticleDelete.php';
 $GLOBALS['wgAutoloadClasses']['PermanentFileDelete'] = __DIR__ . '/PermanentFileDelete.php';
+
+global $wgCityId;
+if ( COMMUNITY_CENTRAL_CITY_ID === intval($wgCityId) ) {
```
mszabo-wikia: Let's create WikiFactory variable instead with default config value false, and enable the extension on Central.

extensions/wikia/Privacy/SpecialRequestToBeForgottenInternalController.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+class SpecialRequestToBeForgottenInternalController extends WikiaSpecialPageController {
+	public function __construct() {
+		parent::__construct( 'RequestToBeForgottenInternal', 'requesttobeforgotten', false );
+		$this->specialPage->checkPermissions();
+	}
+
+	public function index() {
+		if ( $this->getRequest()->wasPosted() ) {
```
mszabo-wikia: We should validate edit token here.

mszabo-wikia (changes requested):

lib/Wikia/src/Service/User/Permissions/data/PermissionsDefinesAfterWikiFactory.php
```diff
@@ -644,3 +644,7 @@
 $wgGroupPermissions['*']['fandom-admin'] = false;
 $wgGroupPermissions['staff']['fandom-admin'] = true;
 $wgGroupPermissions['fandom-editor']['fandom-admin'] = true;
+
+// request to be forgotten
+$wgGroupPermissions['*']['requesttobeforgotten'] = false;
+$wgGroupPermissions['staff']['requesttobeforgotten'] = true;
```
mszabo-wikia: I'd create a separate user group for this and add it to only those people who need it

extensions/wikia/Privacy/Privacy.setup.php
rybmat: Do you think we would like to enable this page on other wikis? If not, then using wikifactory is not necessary here

mszabo-wikia: it's more for conventions and removing logic details from config. plus in dev we might not always have community wiki, or might want to test it elsewhere.

lib/Wikia/src/Service/User/Permissions/data/PermissionsDefinesAfterWikiFactory.php
macbre: I don't think that's really necessary here. This special page is meant mostly for testing purposes and will be there until a generic solution for GDPR is created that will cover MediaWiki and other services.

extensions/wikia/Privacy/templates/SpecialRequestToBeForgottenInternal_index.php
```diff
@@ -0,0 +1,11 @@
+<section>
+	<form method="post">
+		<label for="username">Username</label>
+		<input id="username" type="text" name="username" required>
+		<input name="editToken" type="hidden" value="<?= $editToken ?>">
+		<input type="submit" value="Submit">
+	</form>
+	<? if ( !empty( $message ) ): ?>
+		<span><?= $message ?></span>
```
mszabo-wikia: let's escape this just in case

macbre: 9814a3a

mszabo-wikia (approved):

PR #15293, opened 2018-05-14
SUS-4773: Don't use REPLACE for updating edit counts

mszabo-wikia:
A [`REPLACE`](https://dev.mysql.com/doc/refman/5.7/en/replace.html) statement will attempt an `INSERT` then perform a `DELETE`+ `INSERT` in case of a duplicate key error:
>`REPLACE` works exactly like `INSERT`, except that if an old row in the table has the same value as a new row for a `PRIMARY KEY` or a `UNIQUE` index, the old row is deleted before the new row is inserted.

For our use case here this is suboptimal. The most common scenario is a simple `UPDATE` of already existing rows; an `INSERT` can only occur if this was the user's first edit on the given wiki. So let's optimize for that.

https://wikia-inc.atlassian.net/browse/SUS-4773

### DIFF
--- a/includes/wikia/services/UserStatsService.class.php
+++ b/includes/wikia/services/UserStatsService.class.php
@@ -196,13 +196,15 @@ class UserStatsService extends WikiaModel {
 		// SUS-4773: In the vast majority of cases the above UPDATE will have handled setting proper edit count
 		// We only need to INSERT if this is the user's first edit on this wiki
 		if ( !$dbw->affectedRows() ) {
-			$dbw->insert(
+			$dbw->upsert(
 				'wikia_user_properties',
 				[
 					 'wup_user' => $this->userId,
 					 'wup_property' => $statName,
 					 'wup_value' => $statVal
 				 ],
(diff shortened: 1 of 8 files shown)


includes/wikia/services/UserStatsService.class.php
```diff
@@ -96,6 +101,21 @@ public function increaseEditsCount() {
...
+				'lastContributionTimestamp' => null,
+				'editcount' => 0,
+				'editcountThisWeek' => 0
+			];
+		}
+
+		// SUS-4773: Memoize user stats to remember them in the context of a single request and avoid network overhead
+		if ( isset( $this->cache[$this->wikiId][$this->userId] ) ) {
+			return $this->cache[$this->wikiId][$this->userId];
+		}
```
macbre: Memcache client does that already if the same key is accessed multiple times (and there's no memcache update from the same request).

mszabo-wikia: 👍 true, I'll remove

r0bj (approved):

macbre (approved):

PR #15295, opened 2018-05-14
PLATFORM-3485: handle full address in REQUEST_URI

themech:
`ApiService::foreignCall` uses local apache as a proxy and the full address ends up in the `REQUEST_URI`

### DIFF
--- a/extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
+++ b/extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
@@ -71,7 +71,9 @@ class WikiFactoryLoader {
 			// normal HTTP request
 			$this->mServerName = strtolower( $server['SERVER_NAME'] );
 
-			$path = parse_url( $server['REQUEST_SCHEME'] . '://' . $server['SERVER_NAME'] . $server['REQUEST_URI'], PHP_URL_PATH );
+			$fullUrl =  preg_match( "/^https?:\/\//", $server['REQUEST_URI'] ) ? $server['REQUEST_URI'] :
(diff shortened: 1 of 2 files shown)


extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
```diff
@@ -71,7 +71,9 @@ public function  __construct( array $server, array $environment, array $wikiFact
 			// normal HTTP request
 			$this->mServerName = strtolower( $server['SERVER_NAME'] );
 
-			$path = parse_url( $server['REQUEST_SCHEME'] . '://' . $server['SERVER_NAME'] . $server['REQUEST_URI'], PHP_URL_PATH );
+			$fullUrl =  preg_match( "/^https?:\/\//", $server['REQUEST_URI'] ) ? $server['REQUEST_URI'] :
```
mszabo-wikia, line 74: I think simple substr check is faster here

themech: with substr you have to check both, https and http. I've compared 100000 calls to preg_match with 100000 calls to `substr($url, 7) === 'http://' || substr($url, 8) === 'https://'`:

```
mech@dev-mech:~$ php https_perf.php 
3.2825589179993E-6 sec/preg_match
6.0459589958191E-6 sec/substr
```
`preg_match` is faster. It makes sense, most of the times `preg_match` knows the result after checking the first character

mszabo-wikia (approved):

Grunny (approved):

PR #15306, opened 2018-05-15
SUS-4781 | WikiFactory tags tab should no longer be present in WikiFactory UI

macbre:
https://wikia-inc.atlassian.net/browse/SUS-4781

Relates to #15304


mszabo-wikia (approved):
🔥

PR #15312, opened 2018-05-15
SUS-4765: check opt-out status for Qualaroo, GA, Comscore and Quantcast

mszabo-wikia:
Check opt-out status for Qualaroo, GA, Comscore and Quantcast before loading external code and remove all tracking pixels that'd have been loaded if JavaScript was disabled.

https://wikia-inc.atlassian.net/browse/SUS-4765


extensions/wikia/AnalyticsEngine/AnalyticsProviderComscore.php
```diff
@@ -16,22 +16,22 @@ function trackEvent($event, $eventDetails=array()){
...
-	_comscore.push({ c1: "2", c2: "'.static::$PARTNER_ID.'",
-	options: {
-		url_append: "'.static::$COMSCORE_KEYWORD_KEYNAME.'='.static::getC7Value().'"
-	}
+require(["wikia.trackingOptOut"], function (trackingOptOut) {
+	trackingOptOut.ifNotOptedOut(function () {
+		window._comscore = window._comscore || [];
+		window._comscore.push({ c1: "2", c2: "' . static::$PARTNER_ID . '",
+			options: {
+				url_append: "' . static::$COMSCORE_KEYWORD_KEYNAME . '=' . static::getC7Value() . '"
```
nandy-andy, line 24: Don't we have to `urlencode()` the whole `url_append` value?

resources/wikia/modules/spec/trackingOptOut.spec.js
```diff
@@ -0,0 +1,94 @@
+describe('Tracking opt-out', function () {
```
nandy-andy, line 1: 💎

nandy-andy (commented):
Just one question ;-) Thanks for unit tests! ❤️

extensions/wikia/AnalyticsEngine/AnalyticsProviderComscore.php
mszabo-wikia: I just copied what's been already there for years 😉 If we find it's broken perhaps we should sunset this feature as unused. 🙈

PR #15316, opened 2018-05-16
SUS-4772 | Remove data for renamed user

latata:
https://wikia-inc.atlassian.net/browse/SUS-4772
@Wikia/sus


extensions/wikia/Privacy/RemoveUserDataController.php
```diff
@@ -77,6 +86,17 @@ private function getUserWikis( int $userId ) {
 		return $specialsDbr->selectFieldValues( 'events_local_users', 'wiki_id', ['user_id' => $userId], __METHOD__, ['DISTINCT'] );
 	}
 
+	private function getFakeUser( $username ) {
```
macbre: What does `fake` is this context mean?

latata: This is how the user, which is created during rename process, is named in the code:
https://github.com/Wikia/app/blob/25b19bc52b237f2c74d8d5a07d4d3e57c2228f37/extensions/wikia/UserRenameTool/RenameUserProcess.class.php#L443-L443

latata: I can add comment about it

extensions/wikia/Privacy/RemoveUserDataController.php
```diff
@@ -28,18 +29,28 @@ public function removeUserData() {
 
 		$this->info( "Right to be forgotten request for user $userId", ['userId' => $userId] );
 
+		$username = User::whoIs( $userId );
+		$fakeUser = $this->getFakeUser($username);
```
pwojcik86, line 33: Do we need the whole user object here? Won't `user_id` be enough?

extensions/wikia/Privacy/RemoveUserDataOnWikiTask.php
```diff
@@ -120,14 +120,18 @@ private function removeUserPagesFromRecentChanges( $userDbKey ) {
...
 		$this->removeCheckUserData( $userId );
 		$this->removeAbuseFilterData( $userId );
 		$this->removeIpFromRecentChanges( $userId );
 
 		$userDbKey = Title::newFromText( $username )->getDBkey();
 		$this->removeUserPages( $userDbKey );
 		$this->removeUserPagesFromRecentChanges( $userDbKey );
+		if ( !empty( $oldUsername ) ) {
+			$oldUserDbKey = Title::newFromText( $oldUsername )->getDBkey();
+			$this->removeUserPages( $oldUserDbKey );
```
pwojcik86: We also need `$this->removeUserPagesFromRecentChanges( ... );`

pwojcik86 (approved):

PR #15324, opened 2018-05-16
SUS-4284 | HubService::getComscoreCategoryOverride - use wgComScoreTagOverride WikiFactory variable instead of legacy WikiFactory tags

macbre:
https://wikia-inc.atlassian.net/browse/SUS-4284

[Values from `comscore_*` tags were migrated](https://gist.github.com/macbre/7c10460a5cb846efe321dfbf3bcae38f) to [`wgComScoreTagOverride` WikiFactory variable](https://community.wikia.com/wiki/Special:WikiFactoryReporter?varid=1829) (497 wikis were affected).

`HubService::getComscoreCategoryOverride` returns the same values.

This was the last piece of the code using `WikiFactoryTags` class - sunset PR to follow.

### DIFF
--- a/includes/wikia/services/HubService.class.php
+++ b/includes/wikia/services/HubService.class.php
@@ -86,21 +86,10 @@ class HubService {
 	/**
 	 * Look for a comscore_zzz tag
 	 * @param integer $cityId
-	 * @return hash of category data { id, name, url, short, deprecated, active }
+	 * @return string|null
 	 */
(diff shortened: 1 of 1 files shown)


mszabo-wikia (changes requested):
It seems all usages of this method end up passing the current wiki ID at some point. Maybe we can either simplify the signature, or just check if $cityId == $wgCityId and return already loaded global in that case.

mszabo-wikia (approved):

PR #15326, opened 2018-05-16
SUS-4789 | force the value of wgEnableCategorySelectExt only when it is not set yet

macbre:
https://wikia-inc.atlassian.net/browse/SUS-4789

Allow WikiFactory to set this variable to `false`.

### Before the fix

```
macbre@cron-s1:~$ ./city_eval.sh 3035
LBFactory_Wikia::getSectionForWiki: section c2, wiki
> var_dump($wgEnableCategorySelectExt)
bool(true)
```

### After the fix

```
macbre@cron-s1:~$ ./city_eval.sh 3035
LBFactory_Wikia::getSectionForWiki: section c2, wiki
> var_dump($wgEnableCategorySelectExt)
bool(false)
```

### DIFF
--- a/includes/wikia/Extensions.php
+++ b/includes/wikia/Extensions.php
@@ -711,7 +711,7 @@ if( !empty( $wgEnableHAWelcomeExt ) ) {
 }
 
 // Enable CategorySelect extension for all not RTL wikis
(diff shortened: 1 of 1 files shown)


mszabo-wikia (approved):

macbre:
Backported to prod in 3aca400

PR #15327, opened 2018-05-16
PLATFORM-3487: special page for GDPR data portability

themech:


extensions/wikia/DownloadYourData/DownloadYourData.i18n.php
```diff
@@ -0,0 +1,24 @@
+<?php
+$messages = array();
+
+$messages['en'] = array(
+	'downloadyourdata' => 'Download Your Data',
+	'downloadyourdata-desc' => 'Allows users to download their data',
+	'downloadyourdata-intro' => 'Click the button below to download a .CSV copy of your FANDOM data. The data we provide is username, user ID, real name (if provided), email address, account creation date and a link to a list of your contributions.',
+	'downloadyourdata-button-text' => 'Download data',
+	'downloadyourdata-username-placeholder' => 'Optional username',
+	'downloadyourdata-user-not-found' => 'User <i>$1</i> not found',
```
Grunny: Since this message is parsed, it might be better to do the italics with wikitext which will be a little easier for translators (i.e. `''$1''`).

extensions/wikia/DownloadYourData/DownloadYourData.i18n.php
```diff
@@ -0,0 +1,24 @@
+<?php
+$messages = array();
+
+$messages['en'] = array(
```
Grunny, line 4: We should add `qqq` message documentation.

extensions/wikia/DownloadYourData/DownloadYourDataSpecialController.class.php
```diff
@@ -0,0 +1,85 @@
...
+	public function __construct() {
+		parent::__construct( 'DownloadYourData' );
+	}
+
+	public function index() {
+		wfProfileIn( __METHOD__ );
+
+		$user = $this->getUser();
+		if ( !$user->isLoggedIn() ) {
+			$this->getOutput()->redirect( \SpecialPage::getTitleFor( 'Contact/data-portability' )->getFullURL() );
```
Grunny: Minor: you can use the subpage parameter here to match how MW core would usually do it, i.e. `\SpecialPage::getTitleFor( 'Contact', 'data-portability' )`, but both work.

extensions/wikia/DownloadYourData/DownloadYourDataSpecialController.class.php
```diff
@@ -0,0 +1,85 @@
...
+
+	const CSV_FILE_NAME = 'fandom_account_data.csv';
+	const CSV_CONTENT_TYPE = 'text/csv; charset=UTF-8';
+
+	public function __construct() {
+		parent::__construct( 'DownloadYourData' );
+	}
+
+	public function index() {
+		wfProfileIn( __METHOD__ );
```
Grunny: The `wfProfile*` methods can be removed, we don't actually use them any more.

extensions/wikia/DownloadYourData/templates/DownloadYourDataSpecialController_index.mustache
```diff
@@ -0,0 +1,14 @@
+<p>{{{introText}}}</p>
```
Grunny: This value should be escaped, since the output format of the message is `->text()` so is not previously escaped.

lib/Wikia/src/Service/User/Permissions/data/PermissionsDefinesAfterWikiFactory.php
```diff
@@ -641,3 +641,5 @@
 $wgGroupPermissions['*']['fandom-admin'] = false;
 $wgGroupPermissions['staff']['fandom-admin'] = true;
 $wgGroupPermissions['fandom-editor']['fandom-admin'] = true;
+
+$wgGroupPermissions['staff']['exportuserdata'] = true;
```
Grunny: I think we should probably make this `util` rather than `staff` to be consistent with similar user data features being limited to `util` staff members.

Grunny (changes requested):

extensions/wikia/DownloadYourData/DownloadYourDataSpecialController.class.php
themech: done

lib/Wikia/src/Service/User/Permissions/data/PermissionsDefinesAfterWikiFactory.php
themech: 👍

extensions/wikia/DownloadYourData/templates/DownloadYourDataSpecialController_index.mustache
themech: good point, done for `introText`, `error` message and submit button text

extensions/wikia/DownloadYourData/DownloadYourData.i18n.php
themech: oh, right, forgot about those

Grunny (approved):

PR #15331, opened 2018-05-17
(MAIN-13789) Quick fix for communities with custom forum namespaces

Grunny:
Some older communities override wgExtraNamespaces in WikiFactory in order
to set a custom Forum: namespace ID. As such, the default forum namespace
setting for English wikis at least needs to be set before WikiFactory
overrides it. After the config changes, the more common forum namespace
ID was always getting appended to wgExtraNamespaces causing those communities
to have two forum namespaces defined, with only one taking effect.

This fixes it by moving the default back to before WikiFactory takes over
until we can consider cleaning up the situation so it's more consistent.

/cc @mixoteric

### DIFF
--- a/includes/wikia/Extensions.php
+++ b/includes/wikia/Extensions.php
@@ -94,12 +94,6 @@ switch ( $wgLanguageCode ) {
 		$wgNamespaceAliases["Forum_talk"] = 111;
 
 		break;
-            
-        default:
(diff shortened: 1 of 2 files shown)


wwarren (approved):

wwarren:
LGTM

ljagiello (approved):

PR #15332, opened 2018-05-17
SUS-4790 / SUS-4791 | respect wgEnableHAWelcomeExt value coming from WikiFactory

macbre:
https://wikia-inc.atlassian.net/browse/SUS-4790 and https://wikia-inc.atlassian.net/browse/SUS-4791

And a fix a notice from a previous fix - #15326

### DIFF
--- a/includes/wikia/Extensions.php
+++ b/includes/wikia/Extensions.php
@@ -705,7 +705,7 @@ if( !empty( $wgEnableHAWelcomeExt ) ) {
 }
 
 // Enable CategorySelect extension for all not RTL wikis
-if (!in_array($wgLanguageCode, array('ar', 'fa', 'he', 'ps', 'yi')) && is_null( $wgEnableCategorySelectExt) ) {
+if (!in_array($wgLanguageCode, array('ar', 'fa', 'he', 'ps', 'yi')) && !isset($wgEnableCategorySelectExt) ) {
     $wgEnableCategorySelectExt = true;
 }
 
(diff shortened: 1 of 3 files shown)


ghost (approved):

includes/wikia/Extensions.php
```diff
@@ -705,7 +705,7 @@
 }
 
 // Enable CategorySelect extension for all not RTL wikis
-if (!in_array($wgLanguageCode, array('ar', 'fa', 'he', 'ps', 'yi')) && is_null( $wgEnableCategorySelectExt) ) {
+if (!in_array($wgLanguageCode, array('ar', 'fa', 'he', 'ps', 'yi')) && !isset($wgEnableCategorySelectExt) ) {
```
mszabo-wikia, line 708: I believe this won't work if `wgEnableCategorySelectExt` is explicitly set to `false` in WikiFactory.

mszabo-wikia (dismissed):

includes/wikia/Extensions.php
macbre:
```php
> $a=false;

> var_dump( isset($a))
bool(true)
```

ghost: Gotta love PHP. So `empty()` perhaps?

mszabo-wikia: My bad, I misremembered—this happens when you set an array entry to `null`, then it'll be considered "not set".

PR #15336, opened 2018-05-17
PLATFORM-3491: enable HTTPS on wikis with ID greater than given threshold

harnash:
Still final WIKI_ID is to be decided.

### DIFF
--- a/extensions/wikia/HTTPSOptIn/HTTPSOptInHooks.class.php
+++ b/extensions/wikia/HTTPSOptIn/HTTPSOptInHooks.class.php
@@ -18,8 +18,11 @@ class HTTPSOptInHooks {
 	const WIKI_ID_THRESHOLD_WITH_HTTPS_ON = 1700000;
 
 	public static function onGetPreferences( User $user, array &$preferences ): bool {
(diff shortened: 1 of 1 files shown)


themech (approved):

extensions/wikia/HTTPSOptIn/HTTPSOptInHooks.class.php
```diff
@@ -13,6 +13,10 @@ class HTTPSOptInHooks {
 		'wikia' => [ 'Community_Central:Licensing' ]
 	];
 
+	// wikis with ID greater than this threshold will get HTTPS on by default for
+	// logged-in users despite the opt-in preference
+	const WIKI_ID_THRESHOLD_WITH_HTTPS_ON = 1700000;
+
 	public static function onGetPreferences( User $user, array &$preferences ): bool {
 		global $wgServer, $wgHTTPSForLoggedInUsers;
 		if ( empty( $wgHTTPSForLoggedInUsers ) && wfHttpsAllowedForURL( $wgServer ) ) {
```
Grunny: I think we should check the threshold here as well to not surface the preference on wikis where it won't have an effect, as we do for `wgHTTPSForLoggedInUsers`.

PR #15344, opened 2018-05-18
PLATFORM-3489: fix article path for language wikis

themech:
These lines were removed in https://github.com/Wikia/app/pull/15169 but somehow found their way back during the config refactoring


Grunny (approved):

harnash:
Tested and approved.

PR #15363, opened 2018-05-21
ADEN-7187 Enabling Consent Management Prebid module

dmnsobczak:


fraszczakszymon (approved):

extensions/wikia/TrackingOptIn/js/trackingOptIn.js
```diff
@@ -44,9 +60,54 @@ define('wikia.trackingOptIn', [
 		window.Wikia.consentQueue.push(callback);
 	}
 
+	function geoRequiresTrackingConsent() {
+		return geoRequiresConsent;
+	}
+
+	function getPrebidGlobalVendorList() {
```
fraszczakszymon: Do we really need to request this list on each pv? Does it really matter for us that this list is changing? We have hardcoded list of enabled providers (like `10,32,52,69,76`). Can we "cache"/hardcode it and use it as a module?

extensions/wikia/TrackingOptIn/js/trackingOptIn.js
```diff
@@ -1,11 +1,25 @@
...
 	'wikia.instantGlobals',
 	'wikia.lazyqueue',
 	'wikia.log',
 	'wikia.trackingOptInModal'
-], function (instantGlobals, lazyQueue, log, trackingOptInModal) {
+], function (cookies, instantGlobals, lazyQueue, log, trackingOptInModal) {
 	var optIn = false,
-		logGroup = 'wikia.trackingOptIn';
+		geoRequiresConsent = true,
+		instance = null,
```
fraszczakszymon: It seems that we don't need this variable in this scope

extensions/wikia/TrackingOptIn/js/trackingOptIn.js
```diff
@@ -44,9 +60,54 @@ define('wikia.trackingOptIn', [
...
+	function geoRequiresTrackingConsent() {
+		return geoRequiresConsent;
+	}
+
+	function getPrebidGlobalVendorList() {
+		var request = new XMLHttpRequest();
+
+		log('Requesting IAB Global Vendor List', log.levels.debug, logGroup);
+
+		request.open('GET', prebidVendorListUrl, false);
```
fraszczakszymon: I have doubts about requesting this resource directly from `consensu.org`. Also - what happens in case of failure/adblock?

extensions/wikia/TrackingOptIn/js/trackingOptIn.js
```diff
@@ -44,9 +60,54 @@ define('wikia.trackingOptIn', [
...
+		if (cookie) {
+			log('Serving consent string from cookie', log.levels.debug, logGroup);
+			return cookie;
+		}
+
+		var consentString,
+			consentData = new trackingOptInModal.ConsentString();
+
+		consentData.setGlobalVendorList(getPrebidGlobalVendorList());
+		consentData.setPurposesAllowed(optIn ? prebidPurposesList : []);
```
fraszczakszymon: Can we create this array based on `prebidVendorsList`?

extensions/wikia/TrackingOptIn/js/trackingOptIn.js
```diff
@@ -44,9 +60,54 @@ define('wikia.trackingOptIn', [
...
+		consentData.setVendorsAllowed(optIn ? prebidVendorsList : []);
+
+		consentString = consentData.getConsentString();
+
+		cookies.set(prebidConsentStringCookie, consentString, {
+			path: '/',
+			domain: window.wgCookieDomain,
```
fraszczakszymon: Does not work on mobile

extensions/wikia/TrackingOptIn/js/trackingOptIn.js
```diff
@@ -44,9 +60,54 @@ define('wikia.trackingOptIn', [
...
+
+		log('Requesting IAB Global Vendor List', log.levels.debug, logGroup);
+
+		request.open('GET', prebidVendorListUrl, false);
+		request.send(null);
+
+		return JSON.parse(request.responseText);
+	}
+
+	function getPrebidConsentString(optIn) {
```
fraszczakszymon: Can we move building consent string to trackingOptIn module? Does it make sense? Are we going to reuse it somewhere else?

extensions/wikia/TrackingOptIn/js/trackingOptIn.js
```diff
@@ -44,9 +60,54 @@ define('wikia.trackingOptIn', [
...
+		var cookie = cookies.get(prebidConsentStringCookie);
+
+		if (cookie) {
```
fraszczakszymon: Should we reset cookie once user change decision about opt in?

extensions/wikia/AdEngine/js/lookup/prebid/prebid.js
```diff
@@ -54,7 +54,27 @@ define('ext.wikia.adEngine.lookup.prebid', [
 				return;
 			}
 
-			log('User opt-in for prebid', log.levels.info, logGroup);
+			win.__cmp = function(command, version, callback) {
```
fraszczakszymon, line 59: We can define this function out of `call` function scope, am I right?

dmnsobczak: No, because we need optIn variable which is available here. Also, we don't need it if prebid is not enabled = no call will be performed.

extensions/wikia/TrackingOptIn/js/trackingOptIn.js
dmnsobczak: Agree, removed/moved.

extensions/wikia/TrackingOptIn/js/trackingOptIn.js
dmnsobczak: We don't request it every PV because consent string is cached for 7 days, and then request is not being performed.

extensions/wikia/TrackingOptIn/js/trackingOptIn.js
dmnsobczak: Turns out requesting directly is not a good idea - this server is often unstable. List trimmed and copied into our code.

extensions/wikia/TrackingOptIn/js/trackingOptIn.js
dmnsobczak: Done.

extensions/wikia/TrackingOptIn/js/trackingOptIn.js
dmnsobczak: Right, I will reset cookie if optIn value will be different than the one saved in consent string.

extensions/wikia/TrackingOptIn/js/trackingOptIn.js
dmnsobczak: Right, thanks. Fixed.

extensions/wikia/TrackingOptIn/js/trackingOptIn.js
dmnsobczak: We could, IAB consent string is a standard and other non-prebid services can (I think) get info about consent in this format, but there's nothing bad with moving it to separate module. Done.

extensions/wikia/TrackingOptIn/js/trackingOptInProxy.js
```diff
@@ -41,20 +41,29 @@ define('wikia.trackingOptIn', [
...
 
-	function isOptedIn() {
-		log(['isOptedIn - proxy', optIn], log.levels.info, logGroup);
-		return optIn;
+	function geoRequiresTrackingConsent() {
+		if (win.M === 'undefined' || typeof win.M.continent === 'undefined') {
+			return true;
+		}
+
+		return win.M.continent === 'EU';
```
fraszczakszymon: Is there any chance to get real boolean from trackingOptIn module loaded in mobile-wiki?

dmnsobczak: Done.

fraszczakszymon (approved):

PR #15366, opened 2018-05-21
PLATFORM-3462: removing old sitemap extension

themech:


themech:
part of that extension is still used, I'll try less brutal approach

PR #15367, opened 2018-05-21
PLATFORM-3462: remove old sitemaps code

themech:


extensions/wikia/SitemapXml/SpecialSitemap_body.php
```diff
@@ -0,0 +1,71 @@
...
+		global $wgOut;
+
+		// keeping the checks for old sitemaps for now so we return 404s
+		$showIndex = strpos( $subpage, '-index.xml' ) !== false;
+
+		$forceOldSitemap = strpos( $subpage, '-oldsitemapxml-' ) !== false;
+		$forceNewSitemap = strpos( $subpage, '-newsitemapxml-' ) !== false;
+
+		if ( ( $showIndex && !$forceOldSitemap ) || $forceNewSitemap ) {
+			$wgOut->disable();
```
Grunny: We can avoid the global with `$this->getOutput()->disable();`

extensions/wikia/SitemapXml/SpecialSitemap_body.php
```diff
@@ -0,0 +1,71 @@
...
+
+	/**
+	 * show 404
+	 *
+	 * @access private
+	 */
+	private function print404() {
+		global $wgOut;
+
+		$wgOut->disable();
```
Grunny: Same here.

Grunny (approved):
Minor comment, otherwise looks good.

extensions/wikia/SitemapXml/SpecialSitemap_body.php
themech: Done. In my defence, I just moved this file :)

extensions/wikia/SitemapXml/SpecialSitemap_body.php
themech: ditto

PR #15381, opened 2018-05-22
SUS-4812 | check opt-in status in the internal tracker 🇪🇺

mszabo-wikia:
https://wikia-inc.atlassian.net/browse/SUS-4812

### DIFF
--- a/extensions/wikia/Track/templates/track.mustache
+++ b/extensions/wikia/Track/templates/track.mustache
@@ -1,69 +1,77 @@
 <!-- Wikia Beacon Tracking -->
-<noscript><img src="{{{url}}}&amp;nojs=1" width="1" height="1" border="0" alt="" /></noscript>
 <script>
-(function() {
-	function genUID() {
-		return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, function(c) {
-			var r = Math.random() * 16 | 0, v = c === 'x' ? r : (r & 0x3 | 0x8);
-			return v.toString(16);
+	require(['wikia.trackingOptIn'], function (trackingOptIn) {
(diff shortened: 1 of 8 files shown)


extensions/wikia/Track/templates/track.mustache
```diff
@@ -1,69 +1,77 @@
...
+					'&session_id=' + window.sessionId + '&pv_unique_id=' + window.pvUID + '&pv_number=' + window.pvNumber +
+					'&pv_number_global=' + window.pvNumberGlobal;
 
-	script.src = trackUrl;
-	document.head.appendChild(script);
-})();
+			if (optIn) {
+				trackUrl += '&u=' + window.wgTrackID;
+			} else {
+				trackUrl += '&u=0';
```
rogatty: It should be `u=-1`

skins/oasis/modules/templates/Oasis_Index.php
```diff
@@ -65,6 +65,8 @@
 	<div class="background-image-gradient"></div>
 <? endif ?>
 
+<?= $internalTracker ?>
```
rogatty, line 68: Shouldn't this be loaded after `<?= $trackingOptIn ?>`? I'm not sure what are the dependencies.

mszabo-wikia: the module proper is loaded in the head, the code below will init it to show the modal, so we can lazy queue the tracker as callback to be fired after the user's consent status has been obtained

extensions/wikia/Track/templates/track.mustache
mszabo-wikia: addressed in 0f77ce8bd3f

rogatty (approved):

PR #15386, opened 2018-05-23
PLATFORM-3193: make inter-wiki links protocol relative

harnash:


themech (changes requested):
We should exclude wikis which are not https-ready (`wfHttpsAllowedForURL`)

Grunny (approved):

themech (approved):

PR #15405, opened 2018-05-24
SUS-4820 | Break dependency between mobile-wiki and old WikiaMobile comments

mszabo-wikia:
mobile-wiki relies on old WikiaMobile entry point to get article comment data, but it actually doesn't use the html templating there. Let's break this dependency by using `ArticleCommentList` instead. This approach is also more efficient as we save some database queries for every single individual comment when fetching data.

https://wikia-inc.atlassian.net/browse/SUS-4820

### DIFF
--- a/extensions/wikia/MercuryApi/models/MercuryApi.class.php
+++ b/extensions/wikia/MercuryApi/models/MercuryApi.class.php
@@ -16,20 +16,6 @@ class MercuryApi {
 	 */
 	private $users = [];
 
-	/**
-	 * @desc Fetch Article comments count
-	 *
-	 * @param Title $title - Article title
-	 *
-	 * @return integer
-	 */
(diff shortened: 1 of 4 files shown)


extensions/wikia/MercuryApi/models/MercuryApi.class.php
```diff
@@ -16,20 +16,6 @@ class MercuryApi {
...
 	private $users = [];
 
-	/**
-	 * @desc Fetch Article comments count
-	 *
-	 * @param Title $title - Article title
-	 *
-	 * @return integer
-	 */
-	public function articleCommentsCount( Title $title ) {
```
mszabo-wikia, line 26: not used

rybmat (approved):

hakubo (approved):

PR #15418, opened 2018-05-25
SUS-4421 | Avoid timeouts in background tasks

mszabo-wikia:
Set proper timeout also in Apache request handler.

https://wikia-inc.atlassian.net/browse/SUS-4421


macbre (approved):

ljagiello:
@mszabo-wikia for sure this will reduce the problem, one question from me, why the task never ends in the first place?

PR #15422, opened 2018-05-25
IRIS-4844 | Return meaningful error when DiscussionsActivator fails

rogatty:
## Links
https://wikia-inc.atlassian.net/browse/IRIS-4844

## Description
Instead of an error page with status code 200, let's return JSON with error details and appropriate status code.

### DIFF
--- a/extensions/wikia/Discussions/controllers/EnableDiscussionsController.class.php
+++ b/extensions/wikia/Discussions/controllers/EnableDiscussionsController.class.php
@@ -12,21 +12,32 @@ class EnableDiscussionsController extends \WikiaController {
 	}
 
 	public function activateDiscussions() {
-		$cityId = $this->request->getInt( self::SITE_ID );
-		if ( empty( $cityId ) ) {
-			throw new BadRequestException();
-		}
+		try {
+			$cityId = $this->request->getInt( self::SITE_ID );
+			if ( empty( $cityId ) ) {
+				throw new BadRequestException();
+			}
 
-		$wiki = WikiFactory::getWikiByID( $cityId );
-		if ( empty( $wiki ) ) {
-			throw new NotFoundException();
-		}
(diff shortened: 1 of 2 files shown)


extensions/wikia/Discussions/controllers/EnableDiscussionsController.class.php
```diff
@@ -8,24 +8,36 @@ class EnableDiscussionsController extends \WikiaController {
...
-			$wiki->city_lang ) )->activateDiscussions();
+			( new \DiscussionsActivator( $wiki->city_id, $wiki->city_title,
+				$wiki->city_lang ) )->activateDiscussions();
 
-		if ( $this->request->getBool( self::CREATE_WELCOME_POST, false ) ) {
-			( new \StaffWelcomePoster() )->postMessage( $wiki->city_id, $wiki->city_lang );
+			if ( $this->request->getBool( self::CREATE_WELCOME_POST, false ) ) {
+				( new \StaffWelcomePoster() )->postMessage( $wiki->city_id, $wiki->city_lang );
+			}
+		} catch ( WikiaBaseException $exception ) {
```
xkxd, line 32: what method will cause such exception? is it necessary to try-catch the whole code?

xkxd (approved):

extensions/wikia/Discussions/controllers/EnableDiscussionsController.class.php
rogatty: `DiscussionsActivator::activateDiscussions` throws `WikiaException` and the exceptions thrown above extend `WikiaHttpException`. Both classes extend `WikiaBaseException`.

I explained the reason to try-catch everything in https://github.com/Wikia/app/pull/15422/commits/231f607c0ebe0614db48ce1bb81974215c5acbf2

PR #15423, opened 2018-05-25
XF-84 remove unused parameter from function definition

jsutterfield:
Ticket: https://wikia-inc.atlassian.net/browse/XF-84

### DIFF
--- a/extensions/wikia/SpecialUnsubscribe/Unsubscribe.body.php
+++ b/extensions/wikia/SpecialUnsubscribe/Unsubscribe.body.php
@@ -94,7 +94,7 @@ class UnsubscribePage extends UnlistedSpecialPage {
 			$this->procUnsub( $email );
 		} else {
 			#this is 1st pass, give them a button to push
-			$this->showInfo( $email , $token);
+			$this->showInfo( $email , $token );
 		}
 
 	}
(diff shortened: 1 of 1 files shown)


extensions/wikia/SpecialUnsubscribe/Unsubscribe.body.php
```diff
@@ -94,7 +94,7 @@ public function execute( $subpage ) {
 			$this->procUnsub( $email );
 		} else {
 			#this is 1st pass, give them a button to push
-			$this->showInfo( $email , $token);
+			$this->showInfo( $email , $token );
```
jsutterfield, line 97: Only place this method is called.

extensions/wikia/SpecialUnsubscribe/Unsubscribe.body.php
```diff
@@ -131,7 +131,7 @@ private function email2users($email) {
 	 * @param $email string: email whose info we're looking up
 	 * @param $token string: token from url, to verify this link
 	 */
-	function showInfo( $email, $token, $timestamp ) {
```
jsutterfield, line 134: `$timestamp` isn't used inside this method.

rogatty (approved):

PR #15428, opened 2018-05-28
SUS-4784 | Allow to create wikis with the language code in the path via CNW

mszabo-wikia:
Allow to create non-English wikis with the language code portion in the URL path, rather than a 2nd level subdomain. For now this is only enabled on sandboxes.

https://wikia-inc.atlassian.net/browse/SUS-4784


extensions/wikia/CreateNewWiki/CreateNewWiki_setup.php
```diff
@@ -18,7 +18,7 @@
 $wgAutoloadClasses['CreateWiki'] = $dir."/CreateWiki.php";
 $wgAutoloadClasses['CreateWikiChecks'] = $dir."/CreateWikiChecks.php";
 $wgAutoloadClasses['SpecialCreateNewWiki'] = $dir . 'SpecialCreateNewWiki.class.php';
-$wgAutoloadClasses['CreateNewWikiHooks'] = __DIR__ . '/CreateNewWikiHooks.class.php';
```
macbre, line 21: Why no `.class` suffix?

extensions/wikia/CreateNewWiki/CreateNewWikiHooks.php
```diff
@@ -0,0 +1,13 @@
...
+
+class CreateNewWikiHooks {
+
+	public static function onMakeGlobalVariablesScript( array &$vars, OutputPage $outputPage ) {
+		if ( $outputPage->getTitle()->isSpecial( 'CreateNewWiki' ) ) {
+			global $wgWikiaBaseDomain, $wgCreateLanguageWikisWithPath;
+
+			$vars['wgWikiaBaseDomain'] = $wgWikiaBaseDomain;
+			$vars['wgCreateLanguageWikisWithPath'] = $wgCreateLanguageWikisWithPath;
+		}
```
macbre, line 11: `OutputPage::addJsConfigVars` allows you to emit JS variables when rendering a special page. No need to register a "generic" hook with a check for a specific special page inside.

macbre (approved):

extensions/wikia/CreateNewWiki/CreateNewWiki_setup.php
mszabo-wikia: It's not really needed

extensions/wikia/CreateNewWiki/js/WikiBuilder.js
```diff
@@ -300,14 +304,22 @@ define('ext.createNewWiki.builder', ['ext.createNewWiki.helper', 'wikia.tracker'
...
+			if (selected && selected !== NO_SUBDOMAIN_LANGUAGE) {
+				wikiBaseDomain.text(wikiaBaseDomain + '/' + selected);
+			} else {
+				wikiBaseDomain.text(wikiaBaseDomain);
+			}
 		} else {
-			wikiDomainCountry.html('');
-		}
+			if (selected && selected !== NO_SUBDOMAIN_LANGUAGE) {
+				wikiDomainCountry.html(selected + '.');
```
Grunny, line 315: This should be `.text()` as a best practice since `selected` was retrieved from the DOM using `.val()`, so it is plain text and should be put back as plain text.

extensions/wikia/CreateNewWiki/CreateWikiChecks.php
```diff
@@ -29,33 +29,36 @@ class CreateWikiChecks {
 	 *
 	 * @return integer - 0 or 1
```
macbre, line 30: It's bool actually

macbre (approved):

PR #15433, opened 2018-05-28
PLATFORM-3514 | Set local sitemap URL with language

jmail:


includes/wikia/models/DesignSystemGlobalFooterModel.class.php
```diff
@@ -546,8 +546,12 @@ private function getLicenseUrl() {
 		return $licenseUrl;
 	}
 
+	private function getLocalHref( $hrefKey ) {
+		return DesignSystemSharedLinks::getInstance()->getLocalHref($hrefKey, $this->lang, $this->productInstanceId);
```
Grunny: Minor: There should be spaces between the brackets per coding conventions.

includes/wikia/models/DesignSystemGlobalFooterModel.class.php
```diff
@@ -546,8 +546,12 @@ private function getLicenseUrl() {
...
+	}
+
 	private function getHref( $hrefKey ) {
-		return DesignSystemSharedLinks::getInstance()->getHref( $hrefKey, $this->lang );
+		return DesignSystemSharedLinks::getInstance()->getHref($hrefKey, $this->lang);
```
Grunny: Minor: Same here, there should be spaces between the brackets per coding conventions, so can just undo this change.

includes/wikia/models/DesignSystemSharedLinks.class.php
```diff
@@ -33,6 +33,24 @@ public function getHref( $name, $lang ) {
...
+	 * @param $lang string two letter language code
+	 * @param $cityId int
+	 * @return string full URL, in case of lang specific URL missing, default one is returned
+	 */
+	public function getLocalHref( $name, $lang, $cityId) {
+		$lang = $this->getLangWithFallback( $lang );
+
+		$href = $this->hrefs[$lang][$name] ?? $this->hrefs['default'][$name];
+
+		return WikiFactory::getLocalEnvURL(
```
Grunny: Since this is a local URL, `WikiFactory::getLocalEnvURL` is unnecessary.

includes/wikia/models/DesignSystemSharedLinks.class.php
```diff
@@ -33,6 +33,24 @@ public function getHref( $name, $lang ) {
...
+			WikiFactory::cityUrlToLanguagePath(
+				WikiFactory::cityIDtoUrl($cityId)
```
Grunny: Minor: missing spaces.

includes/wikia/models/DesignSystemSharedLinks.class.php
```diff
@@ -33,6 +33,24 @@ public function getHref( $name, $lang ) {
...
+	 * @param $cityId int
+	 * @return string full URL, in case of lang specific URL missing, default one is returned
+	 */
+	public function getLocalHref( $name, $lang, $cityId) {
+		$lang = $this->getLangWithFallback( $lang );
+
+		$href = $this->hrefs[$lang][$name] ?? $this->hrefs['default'][$name];
+
+		return WikiFactory::getLocalEnvURL(
+			WikiFactory::cityUrlToLanguagePath(
```
Grunny: Since we're now using this in a few places (and may start using this in more), I wonder if it makes sense to have a helper method in WikiFactory to get this for an ID, what do you think?

Grunny (changes requested):

extensions/wikia/WikiFactory/WikiFactory.php
```diff
@@ -3409,6 +3409,17 @@ static public function cityIDtoDomain( $city_id, $master = false ) {
...
 
+	/**
+	 * Returns local address of href
+	 *
+	 * @param int $city_id
+	 * @param string $href
+	 * @return string
+	 */
+	static public function cityIdToLanguagePath( $city_id, $href ) {
+		return static::cityUrlToLanguagePath( WikiFactory::cityIDtoUrl( $city_id ) ) . $href;
```
Grunny: I'd use `static::cityIDtoUrl` here rather than spelling out the class, and I'd suggest removing the `$href` parameter so that this method is usable for other use cases and matches similar methods like `cityIDtoUrl` and just returns the language path.

Grunny (changes requested):

Grunny (approved):

PR #15436, opened 2018-05-29
ADEN-7215 Fixing getting vastContent for wikia adapters

dmnsobczak:


extensions/wikia/AdEngine/js/template/porvata.js
```diff
@@ -185,8 +185,8 @@ define('ext.wikia.adEngine.template.porvata', [
 
 		if (prebid && params.hbAdId) {
 			params.bid = prebid.getBidByAdId(params.hbAdId);
-			params.vastResponse = params.bid.vastContent || null;
-			params.vastUrl = params.bid.vastUrl;
+			params.vastResponse = params.bid && params.bid.vastContent ? params.bid.vastContent : null;
```
fraszczakszymon, line 188: How is it possible that there is `params.hbAdId` but there is no `prebid.getBidByAdId(params.hbAdId)`? Are you sure that this fix isn't hiding real problem?

dmnsobczak: You right, It was a bug in Prebid.js library. Till fixed I applied fix + performance change.

fraszczakszymon (approved):

PR #15439, opened 2018-05-29
🐖 | Make tests great again 🚚

mszabo-wikia:
Drag the tests, kicking and screaming, into Travis CI so they can be executed in a containerized, isolated environment.

A tiny number of tests have dependencies that cannot be replicated inside the container (such as `SwiftStorageTest` calling the aforementioned storage, or ImageServing tests assuming the presence of certain wiki articles with embedded specific images). For now these tests have been annotated with `@group Infrastructure` and are excluded from Travis runs—it should be possible to run them manually on devbox, or via some Jenkins job.

Additionally, all `@group Broken` tests were either fixed and re-enabled or removed, and the handling of the group itself was removed from our PHPUnit setup. In the future, failing tests should either be fixed, or removed, or at least disabled via PHPUnit `markTestSkipped` feature so they actually show up in the test run output as skipped.

Finally, `WikiaBaseTest` class has been deprecated because its main purpose was to provide a wrapper around monkey patching (e.g. mocking static methods). As relying on such monkey patching means tests might break en masse with the next major PHP version, it's preferred to either use integration tests instead, or refactor the offending code.
@Wikia/sus


extensions/wikia/UserRenameTool/tests/RenameUserSetupChecksTest.php
```diff
@@ -25,18 +26,23 @@ function testSpoofUser( string $destName, bool $expectedResult ) {
 		$this->mockStaticMethod( PhalanxServiceFactory::class, 'getServiceInstance', $phalanxService );
 
 		// perform initial checks for user rename process
-		$this->assertEquals( $expectedResult, $process->setup() );
+		$actualResult = $process->setup();
 
 		if ( $expectedResult ) {
+			var_dump($process->getWarnings() );
+			var_dump($process->getErrors());
```
macbre: var_dump :)

macbre (approved):

extensions/wikia/Privacy/tests/UserDataRemoverTest.php
```diff
@@ -92,12 +107,17 @@ public function testUserDataShouldBeRemovedFromUserEmailLogTable() {
 	}
 
 	public function testUserDataShouldBeRemovedFromUserPropertiesTable() {
+		$this->heliosClientMock->expects( $this->once() )
```
pwojcik86: We don't need to check if Helios is called; it's done by the rtbf service already. The only reason this happens is because we use `Wikia::invalidateUser`

extensions/wikia/Privacy/tests/fixtures/audit_log.sql
```diff
@@ -1,19 +1,20 @@
-DROP TABLE IF EXISTS rtbf_log;
-CREATE TABLE rtbf_log (
-  id int PRIMARY KEY AUTOINCREMENT,
+SET foreign_key_checks = 0;
+
+CREATE TABLE IF NOT EXISTS rtbf_log (
+  id int PRIMARY KEY AUTO_INCREMENT,
```
pwojcik86, line 4: I thought this didn't work with sqllite

pwojcik86 (changes requested):

extensions/wikia/Privacy/tests/fixtures/audit_log.sql
mszabo-wikia: There's no SQLite any more, MySQL is used :)

extensions/wikia/Privacy/tests/UserDataRemoverTest.php
mszabo-wikia: sure, I can change it to do no verification—but the stubbing needs to stay there to avoid an HTTP call to a nonexistent service

macbre:
@mszabo-wikia - rotten pull request warning :smile:

PR #15447, opened 2018-05-30
ADEN-6973 Set single size for UAP BLB

v3rth:


extensions/wikia/AdEngine/js/provider/gpt/adSizeFilter.js
```diff
@@ -9,7 +9,12 @@ define('ext.wikia.adEngine.provider.gpt.adSizeFilter', [
 	'use strict';
 
 	var logGroup = 'ext.wikia.adEngine.provider.gpt.adSizeFilter',
-		minSkinWidth = 1240;
+		minSkinWidth = 1240,
+		BLBSize = {
+			desktop: [[728, 90], [3, 3]],
+			desktopSpecial: [[3, 3]],
```
nandy-andy, line 15: What do you think of renaming `desktopSpecial` to `desktopFanTakeover`? This way the special case is demystified ;-) It explicitly says what's the special scenario

nandy-andy (approved):

extensions/wikia/AdEngine/js/provider/gpt/adSizeFilter.js
v3rth: Both cases are for fan takeover :( 
https://github.com/Wikia/app/pull/15447/files#diff-da9b3451abe4f38d85a4ffd28397af89R63

nandy-andy: Yeah, and once we've got it live site-wide we'll remove the IG and we'll be able to simplify the code.

PR #15462, opened 2018-06-04
XW-5001 | simplify response for global nav design system api

rybmat:
https://wikia-inc.atlassian.net/browse/XW-5001

@Wikia/x-wing


includes/wikia/models/DesignSystemGlobalNavigationModelV2.class.php
```diff
@@ -97,9 +55,34 @@ public function getData() {
 		return $data;
 	}
 
+	private function getMainNavigation() {
+		if ( $this->isWikiaOrgCommunity() ) {
+			return $this->getLink( self::COMMUNITY_CENTRAL_LABEL, $this->getHref('community-central'), self::COMMUNITY_CENTRAL_TRACKING_LABEL );
+		} else {
+			return array_merge(
+				$this->getFandomLinks(),
+				[
```
latata: how about extracting this array to method?

latata (approved):
❤️

includes/wikia/models/DesignSystemGlobalNavigationModelV2.class.php
rybmat: makes sense

PR #15473, opened 2018-06-05
SUS-5124 | sane cache invalidation for Wall/Forum 

mszabo-wikia:
In order to ensure changes to Wall/Forum content are immediately visible to users, we must purge the URLs for the Wall/Forum thread and its containing Wall/Board page as well as cached reply information (if it was a thread reply). Wall accomplished the former by hooking into `TitleGetSquidURLs`, however it created a new `Title` object directly from database, which caused issues with replication lag & spawned a number of workarounds including waiting for slaves to catch up. Wall refactorings, most notably 34b668b971093563d625e351226cdd1a4f4012c9 have since removed unnecessary `Title` creation and thus the root cause of replication lag problems, but the logic here is confusing enough to warrant some proper cleanup for future (s)us.

* When the underlying article comment of a Wall/Forum message was changed, AC would schedule purges of its own, which Wall/Forum had to be aware of and intercept. This logic was moved out of shared methods so now it only runs for article comments.
* Similarly `ArticleComment::doAfterPost` was performing work only required by article comments, so it was also removed from shared methods.
* `Wall::invalidateCache` was a no-op method that nevertheless got called in several places—it was removed.
* `WallThread::invalidateCache` call was moved into `WallMessage::invalidateCache` because these methods go side-by-side so it makes sense to call them together.

https://wikia-inc.atlassian.net/browse/SUS-5124


extensions/wikia/ArticleComments/classes/ArticleComment.class.php
```diff
@@ -1094,40 +1095,17 @@ static public function doPurge( $title, $commentTitle ) {
...
 	 * @param Status $status
 	 * @param Article|WikiPage $article
-	 * @param int $parentId
+	 * @param Title $parentPageTitle
+	 * @param int $parentCommentId
 	 * @return array
+	 * @throws Exception
 	 */
-	static public function doAfterPost( Status $status, $article, $parentId = 0 ) {
-		Hooks::run( 'ArticleCommentAfterPost', [ $status, &$article ] );
```
mszabo-wikia, line 1130: unused

extensions/wikia/ArticleComments/classes/ArticleComment.class.php
```diff
@@ -1201,10 +1182,6 @@ static public function doAfterPost( Status $status, $article, $parentId = 0 ) {
 	static public function addArticlePageToWatchlist( $comment ) {
 		global $wgUser, $wgEnableArticleWatchlist, $wgBlogsEnableStaffAutoFollow;
 
-		if ( !Hooks::run( 'ArticleCommentBeforeWatchlistAdd', [ $comment ] ) ) {
```
mszabo-wikia, line 1204: unused after removing sole handler

extensions/wikia/ArticleComments/classes/ArticleCommentsAjax.class.php
```diff
@@ -207,35 +207,22 @@ static public function axPost() {
...
 
-		if (
-			$title->getNamespace() == NS_USER_TALK &&
-			$response[0] == EditPage::AS_SUCCESS_NEW_ARTICLE &&
-			$title->getText() != $wgUser->getName()
-		) {
-			$user = User::newFromName( $title->getText() );
-
-			if ( $user ) {
-				$user->setNewtalk( true );
```
mszabo-wikia, line 218: this feature is already implemented elsewhere, it's called: Message Wall & Wall Notifications

extensions/wikia/Wall/Wall.class.php
```diff
@@ -358,9 +358,4 @@ public function setMaxPerPage( $val ) {
 	public function setSorting( $val ) {
 		$this->mSorting = $val;
 	}
-
-	public function invalidateCache() {
-		// TODO: implent it
-		return true;
-	}
```
mszabo-wikia, line 365: nope

extensions/wikia/Wall/WallHooksHelper.class.php
```diff
@@ -1299,22 +1258,6 @@ static public function onAllowNotifyOnPageChange( $editor, $title ) {
 		return true;
 	}
 
-	/**
-	 * @param User $user
-	 * @param Article $article
-	 * @return bool
-	 */
-	static public function onWatchArticle( &$user, &$article ) {
```
mszabo-wikia, line 1307: hook handler registration was commented out

extensions/wikia/Wall/WallHooksHelper.class.php
```diff
@@ -1348,14 +1291,6 @@ static public function onUnwatchArticle( &$user, &$article ) {
 		return true;
 	}
 
-	static private function isWallMainPage( Title $title ) {
```
mszabo-wikia, line 1351: unused private method

includes/wikia/services/UserStatsService.class.php
```diff
@@ -123,7 +123,7 @@ function () {
 				} else {
 					// make sure this value is an integer
 					$stats['editcount'] = (int) $stats['editcount'];
-					$stats['editcountThisWeek'] = (int) $stats['editcountThisWeek'];
+					$stats['editcountThisWeek'] = (int) ( $stats['editcountThisWeek'] ?? 0);
```
macbre, line 126: :+1:

macbre (approved):

PR #15470, opened 2018-06-05
SUS-4383 | create a new wiki via an offline Celery task

macbre:
https://wikia-inc.atlassian.net/browse/SUS-4383

## Why

Currently CreateNewWiki internal magic is executed when an AJAX request is processed. This takes ~15-30 seconds on production in SJC. As we"re moving towards an active-active setup we'd like to run this from RES datacenter as well. As @mixoteric tests showed it's not currently possible due to the network latency between SJC-RES. CreateNewWiki AJAX request simply times out in Reston.

## What changed

1. Send an AJAX request with all the required parameters and queue an offline Celery task to actually create a new wiki.
2. It finishes quickly and returns queue task ID.
3. Every two seconds send a polling AJAX request that will check if given task completes.
4. Monitor the time required for an offline task to complete + add monitoring and alerting for this metric.

## Database

`wikicities.city_creation_log` table has been introduced to keep track of wiki creation progress and status.

```sql
CREATE TABLE `city_creation_log` (
  `log_id` int(9) NOT NULL AUTO_INCREMENT,
  `task_id` char(39) DEFAULT NULL,
  `city_id` int(11) DEFAULT NULL,
  `creation_started` datetime DEFAULT CURRENT_TIMESTAMP,
  `creation_ended` datetime DEFAULT NULL,
  `completed` tinyint(1) NOT NULL DEFAULT '0',
  `exception_message` varchar(256) DEFAULT NULL,
  PRIMARY KEY (`log_id`),
  KEY `task_id_idx` (`task_id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8
```

Example results:

```sql
-- Wiki creation in progress

ysql@geo-db-dev-db-slave.query.consul[wikicities]>select * from city_creation_log;
+--------+-----------------------------------------+---------+---------------------+----------------+-----------+-------------------+
| log_id | task_id                                 | city_id | creation_started    | creation_ended | completed | exception_message |
+--------+-----------------------------------------+---------+---------------------+----------------+-----------+-------------------+
|      1 | mw-99AEE755-9F5D-4DFA-8F30-C0F1CA76C7F0 | 1657000 | 2018-06-04 09:47:33 | NULL           |         0 | NULL              |
+--------+-----------------------------------------+---------+---------------------+----------------+-----------+-------------------+
1 row in set (0.00 sec)

-- Completed

mysql@geo-db-dev-db-slave.query.consul[wikicities]>select * from city_creation_log;
[...]

PR #15484, opened 2018-06-05
Cake 2908 - link fc community on wiki creation and prevent community creation emails

nmonterroso:
@Wikia/cake 

Does FC community setup as part of wiki creation instead of as a separate request. Also prevents initial wall posting message and founder tip emails for FC-enabled communities during wiki creation.

🤞 note that since wiki creation does not work in the dev environment it's not possible to test these changes on my devbox. But from what I've been able to test things work the same way if there's no fc community passed.🤞 

See Wikia/pandora#4572 for the pandora side changes.


nmonterroso:
#15470 will change the community creation flow to happen as part of an async task, so we should reimplement with those changes.

PR #15488, opened 2018-06-06
SUS-5128 | use GoogleMaps API key in Places extension

macbre:
https://wikia-inc.atlassian.net/browse/SUS-5128

Avoid the following:

```
Google Maps JavaScript API warning: NoApiKeys https://developers.google.com/maps/documentation/javascript/error-messages#no-api-keys
```

Additionally, track when Google Maps API is loaded via `$.loadGoogleMaps` (uses by Places extension):

```
Wikia.Tracker: trackingevent googlemaps/open/apiloaded/ [analytics track]
```

### DIFF
--- a/extensions/wikia/Places/PlacesController.class.php
+++ b/extensions/wikia/Places/PlacesController.class.php
@@ -60,11 +60,15 @@ class PlacesController extends WikiaController {
 	 * Map center can be specified
 	 */
 	public function renderMarkers() {
+		// SUS-5128 | add Google Maps API key
+		global $egMapsGMaps3ApiKey;
+
 		$this->setVal('markers', $this->prepareMarkers($this->getVal('markers')));
 		$this->setVal('center', $this->getVal('center'));
 		$this->setVal('mapId', 'places-map-' . self::$mapId++);
 		$this->setVal('height', $this->getVal('height', 500));
 		$this->setVal('options', $this->getVal('options', array()));
+		$this->setVal('googleMapsApiKey', $egMapsGMaps3ApiKey);
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


extensions/wikia/Places/PlacesController.class.php
```diff
@@ -60,11 +60,15 @@ public function placeFromModel(){
 	 * Map center can be specified
 	 */
 	public function renderMarkers() {
+		// SUS-5128 | add Google Maps API key
+		global $egMapsGMaps3ApiKey;
```
mszabo-wikia: Let's use wgGoogleMapsKey instead

mszabo-wikia (changes requested):

extensions/wikia/Places/PlacesController.class.php
macbre: `egMapsGMaps3ApiKey` has different values for different environments while `wgGoogleMapsKey` is fixed.

mszabo-wikia: let's address that then, egMapsGMaps3ApiKey is Maps extension specific var :)

macbre: Tried wgGoogleMapsKey on devbox. It does not work - `egMapsGMaps3ApiKey` handles this environment.

mszabo-wikia: but then we should add the proper env overrides for wgGoogleMapsKey and have egMapsGMaps3ApiKey just be a simple reassignment—if we ever change or sunset Maps extension, tight coupling to its internals will cause confusion

macbre: See 4ae2620

mszabo-wikia (approved):

PR #15491, opened 2018-06-06
SUS-4895 | Don't let tracking calls hang forever if opt-in modal cannot load

mszabo-wikia:
https://wikia-inc.atlassian.net/browse/SUS-4895

### DIFF
--- a/resources/wikia/modules/tracker.js
+++ b/resources/wikia/modules/tracker.js
@@ -119,7 +119,7 @@
 		function flushInternalTrackingQueue(optIn) {
 			isOptedIn = optIn;
 			isTrackingOptInReady = true;
-			
+
 			while (internalTrackingQueue.length > 0) {
 				var fn = internalTrackingQueue.shift();
 				fn();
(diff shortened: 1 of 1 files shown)


resources/wikia/modules/tracker.js
```diff
@@ -370,8 +370,14 @@
 			track.apply( null, args );
 		}
 
-		require(['wikia.trackingOptIn'], function (trackingOptIn) {
-			trackingOptIn.pushToUserConsentQueue(flushInternalTrackingQueue);
+		require([require.optional('wikia.trackingOptIn')], function (trackingOptIn) {
```
pwojcik86, line 373: Is this the only place we need to do this? What about `universal_analytics`?

pwojcik86 (approved):

resources/wikia/modules/tracker.js
mszabo-wikia: generally trackingOptIn is always available, it's only not there if the asset was somehow blocked (e.g. crazy antivirus). it's fine if GA fails in that case since that cannot block the page, but tracker calls waiting for callbacks can unfortunately cause such blockage that we need to alleviate

PR #15492, opened 2018-06-06
cake 2908 - save FC community id on wiki creation, don't send founder emails

nmonterroso:
@Wikia/cake 

Saves the FC community link during wiki creation if the param is present, rather than requiring a separate request. This is also used as a flag to prevent founder tip emails from being sent out. 

Since this is built on #15470 this PR shouldn't be merged until that PR is merged.


nmonterroso:
This is ready to merge

drsnyder:
LGTM

macbre:
This is causing `Argument 6 passed to CreateWikiTask::create() must be of the type string, null given`

PR #15495, opened 2018-06-07
XW-4988 | Add tracking to featured video attribution link

rybmat:
https://wikia-inc.atlassian.net/browse/XW-4988

@Wikia/x-wing

### DIFF
--- a/extensions/wikia/ArticleVideo/scripts/featured-video.attribution.js
+++ b/extensions/wikia/ArticleVideo/scripts/featured-video.attribution.js
@@ -1,12 +1,25 @@
 define('wikia.articleVideo.featuredVideo.attribution', [
 		'wikia.mustache',
 		'wikia.articleVideo.featuredVideo.templates',
-		'JSMessages'
+		'JSMessages',
+		'wikia.tracker'
 	],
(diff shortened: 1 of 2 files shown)


extensions/wikia/ArticleVideo/scripts/featured-video.attribution.js
```diff
@@ -1,12 +1,26 @@
...
+				category: 'featured-video',
+				action: tracker.ACTIONS.CLICK,
+				trackingMethod: 'analytics'
+			});
+
 		return function (playerInstance) {
+			// tracking for initial video attribution
+			$('.featured-video__attribution-username, .featured-video__attribution-icon').click(function() {
+				track({
+					label: playerInstance.getPlaylistItem(0).username
```
bkoval, line 20: According to https://developer.jwplayer.com/jw-player/docs/javascript-api-reference/ getPlaylistItem with no number returns current playlist item. Giving index 0 will probably return the first one.

bkoval: Ah, but in this case it's ok, since this will always be the first item :)

bkoval (approved):

PR #15507, opened 2018-06-08
ADEN-7229 Fix indexExchange bidder parameters

dmnsobczak:


extensions/wikia/AdEngine/js/lookup/prebid/adapters/indexExchange.js
```diff
@@ -6,6 +6,7 @@ define('ext.wikia.adEngine.lookup.prebid.adapters.indexExchange',[
 	'use strict';
 
 	var bidderName = 'indexExchange',
```
fraszczakszymon, line 8: Is it still needed?

extensions/wikia/AdEngine/js/lookup/prebid/adapters/indexExchange.js
```diff
@@ -125,9 +115,8 @@ define('ext.wikia.adEngine.lookup.prebid.adapters.indexExchange',[
 			},
 			bids: config.sizes.map(function (size) {
 				return {
-					bidder: bidderName,
+					bidder: requestsBidderName,
```
fraszczakszymon, line 118: Does it mean we need to retarget all line items?

extensions/wikia/AdEngine/js/lookup/prebid/adapters/indexExchange.js
dmnsobczak: Yes, I would stick to this full name. Also all lines in DFP are targeted to `indexExchange`.

extensions/wikia/AdEngine/js/lookup/prebid/adapters/indexExchange.js
dmnsobczak: No, Index Exchange only requires `ix` in bid params because... just because. But we can rename all the lines, modules and test variables if we have to much free time :>

PR #15519, opened 2018-06-12
Revert "cake 2908 - save FC community id on wiki creation, don't send founder emails"

macbre:
Reverts Wikia/app#15492

Causes `Argument 6 passed to CreateWikiTask::create() must be of the type string, null given`.


macbre:
Backported to the preview branch in 3ddda75

PR #15524, opened 2018-06-12
Cake 2908 - retry but with default null for FC community id

nmonterroso:
@Wikia/cake @macbre @mszabo-wikia 

Retrying #15492, but with a default for the FC community id set to null. I tested this by deploying to sandbox-s1 before doing `= null`, manually trying CNW -> verify broken + run selenium -> verify busted, then deploying with default `null` and rerunning both tests and making sure they work.

My [success selenium](http://jenkins.wikia-prod:8080/view/Sustaining/view/Sustaining-sandbox/job/sustaining-app-CNW-user-sandbox/263/console) run still has this error:
```
 org.openqa.selenium.TimeoutException: Expected condition failed: waiting for element to be clickable: [[ChromeDriver: chrome on XP (2641910e5df08211621f821cf3f72961)] -> css selector: #ca-delete] (tried for 5 second(s) with 500 milliseconds interval)
```

But it seems like all tests going back to June 6 have this error, so it shouldn't be related to my change.


JoshuaRogan (approved):
LGTM

PR #15546, opened 2018-06-15
SUS-5253 MW in K8s - in nginx - add rewrite from main path to index.php

sqreek:
https://wikia-inc.atlassian.net/browse/SUS-5253

@Wikia/sus


mszabo-wikia (approved):
Note in current setup main rewrite is on CDN level for better hit ratio

PR #15566, opened 2018-06-19
PLATFORM-3537: upgrade Vignette images embedded in wikitext to use HTTPS

harnash:
When dealing with Vignette (-like) url embedded in wikitext we should upgrade them to HTTPS to avoid mixed content issues.


extensions/wikia/HTTPSSupport/HTTPSSupportHooks.class.php
```diff
@@ -68,4 +70,14 @@ private static function httpsEnabledTitle( Title $title ): bool {
 		return array_key_exists( $wgDBname, self::$httpsArticles ) &&
 			in_array( $title->getPrefixedDBKey(), self::$httpsArticles[ $wgDBname ] );
 	}
+
+	public static function parserUpgradeVignetteUrls ( string &$url ): bool {
+		if ( preg_match( self::VIGNETTE_IMAGES_HTTP_UPGRADABLE, $url ) && strpos( $url, "http://" ) === 0 ) {
+			$url = wfHttpToHttps( $url );
+			return true;
```
mszabo-wikia: Other way around—`false` aborts further hook processing, `true` doesn't 😉

mszabo-wikia (changes requested):

extensions/wikia/HTTPSSupport/HTTPSSupportHooks.class.php
harnash: oh... right! Fix: 11bef317cc

mszabo-wikia (approved):

extensions/wikia/HTTPSSupport/HTTPSSupportHooks.class.php
```diff
@@ -68,4 +70,10 @@ private static function httpsEnabledTitle( Title $title ): bool {
 		return array_key_exists( $wgDBname, self::$httpsArticles ) &&
 			in_array( $title->getPrefixedDBKey(), self::$httpsArticles[ $wgDBname ] );
 	}
+
+	public static function parserUpgradeVignetteUrls ( string &$url ): bool {
```
Grunny: This is no longer returning a bool, so the type declaration on the return value should be removed.

Grunny (changes requested):

extensions/wikia/HTTPSSupport/HTTPSSupportHooks.class.php
harnash: fix: 8e855226164

Grunny (approved):

PR #15573, opened 2018-06-20
Fixing tests to not call deleted wiki

harnash:
Another wiki bites the dust.

### DIFF
--- a/includes/wikia/tests/GlobalTitleTest.php
+++ b/includes/wikia/tests/GlobalTitleTest.php
@@ -14,18 +14,18 @@ class GlobalTitleTest extends WikiaBaseTest {
 			->willReturnMap( [
 				// basically all tests where GlobalTitle::load() is executed
 				[ 'wgServer', 177, 'http://community.wikia.com' ],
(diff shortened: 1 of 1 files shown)


includes/wikia/tests/GlobalTitleTest.php
```diff
@@ -58,16 +58,16 @@ function testNewFromText2() {
 	function testUrlsMainNS() {
 		$this->mockProdEnv();
 
-		$title = GlobalTitle::newFromText( "Timeline", NS_MAIN, 410 ); # yugioh
-		$expectedUrl = "http://yugioh.wikia.com/wiki/Timeline";
+		$title = GlobalTitle::newFromText( "Timeline", NS_MAIN, 147 ); # yugioh
```
KockaAdmiralac: You might want to change the `# yugioh` comment here and `# wowwiki` on line 69.

harnash: Got that! Thanks

macbre (approved):

PR #15577, opened 2018-06-20
Remove not used /cache directory

macbre:
It's empty on production as we do not use `FileCacheBase` class.


mszabo-wikia (approved):

sqreek (approved):
almost 10 years

PR #15580, opened 2018-06-21
SUS-5287 | MWHttpRequest::proxySetup - do not use a localhost as a proxy for "local" URLs + use $wgHTTPProxy value only

macbre:
https://wikia-inc.atlassian.net/browse/SUS-5287

When MediaWiki is run inside a container it does not have an access to a localhost proxy as nginx is running in a different container. This commit fixes the following error when trying to perform a HTTP request to a "local" URL, i.e. in *.wikia.com domain. Instead of `localhost:80` proxy a HTTP border will be used.

Additionally, backport a MediaWiki change that simply ignores `http_proxy` environment variable. Let's stick to `$wgHTTPProxy`.

```
http-curl-error, Failed to connect to localhost port 80: Connection refused
```

### DIFF
--- a/includes/HttpFunctions.php
+++ b/includes/HttpFunctions.php
@@ -380,12 +380,11 @@ class MWHttpRequest {
 	 * (override in subclass)
 	 */
 	public function proxySetup() {
-		global $wgHTTPProxy;
-
 		if ( $this->proxy ) {
 			return;
 		}
 
+		global $wgHTTPProxy;
 		if ( $wgHTTPProxy ) {
 			$this->proxy = $wgHTTPProxy ;
 		}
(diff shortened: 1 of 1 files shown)


includes/HttpFunctions.php
```diff
@@ -388,12 +386,8 @@ public function proxySetup() {
 			return;
 		}
 
-		if ( Http::isLocalURL( $this->url ) ) {
-			$this->proxy = 'http://localhost:80';
-		} elseif ( $wgHTTPProxy ) {
+		if ( $wgHTTPProxy ) {
```
sqreek: micro-micro optimization
global $wgHTTPProxy 
could go just before use statement

sqreek (approved):

PR #15588, opened 2018-06-22
PLATFORM-3390: script for uploading videos to a given wiki

harnash:
Simple script to upload videos to a wiki. It takes list of URLs to be uploaded and uploads them using MW mechanisms.


extensions/wikia/VideoHandlers/maintenance/uploadVideo.php
```diff
@@ -0,0 +1,69 @@
...
+
+	public function execute() {
+		global $wgUser, $wgDBname;
+
+		$fileList = realpath( $this->getOption( 'file-list' ) );
+		if ( !file_exists( $fileList ) ) {
+			$this->error( "File with list of files to upload does not exists: $fileList", 1 );
+		}
+
+		$files = file( $fileList );
```
Grunny: Minor: I'd consider passing `FILE_IGNORE_NEW_LINES | FILE_SKIP_EMPTY_LINES` as the second param here. You're trimming so `FILE_IGNORE_NEW_LINES` isn't strictly needed, but `FILE_SKIP_EMPTY_LINES` will avoid us trying to upload an empty string if we have a new line at the end of the file.

Grunny (approved):
One suggestion, otherwise looks good.

extensions/wikia/VideoHandlers/maintenance/uploadVideo.php
harnash: Fix in: 280511188c01cf850f707ff92a444e7930c9dd7d

PR #15599, opened 2018-06-25
(PLATFORM-3524) More handling for downgrading anonymous users

Grunny:
Ensure app handles downgrading anonymous users to HTTP for robots.txt
and sitemaps. Also limit the cache length on anonymous downgrades of
articles for now.

/cc @Wikia/core-platform-team

### DIFF
--- a/extensions/wikia/HTTPSSupport/HTTPSSupportHooks.class.php
+++ b/extensions/wikia/HTTPSSupport/HTTPSSupportHooks.class.php
@@ -32,14 +32,14 @@ class HTTPSSupportHooks {
 	 * to HTTP if necessary.
 	 *
 	 * @param  Title      $title
-	 * @param             $unused
+	 * @param             $article
 	 * @param  OutputPage $output
 	 * @param  User       $user
 	 * @param  WebRequest $request
 	 * @param  MediaWiki  $mediawiki
 	 * @return bool
 	 */
-	public static function onBeforeInitialize( Title $title, $unused, OutputPage $output,
+	public static function onAfterInitialize( Title $title, $article, OutputPage $output,
 		User $user, WebRequest $request, MediaWiki $mediawiki
 	): bool {
 		global $wgDisableHTTPSDowngrade;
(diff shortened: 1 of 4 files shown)


extensions/wikia/HTTPSSupport/HTTPSSupportHooks.class.php
```diff
@@ -56,11 +56,48 @@ public static function onBeforeInitialize( Title $title, $unused, OutputPage $ou
...
+	 * @param  WebRequest $request
+	 * @param  User       $user
+	 * @return boolean
+	 */
+	public static function onSitemapPageBeforeOutput( string $subpage, WebRequest $request, User $user ): bool {
+		global $wgScriptPath;
+		if ( WebRequest::detectProtocol() === 'https' &&
+			!self::httpsAllowed( $user, $request->getFullRequestURL() )
+		) {
+			self::downgradeRedirectForPath( "$wgScriptPath/$subpage", $request );
```
themech, line 78: So we're undoing the rewrite introduced in VCL. Assuming the rewrite actually happened. And skipping all the query params (like a `cb=`). It should be fine in this case, but makes me wonder if we should have any rewrites for app on VCL level once we're doing protocol redirects in the app.

extensions/wikia/HTTPSSupport/HTTPSSupportHooks.class.php
```diff
@@ -56,11 +56,48 @@ public static function onBeforeInitialize( Title $title, $unused, OutputPage $ou
...
+	}
+
+	/**
+	 * Handle downgrading anonymous requests for robots.txt.
+	 *
+	 * @param  WebRequest $request
+	 * @param  User       $user
+	 * @return boolean
+	 */
+	public static function onWikiaRobotsBeforeOutput( WebRequest $request, User $user ): bool {
```
themech, line 91: This looks very similar to `onSitemapPageBeforeOutput`. Maybe we could create a new method that takes the `$user`, `$request` and `redirectUrl` params and call it from those two hook handlers.
And ideally if we manage to solve the issues of undoing VCL redirects, the `redirectUrl` won't be needed anymore, as we will be able to use `getFullRequestURL`

extensions/wikia/HTTPSSupport/HTTPSSupportHooks.class.php
themech: as discussed, we will do some cleanup in a follow-up ticket

themech (approved):

PR #15605, opened 2018-06-26
ADEN-7294 New implementation of InstartLogic recovery

dmnsobczak:


extensions/wikia/AdEngine/js/lookup/prebid/adapters/appnexus.js
```diff
@@ -65,8 +65,8 @@ define('ext.wikia.adEngine.lookup.prebid.adapters.appnexus',[
 		return adContext.get('bidders.appnexus');
```
fraszczakszymon: Does it mean that AN is going to be enabled both for BT and IL rec? Shouldn't we enable it based on `context.opts.wadIL`?

extensions/wikia/AdEngine/js/wad/ilRecLoader.js
```diff
@@ -0,0 +1,33 @@
...
+			isAsync: false,
+			node: doc.head.lastChild
+		});
+	}
+
+	function init() {
+		if (adContext.get('opts.babRec')) {
+			doc.addEventListener('bab.blocking', injectScript);
+		} else {
+			injectScript();
```
fraszczakszymon: Same question here - do we really want to inject this script without babDetection?

dmnsobczak: And same answer: It's just an option. We had that before, why not leave it after cleanup-o-refactor. Waiting for grooming - we probably gonna remove that.

dmnsobczak: Ok, removing.

extensions/wikia/AdEngine/js/lookup/prebid/adapters/appnexus.js
dmnsobczak: That's true, condition expanded.

fraszczakszymon (approved):

fraszczakszymon (approved):

fraszczakszymon (approved):

PR #15614, opened 2018-06-28
SUS-5405 | Use k8s internal ingress for calling services on Kubernetes

mszabo-wikia:
Avoid unnecessary roundtrip outside of K8S cluster when calling services that are themselves running on K8S.

https://wikia-inc.atlassian.net/browse/SUS-5405

### DIFF
--- a/includes/wikia/VariablesBase.php
+++ b/includes/wikia/VariablesBase.php
@@ -8235,6 +8235,13 @@ $wgUseImageResize = true;
 $wgUseInstantCommons = false;
 
 /**
+ * Whether to use Kubernetes internal ingress for making requests to service dependencies on Kubernetes.
+ * This is only enabled if app itself is running on Kubernetes.
+ * @var bool $wgUseKubernetesInternalIngress
+ */
+$wgUseKubernetesInternalIngress = getenv( 'KUBERNETES_POD' );
+
+/**
  * Set this to true to make a local copy of the message cache, for use in
  * addition to memcached. The files will be put in $wgCacheDirectory.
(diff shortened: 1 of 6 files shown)


includes/wikia/VariablesBase.php
```diff
@@ -8234,6 +8234,13 @@
  */
 $wgUseInstantCommons = false;
 
+/**
+ * Whether to use Kubernetes internal ingress for making requests to service dependencies on Kubernetes.
+ * This is only enabled if app itself is running on Kubernetes.
+ * @var bool $wgUseKubernetesInternalIngress
+ */
+$wgUseKubernetesInternalIngress = getenv( 'KUBERNETES_POD' );
```
macbre: Let's cast it to a "real" boolean value.

macbre (approved):

lib/Wikia/src/Service/Gateway/InternalIngressUrlProvider.php
```diff
@@ -0,0 +1,9 @@
+<?php
+namespace Wikia\Service\Gateway;
+
+class InternalIngressUrlProvider implements UrlProvider {
+
+	public function getUrl( $serviceName ) {
+		return $serviceName;
```
sqreek: is namespace not needed here?

mszabo-wikia: 👍

sqreek (approved):

macbre (approved):

PR #15616, opened 2018-06-29
SER-2394 | send internal request header with write operations

idradm:
https://wikia-inc.atlassian.net/browse/SER-2323


mszabo-wikia (changes requested):
This is generated code, so it'd be overwritten the next time someone regenerates the API client.

Plus this change is not needed because we automatically set the internal header for all outgoing Swagger client requests.

PR #15619, opened 2018-07-02
ADEN-7334 Adds getPublisherConsent command

leshniak:


extensions/wikia/TrackingOptIn/js/cmp.js
```diff
@@ -227,6 +266,25 @@ define('wikia.cmp', [
 				}, {})
 			}, true);
 		},
+		getPublisherConsent: function (purposesIds, callback) {
+			var purposesAllowed = (purposesIds || publisherConsent.getPurposesAllowed()).reduce(function (obj, id) {
+					var type = (id > 24) ? 'customPurposeConsents' : 'standardPurposeConsents';
```
nandy-andy: What's the magic `24`? Can we assign it to a variable with a descriptive name? ;-)

extensions/wikia/TrackingOptIn/js/cmp.js
```diff
@@ -47,9 +60,31 @@ define('wikia.cmp', [
 		log('Initializing module', log.levels.debug, logGroup);
 
 		vendorConsentString = cookies.get(vendorConsentCookieName);
+		publisherConsentString = cookies.get(publisherConsentCookieName);
+
+		if (publisherConsentString) {
```
nandy-andy: This `if-else` logic is very similar as for publisher consent string and vendor consent string - can we extract it and re-use it in the `init`?

extensions/wikia/TrackingOptIn/js/cmp.js
ludwikkazmierczak: @nandy-andy id's up to 24 are reserved. 25+ may be our custom purposes :)

extensions/wikia/TrackingOptIn/js/cmp.js
leshniak: yes, it's similar. we could, but the next step is to remove all this code. I think that it's not worth of time.

dmnsobczak (approved):

fraszczakszymon (approved):

PR #15624, opened 2018-07-03
PLATFORM-3548: remove unnecessary path manipulations when doing HTTPS…

harnash:
… downgrades

This affects Sitemaps and robots.txt
related to 
* https://github.com/Wikia/wikia-vcl/pull/1832
* https://github.com/Wikia/wikia-vcl/pull/1833


Grunny (approved):

extensions/wikia/HTTPSSupport/HTTPSSupportHooks.class.php
```diff
@@ -63,36 +63,17 @@ public static function onAfterInitialize( Title $title, $article, OutputPage $ou
...
-	 * @param  WebRequest $request
-	 * @param  User       $user
-	 * @return boolean
-	 */
-	public static function onWikiaRobotsBeforeOutput( WebRequest $request, User $user ): bool {
-		if ( WebRequest::detectProtocol() === 'https' &&
-			!self::httpsAllowed( $user, $request->getFullRequestURL() )
-		) {
-			self::downgradeRedirectForPath( '/robots.txt', $request );
+			self::downgradeRedirectForPath( $request->getFullRequestURL(), $request );
```
themech: For the downgrade, do we need to pass both, the path and the request? Maybe we can just pass the `$request` and rename the method to something like `downgradeRedirect` or `downgradeRequest`.

harnash: No. Currently we don't. 0b8e14a5902

extensions/wikia/HTTPSSupport/HTTPSSupportHooks.class.php
```diff
@@ -66,27 +66,20 @@ public static function onAfterInitialize( Title $title, $article, OutputPage $ou
...
-	 * Handle downgrading anonymous requests for robots.txt.
-	 *
-	 * @param  WebRequest $request
-	 * @param  User       $user
-	 * @return boolean
-	 */
-	public static function onWikiaRobotsBeforeOutput( WebRequest $request, User $user ): bool {
-		return self::handleRedirectForPath( '/robots.txt', $request, $user );
+	public static function onSitemapRobotsPageBeforeOutput( WebRequest $request, User $user ): bool {
+		if ( WebRequest::detectProtocol() === 'https' &&
```
Grunny: This is no longer just downgrading after https://wikia-inc.atlassian.net/browse/PLATFORM-3552 which is why the method below was renamed to just `handleRedirectForPath`, so having this check here undoes the work in PLATFORM-3552 for robots and sitemaps. I think this hook should now just be the content of what was the `handleRedirectForPath` method below.

Grunny (changes requested):

extensions/wikia/HTTPSSupport/HTTPSSupportHooks.class.php
harnash: Oops. Fixed that in 017ce92037c

themech (approved):

Grunny (approved):

PR #15626, opened 2018-07-03
SUS-5462 | Use WikiaLogger to catch exceptions for maintanance scripts

JamesHawking:
https://wikia-inc.atlassian.net/browse/SUS-5462


macbre (changes requested):
`wfShellExec` throws an exception only when the command is longer than 100k characters.

chessky:
The thing is already done.

PR #15627, opened 2018-07-03
SUS-5441 | Redis based localisation cache

mszabo-wikia:
Localisation cache prototype that uses Redis to store data instead of CDB files on disk. It supports also varying the cache between different clients that use the same instance (e.g. devboxes, sandboxes)

https://wikia-inc.atlassian.net/browse/SUS-5441


lib/Wikia/src/Localisation/LCStoreRedis.php
```diff
@@ -0,0 +1,114 @@
...
+	 * @param $key
+	 * @return array
+	 */
+	public function getItemList( $code, $key ) {
+		return $this->getOrCreateConnection()->hKeys( "$code:$key" );
+	}
+
+	private function getOrCreateConnection(): \Redis {
+		if ( $this->redisConnection === null ) {
+			$this->redisConnection = new \Redis();
```
macbre, line 99: I believe we need to install the dependency that provides `Redis` class.

mszabo-wikia: yup, we'll need https://github.com/phpredis/phpredis

PR #15646, opened 2018-07-06
SUS-5396 | Migrate cron tab from cron-s1 to k8s cron jobs

JamesHawking:
Example descriptors of cronjobs for maintenance scripts.

https://wikia-inc.atlassian.net/browse/SUS-5396


mszabo-wikia (approved):
Looks OK to me. Probably we will have to integrate deployment of these cron jobs into the main prod deploy pipeline once it's ready.

sqreek:
Also adding label sus will be nice

docker/maintenance/updateSpecialPages/updateSpecialPages-c1.yaml
```diff
@@ -0,0 +1,30 @@
...
+             - name: LOG_SOCKET_ADDRESS
+               value: "tcp://localhost:9999"
+            args:
+            - php
+            - /usr/wikia/slot1/current/src/maintenance/maintenanceTaskScheduler.php
+            - --params="--quiet"
+            - --cluster=c1
+            - --active=2
+            - maintenance/updateSpecialPages.php
+          restartPolicy: Never
```
macbre: Just discussed this with @drozdo and we can just run `maintenance/updateSpecialPages.php` across all wikis at once - no need to create seven separate YAML files. Celery queue will make sure we do not run too many scripts at once.

docker/maintenance/SMW_conceptCache/SMW_conceptCache-1318.yaml
```diff
@@ -0,0 +1,29 @@
...
+  schedule: "0 10 * * *"
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+          - name: logger
+            image: artifactory.wikia-inc.com/sus/mediawiki-logger:latest
+          - name: php
+            image: artifactory.wikia-inc.com/sus/mediawiki-sandbox-php:54f12b1.9306b50
```
macbre: We need to keep this tag in sync with production deploy. Should we make this a template that Jenkinsfile will use and pass a correct tag?

docker/maintenance/SMW_conceptCache/SMW_conceptCache-490.yaml
```diff
@@ -0,0 +1,29 @@
...
+             - name: SERVER_ID
+               value: "177"
+             - name: LOG_SOCKET_ONLY
+               value: "yes"
+             - name: LOG_SOCKET_ADDRESS
+               value: "tcp://localhost:9999"
+            args:
+            - php
+            - /usr/wikia/slot1/current/src/maintenance/maintenanceTaskScheduler.php
+            - --id=490
```
macbre: We can pass multiple IDs to the `maintenanceTaskScheduler` script. Let's just use a single YAML file for `SMW_conceptCache.php`.

PR #15648, opened 2018-07-09
SUS-5499 | Use internal host name for MW->MW requests on Kubernetes

mszabo-wikia:
Use internal host name for MW->MW requests on Kubernetes. Otherwise, the request goes via borders and ends up on Apache boxes.

https://wikia-inc.atlassian.net/browse/SUS-5499


includes/HttpFunctions.php
```diff
@@ -296,6 +296,16 @@ function __construct( $url, $options = array() ) {
...
+		// SUS-5499: Use internal host name for MW->MW requests when running on Kubernetes
+		if ( $wgKubernetesNamespace && Http::isLocalURL( $url ) ) {
+			$targetHost = $this->parsedUrl['host'];
+
+			$this->url = str_replace( $targetHost, "$wgKubernetesDeploymentName.$wgKubernetesNamespace", $this->url );
+			$this->parsedUrl['host'] = "$wgKubernetesDeploymentName.$wgKubernetesNamespace";
+
+			$this->setHeader( 'Host', $targetHost );
+		}
+
```
macbre: Maybe let's use proxy instead?

mszabo-wikia: hmm, but what would we be proxying through? borders will cause the request to go to apaches.

rogatty (approved):

includes/wikia/VariablesBase.php
```diff
@@ -5152,6 +5152,18 @@
  */
 $wgJsMimeType = 'text/javascript';
 
+/**
```
ghost: I would put these assignments [somewhere here](https://github.com/Wikia/app/blob/dev/LocalSettings.php#L43), closer to other `getenv()` and leave only simple declarations in `VariablesBase.php` (no function calls, no logic in this file). Ideally, I would like to keep this file injecting identical state in any environment. `getenv()` here makes it implicitly env-specific. This is not a hard requirement, though - more of a personal preference and tendency to keep things strict and group similar stuff together.

ghost (changes requested):

rogatty (approved):

ghost (approved):

ghost:
Thanks @macbre

macbre:
```
$ curl -x mediawiki-sandbox-sus2.prod:80  -H 'X-Original-Host: muppet.wikia.com' http://muppet.wikia.com/wiki/Elmo -svo /dev/null 2>&1 | grep -E 'Served|HTTP|Server'
> GET http://muppet.wikia.com/wiki/Elmo HTTP/1.1
< HTTP/1.1 200 OK
< Server: nginx/1.14.0
< X-Served-By: mediawiki-sandbox-sus2-8c557c9d8-kn2sl
```

docker/maintenance/cronjob-template.yaml
```diff
@@ -15,10 +15,16 @@ spec:
           - name: php
             image: artifactory.wikia-inc.com/sus/mediawiki-prod-php:${label}
             env:
-             - name: SERVER_ID
-               value: "${server_id}"
-             - name: LOG_STDOUT_ONLY
-               value: "yes"
+              - name: KUBERNETES_DEPLOYMENT_NAME
+                value: mediawiki-prod
```
rogatty: Is there a reason to not use `mw-cj-${name}` here and avoid calling production pods and slowing them down?

docker/prod/prod.template.yaml
```diff
@@ -246,6 +252,12 @@ spec:
               port: 9000
             timeoutSeconds: 5
           env:
+            - name: KUBERNETES_DEPLOYMENT_NAME
+              value: mediawiki-prod
```
rogatty: Same question as https://github.com/Wikia/app/pull/15648/files#r204779888 but with `mediawiki-tasks`

includes/HttpFunctions.php
```diff
@@ -430,6 +428,12 @@ public function getHeaderList() {
 				);
 		}
 
+		// SUS-5499: Use internal host name for MW->MW requests when running on Kubernetes
```
rogatty, line 431: Do we no longer follow the `// Wikia change` convention?

docker/maintenance/cronjob-template.yaml
macbre: Well, there's no nginx here :)

rogatty: Is it the same in the old setup? I just wonder if we won't overload the production pods with heavy queries which could be confined to the cronjob pods. Would it be worth to add the nginx here?

docker/prod/prod.template.yaml
macbre: Changed in 41e6da58164

docker/maintenance/cronjob-template.yaml
macbre: I'd actually move these requests to `mediawiki-tasks`, so that the offline traffic will be separated from end-user requests. Good point!

includes/HttpFunctions.php
```diff
@@ -430,6 +428,12 @@ public function getHeaderList() {
...
+		global $wgKubernetesNamespace;
+		if ( !empty( $wgKubernetesNamespace ) && Http::isLocalURL( $this->url ) ) {
+			$list[] = sprintf( 'X-Original-Host: %s', $this->parsedUrl['host'] );
```
rogatty, line 434: It smells a bit to have an `if` like that but I don't have a better idea where to put it.

rogatty (approved):

includes/HttpFunctions.php
macbre: You know me, I like smelly things ;)

PR #15650, opened 2018-07-09
SUS--1 | Small linter fixes

hakubo:
Started to use vscode
and it marks folders as red when there are problems in it

I don't like my folders red - so I fixed most of what it complains about

### DIFF
--- a/extensions/wikia/Wall/templates/WallHistory_index.php
+++ b/extensions/wikia/Wall/templates/WallHistory_index.php
@@ -1,5 +1,5 @@
 <div class="WallHistory">
-<? if ( empty( $wallmessageNotFound ) ): ?>
+<?php if ( empty( $wallmessageNotFound ) ): ?>
 	<div class="SortingBar">
 		<div id="pageTitle">
 			<?= $pageTitle; ?>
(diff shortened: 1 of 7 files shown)


macbre:
http://php.net/manual/en/language.basic-syntax.phptags.php says that

> PHP also allows for short open tag `<?` (which is discouraged since it is only available if enabled using the `short_open_tag` `php.ini` configuration file directive, or if PHP was configured with the` --enable-short`-tags option).

So, let's replace `<?` with `<?php`.

mszabo-wikia:
yep, I'm not sure why the linter is complaining there since it is valid syntax

extensions/wikia/Wall/templates/WallHistory_index.php
```diff
@@ -125,7 +125,7 @@
 			<?= $app->renderView( 'PaginationController', 'index', [ 'totalItems' => $totalItems, 'itemsPerPage' => $itemsPerPage, 'currentPage' => $currentPage, 'url' => $wallHistoryUrl ] ); ?>
 		<?php endif;?>
 	<?php endif; ?>
-<?php else: ?>
+<? else: ?>
```
macbre: So this change (and the one below) is not needed, right?

hakubo: grr... missed it :)

macbre (approved):

PR #15684, opened 2018-07-11
SUS-5455 | founder-emails jobs + cronjob-generator.sh

hakubo:
`./cronjob-generator.sh` 
generates all cron job yamls

I am sure it can be done simplier - open to suggestions
currently it needs
```
sudo apt install jq
```

with relative ease we could also support multi jobs per file 
so all founder emails jobs could be in one files

alternatively we could take file name and use it as a job name
so each yaml would only need
args and schedule
and optionally server_id


docker/maintenance/cronjob-generator.sh
```diff
@@ -0,0 +1,38 @@
...
+		exit 1;
+	fi
+
+	if [ "$SCHEDULE" = 'null' ]; then
+		echo "schedule has to be set in $job_description_file_name"
+		exit 1;
+	fi
+
+	if [ "$ARGS" = 'null' ]; then
+		echo "schedule has to be set in $job_description_file_name"
```
macbre: It should say `args`.

macbre:
I share the idea of using filenames instead of duplicating it in `name` field :+1:

macbre:
According to Chef we have `jq` package installed site-wide.

macbre (approved):

docker/maintenance/cronjob-template.yaml
```diff
@@ -0,0 +1,27 @@
+---
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: {name}
```
macbre: Let's prefix the name of cronjob (maybe with `mediawiki-`?) so it's easy to distinguish them from other Kubernetes jobs.

docker/maintenance/cronjob-generator.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+## usage ./cronjob-generator
+SCRIPT_FOLDER=$(dirname $(readlink -f $0))
+JOB_DESCRIPTIONS=(`ls $SCRIPT_FOLDER | grep -v 'example' | grep -v 'cronjob-template.yaml' | grep '.yaml'`)
```
sqreek: instead of filtering this way you can pass list of files as an argument
what that means you don't need to make a script that handle all files in iteration, but script that handle one file, 
then you can encapsulate this line in other file that call `handle one file` script

docker/maintenance/cronjob-generator.sh
```diff
@@ -0,0 +1,38 @@
...
+TEMPLATE=`cat $SCRIPT_FOLDER/cronjob-template.yaml`
+
+for job_description_file_name in "${JOB_DESCRIPTIONS[@]}"; do
+	JOB_JSON=`js-yaml $SCRIPT_FOLDER/$job_description_file_name`
+	NAME=`echo "$JOB_JSON" | jq .name`
```
sqreek: you can use jq to extract all of this at once
fe `js-yaml founder-emails-complete-digest.yaml | jq '(.name | tostring)  + (.schedule | tostring)'`

docker/maintenance/cronjob-generator.sh
```diff
@@ -0,0 +1,38 @@
...
+	SCHEDULE=`echo "$JOB_JSON" | jq .schedule`
+	ARGS=`echo "$JOB_JSON" | jq .args | js-yaml | sed -e 's/^/            /g'`
```
sqreek: as above (instead counting spaces)

sqreek: js-yaml founder-emails-complete-digest.yaml | jq .args[]  -r

docker/maintenance/cronjob-generator.sh
```diff
@@ -0,0 +1,38 @@
...
+		echo "schedule has to be set in $job_description_file_name"
+		exit 1;
+	fi
+
+	if [ "$ARGS" = 'null' ]; then
+		echo "schedule has to be set in $job_description_file_name"
+		exit 1;
+	fi
+
+	job_k8s_descriptor="${TEMPLATE/\{name\}/${NAME}}"
```
sqreek: in sed you can specify all replaces in one command
`sed 's/ab/~~/g; s/bc/ab/g; s/~~/bc/g'`

docker/maintenance/cronjob-generator.sh
macbre: You can at least filter out YAML-less ;) files at ls level - `ls $SCRIPT_FOLDER *.yaml` instead of later `grep`.

hakubo: actually that won't work
as ls does not have any filter option
if you do 
ls *.yaml
you're actually using globbing of your shell and pass it to ls for formatting only 

so to make it work
we'd have to 
ls $SCRIPT_FOLDER/*.yaml

but then it gives me file names with full paths that I'd have to clean there
so I think it's easier to leave it as is

docker/maintenance/cronjob-generator.sh
hakubo: yes you can - are there benefits to it?

docker/maintenance/cronjob-generator.sh
hakubo: neat, did not know that
but then it gives me a long string that I have to cast to an array I guess?
and then use array in script?

macbre (approved):

PR #15712, opened 2018-07-13
SUS-5527 | Configure rolling update for MediaWiki on Kubernetes

mszabo-wikia:
* add readiness and liveness probes for nginx and php containers
* define deadline for deployment
* verify and report deployment rollout status and report it to slack

https://wikia-inc.atlassian.net/browse/SUS-5527


macbre (approved):

docker/prod/prod.template.yaml
```diff
@@ -20,10 +26,34 @@ spec:
...
+              path: /health/check
+              port: 8080
+            initialDelaySeconds: 60
+            timeoutSeconds: 5
+            periodSeconds: 20
+          readinessProbe:
+            httpGet:
+              path: /health/check
+              port: 8080
+            initialDelaySeconds: 5
```
r0bj: `readinessProbe.initialDelaySeconds` will be executed after `livenessProbe.initialDelaySeconds` so it will be 65 seconds until pod is ready for traffic. Is it intended?

mszabo-wikia: image pull/container creation takes around 1 minute so I think this should be oK

r0bj: But I think the timer starts when image is already pulled.

hakubo (approved):

macbre (approved):

PR #15713, opened 2018-07-16
PLATFORM-3575: initial HTTPS redirects cleanup

harnash:
Introduce late HTTP <-> HTTPS redirects to allow some Wiki Factory logic to kick in before we issue 30x to a proper protocol.

ping: @Wikia/core-platform-team, @macbre

### DIFF
--- a/includes/OutputPage.php
+++ b/includes/OutputPage.php
@@ -134,6 +134,10 @@ class OutputPage extends ContextSource {
 
 	var $mRedirectCode = '';
 
+	# start wikia change
+	var $mRedirectProtocol = PROTO_CURRENT;
+	# end wikia change
+
 	var $mFeedLinksAppendQuery = null;
 
 	/**
@@ -256,6 +260,15 @@ class OutputPage extends ContextSource {
 			$this->setContext( $context );
 		}
 	}
+
+	# start wikia change
+	public function redirectProtocol( $protocol, $responsecode = '302' ) {
+		if ( $protocol === PROTO_HTTP || $protocol === PROTO_HTTPS ) {
+			$this->mRedirectProtocol = $protocol;
+			$this->mRedirectCode = $responsecode;
+		}
+	}
+	# end wikia change
 
 	/**
 	 * Redirect to $url rather than displaying the normal page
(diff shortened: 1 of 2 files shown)


includes/OutputPage.php
```diff
@@ -257,6 +261,15 @@ function __construct( IContextSource $context = null ) {
 		}
 	}
 
+	# start wikia change
+	public function redirectProtocol( $protocol, $responsecode = '302' ) {
+		if ( $protocol === PROTO_HTTP || $protocol === PROTO_HTTPS ) {
+			$this->mRedirectProtocol = $protocol;
+			$this->mRedirectCode = $responsecode;
```
harnash, line 268: I wonder if we should override `mRedirectCode` if there is set one already? 🤔

includes/OutputPage.php
```diff
@@ -2065,9 +2078,19 @@ public function output() {
...
+
+			if ( $this->mRedirectProtocol === PROTO_HTTP ) {
+				$this->mRedirect = wfHttpsToHttp( $this->mRedirect );
+			} elseif ( $this->mRedirectProtocol === PROTO_HTTPS ) {
+				$this->mRedirect = wfHttpToHttps( $this->mRedirect );
+			}
+
 			# Standards require redirect URLs to be absolute
-			$this->mRedirect = wfExpandUrl( $this->mRedirect, PROTO_CURRENT );
+			$this->mRedirect = wfExpandUrl( $this->mRedirect, $this->mRedirectProtocol );
```
harnash, line 2093: This will change protocol if we are dealing with relative urls only

includes/OutputPage.php
Grunny: For HTTP -> HTTPS we'll want it to be a 301, so I think that's probably worth being able to do here. Article redirects in MW are already 301s, but we'll want to make sure it's a 301 if it's going from HTTP -> HTTPS (though for now, going HTTPS->HTTP are 302s).

harnash: Udpated in 24ad37f7f54

macbre (approved):

includes/OutputPage.php
```diff
@@ -134,6 +134,10 @@ class OutputPage extends ContextSource {
 
 	var $mRedirectCode = '';
 
+	# start wikia change
```
themech, line 137: nit: in general we use slashes

themech (approved):

PR #15722, opened 2018-07-16
SUS-5533 | Use fixed worker count for PHP FPM

mszabo-wikia:
https://wikia-inc.atlassian.net/browse/SUS-5533


hakubo:
Having to update it in 7 places seems error prone. Can we simplify it?

macbre (approved):

PR #15725, opened 2018-07-17
XW-5105 | Bring back Global nav controls on mobile devices for oasis

rybmat:
https://wikia-inc.atlassian.net/browse/XW-5105

@Wikia/x-wing


bkoval (approved):

extensions/wikia/DesignSystem/styles/design-system.scss
```diff
@@ -20,6 +20,10 @@ $global-navigation-z-index: 5000101;
 
 .wds-global-navigation-wrapper {
 	height: $wds-global-navigation-height;
+
+	@media screen and (max-width: 768px) {
+		width: 768px;
```
hakubo, line 25: I think you could achieve the same with min width, right?

rybmat: yes, but does it make any difference? above 768px there is other breakpoint so this style will not be applied

PR #15738, opened 2018-07-18
ADEN-7303 Disables CMP when the new one is defined

leshniak:


extensions/wikia/TrackingOptIn/js/cmp.js
```diff
@@ -232,6 +232,11 @@ define('wikia.cmp', [
 	}
 
 	function isEnabled() {
```
fraszczakszymon, line 234: This method is used in a9.js. Is it safe to force `false`?

PR #15739, opened 2018-07-18
SUS | dockerignore the docker directory

mszabo-wikia:
changes there don't need to be added to the image

### DIFF
--- a/docker/.dockerignore
+++ b/docker/.dockerignore
@@ -16,3 +16,5 @@
 
 # node_modules in app root weights over 130 MB
 app/node_modules
+
+app/docker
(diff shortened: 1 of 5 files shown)


docker/.dockerignore
```diff
@@ -16,3 +16,5 @@
 
 # node_modules in app root weights over 130 MB
 app/node_modules
+
+app/docker
```
hakubo, line 20: I can see nginx config in docker folder
aren't they needed?

mszabo-wikia: I will check

macbre (approved):

PR #15746, opened 2018-07-18
XW-3053 | Fix missing PI accent colors in horizontal group headers 

lkucharczyk:
Accent colors doesn't apply to horizontal group headers in portable infoboxes, because inlineStyles variable isn't used in creation of horizontal groups. 

[Bug example](https://luq.wikia.com/wiki/PI_Accent)
Jira issue: [XW-3053](https://wikia-inc.atlassian.net/browse/XW-3053)


KockaAdmiralac:
Does a comment count as activity? Hopefully it does, because this PR is yet to be reviewed by Staff and not relevant to the stale limit at all.

PR #15753, opened 2018-07-19
SUS-5508 | Remove PId and add concurencyPolicy to dumps on demand

hakubo:


ghost (approved):

rogatty:
If I understand https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/ correctly, then this cron job won't be scheduled if the script takes more than 500 minutes:
>If there are more than 100 missed schedules, then it does not start the job and logs the error.
>A CronJob is counted as missed if it has failed to be created at its scheduled time. For example, If concurrencyPolicy is set to Forbid and a CronJob was attempted to be scheduled when there was a previous schedule still running, then it would count as missed.

rogatty:
Superseded by https://github.com/Wikia/app/pull/15770

PR #15766, opened 2018-07-20
SUS-4713 | Fix for PI galleries not working with tabber

lkucharczyk:
On Mercury, when the tabber is detected, it tries to extract non-existing "data-ref" parameter, which is present only when the image is a part of a gallery. This fix will use data-file attribute instead, which is tied with every image on Mercury.

Jira issue: [SUS-4713](https://wikia-inc.atlassian.net/browse/SUS-4713)


lkucharczyk:
It can wait, no hurry.

lkucharczyk:
#16013

PR #15775, opened 2018-07-23
SUS-5672 | publish Slack notifications on production deployments

macbre:
https://wikia-inc.atlassian.net/browse/SUS-5672

### DIFF
--- a/docker/prod/Jenkinsfile
+++ b/docker/prod/Jenkinsfile
@@ -5,8 +5,7 @@ def loggerImage = "artifactory.wikia-inc.com/sus/mediawiki-logger"
 
 def rolloutStatus = 1
 
-// Slack notifications
-env.MW_PUSH_CHANNEL="#mw-prod-deployment"
+def slackChannel = "#mw-prod-deployment"
 
 def render(String template, java.util.LinkedHashMap binding) {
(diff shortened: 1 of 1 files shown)


docker/prod/Jenkinsfile
```diff
@@ -5,6 +5,9 @@ def loggerImage = "artifactory.wikia-inc.com/sus/mediawiki-logger"
 
 def rolloutStatus = 1
 
+// Slack notifications
+env.MW_PUSH_CHANNEL="#mw-prod-deployment"
```
rogatty: I think we need to set `env.MW_SANDBOX_K8S_WEBHOOK` too but this is a secret URL. Those variables should probably be passed from Jenkins.

macbre: This one is defined in Jenkins (as a "global" env variable there).

rogatty: So how about using this instead?
```
def slackChannel = "#mw-prod-deployment"
```
It's not an environment variable, after all.

macbre: Good point - e3feb5a

rogatty (approved):

PR #15785, opened 2018-07-24
ADEN-7450 Disables CMP when the new one is defined

fraszczakszymon:


extensions/wikia/TrackingOptIn/js/cmp.js
```diff
@@ -232,6 +232,11 @@ define('wikia.cmp', [
 	}
 
 	function isEnabled() {
+		if (typeof win.__cmp === 'function' && win.isConsentManagementProviderLoadedFromTrackingOptInModal) {
```
jacekbj: Do we need to keep it if we do not use it?

extensions/wikia/TrackingOptIn/js/trackingOptIn.js
```diff
@@ -31,6 +31,9 @@ define('wikia.trackingOptIn', [
 				zIndex: 9999999
 			});
 			geoRequiresConsent = instance.geoRequiresTrackingConsent();
+
+			// TODO: Remove this flag once we fully switch to CMP from TrackingOptIn
```
jacekbj: Can you please add a link to a Jira ticket to make finding it later easier?

jacekbj (approved):

dmnsobczak (approved):

PR #15786, opened 2018-07-24
SRE-89 | Remove wfWaitForSlaves call from VET

mszabo-wikia:
Video upload works fine even from RES without this call.
https://wikia-inc.atlassian.net/browse/SRE-89

### DIFF
--- a/extensions/wikia/VideoEmbedTool/VideoEmbedTool_body.php
+++ b/extensions/wikia/VideoEmbedTool/VideoEmbedTool_body.php
@@ -254,9 +254,6 @@ class VideoEmbedTool {
 		$vHelper = new VideoHandlerHelper();
 		$vHelper->setVideoDescription( $oTitle, $description );
 
-		// SUS-66: let's wait for all slaves to catch up after the above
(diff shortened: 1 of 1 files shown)


ljagiello (approved):
How this will affect original issue: https://wikia-inc.atlassian.net/browse/SUS-66 ?

ping @macbre

ghost:
@macbre is on vacation until the end of August, I believe. As for `wfWaitForSlaves` - @mszabo-wikia if you remove some changes, investigate why they had been made in the first place and test it properly. @ljagiello linked some other case. It's not "make it work no matter what". It's "make it work and also don't break anything else" :) Otherwise moving to active-active will create more problems than it is supposed to solve :)

mszabo-wikia:
@mixoteric the failing code in the original ticket was removed fully, and there was no indication why wfWaitForSlaves was added—the remaining logic doesn't expect newly updated data to be present in slave. Inserting video via placeholder now works fine even in RES where replication delay is not insignificant (although tests still fail because of timeout as the request still takes 14 secs down from original 38).

PR #15791, opened 2018-07-25
refactor resetWeeklyUserContributionsCount.php for better logging

ghost:
https://wikia-inc.atlassian.net/browse/SUS-5469

### DIFF
--- a/maintenance/wikia/cronjobs/resetWeeklyUserContributionsCount.php
+++ b/maintenance/wikia/cronjobs/resetWeeklyUserContributionsCount.php
@@ -15,7 +15,7 @@ class ResetWeeklyUserContributionsCount extends Maintenance {
 
 	public function __construct() {
 		parent::__construct();
-		$this->mDescription = 'reset weekly user contribution count';
+		$this->mDescription = 'reset weekly user contributions count';
 	}
 
 	public function execute() {
--- a/maintenance/wikia/cronjobs/resetWeeklyUserContributionsCount.php
+++ b/maintenance/wikia/cronjobs/resetWeeklyUserContributionsCount.php
@@ -1,22 +1,21 @@
 <?php
-
-use \Wikia\Logger\WikiaLogger;
-
 /**
  * On Special:Community we have weekly users rank with most contributions.
- * We need to reset it after every Sunday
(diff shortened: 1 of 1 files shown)


maintenance/wikia/cronjobs/resetWeeklyUserContributionsCount.php
```diff
@@ -1,19 +1,18 @@
...
 
-require_once( __DIR__ . '/../../Maintenance.php' );
+require_once __DIR__ . '/../../Maintenance.php';
 
-/**
- * Class ResetWeeklyUserContributionsCount
- */
 class ResetWeeklyUserContributionsCount extends Maintenance {
 
+	use Wikia\Logger\Loggable;
```
macbre, line 14: `Maintenance` class has `error` method defined. How will it cooperate with `Loggable` trait?

> An inherited member from a base class is overridden by a member inserted by a Trait. The precedence order is that members from the current class override Trait methods, which in turn override inherited methods.

I believe that's fine for us, right?

rogatty (approved):

maintenance/wikia/cronjobs/resetWeeklyUserContributionsCount.php
rogatty: This will fail with `PHP Fatal error:  Trait 'Wikia\Logger\Loggable' not found` because according to @macbre (I couldn't find it in the code) the classes are autoloaded just before `execute()` is called.

rogatty: See https://github.com/Wikia/app/pull/15797/commits/fffa65184ca0368d7030f8de6bbe9cc747928f32

rogatty, line 14: This will fail with `PHP Fatal error:  Trait 'Wikia\Logger\Loggable' not found` because according to @macbre (I couldn't find it in the code) the classes are autoloaded just before `execute()` is called. See https://github.com/Wikia/app/pull/15797/commits/fffa65184ca0368d7030f8de6bbe9cc747928f32.

rogatty (changes requested):

PR #15792, opened 2018-07-25
ADEN-7420 Call bill the lizard and track data

fraszczakszymon:


extensions/wikia/AssetsManager/config.php
```diff
@@ -309,6 +309,14 @@
 	],
 ];
 
+$config['adengine2_bridge_js'] = [
```
jacekbj: Please explain to me (at your convenience) how groups differ from files. Idk why `adengine2_bridge_js` is a one-file group.

jacekbj (approved):

extensions/wikia/AssetsManager/config.php
fraszczakszymon: It isn't really different. Group with one asset works in the same way as single asset. I left it somewhere in the middle of refactoring and fixing doubled bridge. Fixed :+1:

jacekbj (approved):

PR #15793, opened 2018-07-25
PLATFORM-3592: Set surrogate keys headers in WikiFactoryLoader

themech:
So WFL redirects are also covered

### DIFF
--- a/includes/wikia/Wikia.php
+++ b/includes/wikia/Wikia.php
@@ -1950,6 +1950,26 @@ class Wikia {
 		return wfGetEffectiveHostname() . '-wiki-' . $wikiId;
 	}
 
+	/**
+	 * Add surrogate key(s) preserving those previously emitted.
+	 * @param string|array $surrogateKeys Surrogate keys (array or space-delimited string)
+	 */
+	public static function attachSurrogateKeysToHeaders( $surrogateKeys ) {
+		if ( !is_array( $surrogateKeys ) ) {
+			$surrogateKeys = [ $surrogateKeys ];
+		}
+		// get existing surrogate keys
+		$headers = headers_list();
+		foreach ( $headers as $header ) {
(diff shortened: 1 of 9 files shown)


includes/wikia/Wikia.php
```diff
@@ -1950,6 +1950,26 @@ public static function wikiSurrogateKey( $wikiId ) {
...
+		// get existing surrogate keys
+		$headers = headers_list();
+		foreach ( $headers as $header ) {
+			if ( strtolower( substr( $header, 0, 14 ) ) === 'surrogate-key:' ) {
+				$surrogateKeys[] = trim( substr( $header, 14 ) );
+			}
+		}
+		$surrogateKey = implode( ' ', $surrogateKeys );
+		header( 'Surrogate-Key: ' . $surrogateKey );
+		header( 'X-Surrogate-Key: ' . $surrogateKey );
```
mszabo-wikia, line 1977: What do you think of storing surrogate keys in some array and only emitting headers at the end of the request?

themech: It gets tricky if the request ends with a WikiFactoryLoader, because one would have to make sure the headers are set based on this array

mszabo-wikia: We could call it in RestInPeace and also in LocalSettings.php if WFL aborted processing

themech: given that we're rarely set custom surrogate keys and also we would like to make sure every response (even the one that ended with an exception and due to the body being sent headers cannot be added anymore) has a wiki-wide surrogate key for purging, I'll slightly modify the approach. But still, I would like to emit the header in WikiFactoryLoader to make sure every response has it set

themech: @mszabo-wikia I've changed the method so WFL simply sets the header without processing existing ones. This way only few places are appending those and we also ensure headers set in WFL are present in every response.

harnash (approved):

PR #15796, opened 2018-07-25
SRE-97 | Save user's attributes in bulk

mszabo-wikia:
Currently we save user's attributes to UAS one-by-one when `User::saveSettings` is called. This is a significant performance bottleneck in an active-active setup. Since UAS already has support for bulk updates via a PATCH request, we can utilize that endpoint to update all attributes via a single request. Further, when `User::saveSettings` is called, we only need to update those attributes whose values actually changed during the request (e.g. due to a `User::setGlobalAttribute` call)—other attributes will either have the value that is already in the service, or will have a default value of configuration that wouldn't be saved to the service.

This PR
* regenerates the UAS client in 15078b9
* adds an integration test for saving user's attributes
* makes the save operation use a single bulk update for attributes.

https://wikia-inc.atlassian.net/browse/SRE-97


ljagiello (approved):
With my limited PHP knowledge it looks fine.

lib/Wikia/src/Service/User/Attributes/UserAttributes.php
```diff
@@ -81,35 +85,42 @@ private function loadAttributes( $userId ) {
...
 		$this->loadAttributes( $userId );
 
+		$name = $attribute->getName();
+		$value = $attribute->getValue();
+
 		// If attribute value is null and default exists, set value to default
-		if ( is_null( $attribute->getValue() ) && isset( $this->defaultAttributes[$attribute->getName()] ) ) {
-			$attribute->setValue( $this->defaultAttributes[$attribute->getName()] );
+		if ( is_null( $value ) && isset( $this->defaultAttributes[$name] ) ) {
+			$attribute->setValue( $this->defaultAttributes[$name] );
```
jsutterfield, line 93: Shouldn't this be `$value = $this->defaultAttributes[$name]`?

jsutterfield: So actually, I think this would be fine. Since when we call `getAttribute` for this we'll see the value is `null` and return the default, and when we call `save` well see that the value is the default and skip it.

It still feels a little weird that at times we can have 2 different values for an attribute: the default in `$this->changedAttributes[$name]` and `null` in `$this->attributes[$userId][$name]`. What about not using the local `$value` variable and just using the value from the object? Or we could make `changedAttributes` just an array with attribute names and use those to key into `$this->attributes[$userId][$name]` to get the value.

jsutterfield (approved):
LGTM, just one small suggestion.

Also, as the one who wrote the TODO comment to convert this to a bulk save, thank you for doing this!

Grunny (approved):

lib/Wikia/src/Service/User/Attributes/UserAttributes.php
mszabo-wikia: Thanks! Fixed in 59b1c74192089328745b9d63ccdb592511c28c77.

PR #15815, opened 2018-07-27
Increase successful cron job history limit

rogatty:
I noticed that when `successfulJobsHistoryLimit` is set to 0 we don't have any useful logs in Kibana. I suspect the pod is deleted before the logstash picks up the logs. Let's keep a single pod per cron job to have the logs for debugging purposes.

### DIFF
--- a/docker/maintenance/cronjob-template.yaml
+++ b/docker/maintenance/cronjob-template.yaml
@@ -5,7 +5,7 @@ metadata:
   name: "mw-cj-${name}"
 spec:
   schedule: ${schedule}
-  successfulJobsHistoryLimit: 0
+  successfulJobsHistoryLimit: 1
   failedJobsHistoryLimit: 1
   jobTemplate:
     spec:


macbre:
I do see the logs from crons that run successfully. `syslogng` on our virtual hosts is set up to forward logs to logstash using UDP. I believe Kubernetes is set up similarly.  The "reliability" of UDP can be the reason of missing logs.

rogatty:
@macbre do you have an example I can look at? From my experience the MediaWiki logs inside k8s cron jobs are only present if the job failed or if I changed this setting.

macbre:
`mw-cj-dump-starters` looks fine - https://kibana.wikia-inc.com/goto/505e7c8ca7680bf01bb0a7744391708a

rogatty:
Interesting. There are only k8s logs (nothing from mediawiki) for `mw-cj-update-special-pages` where I forgot to change this param before applying the descriptor: https://kibana.wikia-inc.com/goto/de87fcf432cac2905b4cc787e61117e8
I'm sure it ran successfully because all the tasks were scheduled as expected.

ghost (approved):

ghost:
@macbre - I think, that if nature of UDP (package loss) was the sole reason of missing log entries, we would already have way more complains about it. Unless, of course, the nature of containers and underlying structure somehow contributes to more severe package loss in UDP.  Is there a quick way to switch to TCP and see if that changes anything?

rogatty:
I'm going to merge it as it helps with testing the cron jobs conversion. We can investigate the root cause and improve this independently.

PR #15828, opened 2018-07-30
PLATFORM-3567: merge WFL and MediaWiki redirects - poc

themech:
Major changes:
* WFL redirects can be intercepted and merged later with Main page/Title redirects
* X-Redirected-By header can include several space-separated values
* WFL redirects preserve query parameters and url fragments


index.php
```diff
@@ -48,11 +48,37 @@
...
+	public function getRedirectedBy() {
+		return $this->redirectedBy;
+	}
+};
+
+// This will intercept wiki redirects made in WikiFactoryLoader and pass
+// them later to mediawiki to merge with other redirects (like the main page).
+// This allows us to return a single redirects response instead of a chain of
+// 301s.
+$wgWFLRedirect = new WFLRedirect();
```
themech: Note to myself: replace `wgWFLRedirect` global with a hook listener

extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
```diff
@@ -404,51 +402,58 @@ public function execute() {
...
-				!empty( $_COOKIE["{$wgCookiePrefix}UserID"] );
+			$redirectedBy = [];
+			if ( $shouldUpgradeToHttps ) $redirectedBy[] = 'AnonsHTTPSUpgrade';
+			if ( $cond1 ) $redirectedBy[] = 'NotPrimary';
+			if ( $cond2 ) $redirectedBy[] = 'AlternativeDomain';
+			$redirectedBy = join( ' ', $redirectedBy );
 
-			if ( $hasAuthCookie ) {
-				header( 'Cache-Control: private, must-revalidate, max-age=0' );
+			global $wgWFLRedirect;
```
themech: Note: replace with a hook call, name is something like `WikiFactoryLoaderRedirect`

extensions/wikia/SitemapXml/SpecialSitemap_body.php
```diff
@@ -43,9 +43,6 @@ public function execute( $subpage ) {
 
 		if ( ( $showIndex && !$forceOldSitemap ) || $forceNewSitemap ) {
 			$this->getOutput()->disable();
-			if ( !Hooks::run( 'SitemapPageBeforeOutput', [ $this->getRequest(), $this->getUser() ] ) ) {
```
Grunny, line 46: This will also stop us handling the HTTPS -> HTTP downgrade in the app once we want to scale this and remove Fastly from the equation. I think we should still handle downgrading here, or is this being handled in `onBeforeInitialize`?

themech: Good point. I'll have to think about the downgrades in `onBeforeInitialize`. Because if that part of the code or WFL decides to redirect, this execute method won't be called.

extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
themech: or not. we cannot easily register a hook listener inside index.php, as the `Hooks` class and `$wgHooks` do not exist yet

extensions/wikia/SitemapXml/SpecialSitemap_body.php
themech: discussed on Slack, it should be fine now as `onBeforeInitialize` prevents the sitemap code from ignoring the Output page redirect.

Grunny (approved):

PR #15832, opened 2018-07-30
cake-3437 - separate function to get fandom creator variables

nmonterroso:
@Wikia/cake @fraszczakszymon @vforge 
https://wikia-inc.atlassian.net/browse/CAKE-3437

We're integrating ads into the FC, and rather than relying on the existing n&s functionality we want to have our own function for getting these variables so they don't change for our app without explicit intent. 

I'm updating the helper function to be a bit more flexible to support both apps but the output won't change. Unfortunately this seems to only be testable in production since those variables aren't set in dev.


lizlux:
LGTM

JoshuaRogan (approved):

nmonterroso:
I deployed this to sandbox-s1 and both the n&s and fc variables are showing up as expected

PR #15834, opened 2018-07-31
SRE-62 | Remove dev env specific timeout from TemplateClassificationService

mszabo-wikia:
After https://wikia-inc.atlassian.net/browse/SRE-62 changes it should not be needed any more.


mszabo-wikia:
failing jenkins job is not related (it's disabled now)

PR #15838, opened 2018-08-01
ADEN-7248 Clean after geo module replacement.

jacekbj:


extensions/wikia/AdEngine/AdEngine2Hooks.class.php
```diff
@@ -88,7 +88,6 @@ public static function onInstantGlobalsGetVariables( array &$vars ) {
 		$vars[] = 'wgAdDriverN1DecisionTreeClassifierRabbitCountries';
 		$vars[] = 'wgAdDriverN1LogisticRegressionRabbitCountries';
 		$vars[] = 'wgAdDriverNetzAthletenCountries';
-		$vars[] = 'wgAdDriverNewGeoCountries';
```
fraszczakszymon: Unfortunately we need to remove this var in separate release (in order to avoid missing wg value for cache)

fraszczakszymon: In the same PR we will be able to get rid of:
https://github.com/Wikia/app/pull/15838/files#diff-ac1bcc3ae7cf3e1fa4ec596d457b1bdcR242
am I right?

jacekbj: Regarding: https://github.com/Wikia/app/pull/15838/files#diff-ac1bcc3ae7cf3e1fa4ec596d457b1bdcR242
There are some other modules which use geo module, I don't know if they do not happen to use geo.js from that place.

dmnsobczak (approved):

PR #15840, opened 2018-08-01
 SUS-5694 | Use the target namespace instead of the source namespace

rogatty:
https://wikia-inc.atlassian.net/browse/SUS-5694

Follow up to https://github.com/Wikia/app/pull/15836

`pl_namespace` is the namespace of the target page. `pl_title` is not prefixed, colons there are a part of the title, e.g. https://harrypotter.wikia.com/wiki/Harry_Potter:_Hogwarts_Mystery is in `NS_MAIN`.

Group by target namespace so the DB uses the `pl_namespace` index.

### DIFF
--- a/extensions/wikia/Search/classes/IndexService/Evaluation.php
+++ b/extensions/wikia/Search/classes/IndexService/Evaluation.php
@@ -7,7 +7,6 @@ use MWNamespace;
 
 class Evaluation extends AbstractService {
 	const DISABLE_BACKLINKS_COUNT_FLAG = 'disable_backlinks_count';
-	const BACKLINKS_FROM_ALL_NAMESPACES_FLAG = 'backlinks_from_all_namespaces';
 	const LANGUAGES_SUPPORTED = ['en', 'de', 'es', 'fr', 'it', 'ja', 'pl', 'pt-br', 'ru', 'zh', 'zh-tw'];
 
 	/**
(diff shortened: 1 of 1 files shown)


extensions/wikia/Search/classes/IndexService/Evaluation.php
```diff
@@ -73,38 +72,40 @@ protected function processAllDocuments( $documents ) {
...
+		foreach ( $titlesByNamespace as $namespace => $titles ) {
+			$dbResults = $dbr->select(
+				'pagelinks',
+				[ 'count(*) as cnt', 'pl_title' ],
+				[
+					'pl_namespace' => $namespace,
+					'pl_title' => $titles
+				],
+				__METHOD__,
+				[ 'GROUP BY' => 'pl_title' ]
```
mszabo-wikia, line 102: We need to GROUP BY both columns.

mszabo-wikia (changes requested):

extensions/wikia/Search/classes/IndexService/Evaluation.php
rogatty: `pl_namespace` is the same for all rows so I don't think that's true. Also, I checked and it works ;)

mszabo-wikia (approved):

PR #15845, opened 2018-08-02
(PLATFORM-3616) Make external links to other wikis protocol-relative

Grunny:
Make external links to other wikis protocol-relative for URLs to communities
that support HTTPS. This means we won't need users who paste URLs to other wikis
to update them.

/cc @Wikia/core-platform-team

### DIFF
--- a/extensions/wikia/HTTPSSupport/HTTPSSupportHooks.class.php
+++ b/extensions/wikia/HTTPSSupport/HTTPSSupportHooks.class.php
@@ -91,6 +91,29 @@ class HTTPSSupportHooks {
 		return true;
 	}
 
+	public static function parserUpgradeVignetteUrls( string &$url ) {
+		if ( preg_match( self::VIGNETTE_IMAGES_HTTP_UPGRADABLE, $url ) && strpos( $url, 'http://' ) === 0 ) {
+			$url = wfHttpToHttps( $url );
+		}
+	}
+
+	/**
+	 * Make sure any "external" links to our own wikis that support HTTPS
+	 * are protocol-relative on output.
+	 *
+	 * @param  string  &$url
+	 * @param  string  &$text
+	 * @param  bool    &$link
+	 * @param  array   &$attribs
+	 * @return boolean
+	 */
+	public static function onLinkerMakeExternalLink( string &$url, string &$text, bool &$link, array &$attribs ): bool {
+		if ( wfHttpsAllowedForURL( $url ) ) {
+			$url = wfProtocolUrlToRelative( $url );
+		}
+		return true;
+	}
+
 	private static function httpsAllowed( User $user, string $url ): bool {
 		global $wgEnableHTTPSForAnons;
 		return wfHttpsAllowedForURL( $url ) &&
@@ -101,12 +124,6 @@ class HTTPSSupportHooks {
 		global $wgDBname;
 		return array_key_exists( $wgDBname, self::$httpsArticles ) &&
 			in_array( $title->getPrefixedDBKey(), self::$httpsArticles[ $wgDBname ] );
-	}
-
-	public static function parserUpgradeVignetteUrls( string &$url ) {
-		if ( preg_match( self::VIGNETTE_IMAGES_HTTP_UPGRADABLE, $url ) && strpos( $url, 'http://' ) === 0 ) {
-			$url = wfHttpToHttps( $url );
-		}
 	}
 
 	private static function redirectWithPrivateCache( string $url, WebRequest $request ) {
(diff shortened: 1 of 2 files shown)


extensions/wikia/HTTPSSupport/HTTPSSupportHooks.class.php
```diff
@@ -91,6 +91,29 @@ public static function onSitemapRobotsPageBeforeOutput( WebRequest $request, Use
...
+	 * @param  string  &$text
+	 * @param  bool    &$link
+	 * @param  array   &$attribs
+	 * @return boolean
+	 */
+	public static function onLinkerMakeExternalLink( string &$url, string &$text, bool &$link, array &$attribs ): bool {
+		if ( wfHttpsAllowedForURL( $url ) ) {
+			$url = wfProtocolUrlToRelative( $url );
+		}
+		return true;
```
jmail, line 114: I am not sure if that return is required. Specially that information doesn't give you anything.

extensions/wikia/HTTPSSupport/HTTPSSupportHooks.class.php
```diff
@@ -91,6 +91,29 @@ public static function onSitemapRobotsPageBeforeOutput( WebRequest $request, Use
...
+	 * Make sure any "external" links to our own wikis that support HTTPS
+	 * are protocol-relative on output.
+	 *
+	 * @param  string  &$url
+	 * @param  string  &$text
+	 * @param  bool    &$link
+	 * @param  array   &$attribs
+	 * @return boolean
+	 */
+	public static function onLinkerMakeExternalLink( string &$url, string &$text, bool &$link, array &$attribs ): bool {
```
jmail, line 110: not used args. text, link, attribs.

jmail (changes requested):

extensions/wikia/HTTPSSupport/HTTPSSupportHooks.class.php
Grunny: It's required because the return is checked by the place the hook is called in order to check if a hook method wants to hijack the output completely: https://github.com/Wikia/app/blob/2fb5243/includes/Linker.php#L1063-L1065 a void return would override the output there.

extensions/wikia/HTTPSSupport/HTTPSSupportHooks.class.php
Grunny: This is a standard we use since the hook always passes all of these params here: https://github.com/Wikia/app/blob/2fb5243/includes/Linker.php#L1064 so all callers usually specify all the params the hook provides. It's not strictly necessary in cases like this where we're only using the first param (but would be if we were only using the `$attribs` param for example), but it's consistent.

extensions/wikia/HTTPSSupport/HTTPSSupportHooks.class.php
jmail: Right... Then why this functions doesn't have it? https://github.com/Wikia/app/pull/15845/files#diff-fca4065dcdb1fb06adf96c5420beadf6R94

extensions/wikia/HTTPSSupport/HTTPSSupportHooks.class.php
jmail: Right... Why then this function doesn't have it? https://github.com/Wikia/app/pull/15845/files#diff-fca4065dcdb1fb06adf96c5420beadf6R94

extensions/wikia/HTTPSSupport/HTTPSSupportHooks.class.php
Grunny: That's a different hook that only has one parameter: https://github.com/Wikia/app/blob/2fb5243/includes/parser/Parser.php#L1882

extensions/wikia/HTTPSSupport/HTTPSSupportHooks.class.php
Grunny: That's a different hook call, and due to the way it's checked, it still works as expected: https://github.com/Wikia/app/blob/2fb5243/includes/parser/Parser.php#L1882 However, it probably could explicitly return a value in this case since it's used in a conditional.

But, not returning a value here in `LinkerMakeExternalLink` https://github.com/Wikia/app/blob/2fb5243/includes/Linker.php#L1064 would break the functionality as we don't want to overwrite the output and stop the call to `Html::rawElement`, we just want to change the value of `$url` before the `Html::rawElement` call happens.

Grunny: That's a different hook that only has one parameter: https://github.com/Wikia/app/blob/2fb5243/includes/parser/Parser.php#L1882

extensions/wikia/HTTPSSupport/HTTPSSupportHooks.class.php
Grunny: That's a different hook call, and due to the way it's checked, it still works as expected: https://github.com/Wikia/app/blob/2fb5243/includes/parser/Parser.php#L1882 However, it probably could explicitly return a value in this case since it's used in a conditional.

But, not returning a value here in `LinkerMakeExternalLink` https://github.com/Wikia/app/blob/2fb5243/includes/Linker.php#L1064 would break the functionality as we don't want to overwrite the output and stop the call to `Html::rawElement`, we just want to change the value of `$url` before the `Html::rawElement` call happens.

jmail (approved):

themech (approved):

PR #15859, opened 2018-08-07
ADEN-7470 Sticky BFAB

fraszczakszymon:


skins/oasis/modules/templates/Footer_Index.php
```diff
@@ -1,5 +1,7 @@
 <footer id="WikiaFooter" class="WikiaFooter <?= $showToolbar ? '' : 'notoolbar' ?>">
-	<?= F::app()->renderView('Ad', 'Index', ['slotName' => 'BOTTOM_LEADERBOARD', 'pageTypes' => ['homepage_logged', 'corporate', 'search', 'all_ads'], 'addToAdQueue' => false]); ?>
+	<div id="bottomLeaderboardWrapper">
```
nandy-andy, line 2: Let's keep in mind this change and set data for rolling it out +1 day after the release, so we're sure cache is invalidated and all pages have new wrapper :)

nandy-andy (approved):

PR #15872, opened 2018-08-09
SEO-636 | When #REDIRECT is used make a proper 301 redirect

rogatty:
https://wikia-inc.atlassian.net/browse/SEO-636

For now do it only for anons. `?redirect=no` is still respected to stop the redirect.
The hook is called in https://github.com/Wikia/app/blob/dev/includes/Wiki.php#L400

@Wikia/core-platform-team


extensions/wikia/SEOTweaks/SEOTweaksHooksHelper.class.php
```diff
@@ -284,6 +284,36 @@ static public function onArticleViewHeader( Article $article, bool &$outputDone,
...
+	 *
+	 * @param Title $title
+	 * @param WebRequest $request
+	 * @param $ignoreRedirect
+	 * @param $target
+	 * @param Article $article
+	 * @return bool
+	 */
+	static public function onInitializeArticleMaybeRedirect(
+		Title &$title, WebRequest &$request, &$ignoreRedirect, &$target, &$article
```
Grunny: Minor: Objects are (effectively) passed by reference in PHP, so no need for the `&` on `$title`, `$request`, and `$article`.

extensions/wikia/SEOTweaks/SEOTweaksHooksHelper.class.php
```diff
@@ -284,6 +284,36 @@ static public function onArticleViewHeader( Article $article, bool &$outputDone,
...
+	): bool {
+		global $wgUser;
+
+		if ( !$wgUser->isAnon() || !$title->isRedirect() ) {
```
Grunny: `$article->getContext()->getUser()->isAnon()` is preferred over the global.

Grunny (changes requested):
I see on the sandbox, from domain aliases, this has two redirects, i.e. `http://wookieepedia.wikia.com/wiki/Main_page` -> `http://starwars.wikia.com/wiki/Main_page` -> `http://starwars.wikia.com/wiki/Main_Page`. Since this is manually setting the response headers and `initializeArticle` happens after the `MediaWiki::performRequest` method has redirected due to the domain alias. Can we get this down to one redirect when combined with https://github.com/Wikia/app/pull/15828 in this case too without making the domain alias redirect riskier?

Grunny (approved):

extensions/wikia/SEOTweaks/SEOTweaksHooksHelper.class.php
```diff
@@ -284,6 +284,34 @@ static public function onArticleViewHeader( Article $article, bool &$outputDone,
...
+	): bool {
+		if ( !$user->isAnon() || !$title->isRedirect() ) {
+			return true;
+		}
+
+		$queryParams = $request->getQueryValues();
+		unset( $queryParams['title'] );
+		$targetUrl = $output->getWikiPage()->getRedirectTarget()->getFullURL( $queryParams );
+
+		$output->redirect( $targetUrl, HTTP_REDIRECT_PERM );
```
Grunny: Hmm, actually, is `HTTP_REDIRECT_PERM` definitely defined in our stack? IIRC it's part of a pecl extension that we don't have and a quick test suggests it's not defined:
```
$ run_maintenance --id=147 --script="eval.php"
> var_dump(defined('HTTP_REDIRECT_PERM'));
bool(false)

> echo HTTP_REDIRECT_PERM;
HTTP_REDIRECT_PERM
```

Using constants in place of all these redirect codes we use is probably worthwhile though.

rogatty: But it works... :speak_no_evil: Yeah, I'll update it, might break at some point.

themech: Please use the third param of the  `redirect` method - the ` $redirectedBy`. This identifies the code that triggered a redirect.

rogatty: Done.

Grunny (approved):

PR #15873, opened 2018-08-09
OPS-11879 consul address for relay-mailers

gites:


ljagiello (approved):

ljagiello (commented):
What will happen on dev?

gites:
Consul domains are already used in dev config. https://github.com/Wikia/config/blob/dev/dev.php#L522 and https://github.com/Wikia/config/blob/dev/dev.php#L643

PR #15877, opened 2018-08-10
XW-5138 | Update Global Footer color Scheme

rybmat:
https://wikia-inc.atlassian.net/browse/XW-5183

@Wikia/x-wing


JamesHawking (approved):

extensions/wikia/DesignSystem/node_modules/design-system/dist/scss/wds-mixins/_wds-midlight.scss
```diff
@@ -0,0 +1,7 @@
+@mixin wds-midlight($background-color) {
+	background-image: linear-gradient(to top, $background-color 54%, transparent 54%, transparent 100%);
```
HM100, line 2: Change `transparent` into `rgba($background-color,0)` since Safari leaves an unexpected shade of gray when transparent keyword is used.

PR #15882, opened 2018-08-10
added founding_user_id to wikiDetails

ghost:
added founding_user_id parameter to be returned from Wikis/Details api


includes/wikia/services/WikiDetailsService.class.php
```diff
@@ -202,6 +202,7 @@ protected function getFromService( $id ) {
 				'admins' => count( $this->getWikiAdminIds( $id ) )
 			],
 			'topUsers' => array_keys( $topUsers ),
+            'founding_user_id' => isset( $modelData [$id] ) ? $modelData[$id]['founding_user_id'] : '',
```
KockaAdmiralac: These should probably be tabs instead of spaces.

ghost: ok, will fix that in 5 min

tmszdmsk (approved):

PR #15897, opened 2018-08-14
PLATFORM-3615: Make the sitemap available at the previous domain

themech:


extensions/wikia/SitemapXml/SitemapHooks.class.php
```diff
@@ -6,10 +6,72 @@ class SitemapHooks {
...
+		$currentHost = parse_url( WikiFactoryLoader::getCurrentRequestUri( $_SERVER, true, true ), PHP_URL_HOST );
+
+		// We only want to deal with plain HTTP versions of the wikia.com domain
+		if ( !self::shouldCancelRedirect( $currentHost, $title ) ) {
+			return true;
+		}
+
+		$output->cancelRedirect();
+
+		$request->response()->header( 'X-sitemaps-redirect-cancelled: 1' );
```
rogatty: I think the convention is to use capitalized header names, i.e. `X-Sitemaps-Redirect-Cancelled`

Grunny: :+1: Done.

extensions/wikia/SitemapXml/SitemapHooks.class.php
```diff
@@ -6,10 +6,72 @@ class SitemapHooks {
...
+		$wgServer = "http://{$currentHost}";
+		// Reset script path and related variables on language path wikis
+		// so the URLs in the sitemap are correct
+		$wgScriptPath = '';
+		$wgScript = '/index.php';
+		$wgArticlePath = '/wiki/$1';
+		return false;
+	}
+
+	private static function shouldCancelRedirect( $currentHost, Title $targetTitle ): bool {
```
rogatty, line 44: This method is hard to follow. What do you think about adding a unit test for it? Would be easier to reason about it while looking at the test cases.

Grunny: Good point, I've added a unit test for the hook method, which will cover the private method logic.

extensions/wikia/SitemapXml/tests/SitemapHooksTest.php
```diff
@@ -0,0 +1,97 @@
...
+				true
+			],
+			[
+				'https://wookieepedia.wikia.com/sitemap-newsitemapxml-index.xml',
+				'https://starwars.fandom.com/sitemap-newsitemapxml-index.xml',
+				true
+			],
+			[
+				'http://ja.starwars.wikia.com/sitemap-newsitemapxml-index.xml',
+				'https://starwars.fandom.com/sitemap-newsitemapxml-index.xml',
```
rogatty, line 72: Are you sure this case is right? `ja.starwars.wikia.com` -> `starwars.fandom.com` without the lang path doesn't really make sense.

rogatty (approved):

extensions/wikia/SitemapXml/tests/SitemapHooksTest.php
Grunny: The point was just to check it won't prevent the redirect when a language subdomain redirects to one without the language path, and I was just using `starwars.fandom.com` for all of them for consistency. This will be the case for some of the exception non-English wikis that may stay at single subdomains, so for example, `de.starwars.wikia.com` is an alias for `jedipedia.wikia.com`. After the switch, unless we decide to move it under `starwars.fandom.com` (these will be reviewed on a case-by-case basis), this will become `jedipedia.fandom.com`. This is a bad example, since the subdomain itself is different, but the test case is just there to make sure a multiple subdomain alias does not match if the language code is not there on the redirect target.

rogatty: Makes sense, thanks for the explanation!

PR #15908, opened 2018-08-17
in Indexer Evaluation service do not process empty documents

sqreek:
in indexer - when all documents are redirects there is no pageIds for loading content.
if there is no pageIds then [cond](https://github.com/Wikia/app/blob/d62acf03a09b9d9441aa95fe1fac660f9a4fde75/extensions/wikia/Search/classes/IndexService/Evaluation.php#L236]) stays empty and make query like `select rev_page, rev_text_id from revision`

@jsutterfield 
@mixoteric


jsutterfield (approved):
LGTM

PR #15923, opened 2018-08-21
Update getDiscussionStats to get threadCount from all forums

jsutterfield:
The `WikiDetailsService.class::getDiscussionStats` is out of date and returning incorrect information. It's returning the thread count from just the "General" category of a discussion, instead of the thread count of all categories. This PR updates that, which updates the value returned by this API: https://www.sandbox-s2.wikia.com/wikia.php?controller=WikisApi&method=getDetails&format=json&ids=558247&mcache=writeonly

This is the old URL which was being used which only returns data for the general forum: http://prod.sjc.k8s.wikia.net/discussion/558247/forums/558247

And the new one which returns data for all forums which we iterate over to return the combined thread count: http://prod.sjc.k8s.wikia.net/discussion/558247/forums

/cc @Wikia/mobile-apps


includes/wikia/services/WikiDetailsService.class.php
```diff
@@ -280,18 +280,29 @@ protected function cacheWikiData( $wikiInfo, $method = null ) {
...
-		$response = Http::get( "http://$discussionsServiceUrl/$id/forums/$id?limit=1", 'default', [ 'noProxy' => true ] );
-		if ( $response !== false ) {
-			$decodedResponse = json_decode( $response, true );
-			if ( isset( $decodedResponse['threadCount'] ) && json_last_error() === JSON_ERROR_NONE ) {
-				return $decodedResponse['threadCount'];
+		$response = Http::get( "http://$discussionsServiceUrl/$id/forums", 'default', [ 'noProxy' => true ] );
+		$threadCount = 0;
+
+		if ($response === false) {
+			return $threadCount;
```
JakubCie: Shouldn't we return null here? The current behavior is that if wiki doesn't have discussions "discussions" field is not present in the response. I'm not sure but if we return 0 here would it mean that "discussions" field will be present for all wikis (even without discussions)?

jsutterfield: That's a good point! I'll update that.

jsutterfield:
@JakubCie should be good now. Discussion count is returned for communties with discussions (558247 -- clashofclans), but not for those without discussions (51 -- doom):
```bash
james-s:app (fix_discussion_stats) jamessutterfield$ curl -s https://www.sandbox-s2.wikia.com/wikia.php\?controller\=WikisApi\&method\=getDetails\&format\=json\&ids\=558247,51\&mcache\=writeonly | jq '.items[]' | jq '{id, stats}'
{
  "id": 558247,
  "stats": {
    "edits": 524941,
    "articles": 866,
    "pages": 351656,
    "users": 15500943,
    "activeUsers": -1,
    "images": 9576,
    "videos": 1299,
    "admins": 22,
    "discussions": 1370
  }
}
{
  "id": 51,
  "stats": {
    "edits": 109648,
    "articles": 3674,
    "pages": 15085,
    "users": 15500943,
    "activeUsers": 50,
    "images": 5238,
    "videos": 74,
    "admins": 17
  }
}
```

hakubo (approved):

PR #15929, opened 2018-08-23
ADEN-7566 Increase maxCpm to $50 for AppNexus and Rubicon videos.

jacekbj:
(!)  No not merge before updating dependencies after merging https://github.com/Wikia/ad-products/pull/146.


nandy-andy (approved):
Once we have AdEngine3 EVERYWHERE such PRs will contain only a version bump of the dependencies ;)

PR #15934, opened 2018-08-24
(PLATFORM-3658) Add support for both domains in some places in AdEngine

Grunny:
Fix some low hanging fruit and add support for both domains in some places
in AdEngine code.

/cc @Wikia/adeng @Wikia/core-platform-team


extensions/wikia/AdEngine/XDM_iframe.html
```diff
@@ -1,13 +0,0 @@
-<html>
-<body>
-</body>
-<script src="http://athena-ads.wikia.com/athena/XDM.js"></script>
```
Grunny, line 4: This doesn't work anymore, so I assumed we can just remove it.

rogatty (approved):

nandy-andy (approved):

PR #15941, opened 2018-08-28
IRIS-6408 | make title clickable on FC wikis

xkxd:
@Wikia/iris 

https://wikia-inc.atlassian.net/browse/IRIS-6408

### DIFF
--- a/includes/wikia/models/DesignSystemCommunityHeaderModel.class.php
+++ b/includes/wikia/models/DesignSystemCommunityHeaderModel.class.php
@@ -17,7 +17,7 @@ class DesignSystemCommunityHeaderModel extends WikiaModel {
 	private $wikiLocalNavigation = null;
 
 	public function __construct( string $langCode ) {
-		global $wgCityId;
+		global $wgCityId, $wgFandomCreatorCommunityId;
 
 		parent::__construct();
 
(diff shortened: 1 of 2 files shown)


includes/wikia/models/DesignSystemCommunityHeaderModel.class.php
```diff
@@ -17,15 +17,18 @@ class DesignSystemCommunityHeaderModel extends WikiaModel {
...
+		global $wgCityId, $wgFandomCreatorCommunityId;
 
 		parent::__construct();
 
 		$this->productInstanceId = $wgCityId;
 		$this->langCode = $langCode;
 		$this->themeSettings = new ThemeSettings( $wgCityId );
 		$this->settings = $this->themeSettings->getSettings();
-		$this->mainPageUrl = wfProtocolUrlToRelative( Title::newMainPage()->getFullURL() );
+		$this->mainPageUrl = empty( $wgFandomCreatorCommunityId )
```
rybmat, line 28: @xkxd IMO we should move overriding of mainPageUrl to the hook where previously mainpage url was set to /

rybmat (changes requested):

includes/wikia/models/DesignSystemCommunityHeaderModel.class.php
xkxd: mainPageUrl is used in `workmark` as well

rybmat (approved):

PR #15942, opened 2018-08-28
ADEN-7598 Restrict queen_of_hearts to FV pages

fraszczakszymon:


extensions/wikia/AdEngine/js/ml/billTheLizard.js
```diff
@@ -17,11 +17,15 @@ define('ext.wikia.adEngine.ml.billTheLizard', [
...
 
 	function isApplicable(name) {
-		if (name.indexOf('ctp_desktop') === 0) {
-			return adContext.get('targeting.hasFeaturedVideo');
-		}
+		var parts = name.split(':');
 
-		return true;
+		switch (parts[0]) {
+			case 'ctp_desktop':
```
jacekbj, line 23: I think we should start placing literals (like "ctp_desktop" and "queen_of_hearts") in a separate module or just make them exportable `const` in this module.
For now it's OK for now as these strings appear only once, but later the risk of typo/forgetting about one increases.

jacekbj (approved):

PR #15948, opened 2018-08-29
ADEN-7585 Add ctp/audio to FV events

dmnsobczak:


extensions/wikia/AdEngine/js/video/player/jwplayer/jwplayerTracker.js
```diff
@@ -31,14 +35,25 @@ define('ext.wikia.adEngine.video.player.jwplayer.jwplayerTracker', [
 			return;
 		}
 
-		params.withAudio = !player.getConfig().mute;
-
 		Object.keys(trackingEventsMap).forEach(function (playerEvent) {
 			player.on(playerEvent, function(event) {
-				var errorCode;
+				var errorCode,
+					vastParams = event.tag ? vastParser.parse(event.tag) : null;
```
fraszczakszymon: As far as I understand we need to have those flags on init/load events. So we can't base on vast request beacuse it happens to late. Can't we use `params.withAudio` for that?

Also we need to make sure that value is updated

dmnsobczak: You right. Added some code, now status is properly set on `ready` event (`init` is too soon).

extensions/wikia/AdEngine/js/video/player/jwplayer/jwplayerTracker.js
```diff
@@ -31,14 +35,34 @@ define('ext.wikia.adEngine.video.player.jwplayer.jwplayerTracker', [
...
+					if (
+						vastParams.customParams.ctp !== undefined
+						&& params.withCtp !== (vastParams.customParams.ctp === 'yes')
+					) {
+						params.withCtp = vastParams.customParams.ctp === 'yes';
+					}
+
+					if (
+						vastParams.customParams.audio !== undefined
+						&& params.withAudio !== (vastParams.customParams.audio === 'yes')
```
nandy-andy: What do you think of extracting this condition to a method with a descriptive name? It may save some time in the feature when we'll get back to this code ;) What the name would be? `customParamDefinedAndFlagNotSet()`? :)

extensions/wikia/AdEngine/js/video/player/jwplayer/jwplayerTracker.js
```diff
@@ -1,7 +1,9 @@
 /*global define*/
 define('ext.wikia.adEngine.video.player.jwplayer.jwplayerTracker', [
-	'ext.wikia.adEngine.video.player.playerTracker'
-], function (playerTracker) {
+	'ext.wikia.adEngine.video.player.playerTracker',
```
fraszczakszymon, line 3: Remove comma please

extensions/wikia/AdEngine/js/video/player/jwplayer/jwplayerTracker.js
```diff
@@ -31,14 +33,35 @@ define('ext.wikia.adEngine.video.player.jwplayer.jwplayerTracker', [
...
+
+		updateCtpAudio();
 
 		Object.keys(trackingEventsMap).forEach(function (playerEvent) {
 			player.on(playerEvent, function(event) {
 				var errorCode;
-				if (playerEvent === 'adError') {
-					errorCode = event && event.code;
+
+				if (['adRequest', 'adError', 'ready', 'videoStart'].indexOf(playerEvent) !== -1) {
```
fraszczakszymon, line 47: Why do we have 'adRequest', 'adError' in this array if we skip updating?

fraszczakszymon: Discussed offline 👍

extensions/wikia/AdEngine/js/video/player/jwplayer/jwplayerTracker.js
dmnsobczak: Removed 👍

fraszczakszymon (approved):

PR #15965, opened 2018-08-31
ADEN-7549 Disable incontent player for desktop when wgDisableInconten…

jacekbj:
…tPlayer is true.


extensions/wikia/AdEngine/js/run/desktop.run.js
```diff
@@ -142,14 +144,25 @@ require([
...
+	tracker,
 	trackingOptIn,
 	win
 ) {
 	'use strict';
 
 	function initDesktopSlots() {
 		highImpact.init();
-		inContent.init('INCONTENT_PLAYER');
+		if (instantGlobals.wgDisableIncontentPlayer) {
```
fraszczakszymon: This is not an instant global. We need to pass this variable in the `AdEngine2ContextService` and read value from adContext

fraszczakszymon (approved):

fraszczakszymon (approved):

PR #15977, opened 2018-09-03
ADEN-7619 Adding BOTTOM_LEADERBOARD to Blockthrough recovery

dmnsobczak:


extensions/wikia/AdEngine/js/wad/btRecLoader.js
```diff
@@ -14,37 +14,53 @@ define('ext.wikia.adEngine.wad.btRecLoader', [
...
+			},
+			BOTTOM_LEADERBOARD: {
+				uid: '5b8f13805d-188',
+				style: 'margin-bottom:23px; z-index:100;',
+				lazy: true
 			}
 		};
 
-	function markAdSlots(replace) {
+	function markAdSlots(replace, lazySlotName) {
```
nandy-andy: If I understand correctly this method marks ad slots, replaces the one to be recovered and executes vendor's function depending on the 2nd parameter passed to it. It's quite a lot of duties for such a small method ;) What do you think of splitting the responsibilities a little bit? :)

extensions/wikia/AdEngine/js/wad/btRecLoader.js
```diff
@@ -14,37 +14,53 @@ define('ext.wikia.adEngine.wad.btRecLoader', [
 		placementsMap = {
 			TOP_LEADERBOARD: {
 				uid: '5b33d3584c-188',
-				style: 'margin: 10px 0;'
+				style: 'margin:10px 0; z-index:100;',
```
nandy-andy, line 17: 100 should be good enough and it's in the WDS scale: https://github.com/Wikia/app/blob/dev/extensions/wikia/DesignSystem/node_modules/design-system/dist/scss/wds-variables/_z-index-scale.scss#L11

However, I skimmed through our ads code and we use some outdated vars: https://github.com/Wikia/app/blob/1e4ea22073c29ec97beeccda86d068915366d0c5/skins/oasis/css/core/ads-porvata.scss#L4

Anyway, I don't think there is a way to re-use SASS vars in JS, so I think we need to leave it as-is here. Unless there is a way to re-use it...

dmnsobczak: `100` is the exact same value like for `TOP_RIGHT_BOXAD`. And I also don't know how to connect these places with additionally keeping in mind that we need to be compatible with Blockthrough rules and not to make AdBlock developers job easier :(

extensions/wikia/AdEngine/js/wad/btRecLoader.js
dmnsobczak: That's a good observation. I refactored the code a little bit.

extensions/wikia/AdEngine/js/wad/btRecLoader.js
```diff
@@ -14,40 +14,60 @@ define('ext.wikia.adEngine.wad.btRecLoader', [
...
+			},
+			BOTTOM_LEADERBOARD: {
+				uid: '5b8f13805d-188',
+				style: 'margin-bottom:23px; z-index:100;',
+				lazy: true
 			}
 		};
 
-	function markAdSlots(replace) {
+	function markAdSlots(replace, lazySlotName) {
```
fraszczakszymon: It seems that method isn't called with `replace=true` at all. Why do we need this flag?

extensions/wikia/AdEngine/js/wad/btRecLoader.js
```diff
@@ -14,40 +14,60 @@ define('ext.wikia.adEngine.wad.btRecLoader', [
...
 		Object
 			.keys(placementsMap)
 			.forEach(function (key) {
```
fraszczakszymon: We don't need to iterate when lazySlotName is defined. Am I right?

extensions/wikia/AdEngine/js/wad/btRecLoader.js
```diff
@@ -56,13 +76,17 @@ define('ext.wikia.adEngine.wad.btRecLoader', [
 		});
 	}
 
-	function init() {
-		markAdSlots(false);
+	function triggerScript() {
+		win.BT.clearThrough();
```
fraszczakszymon: Is there any chance that it won't be defined?

extensions/wikia/AdEngine/js/wad/btRecLoader.js
fraszczakszymon: What do you think about extracting separate method for `markAdSlot(slotName)` and reuse it in `markAdSlots`

extensions/wikia/AdEngine/js/wad/btRecLoader.js
dmnsobczak: Actually we are not. I left it here since first BT Epic when I was using this at some point of development. Removed.

extensions/wikia/AdEngine/js/wad/btRecLoader.js
dmnsobczak: Good suggestions. Code refactored.

extensions/wikia/AdEngine/js/wad/btRecLoader.js
dmnsobczak: Oh, there is. Check added.

fraszczakszymon (approved):

PR #15982, opened 2018-09-04
ADEN-7421 Control FV autoplay via bill-the-lizard and remove frontend ml models

fraszczakszymon:


extensions/wikia/AnalyticsEngine/AnalyticsProviderBillTheLizard.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+class AnalyticsProviderBillTheLizard implements iAnalyticsProvider {
+	private static $template = 'extensions/wikia/AnalyticsEngine/templates/bill-the-lizard.mustache';
+
+	public static function isEnabled() {
+		global $wgShowAds;
+
+		return $wgShowAds && AdEngine2Service::areAdsShowableOnPage();
```
nandy-andy, line 9: What do you think of including here `$noExternals`?

nandy-andy (approved):

extensions/wikia/AnalyticsEngine/AnalyticsProviderBillTheLizard.php
fraszczakszymon: It should work with $noExternals too :)
https://github.com/Wikia/app/blob/dev/extensions/wikia/AdEngine/AdEngine2PageTypeService.class.php#L42

PR #15993, opened 2018-09-06
ADEN-7630 Collapse Porvata on mobile if outstream is disabled

fraszczakszymon:


extensions/wikia/AdEngine/js/template/porvata.js
```diff
@@ -200,6 +209,11 @@ define('ext.wikia.adEngine.template.porvata', [
 			return;
 		}
 
+		if (adContext.get('opts.isIncontentPlayerDisabled')) {
+			callCollapse(params, 'disabled');
```
nandy-andy, line 234: Please add here sending a GA event here, the same as we do it in [Oasis](https://github.com/Wikia/app/pull/15965/files#diff-2bbd7807397898087e6ae3886e0075feR156) and [mobile-wiki](https://github.com/Wikia/mobile-wiki/pull/997/files#diff-980fed8db3ff3ff13dcf8d5415610016R96).

nandy-andy: Although, we probably send some kind of an event to our ads profile in GA thanks to: https://github.com/Wikia/app/pull/15993/files#diff-19a602d05fba021e166521ec878dc7d8R48, dodn't we? :)

fraszczakszymon: Solved :+1: I added tracking

nandy-andy (approved):

nandy-andy (approved):

PR #16000, opened 2018-09-07
SUS-5786 | remove hostname entry from wgTransactionContext

macbre:
https://wikia-inc.atlassian.net/browse/SUS-5786

This generates misleading differences between response from Apaches and Kubernetes cluster. And is useless for responses cached on CDN level.

## Variable after this change

```js
wgTransactionContext={"type":"page/file","env":"dev","php_version":"7.0.27","wiki":"plpoznan","entry_point":"page","namespace":6,"logged_in":false,"action":"view","skin":"oasis","parser_cache_used":true,"size_category":"simple"};
```

### DIFF
--- a/includes/wikia/transaction/Transaction.php
+++ b/includes/wikia/transaction/Transaction.php
@@ -19,7 +19,6 @@ class Transaction {
 
 	// Parameters
 	const PARAM_ENVIRONMENT = 'env';
-	const PARAM_HOSTNAME = 'hostname';
 	const PARAM_PHP_VERSION = 'php_version';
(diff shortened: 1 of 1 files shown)


pwojcik86 (approved):

ghost (approved):

ghost:
In addition to that, whenever we talk about some sort of transaction between the client and "us" infrastructure details such as datacenters, hostnames. and whatnot should remain irrelevant and a such should never be exposed to the client.  They should only serve debugging purposes in cases of cache pollution, etc. and be tracked on a different level, e.g. HTTP headers.

PR #16008, opened 2018-09-10
SUS-5789 add rewrite for lyrics api

pwojcik86:
https://wikia-inc.atlassian.net/browse/SUS-5789

Fixes an issue where lyricswiki api requests don't contain the `action` parameter.

## Testing

Check the url in the ticket. It should return the same response from apaches and k8s. Use a sandbox.

### DIFF
--- a/docker/dev/site.conf
+++ b/docker/dev/site.conf
@@ -34,6 +34,11 @@ server {
     rewrite ^/(__cb\d+/)?(skins|resources|extensions)/(.+) /$2/$3 break;
     rewrite ^/favicon.ico /skins/common/images/favicon.ico break;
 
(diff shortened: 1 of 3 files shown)


macbre (approved):

docker/dev/site.conf
```diff
@@ -34,6 +34,11 @@ server {
     rewrite ^/(__cb\d+/)?(skins|resources|extensions)/(.+) /$2/$3 break;
     rewrite ^/favicon.ico /skins/common/images/favicon.ico break;
 
+    # SUS-5789 force pre-existing lyricwiki API calls to use a MediaWiki-compliant query
+    if ($arg_func) {
```
macbre: Let's just document the `if` statement that it checks if there's `func` parameter in the URL

macbre (approved):

PR #16020, opened 2018-09-11
XF-635 | remove unnecessary image-related data from mobile html

rybmat:
https://wikia-inc.atlassian.net/browse/XF-635

@Wikia/x-wing

### DIFF
--- a/extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
+++ b/extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
@@ -6,7 +6,7 @@ class ArticleAsJson {
 		'imageMaxWidth' => false
 	];
 
-	const CACHE_VERSION = 3.38;
+	const CACHE_VERSION = 3.40;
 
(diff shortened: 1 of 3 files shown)


extensions/wikia/ArticleAsJson/ArticleAsJson.class.php
```diff
@@ -82,6 +82,57 @@ private static function renderGallery( $media ) {
 		);
 	}
 
+	public static function getDataAttrsForImage( $media ) {
+		$dataAttrs = [
```
hakubo, line 86: that actually should be an object right? :)

hakubo: or use that method in a method for getting data attrs for gallery to avoid repetion

rybmat: done

hakubo (approved):

PR #16030, opened 2018-09-12
PLATFORM-3677 call cookie syncer once

harnash:
Do not call Cookie-Sync service when sync is done


skins/oasis/modules/templates/Oasis_Index.php
```diff
@@ -95,9 +95,7 @@
 	<!-- Combined JS files and head scripts -->
 	<?= $jsFiles ?>
 <? endif ?>
-<? if ( isset($_COOKIE['tracking-opt-in-status']) ): ?>
-	<?= F::app()->renderView( 'CookieSyncerService', 'Index' ) ?>
-<? endif ?>
+<?= F::app()->renderView( 'CookieSyncerService', 'Index' ) ?>
```
rogatty: I believe that only essential scripts should be inlined in HTML. How about using AssetsManager to load this script asynchronously?
https://github.com/Wikia/app/blob/dev/extensions/wikia/AssetsManager/config.php

extensions/wikia/CookieSyncer/templates/CookieSyncer_Index.php
```diff
@@ -1 +1,9 @@
-<iframe src="<?= $url; ?>" class="cookie-syncer-module-iframe"></iframe>
+<script type="text/javascript">
+	if (window.document.cookie.indexOf('cookiesync_done') === -1 && window.document.cookie.indexOf('tracking-opt-in-status') !== -1) {
```
rogatty: If we use AssetsManager to load this script, then we can use https://github.com/Wikia/app/blob/dev/resources/wikia/modules/cookies.js#L58 instead of `indexOf` which might break in some edge cases.

extensions/wikia/CookieSyncer/templates/CookieSyncer_Index.php
```diff
@@ -1 +1,9 @@
...
+		var iframe = document.createElement('iframe');
+		iframe.style.display = "none";
+		iframe.src = "<?= $url; ?>";
+		iframe.classList.add("cookie-syncer-module-iframe");
```
rogatty: We can skip either `iframe.style.display = "none";` or setting this class.

skins/oasis/modules/templates/Oasis_Index.php
harnash: @rogatty fix in 9d1fe3b97cf

extensions/wikia/CookieSyncer/templates/CookieSyncer_Index.php
harnash: fix in 44a754fb8d8

extensions/wikia/CookieSyncer/templates/CookieSyncer_Index.php
harnash: @rogatty fix in 5259cdd8dc2

extensions/wikia/AssetsManager/config.php
```diff
@@ -2492,3 +2492,11 @@
 		'//extensions/wikia/FeedsAndPosts/js/feedsAndPosts.js',
 	]
 ];
+
+$config['cookiesyncer_js'] = [
```
rogatty: Is it concatenated with other packages or downloaded separately? If the latter, then I'd add the script to `oasis_shared_core_js` or something like that.

harnash: @rogatty fix in bebadc82140

extensions/wikia/AssetsManager/config.php
```diff
@@ -13,6 +13,7 @@
 		'//resources/mediawiki/mediawiki.Uri.js',
 		'#group_banner_notifications_js',
 		'#group_ui_repo_api_js',
+		'//extensions/wikia/CookieSyncer/js/cookiesyncer.js',
```
rogatty, line 16: @mszabo-wikia is there any reason not to do that?

rogatty (approved):

extensions/wikia/AssetsManager/config.php
mszabo-wikia: I think it should be ok, maybe `oasis_noads_extensions_js` is a better place

rogatty:
Closed in favour of https://github.com/Wikia/app/pull/16063

PR #16048, opened 2018-09-17
SRE | Active active phase two merge

mszabo-wikia:
* Simplify chat configuration to accommodate RES sandboxes by getting rid of `$wgChatPublicHostOverride` and setting proper `$wgChatPublicHost` instead in the appropriate files.
* Ensure we use SJC DC when loading starter dumps during CNW process


includes/wikia/Extensions.php
```diff
@@ -3,16 +3,6 @@
...
 ###############################################################################
 
-###############################################################################
-# DC specific settings                                                        #
-###############################################################################
-switch ($wgWikiaDatacenter) {
-	case "res":
-		# PLATFORM-1740: disable task queue in Reston, it was calling SJC broker
-		$wgTaskBroker = false;
-		break;
```
mszabo-wikia, line 13: This override is already in config, so can be removed from here.

PR #16049, opened 2018-09-17
XF-635 | separate endpoints for wiki variables used by mobile-wiki, discussions and announcements

rybmat:
https://wikia-inc.atlassian.net/browse/XF-635

@Wikia/iwing


extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -180,6 +179,48 @@ private function prepareWikiVariables() {
 		return $wikiVariables;
 	}
 
+	public function getMobileWikiVariables() {
+		( new CrossOriginResourceSharingHeaderHelper() )->allowWhitelistedOrigins()
+			->setAllowMethod( [ 'GET' ] )
+			->setHeaders( $this->response );
```
mszabo-wikia, line 80: is it needed here? If it's internal traffic only I think we can remove it

rybmat: @mszabo-wikia requests to these endpoints come also from front-end apps

rogatty: @rybmat which ones? I thought all Ember clients use a server-side request.

rybmat: @rogatty I'll double check that, thanks!

PR #16050, opened 2018-09-17
SUS-5823 | make /api/v1/Docs work

macbre:
https://wikia-inc.atlassian.net/browse/SUS-5823

There were a few issues here:

* `WikiaRequest::getScriptUrl` was reading `$_SERVER['SCRIPT_URL']` which is not set when PHP runs in fpm mode with nginx, use `REQUEST_URI` entry which is set by both Apache and nginx
* there were two notices thrown by Swagger-related code breaking JSON response
* enable debug logging of nginx rewrites on dev


ghost (approved):

ghost:
@macbre - does it make sense to change `SCRIPT_URL` to `REQUEST_URI` across the codebase, in all the following files?

```
michal@dev-michal:/usr/wikia/source/wiki$ grep -r SCRIPT_URL *
includes/wikia/nirvana/WikiaRequest.class.php:	 * Get data from $_SERVER['SCRIPT_URL'], which is original path of the request, before mod_rewrite changed it.
includes/wikia/nirvana/WikiaRequest.class.php:		$scriptUrl = isset( $_SERVER['SCRIPT_URL'] ) ? $_SERVER['SCRIPT_URL'] : null;
includes/wikia/Wikia.php:		if (isset($_SERVER['REQUEST_METHOD']) && ($_SERVER['REQUEST_METHOD'] === 'GET') && ($_SERVER['SCRIPT_URL'] === '/wikia.php')) {
includes/WebStart.php:		( strpos(strtolower($_SERVER['SCRIPT_URL']), 'datacenter') === false ) &&
includes/WebStart.php:		!in_array( strtolower( $_SERVER['SCRIPT_URL'] ), [ 'api.php', 'wikia.php' ] )
includes/GlobalFunctions.php:		# If in this mode, use SCRIPT_URL instead, which mod_rewrite
```

macbre:
@mixoteric - I've replaced `SCRIPT_URL` check in `Wikia::logBacktrace` with a call to `WikiaLogger`.

As for the rest:

* `SCRIPT_NAME` has a different meaning than `REQUEST_URI` - for `/wikia.php?foo=bar` URL it will be set to `/wikia.php` (the latter one will obviously have the entire URL). Let's keep it as it is in `WebStart.php`. `$_SERVER['SCRIPT_NAME']` should be used instead - it is set when run as PHP module inside Apache or fpm process via nginx.
* `WikiaRequest` class was modified originally by this pull request

## Example

```
  'REQUEST_URI' => string '/wikia.php?foo' (length=14)
  'SCRIPT_NAME' => string '/wikia.php' (length=10)
```

Apache also sets:

```
  'SCRIPT_URL' => string '/wikia.php' (length=10)
```

ghost (approved):

PR #16051, opened 2018-09-17
SUS-5826 | prevent concurent cronjobs

xkxd:
https://wikia-inc.atlassian.net/browse/SUS-5826


maintenance/utils/Status.php
```diff
@@ -0,0 +1,42 @@
...
+class MaintenanceStatus {
+	private $cacheKey = null;
+
+	private $STARTED = 'STARTED';
+	private $ABORTED = 'ABORTED';
+	private $FINISHED = 'FINISHED';
+	private $TTL = 7200; // 2h in seconds
+
+	public function __construct($taskName) {
+		$this->cacheKey = $taskName . '_status';
```
mszabo-wikia: This should probably be sharded by wiki ID. If you wrap this in `wfMemcKey` it will do it for you.

Actually that depends on whether the task is per-wiki or global. So this warrants some further discussion.

mszabo-wikia (changes requested):

maintenance/Maintenance.php
```diff
@@ -886,6 +888,10 @@ public function finalSetup() {
 		# Same with these
 		$wgCommandLineMode = true;
 
+		# MaintenanceStatus comes from a separate file (listed in AutoLoader.php) and we need to set it
+		# here as `finalSetup` is called after AutoLoader is included.
+		$this->status = new MaintenanceStatus($this->getName());
```
mszabo-wikia: this code should be spacey, you can get an authoritative IntelliJ code style XML from https://github.com/Wikia/guidelines/blob/master/PHP/bin/MediaWiki.xml

maintenance/doMaintenance.php
```diff
@@ -100,6 +100,16 @@
 	require( MWInit::interpretedPath( 'AdminSettings.php' ) );
 }
 $maintenance->finalSetup();
+
+// check if specific task is already running (in case k8s will accidentally create two instances of cron job)
+if ($maintenance->status->isRunning()) {
```
mszabo-wikia: I wonder if we should do this for every execution of maintenance scripts (since it'll include also manually started ones). Maybe we should check an env var like `IS_MAINTENANCE_JOB` which would be set only inside cronjob containers, so that manual runs are not affected?

macbre: I think we only need to make this work for just a few maintenance scripts. What if we just extend `Maintenance` class to handle locking mechanism and explicitly use it in specific cronjob scripts?

xkxd: the purpose of doing this in `Maintenance` was to make it as generic as possible, since k8s cronjobs can be created more than once no matter what job is running. also having any differences in `$maintenance` would require to add `if` check for every `$status` invocation. 

on the other hand, if we'll agree on extending `Maintenance` class, which scripts should I extend?

mszabo-wikia (approved):

docker/maintenance/cronjobs-generator.sh
```diff
@@ -5,7 +5,7 @@
 ## e.g. ./cronjobs-generator.sh 05c4221.5c2dd23 | kubectl --context context -n environment apply -f -
 
 SCRIPT_FOLDER=$(dirname $(readlink -f $0))
-JOB_DESCRIPTIONS=(`ls $SCRIPT_FOLDER | grep -v 'example' | grep -v 'cronjob-template.yaml' | grep '\.yaml$'`)
+JOB_DESCRIPTIONS=(`ls $SCRIPT_FOLDER | grep -v 'cronjob-template.yaml' | grep '\.yaml$'`)
```
ghost: `ls` can do filtering, too. `ls $SCRIPT_FOLDER/*yaml | grep -v 'cronjob-template'` is shorter. :)

includes/AutoLoader.php
```diff
@@ -939,6 +939,9 @@
 	# maintenance/term
 	'AnsiTermColorer'  => 'maintenance/term/MWTerm.php',
 	'DummyTermColorer' => 'maintenance/term/MWTerm.php',
+
+	# maintenance/utils
+	'MaintenanceStatus' => 'maintenance/utils/Status.php',
```
ghost: Autoloader uses `$wgAutoloadClasses` array. Please, do not modify core MediaWiki's file, but rather add your entry to `includes/wikia/DefaultSettings.php`.

maintenance/utils/Status.php
```diff
@@ -0,0 +1,42 @@
...
+	const ABORTED = 'ABORTED';
+	const FINISHED = 'FINISHED';
+	const TTL = 7200; // 2h in seconds
+
+	public function __construct( $taskName ) {
+		$this->cacheKey = wfMemcKey( $taskName . '_status' );
+	}
+
+	public function markAsStarted() {
+		$this->setCachedStatus( MaintenanceStatus::STARTED );
```
ghost: Consider using `self::STARTED` here instead of explicit class name. It makes refactors easier and `grep` output more meaningful while searching for class name. Also, it is a nice analogy to human comprehension of possession, as we refer to things we own as "ours", and we only use explicit names while referring to someone else's property. So `self` indicates that the constant belongs to the current class more clearly that the class' name. :)

maintenance/doMaintenance.php
```diff
@@ -100,6 +100,16 @@
 	require( MWInit::interpretedPath( 'AdminSettings.php' ) );
 }
 $maintenance->finalSetup();
+
+// check if specific task is already running (in case k8s will accidentally create two instances of cron job)
+if ( $maintenance->status->isRunning() ) {
```
ghost: I don't know what's the risk of failure if we deployed this change, but it clearly breaks backward compatibility.

We completely disallow multiple executions of all scripts. Perhaps we should allow multiple executions by default (backwards compatibility), and explicitly state that some particular script does now allow that (lets us gradually apply that to scripts we are familiar with).

Example: `if ( $maintenance->status->isRunning() && $maintenance->noConcurrency ) { ...`

ghost (changes requested):

ghost (approved):

docs/maintenance.txt
```diff
@@ -55,3 +55,9 @@ That's it. In the execute() method, you have access to all of the normal
 MediaWiki functions, so you can get a DB connection, use the cache, etc.
 For full docs on the Maintenance class, see the auto-generated docs at
 http://svn.wikimedia.org/doc/classMaintenance.html
+
+Note:
```
ghost: I'd document an example implementation here. We may retire FounderEmails at some point or refactor it which will destroy the example. So it's good to have it here. It'll also be simpler to grasp by a person willing to create a cronjob.

ghost (changes requested):

ghost (approved):

ghost (approved):

PR #16054, opened 2018-09-17
PLATFORM-3674 | window.wgTrackID isn't always available so use inlined value instead

rogatty:
https://wikia-inc.atlassian.net/browse/PLATFORM-3674

`track.mustache` is rendered at the top of the `body` while `wgTrackID` is set at the bottom. The easiest fix is to pass the user ID to the template instead of relying on a global variable.

@Wikia/core-platform-team @mszabo-wikia

### DIFF
--- a/extensions/wikia/Track/templates/track.mustache
+++ b/extensions/wikia/Track/templates/track.mustache
@@ -50,7 +50,8 @@
 					'&pv_number_global=' + window.pvNumberGlobal;
 
 			if (optIn) {
-				trackUrl += '&u=' + window.wgTrackID;
+                {{!-- window.wgTrackID isn't available yet as this script is inlined at the top --}}
+				trackUrl += '&u=' + '{{trackID}}';
 			} else {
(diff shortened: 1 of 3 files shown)


extensions/wikia/Track/templates/track.mustache
```diff
@@ -50,7 +50,8 @@
 					'&pv_number_global=' + window.pvNumberGlobal;
 
 			if (optIn) {
-				trackUrl += '&u=' + window.wgTrackID;
+                {{!-- window.wgTrackID isn't available yet as this script is inlined at the top --}}
```
mszabo-wikia, line 53: is `{{` a typo?

harnash (approved):

extensions/wikia/Track/templates/track.mustache
rogatty: Nope, it's a mustache comment syntax, see https://scalate.github.io/scalate/documentation/mustache.html#Comments

PR #16055, opened 2018-09-17
SUS-5825 | make nginx and Apache works similar when handling request paths

macbre:
https://wikia-inc.atlassian.net/browse/SUS-5825

Otherwise nginx handles `/wiki/Special:RecentChanges?feed=rss&title=author` incorrectly, using `author` as a title instead of `Special:RecentChanges`.

Full path parsing to extract the title for MediaWiki logic is controlled by `wgUsePathInfo` global variable. Change the logic behind it to treat nginx+php and Apache+mod_php equally.

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -138,9 +138,7 @@ $wgServer = WebRequest::detectServer();
  * redirect loops when "pretty URLs" are used.
  * @var bool $wgUsePathInfo
  */
-$wgUsePathInfo = ( strpos(php_sapi_name(), 'cgi') === false ) &&
-        ( strpos(php_sapi_name(), 'apache2filter') === false ) &&
-        ( strpos(php_sapi_name(), 'isapi') === false );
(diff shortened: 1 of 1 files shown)


LocalSettings.php
```diff
@@ -138,9 +138,7 @@
  * redirect loops when "pretty URLs" are used.
  * @var bool $wgUsePathInfo
  */
-$wgUsePathInfo = ( strpos(php_sapi_name(), 'cgi') === false ) &&
-        ( strpos(php_sapi_name(), 'apache2filter') === false ) &&
-        ( strpos(php_sapi_name(), 'isapi') === false );
+$wgUsePathInfo = ( php_sapi_name() == 'apache2handler' ) || ( php_sapi_name() == 'fpm-fcgi' ); # Wikia change - SUS-5825
```
pwojcik86, line 141: How does `apache2handler` replace the previous values?

macbre: `apache2handler` is the sapi used when PHP is run as a Apache module.

ghost: You don't have to mark it as Wikia change, since LocalSettings is supposed to be the custom file.

ghost (approved):

PR #16056, opened 2018-09-17
PLATFORM-3669: use proper hostname for sign-in/logout links

harnash:
Should fix sign-in and logout links to be on the same domain as the source wiki.

Ticket: https://wikia-inc.atlassian.net/browse/PLATFORM-3669

/cc: @Wikia/core-platform-team


includes/wikia/models/DesignSystemGlobalNavigationModel.class.php
```diff
@@ -95,11 +95,42 @@ public function getData() {
...
+			global $wgServer;
+			$cityUrl = $wgServer;
+		} else {
+			$cityUrl = WikiFactory::cityIDtoUrl( $cityUrl );
+			// Should probably check for HTTPS in some cases (for now all processed links should be HTTPS or relative
+		}
+
+		$urlHost = parse_url( $url, PHP_URL_HOST );
+		$cityUrlHost = parse_url( $cityUrl, PHP_URL_HOST );
+		print_r("u: $urlHost, c: $cityUrlHost\n");
```
rogatty: :see_no_evil:

includes/wikia/models/DesignSystemGlobalNavigationModel.class.php
```diff
@@ -95,11 +95,42 @@ public function getData() {
 		return $data;
 	}
 
-	private function getHref( $hrefKey, $protocolRelative = false ) {
+	private function fixHostForUrl( $url, $cityId = null ) {
```
rogatty: Let's use a meaningful name, like `updateUrlHostToWikiPrimaryDomain()`

includes/wikia/models/DesignSystemGlobalNavigationModel.class.php
harnash: uupps - too much sandbox debugging :(

includes/wikia/models/DesignSystemGlobalNavigationModel.class.php
```diff
@@ -95,11 +95,42 @@ public function getData() {
...
+
+		return $finalUrl;
+	}
+
+	private function getHref( $hrefKey, $protocolRelative = false, $fixHost = false ) {
 		$url = DesignSystemSharedLinks::getInstance()->getHref( $hrefKey, $this->lang );
 		if ( $protocolRelative ) {
 			$url = wfProtocolUrlToRelative( $url );
 		}
+		if ( $fixHost ) {
```
rogatty: `$useWikiPrimaryDomain`?

includes/wikia/models/DesignSystemGlobalNavigationModel.class.php
harnash: begone! 344a0a842f7

includes/wikia/models/DesignSystemGlobalNavigationModel.class.php
harnash: @rogatty `usePrimaryDomainInUrl` ?

includes/wikia/models/DesignSystemGlobalNavigationModel.class.php
harnash: fix in 96b4330925f

rogatty:
I'm not sure if I like this approach of parsing URLs and replacing the host. It would be cleaner if we explicitly used a method when defining the URLs in https://github.com/Wikia/app/blob/dev/includes/wikia/models/DesignSystemSharedLinks.class.php
Also, I think we're not doing the thing described in the ticket. Sign in link shouldn't point to `fandom-domain-wiki.fandom.com/signin` but to `www.fandom.com/signin`

harnash:
> I'm not sure if I like this approach of parsing URLs and replacing the host. It would be cleaner if we explicitly used a method when defining the URLs in https://github.com/Wikia/app/blob/dev/includes/wikia/models/DesignSystemSharedLinks.class.php

I was thinking about similar approach but decided to go with safer variant that only replaces links in places we are sure and not globally (similar to relative parameter in `getHref()`). Doing it as you suggests may have side effects we are unaware of at the moment. 

> Also, I think we're not doing the thing described in the ticket. Sign in link shouldn't point to `fandom-domain-wiki.fandom.com/signin` but to `www.fandom.com/signin`

I'll fix that.

includes/wikia/models/DesignSystemGlobalNavigationModelV2.class.php
```diff
@@ -92,11 +92,64 @@ private function getCreateWiki( string $trackingLabel ) {
...
+	protected function getDevDomain() {
+		global $wgDevDomain, $wgDevelEnvironment;
+		if ( !$wgDevelEnvironment || empty( $wgDevDomain ) ) {
+			return false;
+		}
+
+		return $wgDevDomain;
+	}
+
+	public function usePrimaryDomainInUrl( $url, $cityId = null ) {
```
rogatty: This name isn't really telling what's happening and the method is complex enough to be hard to follow. How about adding PHPDoc with explanation and some examples?

includes/wikia/models/DesignSystemGlobalNavigationModelV2.class.php
```diff
@@ -92,11 +92,64 @@ private function getCreateWiki( string $trackingLabel ) {
 		];
 	}
 
-	private function getHref( $hrefKey, $protocolRelative = false ) {
+	protected function getDomainForCityId( $cityId ) {
+		global $wgServer;
+		if ( $cityId == null ) {
```
rogatty: `===`

includes/wikia/models/DesignSystemGlobalNavigationModelV2.class.php
```diff
@@ -92,11 +92,64 @@ private function getCreateWiki( string $trackingLabel ) {
...
+		}
+
+		return $wgDevDomain;
+	}
+
+	public function usePrimaryDomainInUrl( $url, $cityId = null ) {
+		global $wgWikiaBaseDomainRegex;
+
+		$cityUrl = $this->getDomainForCityId( $cityId );
+		// Should probably check for HTTPS in some cases (for now all processed links should be HTTPS or relative
```
rogatty: Is this comment still valid? If so, I wouldn't know what to do about HTTPS based on it.

includes/wikia/models/DesignSystemGlobalNavigationModelV2.class.php
```diff
@@ -92,11 +92,64 @@ private function getCreateWiki( string $trackingLabel ) {
...
+		$count = 0;
+		$devDomain = $this->getDevDomain();
+
+		if ( $devDomain ) {
+			$finalUrl = preg_replace( '/' . $wgWikiaBaseDomainRegex . '/', $devDomain, $url, 1, $count );
+		} else {
+			$finalUrl = preg_replace( '/' . $urlParts[1] . '/', $cityUrlParts[1], $url, 1, $count );
+		}
+
+		if ( $finalUrl == null || $count !== 1 ) {
```
rogatty, line 136: When would this happen? Would be nice to have a comment explaining it.

harnash: It is highly unlikely to happen since the values used are captured from the regexp few lines before but function can fail and ignoring its results is not good idea.

includes/wikia/models/DesignSystemGlobalNavigationModelV2.class.php
harnash: Fix in 645ef3f0cd1. It is no longer valid since the substitution method changed.

includes/wikia/models/DesignSystemGlobalNavigationModelV2.class.php
harnash: Fix in 645ef3f0cd1

includes/wikia/models/DesignSystemGlobalNavigationModelV2.class.php
harnash: Update in 9062143475d

includes/wikia/models/DesignSystemGlobalNavigationModel.class.php
```diff
@@ -95,11 +95,74 @@ public function getData() {
...
+		}
+
+		if ( $finalUrl == null || $count !== 1 ) {
+			return $url;
+		}
+
+		return $finalUrl;
+	}
+
+	private function getHref( $hrefKey, $protocolRelative = false, $useWikiPrimaryDomain = false ) {
```
Grunny: Maybe `$useWikiBaseDomain` rather than `Primary` for consistency?

includes/wikia/models/DesignSystemGlobalNavigationModel.class.php
```diff
@@ -95,11 +95,74 @@ public function getData() {
 		return $data;
 	}
 
-	private function getHref( $hrefKey, $protocolRelative = false ) {
+	protected function getDomainForCityId( $cityId ) {
+		global $wgServer;
+		if ( $cityId === null ) {
+			return $wgServer;
+		} else {
```
Grunny: The `else` isn't needed since it just contains a return, and removing it would be consistent with the other methods like `getDevDomain`.

includes/wikia/models/DesignSystemGlobalNavigationModel.class.php
```diff
@@ -95,11 +95,74 @@ public function getData() {
 		return $data;
 	}
 
-	private function getHref( $hrefKey, $protocolRelative = false ) {
+	protected function getDomainForCityId( $cityId ) {
```
Grunny: I assume this was changed to `protected` in order to mock it in tests? I don't really like that as a pattern as it usually suggests your tests are too dependent on implementation details. In this case, I'd suggest on'y mocking the external dependencies and setting this and `getDevDomain` back to private since they are not being overridden by subclasses.

If we remove the `$cityId` per another comment, this might be even simpler.

includes/wikia/models/DesignSystemGlobalNavigationModel.class.php
```diff
@@ -95,11 +95,74 @@ public function getData() {
...
+	 * wiki is setup for (i.e. wikia.com or fandom.com). This will allow to have proper domains in the urls when
+	 * link is stored as .wiki.com but the wiki is running on other domain.
+	 *
+	 * @example https://www.wikia.com/signin -> https://www.fandom.com/signin
+	 *
+	 * @param $url string source url to be updated
+	 * @param null $cityId int id of the wiki for the base domain used in the substitution (null will use current wiki)
+	 * @return string final url with base domain substituted (or original url in case of some errors)
+	 */
+	public function useWikiBaseDomainInUrl( $url, $cityId = null ) {
```
Grunny: Is `$cityId` needed? I don't see anything using it, so it seems like it might be better to remove it and add it back if there's ever a need for it.

includes/wikia/models/tests/useWikiBaseDomainInUrlTest.php
```diff
@@ -0,0 +1,78 @@
...
+
+		$result = $model->usePrimaryDomainInUrl( $url, $cityId );
+
+		$this->assertEquals( $expectedResult, $result );
+	}
+
+	public function getUrlProvider() {
+		return [
+			[
+				'https://muppet.wikia.com',						// server
```
Grunny: The whitespace for the comments here is a bit messy. :P

Grunny (changes requested):

includes/wikia/models/DesignSystemGlobalNavigationModel.class.php
harnash: Right. Fix in 8acfe0a5e86

includes/wikia/models/DesignSystemGlobalNavigationModel.class.php
harnash: @grunny By doing this I would need to mock global variables which is really not encouraged either:
* `$wgServer`
* `$wgDevDomain`
* `$wgDevelEnvironment`

Basically I should use `WikiaBaseTest`.

includes/wikia/models/DesignSystemGlobalNavigationModel.class.php
harnash: OK.  I'll remove it

harnash:
@Grunny This commit should address all the comments: ac2610f

Grunny (approved):

PR #16057, opened 2018-09-17
SUS-5819 | separate fastcgi_read_timeout for mediawiki-tasks

pwojcik86:
https://wikia-inc.atlassian.net/browse/SUS-5819

This PR separates `fastcgi_read_timeout` configuration for `mediawiki-prod` and `mediawiki-tasks`. This allows setting a longer timeout (3600s) for tasks.


macbre (approved):

docker/prod/Dockerfile-nginx
```diff
@@ -18,5 +18,8 @@ ADD extensions /usr/wikia/slot1/current/src/extensions
 
 USER nginx
 
-# validate configuration files for syntax errors
-RUN nginx -c /etc/nginx/nginx.conf -t
```
mszabo-wikia, line 22: why was it removed?

pwojcik86: After this change site configuration is available only at runtime, so this would test only `nginx.conf` without includes.

xkxd (approved):

PR #16068, opened 2018-09-20
PLATFORM-3661: handle wikia and fandom domains on dev

themech:
Distinguish between fandom and wikia domains on devboxes (to fix fandom-dev redirects and to make it more prod-like).
After this change, only wikis with `fandom.com` in the `city_url` will be available under the `fandom-dev.(pl|us)` domain. Also redirects between fandom-dev and wikia-dev will work, given a wikis has both domains configured in the `city_domains` table.

### DIFF
--- a/includes/wikia/tests/UrlHelpersTest.php
+++ b/includes/wikia/tests/UrlHelpersTest.php
@@ -1,0 +1,28 @@
+<?php
+
+class UrlHelpersTest extends WikiaBaseTest {
+
+	/**
+	 * @dataProvider normalizeHostDataProvider
+	 */
+	public function testNormalizeHost( $environment, $host, $expected ) {
+		$this->mockEnvironment( $environment );
+		$url = wfNormalizeHost( $host );
+		$this->assertEquals( $expected, $url );
+	}
+
+	public function normalizeHostDataProvider() {
+		return [
+			[
+				'env' => WIKIA_ENV_DEV,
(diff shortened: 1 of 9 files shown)


includes/wikia/tests/UrlHelpersTest.php
```diff
@@ -0,0 +1,28 @@
...
+		return [
+			[
+				'env' => WIKIA_ENV_DEV,
+				'url' => 'http://muppet.' . static::MOCK_DEV_NAME . '.wikia-dev.us',
+				'expected' => 'http://muppet.wikia.com'
+			],
+			[
+				'env' => WIKIA_ENV_DEV,
+				'url' => 'http://muppet.' . static::MOCK_DEV_NAME . '.fandom-dev.us',
+				'expected' => 'http://muppet.fandom.com'
```
rogatty: There is no http without s in fandom.com :lock:

rogatty (approved):

harnash (approved):

includes/wikia/tests/UrlHelpersTest.php
themech: 👍 https://github.com/Wikia/app/pull/16068/commits/f4985582c5c63580bd3ca401f79e5fc7c5758949

sczerwinski-wikia:
It was merged to devfandom branch, which has already been merged so I'll close this PR

PR #16070, opened 2018-09-20
ADEN-7677 Adding timeOut for prebid bidding

dmnsobczak:


fraszczakszymon:
Isn't this issue related to this part of code:
https://github.com/Wikia/app/blob/dev/extensions/wikia/ArticleVideo/scripts/featured-video.jwplayer.instant.js#L121
?
If user has disabled ads we don't need to run bidders/workaround it with special timeout. Am I wrong?

dmnsobczak:
Discussed on Slack. Changes applied.

fraszczakszymon (approved):

PR #16076, opened 2018-09-21
PLATFORM-3693 | wgEnableHTTPSForAnons no longer needed for fandom.com

rogatty:
https://wikia-inc.atlassian.net/browse/PLATFORM-3693

`.fandom.com` wikis will use HTTPS for everyone, no matter the value of `wgEnableHTTPSForAnons`.

@Wikia/core-platform-team


includes/wikia/GlobalFunctions.php
```diff
@@ -1557,3 +1557,12 @@ function wfGetBaseDomainForHost( $host ) {
 
 	return $wgWikiaBaseDomain;
 }
+
+function wfDoesURLHaveFandomDomain( string $url ): bool {
+	global $wgFandomBaseDomain;
+
+	$host = parse_url( $url, PHP_URL_HOST );
+	$host = WikiFactory::normalizeHost( $host );
```
rogatty: If https://github.com/Wikia/app/pull/16068 gets merged first, this should be updated as the `normalizeHost` method is going to be moved out of `WikiFactory` class by @themech

includes/wikia/GlobalFunctions.php
```diff
@@ -1557,3 +1557,17 @@ function wfGetBaseDomainForHost( $host ) {
...
+function wfDoesURLHaveFandomDomain( $url ): bool {
+	global $wgFandomBaseDomain;
+
+	// e.g. not existing wikis
+	if ( empty( $url ) ) {
+		return false;
+	}
+
+	$host = parse_url( $url, PHP_URL_HOST );
+	$host = WikiFactory::normalizeHost( $host );
```
rogatty: If https://github.com/Wikia/app/pull/16068 gets merged first, this should be updated as the `normalizeHost` method is going to be moved out of `WikiFactory` class by @themech

includes/wikia/GlobalFunctions.php
```diff
@@ -1557,3 +1557,17 @@ function wfGetBaseDomainForHost( $host ) {
...
+
+function wfDoesURLHaveFandomDomain( $url ): bool {
+	global $wgFandomBaseDomain;
+
+	// e.g. not existing wikis
+	if ( empty( $url ) ) {
+		return false;
+	}
+
+	$host = parse_url( $url, PHP_URL_HOST );
```
harnash: `$host` can be `false`

includes/wikia/GlobalFunctions.php
```diff
@@ -1557,3 +1557,17 @@ function wfGetBaseDomainForHost( $host ) {
...
+	$host = WikiFactory::normalizeHost( $host );
+
+	return wfGetBaseDomainForHost( $host ) === $wgFandomBaseDomain;
```
harnash: This might not work for dev environment. For example `normalizeHost()` will return `fandom-dev.us` then `wfGetBaseDomainForHost()` will return `wikia.com` which will be not the same as `$wgFandomBaseDomain`.

harnash (changes requested):

includes/wikia/GlobalFunctions.php
rogatty: `normalizeHost()` won't return `fandom-dev.us`, @themech fixed it in https://github.com/Wikia/app/pull/16068

includes/wikia/GlobalFunctions.php
rogatty: I moved the check in https://github.com/Wikia/app/pull/16076/commits/6a0e895d0637680f6827608e26306b4ecb5d6e01

includes/wikia/GlobalFunctions.php
themech: 👍

includes/wikia/GlobalFunctions.php
```diff
@@ -1557,3 +1557,18 @@ function wfGetBaseDomainForHost( $host ) {
...
+	global $wgFandomBaseDomain;
+
+	$host = parse_url( $url, PHP_URL_HOST );
+
+	// e.g. not existing wikis
+	if ( empty( $host ) ) {
+		return false;
+	}
+
+	$host = WikiFactory::normalizeHost( $host );
```
rogatty: If https://github.com/Wikia/app/pull/16068 gets merged first, this should be updated as the `normalizeHost` method will be moved out of `WikiFactory` class.

includes/wikia/GlobalFunctions.php
```diff
@@ -1557,3 +1557,18 @@ function wfGetBaseDomainForHost( $host ) {
 
 	return $wgWikiaBaseDomain;
 }
+
+function wfDoesURLHaveFandomDomain( $url ): bool {
```
themech: short comment about what this method does and that it should work on all env would be nice.
+ maybe renaming it to something like `wfIsFadomUrl`.

rogatty: `wfIsFandomUrl` was my first idea but I think the current name is better because it doesn't need the explanation, it just checks if the url has a fandom domain :man_shrugging: The fact that it works on all envs should be obvious, no? I'd add `TODO` if it didn't :wink:

themech (approved):

includes/wikia/GlobalFunctions.php
```diff
@@ -1557,3 +1557,28 @@ function wfGetBaseDomainForHost( $host ) {
...
+ * Returns true for URLs with fandom domain, some examples:
+ * - https://starwars.fandom.com/wiki/Yoda
+ * - starwars.fandom.com/wiki/Yoda
+ * - starwars.fandom-dev.pl/
+ * - starwars.fandom-dev.us
+ *
+ * @param $url
+ * @return bool
+ */
+function wfDoesURLHaveFandomDomain( $url ): bool {
```
Grunny: One thought here, what do you think of renaming this method to something like `wfForceHttpsForURL` or something similar? My thinking is this is the intent of the method, and decouples it from a specific domain/implementation.

rogatty: @Grunny I think you misclicked or misspelled, this method is not about forcing HTTPS at all.

Grunny: This method itself isn't, but everywhere it's used is, right? It's used in place of `wgEnableHTTPSForAnons`. What I'm thinking is that this method name should be decoupled from the fandom.com domain implementation, so if we enabled HTTPS by default on any other domains in the future, we don't need to change all these method calls or add extra checks, but instead just add the domains to the implementation of this method with a name like `wfHttpsEnabledForURL`. There may be a use for `wfDoesURLHaveFandomDomain` separately in the future, which could be used by `wfHttpsEnabledForURL`, but right now we only need to know if HTTPS is enabled by default for the URL. What do you think?

rogatty: I see. It'll be similar to `wfHttpsAllowedForURL` but I think it makes sense :+1: Updated in https://github.com/Wikia/app/pull/16076/commits/39269b4aec40657e0907b12a01d39871d75edffb

Grunny (approved):

PR #16078, opened 2018-09-24
ADEN-7702 Introduce the idea of firstCallSlots and secondCallSlots.

jacekbj:
The idea:
* we can configure which slots should be queried in the first call
* we can disable BTF slots


extensions/wikia/AdEngine/js/provider/directGpt.js
```diff
@@ -19,7 +19,8 @@ define('ext.wikia.adEngine.provider.directGpt', [
 		sraEnabled = !context.opts.disableSra,
 		atfSlots = [
 			'TOP_LEADERBOARD',
-			'GPT_FLUSH'
+			'GPT_FLUSH',
+			'TOP_RIGHT_BOXAD'
```
fraszczakszymon: It breaks non-SRA, isn't it?

jacekbj: It does not. SRA added `TOP_RIGHT_BOXAD` to atfSlots. Now it does not since it's already there.

extensions/wikia/AdEngine/js/provider/btfBlocker.js
```diff
@@ -69,32 +64,54 @@ define('ext.wikia.adEngine.provider.btfBlocker', [
...
+			}
+
+			// ATF slots are always called in second call if not called in first
+			if (config.atfSlots) {
+				config.atfSlots
+				// remove slot already called in first call
+				.filter(function(slotName) {
+					return config.firstCallSlots.indexOf(slotName) === -1;
+				})
+				// unblock slot
```
fraszczakszymon, line 89: What about unblocking BTF slots?

jacekbj: I really dislike the idea of variable `unblockedSlots`. Why not set the property `disabled` in slot config instead? What do you think about that?

jacekbj: I deleted the previous comment even though it was right :)
So:
ATF slots are added to `unblockedSlots` to make sure they're always used.
BTF slots will be used if `isSecondCallEnabled()` returns value `true` https://github.com/Wikia/app/blob/ADEN-7702/extensions/wikia/AdEngine/js/provider/btfBlocker.js#L55

fraszczakszymon: Got it 👍

fraszczakszymon (approved):

PR #16086, opened 2018-09-24
SRE-103 | Optimize MW Rabbit resources usage

mszabo-wikia:
We use Rabbit extensively in our MW write operations to publish various updates and background tasks. Currently there's no centralized way to acquire a Rabbit channel or connection so we are pretty wasteful with resources here—we can create as many as 20 connections and channels during the course of a write operation.

What's changed:
* there's now a centralized gateway, the `ConnectionManager` which is a one stop shop to acquire a Rabbit channel for publishing messages
* Rabbit publishers no longer need to be aware of connection details like host or credentials, they can just call the `ConnectionManager`, so there is no longer a need to duplicate that info for every publisher
* refactored `AsyncTaskList` by moving out publishing logic into `TaskPublisher` implementations, allowing us to publish all background tasks created during a request in a single operation
* basic integration tests for tasks publishing

Preliminary testing on sandbox suggests that batching up Rabbit resources (i.e. create 1 channel+connection per virtual host) can improve speed by up to **50%**—see callgraphs.

https://wikia-inc.atlassian.net/browse/SRE-103

### Before
callgraph_before
### After
callgraph_after


includes/wikia/rabbitmq/ConnectionBase.class.php
```diff
@@ -58,7 +44,7 @@ public function publish( $routingKey, $body ) {
 				$routingKey
 			);
 
-			$channel->wait_for_pending_acks(self::ACK_WAIT_TIMEOUT_SECONDS);
+			$channel->wait_for_pending_acks(self::ACK_WAIT_TIMEOUT_SECONDS );
```
ljagiello: Style: do you want a whitespace between parenthesis or not (one side is without and one is with).

ljagiello (approved):

PR #16090, opened 2018-09-25
configure Traefik to support fandom.com

ghost:
Per conversation with @pwojcik86 , I've decided to reconfigure Traefik to support `fandom.com` domain in the same way it supports `wikia.com`.

https://wikia-inc.atlassian.net/browse/SUS-5843

### DIFF
--- a/docker/sandbox/sandbox.template.yaml
+++ b/docker/sandbox/sandbox.template.yaml
@@ -166,7 +166,7 @@ metadata:
     kubernetes.io/ingress.class: traefik
 spec:
   rules:
-    - host: "${SANDBOX_NAME}.wikia.com"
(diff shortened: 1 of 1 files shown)


docker/sandbox/sandbox.template.yaml
```diff
@@ -166,7 +166,7 @@ metadata:
     kubernetes.io/ingress.class: traefik
 spec:
   rules:
-    - host: "${SANDBOX_NAME}.wikia.com"
+    - host: "${SANDBOX_NAME}.wikia.com", "${SANDBOX_NAME}.fandom.com"
```
pwojcik86: I just checked this manually, and it won't work. You need to create a separate entry.

pwojcik86 (changes requested):

pwojcik86 (approved):

PR #16095, opened 2018-09-25
PLATFORM-3725: fix wgServicesExternalDomain on dev

harnash:
Quick fix. I'm not 100% sure this (hack in WFL) is still needed though.
Mind the source branch

### DIFF
--- a/extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
+++ b/extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
@@ -214,8 +214,7 @@ class WikiFactoryLoader {
 	 */
 	public function execute() {
 		global $wgCityId, $wgDBservers, $wgLBFactoryConf, $wgDBserver, $wgContLang,
-			   $wgEnableHTTPSForAnons, $wgFandomBaseDomain, $wgDevelEnvironment,
-			   $wgFandomDevDomain;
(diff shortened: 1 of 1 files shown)


extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
```diff
@@ -232,8 +233,10 @@ public function execute() {
 
 		// Override wikia.com related config early when requesting a fandom.com wiki
 		if ( strpos( $this->mServerName, '.' . $wgFandomBaseDomain ) !== false ) {
-			$GLOBALS['wgServicesExternalDomain'] = "https://services.{$wgFandomBaseDomain}/";
-			$GLOBALS['wgCookieDomain'] = ".{$wgFandomBaseDomain}";
+			$domain = $wgDevelEnvironment ? $wgFandomDevDomain : $wgFandomBaseDomain;
```
themech: Doesn't `$wgFandomDevDomain` contain the devbox name like `mech.fandom-dev.pl`? If so, the `wgServicesExternalDomain` will become `https://services.mech.fandom-dev.pl/` which is wrong.

themech (changes requested):

extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
harnash: fix in 6f3b20d2297

themech (approved):

PR #16097, opened 2018-09-25
CAKE-3834 | Update resource route

JoshuaRogan:
@Wikia/cake @Wikia/sus

Uses the new resource (https://github.com/Wikia/pandora/blob/master/service/content-graph/content-graph-service/src/main/java/com/wikia/services/contentgraph/resources/NavigationResource.java#L57) but retains the original structure. 

Waiting for the final migration to populate site_navigation table before merging this.


nmonterroso:
👍 when the migration is done LGTM

macbre:
Why SUS was pinged here? :)

PR #16105, opened 2018-09-26
SRE-106 | Don't fetch pending SiteStatsUpdate deltas if 0 <= $wgSiteStatsAsyncFactor <= 1

mszabo-wikia:
Because of https://github.com/Wikia/app/blob/dev/includes/SiteStatsUpdate.php#L60 we don't need to deal with pending deltas if the async factor is less than or equal to 1, so we can spare the explicit locking.

https://wikia-inc.atlassian.net/browse/SRE-106

### DIFF
--- a/includes/SiteStatsUpdate.php
+++ b/includes/SiteStatsUpdate.php
@@ -71,7 +71,10 @@ class SiteStatsUpdate implements DeferrableUpdate {
 		$dbw = wfGetDB( DB_MASTER );
 		$lockKey = wfMemcKey( 'site_stats' ); // prepend wiki ID
 		$pd = [];
-		if ( $wgSiteStatsAsyncFactor ) {
+
+		// SRE-106: Only fetch pending deltas if we have an async factor of > 1
+		// Otherwise we never actually save any pending deltas into memcached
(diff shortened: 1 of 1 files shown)


ljagiello:
@mszabo-wikia my PHP knowledge is a bit limited but I have a hard time to find a place that modify `wgSiteStatsAsyncFactor` in the code. The only other place seems to be: https://github.com/Wikia/app/blob/8d9874172e357880516756afa319be4d15dafa94/includes/wikia/VariablesBase.php#L7367 where we always initialize value 1. Is that mean at least with the current code you will basically exclude that code always?

ljagiello (approved):

mszabo-wikia:
@ljagiello yeah, because of https://github.com/Wikia/app/blob/dev/includes/SiteStatsUpdate.php#L60 we never actually save pending deltas to memcached (`mt_rand( 0, $rate - 1 ) != 0` will never be true) so there is no use trying to fetch them later. As for the config value this variable is not in WikiFactory so it's consistently `1` everywhere.

ljagiello:
Yep, makes sense. Wasn't sure if I miss something. Thanks for clarification.

LGTM.

PR #16109, opened 2018-09-26
SRE | in proxy.php perform normal MW shutdown procedure

mszabo-wikia:
Ensure that the standard MW shutdown process is performed when executing background tasks—deferred updates and `RestInPeace` hook should be run, then we should commit any pending writes over our master connections.

This will be needed for example for https://github.com/Wikia/app/pull/16086


macbre (approved):

macbre (approved):

macbre:
Will be merged as a part of #16115

PR #16121, opened 2018-09-30
Remove an inline style in chat

LunarTwilight:
This PR removes an inline style in chat so that users no longer have to use `!important` for this in their CSS.


LunarTwilight:
Per chat with @KockaAdmiralac, this probably isn't the correct way to do this.

PR #16122, opened 2018-10-01
ADEN-7695 Update A9 apstag with deals

fraszczakszymon:


extensions/wikia/AdEngine/js/lookup/a9.js
```diff
@@ -166,28 +222,17 @@ define('ext.wikia.adEngine.lookup.a9', [
...
+		return bids[slotName] || {};
 	}
 
 	function getPrices() {
 		return priceMap;
 	}
 
 	function isSlotSupported(slotName) {
-		return slots[slotName] || VIDEO_SLOTS.indexOf(slotName) >= 0;
+		return slots[slotName];
```
jacekbj: I'd expect the function `isSlotSupported` to return a boolean value.

extensions/wikia/AdEngine/js/lookup/a9.js
```diff
@@ -84,11 +123,13 @@ define('ext.wikia.adEngine.lookup.a9', [
 			priceMap = {};
 
 			slots = slotsContext.filterSlotMap(config[skin]);
-			a9Slots = Object.keys(slots).map(createSlotDefinition);
 
-			if (isVideoBidderEnabled()) {
-				a9Slots = a9Slots.concat(VIDEO_SLOTS.map(createVideoSlotDefinition));
-			}
+			var a9Slots = Object
```
jacekbj, line 127: Is it possible that `a9Slots` will be empty? If so, should we proceed with execution?

fraszczakszymon: Handled 👍

dmnsobczak (approved):

PR #16120, opened 2018-09-28
CAKE-2151 - FC CNW opt in prompt/modal

nmonterroso:
https://wikia-inc.atlassian.net/browse/CAKE-2151
@Wikia/cake @Wikia/sus 

Adds prompts during CNW flow to create an FC community. The desired flow is described in the JIRA ticket.

TODO:
- [x] add global to switch off functionality
- [x] ~will the existing selenium tests work or do they need to be updated?~ They do not need to be updated

PR #16123, opened 2018-10-01
Revert "CAKE-2151 - FC CNW opt in prompt/modal"

macbre:
Reverts Wikia/app#16120


macbre:
This one does not seem to be a root cause of a release blocker from today.

PR #16126, opened 2018-10-01
SEO-669 | Show banner notification about fandom.com migration

rogatty:
https://wikia-inc.atlassian.net/browse/SEO-669

When a wiki is chosen to be migrated to `fandom.com` domain, we'll put the migration date in the `wgFandomComMigrationDate` WikiFactory variable. When a wiki is migrated, we'll set `wgFandomComMigrationDone` to `true`. It will cause a banner notification to display before and after the migration. When user closes the notification, we'll create an entry in LocalStorage to not show it again.

@Wikia/core-platform-team

### DIFF
--- a/extensions/wikia/FandomComMigration/FandomComMigration.i18n.php
+++ b/extensions/wikia/FandomComMigration/FandomComMigration.i18n.php
@@ -2,8 +2,8 @@
 $messages = array();
 
 $messages['en'] = array(
-	'fandom-com-migration-after' => 'Your community domain has been migrated to fandom.com. Find out more in <a href="https://community.wikia.com/wiki/User_blog:Brandon_Rhea/Wiki_domains_will_be_changing_from_wikia.com_to_fandom.com_in_early_2019?">this FANDOM news blog</a>.',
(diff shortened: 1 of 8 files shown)


extensions/wikia/FandomComMigration/FandomComMigration.i18n.php
```diff
@@ -0,0 +1,12 @@
+<?php
+$messages = array();
+
+$messages['en'] = array(
+	'fandom-com-migration-after' => 'Attention: We have migrated your community\'s domain to fandom.com. Find out more in <a href="https://community.wikia.com/wiki/User_blog:Brandon_Rhea/Wiki_domains_will_be_changing_from_wikia.com_to_fandom.com_in_early_2019?">this FANDOM news blog</a>.',
```
Grunny: Can we avoid the raw HTML here? I know it's temporary, so isn't a big risk security-wise, but it's still something translators have to deal with. We could change it to a wikitext external link for example (which can be parsed in JS too with `mw.message('fandom-com-migration-after').parse()`).

Grunny (commented):
One question/suggestion, otherwise looks good.

extensions/wikia/FandomComMigration/FandomComMigration.i18n.php
rogatty: :+1: updated in https://github.com/Wikia/app/pull/16126/commits/28adc8cce8cf06071207c84246172bd28376475b

Grunny (approved):

PR #16128, opened 2018-10-01
CAKE-2151 - product review feedback

nmonterroso:
@Wikia/cake 
https://wikia-inc.atlassian.net/browse/CAKE-2151

- adds tracking
- adds specific class to left column so that "ghost" elements (like grammarly) don't take up width
- links/buttons/checkboxes/dropdowns all use the proper DS link color
- button in prompt ad was submitting the form on the second screen, that's now fixed

### DIFF
--- a/extensions/wikia/CreateNewWiki/js/CommunityBuilderOptIn.js
+++ b/extensions/wikia/CreateNewWiki/js/CommunityBuilderOptIn.js
@@ -1,13 +1,18 @@
-define('ext.createNewWiki.communityBuilderOptIn', [], function() {
+define('ext.createNewWiki.communityBuilderOptIn', ['wikia.tracker'], function(tracker) {
 	'use strict';
 
 	var currentLanguage,
 		optInPrompts,
+		learnMoreLinks,
 		sectionWrapper,
 		optInModal,
 		nameInput,
 		domainInput,
 		viewedOptInModal = false,
+		track = tracker.buildTrackingFunction({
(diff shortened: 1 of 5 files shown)


extensions/wikia/CreateNewWiki/js/CommunityBuilderOptIn.js
```diff
@@ -81,7 +88,21 @@ define('ext.createNewWiki.communityBuilderOptIn', [], function() {
...
 
-		window.location = 'http://community-builder.wikia.com/community/create?'+params.join('&');
+		track({
+			action: tracker.ACTIONS.CLICK,
+			label: 'use-opt-in'
+		});
+
+		setTimeout(function() {
+			window.location = 'http://community-builder.wikia.com/community/create?'+params.join('&');
+		}, 50);
```
lizlux, line 98: Is `50` really better than `0`? I know we've run into this issue in the past (where navigation happens right after tracking) but every time I come across it I have to refresh my memory as to the solution.

nmonterroso: It's what happens in the core tracker, I just used a smaller number. Not sure it's worth trying to figure out what the right amount is.

lizlux: Cool

drsnyder:
LGTM

PR #16130, opened 2018-10-01
SRE-107 | Refactor CreateNewWiki flow for performance

mszabo-wikia:
* Use a new bulk import class to import starter XML dump instead of the vanilla importer (**280x** faster)
* Don't import links tables SQL dump from starter—it's just invalid data that'll be cleaned up by the post-creation `refreshLinks.php` run
* Insert WikiFactory variables in a single bulk step during wiki creation
* Don't spawn maintenance script processes—switch task execution context to the new wiki after it's been initialized and perform the operations within the task
* Don't use `wfWaitForSlaves` due to abysmal performance—in replication-sensitive situations just wait until the slave we're connected to has the latest data; if a next element of the process depends on it it will be using the same slave connection.
* Improve integration test coverage of the process
* Remove the practically unused `wgUniversalCreationVariables` / `wgLangCreationVariables` pair in favor of conditionals in the `ConfigureWikiFactory` step
* Use [mysqli::multi_query](http://php.net/manual/en/mysqli.multi-query.php) to initialize the DB schema for the new wiki (**16x** faster)

https://wikia-inc.atlassian.net/browse/SRE-107


JoshuaRogan:
I went through the code and it appears there aren't any interface changes that we would need to update when we create communities from Fandom Creator (https://github.com/Wikia/pandora/blob/12c6f44059768850fea405181584ce81e318e82c/service/content-graph/content-graph-service/src/main/java/com/wikia/services/contentgraph/service/mediawiki/communitycreate/MwCommunityCreator.java#L94) 

If you can confirm that would be great.

If you could in the future request a review from "Wikia/cake" whenever the create new wiki flow changes.

mszabo-wikia:
@JoshuaRogan yes that is correct, neither the user facing frontend nor the exposed HTTP APIs have changed.

I was planning to mention CAKE on this ticket but forgot about that; my bad :C Thanks for the review!

ghost:
LGTM, I am taking this for a ride with Kubernetes. I need to debug CNW on Kubernetes so it makes more sense to consolidate our efforts and "fix" your branch to meet my needs. :)

extensions/wikia/CreateNewWiki/classes/BulkRevisionImporter.php
```diff
@@ -0,0 +1,165 @@
+<?php
+
+use Wikia\CreateNewWiki\Tasks\TaskContext;
+
+/**
+ * Import a set of pages into the wiki database in a bulk operation
```
mszabo-wikia, line 6: The MediaWiki provided importer class doesn't suit our needs—it makes way too many queries for the 200+ pages we need to import. So I created a class to support bulk imports.

extensions/wikia/CreateNewWiki/tasks/ConfigureWikiFactory.php
```diff
@@ -72,7 +72,19 @@ public function getStaticVariables( $siteName, $imagesURL, $imagesDir, $dbName,
...
-			'wgWikiDescription' => $description
+			'wgWikiDescription' => $description,
+
+			// Enable Discussions
+			'wgEnableDiscussions' => true,
+			'wgEnableDiscussionsNavigation' => true,
+			'wgArchiveWikiForums' => true,
+			'wgEnableRecirculationDiscussions' => true,
+
+			// Features to enable for new wikis, moved here from wgUniversalCreationVariables
```
mszabo-wikia, line 83: These are from the value of wgUniversalCreationVariables on wikiaglobal wiki. The variable was referring to a lot of nonexistent variable IDs too, which I got rid of.

The reason behind this change was to eschew separate queries to change variables and do it all at once.

extensions/wikia/CreateNewWiki/tasks/ConfigureWikiFactory.php
```diff
@@ -85,34 +97,47 @@ public function getStaticVariables( $siteName, $imagesURL, $imagesDir, $dbName,
 			$wikiFactoryVariables[ 'wgWikiDirectedAtChildrenByFounder' ] = true;
 		}
 
+		// Language specific settings, moved here from wgLangCreationVariables
```
mszabo-wikia, line 100: Same situation as wgUniversalCreationVariables

extensions/wikia/CreateNewWiki/tasks/ConfigureWikiFactory.php
```diff
@@ -85,34 +97,47 @@ public function getStaticVariables( $siteName, $imagesURL, $imagesDir, $dbName,
...
 	 */
-	public function getVariablesFromDB( $sharedDBW, $wikiFactoryVariables ) {
-		wfProfileIn( __METHOD__ );
+	public function getVariablesFromDB( array $wikiFactoryVariables ): array {
+		global $wgExternalSharedDB;
+
+		$dbr = wfGetDB( DB_SLAVE, [], $wgExternalSharedDB );
 
-		$oRes = $sharedDBW->select(
+		$res = $dbr->select(
```
mszabo-wikia, line 129: We can safely read from slave here and save some time with that.

extensions/wikia/CreateNewWiki/tasks/ConfigureWikiFactory.php
```diff
@@ -124,36 +149,26 @@ public function getVariablesFromDB( $sharedDBW, $wikiFactoryVariables ) {
...
-						"cv_variable_id" => $cvId
-					],
-					__METHOD__
-				);
+			if ( !empty( $cvId ) ) {
+				$rows[] = [
+					"cv_value" => serialize( $value ),
+					"cv_city_id" => $cityId,
+					"cv_variable_id" => $cvId,
+				];
```
mszabo-wikia, line 166: previously we'd insert every variable one-by-one—we can batch that for performance

extensions/wikia/CreateNewWiki/tasks/CreateDatabase.php
```diff
@@ -47,7 +47,12 @@ public function run() {
 		$dbw = wfGetDB( DB_MASTER, [ ], $this->clusterDB );
 		$this->taskContext->setWikiDBW($dbw);
 
-		$dbw->query( sprintf( "CREATE DATABASE `%s`", $this->taskContext->getDBname() ) );
+		$safeDbName = $dbw->addIdentifierQuotes( $this->taskContext->getDbName() );
+
+		$dbw->query( "CREATE DATABASE $safeDbName" );
+		$dbw->commit( __METHOD__ );
+
+		$this->waitUntilDatabaseExistsOnSlave();
```
mszabo-wikia, line 55: We need to make sure that the database exists on the cluster slave we're connected to when we move on to the next steps (creating tables). wfWaitForSlaves is expensive (can take over 30+ seconds), so let's just ensure the status of the slave we are connected to, as subsequent wfGetDB invocation for this database during this request will return that connection.

extensions/wikia/CreateNewWiki/tasks/CreateTables.php
```diff
@@ -55,21 +55,49 @@ public function run() {
 		$dbw = wfGetDB( DB_MASTER, [ ], $this->taskContext->getDBname() );
 		$this->taskContext->setWikiDBW( $dbw );
 
-		foreach ( $this->sqlFiles as $file ) {
-			$this->debug( __METHOD__ . ": Populating database with {$file}" );
+		$dbw->multiQuery( $this->getSchemaSQL(), __METHOD__ );
+		$dbw->commit( __METHOD__ );
```
mszabo-wikia, line 59: mysqli::multi_query is 16x faster than using DatabaseBase::sourceStream that sends query for each statement

extensions/wikia/CreateNewWiki/tasks/CreateTables.php
```diff
@@ -55,21 +55,49 @@ public function run() {
...
+
+		$this->waitUntilTableExistsOnSlave();
```
mszabo-wikia, line 61: As with creating the database, we want to poll the slave we're connected to to ensure the tables are ready before we move on to import data.

extensions/wikia/CreateNewWiki/tasks/EnableDiscussionsTask.php
```diff
@@ -28,19 +27,6 @@ private function activateDiscussions() {
...
-	 * Sets the appropriate wg variables to make the new discussions
-	 * instance enabled and available on this new wiki.
-	 */
-	private function enableDiscussions() {
-		( new \DiscussionsVarToggler( $this->taskContext->getCityId() ) )
-			->setEnableDiscussions( true )
-			->setEnableDiscussionsNav( true )
-			->setArchiveWikiForums( true )
-			->setEnableRecircDiscussions( true )
-			->save();
```
mszabo-wikia, line 41: These were moved to ConfigureWikiFactory step bulk variables insert.

extensions/wikia/CreateNewWiki/tasks/ImportStarterData.php
```diff
@@ -9,72 +9,168 @@ class ImportStarterData extends Task {
...
+
+		$bulkRevisionImporter = new \BulkRevisionImporter( $this->taskContext );
+
+		// We don't want to invoke tasks (namely run Page::doEditUpdates) on each insert to revision table
+		// The links will be updated by a run of refreshLinks.php script in post creation tasks
+		$importer->setNoUpdates( true );
+
+		$importer->setShouldCheckPermissions( false );
+
+		$importer->setRevisionCallback( [ $bulkRevisionImporter, 'addRevision' ] );
```
mszabo-wikia, line 115: We can skip permissions check here (expensive and redundant) and we need to use the bulk importer for efficiency.

extensions/wikia/CreateNewWiki/tasks/SetupWikiCities.php
```diff
@@ -12,42 +12,42 @@ class SetupWikiCities extends Task {
...
 		$this->debug(
 			implode( ":",
-				[ __METHOD__, "Row added into city_domains table, city_id = {$cityId}" ]
+				[ __METHOD__, "Row added into city_domains table, city_id = {$wikiId}" ]
 			)
 		);
 
 		$this->taskContext->getSharedDBW()->commit( __METHOD__ ); // commit shared DB changes
 
+		$this->overrideLocalContext( $wikiId );
```
mszabo-wikia, line 43: This allows us to perform subsequent operations like importing data without having to spawn maintenance scripts. So the relevant code can be inlined into their respective tasks.

lib/Wikia/src/Tasks/Tasks/CreateNewWikiTask.php
```diff
@@ -65,7 +64,6 @@ public function postCreationSetup( $params ) {
 			$this->setWelcomeTalkPage();
 		}
 
-		$this->populateCheckUserTables();
```
mszabo-wikia, line 68: There is no need to retroactively populate CU tables because there is simply no data—wiki is brand new

ghost (approved):

extensions/wikia/CreateNewWiki/tasks/ConfigureWikiFactory.php
ghost: I am wondering, if we could abstract the initial configuration to a template and separate it from the codebase, so that we don't have a couple of variables here, another couple of variables there. That's for the future. Do you think it's worth doing as part of some general cleanup?

extensions/wikia/CreateNewWiki/tasks/ConfigureWikiFactory.php
ghost: Please, refer to my previous comment. :)

extensions/wikia/CreateNewWiki/tasks/CreateTables.php
ghost: _Ah, I hate waiting for slaves_ ― said Spartacus and escaped :)

But yeah, in this particular case it will address some issues and won't hurt a user, since they are waiting for wiki to be created anyway.

includes/wikia/VariablesBase.php
```diff
@@ -1277,11 +1277,11 @@
...
 /**
- * Wiki description enter by the user on the CNW page. Removed after wiki is created
- * @see extensions/wikia/CreateNewWiki/maintenance/setMainPageContent.php
- * @var string $wgWikiDescription
+ * The datacenter that hosts the starter dumps for CreateNewWiki
+ * @see \Wikia\CreateNewWiki\Starters
+ * @var string $wgCreateWikiStarterDumpsDatacenter
  */
-$wgWikiDescription = '';
+$wgCreateWikiStarterDumpsDatacenter = WIKIA_DC_SJC;
```
ghost: I'd put this to `VariablesExpansions.php`. This file should change the state in the same way, always. Now it depends on a constant defined elsewhere.

ghost (changes requested):

ghost (approved):
All tests pass on k8s sandbox, good job!

docker/sandbox/site.conf
```diff
@@ -28,13 +28,14 @@ server {
     # rewrites for language wiki
     rewrite "^/[a-z]{2,3}(?:-[a-z-]{2,12})?/(sitemap.+\.xml[.gz]*)$" /index.php?title=Special:Sitemap/$1&uselang=en break;
 
-    rewrite "^/(?:[a-z]{2,3}(?:-[a-z-]{2,12})?/)api.php(.*)" /api.php$1;
+    rewrite "^/(?:[a-z]{2,3}(?:-[a-z-]{2,12})?/)?api.php(.*)" /api.php$1;
```
ghost, line 31: Lack of consistency with the rest of rewrites for language wikis. I've added `?` to make the language prefix optional, as it is in other rules, e.g. `load.php` and `wikia.php`.

docker/sandbox/site.conf
```diff
@@ -76,7 +77,8 @@ server {
 
     # SUS-5792: rewrite for legacy corporate pages with short article path
     # see wgShortArticlePathWikis MediaWiki variable for the list of them
-    if ($http_x_original_host ~ ^(www|de|fr|pl|es|ja|ru|it).sandbox[^\.]+.wikia.com$) {
+    if ($http_x_original_host ~ "^(www|de|fr|pl|es|ja|ru|it)\.wikia\.com$") {
```
ghost, line 80: This regex was bad - `.` had to be escaped and the entire regex encapsulated inside `"`. In addition, I've added `api.php` specific rule here as there was no `break` in the above rule for `api.php`. Other rules, such as `load.php` and `wikia.php` has `break` so such URLs will never end up here.

ghost:
This solves https://wikia-inc.atlassian.net/browse/SUS-5872 and https://wikia-inc.atlassian.net/browse/SUS-5863

ghost:
@chessky will run an additional round of tests and merge it if passed QA

PR #16131, opened 2018-10-02
DE-3843 | wgIsTestWiki in DW DimensionApi

ghost:


includes/wikia/api/DWDimensionApiControllerSQL.class.php
```diff
@@ -16,10 +16,15 @@ class DWDimensionApiControllerSQL {
            city.city_vertical      AS vertical_id,
            IFNULL(city.city_cluster, \'c1\') AS cluster,
            city.city_created       AS created_at,
-           0                       AS deleted
+           0                       AS deleted,
+           IF(cv_test_wiki.cv_value=\'b:1;\', \'1\', \'0\') AS is_test_wiki
```
jcellary: Is there a reason you return 1 and 0 as string instead of number?

includes/wikia/api/DWDimensionApiControllerSQL.class.php
```diff
@@ -16,10 +16,15 @@ class DWDimensionApiControllerSQL {
...
       FROM city_list city
       LEFT JOIN city_lang lang
         ON lang.lang_code = city.city_lang
+      LEFT JOIN city_variables cv_test_wiki
```
jcellary: Is this variable always set? Maybe we should do an outer join if it's optional?

jcellary (approved):
Asked some questions, other than than lgtm

includes/wikia/api/DWDimensionApiControllerSQL.class.php
ghost: You're right, there is no need to keep here string value

includes/wikia/api/DWDimensionApiControllerSQL.class.php
ghost: LEFT join is OUTER by default

macbre (approved):

PR #16132, opened 2018-10-02
(PLATFORM-3728) Add script to migrate a list of wikis to fandom.com

Grunny:
/cc @Wikia/core-platform-team


harnash (approved):

rogatty (approved):

harnash:
@Grunny during tests I've made a mistake in the domain and the script allowed me to change it which caused my wiki to be inaccessible. I think it would be worth to at least check if the `$targetDomain` at least matches `$wgWikiaBaseDomainRegex`. What do you think?

Grunny:
@harnash good point, done in 58440b82c5b.

harnash (approved):

PR #16135, opened 2018-10-02
SUS-5863 Migrate missing Apache rewrite rules

ghost:
https://wikia-inc.atlassian.net/browse/SUS-5863

### DIFF
--- a/docker/dev/site.conf
+++ b/docker/dev/site.conf
@@ -79,7 +79,7 @@ server {
 
     # SUS-5792: rewrite for legacy corporate pages with short article path
     # see wgShortArticlePathWikis MediaWiki variable for the list of them
-    if ($host ~ ^(www|de|fr|pl|es|ja|ru|it).dev.wikia-local.com$) {
+    if ($host ~ ^(www|de|fr|pl|es|ja|ru|it)\.dev\.wikia-local\.com$) {
       rewrite ^/(.*)$ /index.php?title=$1 break;
     }
 
--- a/docker/dev/site.conf
+++ b/docker/dev/site.conf
@@ -77,11 +77,5 @@ server {
         fastcgi_param SERVER_NAME $host;
     }
 
-    # SUS-5792: rewrite for legacy corporate pages with short article path
-    # see wgShortArticlePathWikis MediaWiki variable for the list of them
-    if ($host ~ "^(www|de|fr|pl|es|ja|ru|it)\.dev\.wikia-local\.com$") {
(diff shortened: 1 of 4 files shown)


mixoteric:
The rules should also be migrated to other environments (`dev`, `prod`), although behaviour might be different. Some URL patterns are handled by Varnish.

docker/sandbox/site.conf
```diff
@@ -80,5 +85,9 @@ server {
       rewrite ^/(.*)$ /index.php?title=$1 break;
     }
 
+    # SUS-5863 Short URLs as http://www.wikia.com/Special:Version should not be processed by redirect-canonical.php
```
ghost: Actually, this one should be handled by the above rule, but for some reason is not.

ghost:
```
    # SUS-5863 Some www.wikia.com URLs should be redirected immediately.
    if ($http_x_original_host = 'www.wikia.com') {
        rewrite "^/(index.php/)?(Main_Page|Wikia)?$" http://www.wikia.com/fandom permanent;
    }
```
skipping this one for `dev`, it'll be handled by the last rule as regular main pages.

chessky (commented):
💠 🏃 ....... :)

docker/dev/site.conf
```diff
@@ -79,7 +79,7 @@ server {
 
     # SUS-5792: rewrite for legacy corporate pages with short article path
     # see wgShortArticlePathWikis MediaWiki variable for the list of them
-    if ($host ~ ^(www|de|fr|pl|es|ja|ru|it).dev.wikia-local.com$) {
+    if ($host ~ "^(www|de|fr|pl|es|ja|ru|it)\.dev\.wikia-local\.com$") {
```
hakubo: This is broken. 
as this will break requests to e.g. api.php

I have just got to know that corporate pages are being migrated
https://wikia-inc.atlassian.net/browse/CT-2282

I propose that we do not handle them at all in the new setup. And treat this ticket as a blocker

ghost: @hakubo - to some extent, we may treat CT-2282 as a blocker, depending on how long it's going to take. Apart from corporate pages, www.wikia.com has short article paths. Perhaps we should unify the configuration after the corporate pages are moved? ( e.g. so that www.wikia.com/Special:CreateNewWiki becomes www.wikia.com/wiki/Special:CreateNewWiki?

hakubo (approved):

macbre (approved):

PR #16140, opened 2018-10-03
SUS-5847: catch all Exceptions and Errors in proxy.php

macbre:
https://wikia-inc.atlassian.net/browse/SUS-5847

When debugging problems with celery-workers and executing MediaWiki tasks "manually" from Python console, I've found it useful to properly handled exceptions and errors thrown by tasks code. Make `proxy.php` return a proper JSON response with "failure" field. `celery-worker` will log the following:

```
[2018-10-03 09:48:40,343: ERROR/Worker-4] {"exception_type": "RemoteExecuteError", "exception": "RemoteExecuteError(u'TypeError: Argument 6 passed to Wikia\\\\Tasks\\\\TaskRunner::__construct() must be of the type float, null given, called in /usr/wikia/source/deploytools/24231/src/lib/Wikia/src/Tasks/TaskRunner.php on line 41',)", "event": "Task failed", "task_id": "ecb4a7e4-438b-430c-9efa-ac64b1708b39"}
```

### DIFF
--- a/extensions/wikia/Tasks/proxy/proxy.php
+++ b/extensions/wikia/Tasks/proxy/proxy.php
@@ -52,11 +52,19 @@ require ( $IP . '/includes/WebStart.php' );
 $request = new FauxRequest( $_POST, true );
 
 // finally, execute the task
-$runner = Wikia\Tasks\TaskRunner::newFromRequest( $request );
-$runner->run();
+try {
+	$runner = Wikia\Tasks\TaskRunner::newFromRequest( $request );
+	$runner->run();
+	$resp = $runner->format();
+} catch ( Throwable $ex ) {
+	$resp = [
+		'status' => 'failure',
+		'reason' => sprintf('%s: %s', get_class( $ex ), $ex->getMessage() ),
+	];
+}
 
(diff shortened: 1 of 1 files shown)


extensions/wikia/Tasks/proxy/proxy.php
```diff
@@ -52,11 +52,19 @@
 $request = new FauxRequest( $_POST, true );
 
 // finally, execute the task
-$runner = Wikia\Tasks\TaskRunner::newFromRequest( $request );
-$runner->run();
+try {
+	$runner = Wikia\Tasks\TaskRunner::newFromRequest( $request );
+	$runner->run();
+	$resp = $runner->format();
+} catch ( Throwable $ex ) {
```
hakubo, line 59: $exception?

hakubo (approved):

extensions/wikia/Tasks/proxy/proxy.php
ghost: Both, `$ex` and `$exception`, as well as `$e` are used quite frequently across our codebase. Personally, I prefer the shortest as `Throwable` provides all the relevant information and `catch` blocks are usually very short.

ghost (approved):

PR #16143, opened 2018-10-03
SUS-5861 add x-original-host when using local proxy

pwojcik86:
https://wikia-inc.atlassian.net/browse/SUS-5861

Whenever we proxy a HTTP request locally on k8s, we need to set the proper `x-original-host` header, since nginx needs it to forward the correct host to PHP.

### DIFF
--- a/includes/HttpFunctions.php
+++ b/includes/HttpFunctions.php
@@ -382,9 +382,13 @@ class MWHttpRequest {
 			return;
 		}
 
-		global $wgHTTPProxy;
+		global $wgHTTPProxy, $wgSetOriginalHostHeader;
 		if ( $wgHTTPProxy ) {
-			$this->proxy = $wgHTTPProxy ;
+			$this->proxy = $wgHTTPProxy;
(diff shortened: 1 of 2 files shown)


hakubo (approved):

ghost (approved):

includes/HttpFunctions.php
```diff
@@ -382,9 +382,13 @@ public function proxySetup() {
 			return;
 		}
 
-		global $wgHTTPProxy;
+		global $wgHTTPProxy, $wgSetOriginalHostHeader;
 		if ( $wgHTTPProxy ) {
-			$this->proxy = $wgHTTPProxy ;
+			$this->proxy = $wgHTTPProxy;
+			if ( $wgSetOriginalHostHeader ) {
```
macbre, line 388: `!empty($wgSetOriginalHostHeader)` - this variable is not always set via `LocalSettings.php`.

macbre (changes requested):
See above

pwojcik86:
Turns out this was already solved in scope of another ticket

PR #16144, opened 2018-10-03
SEO-670 | SEO friendly category pages

rogatty:
https://wikia-inc.atlassian.net/browse/SEO-670

@Wikia/core-platform-team @mszabo-wikia


extensions/wikia/CategoryPage3/CategoryPage3Model.class.php
```diff
@@ -0,0 +1,284 @@
...
+	/**
+	 * @param DatabaseMysqli $dbr
+	 * @return ResultWrapper
+	 * @throws FatalError
+	 * @throws MWException
+	 */
+	private function getMembersFromDB( DatabaseMysqli $dbr ): ResultWrapper {
+		$extraConditions = $this->getExtraConditionsForQuery( $dbr );
+
+		return $dbr->select(
```
rogatty: Logs show it takes ~10-20 ms under normal load.

extensions/wikia/CategoryPage3/CategoryPage3Model.class.php
```diff
@@ -0,0 +1,284 @@
...
+	private function addPage( Title $title, $sortkey ) {
+		global $wgContLang;
+
+		$firstChar = $wgContLang->convert( $this->collation->getFirstLetter( $sortkey ) );
+
+		$this->members[ $title->getArticleID() ] = new CategoryPage3Member( $title, $firstChar );
+	}
+
+	private function findTotalNumberOfMembers( DatabaseMysqli $dbr ) {
+		$totalNumberOfMembers = $dbr->selectField(
```
rogatty: Logs show it takes ~10 ms for a category with ~44k members and ~1ms for a category with ~3k members under normal load.

extensions/wikia/CategoryPage3/CategoryPage3Model.class.php
```diff
@@ -0,0 +1,284 @@
...
+
+		$this->totalNumberOfMembers = $totalNumberOfMembers;
+	}
+
+	private function findPrevPage( DatabaseMysqli $dbr ) {
+		if ( $this->from === null ) {
+			return;
+		}
+
+		$res = $dbr->select(
```
rogatty: Logs show it takes ~200 ms for a category with ~44k members and ~2-4ms for a category with ~3k members under normal load.

extensions/wikia/CategoryPage3/CategoryPage3Model.class.php
```diff
@@ -0,0 +1,284 @@
...
+		}
+
+		$title = Title::newFromRow( $lastRow );
+		$this->pagination->setPrevPageKey( $this->getHumanSortkey( $lastRow, $title ) );
+	}
+
+	private function findLastPage( DatabaseMysqli $dbr ) {
+		$lastPageMembersCount = $this->totalNumberOfMembers % static::MEMBERS_PER_PAGE_LIMIT;
+
+		$res = $dbr->select(
```
rogatty: Logs show it takes ~200 ms for a category with ~44k members and ~10ms for a category with ~3k members under normal load.

includes/db/Database.php
```diff
@@ -203,7 +203,7 @@ function getServerInfo();
 abstract class DatabaseBase implements DatabaseType {
 
 	// @const log 5% of queries (increased from 1% in SUS-2974)
-	const QUERY_SAMPLE_RATE = 0.05;
+	const QUERY_SAMPLE_RATE = 1;
```
rogatty: Don't forget about it.

extensions/wikia/CategoryPage3/CategoryPage3Hooks.class.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+class CategoryPage3Hooks {
+	/**
+	 * @param Title $title
+	 * @param Article $article
+	 * @return bool
+	 */
+	static public function onArticleFromTitle( &$title, &$article ): bool {
+		if ( !$title || $title->getNamespace() !== NS_CATEGORY ) {
```
Grunny: `!$title->inNamespace( NS_CATEGORY )` would be preferred.

extensions/wikia/CategoryPage3/CategoryPage3Member.class.php
```diff
@@ -0,0 +1,45 @@
...
+
+	/**
+	 * @return Title
+	 */
+	public function getTitle(): Title {
+		return $this->title;
+	}
+
+	public function isSubcategory(): bool {
+		return $this->getTitle()->getNamespace() === NS_CATEGORY;
```
Grunny: `$this->getTitle()->inNamespace( NS_CATEGORY )`

harnash (approved):

extensions/wikia/CategoryPage3/CategoryPage3CacheHelper.class.php
```diff
@@ -0,0 +1,30 @@
...
+class CategoryPage3CacheHelper {
+	const EXPIRATION_TIME = 60 * 60;
+
+	/**
+	 * Can be used as a part of memcache key
+	 *
+	 * @param $title
+	 * @return bool|Object
+	 */
+	public static function getTouched( $title ) {
```
Grunny: I'd add a type here and on `setTouched`, given it would fail on `getKey` if it wasn't an instance of `Title` anyway, so I assume we can be strict.

extensions/wikia/CategoryPage3/CategoryPage3Hooks.class.php
```diff
@@ -0,0 +1,38 @@
...
+
+		return true;
+	}
+
+	/**
+	 * @param Title $title
+	 * @param Article $article
+	 * @return bool
+	 */
+	static public function onArticleFromTitle( &$title, &$article ): bool {
```
Grunny: No need for the `&` on these, objects are effectively passed by reference in PHP regardless.

mszabo-wikia: I think it's needed in this case for `$article` at least because we want to overwrite its value entirely

Grunny: Ah, yeah, you're right, I missed that we were doing that here. So, yeah, only unnecessary on `$title`.

extensions/wikia/CategoryPage3/CategoryPage3Hooks.class.php
```diff
@@ -123,13 +123,51 @@ public static function onGetPreferences( $user, &$defaultPreferences ) {
...
 
-	static public function onUserGetDefaultOptions( &$defaultOptions ) {
+	public static function onLinkerMakeExternalLink( string $url, string $text, string &$link, array $attribs ): bool {
+		global $wgCityId, $wgContLang, $wgFandomBaseDomain, $wgWikiaBaseDomain;
+
+		$host = parse_url( $url, PHP_URL_HOST );
+		$host = wfNormalizeHost( $host );
+
+		// External-external links
+		if ( !endsWith( $host, $wgFandomBaseDomain ) && !endsWith( $host, $wgWikiaBaseDomain ) ) {
```
Grunny: This would pass for domains like `examplefandom.com`, right? It should check if it ends in `".{$wgFandomBaseDomain}"` as we do elsewhere to be safe.

extensions/wikia/CategoryPage3/CategoryPage3Hooks.class.php
```diff
@@ -123,13 +123,51 @@ public static function onGetPreferences( $user, &$defaultPreferences ) {
...
+		// Not this wiki
+		if ( WikiFactory::UrlToID( $host ) !== $wgCityId ) {
+			return true;
+		}
+
+		$path = parse_url( $url, PHP_URL_PATH );
+		$categoryNsText = $wgContLang->getNsText( NS_CATEGORY );
+
+		// Not a category
+		if ( strpos( $path, $categoryNsText ) === false ) {
```
Grunny: Maybe include the colon here just in case "Category" as a string appears in another page title.

extensions/wikia/CategoryPage3/CategoryPage3Hooks.class.php
```diff
@@ -170,4 +208,11 @@ private static function getLayoutForKey( Title $title, Article $article, $key )
 
 		return null;
 	}
+
+	private static function getNonCrawlableLink( $url, $text, $attribs ) {
+		$attribs['href'] = '#';
+		$attribs['onclick'] = 'window.location.assign("' . rawurlencode( $url ) . '"); return false;';
```
Grunny: We should wrap the encoded URL in `Xml::escapeJsString`, just in case.

Grunny (changes requested):

extensions/wikia/CategoryPage3/CategoryPage3Hooks.class.php
```diff
@@ -130,7 +130,10 @@ public static function onLinkerMakeExternalLink( string $url, string $text, stri
 		$host = wfNormalizeHost( $host );
 
 		// External-external links
-		if ( !endsWith( $host, $wgFandomBaseDomain ) && !endsWith( $host, $wgWikiaBaseDomain ) ) {
+		if (
+			!endsWith( $host, ".${$wgFandomBaseDomain}" ) &&
+			!endsWith( $host, ".${$wgWikiaBaseDomain}" )
```
Grunny: One too many `$`.

extensions/wikia/CategoryPage3/CategoryPage3Hooks.class.php
```diff
@@ -143,7 +146,7 @@ public static function onLinkerMakeExternalLink( string $url, string $text, stri
 		$categoryNsText = $wgContLang->getNsText( NS_CATEGORY );
 
 		// Not a category
-		if ( strpos( $path, $categoryNsText ) === false ) {
+		if ( strpos( $path, "${$categoryNsText}:" ) === false ) {
```
Grunny: One too many `$`. Also, can `$path` be URL encoded here? If so, should we account for any case in which `:` might be `%3A` (maybe just by unencoding `$path` for the comparison)?

extensions/wikia/CategoryPage3/CategoryPage3Hooks.class.php
```diff
@@ -210,8 +213,9 @@ private static function getLayoutForKey( Title $title, Article $article, $key )
 	}
 
 	private static function getNonCrawlableLink( $url, $text, $attribs ) {
+		$urlEscaped = Xml::escapeJsString( rawurlencode( $url ) );
 		$attribs['href'] = '#';
-		$attribs['onclick'] = 'window.location.assign("' . rawurlencode( $url ) . '"); return false;';
+		$attribs['onclick'] = "window.location.assign('${$urlEscaped}'); return false;";
```
Grunny: One too many `$`.

Grunny (changes requested):

extensions/wikia/CategoryPage3/CategoryPage3Hooks.class.php
```diff
@@ -0,0 +1,222 @@
...
+	 * @param Title $title
+	 * @param Article $article
+	 * @return bool
+	 */
+	public static function onArticleFromTitle( $title, &$article ): bool {
+		if ( $title->isRedirect() ) {
+			$title = $article->getRedirectTarget();
+		}
+
+		if ( is_null( $title ) || !$title->inNamespace( NS_CATEGORY ) ) {
```
mszabo-wikia, line 49: this null check is redundant, since the method call above will cause an error if `$title` is null

extensions/wikia/CategoryPage3/CategoryPage3Hooks.class.php
```diff
@@ -0,0 +1,222 @@
...
+		$options[ wfMessage( 'category-page3-layout-selector-mediawiki' )->escaped() ] = CategoryPageWithLayoutSelector::LAYOUT_MEDIAWIKI;
+		$options[ wfMessage( 'category-page3-layout-selector-category-exhibition' )->escaped() ] = CategoryPageWithLayoutSelector::LAYOUT_CATEGORY_EXHIBITION;
+		$preference['options'] = $options;
+
+		$defaultPreferences[ static::GLOBAL_PREFERENCE_NAME ] = $preference;
+
+		return true;
+	}
+
+	public static function onLinkerMakeExternalLink( string $url, string $text, string &$link, array $attribs ): bool {
```
mszabo-wikia, line 128: seems like there's some logic in here, do you think it'd be worth to add a unit test?

extensions/wikia/CategoryPage3/CategoryPage3Hooks.class.php
rogatty: I think if there is a redirect but it's invalid, then it could be `null`.

extensions/wikia/CategoryExhibition/CategoryExhibitionPage.php
```diff
@@ -3,7 +3,31 @@
 /**
  * Custom category page showing exhibition of pages, subcategories and media in the category
  */
-class CategoryExhibitionPage extends CategoryPageII {
+class CategoryExhibitionPage extends CategoryPageWithLayoutSelector {
+	public function openShowCategory() {
+		parent::openShowCategory();
+
+		global $wgOut, $wgExtensionsPath, $wgJsMimeType, $wgTitle, $wgRequest, $wgUser;
```
Grunny:
```
$context = $this->getContext();
$context->getOutput();
$context->getTitle();
$context->getRequest();
$context->getUser();
```
^ would be preferred over the globals.

extensions/wikia/CategoryPage3/CategoryPage3.class.php
```diff
@@ -0,0 +1,194 @@
...
+		// Keep 4:3 ratio
+		$height = floor( $width / 4 * 3 );
+
+		try {
+			$url = VignetteRequest::fromUrl( $url )
+				->topCrop()
+				->width( $width )
+				->height( $height )
+				->url();
+		} catch ( Exception $e ) {
```
Grunny, line 165: It looks like it only explicitly throws `InvalidArgumentException`, so being more specific here might be good to avoid potentially hiding other exceptions that we would want to know happened.

extensions/wikia/CategoryPage3/CategoryPage3Model.class.php
```diff
@@ -0,0 +1,320 @@
...
+		$pageIds = array_keys(
+			array_filter(
+				$this->members,
+				function ( $member ) {
+					/** @var CategoryPage3Member $member */
+					return !$member->isSubcategory();
+				}
+			)
+		);
+		$imageServing = new ImageServing( $pageIds, $thumbWidth, array( 'w' => $thumbWidth, 'h' => $thumbHeight ) );
```
Grunny: Minor: should use short array syntax per coding conventions.

extensions/wikia/CategoryPage3/CategoryPage3TrendingPages.class.php
```diff
@@ -0,0 +1,82 @@
...
+		$requiredParams = [ 'title', 'thumbnail', 'url' ];
+
+		$processedItem = [];
+
+		if ( !empty( $item ) && is_array( $item ) ) {
+			foreach ( $requiredParams as $param ) {
+				if ( !empty( $item[ $param ] ) ) {
+					$processedItem[ $param ] = $item[ $param ];
+				} else {
+					return null;
```
Grunny: Minor: is there any benefit to returning null rather than an empty array here? Would allow us to be strict about the return type rather than it being mixed/nullable, and the `empty` checks will still work.

extensions/wikia/CategoryPage3/styles/category-layout-selector.scss
```diff
@@ -0,0 +1,39 @@
...
+			cursor: pointer;
+			padding: 0 13px;
+
+			&.is-active {
+				color: $color-buttons;
+				cursor: default;
+			}
+
+			&:last-child {
+				border-right: none;
```
Grunny: Minor: I believe our [conventions](https://github.com/Wikia/guidelines/blob/47d00a7/SCSS/CodingConventions.md#borderzero) prefer `border-right: 0;` here.

extensions/wikia/CategoryPage3/styles/category-page3.scss
```diff
@@ -0,0 +1,195 @@
...
+		}
+
+		&__trending-pages {
+			// Grid is totally broken in IE11, -ms-inline-flexbox is better
+			display: -ms-inline-flexbox;
+			-ms-flex-wrap: wrap;
+
+			border-bottom: 1px solid $color-page-border;
+			display: grid;
+			grid-gap: 20px;
```
Grunny, line 28: Does `grid-gap` work OK in Safari: https://developer.mozilla.org/en-US/docs/Web/CSS/gap#Browser_compatibility ?

extensions/wikia/CategoryPage3/styles/category-page3.scss
```diff
@@ -0,0 +1,195 @@
...
+			font-weight: $wds-font-weight-bold;
+			margin-bottom: 30px;
+			margin-top: 40px;
+			text-transform: uppercase;
+		}
+
+		&__alphabet-shortcuts {
+			font-weight: $wds-font-weight-bold;
+			list-style: none;
+			margin: 0 0 35px 0;
```
Grunny: Could be `0 0 35px` I think.

extensions/wikia/CategoryPage3/templates/members-header.php
```diff
@@ -0,0 +1,15 @@
+<p class="category-page__total-number">
+	<?php /** @var $totalNumberOfMembers int */ ?>
+	<?= wfMessage( 'category-page3-total-number', $totalNumberOfMembers )->escaped() ?>
+</p>
+<ul class="category-page__alphabet-shortcuts">
+	<?php foreach ( $alphabetShortcuts as $shortcut ) : ?>
+		<li class="category-page__alphabet-shortcut<?= $shortcut['isActive'] ? ' is-active': '' ?>">
+			<?= Html::rawElement( 'a', [
```
Grunny: Can we use  `element` rather than `rawElement` to make sure it's escaped at the output?

extensions/wikia/CategoryPage3/templates/pagination.php
```diff
@@ -0,0 +1,27 @@
...
+		<?php endif; ?>
+		<?php if ( !empty( $pagination->getNextPageUrl() ) ): ?>
+            <a href="<?= $pagination->getNextPageUrl() ?>" class="category-page__pagination-next wds-button wds-is-secondary">
+				<span><?= wfMessage( 'category-page3-pagination-next' )->escaped() ?></span>
+				<?= DesignSystemHelper::renderSvg( 'wds-icons-menu-control-tiny', 'wds-icon wds-icon-tiny' ); ?>
+            </a>
+		<?php endif; ?>
+		<?php if ( !empty( $pagination->getLastPageUrl() ) ): ?>
+            <a href="<?= $pagination->getLastPageUrl() ?>" class="wds-button wds-is-text">
+				<?= wfMessage( 'category-page3-pagination-last' )->escaped() ?>
```
Grunny, line 25: Minor: whitespace seems very inconsistent in this file.

extensions/wikia/CategoryPage3/templates/trending-pages.php
```diff
@@ -0,0 +1,18 @@
+<div class="category-page__trending-pages-header"><?= wfMessage( 'category-page3-trending-pages' ) ?></div>
```
Grunny: The default output format of `wfMessage` is `parse`. Since this message does not contain wikitext, to save using the parser, I'd explicitly use `->escaped()` here.

extensions/wikia/CategoryPage3/CategoryPage3.class.php
```diff
@@ -0,0 +1,194 @@
...
+		$this->from = $request->getVal( 'from' );
+
+		$this->model = new CategoryPage3Model( $context->getTitle(), $this->from );
+		$this->model->loadData();
+
+		if ( $this->model->getTotalNumberOfMembers() === 0 ) {
+			return;
+		}
+
+		$this->model->loadImages( 40, 30 );
```
themech: please move these number to some consts. It will be easier to read the code, as `40, 30` doesn't tell me anything

extensions/wikia/CategoryExhibition/templates/form.tmpl.php
```diff
@@ -1,40 +1,25 @@
...
-			if ( $current == $sortType ) {
-				$el["class"] = "selected";
-			}
-			$el["href"] = "$path?sort=$sortType&display=$displayType";
-			$el["id"] = Sanitizer::escapeId( "category-exhibition-form-$sortType" );
-			$el["text"] = wfMessage( "category-exhibition-$sortType" )->escaped();
-			$dropdown[] = $el;
+	<?= wfMessage( 'category-exhibition-sorttype' )->escaped(); ?>
+	<?php
+	$dropdown = array();
```
themech: just a note: some of the code in this PR uses the `[]` syntax while there are still some places that use `array()`

extensions/wikia/CategoryPage3/CategoryPage3CacheHelper.class.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+class CategoryPage3CacheHelper {
+	const EXPIRATION_TIME = 60 * 60;
```
themech: maybe something like `EXPIRATION_TIME_SEC`?

extensions/wikia/CategoryPage3/CategoryPage3CacheHelper.class.php
```diff
@@ -0,0 +1,34 @@
...
+	public static function getTouched( Title $title ) {
+		global $wgMemc;
+		return $wgMemc->get( static::getKey( $title ) );
+	}
+
+	public static function setTouched( Title $title ) {
+		global $wgMemc;
+		$wgMemc->set(
+			static::getKey( $title ),
+			time() . rand( 0, 9999 ),
```
themech: :)
I know, you just copied it from `CategoryExhibitionCacheHelper`

extensions/wikia/CategoryPage3/CategoryPage3Hooks.class.php
```diff
@@ -0,0 +1,221 @@
...
+		'subcatuntil',
+	];
+
+	/**
+	 * @param $categoryInserts
+	 * @param $categoryDeletes
+	 * @return bool
+	 * @throws MWException
+	 */
+	public static function onAfterCategoriesUpdate( $categoryInserts, $categoryDeletes/*, $title*/ ): bool {
```
themech, line 26: do we need this commented-out param?

extensions/wikia/CategoryPage3/CategoryPageWithLayoutSelector.class.php
```diff
@@ -0,0 +1,53 @@
...
+				'currentLayout' => $this->getCurrentLayout()
+			] )
+			->render( 'extensions/wikia/CategoryPage3/templates/layout-selector.php' );
+	}
+
+	/**
+	 * @throws Exception
+	 */
+	protected function getCurrentLayout() {
+		throw new Exception( 'getCurrentLayout method needs to be overriden' );
```
themech: So basically you want to define `CategoryPageWithLayoutSelector` as an abstract class and `getCurrentLayout` as an abstract method that has to be overriden, right?

extensions/wikia/DesignSystem/node_modules/design-system/dist/css/styles.css
```diff
@@ -1 +1 @@
...global-navigation__modal-control-close svg,.wds-global-navigation__modal-control:hover svg{fill:#fff}.wds-global-navigation__modal-control:hover .wds-avatar__image{border-color:#fff;fill:#fff}.wds-global-navigation__links{display:none;font-size:18px;font-weight:700;height:100%}@media only screen and (min-width:768px){.wds-global-navigation__links{display:flex}}@media only screen and (max-width:1279px){.wds-global-navigation__links{font-size:16px}}.wds-global-navigation__link{text-transform:uppercase;transition:color .3s,font-size .2s,margin-left .2s}.wds-global-navigation__link,.wds-global-navigation__start-a-wiki{align-items:center;color:#00d6d6;display:flex;height:100%;margin-left:30px;text-decoration:none}.wds-global-navigation__link:visited,.wds-global-navigation__start-a-wiki:visited{color:#00d6d6!important}.wds-global-navigation__link:hover,.wds-global-navigation__start-a-wiki:hover{color:#fff!important}.wds-global-navigation__link .wds-dropdown__toggle-chevron,.wds-global-navigation__start-a-wiki .wds-dropdown__toggle-chevron{fill:#ff6a64}.wds-global-navigation__link .wds-global-navigation__link-button,.wds-global-navigation__start-a-wiki .wds-global-navigation__link-button{
```
themech: newline

extensions/wikia/MercuryApi/models/MercuryApiCategoryModel.class.php
```diff
@@ -1,5 +1,10 @@
 <?php
 
+/**
+ * TODO Remove after SEO-670 is released on mobile-wiki and no cached JS tries to use it anymore
+ * @deprecated
+ * Class MercuryApiCategoryModel
```
themech, line 6: 👍 +1

extensions/wikia/MercuryApi/handlers/MercuryApiMainPageHandler.class.php
```diff
@@ -7,7 +7,11 @@ class MercuryApiMainPageHandler {
 	public static function getMainPageData( MercuryApi $mercuryApiModel ) {
 		$mainPageData = [ ];
 		$curatedContent = self::getCuratedContentData( $mercuryApiModel, null );
-		$trendingArticles = $mercuryApiModel->getTrendingArticlesData();
+		$trendingArticles = $mercuryApiModel->getTrendingPagesData(
+			10,
+			null,
+			false
```
themech, line 13: it's not really obvious what these three values mean

extensions/wikia/CategoryPage3/CategoryPage3.class.php
themech: same `loadImages( 40, 30 )` is used in another place, that makes me even more convinced that these two values should be moved to consts

extensions/wikia/CategoryPage3/CategoryPage3.class.php
rogatty: `->url()` throws `\Exception`:
https://github.com/Wikia/app/blob/6c60bc14c47ed41d2d72683c137e3c1f9acbaf27/lib/Wikia/src/Vignette/UrlGenerator.php#L283-L285

extensions/wikia/CategoryPage3/styles/category-page3.scss
rogatty: From what I saw on @themech's Mac, it's fine. The standard property name is `gap` but all browsers I checked support `grid-gap` and Safari supports only `grid-gap`. That's why I used it here.

extensions/wikia/CategoryPage3/CategoryPage3.class.php
rogatty: screenshot from 2018-11-02 13-27-12
Isn't that redundant? ;)

extensions/wikia/CategoryPage3/CategoryPage3Hooks.class.php
rogatty: I think that's a convention (one of many we have), in case someone wanted to do something with this param in the future.

extensions/wikia/DesignSystem/node_modules/design-system/dist/css/styles.css
rogatty: This file is installed with npm, I'm not touching it.

extensions/wikia/MercuryApi/handlers/MercuryApiMainPageHandler.class.php
rogatty: screenshot from 2018-11-02 13-38-44

screenshot from 2018-11-02 13-39-38

Grunny (approved):

extensions/wikia/CategoryPage3/CategoryPage3Model.class.php
mszabo-wikia: 200 is a lot

rogatty: Yup, that's why it's cached:
https://github.com/Wikia/app/blob/8504a5e27ec7cb65c6824920fe703020dad2749b/extensions/wikia/CategoryPage3/CategoryPage3Model.class.php#L242-L248

mszabo-wikia: 👍

PR #16149, opened 2018-10-04
SRE-109 | Improve edit performance

mszabo-wikia:
* Moved stuff that can be deferred to task queue
* ImageServing index refresh
  * edit count++
  * site stats++
* Made `PortableInfoboxDataService` less wasteful by only purging cache on edit—`LinksUpdate` will make sure that we update page props with the new value
* Made `FounderEmailsEditEvent` not query master after edit to determine user edit count. We can simply use `UserStatsService` instead.
* Replaced calls to `User::invalidateCache` with `User::touch` in places where it was only used to bust caches to ensure we serve a properly updated UI (see [r59993](https://www.mediawiki.org/wiki/Special:Code/MediaWiki/59993#c4885)). `User::invalidateCache` has expensive side-effects that we do not need.
* Moved ThemeDesigner hook that invalidates settings when a new background is uploaded to a proper place—`UploadComplete` hook. Previously it'd fire for every edit and would fetch a revision from master.
* Optimized `PageStatsService` to not read from master but use the passed `Title` object properly
* Deferred `LinksUpdate` until the end of the request and made it read existing page info from the replica. There's no need to use a master for those reads. Also made it skip reads altogether if we're creating a page because the new page will not have any existing information about it yet.
* Passed an explicit `parent_id` of 0 to `Revision` when we're creating a page. Otherwise `Revision` would go looking for the parent on the master & would never find anything (since it'd be the first revision)
* Made `EmailNotification` get the list of users to notify from a replica
* Made `WikiPage::updateRevisionOn` eschew a `DELETE` if the previous revision is a redirect if we're creating a page (since there can be no previous revisions on that case)
* Made `Title::checkUserBlock` read user block from a replica
* Removed implicit transactions from external store cluster (see also: https://github.com/wikimedia/operations-mediawiki-config/blob/master/wmf-config/db-codfw.php#L694)


extensions/wikia/FounderEmails/events/FounderEmailsEditEvent.class.php
```diff
@@ -318,23 +318,15 @@ public static function recentChanges( $oRecentChange ) {
 			$wasNotificationSent = ( static::getFirstEmailSentFlag( $editor->getName() ) === '1' ) ;
 
 			if ( !$wasNotificationSent ) {
-				$userEditStatus = static::getUserEditsStatus( $editor, true );
```
mszabo-wikia, line 321: This method would try to calculate edit count of user from master DB. However, we don't care about the actual edit count—we just want to know if it is 0. If it's 0, then it means the user is making their first edit now. So we can use the existing `UserStatsService` to give us that data, which doesn't involve reading from master.

extensions/wikia/ImageServing/imageServingHelper.class.php
```diff
@@ -105,75 +107,4 @@ public static function getPlaceholder($file) {
...
-	/**
-	 * buildIndex - save image index in db
-	 *
-	 * @param int $articleId article ID
-	 * @param array|string $images
-	 * @param $ignoreEmpty boolean
-	 * @param bool $dryRun don't store results in DB (think twice before passing true, used by imageServing.php maintenance script)
-	 * @return mixed|bool set of images extracted from given article
-	 */
-	private static function buildIndex( $articleId, $images, $ignoreEmpty = false, $dryRun = false ) {
```
mszabo-wikia, line 118: dryRun parameter was never used, and referred maintenance script does not exist

extensions/wikia/ImageServing/imageServingHelper.class.php
```diff
@@ -9,25 +9,27 @@ class ImageServingHelper {
...
+		$task->title( $title );
+
+		// SRE-109: Don't trigger a re-parse if the article has only one image or no images at all
+		if ( count( $images ) <= 1 ) {
+			$task->call( 'createIndexFromImages', array_keys( $images ) );
+		} else {
+			$task->call( 'createIndexFromPageContent' );
+		}
+
+		$task->queue();
```
mszabo-wikia, line 31: The main change here is deferring the ImageServing index update into a background task.

extensions/wikia/PortableInfobox/PortableInfobox.setup.php
```diff
@@ -81,7 +81,7 @@
 $wgHooks[ 'wgQueryPages' ][] = 'PortableInfoboxHooks::onWgQueryPages';
 $wgHooks[ 'AllInfoboxesQueryRecached' ][] = 'PortableInfoboxHooks::onAllInfoboxesQueryRecached';
 $wgHooks[ 'ArticlePurge' ][] = 'PortableInfoboxHooks::onArticlePurge';
-$wgHooks[ 'ArticleSave' ][] = 'PortableInfoboxHooks::onArticleSave';
+$wgHooks[ 'ArticleSaveComplete' ][] = 'PortableInfoboxHooks::onArticleSaveComplete';
```
mszabo-wikia, line 84: We only want to run this hook for successful edit operations.

extensions/wikia/ThemeDesigner/ThemeDesignerHooks.class.php
```diff
@@ -43,6 +29,8 @@ public static function onUploadComplete( $image ) {
 			Wikia::invalidateFavicon();
 		}
 
+		self::resetThemeBackgroundSettings( $image->getTitle(), false );
```
mszabo-wikia, line 32: The previous hook handler would be called for every edit and would try to read title from master. But we only care about image uploads to the background image controlled by theme designer.

includes/EmailNotification.php
```diff
@@ -185,8 +185,9 @@ private function getTimeOutSql() {
 
 	private function getWatchersToNotify( $notificationTimeoutSql ) {
 		$watchers = [];
-		$dbw = wfGetDB( DB_MASTER );
-		$res = $dbw->select(
+		// SRE-109: Get the set of users to notify from the slave
+		$dbr = wfGetDB( DB_SLAVE );
```
mszabo-wikia, line 189: There's no need to read this from master.

includes/LinksUpdate.php
```diff
@@ -22,7 +22,7 @@
 
 use Wikia\Tasks\Tasks\BatchRefreshLinksForTemplate;
 
-class LinksUpdate {
+class LinksUpdate implements DeferrableUpdate {
```
mszabo-wikia, line 25: This will allow us to defer the update to the end of the request, so in FastCGI it won't block serving a response to the user.

extensions/wikia/AchievementsII/services/AchAwardingService.class.php
```diff
@@ -289,7 +289,8 @@ private function saveBadges() {
 			}
 
 			//touch user when badges are given
-			$this->mUser->invalidateCache();
+			// SRE-109: Use touch() to avoid needless DB queries; it's sufficient as per r59993
+			$this->mUser->touch();
```
mszabo-wikia, line 293: See: https://www.mediawiki.org/wiki/Special:Code/MediaWiki/59993#c4885

includes/wikia/services/PageStatsService.class.php
```diff
@@ -88,40 +55,25 @@ static function onArticleSaveComplete(
...
-
-		wfProfileIn(__METHOD__);
-
-		wfDebug(__METHOD__ . ": page #{$this->pageId}\n");
-
-		// invalidate cached data from getCurrentRevision()
-		$wgMemc->delete($this->getKey('current-revision'));
-
-		// invalidate cached data from getPreviousEdits()
-		$wgMemc->delete($this->getKey('previous-edits'));
```
mszabo-wikia: The referred methods do not exist in code.

includes/WikiPage.php
```diff
@@ -1327,12 +1333,6 @@ public function doEdit( $text, $summary, $flags = 0, $baseRevId = false, $user =
 			$summary = self::getAutosummary( $oldtext, $text, $flags );
 		}
 
-		// <Wikia>
-		if ( is_string( $user ) ) {
-			error_log( "MOLI: " . __METHOD__ . ": invalid User : " . print_r( $user, true ) );
-			Wikia::debugBacktrace( "MOLI: invalid User:" );
```
pwojcik86, line 1333: nice.

includes/WikiPage.php
```diff
@@ -1659,8 +1663,11 @@ public function doEditUpdates( Revision $revision, User $user, array $options =
 		}
 
 		# Update the links tables
-		$u = new LinksUpdate( $this->mTitle, $editInfo->output );
-		$u->doUpdate();
+		// SRE-109: tell LinksUpdate if this is a new page and defer it till the end of the request
+		$linksUpdate = new LinksUpdate( $this->mTitle, $editInfo->output );
+		$linksUpdate->setWasPageNewlyCreated( $options['created'] );
+
+		DeferredUpdates::addUpdate( $linksUpdate );
```
pwojcik86, line 1670: Are you sure that subsequent operations don't rely on the result of this one?

pwojcik86 (approved):
Looks ok to me, but I'm afraid I don't understand all the details.

includes/WikiPage.php
mszabo-wikia: Yeah, if something was relying on that, chances are they'd have used the `LinksUpdateComplete` hook which will still work just a same.

ghost (approved):

mszabo-wikia:
I'll split this up into smaller changes.

PR #16163, opened 2018-10-05
Also depend on mediawiki.jqueryMsg to make sure we parse links

Grunny:
/cc @rogatty  @mszabo-wikia


Grunny:
Reviewed in https://github.com/Wikia/app/pull/16162

PR #16182, opened 2018-10-11
SRE-113 | Fix MW celery purger

mszabo-wikia:
Ensure tasks are published at the right time and add tests around this flow.

https://wikia-inc.atlassian.net/browse/SRE-113


macbre:
Some context will be useful here.

mszabo-wikia:
@macbre sure :) the main issue was that because of the `TaskPublisher` refactor, the `onRestInPeace` hook handler was invoked too late, after the tasks have already been queued. So we need a way to ensure these tasks are properly sent to the publisher before it flushes all tasks. That was the object of this change, along with improving test coverage.

Grunny (approved):

PR #16186, opened 2018-10-11
ADEN-7733 autoplay info in events 

pnather:


extensions/wikia/AdEngine/js/video/player/playerTracker.js
```diff
@@ -64,7 +64,8 @@ define('ext.wikia.adEngine.video.player.playerTracker', [
 				'additional_2': floatingState,
 				'vast_id': params.vastId || emptyValue.string,
 				'video_id': params.videoId || '',
-				'btl': billTheLizard && billTheLizard.hasResponse() ? 1 : 0
+				'btl': billTheLizard && billTheLizard.hasResponse() ? 1 : 0,
+				'user_block_autoplay': params.withAudio
```
fraszczakszymon: We should send 1/0 instead of true/false. Look at lines L59-L60. It's already sent to DW. 

However - imo this solution doesn't cover the problem from ticket. We need to know whether ctp was triggered by user (by setting the cookie). Queen_of_hearts/ab test impacts `params.withAudio` so we can't rely on this flag to address this issue.

extensions/wikia/AdEngine/js/spec/video/playerTracker.spec.js
```diff
@@ -62,6 +67,7 @@ describe('ext.wikia.adEngine.video.player.playerTracker', function () {
 			mocks.adLogicPageParams,
 			mocks.adTracker,
 			mocks.slotTargeting,
+			mocks.autoplay,
```
jacekbj, line 70: Please add a test if it's really tracked.

extensions/wikia/AdEngine/js/video/player/playerTracker.js
```diff
@@ -64,7 +66,8 @@ define('ext.wikia.adEngine.video.player.playerTracker', [
 				'additional_2': floatingState,
 				'vast_id': params.vastId || emptyValue.string,
 				'video_id': params.videoId || '',
-				'btl': billTheLizard && billTheLizard.hasResponse() ? 1 : 0
+				'btl': billTheLizard && billTheLizard.hasResponse() ? 1 : 0,
+				'user_block_autoplay': !autoplay.isAutoplayEnabled() ? 1 : 0
```
fraszczakszymon, line 70: `isAutoplayEnabled()` still includes queen_of_hearts/ab_test decisions. Take a look at `featuredVideoCookieService` only

jacekbj: 💎

PR #16187, opened 2018-10-11
(SEO-674) Redirect to the first category page when old params are used

Grunny:
Redirect to the first category page when old pagination query params are used.

/cc @rogatty

### DIFF
--- a/extensions/wikia/CategoryPage3/CategoryPage3Hooks.class.php
+++ b/extensions/wikia/CategoryPage3/CategoryPage3Hooks.class.php
@@ -4,6 +4,18 @@ class CategoryPage3Hooks {
 
 	const COOKIE_NAME = 'category-page-layout';
 
+	private static $oldQueryParams = [
+		'display',
+		'filefrom',
+		'fileuntil',
+		'page',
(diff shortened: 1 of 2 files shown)


extensions/wikia/CategoryPage3/CategoryPage3Hooks.class.php
```diff
@@ -54,4 +66,39 @@ static public function onArticleFromTitle( $title, &$article ): bool {
...
+	 * @param WebRequest $request
+	 * @param MediaWiki $mediawiki
+	 * @return bool
+	 */
+	static public function onBeforeInitialize(
+		Title $title, $unused, OutputPage $output,
+		User $user, WebRequest $request, MediaWiki $mediawiki
+	): bool {
+		if ( !$title->inNamespace( NS_CATEGORY ) ||
+			$user->isLoggedIn()
```
rogatty, line 87: I think we should check the current layout in addition to that. There is no existing method that could be used here, so I'll update it as a part of the layout selector task.

rogatty (approved):

PR #16202, opened 2018-10-15
ADEN-7595-2 CTP Cookie

pnather:


jacekbj (approved):

extensions/wikia/AssetsManager/config.php
```diff
@@ -208,6 +208,7 @@
 		'//extensions/wikia/AdEngine/js/utils/math.js',
 		'//extensions/wikia/AdEngine/js/utils/time.js',
 		'//extensions/wikia/AdEngine/js/video/articleVideoAd.js',
+		'//extensions/wikia/ArticleVideo/scripts/featured-video.cookies.js',
```
fraszczakszymon: Do we need to add this asset on all pages? I believe it's already available on FV pages. We can use require.optional in playerTracker.js so it won't throw errors on non-FV pages

extensions/wikia/AssetsManager/config.php
```diff
@@ -899,6 +900,7 @@
 		'//extensions/wikia/AdEngine/js/video/videoFrequencyMonitor.js',
 		'//extensions/wikia/AdEngine/js/video/videoFrequencySettings.js',
 		'//extensions/wikia/AdEngine/js/video/videoFrequencyStore.js',
+		'//extensions/wikia/ArticleVideo/scripts/featured-video.cookies.js',
```
fraszczakszymon: Is it fine for mobile-wiki? There is a bit different implementation of jwplayer. Check how we set autoplay flag there please:
https://github.com/Wikia/mobile-wiki/blob/dev/app/modules/video-players/jwplayer.js

extensions/wikia/AdEngine/js/video/player/playerTracker.js
```diff
@@ -72,6 +72,10 @@ define('ext.wikia.adEngine.video.player.playerTracker', [
 			trackingData['vast_id'] = params.bid.creativeId || emptyValue.string;
 		}
 
+		if ([0, 1].indexOf(params.userBlockAutoplay) > -1) {
```
fraszczakszymon:
```suggestion
		if ([-1, 0, 1].indexOf(params.userBlockAutoplay) > -1) {
```

fraszczakszymon (approved):

PR #16207, opened 2018-10-15
(PLATFORM-3765) Use full URLs in in content links on language wikis

Grunny:
Use full URLs in in content links on language wikis so that the parser
cache has valid URLs when migrating those wikis.

/cc @Wikia/core-platform-team

### DIFF
--- a/includes/wikia/Extensions.php
+++ b/includes/wikia/Extensions.php
@@ -94,6 +94,13 @@ if ( ! empty( $wgEnableWikisApi ) ) {
 	F::app()->registerApiController( 'WikisApiController', "{$IP}/includes/wikia/api/WikisApiController.class.php" );
 }
 
+// During migration drop parser expiry on non-English wikis to 24 hours (PLATFORM-3765)
+if ( !wfHttpsAllowedForURL( $wgServer ) ||
+	( wfHttpsEnabledForURL( $wgServer ) && $wgLanguageCode !== 'en' )
+) {
+	$wgParserCacheExpireTime = 24 * 3600;
+}
+
 /*
  * Code for http://lyrics.wikia.com/
  */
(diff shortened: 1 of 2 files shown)


includes/wikia/Extensions.php
```diff
@@ -94,6 +94,13 @@
 	F::app()->registerApiController( 'WikisApiController', "{$IP}/includes/wikia/api/WikisApiController.class.php" );
 }
 
+// During migration drop parser expiry on non-English wikis to 24 hours (PLATFORM-3765)
+if ( !wfHttpsAllowedForURL( $wgServer ) ||
+	( wfHttpsEnabledForURL( $wgServer ) && $wgLanguageCode !== 'en' )
+) {
+	$wgParserCacheExpireTime = 24 * 3600;
```
rogatty, line 101: Is there a specific reason it's in `Extensions.php`? Can it be set [earlier](https://github.com/Wikia/config#whats-the-order-in-which-the-configuration-is-loaded)? It's just weird as it's not an extension.

includes/wikia/Extensions.php
```diff
@@ -94,6 +94,13 @@
 	F::app()->registerApiController( 'WikisApiController', "{$IP}/includes/wikia/api/WikisApiController.class.php" );
 }
 
+// During migration drop parser expiry on non-English wikis to 24 hours (PLATFORM-3765)
+if ( !wfHttpsAllowedForURL( $wgServer ) ||
+	( wfHttpsEnabledForURL( $wgServer ) && $wgLanguageCode !== 'en' )
```
rogatty, line 99: What's the reason to keep it that short after migration? Just in case we find something else?

includes/wikia/Extensions.php
Grunny: Because it needs to be after WikiFactory is loaded, otherwise wgServer and wgLanguageCode may not be set yet.

rogatty (approved):

PR #16209, opened 2018-10-16
SUS-6049 | point CreateNewWiki links to community.wikia.com

macbre:
https://wikia-inc.atlassian.net/browse/SUS-6049

W ewant to move away all MediaWiki-related things from www.wikia.com domain and make it being served by Wordpress only. Let's start with CreateNewWiki.

I've just enabled it on Community Wiki. Let's update the links. Once this goes live, we'll handle redirects from an old address and then disable the extension on www.wikia.com


includes/wikia/models/DesignSystemSharedLinks.class.php
```diff
@@ -88,7 +88,7 @@ private function getLangWithFallback( $lang ) {
 			'local-sitemap-fandom' => '//fandom.wikia.com/local-sitemap',
 			'community-central' => '//community.wikia.com/wiki/Community_Central',
 			'support' => '//community.wikia.com/wiki/Special:Contact',
-			'create-new-wiki' => '//www.wikia.com/Special:CreateNewWiki',
+			'create-new-wiki' => '//community.wikia.com/wiki/Special:CreateNewWiki',
```
rybmat, line 91: @macbre it needs to be changed for every language in this array as well

rybmat (changes requested):

rybmat (approved):

PR #16213, opened 2018-10-16
DE-3752 | sending edit events directly to Kinesis

ghost:


extensions/wikia/DataWarehouse/DataWarehouseEventProducer.class.php
```diff
@@ -375,10 +378,21 @@ public function setCategory() {
...
 
+	public function setBeacon( $beacon ) {
+		$this->mParams['beacon'] = $beacon;
+	}
+
 	public function sendLog() {
 		wfProfileIn( __METHOD__ );
 		$data = json_encode($this->mParams);
 		$this->getRabbit()->publish( $this->mKey, $data );
+		if ( ( ! Wikia::isDevEnv()) && ( $this->mParams['cityId'] % 100 == 28 ) ) { // enabled on 1% of wikis
```
macbre: Don't you want to use `<` operator instead and just increase the percentage of wikis as we go?

extensions/wikia/DataWarehouse/DataWarehouseEventProducer.class.php
```diff
@@ -375,10 +378,21 @@ public function setCategory() {
...
+			$this->mParams['action'] = $this->mKey;
+			$kinesis = new AsyncKinesisProducerTask();
+			( new AsyncTaskList() )
```
macbre: use `newLocalTask` static constructor instead. AFAIR you need to specify wiki ID when queuing a task. This constructor will use the ID of the current wiki.

extensions/wikia/DataWarehouse/DataWarehouseEventProducer.class.php
ghost: Good idea, I will use `<` indeed.

extensions/wikia/DataWarehouse/DataWarehouseEventProducer.class.php
ghost: OK, thanks!

jcellary (approved):

macbre (approved):

PR #16214, opened 2018-10-16
(PLATFORM-3765) Use full URLs in in content links on language wikis

Grunny:
/cc @Wikia/core-platform-team

PR #16215, opened 2018-10-16
(PLATFORM-3765) Limit number of wikis with lower parser cache expiry

Grunny:
Limit the parser cache expiry drop to wikis scheduled for migration.


Grunny:
Reviewed as part of https://github.com/Wikia/app/pull/16214

PR #16226, opened 2018-10-17
SRE-116 | Redo ProtectSite with a database backend

mszabo-wikia:
The current implementation of ProtectSite uses exclusively memcached as a "backend" so it will not work with multiple active datacenters. Since the extension is kind of a mess I've decided to start from scratch and rebuild it while using a database backend to store protection information.

The old extension will be removed separately, after the new one is live.

https://wikia-inc.atlassian.net/browse/SRE-116

### DIFF
--- a/extensions/wikia/ProtectSiteII/ProtectSiteSpecialController.php
+++ b/extensions/wikia/ProtectSiteII/ProtectSiteSpecialController.php
@@ -42,6 +42,12 @@ class ProtectSiteSpecialController extends WikiaSpecialPageController {
 		];
 
 		$inputs[] = [
+			'type' => 'text',
+			'label' => $this->msg( 'protectsite-label-reason' )->escaped(),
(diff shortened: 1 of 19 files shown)


extensions/wikia/ProtectSiteII/ProtectSiteSpecialController.php
```diff
@@ -0,0 +1,168 @@
...
+		$this->setHeaders();
+		$this->checkPermissions();
+		$this->checkIfUserIsBlocked();
+
+		$active = $this->model->getProtectionSettings( $this->wg->CityId );
+
+		$inputs = [];
+
+		foreach ( ProtectSiteModel::getValidActions() as $action ) {
+			// For grepping - possible message keys used here:
```
ghost, line 29: :heart:

ghost (approved):

PR #16232, opened 2018-10-18
DE-3752 | event structure cleanup, events on 10% of edits

ghost:


esthomw (approved):

extensions/wikia/DataWarehouse/DataWarehouseEventProducer.class.php
```diff
@@ -287,17 +283,13 @@ public function setGeoContinent ( $continent ) {
 	}
 
 	public function setCityId ( $city_id ) {
```
macbre: You can also use type hinting in method declaration and PHP will cast the parameter to int.

```suggestion
	public function setCityId ( int $city_id ) {
```

macbre (approved):

PR #16233, opened 2018-10-18
SER-2686 | Publish events when Wiki gets hard closed (removed)

slayful:


mszabo-wikia (approved):
Note that hook handlers are no longer required to return a boolean—returning nothing is fine as well.

tmszdmsk (approved):

PR #16234, opened 2018-10-18
IW-1070 | feeds module updates

latata:
@leshniak @xkxd 

* styles changes
* render feeds module in the recirculation container so it will be sticky as right rail recirculation module

### DIFF
--- a/extensions/wikia/FeedsAndPosts/scripts/index.js
+++ b/extensions/wikia/FeedsAndPosts/scripts/index.js
@@ -1,9 +1,13 @@
-require(['jquery'], function ($) {
+require(['jquery', 'wikia.tracker'], function ($, tracker) {
 	// Use number of hours passed since Jan. 1, 1970. That way cache is busted at most every hour.
 	var version = Math.floor((new Date()).getTime() / (60 * 60 * 1000));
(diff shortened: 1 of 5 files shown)


extensions/wikia/FeedsAndPosts/scripts/index.js
```diff
@@ -8,12 +8,13 @@ require(['jquery'], function ($) {
 			var wikiName = $('meta[property="og:site_name"]').prop('content');
 
 			// Load FP into a newly inserted element
-			$('#wikia-recent-activity').after('<div class="rail-module feed-posts-module"></div>');
+			$('#recirculation-rail').append('<div class="rail-module feed-posts-module"></div>');
 			var fpContainer = $('.feed-posts-module').get(0);
 			window.fp.default(fpContainer, {
```
xkxd: not related, but considering IDs as globals I'd say `fp` can be overridden via some id in custom markup much easier that sth like `feedsAndPosts`

xkxd (approved):

extensions/wikia/FeedsAndPosts/scripts/index.js
latata: I'd rather go with something like `nbgToAQTol` :D

PR #16241, opened 2018-10-19
PLATFORM-3761 & PLATFORM-3762: robots and wikis index page on empty domains

themech:
1. Serve robots.txt on domains containing active language wikis.
2. Render a special page that will contain a list of language wikis. This page is displayed:
 * on empty domain root, where there are language path wikis but the English wikis does not exist
 * on closed English wikis, where there are active language path wikis


extensions/wikia/LanguageWikisIndex/LanguageWikisIndexHooks.class.php
```diff
@@ -0,0 +1,88 @@
...
+				$context->setSkin( Skin::newFromKey( 'oasis' ) );
+				$wgSuppressCommunityHeader = true;
+				$wgSuppressPageHeader = true;
+
+				SpecialPageFactory::executePath( $wgTitle, $context );
+				$wgOut->output();
+				return true;
+			default:
+				if ( $redirectAll ) {
+					$wgRequest->response()->header( 'Location: ' . self::WIKIS_INDEX_PAGE , 302 );
```
Grunny: We should probably check with Becky about these being 302s. Even if the wiki is (re)created in the future, we may not do that any time soon and so 301s might be better to avoid junk sticking around in Google. Any thoughts @rogatty?

extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
```diff
@@ -374,6 +374,39 @@ public function execute() {
 						"db"     => $oRow->city_dbname,
 						"cluster" => $oRow->city_cluster,
 					);
+				} else {
+					// no city found, check if this is a language wikis index page
+					if ( empty( $this->mWikiID ) && $this->hasLanguagePathWikis() ) {
```
Grunny: Minor: `elseif`.

extensions/wikia/LanguageWikisIndex/LanguageWikisIndexHooks.class.php
rogatty: Can we tell if it's a closed wiki or not? For closed wikis I'd use 301, for not-yet-created wikis I think 302 is fine. After all, we're returning 200 (come back later) on https://community.wikia.com/wiki/Community_Central:Not_a_valid_community?from=szl.starwars.wikia.com, not 410 (never come back). But yeah, I'd double check with Becky.

extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
themech: 👍

Grunny (approved):

PR #16242, opened 2018-10-19
SRE-117 | Add process cache to MercuryApi templates processing

mszabo-wikia:
When MercuryApi is processing an article that transcludes the same template many times (e.g. an icon template used several times, like the ones in the table headers on https://fallout.wikia.com/wiki/Fallout_4_characters), it repeatedly fetches template classification data and media details for the same template. We can leverage process cache here to improve performance.

@Wikia/iwing 

https://wikia-inc.atlassian.net/browse/SRE-117

### DIFF
--- a/includes/wikia/parser/templatetypes/handlers/InfoIconTemplate.class.php
+++ b/includes/wikia/parser/templatetypes/handlers/InfoIconTemplate.class.php
@@ -40,21 +40,28 @@ class InfoIconTemplate {
 	private static function getIconLink($image ) {
 		// SRE-117: Use process cache to avoid repeated expensive media queries
 		if ( !isset( self::$infoIconCache[$image] ) ) {
-			$title = Title::newFromText( $image );
+			self::$infoIconCache[$image] = self::makeIconLink( $image );
+		}
+
+		return self::$infoIconCache[$image];
+	}
+
+	private static function makeIconLink( $image ): string {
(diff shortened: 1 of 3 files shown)


includes/wikia/parser/templatetypes/handlers/InfoIconTemplate.class.php
```diff
@@ -33,19 +37,24 @@ public static function handle( $wikitext, Parser $parser ) {
...
-			$link = Linker::makeImageLink2(
-				$title,
-				$file,
-				[],
-				[ 'template-type' => TemplateClassificationService::TEMPLATE_INFOICON ]
-			);
+	private static function getIconLink($image ) {
+		// SRE-117: Use process cache to avoid repeated expensive media queries
+		if ( !isset( self::$infoIconCache[$image] ) ) {
+			$title = Title::newFromText( $image );
```
hakubo: while at it - Title newFromText can return null, should we guard against it here?

mszabo-wikia: hmm, good point. I could add a check for safety, although I don't recall seeing errors from here.

hakubo (approved):

rybmat (approved):

rybmat (approved):

PR #16252, opened 2018-10-23
SUS-6081 | htmlescape serchFormId in Search_Index.php template

rybmat:
https://wikia-inc.atlassian.net/browse/SUS-6081

### DIFF
--- a/skins/oasis/modules/templates/Search_Index.php
+++ b/skins/oasis/modules/templates/Search_Index.php
@@ -1,14 +1,14 @@
 <?php $searchterm = isset($searchterm) ? $searchterm : ''; ?>
 <?php $searchFormId = isset($searchFormId) ? $searchFormId : 'WikiaSearch'; ?>
 
-<form id="<?= $searchFormId; ?>" class="WikiaSearch<?= empty($noautocomplete) ? '' : ' noautocomplete' ?>" action="<?= $specialSearchUrl; ?>" method="get">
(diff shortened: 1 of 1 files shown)


skins/oasis/modules/templates/Search_Index.php
```diff
@@ -1,7 +1,7 @@
 <?php $searchterm = isset($searchterm) ? $searchterm : ''; ?>
 <?php $searchFormId = isset($searchFormId) ? $searchFormId : 'WikiaSearch'; ?>
 
-<form id="<?= $searchFormId; ?>" class="WikiaSearch<?= empty($noautocomplete) ? '' : ' noautocomplete' ?>" action="<?= $specialSearchUrl; ?>" method="get">
+<form id="<?= htmlspecialchars($searchFormId); ?>" class="WikiaSearch<?= empty($noautocomplete) ? '' : ' noautocomplete' ?>" action="<?= $specialSearchUrl; ?>" method="get">
```
Grunny: Since this is an attribute, `Sanitizer::encodeAttribute` would be preferred.

ghost (approved):

Grunny (changes requested):
For good measure/paranoia, I would also consider wrapping the other vars that are not currently escaped in `Sanitizer::encodeAttribute` (i.e. `$placeholder` which is the output of a message that is not escaped (but should be), the output of `$searchParams` (which right now holds only validated numbers, but what if that changes?), `$fulltext`, etc.). The best place to escape is in the template at output, and makes it safer against any implementation changes in the future.

Grunny (approved):

PR #16257, opened 2018-10-24
(PLATFORM-3769) Bump cache of ArticlesApi responses on lang path wikis

Grunny:
Bump cache of ArticlesApi responses on lang path wikis to that the local URLs
are updated immediately after migrating communities.


Grunny:
Reviewed in https://github.com/Wikia/app/pull/16256.

PR #16259, opened 2018-10-24
ADEN-7773 Bump ad-engine version and fix wikia adapter after moving prebid to ad-engine

nandy-andy:


fraszczakszymon (approved):

extensions/wikia/AdEngine/src/bidders.js
```diff
@@ -405,7 +405,10 @@ export function getBiddersContext(skin) {
 						enabled: false,
 						slots: {
 							FEATURED: {},
```
fraszczakszymon: What about `FEATURED`?

nandy-andy: When I add the properties we've added I can notice the bid goes on `SyntheticTests/Premium/FeaturedVideo?wikia_video_adapter=2000&&pbjs_debug=1` but there is no cow ad - as we discussed we probably need to change targeting of blue ad :)

fraszczakszymon (approved):

dmnsobczak (approved):

PR #16263, opened 2018-10-25
ADEN-7814 Disable INVISIBLE_SKIN call on non-Roadblock pages

dmnsobczak:


extensions/wikia/AdEngine/js/provider/gpt/sraHelper.js
```diff
@@ -10,8 +10,15 @@ define('ext.wikia.adEngine.provider.gpt.sraHelper', [
 		sraSlots = [
 			'INVISIBLE_SKIN',
 			'TOP_LEADERBOARD'
+		],
+		optionalSlots = [
+			'INVISIBLE_SKIN'
 		];
 
+	function isOptional(slotName) {
```
jacekbj: Please add a description of what it means that a slot is optional.

dmnsobczak: Those are the slots used only in UAP:Roadblock template. I renamed some parts of code to make it clear.

fraszczakszymon:
What would happen if we simple disable `INVISIBLE_SKIN` in [slotsContext](https://github.com/Wikia/app/blob/dev/extensions/wikia/AdEngine/js/context/slotsContext.js#L38)? Wouldn't it be disabled by default? Roadblock is going to re-enable this slot for its purposes again

dmnsobczak:
No, it's not gonna disable slot - we still have INVISIBLE_SKIN requests on non-Roadblock pages.

fraszczakszymon (approved):

PR #16272, opened 2018-10-25
IW-1152 | "Just updated" for Community Feeds

latata:
https://wikia-inc.atlassian.net/browse/IW-1152

* returns 4 recent updated articles from content namespace with image urls
* results cached for 60 minutes

@Wikia/iwing


extensions/wikia/FeedsAndPosts/FeedsAndPosts.setup.php
```diff
@@ -4,9 +4,19 @@
  * Injects discussion feed into an article page
  */
 
+$dir = dirname( __FILE__ ) . '/';
+
 // Autoload
 $wgAutoloadClasses['FeedsAndPostsHooks'] =  __DIR__ . '/FeedsAndPostsHooks.class.php';
```
jsutterfield: Could we update this line so it's using `$dir` as well? That way we're not doing the same thing in 2 different ways.

extensions/wikia/FeedsAndPosts/JustUpdatedAPIProxy.class.php
```diff
@@ -0,0 +1,65 @@
+<?php
+
+namespace Wikia\FeedsAndPosts;
+
+class JustUpdatedAPIProxy {
+
+	const LIMIT = 4;
+
+	private function requestAPI($params) {
```
jsutterfield: Our [PHP guidelines](https://github.com/Wikia/guidelines/blob/master/PHP/CodingConventions.md) are a superset of the MediaWiki guidelines, and those call for spaces inside of all of those parentheses: https://www.mediawiki.org/wiki/Manual:Coding_conventions/PHP#Spaces. So we should probably update those.

jsutterfield (approved):
2 super minor comments, apart from that LGTM!

extensions/wikia/FeedsAndPosts/FeedsAndPostsController.class.php
```diff
@@ -0,0 +1,11 @@
...
+
+use Wikia\FeedsAndPosts\JustUpdatedAPIProxy;
+
+class FeedsAndPostsController extends WikiaController {
+	public function getJustUpdated() {
+		$this->response->setFormat( WikiaResponse::FORMAT_JSON );
+		$this->response->setValues( ( new JustUpdatedAPIProxy() )->get() );
+
+	}
+}
```
rybmat: newline at the end of file is missing

extensions/wikia/FeedsAndPosts/JustUpdatedAPIProxy.class.php
```diff
@@ -0,0 +1,65 @@
...
+
+				foreach($recentChanges as &$article) {
+					$article['image'] = $this->getImage($article['title']);
+				}
+
+				return $recentChanges;
+			},
+			[
+				'command' => \WikiaDataAccess::USE_CACHE,
+				'cacheTTL' => \WikiaResponse::CACHE_VERY_SHORT,
```
rybmat: according to ticket it can be cached up to an hour, why do we use 5 min instead?

extensions/wikia/FeedsAndPosts/JustUpdatedAPIProxy.class.php
```diff
@@ -0,0 +1,65 @@
...
+		return $api->GetResultData();
+	}
+
+	private function getRecentChanges() {
+		$response = $this->requestAPI( [
+			'action' => 'query',
+			'list' => 'recentchanges',
+			'rcprop' => 'title',
+			'rcshow' => '!bot|!minor|!redirect',
+			'rcnamespace' => '0',
```
rybmat: please use $wgContentNamespaces instead of hardcoded 0, some wikis have their custom namespaces that are used as regular articles. To get results for multiple namespaces just separate them with `|`

extensions/wikia/FeedsAndPosts/JustUpdatedAPIProxy.class.php
```diff
@@ -0,0 +1,65 @@
...
+		$api->execute();
+
+		return $api->GetResultData();
+	}
+
+	private function getRecentChanges() {
+		$response = $this->requestAPI( [
+			'action' => 'query',
+			'list' => 'recentchanges',
+			'rcprop' => 'title',
```
rybmat: what about this requirement? `Nice-to-have: Filtered by those with a before and after character count data of X`
To do that just add `|size` here. The size represents number of bytes changed (can be negative for deletions)

includes/wikia/VariablesBase.php
```diff
@@ -2664,7 +2664,14 @@
  * @see extensions/wikia/FeedsAndPosts
  * @var bool $wgEnableFeedsAndPostsExt
  */
-$wgEnableFeedsAndPostsExt = false;
+$wgEnableFeedsAndPostsExt = true;
```
rybmat, line 2675: does it mean that fepo will be enabled on all wikis by default?

extensions/wikia/FeedsAndPosts/FeedsAndPostsController.class.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+use Wikia\FeedsAndPosts\JustUpdatedAPIProxy;
+
+class FeedsAndPostsController extends WikiaController {
+	public function getJustUpdated() {
```
rybmat: where this data is used? Is there any template? or is it called from other front-end apps?

extensions/wikia/FeedsAndPosts/JustUpdatedAPIProxy.class.php
```diff
@@ -0,0 +1,65 @@
...
+
+		if ( isset( $response['image']['imageserving'] ) ) {
+			return $response['image']['imageserving'];
+		}
+
+		return null;
+	}
+
+	public function get() {
+		return \WikiaDataAccess::cacheWithOptions(
```
rybmat: instead of using `cacheWithOptions` you can use `WikiaDataAccess::cache(wfMemcKey( 'feeds-just-updated' ), WikiaDataAccess::CACHE_VERY_SHORT, function(){...});` the command is `USE_CACHE` by default

rybmat (changes requested):

rybmat:
@latata one of the unit tests fails

extensions/wikia/MercuryApi/models/MercuryApi.class.php
```diff
@@ -261,7 +265,8 @@ public function getWikiVariables() {
 				'enableDiscussions' => $wgEnableDiscussions,
 				'enableDiscussionsImageUpload' => $wgEnableDiscussionsImageUpload,
 				'enableFandomAppSmartBanner' => !empty( $enableFAsmartBannerCommunity ),
-				'enableFeedsAndPosts' => $wgEnableFeedsAndPostsExt,
+				'enableFeedsAndPostsEmbeddedFeed' => $wgEnableFeedsAndPostsExt &&
```
rybmat: Does any of our front end apps use this endpoint? If not then I would remove this variable from here (as well as few others)

extensions/wikia/MercuryApi/models/MercuryApi.class.php
```diff
@@ -197,7 +200,8 @@ public function getDiscussionsWikiVariables() {
 				'discussionColorOverride' => SassUtil::sanitizeColor( $wgDiscussionColorOverride ),
 				'enableDiscussions' => $wgEnableDiscussions,
 				'enableDiscussionsImageUpload' => $wgEnableDiscussionsImageUpload,
-				'enableFeedsAndPosts' => $wgEnableFeedsAndPostsExt,
+				'enableFeedsAndPostsEmbeddedFeed' => $wgEnableFeedsAndPostsExt &&
```
rybmat: since you are renaming this variable, you should support both names for some time, because when mediawiki is released, mobile wiki will not be able to get value of this variable until it is released.

includes/wikia/VariablesBase.php
latata: oops. that was temporary for testing purpose. :D I'll revert it.

extensions/wikia/FeedsAndPosts/FeedsAndPostsController.class.php
latata: Fepo will use it

extensions/wikia/MercuryApi/models/MercuryApi.class.php
latata: Whatever, it's not on prod yet ;) In the worst scenerio embedded feeds won't be visible for some time.

extensions/wikia/FeedsAndPosts/JustUpdatedAPIProxy.class.php
```diff
@@ -0,0 +1,84 @@
...
+		global $wgContentNamespaces;
+
+		$response = $this->requestAPI( [
+			'action' => 'query',
+			'list' => 'recentchanges',
+			'rcprop' => 'title|sizes',
+			'rcshow' => '!bot|!minor|!redirect',
+			'rcnamespace' => implode( '|', $wgContentNamespaces ),
+			// load 20 times more items so we'll be able to filter out articles with minor changes
+			'rclimit' => self::LIMIT * 20,
```
rybmat: 100*20? why so much?

rybmat (changes requested):

extensions/wikia/FeedsAndPosts/JustUpdatedAPIProxy.class.php
rybmat: I thought increasing limit to 100 would be enough

latata: it's `4*20`. we get 80 and then return 4.

rybmat: ok, I missread something

rybmat (approved):

extensions/wikia/MercuryApi/models/MercuryApi.class.php
```diff
@@ -261,7 +261,6 @@ public function getWikiVariables() {
 				'enableDiscussions' => $wgEnableDiscussions,
 				'enableDiscussionsImageUpload' => $wgEnableDiscussionsImageUpload,
 				'enableFandomAppSmartBanner' => !empty( $enableFAsmartBannerCommunity ),
-				'enableFeedsAndPosts' => $wgEnableFeedsAndPostsExt,
```
xkxd, line 257: why do we remove it?

rybmat: bc this endpoint is not used by any of ember apps

extensions/wikia/FeedsAndPosts/JustUpdatedAPIProxy.class.php
```diff
@@ -0,0 +1,125 @@
...
+		$articlesWithMinorChange = [];
+		$articlesWithMajorChange = [];
+		$mainPageId = \Title::newMainPage()->getArticleID();
+
+		foreach ( $articles as &$article ) {
+			$diffSize = abs( $article['newlen'] - $article['oldlen'] );
+			$articleId = $article['pageid'];
+
+			unset(
+				$article['newlen'],
```
rybmat: instead of this `unset` can we just create a new object with only the data we need? IMO it will be much more readable

rybmat (changes requested):

rybmat:
@latata one of unittests is still failing

extensions/wikia/FeedsAndPosts/JustUpdatedAPIProxy.class.php
latata: Totally make sense :)

rybmat (approved):

rybmat (approved):

extensions/wikia/FeedsAndPosts/WikiRecentChangesAPIProxy.class.php
```diff
@@ -0,0 +1,129 @@
...
+		}
+	}
+
+	private function filterOutSmallChangesAndCleanUp( $articles ) {
+		$articlesWithMinorChange = [];
+		$articlesWithMajorChange = [];
+		$titlesMap = [];
+		$mainPageId = \Title::newMainPage()->getArticleID();
+
+		foreach ( $articles as &$article ) {
```
jsutterfield: Do we need to pass `$article` by reference here?

extensions/wikia/FeedsAndPosts/WikiRecentChangesAPIProxy.class.php
```diff
@@ -0,0 +1,129 @@
...
+				} else {
+					$articlesWithMinorChange[] = $article;
+				}
+				$titlesMap[$title] = true;
+			}
+		}
+
+		$resultArticles = array_slice( $articlesWithMajorChange, 0, self::LIMIT );
+
+		if ( count( $resultArticles ) < 4 ) {
```
jsutterfield:
```suggestion
		if ( count( $resultArticles ) < self::LIMIT ) {
```
?

extensions/wikia/FeedsAndPosts/WikiRecentChangesAPIProxy.class.php
```diff
@@ -0,0 +1,129 @@
...
+			$resultArticles =
+				array_slice( array_merge( $articlesWithMajorChange, $articlesWithMinorChange ), 0,
+					self::LIMIT );
```
jsutterfield: This spacing is a little hard to read. Maybe: 
```php
			$resultArticles = array_slice(
				array_merge( $articlesWithMajorChange, $articlesWithMinorChange ),
				0,
				self::LIMIT
			);
```

extensions/wikia/FeedsAndPosts/WikiRecentChangesAPIProxy.class.php
```diff
@@ -0,0 +1,129 @@
...
+
+	private function getThumbnailUrl( $url ) {
+		try {
+			return \VignetteRequest::fromUrl( $url )
+				->zoomCrop()
+				->width( self::IMAGE_WIDTH )
+				->height( floor( self::IMAGE_WIDTH / self::IMAGE_RATIO ) )
+				->url();
+		}
+		catch ( \Exception $exception ) {
```
jsutterfield: Can we pull this `catch` up to the line above with the closing `}`?

extensions/wikia/FeedsAndPosts/WikiRecentChangesAPIProxy.class.php
```diff
@@ -0,0 +1,129 @@
...
+		$articlesWithMinorChange = [];
+		$articlesWithMajorChange = [];
+		$titlesMap = [];
+		$mainPageId = \Title::newMainPage()->getArticleID();
+
+		foreach ( $articles as &$article ) {
+			$diffSize = abs( $article['newlen'] - $article['oldlen'] );
+			$title = $article['title'];
+
+			// filter out main page
```
jsutterfield:
```suggestion
			// filter out main page and pages we've already seen
```
?

jsutterfield (approved):
Few last minor suggestions. Other than that, LGTM!

extensions/wikia/FeedsAndPosts/WikiRecentChangesAPIProxy.class.php
latata: Of course. I had been updating `$articles` array before. Now it's not touched.

PR #16274, opened 2018-10-25
IW-1129 escape wiki text in post titles

jsutterfield:
Ticket: https://wikia-inc.atlassian.net/browse/IW-1129

Post titles are user generated content and may contain text which could be interpreted as wiki markup. This is a problem since we're interpolating that text into our i18n messages and then running them through the parser. We should escape that text so it doesn't cause unintended parsing side effects (eg, adding extra links if the text uses brackets `[]`).

cc @Wikia/iwing

### DIFF
--- a/extensions/wikia/Email/Controller/DiscussionController.class.php
+++ b/extensions/wikia/Email/Controller/DiscussionController.class.php
@@ -80,6 +80,16 @@ abstract class DiscussionController extends EmailController {
 		return $this->wiki->city_url . 'd';
 	}
 
+	/**
+	 * Wrap the provided text in <nowiki> tags. This escapes the text preventing
+	 * it from being parsed as wikitext.
+	 * @param $text
+	 * @return string
+	 */
+	protected function wrapTextInNoWikiTags( $text ) {
+		return "<nowiki>" . $text . "</nowiki>";
+	}
+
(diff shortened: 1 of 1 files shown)


rybmat (approved):

extensions/wikia/Email/Controller/DiscussionController.class.php
```diff
@@ -80,6 +80,16 @@ private function getDiscussionsLink() {
...
 	}
 
+	/**
+	 * Wrap the provided text in <nowiki> tags. This escapes the text preventing
+	 * it from being parsed as wikitext.
+	 * @param $text
+	 * @return string
+	 */
+	protected function wrapTextInNoWikiTags( $text ) {
+		return "<nowiki>" . $text . "</nowiki>";
```
latata: don't we use apostrophes for strings?

latata (approved):

extensions/wikia/Email/Controller/DiscussionController.class.php
jsutterfield: Single quotes? Yea, does look like those are preferred: https://www.mediawiki.org/wiki/Manual:Coding_conventions/PHP#String_literals

I'll update.

PR #16278, opened 2018-10-26
SRE | Move tasks publishing to RestInPeace hooks

mszabo-wikia:
Move tasks publishing to `RestInPeace` hook; this avoids an issue where tasks could get lost in case a wiki page edit happened which flushes deferred updates prematurely. With `RestInPeace` this is not an issue as it is guaranteed to run once & only once, at the very end of the request.

For this to work, I converted `SwiftSync` logic to be a task producer (in a similar fashion as the purger). Tests were added to cover `SwiftSync`.


ghost (approved):

lib/Wikia/src/Rabbit/ConnectionManager.php
```diff
@@ -35,7 +35,8 @@ public function __construct( string $host, int $port, string $user, string $pass
 		$this->user = $user;
 		$this->pass = $pass;
 
-		\Hooks::register( 'RestInPeace', [ $this, 'close' ] );
+		// free resources on request shutdown
+		register_shutdown_function( [ $this, 'close' ] );
```
pwojcik86, line 39: Aren't `RestInPeace` hooks triggered on request shutdown?

lib/Wikia/src/SwiftSync/SwiftSync.setup.php
```diff
@@ -18,17 +18,7 @@
...
-$wgAutoloadClasses[ 'Wikia\\SwiftSync\\Hooks'         ] = $dir . "classes/Hooks.php";
-$wgAutoloadClasses[ 'Wikia\\SwiftSync\\ImageSyncTask' ] = $dir . "classes/ImageSyncTask.class.php";
-
 # hooks
 $wgHooks[ 'SwiftFileBackend::doStoreInternal'     ][] = 'Wikia\SwiftSync\Hooks::doStoreInternal';
 $wgHooks[ 'SwiftFileBackend::doCopyInternal'      ][] = 'Wikia\SwiftSync\Hooks::doCopyInternal';
 $wgHooks[ 'SwiftFileBackend::doDeleteInternal'    ][] = 'Wikia\SwiftSync\Hooks::doDeleteInternal';
-
-# DFS operations triggered by the above hooks are pushed to tasks queue via this hook
-$wgHooks[ 'RestInPeace' ][] = 'Wikia\SwiftSync\Hooks::onRestInPeace';
```
pwojcik86, line 34: If we remove this hook, when will `operaitons` be triggered?

lib/Wikia/src/Rabbit/ConnectionManager.php
mszabo-wikia: They're run manually with a `Hooks::run` call before request ends, while this one will be run by the web server environment after all other code is finished. So it's like a separate Hooks system.

ghost (approved):

lib/Wikia/src/SwiftSync/SwiftSync.setup.php
mszabo-wikia: Tasks publishing is registered to run here: https://github.com/Wikia/app/pull/16278/files#diff-22578ec726a5815e66b11de97dc760a7R31 and SwiftSync registers itself as a task producer here: https://github.com/Wikia/app/pull/16278/files#diff-6876f49462d0cbc6129124989dfae604R17 . So the TaskPublisher will fetch its tasks right before publishing. See DefaultTaskPublisher & its integration test.

lib/Wikia/src/Rabbit/ConnectionManager.php
pwojcik86: Why is this needed though? To run `close` after `RestInPeace`?

ghost: I am hesitant about this change. Note, that if something goes wrong and PHP encounters a fatal error before reaching `RestInPeace` event, it won't execute the hooks, but will execute all shutdown functions (there is more than one for sure). So this may have unwanted side effects.

mszabo-wikia: Yeah. Actually, the PHP runtime might be smart enough to take care of closing resources (rabbit connections) on shutdown, so maybe we could just remove this.

mszabo-wikia: @mixoteric Yes that is true, but the `close` method here only takes care of closing Rabbit channels and connections—there is no publishing or business logic that happens there, it's only for cleanup.

ghost (changes requested):
As I mentioned in the previous comment, I am hesitant about this change. The stuff around `RestInPeace` is core MediaWiki's concern and moving it to general PHP's concern might have unwanted side effects.

lib/Wikia/src/Rabbit/ConnectionManager.php
ghost: Okay then :+1:

ghost (approved):

PR #16281, opened 2018-10-29
SUS-6160 | Explicitly set timeouts for nginx - fpm communication

macbre:
https://wikia-inc.atlassian.net/browse/SUS-6160

http://nginx.org/en/docs/http/ngx_http_fastcgi_module.html


docker/base/nginx.conf
```diff
@@ -1,7 +1,7 @@
 worker_processes  1;
 
 error_log  /var/log/nginx/error.log warn;
-pid        /var/run/nginx/nginx.pid;
+pid        /tmp/nginx.pid;
```
pwojcik86: I don't think we need to change this

pwojcik86 (changes requested):

docker/base/nginx.conf
macbre: We have to, for dev instance - I'm getting permission errors

pwojcik86: Then we should fix the dev configuration, not do a workaround in prod

macbre: See e09d1295c5da0f936b8202c62236a15d4884c402

hakubo (approved):

pwojcik86 (approved):

PR #16282, opened 2018-10-29
Hackathon project: circuit breakers PoC

themech:
* APCU-based circuit breakers POC
* two usage examples (for a single call and the whole api client)
* added client-side timeouts (based on @mszabo-wikia PR)


lib/Wikia/src/Persistence/User/Attributes/AttributePersistence.php
```diff
@@ -13,13 +14,22 @@
...
 
+	/** @var CircuitBreaker\CircuitBreaker */
+	private $circuitBreaker;
+
 	public function __construct( ApiProvider $apiProvider ) {
 		$this->apiProvider = $apiProvider;
+		$this->circuitBreaker = new CircuitBreaker\RatioBasedCircuitBreaker( self::SERVICE_NAME,
+			new CircuitBreaker\ApcuStorage(),
+			new CircuitBreaker\Ratio( 3, 5 ),	// failure ratio
+			new CircuitBreaker\Ratio( 2, 5 ),	// success ratio
```
slayful, line 31: I'd probably use (10, 20) for failures and (3, 10) for success to start with.

themech: good suggestion, thanks!

lib/Wikia/src/CircuitBreaker/RatioBasedCircuitBreaker.php
```diff
@@ -0,0 +1,92 @@
...
+		$this->dataStorage->store( $this->serviceName, json_encode( $data ) );
+	}
+
+	public function isAvailable() {
+		$data = $this->loadData();
+		return ( $data['state'] !== State::OPEN );
+	}
+
+	protected function addOperation($success) {
+		$data = $this->loadData();
```
themech, line 56: This method should execute very quickly, so there just a little time between the `loadData ` and `saveData` calls. Still, a race condition is possible and one process can overwrite an operation of another process, which mean the circuit breaker opening/closing may not be accurate.

themech: @mszabo-wikia suggested playing with the `apcu` extension and either using `apcu_inc`/`apcu_dev` or implementing a shift operation and pushing it upstream

PR #16283, opened 2018-10-29
SRE-120 | Make Images Consistent Again

mszabo-wikia:
Prepare maintenance script to replicate past DFS operations SJC -> RES to restore cluster consistency for MW uploads.

Also make `ImageSyncTask` not throw an error if an operation failed. A single failed operation does not mean the other operations are invalid.

https://wikia-inc.atlassian.net/browse/SRE-120

### DIFF
--- a/maintenance/wikia/makeImagesConsistentAgain.php
+++ b/maintenance/wikia/makeImagesConsistentAgain.php
@@ -128,10 +128,12 @@ class MakeImagesConsistentAgain extends Maintenance {
 		$safeUntil = $dbr->addQuotes( $dbr->timestamp( $this->getOption( 'until' ) ) );
 
 		$res = $dbr->select(
-			'filearchive',
+			[ 'filearchive', 'image' ],
 			[ 'fa_name', 'fa_archive_name', 'fa_storage_key' ],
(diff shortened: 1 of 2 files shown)


ghost (approved):

maintenance/wikia/makeImagesConsistentAgain.php
```diff
@@ -0,0 +1,170 @@
...
+				$task->queue();
+			}
+		} else {
+			var_dump( $batches );
+		}
+	}
+
+	private function getAllOperations() {
+		$uploads = 0;
+		foreach ( $this->getNewUploads() as $operation ) {
```
ghost, line 52: Is it possible to get rid of the `foreach`:
```
$newUploads = $this->getNewUploads();
$uploads = count( $newUploads );
yield $newUploads;
```

And the same applies to deletes below. It seems the `foreach`s are used just to increment the counters.

mszabo-wikia: `getNewUploads` will also return a `Generator`, so we'll have to iterate through it anyways and `yield` its values

ghost: Ah, all clear.

PR #16284, opened 2018-10-29
support other environments in sandbox/Jenkinsfile

hakubo:
This way we can reuse whole script for preview deployments

### DIFF
--- a/docker/sandbox/Jenkinsfile
+++ b/docker/sandbox/Jenkinsfile
@@ -148,7 +148,7 @@ EOL""")
 				'SANDBOX_ENVIRONMENT': environment,
 				'SANDBOX_DATACENTER': datacenter,
 				'APP_VERSION': params.app_branch,
-				'CONFIG_VERSION': params.config_branch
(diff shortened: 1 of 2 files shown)


docker/sandbox/Jenkinsfile
```diff
@@ -101,8 +106,8 @@ node("docker-daemon") {
 				imageExists = true
 			}
 
-            if (!imageExists) {
-                sh("docker pull artifactory.wikia-inc.com/sus/php-wikia-base:da4390f")
+			if (!imageExists) {
+					sh("docker pull artifactory.wikia-inc.com/sus/php-wikia-base:da4390f")
```
macbre, line 110: Something is weird with this indent

macbre (approved):

docker/sandbox/Jenkinsfile
hakubo: all good, I have actually 'fixed it' and replaced spaces with tabs in this file

PR #16290, opened 2018-10-29
(PLATFORM-3760) Create content of language wikis index page

Grunny:
/cc @Wikia/core-platform-team


extensions/wikia/LanguageWikisIndex/LanguageWikisIndexController.class.php
```diff
@@ -8,11 +8,29 @@ public function __construct() {
...
+		global $wgIncludeClosedWikiHandler;
 
 		$this->specialPage->setHeaders();
 
-		$this->response->setTemplateEngine( \WikiaResponse::TEMPLATE_ENGINE_MUSTACHE );
+		Wikia::addAssetsToOutput( 'language_wikis_index_scss' );
 
+		$this->setVal( 'langWikis', WikiFactory::getLanguageWikis() );
+
+		if ( !empty( $wgIncludeClosedWikiHandler ) ) {
```
themech: I think you should use `isEmptyDomainWithLanguageWikis` defined in `LanguageWikisIndexHooks` (and/or move it to another class). It would not depend on `$wgIncludeClosedWikiHandler` which is a config variable used by another extension.
`isEmptyDomainWithLanguageWikis` just check the current city id so it should be quick and reliable.

themech (commented):
Not commenting on the CSS, but the PHP part looks good.

extensions/wikia/LanguageWikisIndex/LanguageWikisIndexController.class.php
themech: Cross-extension dependency. Maybe call a hook here and overwrite the intro message in a hook listener defined in `closedWikiHandler`?

extensions/wikia/LanguageWikisIndex/LanguageWikisIndexHooks.class.php
```diff
@@ -41,6 +41,18 @@ public static function onClosedWikiPage( $requestUrl ) {
 		return true;	// proceed with rendering the closed wiki page
 	}
 
+	public static function onGetHTMLBeforeWikiaPage( &$beforeWikiaPageHtml ) {
+		$path = parse_url( RequestContext::getMain()->getRequest()->getRequestURL(), PHP_URL_PATH );
+		if ( self::isEmptyDomainWithLanguageWikis() && $path === self::WIKIS_INDEX_PAGE ) {
+			$beforeWikiaPageHtml .= Html::element(
+				'header',
+				[ 'class' => 'wds-community-header' ],
+				''
```
rogatty, line 54: Empty element? What is it for? Does it look good?

extensions/wikia/LanguageWikisIndex/styles/language-wikis-index.scss
```diff
@@ -0,0 +1,72 @@
+@import 'extensions/wikia/DesignSystem/node_modules/design-system/dist/scss/wds-variables/index';
+
+.language-wiki-index {
+	margin: 50px auto 0;
+	max-width: 710px;
+	text-align: center;
+
+	&__intro {
+		font-size: 22px;
```
rogatty: Let's use variables from https://fandomdesignsystem.com/base-styles/typography. There is no 22px but IMO it only shows why we need to use standard variables. I'd make it `$wds-font-size-xl`

extensions/wikia/LanguageWikisIndex/styles/language-wikis-index.scss
```diff
@@ -0,0 +1,72 @@
...
+		font-weight: bold;
```
rogatty: `$wds-font-weight-bold`

extensions/wikia/LanguageWikisIndex/templates/LanguageWikisIndexController_index.php
```diff
@@ -0,0 +1,39 @@
...
+		</a>
+	</div>
+	<hr />
+	<div class="language-wiki-index__links">
+		<?php
+		/**
+		 * Message keys used here:
+		 *     languagewikisindex-links-cnw
+		 *     languagewikisindex-links-fandom
+		 *     languagewikisindex-links-help
```
rogatty, line 30: :+1:

extensions/wikia/LanguageWikisIndex/LanguageWikisIndexHooks.class.php
Grunny: It's the community header without everything else in it. It puts the same block at the top that we will set the 404 background image on (once we have that asset) as shown in the designs.

extensions/wikia/LanguageWikisIndex/LanguageWikisIndexController.class.php
Grunny: I thought this global might be cheap since it's set in WFL rather than an extension, but rather than using hooks and having that flow complexity, I think we should just check if the domains city ID is not 3 here (or check `WikiFactory::isPublic`). What do you think?

extensions/wikia/LanguageWikisIndex/LanguageWikisIndexHooks.class.php
rogatty: I get how this makes things easier but it can backfire when the community header styles are updated and we need to remember that it's used in a hacky way here. I'd prefer to copy the styles that we need just for this element.

Grunny: Good point, my concern was that the community header styles would be updated and we'd forget to do the same update here, as we'd want this to look the same as any new community header changes too, but it's better for it to look inconsistent if we make changes than potentially break. :+1:

Grunny: Done.

extensions/wikia/LanguageWikisIndex/styles/language-wikis-index.scss
```diff
@@ -0,0 +1,137 @@
+@import 'skins/oasis/css/core/layout';
+@import 'skins/oasis/css/core/breakpoints-variables';
+@import 'extensions/wikia/DesignSystem/node_modules/design-system/dist/scss/wds-variables/index';
+
+$language-wikis-index-header-background-color: #006CB0;
+$language-wikis-index-header-background-height: 105px;
+// Keep in sync with ThemeSettings::BACKGROUND_HEIGHT
```
rogatty: We're not displaying community's background so there is no need to keep it in sync.

extensions/wikia/LanguageWikisIndex/styles/language-wikis-index.scss
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	hr {
+		color: $wds-fandom-color-light-gray;
+	}
+}
+
+.language-wikis-index-header {
+	-moz-osx-font-smoothing: grayscale;
+	-webkit-font-smoothing: antialiased;
```
rogatty: Are we going to display any text there? If not, I'd remove it.

extensions/wikia/LanguageWikisIndex/styles/language-wikis-index.scss
```diff
@@ -0,0 +1,137 @@
...
+	background-color: $language-wikis-index-header-background-color;
+	background-position: right center;
```
rogatty: Everything background-related (apart from color) was crafted for community's background. I see no need to keep it.

Grunny: What about when we add the 404 graphic?

extensions/wikia/LanguageWikisIndex/styles/language-wikis-index.scss
```diff
@@ -0,0 +1,137 @@
...
+			width: $page-width-medium + $padding-width-medium * 2 + $content-border-width * 2;
+		}
+
+		@media #{$breakpoint-large-up} {
+			width: $page-width-large + $padding-width-large * 2 + $content-border-width * 2;
+		}
+	}
+
+	&::before {
+		background: linear-gradient(to right, rgba($language-wikis-index-header-background-color, .5), rgba($language-wikis-index-header-background-color, .5)),
```
rogatty: This is to have a gradient overlay over the background. I don't think we need it anymore.

extensions/wikia/LanguageWikisIndex/styles/language-wikis-index.scss
rogatty: It's not styled exactly like community header. I'd prefer to add only the styles that we need and without the graphic it's hard to say what specifically we'll need.

extensions/wikia/LanguageWikisIndex/styles/language-wikis-index.scss
rogatty: Whole `&::before` can be removed.

extensions/wikia/LanguageWikisIndex/LanguageWikisIndexController.class.php
```diff
@@ -8,11 +8,32 @@ public function __construct() {
...
+		$this->setVal( 'links', [
+			'cnw' => $createNewWikiLink,
+			'fandom' => '//fandom.wikia.com',
+			'fandom-university' => GlobalTitle::newFromText( 'FANDOM University', NS_MAIN, Wikia::COMMUNITY_WIKI_ID )->getFullURL(),
+		] );
 	}
 
+	private function isClosedWiki() {
+		global $wgCityId;
+		return $wgCityId !== WikiFactory::LANGUAGE_WIKIS_INDEX &&
```
themech: ` $wgCityId !== WikiFactory::LANGUAGE_WIKIS_INDEX` duplicates what `!isEmptyDomainWithLanguageWikis()` does. Still, I don't like the fact that we make an assumption here that if this is not an "empty wiki", it must be a closed one. Do you think a hook is much more expensive?

rogatty (approved):
Not commenting on the PHP, but the CSS part looks good.

extensions/wikia/LanguageWikisIndex/LanguageWikisIndexController.class.php
Grunny: We're not making that assumption, we're checking `WikiFactory::isPublic( $wgCityId )` below to see if it was closed. I only included this so we didn't bother with the `WikiFactory::isPublic( $wgCityId )` if we know we're on the "empty" wiki.

I don't think hooks are more expensive, but they're a more complex flow. All we really want to know is "is there a city_list entry and it's closed" which is what `WikiFactory::isPublic( $wgCityId )` will check. Using a hook to lead to the closed wiki handler makes following the code path more complex as you have to see where the hook is used, what it's changing, etc. without benefit over using `WikiFactory::isPublic( $wgCityId )` directly in my opinion.

themech (approved):

PR #16298, opened 2018-10-31
IW-1230 | Install JWP

JamesHawking:
https://wikia-inc.atlassian.net/browse/IW-1230


rybmat (approved):
The same should go to release-758

PR #16319, opened 2018-11-06
IW-1233 Create price update bot for runescape wikis

jsutterfield:
Ticket: https://wikia-inc.atlassian.net/browse/IW-1233

This PR replicates a price update bot which used to exist on the runescape and oldschoolrunescape wikis. The bot queries for all items which need to be updated (those in the "Grand Exchange" category), fetches the latest price info for those items from the Jagex API (Jagex is the creator of Runescape), and updates the Module:Exchange/\<itemName\> and Module:Exchange/\<itemName\>/Data pages with the new price info. Additionally, it updates Module:GEPrices/data page (which is a list of all items and their latest prices), and a series of index pages.

It's based on the old TyBot which was an admin created and maintained bot which updates prices in the same way. After the fork on October 2nd, that bot was moved to the new wikis and removed from our wikis. This code recreates it's functionality so items on our wikis have up-to-date pricing information again.

### DIFF
--- a/maintenance/wikia/runescapeBot/grandExchangePageFetcher.class.php
+++ b/maintenance/wikia/runescapeBot/grandExchangePageFetcher.class.php
@@ -1,50 +1,57 @@
 <?php
 
+/**
+ * Fetches all pages which in the "Grand Exchange" with
+ * a namespace of 112
+ * Class GrandExchangePageFetcher
+ */
 class GrandExchangePageFetcher {
 
-	const GRAND_EXCHANGE_CATEGORY_NAME = "Category: Grand Exchange";
-	const EXCHANGE_NAMESPACE = 112;
+	const CATEGORY_NAME = "Grand Exchange";
+
+	const RUNESCAPE_CITY_ID = 304;
+	const OLDSCHOOL_RUNESCAPE_CITY_ID = 691244;
+
+	const RUNESCAPE_NS_FILTER = 112;
+	const OLDSCHOOL_RUNESCAPE_NS_FILTER = 114;
 
 	/**
 	 * @return array
-	 * @throws MWException
+	 * @throws Exception
 	 */
 	public function fetchAllGrandExchangePages() {
 		$pages = [];
-		$requestParams = [
-			'format' => 'json',
-			'maxlag' => 5,
-			'action' => 'query',
-			'list' => 'categorymembers',
-			'cmprop' => 'title',
-			'cmlimit' => 'max',
-			'cmtitle' => self::GRAND_EXCHANGE_CATEGORY_NAME,
-			'cmnamespace' => self::EXCHANGE_NAMESPACE,
-		];
+		$category = Category::newFromName( self::CATEGORY_NAME );
+		$namespaceFilter = $this->getNameSpaceFilter();
 
-		$hasNext = true;
-		while ( $hasNext ) {
-			$fauxRequest = new FauxRequest( $requestParams );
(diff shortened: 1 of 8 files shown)


maintenance/wikia/runescapeBot/grandExchangePageFetcher.class.php
```diff
@@ -0,0 +1,50 @@
...
+		$requestParams = [
+			'format' => 'json',
+			'maxlag' => 5,
+			'action' => 'query',
+			'list' => 'categorymembers',
+			'cmprop' => 'title',
+			'cmlimit' => 'max',
+			'cmtitle' => self::GRAND_EXCHANGE_CATEGORY_NAME,
+			'cmnamespace' => self::EXCHANGE_NAMESPACE,
+		];
```
mszabo-wikia, line 29: It might be better to create a `Category` instance with `Category::newFromTitle` and use `Category::getMembers( int $limit, string $offset )` to get the members. It's a bit more efficient/straightforward and will also return a `TitleArray` that can be used to iterate over titles.

maintenance/wikia/runescapeBot/updateGrandExchangeItemPrices.php
```diff
@@ -0,0 +1,264 @@
...
+		$this->runescapeApi = new RunescapeApi();
+		$this->pageFetcher = new GrandExchangePageFetcher();
+		$this->textUpdater = new RawTextUpdater();
+	}
+
+	/**
+	 * @throws Exception
+	 */
+	public function execute() {
+		$this->botUser = User::newFromName( self::BOT_USERNAME );
```
mszabo-wikia: There's an existing constant `Wikia::BOT_USER` which should be preferred.

maintenance/wikia/runescapeBot/updateGrandExchangeItemPrices.php
```diff
@@ -0,0 +1,264 @@
...
+	 * Grab the latest update timestamp from Runescape. We can get this value from any item since they're all updated
+	 * once a day and share a latest update timestamp. We'll just use the top most traded item since we have its ID
+	 * already stored in our topItemsTradeCount array.
+	 *
+	 * @return int
+	 */
+	private function getTimeStampForRuneScape() {
+		try {
+			return $this->runescapeApi->getItemById( $this->getFirstItemIdFromTop100Items() )->getTimeStamp();
+		} catch (Exception $e) {
```
mszabo-wikia: Should be spacey

mszabo-wikia (changes requested):

maintenance/wikia/runescapeBot/grandExchangePageFetcher.class.php
jsutterfield: Ah! I didn't know we could do it that way. That does sounds like a better approach. I'll update.

mszabo-wikia (approved):
Looks OK. Depending on the number of pages in the category, we might not want to fetch them all in a single query and use multiple batches instead.

jsutterfield:
Just a heads-up @mszabo-wikia, I ended reverting the changes I made which used the Category class. I tried to implement paging (since we are pulling ~4k pages out of the db), but I was struggling to get it working. Even if using the `ApiMain` approach is a bit less efficient, it does use paging which I think is a greater net benefit. I'm going to stick with that for now.

Thanks again for the eyes and insights :)

PR #16322, opened 2018-11-06
IW-1116 | Remove 'Content-Security-Policy: upgrade-insecure-requests'

JamesHawking:
It causes Edge to fall into redirect loop.

https://wikia-inc.atlassian.net/browse/IW-1116

### DIFF
--- a/includes/wikia/Wikia.php
+++ b/includes/wikia/Wikia.php
@@ -1201,7 +1201,6 @@ class Wikia {
 	 */
 	static public function outputHTTPSHeaders( WebRequest $request ) {
 		if ( WebRequest::detectProtocol() === 'https' ) {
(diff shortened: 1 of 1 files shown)


includes/wikia/Wikia.php
```diff
@@ -1201,7 +1201,6 @@ static public function onWebRequestInitialized( WebRequest $request ) {
 	 */
 	static public function outputHTTPSHeaders( WebRequest $request ) {
 		if ( WebRequest::detectProtocol() === 'https' ) {
-			$request->response()->header('Content-Security-Policy: upgrade-insecure-requests' );
```
rogatty, line 1204: We should remove it only during a redirect.

rogatty:
Fixed in https://github.com/Wikia/app/pull/16415

PR #16323, opened 2018-11-06
 SEO-696 | Disable migration banner 1 week after migration

rogatty:
https://wikia-inc.atlassian.net/browse/SEO-696

@Wikia/core-platform-team

### DIFF
--- a/extensions/wikia/FandomComMigration/FandomComMigrationHooks.class.php
+++ b/extensions/wikia/FandomComMigration/FandomComMigrationHooks.class.php
@@ -50,9 +50,9 @@ class FandomComMigrationHooks {
 
 		if ( $wgFandomComMigrationDone && !empty( $wgDomainChangeDate ) ) {
 			$migrationDateTime = new DateTime( $wgDomainChangeDate );
-			$twoWeeksAgo = ( new DateTime() )->sub( new DateInterval( 'P14D' ) );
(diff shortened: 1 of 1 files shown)


extensions/wikia/FandomComMigration/FandomComMigrationHooks.class.php
```diff
@@ -46,8 +46,15 @@ public static function onWikiaSkinTopScripts( &$vars, &$scripts ) {
...
+
+		if ( $wgFandomComMigrationDone && !empty( $wgDomainChangeDate ) ) {
+			$migrationDateTime = new DateTime( $wgDomainChangeDate );
+			$weekAgo = ( new DateTime() )->sub( new DateInterval( 'P7D' ) );
+
+			return $migrationDateTime > $weekAgo;
+		}
 
-		return !empty( $wgFandomComMigrationScheduled ) || $wgFandomComMigrationDone;
+		return !empty( $wgFandomComMigrationScheduled );
```
themech, line 58: what if `wgFandomComMigrationDone` is set to true but the `wgDomainChangeDate` is empty?

rogatty: Then the data is corrupted and it's better to not show the banner than show it forever.

PR #16339, opened 2018-11-08
Allow for redeploying latest tag to preview

hakubo:
This is needed for a mirror deploy

### DIFF
--- a/docker/preview/Jenkinsfile
+++ b/docker/preview/Jenkinsfile
@@ -122,7 +122,7 @@ def notifyBuild(String buildStatus = 'STARTED') {
   } else {
     color = 'RED'
     colorCode = '#FF0000'
-    //summary = summary + " <!here>"
+    summary = summary + " <!here>"
   }
 
   // Send notifications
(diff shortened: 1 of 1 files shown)


docker/preview/Jenkinsfile
```diff
@@ -122,7 +122,7 @@ def notifyBuild(String buildStatus = 'STARTED') {
   } else {
     color = 'RED'
     colorCode = '#FF0000'
-    summary = summary + " <!here>"
+    //summary = summary + " <!here>"
```
hakubo: le't leave it like that for now - I'll uncomment it when this pipeline will be the main one for preview release

docker/preview/Jenkinsfile
```diff
@@ -293,12 +308,20 @@ node("docker-daemon") {
...
+                createTag(nextAppReleaseTag)
+              } else {
+                println("Source: latest; skip creating a tag")
+              }
             }
           },
           "config": {
             dir("config") {
-              createTag(nextConfigReleaseTag)
+              if (params.source != 'latest') {
```
ghost: Would not it be better to reverse this and have `== latest` in `if` and then everything `else` ? It sounds more human `if it's latest, then skip creating a tag, otherwise, create a tag` :)

ghost (changes requested):
My remark about `!= latest` applies to all cases / occurrences.

macbre (approved):

ghost (approved):

PR #16346, opened 2018-11-08
ADEN-7867 Add "document_hidden" field to adengadinfo and adengadplaye…

jacekbj:
…r DW calls.


dmnsobczak (approved):

extensions/wikia/AdEngine/js/tracking/adInfoTracker.js
```diff
@@ -111,6 +111,7 @@ define('ext.wikia.adEngine.tracking.adInfoTracker',  [
 			'btl': (billTheLizard && billTheLizard.serialize()) || '',
 			'page_width': win.document.body.scrollWidth || '',
 			'page_layout': pageLayout.getSerializedData(slotName) || '',
+			'document_visibility': bridge.geo.getDocumentVisibilityStatus(),
```
fraszczakszymon, line 114: geo? How does it work?

extensions/wikia/AdEngine/js/video/player/playerTracker.js
```diff
@@ -35,6 +35,7 @@ define('ext.wikia.adEngine.video.player.playerTracker', [
 
 	function isEnabled() {
 		return !!context.opts.playerTracking;
+
```
fraszczakszymon: Not needed :)

extensions/wikia/AdEngine/js/tracking/adInfoTracker.js
fraszczakszymon: https://github.com/Wikia/app/blob/dev/extensions/wikia/AdEngine/src/ad-engine.bridge.js#L274
Got it 🤦‍♂️

fraszczakszymon: What do you think about renaming `geo` to `utils`?

jacekbj: That's a perfect idea - I made a naming in one of the previous tasks. I will not do this here because the scope of change would be too big and let's hope we will get rid of bridge soon on Oasis 🤞

dmnsobczak (approved):

PR #16347, opened 2018-11-09
SUS-6166: PHP session handling broken on Kubernetes

macbre:
https://wikia-inc.atlassian.net/browse/SUS-6166

Because we were calling` fastcgi_finish_request()` client received a response before the request fully completed on the backend side. The session data was not yet saved before the next request from the client hit the backend. The second request had the outdated session data.

This caused an issue with notifications that are handled using PHP session.

Partially revert 7900bd05fdd788930af2b9c9ee69e290e482b5d4

### DIFF
--- a/includes/Wiki.php
+++ b/includes/Wiki.php
@@ -441,23 +441,13 @@ class MediaWiki {
 	 * If possible this method will be executed only after the response has been flushed.
 	 */
 	public function restInPeace() {
-		// SUS-6164 | move emitting of profiling data before fastcgi_finish_request is called
-		wfLogProfilingData();
-
-		// If using FastCGI we can flush the output now and do any further updates without blocking
-		if ( function_exists( 'fastcgi_finish_request' ) ) {
-			fastcgi_finish_request();
-		} else {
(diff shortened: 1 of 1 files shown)


ghost:
Per off-line conversation with @macbre - things such as the code removed in this changeset should be introduce with more "meat and potatoes", meaning: it was borrowed from the latest MediaWiki, but it should have been borrowed along with the entire session handling implementation.

ghost:
In addition to that, this is somewhat general change - given the recent "restInPeace" refactor by @mszabo-wikia - I am not so sure about what can we break by fixing this particular issue this way.

@mszabo-wikia - thoughts?

ghost:
Alright, per @mszabo-wikia _There is nothing that relies on the `fastcgi_finish_request` call being there, it was  backported only to take advantage of the performance gains it can bring, so it's safe to remove. So removing it will restore us to the old and working state basically._

ghost (approved):

hakubo (approved):

mszabo-wikia (approved):

PR #16356, opened 2018-11-10
Nginx version bump - security fix

ljagiello:
Changelog:
http://nginx.org/en/CHANGES-1.14

### DIFF
--- a/docker/dev/docker-compose.yml
+++ b/docker/dev/docker-compose.yml
@@ -2,7 +2,7 @@ version: '2'
 
 services:
   web:
-    image: nginx:1.14.1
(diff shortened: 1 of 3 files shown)


macbre:
We can use [the most recent mainline nginx version - 1.15.6](http://nginx.org/en/CHANGES)

> We recommend that in general you deploy the NGINX mainline branch at all times.

macbre (approved):

ljagiello:
@macbre 1.14.x is a stable branch, I would recommend to stay with a stable version if there is no need for any particular feature for mainline tree.

PR #16363, opened 2018-11-13
SRE-109 | Prefer reading data from replicas rather than master during edit

mszabo-wikia:
For queries which are not sensitive to state changes made by the edit, we should use replicas rather than the master. If a replica ends up being too far behind, infrastructure will transparently remove it from the pool.

* `FounderEmailsEditEvent` would query master after each edit to determine the user's edit count. However, we're actually only interested in if this was the user's first edit, i.e. is the user's edit count 0 or not 0—we do not care about the exact edit count. For this, we can simply use `UserStatsService` instead, which avoids referred heavy calculation.
* Make `Title::checkUserBlock` read user's block status from a replica instead of the master.
* Make `LinksUpdate` read existing state from a replica instead of the master.
* Made `EmailNotification` get the list of users who are watching a given page from a replica instead of the master.

### DIFF
--- a/extensions/wikia/FounderEmails/events/FounderEmailsEditEvent.class.php
+++ b/extensions/wikia/FounderEmails/events/FounderEmailsEditEvent.class.php
@@ -318,23 +318,15 @@ class FounderEmailsEditEvent extends FounderEmailsEvent {
 			$wasNotificationSent = ( static::getFirstEmailSentFlag( $editor->getName() ) === '1' ) ;
 
 			if ( !$wasNotificationSent ) {
-				$userEditStatus = static::getUserEditsStatus( $editor, true );
-				/*
-					If there is at least one edit, flag that we should not send this email anymore;
-					either first email is sent out as a result of this request,
-					or there was more than 1 edit so we will never need to send it again
-				 */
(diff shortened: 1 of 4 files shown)


extensions/wikia/FounderEmails/events/FounderEmailsEditEvent.class.php
```diff
@@ -318,23 +318,15 @@ public static function recentChanges( $oRecentChange ) {
 			$wasNotificationSent = ( static::getFirstEmailSentFlag( $editor->getName() ) === '1' ) ;
 
 			if ( !$wasNotificationSent ) {
-				$userEditStatus = static::getUserEditsStatus( $editor, true );
```
Grunny, line 321: Can `getUserEditsStatus` be removed then? I don't see anything else using it.

mszabo-wikia: Indeed it was unused—removed it!

ghost (approved):

Grunny (approved):

themech (approved):

PR #16369, opened 2018-11-13
(CAKE-4069) Update DS API links for N&S migration

Grunny:
/cc @Wikia/cake @themech @harnash


includes/wikia/models/DesignSystemSharedLinks.class.php
```diff
@@ -76,90 +76,90 @@ private function getLangWithFallback( $lang ) {
 
 	private $hrefs = [
 		'default' => [
-			'explore-wikis' => '//fandom.wikia.com/explore',
-			'about' => '//www.wikia.com/about',
+			'explore-wikis' => '//www.wikia.com',
```
vforge, line 79: @Grunny we're changing the explore page to N&S feed?

Grunny: We're placing the explore page as the home page of www.wikia.com, so www.wikia.com will serve the content from www.wikia.com/fandom/explore after the switch, rather than the N&S feed.

vforge: OK :)

vforge (approved):

harnash (approved):

PR #16374, opened 2018-11-14
PLATFORM-3800 | Support language wikis index and closed wikis in mobile-wiki

rogatty:
## Links
- http://wikia-inc.atlassian.net/browse/PLATFORM-3800
- https://github.com/Wikia/mobile-wiki/pull/1199

## Description
- Add the following fields to `WikiVariables` when applicable:
  - `isEmptyDomainWithLanguageWikis`
  - `isClosed`
  - `languageWikis`
  - `scriptPath`
- Enable `MercuryApi::getMobileWikiVariables` on wikis where we display language wikis index
- Enable `DesignSystemApi` on wikis where we display language wikis index
- Remove community header from `DesignSystemApi::getAllElements` on wikis where we display language wikis index, there should be no local content

## Reviewers
@Wikia/core-platform-team @Wikia/iwing


extensions/wikia/LanguageWikisIndex/LanguageWikisIndexController.class.php
```diff
@@ -17,8 +18,7 @@ public function index() {
 
 		$this->setVal( 'langWikis', WikiFactory::getLanguageWikis() );
 
-		$isClosed = $this->isClosedWiki();
-		if ( $isClosed ) {
+		if ( $wgIncludeClosedWikiHandler ) {
```
themech: out of curiosity - when does the `$wgIncludeClosedWikiHandler` work differently than the `isClosedWiki` method that was doing the `!WikiFactory::isPublic( $wgCityId )` check?
Also: should we update other places (like https://github.com/Wikia/app/pull/16374/files#diff-637fd602cdb768ee21823835c4ff2426R17) for consistency?

extensions/wikia/LanguageWikisIndex/LanguageWikisIndexHooks.class.php
```diff
@@ -65,6 +65,40 @@ public static function onGetHTMLBeforeWikiaPage( &$beforeWikiaPageHtml ) {
 		return true;
 	}
 
+	public static function onMercuryWikiVariables( array &$wikiVariables ): bool {
+		if (
+			self::isEmptyDomainWithLanguageWikis() ||
+			self::isClosedWikiWithLanguageWikis()
```
themech: This duplicates the `WikiFactory::isLanguageWikisIndex` method you're adding it this PR.

extensions/wikia/WikiFactory/WikiFactory.php
```diff
@@ -625,6 +625,27 @@ public static function getLanguageWikis() {
 		return self::getWikisUnderDomain( $url['host'] );
 	}
 
+	public static function isLanguageWikisIndex( $cityId = null ) {
```
themech, line 637: Please add a method doc

themech (changes requested):

themech:
In general it looks ok, but I think we need to unify the code detecting language index/closed wikis, as we have several implementations doing the same checks.

extensions/wikia/LanguageWikisIndex/LanguageWikisIndexController.class.php
rogatty: When `city_public === -1` then `$wgIncludeClosedWikiHandler === false` while `$this->isClosedWiki() === true`. I'm not sure if it matters but I wanted to have a consistent check for a closed wiki.

In the case of "on active wikis the `/language-wikis` path should redirect to the main page" it's better to use `WikiFactory::isPublic` because we want to redirect only on `city_public === 1`.

rogatty: We've agreed to use `WikiFactory::isPublic()` instead of `$wgIncludeClosedWikiHandler`

themech (approved):

PR #16375, opened 2018-11-14
LORE-101 - allow internal requests to use WF set api

nmonterroso:
@Wikia/lore @mszabo-wikia 
https://wikia-inc.atlassian.net/browse/LORE-101

Updates `ApiWikiFactorySaveVariable` to allow internal requests without authentication to successfully save wf variables.


extensions/wikia/WikiFactory/api/ApiWikiFactorySaveVariable.php
```diff
@@ -6,7 +6,7 @@ class ApiWikiFactorySaveVariable extends ApiBase {
 	 * @throws UsageException
 	 */
 	public function execute() {
-		if ( !$this->getUser()->isAllowed( 'wikifactory' ) ) {
+		if ( !$this->isInternalRequest() && !$this->getUser()->isAllowed( 'wikifactory' ) ) {
```
drsnyder, line 9: Would this be more clear as `!($this->isInternalRequest() || $this->getUser()->isAllowed( 'wikifactory' ))`?

drsnyder:
LGTM

mszabo-wikia (changes requested):
We should check `X-Mw-Auth-Ok` header too

extensions/wikia/WikiFactory/api/ApiWikiFactorySaveVariable.php
nmonterroso: I think that's a matter of personal preference. Personally, I think the less parens the easier it is to read code, especially here where I'd have to mentally expand the `!` to the two conditions.

nmonterroso:
@mszabo-wikia can you explain what purpose `X-Mw-Auth-Ok` serves? I know we use it in several places to determine if a request is internal, but never really understood why.

nmonterroso:
@mszabo-wikia I added the check for `X-Mw-Auth-Ok` header in 5780d9d2bdb

mszabo-wikia (approved):

PR #16377, opened 2018-11-15
getWikisUnderDomain: return info about the primary domain

themech:
return localized domain prefixed with the protocol


harnash (approved):
We might consider using `wfHttpsAllowedForURL` if possible

themech:
or rather `wfHttpsEnabledForURL`

harnash (approved):

PR #16378, opened 2018-11-15
SUS-6263 | internal requests inside k8s fix

macbre:
https://wikia-inc.atlassian.net/browse/SUS-6263

`X-Original-Host` is required as we proxy "directly" to nginx when we perform internal HTTP requests within Kubernetes cluster. However, we did not add `fandom.com` to the list of internal domains, hence this request header was not set (see #15648).

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -258,6 +258,7 @@ $wgConf->localVHosts = array_merge(
     $wgWikiFactoryDomains,
     [
         $wgWikiaBaseDomain,
+        $wgFandomBaseDomain,
         'uncyclopedia.org',
         'memory-alpha.org',
         'wowwiki.com',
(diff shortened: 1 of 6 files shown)


pwojcik86 (approved):

LocalSettings.php
```diff
@@ -258,6 +258,7 @@
     $wgWikiFactoryDomains,
     [
         $wgWikiaBaseDomain,
+        $wgFandomBaseDomain,
```
hakubo, line 261: spaces?

hakubo (approved):

includes/HttpFunctions.php
```diff
@@ -431,7 +432,17 @@ public function getHeaderList() {
 		// SUS-5499: Use internal host name for MW->MW requests when running on Kubernetes
 		global $wgKubernetesNamespace;
 		if ( !empty( $wgKubernetesNamespace ) && Http::isLocalURL( $this->url ) ) {
-			$list[] = sprintf( 'X-Original-Host: %s', $this->parsedUrl['host'] );
+			$originalHost = $this->parsedUrl['host'];
```
themech: cannot we simply go with this instead of duplicating the code that replaces part of the host?
`
$list[] = sprintf( 'X-Original-Host: %s', wfNormalizeHost( $this->parsedUrl['host'] ) );
`

macbre: Implemented in 6454b28

PR #16395, opened 2018-11-19
Add autoscaling to sandbox and preview

hakubo:
It's easier to add to both envs now
also I think since they are so similar they should behave similarly

preview autoscale max replicas: 10 - since we run tests there
sandbox autoscale max replicas: 2


pwojcik86 (commented):
I would also reduce the `cpu` request and limit for the `php` container.

pwojcik86 (approved):

PR #16396, opened 2018-11-19
SRE-109 | Optimize PageStatsService

mszabo-wikia:
Optimize `PageStatsService` to not read from master but use the passed `Title` object properly. Previously, the class constructor worked only with article IDs, which forced a lookup from the master DB. Using `Title` objects everywhere removes dependency on master DB without risking reading stale data.

### DIFF
--- a/includes/wikia/services/PageStatsService.class.php
+++ b/includes/wikia/services/PageStatsService.class.php
@@ -3,43 +3,18 @@ class PageStatsService {
 
 	const CACHE_TTL = 86400;
 
-	private $mTitle = null;
-	private $pageId;
+	/** @var Title $mTitle */
+	private $mTitle;
 
-	/**
-	 * Pass page ID of an article you want to get data about
-	 */
-	function __construct($pageId) {
-		$this->pageId = intval($pageId);
-	}
-
-	/**
-	 * @static
-	 * @param Title $title
-	 * @return PageStatsService
-	 */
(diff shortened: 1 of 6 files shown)


includes/wikia/services/PageStatsService.class.php
```diff
@@ -88,60 +59,40 @@ static function onArticleSaveComplete(
...
-		$title = $article->getTitle();
-
-		// tell service to invalidate cached data for deleted page
-		if (!empty($title)) {
-			$service = self::newFromTitle($title);
-			$service->regenerateData();
-
-			wfDebug(__METHOD__ . ": cache cleared for page #{$articleId}\n");
-		}
+		$service = new self( $article->getTitle() );
```
themech, line 62: previously there was an emptiness check so we assumed it is possible that `$article->getTitle()` can return null. Now we don't do this check which means we can create new `PageStatsService` with an empty title. But the `PageStatsService` does not check for emptiness when making calls like `$this->mTitle->getArticleID()`. Isn't this an issue?

mszabo-wikia: `WikiPage::doDeleteArticleReal` would fail if somehow the `WikiPage`  instance had a null title, so if we got here, it's safe to assume we have an instance of `Title`.

themech: 👍

themech (approved):

PR #16398, opened 2018-11-19
SUS-6267 | give mediawiki-prod some breathing space for heavy requests spikes

macbre:
See https://wikia-inc.atlassian.net/browse/SUS-6267 for load-testing details.

### DIFF
--- a/docker/prod/prod.template.yaml
+++ b/docker/prod/prod.template.yaml
@@ -90,10 +90,10 @@ spec:
               value: "tcp://localhost:9999"
           resources:
             limits:
-              cpu: "6"
+              cpu: "4"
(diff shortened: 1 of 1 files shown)


docker/prod/prod.template.yaml
```diff
@@ -127,13 +127,13 @@ spec:
...
     name: mediawiki-prod
-  maxReplicas: 50
+  maxReplicas: 75
   minReplicas: 20
   metrics:
   - type: Resource
     resource:
       name: cpu
-      targetAverageUtilization: 50
+      targetAverageUtilization: 25
```
pwojcik86: I think it's better to reduce the `cpu` request instead

pwojcik86 (approved):

PR #16400, opened 2018-11-20
SEO-689 | Modify "red links" to be non-crawlable

maaartyyynaa:


includes/Linker.php
```diff
@@ -240,11 +240,18 @@ public static function link(
 			$html = self::linkText( $target );
 		}
 
+		# decode data-uncrawlable-url to set it as href for red link
+		if ($attribs['href'] === '#'){
```
rogatty, line 244: Take a look at https://github.com/Wikia/guidelines/blob/master/PHP/CodingConventions.md and https://www.mediawiki.org/wiki/Manual:Coding_conventions/PHP, there should be more spaces in here and the rest of your PHP changes.

mszabo-wikia: There's an importable IntelliJ/PHPStorm code style file at https://github.com/Wikia/guidelines/blob/master/PHP/bin/MediaWiki.xml that can be used to automatically apply these settings.

PR #16406, opened 2018-11-20
SEO-700 | Add hreflang links to main pages on fandom.com wikis

rogatty:
## Links

* https://wikia-inc.atlassian.net/browse/SEO-700
* https://support.google.com/webmasters/answer/189077
* https://github.com/Wikia/mobile-wiki/pull/1209

## Description

Restore `SeoLinkHreflang` and update it to use `WikiFactory::getWikisUnderDomain()` as the data source. Add hreflang links to MediaWiki's `<head>` element and return them in `MercuryApi::getPage()` to be used in mobile-wiki.

## Reviewers

@Wikia/core-platform-team


includes/wikia/VariablesBase.php
```diff
@@ -3523,6 +3523,13 @@
  */
 $wgEnableSendGridPostback = true;
 
+/**
+ * Enable SEO Link Hreflang extension.
+ * @see extensions/wikia/SeoLinkHreflang
+ * @var bool $wgEnableSeoLinkHreflangExt
+ */
+$wgEnableSeoLinkHreflangExt = false;
```
rogatty: We need to either change it to `true` or set it in WikiFactory, depending on Becky's decision.

includes/wikia/GlobalFunctions.php
```diff
@@ -1658,3 +1658,15 @@ function wfForceBaseDomain( $url, $targetServer ) {
 	$finalUrl = http_build_url( $url, [ 'host' => $finalHost, ] );
 	return WikiFactory::getLocalEnvURL( $finalUrl );
 }
+
+/**
+ * @return bool
+ */
+function wfIsCurrentWikiAFandomComWiki() {
```
Grunny: Is it worth avoiding some code duplication here, as this method is just: `wfHttpsEnabledForURL( $wgServer );`?

rogatty: Well, IIRC I wanted to name `wfHttpsEnabledForURL` something like this one but you stopped me and wanted it to be more generic :stuck_out_tongue: If the logic there changes and more URLs are https enabled then those hreflang links will break. IMO we need to choose either deduplication or being generic.

Grunny: I more meant reuse code, but make it named for a purpose so it's more resistant to implementation changes. So, I still think `wfHttpsEnabledForURL` is better than `wfIsCurrentWikiAFandomComWiki`, since presumably, this could also involve other domains, like .org in the future, at which point, `wfHttpsEnabledForURL` still wouldn't need to have the method name changed, but `wfIsCurrentWikiAFandomComWiki` would. Here, I'm thinking whether we want to say this is only enabled on fandom.com, or if it's only enabled on a wiki that is migrated/on HTTPS by default, etc. in which case you can use a generic name, without duplicating code. `wfIsCurrentWikiAFandomComWiki` is the kind of name I want to avoid for a method ideally. :) Even if this is a separate method, I'd make one of them reuse the other so one is just an alias to the other.

rogatty: Makes sense, how about https://github.com/Wikia/app/pull/16406/commits/7275323c092b1476a3babbd7558b58a498ea82f5?

Grunny (approved):

PR #16413, opened 2018-11-21
Do not leak StubObject from WikiFactoryLoaderIntegrationTest

mszabo-wikia:
`WikiFactoryLoader` internally initializes `$wgContLang` with a `StubObject` because MW LB requires `$wgContLang` to be defined in order to run, but `WikiFactoryLoader` runs very early in the request lifecycle so the variable is empty at the time. In tests, we must properly restore the value of `$wgContLang` after execution to avoid leaking the `StubObject` to other tests.

### DIFF
--- a/extensions/wikia/WikiFactory/tests/WikiFactoryLoaderIntegrationTest.php
+++ b/extensions/wikia/WikiFactory/tests/WikiFactoryLoaderIntegrationTest.php
@@ -18,6 +18,9 @@ class WikiFactoryLoaderIntegrationTest extends WikiaDatabaseTest {
 		WikiFactory::isUsed( false );
 		$wgExtensionFunctions = [];
 		$this->dbName = $wgDBname;
+		
+		// WikiFactoryLoader has side effects that initialize the global content language with a StubObject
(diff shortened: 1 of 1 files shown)


extensions/wikia/WikiFactory/tests/WikiFactoryLoaderIntegrationTest.php
```diff
@@ -8,12 +8,17 @@ class WikiFactoryLoaderIntegrationTest extends WikiaDatabaseTest {
 	/** @var string $dbName */
 	private $dbName;
 
+	/** @var Language $oldContentLanguage */
```
slayful, line 11: Can add the explanation you provided in the PR as a comment here, for context?

slayful (approved):
I'd add context, but besides that LGTM. Shame we rely on globals so much. :<

extensions/wikia/WikiFactory/tests/WikiFactoryLoaderIntegrationTest.php
mszabo-wikia: Added

PR #16414, opened 2018-11-21
SEO-689

maaartyyynaa:


extensions/wikia/AssetsManager/config.php
```diff
@@ -2460,7 +2460,7 @@
 	'type' => AssetsManager::TYPE_JS,
 	'skin' => [ 'oasis' ],
 	'assets' => [
-		'//extensions/wikia/CategoryPage3/scripts/alphabet-shortcuts.js',
+		'//extensions/wikia/CategoryPage3/scripts/decode-redlink-href.js',
```
rogatty: We have an asset group called `category_page3_alphabet_shortcuts_js` that loads `decode-redlink-href.js` which is in `/extensions/wikia/CategoryPage3` and handles `a[data-uncrawlable-url]`. It's very confusing, let's clean it up.
My suggestion:
- Move this file to `/extensions/wikia/SEOTweaks/scripts`
- Rename it to `uncrawlable-urls.js`
- Change the asset group name to `seo_tweaks_uncrawlable_urls`

extensions/wikia/RTE/RTELinkerHooks.class.php
```diff
@@ -133,4 +133,26 @@ public static function makeBrokenImageLinkObj( $title, $html = '', $query = '',
...
+
+	/**
+	 * Sets href as # and store proper href in
+	 * data-uncrawlable-url to make redlink
+	 * uncrawlable.
+	 *
+	 * @param $attribs array attributes
+	 *
+	 */
+	public static function onSetHrefForRedLink (&$attribs) {
```
rogatty: Please reformat this method according to https://github.com/Wikia/guidelines/blob/master/PHP/CodingConventions.md

extensions/wikia/RTE/RTELinkerHooks.class.php
```diff
@@ -133,4 +133,26 @@ public static function makeBrokenImageLinkObj( $title, $html = '', $query = '',
...
+		wfProfileIn(__METHOD__);
```
rogatty: This is a legacy thing, we don't add it in a new code. The same for `wfProfileOut`

includes/Linker.php
```diff
@@ -228,6 +228,8 @@ public static function link(
 
 		# Note: we want the href attribute first, for prettiness.
 		$attribs = array( 'href' => self::linkUrl( $target, $query, $options ) );
+		Hooks::run( 'SetHrefForRedLink', array(&$attribs));
```
rogatty: There is no need to add a new hook, you can use `LinkEnd` (https://github.com/Wikia/app/pull/16414/files#diff-8ffbd08183616445af94793926721964R247)

extensions/wikia/RTE/RTELinkerHooks.class.php
rogatty: Also, RTE extensions isn't a good place for this hook. RTE stands for Rich Text Editor and it's unrelated to redlinks. I think the best place would be to put it in `/extensions/wikia/SEOTweaks` as this is strictly a SEO improvement.

extensions/wikia/RTE/RTELinkerHooks.class.php
```diff
@@ -133,4 +133,6 @@ public static function makeBrokenImageLinkObj( $title, $html = '', $query = '',
 		wfProfileOut(__METHOD__);
 		return $ret;
 	}
+
+
```
rogatty: No need for that.

includes/Linker.php
```diff
@@ -228,6 +228,7 @@ public static function link(
 
 		# Note: we want the href attribute first, for prettiness.
 		$attribs = array( 'href' => self::linkUrl( $target, $query, $options ) );
+
```
rogatty: No need for that.

extensions/wikia/SEOTweaks/SEOTweaksHooksHelper.class.php
```diff
@@ -352,6 +352,12 @@ public static function onLinkEnd( $skin, Title $target, array $options, &$text,
 			$attribs['rel'] = 'nofollow';
 		}
 
+		if ( strpos( $attribs['href'], 'redlink=1' ) !== false ) {
```
rogatty: We add `redlink=1` when there is `'broken'` value in `$options`, see https://github.com/Wikia/app/blob/98d2c8db4d82daaa6b7d3634a0029d8370c5332f/includes/Linker.php#L300-L307
There is no need for additional check, you can just move the actions inside this block to the block above.

rogatty (approved):

PR #16415, opened 2018-11-21
PLATFORM-3828 | Remove Content-Security-Policy when downgrading protocol

rogatty:
https://wikia-inc.atlassian.net/browse/PLATFORM-3828
https://wikia-inc.atlassian.net/browse/IW-1116

Remove Content-Security-Policy header when downgrading HTTPS to HTTP.
Microsoft Edge starting from v17 misinterprets `Content-Security-Policy: upgrade-insecure-requests` during redirects and falls into a loop HTTPS -> HTTP -> HTTPS -> ...

@Wikia/core-platform-team @JamesHawking

### DIFF
--- a/extensions/wikia/HTTPSSupport/HTTPSSupportHooks.class.php
+++ b/extensions/wikia/HTTPSSupport/HTTPSSupportHooks.class.php
@@ -67,6 +67,26 @@ class HTTPSSupportHooks {
 	}
 
 	/**
+	 * Remove Content-Security-Policy header when downgrading HTTPS to HTTP
+	 * Microsoft Edge starting from v17 misinterprets Content-Security-Policy: upgrade-insecure-requests
+	 * during redirects and falls into a loop HTTPS -> HTTP -> HTTPS -> ...
+	 *
+	 * @param $out
+	 * @param $redirect
+	 * @param $code
+	 * @param $redirectedBy
+	 */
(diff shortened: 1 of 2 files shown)


extensions/wikia/HTTPSSupport/HTTPSSupportHooks.class.php
```diff
@@ -66,6 +66,26 @@ public static function onBeforeInitialize( Title $title, $unused, OutputPage $ou
...
+	 * during redirects and falls into a loop HTTPS -> HTTP -> HTTPS -> ...
+	 *
+	 * @param $out
+	 * @param $redirect
+	 * @param $code
+	 * @param $redirectedBy
+	 */
+	public static function onBeforePageRedirect( $out, &$redirect, &$code, &$redirectedBy ) {
+		$currentUri = \WikiFactoryLoader::getCurrentRequestUri( $_SERVER, false, true );
+		$currentProtocol = parse_url( $currentUri, PHP_URL_SCHEME );
```
themech: I think you can use `WebRequest::detectProtocol()` if you don't need the full uri

rogatty: Yup, that works too, thanks!

themech (approved):

rogatty:
It doesn't work as expected.

PR #16420, opened 2018-11-22
SRE-109 | Update cu_changes table in a background task

mszabo-wikia:
The `cu_changes` table is used by the CheckUser staff tool to log some user data after each edit—this logging should be performed asynchronously via a background task instead of the request thread.


Grunny (approved):
Looks good, I assume the other updates to cu_changes are infrequent enough that it's not worth doing the same (account creation, email changes, etc.)?

mszabo-wikia:
@Grunny yup, the one for account creation is already asynchronous (it's executed in a task published from the user-registration services), while the email hook was only run by the EmailUser feature, which we disabled.

PR #16421, opened 2018-11-22
SRE-109 | Increment users' edit counts via a background task

mszabo-wikia:
Increment users' edit counts via a background task instead of performing it within the request thread. This can be especially beneficial if multiple edits happen within a single request, for example in certain media upload scenarios. At most one task will be published per request (but the tasks' `increment` method will be scheduled to be executed multiple times on the task runner, if needed).


Grunny (approved):

ghost (approved):
Yeah, it'll work, it's strict when it comes to design pattern use and implementation. Yet, to my taste, it's too much code to read, grasp, understand and - more importantly - execute, simply to increment a number. :)

PR #16439, opened 2018-11-29
API: communities and robots rules for a given domain

themech:
API allows to fetch:
* list of communities under a given domain (with FandomCreator data)
* list of robots rules for mediawiki communities


extensions/wikia/FandomCreator/FandomCreator.setup.php
```diff
@@ -12,39 +12,46 @@
...
 } );
 
-$wgAutoloadClasses['FandomCreator\Api\CommunitySetupController'] = __DIR__ . '/Api/CommunitySetupController.php';
+// These hooks are active on every wiki
+$wgHooks['GetWikisUnderDomain'][] = 'FandomCreator\Hooks::onGetWikisUnderDomain';
+$wgHooks['GenerateRobotsRules'][] = 'FandomCreator\Hooks::onGenerateRobotsRules';
 
-$wgHooks['NavigationApigetDataAfterExecute'][] = function( WikiaDispatchableObject $dispatchable ) {
-	global $wgMaxLevelOneNavElements, $wgFandomCreatorCommunityId, $wgMaxLevelTwoNavElements, $wgMaxLevelThreeNavElements;
+if ( !empty($wgFandomCreatorCommunityId ) ) {
```
themech: as this setup file is included on every wiki, some of the code was moved into this `if` block

Grunny (approved):

PR #16455, opened 2018-12-06
IW-1372 | Remove Forum highlights

mszabo-wikia:
Remove Forum thread highlight feature

https://wikia-inc.atlassian.net/browse/IW-1372
@Wikia/iwing


extensions/wikia/WallNotifications/WallNotifications.class.php
```diff
@@ -537,7 +533,6 @@ protected function addNotificationLink( $userId, WallNotificationEntity $notific
 				'entity_key' => $notification->id,
 				'author_id' => $notifData->msg_author_id,
 				'is_reply' => $notification->isReply(),
-				'notifyeveryone' => $notifData->notifyeveryone
```
mszabo-wikia, line 540: I've added a default value to the corresponding DB column, so we can now omit this in INSERTs (and drop the column post release).

rybmat (approved):

PR #16457, opened 2018-12-06
IW-1348 | add data-source and data-image-name attributes to PI html markup

rybmat:
https://wikia-inc.atlassian.net/browse/IW-1348
https://wikia-inc.atlassian.net/browse/IW-1376

* `<infobox>` now takes `type` attribute which is then translated to `type-<value>` class for the `<asside>` element in html markup
* `<group>`, `<data>`, `<header>`, `<image>`, `<navigation>` tags now take `name` attribute which then is translated to `data-item-name` attribute in html
* each xml tag that takes `source` now produces `data-source` attribute in output html

@Wikia/iwing


extensions/wikia/PortableInfobox/services/PortableInfoboxRenderService.class.php
```diff
@@ -244,14 +248,27 @@ protected function createHorizontalGroupData( $groupData ) {
 			$data = $item['data'];
 
 			if ( $item['type'] === 'data' ) {
-				array_push( $horizontalGroupData['labels'], $data['label'] );
-				array_push( $horizontalGroupData['values'], $data['value'] );
+				array_push( $horizontalGroupData['labels'], [
+					'text' => $data['label'],
+					'item-name' => $data['item-name'],
+					'source' => $data['source'],
+				] );
```
mszabo-wikia: `$horizontalGroupData['labels'][] = ...`

extensions/wikia/PortableInfobox/services/PortableInfoboxRenderService.class.php
```diff
@@ -306,17 +323,33 @@ private function createSmartGroupItem( $rowItems, $rowSpan ) {
...
-		return array_reduce( $rowItems, function ( $result, $item ) use ( $capacity ) {
-			$styles = "width: calc({$item['data']['span']} / $capacity * 100%);";
-
-			$label = $item['data']['label'] ?? "";
-			if ( !empty( $label ) ) {
-				$result['renderLabels'] = true;
-			}
-			$result['labels'][] = [ 'value' => $label, 'inlineStyles' => $styles ];
-			$result['values'][] = [ 'value' => $item['data']['value'], 'inlineStyles' => $styles ];
+		return array_reduce(
```
mszabo-wikia, line 327: It might be more readable to do this with a `foreach` loop, PHP's "functional" APIs are not too intuitive.

extensions/wikia/PortableInfobox/templates/PortableInfoboxItemGroup.mustache
```diff
@@ -1 +1,3 @@
-<section class="pi-item pi-group pi-border-color{{#cssClasses}} {{{cssClasses}}}{{/cssClasses}}">{{{content}}}</section>
\ No newline at end of file
+<section class="pi-item pi-group pi-border-color{{#cssClasses}} {{{cssClasses}}}{{/cssClasses}}"{{#item-name}} data-item-name="{{item-name}}"{{/item-name}}>
+	{{{content}}}
+</section>
```
mszabo-wikia: newline

mszabo-wikia (changes requested):

extensions/wikia/PortableInfobox/services/PortableInfoboxRenderService.class.php
rybmat: actually I would leave it as is, since I'd need to modify data in array I'm looping over or create a second array which also would not be the prettiest solution

mszabo-wikia (approved):

PR #16462, opened 2018-12-10
IW-1373 | SiteWide message task for disabling forum highlgiht

bkoval:
@mszabo-wikia @rybmat Still in progress but my ideas for maintenance task are:

1) Get all wikis with forum enabled
2) Get all admins of these wikis and sort them by language their wikis are using
3) Localize message in their languages
4) Send localized message to all the admins using a SiteWideMessageTask, passing it a list of admins in a specific lang


extensions/wikia/SiteWideMessages/ForumHighlightSiteWideMessageMaintenance.php
```diff
@@ -0,0 +1,97 @@
...
+	 * Main entry point, only public function
+	 * Function runs once to send sitewide message about retiring forum higlight feature
+	 *
+	 * @author bkoval
+	 * @access public
+	 * @throws MWException
+	 */
+	public function execute() {
+		// Get all the Wikis with forum enabled
+		$wikisWithForumEnabled = WikiFactory::getListOfWikisWithVar( 'wgEnableForumExt', 'bool', '=', true );
```
rybmat: I wonder if it is better to do it this way or just run maintenance script for every wiki and check if wgEnableForumExt is set to true. In current approach you need to remember to specify single wiki to run this script

rybmat (approved):

mszabo-wikia:
Message sent.

PR #16464, opened 2018-12-10
CORE-17 | Remove second escaping for text nodes in portable infoboxes.

JamesHawking:
[Ticket](https://wikia-inc.atlassian.net/browse/CORE-17)


Grunny (approved):
Looks OK, but I'd make sure we test every node type to make sure this extra escaping wasn't being relied upon to prevent XSS in any output situation.

hakubo (approved):

PR #16471, opened 2018-12-11
PLATFORM-3881 | Redlink behaviour is broken on in classic editor previews

maaartyyynaa:


extensions/wikia/SEOTweaks/scripts/uncrawlable-urls.js
```diff
@@ -6,14 +6,19 @@ require([
 	window
 ) {
 	'use strict';
-	$(function () {
+
+	const decodeUncrawlableURL = function() {
```
KockaAdmiralac: `const` is a JavaScript ES6 keyword and therefore does not work on older browsers. Since Internet Explorer 11 is a [supported browser](https://c.wikia.com/wiki/Help:Supported_browsers) I believe just:
```javascript
function decodeUncrawlableURL() {
    // ...
}
```
is preferred here.

extensions/wikia/SEOTweaks/scripts/uncrawlable-urls.js
```diff
@@ -6,14 +6,19 @@ require([
...
 			var $this = $(this);
 			var url = window.atob($this.attr('data-uncrawlable-url'));
 			$this.attr('href', url);
 		});
-	});
+	};
+
+	$(decodeUncrawlableURL);
+
+	$( window ).on( 'EditPageAfterRenderPreview', decodeUncrawlableURL);
```
KockaAdmiralac: Way more content-displaying asynchronous code is using the `wikipage.content` hook. This only covers edit previews, while `wikipage.content` is used (besides being fired for edit previews) on, for example, setting hooks on new posted threads:
https://github.com/Wikia/app/blob/1e4ea22073c29ec97beeccda86d068915366d0c5/extensions/wikia/Wall/js/WallMessageForm.js#L71
It is also used a lot in userscripts, for example [AddRailModule](https://dev.wikia.com/wiki/MediaWiki:AddRailModule/code.js) (so if the `wikipage.content` hook isn't used here, user-created Rail modules will contain redlinks that don't work. Clearly, this isn't a SEO concern, but since the issue affects more than just edit previews this easy change should fix it for more than just them).

extensions/wikia/SEOTweaks/scripts/uncrawlable-urls.js
```diff
@@ -6,14 +6,19 @@ require([
...
 			var $this = $(this);
 			var url = window.atob($this.attr('data-uncrawlable-url'));
 			$this.attr('href', url);
 		});
-	});
+	};
+
+	$(decodeUncrawlableURL);
+
+	mw.hook('wikipage.content').add(decodeUncrawlableURL);
```
themech, line 22: Code looks ok. Just wondering why did you decide to use `wikipage.content` instead of `EditPageAfterRenderPreview` suggested in the ticket.

maaartyyynaa: We were talking about that with @Grunny and final decision was to use `wikipage.content` instead of `EditPageAfterRenderPreview`

PR #16472, opened 2018-12-11
MAIN-18407: Fix "password too short" message

KockaAdmiralac:
The [password changing form](https://c.wikia.com/wiki/Special:ChangePassword) is showing "password must be at least $1 characters" without replacing the `$1` variable. That is because the variable is never passed to the error message, and this pull request fixes that.

Bug ticket: [MAIN-18407](https://wikia-inc.atlassian.net/browse/MAIN-18407)
Support ticket: [ZD#424553](https://support.wikia-inc.com/hc/en-us/requests/424553)

### DIFF
--- a/includes/User.php
+++ b/includes/User.php
@@ -2180,7 +2180,15 @@ class User implements JsonSerializable {
 				'userId' => $this->getId(),
 				'err'    => $heliosPasswordChange,
 			] );
-			throw new PasswordError( wfMessage( $heliosPasswordChange->errors[0]->description )->text() );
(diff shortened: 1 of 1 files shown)


mszabo-wikia (changes requested):
It's not the right method—`heliosDeletePassword` is responsible for deleting the password, probably you were looking to add it to `heliosSetNewPassword`

KockaAdmiralac:
@mszabo-wikia Whoops, it should be fixed with https://github.com/Wikia/app/pull/16472/commits/d0854bbde60dece54d8e338022856446fe84ddcc now.

PR #16483, opened 2018-12-13
(PLATFORM-3874) Add support for creating lang path wikis on fandom.com

Grunny:
/cc @Wikia/core-platform-team

### DIFF
--- a/extensions/wikia/CreateNewWiki/CreateNewWikiHooks.php
+++ b/extensions/wikia/CreateNewWiki/CreateNewWikiHooks.php
@@ -4,12 +4,10 @@ class CreateNewWikiHooks {
 
 	public static function onMakeGlobalVariablesScript( array &$vars, OutputPage $outputPage ) {
 		if ( $outputPage->getTitle()->isSpecial( 'CreateNewWiki' ) ) {
-			global $wgWikiaBaseDomain, $wgFandomBaseDomain,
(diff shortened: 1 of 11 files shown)


extensions/wikia/CreateNewWiki/CreateNewWikiHooks.php
```diff
@@ -4,13 +4,10 @@ class CreateNewWikiHooks {
 
 	public static function onMakeGlobalVariablesScript( array &$vars, OutputPage $outputPage ) {
 		if ( $outputPage->getTitle()->isSpecial( 'CreateNewWiki' ) ) {
-			global $wgWikiaBaseDomain, $wgFandomBaseDomain,
-				   $wgCreateLanguageWikisWithPath, $wgCreateEnglishWikisOnFandomCom;
+			global $wgWikiaBaseDomain, $wgFandomBaseDomain;
 
 			$vars['wgWikiaBaseDomain'] = $wgWikiaBaseDomain;
```
themech: Do we still need `wgWikiaBaseDomain` in JS?

themech (approved):

extensions/wikia/CreateNewWiki/CreateNewWikiHooks.php
Grunny: Nope, it is used in other JS but it's set in the RL startup module for that case. Removed.

PR #16491, opened 2018-12-14
 (PLATFORM-3878) Update various CC references to use fandom.com

Grunny:
/cc @Wikia/core-platform-team


Grunny:
This is the prepared PR in advance of the migration next year.

themech:
Apparently the integration tests are failing, does it mean we have to update the community domain on dev?

themech (approved):

PR #16493, opened 2018-12-17
CORE-6 | Not all pages using Portable Infoboxes have right page props

macbre:
https://wikia-inc.atlassian.net/browse/CORE-6

@Wikia/core-team 

The root cause here seems to be excessive replace queries performed on `infoboxes` property rows in `page_props` tables. I've replaced delete calls with cache invalidation.

`LinksUpdate::incrTableUpdate`  takes care of keeping `page_props` entries up to date.


extensions/wikia/PortableInfobox/services/PortableInfoboxDataService.class.php
```diff
@@ -233,7 +233,21 @@ protected function clear() {
...
+
+	/**
+	 * This method performs rendering of a current template and saves its metadata
+	 * in the page_props table.
+	 *
+	 * This method will be used to fix portable templates with missing database entries.
+	 *
+	 * @see CORE-6
+	 */
+	public function regenerate() {
```
hakubo, line 229: where is this being used?

macbre: It will be used by a following script that fixes the issue for affected portable infoboxes:

```php
$pages = wfGetDB(DB_SLAVE)->selectFieldValues('page', 'page_id', 'page_namespace = ' . NS_TEMPLATE);
foreach($pages as $page) PortableInfoboxDataService::newFromPageID($page)->regenerate();
```

PR #16497, opened 2018-12-17
ADEN-8037 | Track X for outstream

bckatarzyna:


nandy-andy (approved):

extensions/wikia/AdEngine/js/video/player/porvata/floater.js
```diff
@@ -79,6 +89,8 @@ define('ext.wikia.adEngine.video.player.porvata.floater', [
 
 		function createOnCloseListener(floatingContext) {
 			return function () {
+				var slot = adEngineBridge.slotService.get('INCONTENT_PLAYER');
+				slot.setStatus('force-unstick');
```
dmnsobczak: How about using `slot.emitEvent('force-unstick')` instead? The functionality is the same, but we keep it consistent with AdEngine code:
https://github.com/Wikia/ad-engine/blob/dev/src/ad-products/templates/sticky-ad.js#L165

dmnsobczak (approved):

PR #16498, opened 2018-12-17
ADEN-8070 Get last id from wrappers

fraszczakszymon:


extensions/wikia/AdEngine/js/video/vastParser.js
```diff
@@ -6,18 +6,16 @@ define('ext.wikia.adEngine.video.vastParser', [
 	'use strict';
 	var logGroup = 'ext.wikia.adEngine.video.vastParser';
 
-	function getFirstNumber(possibleValues) {
+	function getLastNumber(possibleValues) {
```
nandy-andy, line 9: How about the scenario when `possibleValues` will have only not a number elements? Previously we've returned an empty string now it can be an `undefined`, I think. Whatever the decision is, my suggestion is to cover this negative case also in unit tests :) Thanks to this small change within 6-12 months we'll be able to tell how we thought the method was supposed to work ;)

fraszczakszymon: Great catch 💎

nandy-andy (approved):

PR #16499, opened 2018-12-18
ADEN-8096 Line item and creative ids without quotes

fraszczakszymon:


extensions/wikia/AdEngine/js/spec/provider/gpt/adElement.spec.js
```diff
@@ -125,4 +125,21 @@ describe('ext.wikia.adEngine.provider.gpt.adElement', function () {
 		expect(element.getNode().getAttribute('data-gpt-creative-id')).toEqual('456');
 		expect(element.getNode().getAttribute('data-gpt-creative-size')).toEqual('[728,90]');
 	});
+
+	it('Add response event details and keep strings without quotes', function () {
```
nandy-andy, line 129: 💎 ❤️

nandy-andy (approved):
It looks fine. I spent some time wondering if we should always trust stuff coming from GPT event. According to [the GPT docs](https://developers.google.com/doubleclick-gpt/reference#googletageventsslotrenderendedevent) both `lineItemId` and `creativeId` may be empty. On the other hand, this piece of code we always have trusted so if there is no easy way to validate/make sure it can be safely passed to DOM we can think about it outside this changes/ticket. Feel free to ping me and discuss the topic ;)

PR #16516, opened 2019-01-02
PLATFORM-3863 closed wikis

harnash:
getWikisUnderDomain() returns 410 for closed wikis

Ticket: https://wikia-inc.atlassian.net/browse/PLATFORM-3863
Linked PR: https://github.com/Wikia/robots-txt/pull/7

### DIFF
--- a/includes/wikia/api/WikisApiController.class.php
+++ b/includes/wikia/api/WikisApiController.class.php
@@ -296,6 +296,10 @@ class WikisApiController extends WikiaApiController {
 			}
 		}
 
+		if ( !WikiFactory::isPublic( $cityId ) ) {
+			throw new ResourceGoneException();
+		}
+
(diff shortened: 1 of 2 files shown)


includes/wikia/api/WikisApiController.class.php
```diff
@@ -296,6 +296,10 @@ public function getWikisUnderDomain() {
 			}
 		}
 
+		if ( !WikiFactory::isPublic( $cityId ) ) {
+			throw new ResourceGoneException();
```
Grunny: What would happen in cases where the English wiki at the root was closed, but language path wikis exist (and so we serve the index page at the root, and should have robots rules for the language versions of the wiki)?

Grunny (approved):

PR #16520, opened 2019-01-03
IW-1466 | Implement wiki article recommendations in MCF and rail

mszabo-wikia:
https://wikia-inc.atlassian.net/browse/IW-1466


extensions/wikia/Rail/templates/Rail_stickyModule.php
```diff
@@ -1,7 +0,0 @@
-<div id="WikiaAdInContentPlaceHolder" class="rail-sticky-module">
```
rybmat, line 1: I have no idea in what way it could be used but I'd ask somebody from AdEng if they don't need it

extensions/wikia/Recirculation/RecirculationApiController.class.php
```diff
@@ -15,109 +14,20 @@ public function __construct() {
...
-		}
-
-		$this->response->setCacheValidity( WikiaResponse::CACHE_STANDARD );
-		$this->response->setData( [
-			'title' => $title,
-			'posts' => $posts,
-		] );
-	}
-
 	public function getPopularWikiArticles() {
```
rybmat, line 18: I can not find any usages of this method, can we get rid of it?

extensions/wikia/Recirculation/js/recirculation.js
```diff
@@ -62,143 +40,52 @@ require([
...
-				height: 337,
-				modelName: 'wiki',
-				opts: {
-					rule_language: window.wgContentLanguage
-				}
-			}
 		};
 
-	function prepareRailRecirculation(options) {
-		if (mw.config.get('showEmbeddedFeed')) {
```
rybmat, line 78: I think we'd like to have this condition fulfilled somehow, did you implement it in new approach? I can not find it

extensions/wikia/Recirculation/services/PopularPagesService.php
```diff
@@ -0,0 +1,57 @@
+<?php
+
+class PopularPagesService {
+	public function getPopularPages( int $limit, int $thumbWidth, int $thumbHeight ): array {
+		global $wgCityId, $wgContentNamespaces;
+
+		$articles = DataMartService::getTopArticlesByPageview(
```
rybmat, line 7: as far as I can see this method queries DB directly. Do we want to cache it somehow? Or use WikiService class maybe? It caches data for 24h

rybmat (changes requested):

extensions/wikia/Rail/templates/Rail_stickyModule.php
mszabo-wikia: It's still there, I just moved it to our recirculation template. This way one can see at a glance what exactly is rendered.

extensions/wikia/Recirculation/RecirculationApiController.class.php
mszabo-wikia: For the time being it is used by mobile-wiki at https://github.com/Wikia/mobile-wiki/blob/dev/app/components/recirculation-prefooter.js#L100, we can definitely remove it after we update mobile-wiki

extensions/wikia/Recirculation/services/PopularPagesService.php
mszabo-wikia: This method uses a cache internally for 1 day

rybmat: I looked at wrong method :p

rybmat (approved):

PR #16533, opened 2019-01-08
IW-1293 Adds Feeds Wiki Variables

leshniak:


latata (approved):

extensions/wikia/FeedsAndPosts/WikiVariables.class.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+namespace Wikia\FeedsAndPosts;
+
+class WikiVariables {
+	public function get() {
+		$cacheTTL = 60; // a minute
+
+		return \WikiaDataAccess::cache(
+			wfMemcKey( 'feeds-wiki-variables' ),
```
mszabo-wikia: probably redundant, these are already memcached b/c of wikifactory

mszabo-wikia (changes requested):

extensions/wikia/FeedsAndPosts/WikiVariables.class.php
leshniak: see changes, can we still get rid of memcache?

mszabo-wikia: yep, all these variables are prefetched and cached by WFL at the start of the request

leshniak: fixed

mszabo-wikia (approved):

extensions/wikia/FeedsAndPosts/WikiVariables.class.php
```diff
@@ -0,0 +1,18 @@
...
+
+class WikiVariables {
+	public function get() {
+		global $wgServer, $wgDBname, $wgDisableHTTPSDowngrade, $wgEnableHTTPSForAnons;
+
+		$wikiVariables = [
+			'basePath' => $wgServer,
+			'dbName' => $wgDBname,
+		];
+
```
themech, line 13: Cannot we use the existing `MercuryApi::getWikiVariables` method? It has the basePath, dbName and all the rest: https://witcher.fandom.com/wikia.php?controller=MercuryApi&method=getWikiVariables

leshniak: Actually, we can. We did this to minimize the payload for feeds.

PR #16534, opened 2019-01-08
IW-1475 | add sponsored content mock to MCF

xkxd:
https://wikia-inc.atlassian.net/browse/IW-1475

- response is mocked
- does not contain changes in Rail.

@Wikia/iwing


extensions/wikia/Recirculation/js/helpers/sponsoredContent.js
```diff
@@ -0,0 +1,97 @@
...
+                    "US",
+                    "AU"
+                ],
+                "title": "Kermit",
+                "siteName": "Netflix"
+            }
+        ];
+
+		return mock;
+		// return $.ajax();
```
rybmat: services.wikia(-dev).(pl|us|com)/wiki-recommendations/sponsored-articles

extensions/wikia/Recirculation/js/helpers/sponsoredContent.js
```diff
@@ -0,0 +1,97 @@
...
+                "siteName": "Hulu"
+            },
+            {
+                "id": 0,
+                "url": "https://muppet.wikia.com/wiki/Elmo",
+                "thumbnailUrl": "https://vignette.wikia.nocookie.net/starwars/images/d/d6/Yoda_SWSB.png/revision/latest/scale-to-width-down/500?cb=20150206140125",
+                "weight": 25,
+                "geos": [],
+                "title": "Elmo",
+                "siteName": "Hulu"
```
rybmat: how about `attribution` for field name here and the second one would be `attributionLabel`?

rybmat (approved):

extensions/wikia/Recirculation/js/views/mixedFooter.js
```diff
@@ -46,14 +46,22 @@ define('ext.wikia.recirculation.views.mixedFooter', [
...
+			$sponsoredContentHook.replaceWith(
+				utils.renderTemplate(
+					templates[templatePaths.sponsoredContent],
+					$.extend(
+						true,
+						{},
+						sponsoredContent,
+						{
+							shortTitle: sponsoredContent.title.substring(0, 80) + '...',
+							attributionLabel: sponsoredContent.attributionLabel || 'Sponsored by'
```
rybmat: do we have translations for default "sponsored by" value? I think the assumption is we have it. If not lets notify stanley and ask if it is required or we'll be relying on the overrides put into the sponsored articles UI

extensions/wikia/Recirculation/templates/client/Recirculation_sponsoredContent.mustache
```diff
@@ -12,7 +12,7 @@
 			{{title}}
 		</span>
 		<span class="mcf-card-article__subtitle">
-			Sponsored by {{siteName}}
+			{{attributionLabel}} {{attribution}}
```
rybmat: Can we just in case make the attributionLabel visible only if attribution is not empty?

rybmat (changes requested):

extensions/wikia/Recirculation/js/views/mixedFooter.js
xkxd: based on what Brett Bates wrote on Jira about having full control we won't translate that label, so I'd expect to have this field mandatory in form (or note that `Sponsored by` is the default somewhere)

rybmat: It is not mandatory in form and per my comment in the ticket it they know it will not be mandatory:
`so it would be one more input field in UI and if filled with text then use it instead of standard 'sponsored by', right?`

so we should make a note somewhere there

rybmat (approved):

extensions/wikia/Recirculation/js/views/mixedFooter.js
xkxd: I'll update a label in service

PR #16539, opened 2019-01-08
MAIN-18542 | Prevent CKE overriding native browser notifs

speeditor:
While in CKE, userscripts are unable to send native browser notifications. This is because `window.Notification`  is overriden by leak (and is also [broken](http://puu.sh/CtyOe/f9bdab2eca.png) as a result).

Here's a ~possible~ valid(!) fix for it using IIFE. 😄 

ZD: https://support.wikia-inc.com/hc/requests/428617
JIRA: https://wikia-inc.atlassian.net/browse/MAIN-18542

### DIFF
--- a/extensions/wikia/RTE/ckeditor/plugins/notification/plugin.js
+++ b/extensions/wikia/RTE/ckeditor/plugins/notification/plugin.js
@@ -8,918 +8,920 @@
  *
  */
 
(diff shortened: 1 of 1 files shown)


speeditor:
Tested in Chrome (Dev tools "Overrides")

PR #16540, opened 2019-01-09
Platform 3901 - changing how sitemaps paginate

abador:
Unfortunately there wasn't a way to get rid of ORDER BY page_id (it seems they aren't in the correct order). The query that takes out all the data to generate sub sitemaps was a little bit tricky with the current query builder(the to string method returns the query for the prepare statement, without parameters), yet I decided to use the existing getQuery method and build it on top of the builder. The performance on subpages is almost linear, the sitemap index generates in about 4s, and most of this time is waiting for the db.

### DIFF
--- a/extensions/wikia/SitemapXml/SitemapXmlController.class.php
+++ b/extensions/wikia/SitemapXml/SitemapXmlController.class.php
@@ -64,17 +64,19 @@ class SitemapXmlController extends WikiaController {
 		if ( $parsedPath->index ) {
 			$out = $this->generateSitemapIndex();
 		} elseif ( is_int( $parsedPath->namespace ) ) {
-			$out = $this->generateSitemapPage( $parsedPath);
+			$out = $this->generateSitemapPage( $parsedPath );
 		} else {
 			$title = ( new WikiaHtmlTitle() )->setParts( [ 'Not Found' ] )->getTitle();
 			$body = '<title>%s</title><h1>Not Found</h1><p>The requested URL was not found</p>';
 
 			$response->setCode( 404 );
 			$response->setBody( sprintf( $body, htmlspecialchars( $title ) ) );
+
 			return;
 		}
 
-		$out .= sprintf( '<!-- Generation time: %dms -->' . PHP_EOL, microtime( true ) * 1000 - $start );
+		$out .= sprintf( '<!-- Generation time: %dms -->' . PHP_EOL,
+			microtime( true ) * 1000 - $start );
 		$out .= sprintf( '<!-- Generation date: %s -->' . PHP_EOL, wfTimestamp( TS_ISO_8601 ) );
 
 		if ( !$this->isRequestedGzipped() ) {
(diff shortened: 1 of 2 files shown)


themech:
Is this backwards compatible? I mean what if the main sitemap is served from the CDN (where the `$end` is not a part of the sitemap name). And the the request for the sub-sitemap gets to our backend (as for any reason it is not in the CDN). What would be the result?

abador:
> Is this backwards compatible? I mean what if the main sitemap is served from the CDN (where the `$end` is not a part of the sitemap name). And the the request for the sub-sitemap gets to our backend (as for any reason it is not in the CDN). What would be the result?

As far as I can tell right now the result should be a 404 http code. I can add some logic to redirect the request based on the page param. It would involve some caching for the results used by the sitemap index generator or else the redirect will take a long time.

abador:
> Is this backwards compatible? I mean what if the main sitemap is served from the CDN (where the `$end` is not a part of the sitemap name). And the the request for the sub-sitemap gets to our backend (as for any reason it is not in the CDN). What would be the result?

I've added the legacy route and sitemap generation via page param. RIght now the old routing should work. I've added some todos where the parts can be deleted after some time(when cache is refreshed)

extensions/wikia/SitemapXml/SitemapXmlController.class.php
```diff
@@ -114,21 +114,33 @@ private function getLastMod( $touched ) {
 		return date( 'Y-m-d\TH:i:s\Z', $touchedTime );
 	}
 
-	private function generateSitemapPage( $namespace, $page ) {
+	private function generateSitemapPage($parsedPath) {
```
themech: Here are our code style guidelines for PHP: https://github.com/Wikia/guidelines/blob/master/PHP/CodingConventions.md. I think you need to apply those.

extensions/wikia/SitemapXml/SitemapXmlController.class.php
```diff
@@ -64,7 +64,7 @@ public function index() {
 		if ( $parsedPath->index ) {
 			$out = $this->generateSitemapIndex();
 		} elseif ( is_int( $parsedPath->namespace ) ) {
-			$out = $this->generateSitemapPage( $parsedPath->namespace, $parsedPath->page );
+			$out = $this->generateSitemapPage( $parsedPath);
```
themech:
```suggestion
			$out = $this->generateSitemapPage( $parsedPath );
```

extensions/wikia/SitemapXml/SitemapXmlController.class.php
themech:
```suggestion
	private function generateSitemapPage( $parsedPath ) {
```

themech (approved):

extensions/wikia/SitemapXml/SitemapXmlModel.class.php
```diff
@@ -25,6 +24,36 @@ public function getPageNumber( $namespace, $limit ) {
...
+				->FIELD( 'page_id', 'rownum' )
+				->FROM( '(SELECT @x:=-1) AS x, (' . $mainSQL->injectParams( $this->dbr,
+						$this->getQuery( $namespace )
+							->FIELD( 'page_id' )
+							->FIELD( '(@x:=@x+1)' )
+							->AS_( 'rownum' )
+							->build() ) . ')' )
+				->AS_( 'db' )
+				->WHERE( 'rownum MOD ' . $limit )
+				->EQUAL_TO( '0' )
```
themech, line 46: respect :)

themech (approved):

PR #16543, opened 2019-01-09
PLATFORM-3864| do not do foreign call to the WikiaRobotsController when it is not needed

harnash:
Ticket: https://wikia-inc.atlassian.net/browse/PLATFORM-3864
Related PR: https://github.com/Wikia/robots-txt/pull/10


harnash:
First part of the fix is to not do `ApiService::foreignCall()` when querying rules for a local wiki and just return them right away.
Second part is optional but I believe it fixes some issues we have currently on prod when root domain is calling for rules on sub-domains which have `read` permission set to `0`. It introduces optional parameter to `ApiService::foreignCall()` which allows request to be made internally (recognised by MW).

extensions/wikia/RobotsTxt/WikiaRobotsController.class.php
```diff
@@ -26,15 +29,24 @@ public function getRulesForDomain() {
...
-				];
-				if ( $this->request->getBool( 'forcerobots' ) ) {
-					$params['forcerobots'] = '1';
+				if ( $wikiData['city_id'] === $wgCityId ) {
+					$robots = $this->getRobotsBuilder();
+					$response = [
+						'allowed' => $robots->getAllowedPaths(),
+						'disallowed' => $robots->getDisallowedPaths(),
+						'sitemaps' => $robots->getSitemaps(),
+					];
```
themech: This duplicates the `getAllowedDisallowed` and we also want to make sure the response is the same. Maybe we can add a new private `getAllowedDisallowedRules` method:
```
$wikiaRobots = new WikiaRobots( new PathBuilder() );
$robots = $wikiaRobots->configureRobotsBuilder( new RobotsTxt() );
return [
'allowed' => $robots->getAllowedPaths(),
...
];
```
Then we can call it here and in the `getAllowedDisallowed` method (where we just iterate over the result and call `setVal`)

harnash: Fix: c0a44bfd869

themech (approved):

PR #16545, opened 2019-01-09
IW-1467 | Populate MCF N&S slots with trending Fandom articles

mszabo-wikia:
https://wikia-inc.atlassian.net/browse/IW-1467]
@Wikia/iwing

### DIFF
--- a/extensions/wikia/Recirculation/RecirculationApiController.class.php
+++ b/extensions/wikia/Recirculation/RecirculationApiController.class.php
@@ -1,7 +1,8 @@
 <?php
 
 class RecirculationApiController extends WikiaApiController {
-	const ALLOWED_TYPES = ['recent_popular', 'vertical', 'community', 'curated', 'hero', 'category', 'latest', 'posts', 'all', 'stories'];
+
+	const MAX_MIXED_CONTENT_FOOTER_SLOTS = 13;
 
 	/**
 	 * @var CrossOriginResourceSharingHeaderHelper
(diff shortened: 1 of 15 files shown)


extensions/wikia/Recirculation/RecirculationApiController.class.php
```diff
@@ -23,11 +24,31 @@ public function getPopularPages() {
...
+	public function getTrendingFandomArticles() {
+		global $wgParselyApiUrl, $wgParselyApiKey, $wgParselyApiSecret, $wgMemc;
+		$this->response->setFormat( WikiaResponse::FORMAT_JSON );
+
+		$limit = min(
+			$this->request->getInt( 'limit', static::MAX_MIXED_CONTENT_FOOTER_SLOTS ),
+			static::MAX_MIXED_CONTENT_FOOTER_SLOTS
+		);
+
+		$articleService = new ParselyService( $wgParselyApiUrl, $wgParselyApiKey, $wgParselyApiSecret );
```
latata, line 47: Can't we just use these all globals inside the `ParselyService`?

extensions/wikia/Recirculation/services/PopularPagesService.php
```diff
@@ -34,6 +35,10 @@ public function getPopularPages( int $limit, int $thumbWidth, int $thumbHeight )
 					'thumbnail' => $images[$articleId][0]['url'],
 					'hasVideo' => false,
 				];
+
+				if ( ++$count >= $limit ) {
```
latata: can we don't use `++$count` and increment it just before the `if`? It would more readable than IMO

latata (approved):

extensions/wikia/Recirculation/RecirculationApiController.class.php
mszabo-wikia: Passing them in the constructor allows us to integration test this class

rybmat (approved):

PR #16549, opened 2019-01-10
PLATFORM-3904 | Update redirected URLs for logged in users, to (mostly) match anon experience

maaartyyynaa:


extensions/wikia/SEOTweaks/SEOTweaksHooksHelper.class.php
```diff
@@ -357,4 +360,13 @@ public static function onLinkEnd( $skin, Title $target, array $options, &$text,
 
 		return true;
 	}
+
+	public static function onDisplayRedirectedFrom( $rdfrom, Article $article): bool {
+		if ( !$article->getContext()->getUser()->isAnon() ) {
+			$redir = Linker::makeExternalLink( $rdfrom, $rdfrom );
+			$article->getContext()->getOutput()->addSubtitle( wfMessage( 'redirectedfrom' )->rawParams( $redir ) );
```
Grunny: Minor: I'd suggest putting `$article->getContext()` in a variable to be used here and in the `if`, and you can then also make use of it in place of `wfMessage` with `$context->msg( 'redirectedfrom' )->rawParams( $redir )`.

Grunny (changes requested):

Grunny (approved):

extensions/wikia/SEOTweaks/SEOTweaksHooksHelper.class.php
```diff
@@ -364,7 +364,15 @@ public static function onLinkEnd( $skin, Title $target, array $options, &$text,
...
 		if ( !$context->getUser()->isAnon() ) {
-			$redir = Linker::makeExternalLink( $rdfrom, $rdfrom );
+			$redirectNamespace = $article->getTitle()->getNamespace();
+			$title = Title::makeTitleSafe( $redirectNamespace, $rdfrom );
+			$article->setRedirectedFrom( $title );
+			$redir = Linker::linkKnown(
+				$article->mRedirectedFrom,
+				null,
+				array(),
+				array( 'redirect' => 'no' )
```
Grunny: Minor: These should use short array syntax, per conventions.

Grunny (approved):

PR #16550, opened 2019-01-10
ADEN-8164 Fixing double player initialisation

dmnsobczak:


nandy-andy (dismissed):
It looks fine as a quick fix ;) And since we won't release it tomorrow I'd suggest thinking about refactoring the code since it's getting less readable because of all those tabs :/

Right now only two ideas:
a) go with opposite: `if (!adContext.get('opts.babDetectionDesktop'))` and return early, or/and
b) create small functions which will help making anonymous function added to consent queue more readable.

One more comment from a different domain: does it make sense to expand FeaturedVideo tests of ours and check within them if the widget is present or we should create a separate job which will cover the scenarios?

nandy-andy (approved):

extensions/wikia/ArticleVideo/scripts/featured-video.jwplayer.instant.js
```diff
@@ -109,30 +109,40 @@ require([
...
 
-		if (!adContext.get('opts.showAds')) {
+			return;
+		}
+
+		setupPlayer();
+	}
+
+	trackingOptIn.pushToUserConsentQueue(function () {
+		if (!adContext.get('opts.showAds') || !adContext.get('opts.babDetectionDesktop')) {
```
fraszczakszymon: It seems that it skips bidders when bab is disabled

extensions/wikia/ArticleVideo/scripts/featured-video.jwplayer.instant.js
```diff
@@ -109,30 +109,40 @@ require([
...
 				setupPlayer();
-			}
-		});
+			});
 
-		if (!adContext.get('opts.showAds')) {
+			return;
+		}
+
+		setupPlayer();
```
fraszczakszymon, line 130: It's getting hard to see what's blocking the player setup and in what order. What do you think about calling it in one place and blocking with all dependencies? Something like:
```suggestion
function waitForBab() {
	if (!adContext.get('opts.babDetectionDesktop')) {
		return Promise.resolve();
	}

	return new Promise(function (resolve, reject) {
		doc.addEventListener('bab.blocking', function () {
			reject();
		});
		doc.addEventListener('bab.not_blocking', function () {
			resolve();
		});
	});
}

function waitForRec() {
	if (!adContext.get('opts.wadHMD')) {
		return Promise.resolve();
	}
	
	return new Promise(function (resolve) {
		hmdRecLoader.setOnReady(function () {
			resolve();
		});
	});
}

function waitForBidders() {
	if (!bidders || !bidders.isEnabled()) {
		return Promise.resolve();
	}
	
	return new Promise(function (resolve) {
		bidders.runOnBiddingReady(function () {
			bidParams = bidders.updateSlotTargeting(featuredVideoSlotName);
			resolve();
		});
	});
}

trackingOptIn.pushToUserConsentQueue(function () {
	if (adContext.get('opts.showAds')) {
		waitForBab()
			.then(waitForBidders, waitForRec)
			.then(setupPlayer);
	} else {
		setupPlayer();
	}
});
```

fraszczakszymon: :ie11:

extensions/wikia/ArticleVideo/scripts/featured-video.jwplayer.instant.js
dmnsobczak: You right, fixed.

PR #16558, opened 2019-01-14
IW-1468 | adjust trackign for recirculation

rybmat:
https://wikia-inc.atlassian.net/browse/IW-1468

@Wikia/iwing


extensions/wikia/Recirculation/js/recirculation.js
```diff
@@ -173,6 +175,19 @@ require([
 							}
 						))
 				);
+
+				tracker.trackImpression('rail');
+
+				$rail.on('click', '[data-tracking]', function () {
+					var $this = $(this),
+						labels = $this.data('tracking').split(','),
+						href = $this.attr('href');
```
xkxd, line 184: var may be omitted

extensions/wikia/Recirculation/js/views/mixedFooter.js
```diff
@@ -115,37 +98,18 @@ define('ext.wikia.recirculation.views.mixedFooter', [
...
+				href = $this.attr('href');
+
 			labels.forEach(function (label) {
-				track({
-					action: tracker.ACTIONS.CLICK,
-					label: label
-				});
+				tracker.trackClick(label);
 			});
+			tracker.trackSelect(href);
```
xkxd, line 112: would it be reasonable to extract that callback as we use the same for rail and mcf?

xkxd (approved):

PR #16563, opened 2019-01-14
IW-1509 | Fix thumbnails in Rail recirculation

mszabo-wikia:
Ensure all thumbnails share the same dimensions

https://wikia-inc.atlassian.net/browse/IW-1509

@Wikia/iwing

### DIFF
--- a/extensions/wikia/Recirculation/RailContentService.php
+++ b/extensions/wikia/Recirculation/RailContentService.php
@@ -14,13 +14,16 @@ class RailContentService extends WikiaService {
 		);
 
 		foreach ( $articles as $pageId => $article ) {
-			$urlGenerator = ( new VignetteUrlToUrlGenerator( $article['thumbnail'] ) )->build();
-
-			$articles[$pageId]['thumbnail'] =
-				$urlGenerator->zoomCrop()
-					->width( static::RAIL_THUMBNAIL_WIDTH )
-					->height( static::RAIL_THUMBNAIL_HEIGHT )
-					->url();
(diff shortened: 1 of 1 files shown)


extensions/wikia/Recirculation/RailContentService.php
```diff
@@ -2,9 +2,26 @@
...
 		$popularPagesService = new PopularPagesService();
-		$articles = $popularPagesService->getPopularPages( 5, 80, 50 );
+		$articles = $popularPagesService->getPopularPages(
+			5,
+			static::RAIL_THUMBNAIL_WIDTH,
+			static::RAIL_THUMBNAIL_HEIGHT
+		);
+
+		foreach ( $articles as $pageId => $article ) {
+			$gen = ( new VignetteUrlToUrlGenerator( $article['thumbnail'] ) )->build();
```
rybmat: `$gen` is not very descriptive name ;) how about `$urlGenerator`?

rybmat (approved):

extensions/wikia/Recirculation/RailContentService.php
latata: you can use `VignetteRequest::fromUrl()` here. I'd also wrap it in `try...catch`
Like here: https://github.com/Wikia/app/blob/f74556613331e60006a4a7a49f6368657a8625f9/extensions/wikia/FeedsAndPosts/Thumbnails.class.php#L6

rybmat (approved):

PR #16565, opened 2019-01-14
Platform 3914

abador:
Redirect anon to a specific article containing searched image. If there is no suitable page or privileges don't allow public read redir to mainpage.

### DIFF
--- a/extensions/wikia/FilePage/WikiaFilePage.php
+++ b/extensions/wikia/FilePage/WikiaFilePage.php
@@ -85,12 +85,12 @@ class WikiaFilePage extends ImagePage {
 			}
 		}
 
-		if ( !$img ) {
-
-            $out->redirect( $url );
+		if ( !$img || $img && !$img->fileExists ) {
+            parent::view();
 
 			return;
 		}
+
 		$res = $this->fetchLinks( $img->getTitle()->getDBkey() );
 
 		foreach ( $res as $row ) {
--- a/extensions/wikia/FilePage/WikiaFilePage.php
+++ b/extensions/wikia/FilePage/WikiaFilePage.php
@@ -60,7 +60,7 @@ class WikiaFilePage extends ImagePage {
 		}
 		$redirKey = "redir:".$wgDBprefix.":".$this->getTitle();
 		$url = $wgMemc->get($redirKey);
-		if(url) {
+		if($url) {
 			$wgOut->redirect($url);
 		}
 		$displayImg = $img = false;
@@ -81,11 +81,11 @@ class WikiaFilePage extends ImagePage {
 			$dbr->select( [ 'imagelinks', 'page' ], [
 					'page_title',
 				], [ 'il_to' => $img->getTitle()->getDBkey(), 'page_is_redirect' => 0, 'il_from = page_id' ], __METHOD__,
-				[ 'LIMIT' => 3, 'ORDER BY' => 'page_namespace, page_id', ] );
-
+				[ 'LIMIT' => 5, 'ORDER BY' => 'page_namespace, page_id', ] );
 
 		$res->seek(0);
 		$row = $res->current();
+		$page = WikiPage::loadFromRow($row);
 		$res->free();
 		$url =  wfExpandUrl("/wiki/".$row->page_title);
 		$wgMemc->add($redirKey, $url);
--- a/extensions/wikia/FilePage/WikiaFilePage.php
+++ b/extensions/wikia/FilePage/WikiaFilePage.php
@@ -63,6 +63,7 @@ class WikiaFilePage extends ImagePage {
         var_dump("2");
 		//fallback to main page
 		$url = Title::newMainPage()->getFullURL();
+        var_dump($url);
 		//wiki needs read privileges
 		if ( !$this->getTitle()->userCan( 'read' ) ) {
             var_dump("21");
--- a/extensions/wikia/FilePage/WikiaFilePage.php
+++ b/extensions/wikia/FilePage/WikiaFilePage.php
@@ -74,11 +74,11 @@ class WikiaFilePage extends ImagePage {
 		$redirKey = wfMemcKey( 'redir', $this->getTitle()->getPrefixedText() );
 
         var_dump("3");
-		$url = $wgMemc->get( $redirKey );
-		if ( $url ) {
+		$urlMem = $wgMemc->get( $redirKey );
+		if ( $urlMem ) {
 
             var_dump("31");
-			$out->redirect( $url );
+			$out->redirect( $urlMem );
 
 			return;
 		}
--- a/extensions/wikia/FilePage/WikiaFilePage.php
+++ b/extensions/wikia/FilePage/WikiaFilePage.php
@@ -118,6 +118,7 @@ class WikiaFilePage extends ImagePage {
         var_dump("7");
 		$wgMemc->add( $redirKey, $url );
 		$out->redirect( $url );
+        var_dump($url);
 	}
 
 	/**
--- a/extensions/wikia/FilePage/WikiaFilePage.php
+++ b/extensions/wikia/FilePage/WikiaFilePage.php
@@ -52,7 +52,7 @@ class WikiaFilePage extends ImagePage {
 	 * Render the image or video
 	 */
 	public function view() {
-		global $wgMemc, $wgDBprefix;
+		global $wgMemc;
 		$out = $this->getContext()->getOutput();
 		if ( !$this->getContext()->getUser()->isAnon() ) {
 			parent::view();
@@ -67,7 +67,7 @@ class WikiaFilePage extends ImagePage {
 
 			return;
 		}
-		$redirKey = wfMemcKey( "redir", $wgDBprefix, $this->getTitle()->getPrefixedText() );
+		$redirKey = wfMemcKey( 'redir', $this->getTitle()->getPrefixedText() );
 
 		$url = $wgMemc->get( $redirKey );
 		if ( $url ) {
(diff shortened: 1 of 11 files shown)


extensions/wikia/FilePage/WikiaFilePage.php
```diff
@@ -40,29 +40,96 @@ protected function isVideo() {
...
+	 */
+	public function view() {
+		global $wgMemc, $wgDBprefix, $wgOut, $wgUser, $wgGroupPermissionsLocal, $wgNamespaceProtection;
+		if ( !$wgUser->isAnon() ) {
+			parent::view();
+
+			return;
+		}
+		//fallback to main page
+		$url = wfExpandUrl( "/wiki/Main_Page" );
```
Grunny: `Title::newMainPage()->getFullURL();` is preferred, since the main page may not always be located at the title "Main Page".

extensions/wikia/FilePage/WikiaFilePage.php
```diff
@@ -40,29 +40,96 @@ protected function isVideo() {
...
+		//wiki needs read privileges
+		if ( preg_match( "/\*\|read\|0/", $wgGroupPermissionsLocal ) ||
```
Grunny: `$this->getTitle()->userCan( 'read' );`

extensions/wikia/FilePage/WikiaFilePage.php
```diff
@@ -40,29 +40,96 @@ protected function isVideo() {
...
+			 preg_match( "/\user\|read\|0/", $wgGroupPermissionsLocal ) ) {
+			$wgOut->redirect( $url );
+
+			return;
+		}
+		$redirKey = "redir:" . $wgDBprefix . ":" . $this->getTitle();
```
Grunny: I'd use `wfMemcKey` to build this, since then that makes sure the key is varied on environment and per wiki for you. I'd also use `$this->getTitle()->getPrefixedText()` rather than relying on `toString`.

extensions/wikia/FilePage/WikiaFilePage.php
```diff
@@ -40,29 +40,96 @@ protected function isVideo() {
...
+		$url = $wgMemc->get( $redirKey );
+		if ( $url ) {
+			$wgOut->redirect( $url );
```
Grunny: Should there be a return here? The rest of the method will still be executed after `OutputPage::redirect` is called I believe.

extensions/wikia/FilePage/WikiaFilePage.php
```diff
@@ -40,29 +40,96 @@ protected function isVideo() {
...
+			return;
+		}
+		$res = $this->fetchLinks( $img->getTitle()->getDBkey() );
+
+		foreach ( $res as $row ) {
+			$title = Title::newFromRow( $row );
+			if ( $title->isRedirect() ) {
+				continue;
+			}
+			$url = wfExpandUrl( "/wiki/" . $row->page_title );
```
Grunny: `$title->getFullURL()` should be used to get hte URL from the title. Hardcoding things like `/wiki/` will break situations like language path wikis, or if we ever changed `wgArticlePath`, so using the methods to build the URL will account for all that for you.

extensions/wikia/FilePage/WikiaFilePage.php
```diff
@@ -40,29 +40,96 @@ protected function isVideo() {
...
+			parent::view();
+
+			return;
+		}
+		//fallback to main page
+		$url = wfExpandUrl( "/wiki/Main_Page" );
+		//wiki needs read privileges
+		if ( preg_match( "/\*\|read\|0/", $wgGroupPermissionsLocal ) ||
+			 preg_match( "/\user\|read\|0/", $wgGroupPermissionsLocal ) ) {
+			$wgOut->redirect( $url );
```
Grunny: `$this->getContext()->getOutput()->redirect( ... );` is preferred over using the global in newer MediaWiki code.

extensions/wikia/FilePage/WikiaFilePage.php
```diff
@@ -40,29 +40,96 @@ protected function isVideo() {
...
 			parent::openShowImage();
 		}
 	}
 
+	/**
+	 * Render the image or video
+	 */
+	public function view() {
+		global $wgMemc, $wgDBprefix, $wgOut, $wgUser, $wgGroupPermissionsLocal, $wgNamespaceProtection;
+		if ( !$wgUser->isAnon() ) {
```
Grunny: `$this->getContext()->getUser()->isAnon()` is preferred over using the global in newer versions of MW.

extensions/wikia/FilePage/WikiaFilePage.php
```diff
@@ -40,29 +40,96 @@ protected function isVideo() {
...
+	 * @param string $dbKey
+	 * @return string
+	 */
+	private function fetchLinks( $dbKey ) {
+		$dbr = wfGetDB( DB_SLAVE );
+
+		return $dbr->select( [ 'imagelinks', 'page' ], [
+			'page_title',
+			'page_namespace',
+		], [ 'il_to' => $dbKey, 'page_is_redirect' => 0, 'il_from = page_id' ], __METHOD__,
```
Grunny: I think we should consider limiting this to article pages only, so `'page_namespace' => NS_MAIN,` rather than redirecting to pages that potentially won't be indexed or aren't valuable. We can double check if this should be the case.

I'd also suggest reformatting this to be a little more in line with similar code elsewhere which is a bit more readable:
```php
return $dbr->select(
	[ 'imagelinks', 'page' ],
	[ 'page_title', 'page_namespace' ],
	[
		'il_to' => $dbKey,
		'page_is_redirect' => 0,
		'il_from = page_id'
	],
	__METHOD__,
	[
		'LIMIT' => 5,
		'ORDER BY' => 'page_namespace, page_id',
	]
);
```

Grunny (changes requested):

extensions/wikia/FilePage/WikiaFilePage.php
```diff
@@ -40,29 +40,105 @@ protected function isVideo() {
...
+		}
+		//fallback to main page
+		$url = Title::newMainPage()->getFullURL();
+		//wiki needs read privileges
+		if ( !$this->getTitle()->userCan( 'read' ) ) {
+			$this->getContext()->getOutput()->redirect( $url );
+
+			return;
+		}
+		$redirKey = wfMemcKey( "redir:", $wgDBprefix, $this->getTitle()->getPrefixedText() );
```
Grunny: `$wgDBprefix` is unneeded, since `wfMemcKey` will prefix the memcache key with `$wgDBname-$wgDBprefix`. `"redir:"` can also just be `'redir'` since it joins the parameters with a colon.

extensions/wikia/FilePage/WikiaFilePage.php
```diff
@@ -40,29 +40,105 @@ protected function isVideo() {
...
+			$title = Title::newFromRow( $row );
+			if ( $title->isRedirect() ) {
+				continue;
+			}
+			$url = $title->getFullURL();
+			$res->free();
+			break;
+		}
+		$wgMemc->add( $redirKey, $url );
+		$this->getContext()->getOutput()->redirect( $url );
```
Grunny: I think it's worth putting `$this->getContext()->getOutput()` in a variable since it's used multiple times.

Grunny (changes requested):
Some minor comments, otherwise looks good.

Grunny (approved):

includes/specials/SpecialUndelete.php
```diff
@@ -1452,6 +1453,9 @@ function undelete() {
 			$out->addWikiMsg( 'undeleterevdel' );
 		}
 
+		$redirKey = wfMemcKey( 'redir', $this->mTargetObj->getPrefixedText() );
+		$wgMemc->delete( $redirKey );
```
Grunny: Using the `UndeleteComplete` hook to do this would be preferred over modifying the core MW file and having this code know about an extension-specific Memcache key.

includes/FileDeleteForm.php
```diff
@@ -134,6 +134,9 @@ public static function doDelete( &$title, &$file, &$oldimage, $reason, $suppress
 			global $wgUser;
 			$user = $wgUser;
 		}
+		global $wgMemc;
+		$redirKey = wfMemcKey( 'redir', $title->getPrefixedText() );
+		$wgMemc->delete( $redirKey );
```
Grunny: Using the `FileDeleteComplete` hook to do this would be preferred over modifying the core MW file and having this code know about an extension-specific Memcache key.

Grunny (changes requested):

includes/WikiPage.php
```diff
@@ -2241,6 +2248,29 @@ public function doDeleteUpdates( $id ) {
 		# Wikia
 	}
 
+	/**
+	 * Clear memcache redirs before db changed
+	 *
+	 * @param $id Int: page_id value of the page being deleted
+	 */
+	public function doClearLinkedFilesCache( $id ) {
```
Grunny: Likewise to the above comments, this should all live in a hook method in an extension (the FilePage extension is fine for this), rather than modifying core MediaWiki files, which shouldn't know about the file page redirect memcache entry.

includes/specials/SpecialUndelete.php
```diff
@@ -1412,7 +1413,7 @@ function getFileComment( $file ) {
 	}
 
 	function undelete() {
-		global $wgUploadMaintenance;
+		global $wgUploadMaintenance, $wgMemc;
```
Grunny: Unused.

Grunny (changes requested):
Additional comments to the ones above.

abador:
I've created a new branch and PR. This branch is too far behind, also I used this branch for debugging and It's too messy. new pr: https://github.com/Wikia/app/pull/16651

PR #16570, opened 2019-01-16
(PLATFORM-3905) Convert embedded Special:FilePath URLs to Vignette URLs

Grunny:
This will attempt to fix plain HTTP Special:FilePath URLs in the simplest
cases by converting them to HTTPS Vignette URLs.

This will not account for non-English translations of Special:FilePath or
language path wiki URLs (which should only have ever been on HTTPS).

/cc @Wikia/core-platform-team

### DIFF
--- a/extensions/wikia/HTTPSSupport/HTTPSSupportHooks.class.php
+++ b/extensions/wikia/HTTPSSupport/HTTPSSupportHooks.class.php
@@ -96,6 +96,55 @@ class HTTPSSupportHooks {
 	}
 
 	/**
+	 * Hacky support for some non-HTTPS Special:FilePath URLs.
+	 *
+	 * This will attempt to fix plain HTTP file path URLs in the simplest
+	 * cases by converting them to HTTPS Vignette URLs.
+	 *
+	 * This will not account for non-English translations of Special:FilePath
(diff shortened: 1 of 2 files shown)


extensions/wikia/HTTPSSupport/HTTPSSupportHooks.class.php
```diff
@@ -95,6 +95,55 @@ public static function parserUpgradeVignetteUrls( string &$url ) {
...
+		// here is necessary, rather than using article path (which could be different on
+		// the current wiki). We will also not support translated versions of "Special:FilePath"
+		$filePathPrefix = '/wiki/Special:FilePath/';
+		$path = urldecode( $parts['path'] );
+
+		if (
+			preg_match( "/\\.{$wgWikiaBaseDomainRegex}$/", $parts['host'] )
+			&& isset( $parts['scheme'] )
+			&& $parts['scheme'] === 'http'
+			&& stripos( $path, $filePathPrefix ) === 0
```
themech, line 129: I'm not a big fan of using Special:FilePath so it makes me think - cannot we use this for https and fandom domain? :)

themech (approved):

PR #16571, opened 2019-01-16
IW-1518 | Fix saving attributes for multiple users

mszabo-wikia:
The UAS client stores a 2D array of attributes (user ID, name). However, a 1D array was  used to store attributes that were changed during a request, which meant that users' attributes could be mixed up in situations where attributes were changed for user(s) other than the request user. By making sure a 2D array is used in this case as well, we avoid this.

@Wikia/iwing 

https://wikia-inc.atlassian.net/browse/IW-1518


jsutterfield (commented):
LGTM

PR #16572, opened 2019-01-16
CORE-60 | Remove restriction of moving pages out of User_blog namespace

JamesHawking:
https://wikia-inc.atlassian.net/browse/CORE-60

### DIFF
--- a/extensions/wikia/Blogs/tests/BlogsHelperTest.php
+++ b/extensions/wikia/Blogs/tests/BlogsHelperTest.php
@@ -8,8 +8,7 @@ class BlogsHelperTest extends WikiaBaseTest {
 
 	/**
 	 * SUS-1635: Regression test for BlogsHelper::AbortMove hook handler
-	 * Verify blog pages and comments can be renamed, pages can be moved into blog namespace and that pages can't be moved
-	 * out of blogs namespaces or into comment namespace
+	 * Verify blog pages and comments can be renamed, pages can be moved into blog namespace and back.
 	 *
 	 * @covers BlogsHelper::onAbortMove()
 	 * @dataProvider blogsMoveRestrictionsDataProvider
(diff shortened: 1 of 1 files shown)


macbre:
@JamesHawking, unit tests should be updated as well:

```
1) BlogsHelperTest::testBlogsMoveRestrictions with data set "renaming user blog" (500, 500, true)
Expectation failed for method name is equal to "getNamespace" when invoked at least once.
Expected invocation at least once but it never occurred.
```

extensions/wikia/Blogs/BlogsHelper.class.php
```diff
@@ -163,12 +163,6 @@ public static function onTitleMoveComplete( Title $oOldTitle, Title $oNewTitle,
 	 * @return bool false if we're trying to move out of or into blog comment namespace, true otherwise
 	 */
 	public static function onAbortMove( Title $sourceTitle, Title $targetTitle, User $user, &$err, string $reason ): bool {
-		$blogsNS = [ NS_BLOG_ARTICLE, NS_BLOG_ARTICLE_TALK ];
-		if ( $sourceTitle->inNamespaces( $blogsNS ) && !$targetTitle->inNamespace( $sourceTitle->getNamespace() ) ) {
-			$err = wfMessage( 'immobile-source-namespace', $sourceTitle->getNsText() )->escaped();
-			return false;
-		}
-
```
macbre, line 171:
```
SUS-260: Prevent moving pages into User blog comment namespace, or moving comments out of this namespace
```

This comment needs to be updated to avoid confusion in the future.

extensions/wikia/Blogs/tests/BlogsHelperTest.php
```diff
@@ -8,8 +8,7 @@ public function setUp() {
 
 	/**
 	 * SUS-1635: Regression test for BlogsHelper::AbortMove hook handler
-	 * Verify blog pages and comments can be renamed, pages can be moved into blog namespace and that pages can't be moved
-	 * out of blogs namespaces or into comment namespace
+	 * Verify blog pages and comments can be renamed, pages can be moved into blog namespace and back.
```
macbre, line 11: Confusing comment :) The code that is tested says that we want to prevent moving pages **into** blog namespace.

PR #16573, opened 2019-01-16
DE-4109 | if cannot determine language id from language code a null is expected

szczeles:


extensions/wikia/DataWarehouse/DataWarehouseEventProducer.class.php
```diff
@@ -344,7 +344,8 @@ public function setLanguage( $lang_code = '' ) {
 		if ( empty( $lang_code ) ) {
 			$lang_code = $this->app->wg->LanguageCode;
 		}
-		$this->mParams['languageId'] = WikiFactory::LangCodeToId($lang_code);
+		$languageIdOrFalse = WikiFactory::LangCodeToId($lang_code);
+		$this->mParams['languageId'] = $languageIdOrFalse !== false ? intval($languageIdOrFalse) : NULL;
```
jcellary, line 348: You could also simply not add languageId if it's not available, but this also works :)

jcellary (approved):

extensions/wikia/DataWarehouse/DataWarehouseEventProducer.class.php
macbre: You can try `intval(false) ?: null` as well (assuming the value of zero is incorrect here).

szczeles: @macbre I'm not 100% sure if language id 0 is invalid - it's currently not used but there is no guarantee it won't be used in the future (as it's a valid integer)

PR #16579, opened 2019-01-17
IW-1471 | Cleanup after getting rid of liftigniter recirculation

rybmat:
https://wikia-inc.atlassian.net/browse/IW-1471

@Wikia/iwing


mszabo-wikia (approved):
LGTM, let's run `composer dump-autoload` to remove LI client classes from generated autoloader files

latata (approved):

PR #16580, opened 2019-01-17
Platform 3913

abador:
Changing the link forced me to change a cache key, so that a logged in and anon user get different pages. I don't like the part of the code that fetches the file, if there is a cleaner option let me know.

### DIFF
--- a/includes/parser/ParserCache.php
+++ b/includes/parser/ParserCache.php
@@ -46,11 +46,14 @@ class ParserCache {
 	 * @param $hash string
 	 * @return mixed|string
 	 */
-	protected function getParserOutputKey( $article, $hash ) {
+	protected function getParserOutputKey( $article, $hash, $isAnon = false ) {
 		global $wgRequest;
 
 		// idhash seem to mean 'page id' + 'rendering hash' (r3710)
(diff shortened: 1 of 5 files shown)


includes/parser/ParserCache.php
```diff
@@ -46,11 +46,14 @@ protected function __construct( $memCached ) {
...
 		global $wgRequest;
 
 		// idhash seem to mean 'page id' + 'rendering hash' (r3710)
 		$pageid = $article->getID();
+		if($isAnon){
+			$pageid = "anon:".$pageid;
+		}
 		$renderkey = (int)($wgRequest->getVal('action') == 'render');
 
 		$key = wfMemcKey( 'pcache', 'idhash', "{$pageid}-{$renderkey}!{$hash}" );
```
mszabo-wikia, line 59: this will invalidate whole parser cache across entire site for anons, do we want to do this?

harnash: @mszabo-wikia we will be serving new and different content for anons so we need to invalidate it. Unless you have better alternatives?

harnash (approved):

harnash (approved):

PR #16583, opened 2019-01-18
IW-1441 | expose community description in FeedsAndPostsController

rybmat:
https://wikia-inc.atlassian.net/browse/IW-1441

In order to get rid of hardcoded data from Feeds, we need to provide this app with description from CommunityData service. Also link to community page (or main page if the first one is not available) is needed as well.

@Wikia/iwing


extensions/wikia/FeedsAndPosts/WikiInfo.class.php
```diff
@@ -0,0 +1,21 @@
...
+namespace Wikia\FeedsAndPosts;
+
+use CommunityDataService;
+use Title;
+
+class WikiInfo {
+	public function get(): array {
+		global $wgCityId, $wgEnableCommunityPageExt;
+
+		$getStartedUrl = !empty($wgEnableCommunityPageExt)
```
jsutterfield: Minor: our PHP guidelines call for padding in all of these function calls with arguments.

jsutterfield (approved):

extensions/wikia/FeedsAndPosts/WikiInfo.class.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+namespace Wikia\FeedsAndPosts;
+
+use CommunityDataService;
+use Title;
+
+class WikiInfo {
+	public function get(): array {
+		global $wgCityId, $wgEnableCommunityPageExt;
```
latata: shouldn't we add some caching here?

rybmat: good catch

latata (approved):

PR #16585, opened 2019-01-18
IW-1517 perform reupload check inside of WikiaMiniUpload::insertImage

jsutterfield:
Ticket: https://wikia-inc.atlassian.net/browse/IW-1517

This PR fixes a vulnerability which would allow a user to circumvent our `reupload` permission using the WMO insertImage API. This PR adds a call to `\UploadBase::userCanReUpload` which does the reupload checks for us.

Also, credit to @mszabo-wikia for identifying where we were missing that permission check and suggesting the proper fix.

cc @Wikia/iwing

### DIFF
--- a/extensions/wikia/WikiaMiniUpload/WikiaMiniUpload_body.php
+++ b/extensions/wikia/WikiaMiniUpload/WikiaMiniUpload_body.php
@@ -492,17 +492,12 @@ class WikiaMiniUpload {
 				}
 				if ( $title->exists() ) {
 					if ( $type == 'overwrite' ) {
-						$title = Title::newFromText($name, 6);
-						// is the target protected?
-						$permErrors = $title->getUserPermissionsErrors( 'edit', $wgUser );
-						$permErrorsUpload = $title->getUserPermissionsErrors( 'upload', $wgUser );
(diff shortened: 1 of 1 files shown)


mszabo-wikia (changes requested):
We need to keep the original checks & apply the reupload check as an addition

extensions/wikia/WikiaMiniUpload/WikiaMiniUpload_body.php
```diff
@@ -747,6 +743,14 @@ function insertImage() {
 		return $tmpl->render( 'summary' );
 	}
 
+	private function userCanOverwrite( Title $title, File $file ) {
+		global $wgUser;
+
+		return empty( $title->getUserPermissionsErrors( 'edit', $wgUser ) )
+			&& empty( $title->getUserPermissionsErrors( 'upload', $wgUser ) )
+			&& ( $file instanceof File && UploadBase::userCanReUpload( $wgUser, $file->getName() ) );
+	}
```
jsutterfield, line 752: Shouldn't need to do a `upload` perm check since we've already verified that the title exists before calling this method.

jsutterfield:
Thanks for catching that @mszabo-wikia! Updated. Could you take another look?

mszabo-wikia (approved):

PR #16586, opened 2019-01-18
Handle exceptions in Nirvana API

themech:
For internal requests with the `X-Wikia-Internal-Request` header,  the dispatcher rethrows exceptions instead of returning the response: https://github.com/Wikia/app/blob/1e4ea22073c29ec97beeccda86d068915366d0c5/includes/wikia/nirvana/WikiaDispatcher.class.php#L236-L240. 
Because there was no error handling in `wikia.php`, this was resulting in empty responses with 200 response status.

### DIFF
--- a/wikia.php
+++ b/wikia.php
@@ -25,7 +25,6 @@ if ( function_exists( 'newrelic_disable_autorum' ) ) {
 	newrelic_disable_autorum();
 }
 
-
 if ( !empty( $wgEnableNirvanaAPI ) ) {
 	// temporarily force ApiDocs extension regardless of config
 	require_once $IP . "/extensions/wikia/ApiDocs/ApiDocs.setup.php";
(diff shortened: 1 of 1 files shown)


wikia.php
```diff
@@ -48,41 +47,56 @@
...
+
+		$response->render();
+	} catch ( WikiaHttpException $e ) {
+		http_response_code( $e->getCode() );
+		if ( $e->getCode() >= WikiaResponse::RESPONSE_CODE_INTERNAL_SERVER_ERROR ) {
+			Wikia\Logger\WikiaLogger::instance()->error( 'Unhandled API error', [
+				'status'=> $e->getCode(),
+				'exception' => $e
+			] );
+		}
```
mszabo-wikia, line 99: Should we render some kind of error body here?

themech: I'm not sure. I think API clients can specify the response format which may also involve a templating engine when generating a response. But I would not like to duplicate the code from `WikiaDispatcher.class.php` here.
We can return a simple JSON response though, just with a status code and error message. This probably won't be used by clients, but may be helpful when debugging. What do you think?

macbre: A simple JSON response sounds good.

macbre (approved):

wikia.php
mszabo-wikia: That can work - alternatively, we could add another field to WikiaRequest that'd truly distinguish between PHP -> PHP nirvana calls and web requests, and update the dispatcher to take that into account when handling exceptions

mszabo-wikia (approved):

wikia.php
themech: Added the JSON response. I think exception handling should be implemented regardless of additional flags - if something is broken, we should not silently ignore it and return 200 responses. Whether we need Yet Another Flag for PHP -> PHP nirvana calls - that is a good question.

PR #16595, opened 2019-01-19
Update AdminDashboard.alias.php

kizule:
Added aliases for Serbian language.


mszabo-wikia:
Hi, thanks for your contribution. Please contact [MtaÄ](https://community.wikia.com/wiki/Message_Wall:Mta%C3%84) about adding these translations.

PR #16598, opened 2019-01-21
IW | check if sponsored item is present

rybmat:
check for presence of sponsoredItem before using its properties

@Wikia/iwing

### DIFF
--- a/extensions/wikia/Recirculation/js/recirculation.js
+++ b/extensions/wikia/Recirculation/js/recirculation.js
@@ -162,7 +162,11 @@ require([
 					$firstItem = $rail.find('.premium-recirculation-rail .thumbnails li').first(),
 					sponsoredItem = sponsoredContentHelper.getSponsoredItem(sponsoredContent);
 
-				if (sponsoredItem.title.length > 90) {
+				if (!sponsoredItem) {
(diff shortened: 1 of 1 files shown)


leshniak (approved):

extensions/wikia/Recirculation/js/recirculation.js
```diff
@@ -162,18 +162,20 @@ require([
...
-					sponsoredItem.shortTitle = sponsoredItem.title;
-				}
-
-				if (sponsoredItem.thumbnailUrl && window.Vignette) {
-					sponsoredItem.thumbnailUrl = window.Vignette.getThumbURL(sponsoredItem.thumbnailUrl, {
-						mode: window.Vignette.mode.zoomCrop,
-						height: 53,
-						width: 53
-					});
+				if (sponsoredItem) {
```
xkxd: id suggest 
```
if (!sponsoredItem) {
   return;
}

...
```

xkxd (approved):

extensions/wikia/Recirculation/js/recirculation.js
rybmat: done

PR #16608, opened 2019-01-23
ADEN-7819 Run AE3 on FV pages

fraszczakszymon:


extensions/wikia/AdEngine3/src/slots.js
```diff
@@ -0,0 +1,247 @@
...
+  return isSupportedPageType &&
+    window.wgIsContentNamespace &&
+    context.get('wiki.opts.adsInContent') &&
+    !context.get('wiki.targeting.wikiIsCorporate');
+}
+
+export default {
+  getContext() {
+    return {
+      TOP_LEADERBOARD: {
```
dmnsobczak, line 30: Don't we wanna switch slot names to lowercase?

dmnsobczak (approved):

extensions/wikia/AdEngine3/AdEngine3.setup.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+$wgExtensionCredits['other'][] = [
+	'name' => 'AdEngine',
+	'author' => 'Wikia',
+	'description' => 'AdEngine',
+	'version' => '1.0',
```
nandy-andy: `3.0` ;)

extensions/wikia/AdEngine3/src/slots.js
fraszczakszymon: Discussed offline. Not at this moment :)

extensions/wikia/AdEngine3/AdEngine3.setup.php
fraszczakszymon: Let it be :D

PR #16569, opened 2019-01-16
PLATFORM-3912 image lightbox

harnash:
Adds an tooltip/dropdown with image description for anonymous users instead of link to the file page.

PR #16610, opened 2019-01-23
Remove share link to file, PLATFORM-3911

abador:
Remove share link to file from lightbox.

### DIFF
--- a/extensions/wikia/Lightbox/templates/Lightbox_lightboxModalContent.php
+++ b/extensions/wikia/Lightbox/templates/Lightbox_lightboxModalContent.php
@@ -128,22 +128,22 @@
 					<?php
 						$formHeader = array (
 							'inputs' => array (
-							array(
-								'type' => 'custom',
-								'output' => '<h2>'. wfMessage('lightbox-urls-form-header')->parse() .'</h2>',
-							),
+								array(
+									'type' => 'custom',
+									'output' => '<h2>'. wfMessage('lightbox-urls-form-header')->parse() .'</h2>',
+								),
 							),
 						);
 					?>
 					<?php
 						$formFilePage = array (
 							'inputs' => array (
-							array(
-								'label' => wfMessage('lightbox-file-page-url')->parse(),
-								'type' => 'text',
-								'name' => 'lightbox-file-page-url',
-								'value' => "{{fileUrl}}",
-							)
+								array(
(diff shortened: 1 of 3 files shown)


harnash (changes requested):
@abador I think your change is not exactly what is in the Acceptance Criterias of the ticket:
```
• File Page URL section is not displayed for logged out users.
```
Your change will affect both: logged in users and anons alike.
This part is completely missing too:
```
• Acknowledgement: if a file page link is shared by a logged in user, logged out users who click it will end up on the first article with that image, per SEO-706. 
```

extensions/wikia/Lightbox/templates/Lightbox_lightboxModalContent.php
```diff
@@ -123,6 +123,36 @@
 				</div>
 			</div>
 			<div class="bottom-forms">
+                <?php
+                    if(!$wg->Out->getContext()->getUser()->isAnon()) {
```
harnash: 1. There is already a property added in #16569 you could and should use.
2. Mustache templates have conditionals - use them instead of falling back to PHP (see: https://mustache.github.io/mustache.5.html)

harnash (changes requested):

extensions/wikia/Lightbox/templates/Lightbox_lightboxModalContent.php
```diff
@@ -123,32 +123,36 @@
...
+                ?>
+                <div class="more-links">
+                    <?php
+                    $formHeader = array (
+                        'inputs' => array (
+                            array(
+                                'type' => 'custom',
+                                'output' => '<h2>'. wfMsg('lightbox-urls-form-header') .'</h2>',
```
harnash: `wfMsg` is deprecated please use `wfMessage` with proper escaping.

extensions/wikia/Lightbox/templates/Lightbox_lightboxModalContent.php
abador: Different scope, but I've added something similar

extensions/wikia/Lightbox/templates/Lightbox_lightboxModalContent.php
abador: changed

harnash (approved):

ludwikkazmierczak (approved):

PR #16620, opened 2019-01-25
IW-1261 | Design System icons update

ustczanka:
https://wikia-inc.atlassian.net/browse/IW-1261

And here is icons changelog:
https://docs.google.com/spreadsheets/d/1LwmpRYpHyzV-C3gHoAXjNnUjeF0KPAa0iG_lphSWWno/edit#gid=458506291


extensions/wikia/BannerNotifications/js/BannerNotifications.js
```diff
@@ -19,22 +19,30 @@
 			'<div class="wds-banner-notification__icon">' +
 			'{{{icon}}}</div>' +
 			'<span class="wds-banner-notification__text">{{{content}}}</span>' +
-			// DS icon: wds-icons-cross-tiny
+			// DS icon: wds-icons-close-tiny
```
rybmat, line 22: maybe it would be worth writing some js helper to render DS icons?

extensions/wikia/PageHeader/styles/index.scss
```diff
@@ -59,8 +59,16 @@
 		padding-bottom: 12px;
 
 		.wds-dropdown__toggle-chevron {
+			margin-left: 4px;
+			transform: scale(.75);
```
rybmat, line 63: why such style? shouldn't we use just smaller icon?

extensions/wikia/PageHeader/styles/index.scss
```diff
@@ -59,8 +59,16 @@
...
 			vertical-align: text-top;
 		}
+
+		&:hover {
+			.wds-dropdown__toggle-chevron {
+				transform: rotate(180deg) scale(.75);
```
rybmat, line 69: same as above

rybmat (changes requested):

extensions/wikia/PageHeader/styles/index.scss
ustczanka: we don't have smaller icon, this is the smallest one... but it is a bit bigger than the one we have on production right now and Megan asked to make it smaller. I don't like it either, but it doesn't look good without this minimizing.

extensions/wikia/PageHeader/styles/index.scss
ustczanka: same as above :)

rybmat (approved):

extensions/wikia/PageHeader/styles/index.scss
latata: shouldn't this be done in DS?

extensions/wikia/PageHeader/styles/index.scss
latata: shouldn't it be done in DS

extensions/wikia/PageHeader/styles/index.scss
ustczanka: I don't think so. Right now we use this xxs icon in only one place, and I think this is an exception,  which we should not copy.

extensions/wikia/PageHeader/styles/index.scss
ustczanka: same as above

PR #16647, opened 2019-02-04
CORE-86 | COPPA tool leaves a remnant of the user's email

macbre:
https://wikia-inc.atlassian.net/browse/CORE-86


chessky (approved):

chessky (commented):
We should first remove the mail and only after delete the log.

PR #16648, opened 2019-02-04
IW-1338 | Add Feeds landing page preference

mszabo-wikia:
Redirect to `/f` based on value of landing page preference
https://wikia-inc.atlassian.net/browse/IW-1338


latata (approved):
LGTM. Consider localise the new option string.

PR #16649, opened 2019-02-04
ADEN-8232 Add bidder_won, bidder_won_price and scroll_y to AdEngine3 slot tracker.

jacekbj:


extensions/wikia/AdEngine3/src/tracking/slot-tracker.js
```diff
@@ -63,7 +68,10 @@ function prepareData(slot, data) {
     opt_in: checkOptIn(),
     document_visibility: utils.getDocumentVisibilityStatus(),
     // Missing:
-    // bidder_won, bidder_won_price, page_layout, rabbit, scroll_y, product_chosen
+    // page_layout, rabbit, product_chosen
+    bidder_won: slotParams.hb_bidder || '',
```
fraszczakszymon: We were already there. This approach (taking params from DOM element) is problematic when we're collapsing slot before it's created/dispatching some custom event when DOM isn't created. I believe we're storing prices received before call in `realPrices` map. Could we reuse it?

fraszczakszymon (approved):

fraszczakszymon (approved):

fraszczakszymon (approved):

dmnsobczak (approved):

PR #16651, opened 2019-02-04
Platform-3914: Release the file pages

abador:
Clean branch with changes to redirects of file pages and additional hooks cleaning caches after article changes


extensions/wikia/CachedRedirects/CachedRedirects.class.php
```diff
@@ -0,0 +1,113 @@
...
+
+
+	/**
+	 * Hook to fetch linked materials
+	 *
+	 * @param $id Int: page_id value of the page being deleted
+	 * @param $links container for links
+	 */
+	public static function onGetFileLinks( $id, &$links ) {
+		$links = self::getFileLinks( $id );
```
harnash: No return value here?

extensions/wikia/CachedRedirects/CachedRedirects.class.php
```diff
@@ -0,0 +1,113 @@
...
+	 *
+	 * @param Title $title -- instance of Title class
+	 *
+	 */
+	private static function purgeRedir( $title ) {
+		global $wgMemc;
+		$redirKey = wfMemcKey( 'redir', $title->getPrefixedText() );
+		$wgMemc->delete( $redirKey );
+		$page = WikiPage::factory( $title );
+		$page->doPurge();
```
harnash: No return value

extensions/wikia/FilePage/WikiaFilePage.php
```diff
@@ -148,4 +148,87 @@ protected function newPage( Title $title ) {
...
+			$title = Title::newFromRow( $row );
+			if ( $title->isRedirect() ) {
+				continue;
+			}
+			if ( !$title->userCan( 'read' ) ) {
+				$out->redirect( $url , '301');
+				continue;
+			}
+			$url = $title->getFullURL();
+			$res->free();
```
harnash: You do not free the resource when one of the above if statements is true. Is that intentional?

includes/WikiPage.php
```diff
@@ -1554,6 +1560,12 @@ public function doEdit( $text, $summary, $flags = 0, $baseRevId = false, $user =
 		// Return the new revision (or null) to the caller
 		$status->value['revision'] = $revision;
 
+		$links2= null;
```
harnash: Is that correct? Seems like $links2 is unused.

harnash (changes requested):

extensions/wikia/CachedRedirects/CachedRedirects_setup.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+$wgExtensionCredits['other'][] = array(
```
Grunny: Given all the cache is used inside the FilePage extension in the WikiaFilePage class, I think it would be better to contain all the file page improvement logic in that one extension, rather than having the cache entries used and set by one extension and purged by a different extension.

includes/WikiPage.php
```diff
@@ -1297,6 +1297,12 @@ public function doEdit( $text, $summary, $flags = 0, $baseRevId = false, $user =
 
 		$flags = $this->checkFlags( $flags );
 
+		$links = null;
+		Hooks::run( 'ArticleGetFileLinks', [ $this->mTitle->getArticleID(), &$links ] );
```
Grunny: I don't think this flow is a good way to handle this. Having a hook populate a variable to be used by a later hook is a very complex flow. I assume this is because the data is gone to generate the links list by this stage. I think instead, we should either purge the cache on the earlier `ArticleDelete` hook (it's probably OK if a few purges happen when the article isn't successfully deleted), or at worst, add a hook in a similar place to `ArticleGetFileLinks` that does the purge there, but I think `ArticleDelete` would be fine.

And why is `ArticleGetFileLinks` used twice?

Grunny (changes requested):

extensions/wikia/CachedRedirects/CachedRedirects.class.php
abador: It's a pointer

extensions/wikia/FilePage/WikiaFilePage.php
abador: I think it can be deleted. Right now it isn't needed.

extensions/wikia/CachedRedirects/CachedRedirects.class.php
abador: to be added

includes/WikiPage.php
abador: to be deleted

includes/WikiPage.php
abador: We have two cases: a new link is added(it seems right now i doesn't clear all the data from cache, so we need to run a full purge on all the links ) and an old link is deleted (that's why I've added the hook to populate links before changes happen). Purging cache before the data is deleted from the db might cause caching of an old redirect (sometimes taskPublisher->doUpdate() fires before action ends and instead of one request to celery we have few more). I'll try to check it tomorrow

extensions/wikia/CachedRedirects/CachedRedirects_setup.php
abador: ok

extensions/wikia/CachedRedirects/CachedRedirects.class.php
abador: Phpdoc should be updated, it's not a hook anymore. The return value isn't used anywhere, but I can add some logs based on those actions

abador (commented):
It seems I must have forced sending data to celery while I was debuging. For now the requests are sent after the page loop finished. thanks :)

includes/WikiPage.php
abador: It seems I must have forced sending data to celery while I was debuging. For now the requests are sent after the page loop finished. thanks :)

extensions/wikia/FilePage/FilePageController.class.php
```diff
@@ -73,6 +73,92 @@ public function fileUsage() {
...
+			if ( $title->isRedirect() ) {
+				continue;
+			}
+			if ( !$title->userCan( 'read' ) ) {
+				continue;
+			}
+			$url = $title->getFullURL();
+			break;
+		}
+		$wgMemc->add( $redirKey, $url );
```
harnash: Shouldn't we check if `$url` is actually populated correctly? In worst case it will be the value from `Title::newMainPage()->getFullURL()` which can cause some issues.

extensions/wikia/FilePage/FilePageController.class.php
```diff
@@ -73,6 +73,92 @@ public function fileUsage() {
...
+			return;
+		}
+
+		$urlMem = $wgMemc->get( $redirKey );
+		if ( $urlMem ) {
+			$this->url = $urlMem;
+			return;
+		}
+
+		$res = $this->fetchLinks( $img->getTitle()->getDBkey() );
```
harnash: IIRC this can result `false` so `foreach` below may have some issues when trying to iterate over it

extensions/wikia/FilePage/FilePageController.class.php
```diff
@@ -73,6 +73,92 @@ public function fileUsage() {
...
+
+		wfProfileOut( __METHOD__ );
+	}
+
+
+
+	/**
+	 * Fetch informationabout pages linked to image
+	 * @param string $dbKey
+	 * @return string
```
harnash: That seems incorrect

extensions/wikia/FilePage/FilePageHooks.class.php
```diff
@@ -173,4 +173,141 @@ static public function onGlobalUsageLinksUpdateComplete( &$images ) {
...
+
+	/**
+	 * getFileLinks get links to material
+	 *
+	 * @param $id Int: page_id value of the page being deleted
+	 */
+	public static function getFileLinks( $id ) {
+		$dbr = wfGetDB( DB_SLAVE );
+
+		return $dbr->select(
```
harnash, line 295: please add `@return` section to do PHPDoc

extensions/wikia/MercuryApi/handlers/MercuryApiFilePageHandler.class.php
```diff
@@ -48,7 +48,15 @@ function ( $acc, $child ) {
 			return $item;
 		}, $fileUsageData['fileList']);
 
+		$anonRedir = F::app()->sendRequest(
```
harnash: We might check if the call actually succeeds before fetching and using the data

extensions/wikia/FilePage/WikiaFilePage.php
```diff
@@ -152,83 +152,17 @@ protected function newPage( Title $title ) {
...
-			'il_from = page_id',
-		], __METHOD__, [
-			'LIMIT' => 5,
-			'ORDER BY' => 'page_namespace, page_id',
-		] );
+		$data = F::app()->sendRequest(
+			'FilePage',
+			'fileRedir',
+			[ 'type' => 'local', 'format' => 'json' ]
+		)->getData();
```
harnash: `sendRequest` may fail - we should check if there is actual data to be used

harnash (changes requested):

extensions/wikia/FilePage/FilePageController.class.php
```diff
@@ -73,6 +73,92 @@ public function fileUsage() {
...
 
+	/**
+	 * Collects data about what articles the current file appears in, either
+	 * locally and globally.
+	 *
+	 * @requestParam string type (optional) Possible values are 'global' or 'local', defaulting to 'local' if not given.
+	 *                                      Determines whether this collects a local or global file usage data
+	 */
+	public function fileRedir() {
+		wfProfileIn( __METHOD__ );
```
Grunny: Minor: We don't need `wfProfileIn`/`wfProfileOut` anymore.

extensions/wikia/FilePage/FilePageHooks.class.php
```diff
@@ -173,4 +173,141 @@ static public function onGlobalUsageLinksUpdateComplete( &$images ) {
...
+	 * Hook to clear caches for linked materials
+	 *
+	 * @param Title $title -- instance of Title class
+	 * @param User $user -- current user
+	 * @param string $reason -- undeleting reason
+	 *
+	 * @return true -- because it's hook
+	 */
+	public static function onUndeleteComplete( Title $title ) {
+		self::purgeTitle($title);
```
Grunny: Minor: Should be spacey within brackets per conventions (here and in the other calls).

extensions/wikia/FilePage/FilePageHooks.class.php
```diff
@@ -173,4 +173,141 @@ static public function onGlobalUsageLinksUpdateComplete( &$images ) {
...
+		$page->doPurge();
+	}
+
+
+	/**
+	 * getFileLinks get links to material
+	 *
+	 * @param $id Int: page_id value of the page being deleted
+	 */
+	public static function getFileLinks( $id ) {
```
Grunny: Minor: Should be `private`.

extensions/wikia/FilePage/FilePageHooks.class.php
```diff
@@ -173,4 +173,141 @@ static public function onGlobalUsageLinksUpdateComplete( &$images ) {
...
+			__METHOD__,
+			[ 'ORDER BY' => 'il_to', ] );
+	}
+
+	/**
+	 * Clear memcache redirs before db changed
+	 *
+	 * @param $id Int: page_id value of the page being deleted
+	 */
+	public static function clearLinkedFilesCache( $id, $results = null ) {
```
Grunny: Minor: Should be `private`.

extensions/wikia/FilePage/FilePageHooks.class.php
```diff
@@ -173,4 +173,141 @@ static public function onGlobalUsageLinksUpdateComplete( &$images ) {
...
+
+
+	/**
+	 * Clear memcache and purge page
+	 *
+	 * @param Title $title -- instance of Title class
+	 *
+	 */
+	private static function purgeTitle( Title $title ) {
+		if ( $title->getNamespace() == NS_FILE ) {
```
Grunny: Minor: `$title->inNamespace( NS_FILE )` is preferred over comparing to `getNamespace`.

includes/wikia/nirvana/WikiaDispatchableObject.class.php
```diff
@@ -314,8 +314,13 @@ public static function getLocalUrl( $method, $params = null, $format = null ) {
 	 * @return string The absolute URL
 	 */
 	public static function getFullUrl( $method, $params = null, $format = null ) {
+		global $wgWikiaEnvironment;
 		$app = F::app();
-		return wfExpandUrl( $app->wg->Server . self::getLocalUrl( $method, $params, $format ) );
+		$url = wfExpandUrl( $app->wg->Server . self::getLocalUrl( $method, $params, $format ) );
+		if ( $wgWikiaEnvironment != WIKIA_ENV_PROD ) {
+			$url = preg_replace("/^https/", "http", $url);
```
Grunny: `wfHttpsToHttp` (makes grepping for everywhere we are doing this easier).

Grunny (changes requested):

extensions/wikia/MercuryApi/handlers/MercuryApiFilePageHandler.class.php
abador: As far as I can tell this usage is standard and I can't see any place to check if it succeeded. I've added a check if $data['url'] is set.

extensions/wikia/FilePage/FilePageController.class.php
abador: I've changed the logic a bit so everything works on $this->url. It must be populated at start with the mainpage url. The changes to image purges should take care of redirecting to the mainpage. The problem I can see is deploing this branch without PLATFORM-3932 @Grunny . After memcache stores a mainpage redirect it will stay there until it's purged.

extensions/wikia/FilePage/FilePageController.class.php
```diff
@@ -73,6 +73,97 @@ public function fileUsage() {
 		wfProfileOut( __METHOD__ );
 	}
 
+	/**
+	 * Collects data about what articles the current file appears in, either
+	 * locally and globally.
+	 *
+	 * @requestParam string type (optional) Possible values are 'global' or 'local', defaulting to 'local' if not given.
```
Grunny: All this documentation is for a different method.

extensions/wikia/FilePage/WikiaFilePage.php
```diff
@@ -148,4 +148,24 @@ protected function newPage( Title $title ) {
...
+	 * Render the image or video
+	 */
+	public function view() {
+		global $wgRedirectFilePagesForAnons;
+		if ( !$this->getContext()->getUser()->isAnon() || empty( $wgRedirectFilePagesForAnons ) ) {
+			parent::view();
+
+			return;
+		}
+		$data = F::app()->sendRequest(
```
Grunny: I don't think this should be using a controller and `sendRequest` to get this redirect URL. The controller would be useful if this was also used externally or was doing something that required Nirvana controllers. For this purpose, I think this would be cleaner if `fileRedir` was in just a helper class (`FilePageHelper` is fine) and you could just call this with `$url = ( new FilePageHelper() )->fileRedir( $fileTitle );`. Rather than `F::app()->sendRequest` in multiple places that don't need it to be a Nirvana controller.

Grunny (changes requested):

extensions/wikia/FilePage/FilePage.setup.php
```diff
@@ -43,4 +43,11 @@
 $wgHooks['GlobalUsageImagePageWikiLink'][] = 'FilePageHooks::onGlobalUsageImagePageWikiLink';
 $wgHooks['GlobalUsageLinksUpdateComplete'][] = 'FilePageHooks::onGlobalUsageLinksUpdateComplete';
 $wgHooks['BeforePageDisplay'][] = 'FilePageHooks::onBeforePageDisplay';
-$wgHooks['WikiaMobileAssetsPackages'][] = 'FilePageHooks::onWikiaMobileAssetsPackages';
\ No newline at end of file
+$wgHooks['WikiaMobileAssetsPackages'][] = 'FilePageHooks::onWikiaMobileAssetsPackages';
+if ( !empty( $wgRedirectFilePagesForAnons ) ) {
```
abador: @Grunny I think that those hooks should run without the if. The PLATFORM-3936 fixes a bug that is causing some file pages not to refresh

extensions/wikia/FilePage/FilePageHelper.class.php
```diff
@@ -13,4 +13,90 @@ public static function stripCategoriesFromDescription( $content ) {
...
+		//fallback to main page
+		$url = Title::newMainPage()->getFullURL();
+		//wiki needs read privileges
+		if ( !$page->getTitle()->userCan( 'read' ) ) {
+			return $url;
+		}
+		$redirKey = wfMemcKey( 'redir', $page->getTitle()->getPrefixedText() );
+
+		$displayImg = $img = false;
+		Hooks::run( 'ImagePageFindFile', [ $page, &$img, &$displayImg ] );
```
Grunny: Is this hook used anywhere?

extensions/wikia/FilePage/FilePageHelper.class.php
```diff
@@ -13,4 +13,90 @@ public static function stripCategoriesFromDescription( $content ) {
...
+	 * Returns an url when file needs to be redirected.
+	 *
+	 * @requestParam Title
+	 *
+	 * @return string $url - url to redirect to
+	 */
+	public static function fileRedir(ImagePage $page) {
+		global $wgMemc;
+
+		$page = $page->getContext();
```
Grunny: Rather than overwriting `$page` I'd suggest making this `$context`. Though, other than the hook (see my question below), this is only used to get the title, so can this either be `$title = $page->getContext()->getTitle()`, or probably better is to just have the method take the title as the parameter, rather than the Page object?

extensions/wikia/FilePage/FilePageHelper.class.php
```diff
@@ -13,4 +13,90 @@ public static function stripCategoriesFromDescription( $content ) {
...
 	}
+
+	/**
+	 * Returns an url when file needs to be redirected.
+	 *
+	 * @requestParam Title
+	 *
+	 * @return string $url - url to redirect to
+	 */
+	public static function fileRedir(ImagePage $page) {
```
Grunny: The method name might be clearer as `getFilePageRedirect`.

extensions/wikia/MercuryApi/handlers/MercuryApiFilePageHandler.class.php
```diff
@@ -36,11 +36,26 @@ function ( $acc, $child ) {
...
+				$item['canRead'] = false;
+			}
 
 			return $item;
 		}, $fileUsageData['fileList']);
 
+		$url = Title::newMainPage()->getFullURL();
+		$page = new ImagePage( $title );
+		$redir = FilePageHelper::fileRedir( $page );
+		if( !is_null($redir) ) {
```
Grunny: Can this actually be `null` or is it always going to be a string?

Grunny (changes requested):

extensions/wikia/FilePage/FilePageHelper.class.php
abador: It looks that it was used in the original ImagePage:65, but as far as I can tell this hook really isn't connected to any action.

extensions/wikia/FilePage/FilePage.setup.php
Grunny: OK, this was mostly to prevent purging cache keys on wikis where nothing would be cached for file redirects, but it should be relatively harmless to re-enable them. So, I'm OK with removing the feature flag here. :)

includes/WikiPage.php
```diff
@@ -1297,6 +1297,8 @@ public function doEdit( $text, $summary, $flags = 0, $baseRevId = false, $user =
 
 		$flags = $this->checkFlags( $flags );
 
+		Hooks::run( 'ArticleFilesRefresh', [ $this->mTitle ] );
```
Grunny: Can we just use the `ArticleSave` hook instead of adding a new hook? The title can be retrieved from the `ArticleSave` parameters so it should work the same.

Grunny (changes requested):

Grunny (approved):

extensions/wikia/Lightbox/LightboxController.class.php
```diff
@@ -255,9 +255,12 @@ public function getShareCodes() {
 		$fileUrl = '';
 		$thumbUrl = '';
 		$networks = array();
+		$fileTitleObj =  Title::newFromText( $fileTitle, NS_FILE );
```
Grunny: This is causing the unit tests to fail because at this stage `fileTitle` may not be valid (or may be empty) so the `getPrefixedText` is failing on one text. Previously this was within the `if ( !empty( $file ) ) {` check, so it only created a title object if the file was valid.

extensions/wikia/Lightbox/LightboxController.class.php
```diff
@@ -255,9 +255,12 @@ public function getShareCodes() {
...
+
+		Wikia::setSurrogateKeysHeaders( "sharecodes-" . $fileTitleObj->getPrefixedText(),
```
Grunny: Minor: Should use single quotes on the string per conventions.

Also, can this also use `getShareSurrogateKey`?

extensions/wikia/Lightbox/LightboxHelper.class.php
```diff
@@ -152,4 +152,12 @@ public function getTimestamp() {
 		return $timestamp;
 	}
 
+	/**
+ * Get a surrogate key for share
```
Grunny: Minor: Indentation issues.

extensions/wikia/Lightbox/LightboxHelper.class.php
```diff
@@ -152,4 +152,12 @@ public function getTimestamp() {
...
+ * @return string - key
+ */
+	public function getShareSurrogateKey(Title $title) {
+		return "sharecodes-" . $title->getPrefixedText();
```
Grunny: Minor: Single quotes per conventions.

extensions/wikia/Lightbox/LightboxHooks.class.php
```diff
@@ -26,4 +26,18 @@ public static function onGetHTMLAfterBody( Skin $skin, &$html ) {
...
+	public static function onArticlePurge( WikiPage $page ) {
+		if ( !$page->getTitle()->inNamespace( NS_FILE ) ) {
+			return true;
+		}
+		Wikia::purgeSurrogateKey( (new LightboxHelper)->getShareSurrogateKey( $page->getTitle() ) );
+		return true;
+	}
+
+
+
```
Grunny: Minor: Too much space.

Grunny (changes requested):

Grunny (approved):

PR #16652, opened 2019-02-05
ADEN-8278 Collapse outstream when there is hivi iab

fraszczakszymon:


extensions/wikia/AdEngine/js/slot/service/slotRegistry.js
```diff
@@ -40,9 +41,10 @@ define('ext.wikia.adEngine.slot.service.slotRegistry',  [
 		}
 	}
 
-	function setState(slotName, state) {
+	function setState(slotName, state, status) {
 		var slot = get(slotName);
 		slotStates[slotName] = state;
+		slotStatuses[slotName] = status;
```
jacekbj, line 47: Can we have a separate `setStatus` method?

jacekbj: Or rename `setState` to sth not tightly connected with state.

jacekbj: Status can be set only when enabling/disabling slot. I resolve all comments here.

PR #16663, opened 2019-02-07
ADEN-8230 Register skin template in AE3

fraszczakszymon:


extensions/wikia/AdEngine3/src/templates/background-changer.js
```diff
@@ -0,0 +1,57 @@
...
+			document.querySelector('.background-image-gradient').classList.add('no-gradients');
+		}
+	} else {
+		document.body.classList.remove('background-dynamic');
+		document.body.classList.remove('background-not-tiled');
+		document.body.classList.remove('background-fixed');
+	}
+}
+
+class BackgroundChanger {
```
nandy-andy, line 19: Inspired by https://github.com/Wikia/app/blob/61d6d1b394768cf9ea78a22488d40750ce7eec35/skins/oasis/js/BackgroundChanger.js :)

extensions/wikia/AdEngine3/src/templates/skin.js
```diff
@@ -0,0 +1,93 @@
...
+			pixelElement = document.createElement('img');
+			pixelElement.src = pixelUrl;
+			pixelElement.width = 1;
+			pixelElement.height = 1;
+			parentNode.appendChild(pixelElement);
+		}
+	});
+}
+
+export class Skin {
```
nandy-andy, line 20: Inspired by https://github.com/Wikia/app/blob/b34866eef012d99285c732311d60fcf7b541df03/extensions/wikia/AdEngine/js/template/skin.js ;)

nandy-andy (approved):

extensions/wikia/AdEngine3/src/templates/background-changer.js
fraszczakszymon: Yes, I didn't want to hack and use original backgroundChanger attached to window. I believe we can remove BackgroundChanges.js once we finish with AE3. It seems that it's used only by AE2.

extensions/wikia/AdEngine3/src/templates/skin.js
fraszczakszymon: :+1:

PR #16673, opened 2019-02-11
IW-1499 | Additional tracking for search

rybmat:
https://wikia-inc.atlassian.net/browse/IW-1603

@Wikia/iwing


extensions/wikia/Search/js/WikiaSearch.js
```diff
@@ -81,8 +89,103 @@ var WikiaSearch = {
...
+		}
+
+		var payload = {
+			searchPhrase: query,
+			clicked: {
+				type: 'article', // we don't show wikis results right now
+				id: parseInt(clickedElement.getAttribute('data-page-id')),
+				title: clickedElement.text,
+				position: parseInt(clickedElement.getAttribute('data-pos')),
+				thumbnail: !!clickedElement.getAttribute('data-thumbnail'),
```
jsutterfield: Is this meant to be a boolean? If so, maybe `hasThumbnail` would be more appropriate?

extensions/wikia/Search/js/WikiaSearch.js
```diff
@@ -81,8 +89,103 @@ var WikiaSearch = {
...
+
+		if (!query) {
+			return;
+		}
+
+		var payload = {
+			searchPhrase: query,
+			clicked: {
+				type: 'article', // we don't show wikis results right now
+				id: parseInt(clickedElement.getAttribute('data-page-id')),
```
jsutterfield: Do we need to cast these values to ints?

extensions/wikia/Search/js/WikiaSearch.js
```diff
@@ -81,8 +89,103 @@ var WikiaSearch = {
...
+			var originalUrl = $elem.attr('href');
+			var modifiedUrl = new URL(originalUrl);
+
+			modifiedUrl.searchParams.append('searchUID', searchUID);
+			$elem.attr('href', modifiedUrl);
+		});
+
+		return searchUID;
+	},
+	genUID: function() {
```
jsutterfield: We should probably use the shared uuid module you created instead of this, yea?

resources/wikia/modules/uuid.js
```diff
@@ -0,0 +1,8 @@
+define('uuid', function () {
```
jsutterfield, line 1: This comes from [here](https://stackoverflow.com/questions/105034/create-guid-uuid-in-javascript/2117523#2117523), right? Maybe we could include that link in the method docs for more context?

extensions/wikia/Search/js/WikiaSearch.js
```diff
@@ -81,8 +89,103 @@ var WikiaSearch = {
...
+	getUniqueSearchId: function() {
+		if (this.searchUID) {
+			return this.searchUID;
+		}
+
+		var queryParams = new URL(window.location).searchParams;
+		var searchUID = queryParams.get('searchUID') || this.genUID();
+		this.searchUID = searchUID;
+
+		$('a.paginator-prev, a.paginator-next, a.paginator-page').each(function() {
```
jsutterfield: How about splitting this into a new `appendUuidToPaginationLinks` method? (or some equivalent name).

extensions/wikia/Search/js/WikiaSearch.js
```diff
@@ -81,8 +89,103 @@ var WikiaSearch = {
 			searchForm.submit();
 		});
 
+	},
+	trackSearchResultClick: function(clickedElement) {
+		var queryparams = new URL(window.location).searchParams;
```
xkxd: IE 11 will explode here unless we have some polyfill

consider `url` package, it's used eg. in `auth-pages`.

can't you go with `window.location.search`?

extensions/wikia/DesignSystem/scripts/DesignSystemGlobalNavigationSearchSuggestions.js
```diff
@@ -1,90 +1,140 @@
...
+			return {
+				enteredPhrase: query,
+				suggestions: Object.keys(suggestions).map(function (articleId) {
+					return {
+						type: 'article',
+						value: articleId,
+						id: suggestions[articleId]
+					};
+				}),
+				app: 'mobile-wiki',
```
jsutterfield: Is this only used on mobile-wiki?

resources/wikia/modules/uuid.js
```diff
@@ -0,0 +1,8 @@
...
+	return function () {
+		return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, function (c) {
+			var r = Math.random() * 16 | 0, v = c === 'x' ? r : (r & 0x3 | 0x8);
```
xkxd, line 5: that's some hardcore magic 👌

xkxd (changes requested):

extensions/wikia/Search/js/WikiaSearch.js
rybmat: then I'd need to parse manually whole query string. Do we even support IE11?

resources/wikia/modules/uuid.js
rybmat: it is copy-pasted from generating pageview unique id, @bkoval is working on sharing it

extensions/wikia/Search/js/WikiaSearch.js
```diff
@@ -33,7 +35,13 @@ var WikiaSearch = {
 			}
 		});
 
+		$('.result-link').on('click', function(event) {
+			console.log(event.target);
```
jsutterfield: So I see we're doing a console.log here, but not in other places where we're calling into the search tracking module. Eg, there's no console.log for `trackSearchResultsImpression`. Is there a reason to log this and not others? Or a reason to log this at all?

jsutterfield (approved):
LGTM. Just a few questions/comments.

extensions/wikia/Search/js/WikiaSearch.js
rybmat: names of properties are defined by data-eng team :)

extensions/wikia/Search/js/WikiaSearch.js
rybmat: yes, @bkoval is working on sharing it, so we can use it in search suggestions as well

resources/wikia/modules/uuid.js
rybmat: actually we copied it from here: https://github.com/Wikia/app/blob/62beff8152ae38d1377cc1c187d5aa8dcc61946b/extensions/wikia/Track/templates/tracking_cookies.mustache#L5, but including link to StackOverflow is a good point

extensions/wikia/DesignSystem/scripts/DesignSystemGlobalNavigationSearchSuggestions.js
rybmat: @bkoval 😈

extensions/wikia/Search/js/WikiaSearch.js
rybmat: this is just leftover I forgot to remove :)

extensions/wikia/Search/js/WikiaSearch.js
rybmat: @xkxd I checked how many sessions we have from IE users, and all of IE versions are getting us 1,52% of sessions (data for the last month). @szczeles did the same check in DW and looks like IE 11 itself is getting us 0.8%. So we both think there is no big deal with broken tracking for this browser

extensions/wikia/Search/js/WikiaSearch.js
rybmat: yes, this is requirement from data-eng

extensions/wikia/Search/js/WikiaSearch.js
rybmat: done

resources/wikia/modules/uuid.js
rybmat: done

extensions/wikia/Search/js/WikiaSearch.js
jsutterfield: 👍

extensions/wikia/Search/js/WikiaSearch.js
jsutterfield: 👍

jsutterfield:
:shipit:

xkxd (approved):

PR #16682, opened 2019-02-12
PLATFORM-3753 | Connect Image Classifier to Image Review process and service

maaartyyynaa:


extensions/wikia/ImageReview/ImageReviewEventsHooks.class.php
```diff
@@ -213,8 +213,12 @@ public static function fetchReviewHistory( $cityId, Title $title ) {
 
 			return array_map(
 				function ( ImageHistoryEntry $item ) use ( $statusMessages ) {
+					$user = $item->getUser();
+					if ( $user == 'FANDOM' ) {
+						$user = 'automatic image classifier';
```
harnash: I would prefer not use literals directly but use `const` instead. Would be much easier to reference if needed

harnash (changes requested):

harnash (approved):

PR #16684, opened 2019-02-13
ADEN-8254 | Add eXtended size Not Available targeting

bckatarzyna:


extensions/wikia/AdEngine3/src/slots.js
```diff
@@ -212,6 +214,13 @@ export default {
 		});
 	},
 
+	setupSizes() {
```
dmnsobczak: This function name is not really specific but currently I don't have a better idea for it :/

dmnsobczak (approved):

extensions/wikia/AdEngine/js/provider/gpt/helper.js
```diff
@@ -104,6 +104,10 @@ define('ext.wikia.adEngine.provider.gpt.helper', [
 			slotTargetingData.uap_c = uapContext.getCreativeId();
 			slotTargetingData.outstream = slotTargeting.getOutstreamData() || 'none';
 
+			if (['TOP_LEADERBOARD', 'BOTTOM_LEADERBOARD'].includes(slot.name)) {
```
dmnsobczak: Function `includes` is ES6 method so it won't work on IE. Please replace it with:
`['TOP_LEADERBOARD', 'BOTTOM_LEADERBOARD'].indexOf(slot.name') !== -1`

dmnsobczak (approved):

PR #16687, opened 2019-02-13
CAKE-3572: Improve Flow Atomicity, Force Checks Against Master DB, During Wiki Close

amalcolm-wikia:
Make checks, flow, and related error output more atomic.
Force use of master DB, for protected flag check, when setting wiki status to closed. 

@Wikia/cake

### DIFF
--- a/extensions/wikia/WikiFactory/api/MarkWikiAsClosedController.class.php
+++ b/extensions/wikia/WikiFactory/api/MarkWikiAsClosedController.class.php
@@ -123,7 +123,7 @@ class MarkWikiAsClosedController extends WikiaController {
 	}
 
 	static private function isValidFandomCreatorCommunityId( $fandomCreatorCommunityId, $wikiId ) {
-		if (!is_numeric( $fandomCreatorCommunityId )) {
+		if ( !is_numeric( $fandomCreatorCommunityId ) ) {
 			return false;
 		}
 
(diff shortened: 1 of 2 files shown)


extensions/wikia/WikiFactory/api/MarkWikiAsClosedController.class.php
```diff
@@ -30,32 +30,48 @@ public function markWikiAsClosed() {
...
 			return;
 		}
 
-		if ( !empty( $fandomCreatorCommunityId ) && !is_numeric( $fandomCreatorCommunityId ) ) {
-			$this->response->setCode( 400 );
-			$this->info('invalid  fandomCreatorCommunityId parameter in request');
+		if ( !empty( $fandomCreatorCommunityId ) ) {
+			if ( !static::isValidFandomCreatorCommunityId( $fandomCreatorCommunityId, $wikiId )) {
+				$this->response->setCode( 400 );
+				$this->info( 'invalid fandomCreatorCommunityId parameter in request: ' . $fandomCreatorCommunityId );
```
themech: I think it is better to put variables into the log message context like `$this->info( 'message', [ 'fandomCreatorCommunityId' => $fandomCreatorCommunityId] );` 
It makes using Kibana logs easier: one can filter out some types of error messages or search by context field values. Same applies to other log messages in this PR.

amalcolm-wikia: Done.

mszabo-wikia (approved):

PR #16704, opened 2019-02-18
(PLATFORM-3929) Enable migration notice on all remaining wikis

Grunny:
Enable migration notice on all remaining wikis except those that may migrate to
wikia.org which will be set in WikiFactory.

/cc @Wikia/core-platform-team

### DIFF
--- a/includes/wikia/VariablesBase.php
+++ b/includes/wikia/VariablesBase.php
@@ -8890,6 +8890,13 @@ $wgAllowCommunityBuilderCNWPrompt = true;
 $wgFandomComMigrationScheduled = false;
 
 /**
+ * Whether the community is scheduled to be migrated to a fandom.com domain, triggers a banner notification
+ * @see SEO-669
+ * @var bool $wgFandomComMigrationScheduled
+ */
+$wgWikiaOrgMigrationScheduled = false;
+
+/**
  * Custom messages to show on the migration banner (before and after migration).
  * @see PLATFORM-3895
  * @var string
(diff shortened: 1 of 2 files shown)


includes/wikia/VariablesBase.php
```diff
@@ -8889,6 +8889,13 @@
  */
 $wgFandomComMigrationScheduled = false;
 
+/**
+ * Whether the community is scheduled to be migrated to a fandom.com domain, triggers a banner notification
+ * @see SEO-669
+ * @var bool $wgFandomComMigrationScheduled
```
themech: I can see some copy-pasting here :)

themech (approved):

includes/wikia/VariablesBase.php
```diff
@@ -8889,6 +8889,13 @@
  */
 $wgFandomComMigrationScheduled = false;
 
+/**
+ * Whether the community is scheduled to be migrated to a fandom.com domain, triggers a banner notification
```
themech:
```suggestion
 * Whether the community is scheduled to be migrated to a wikia.org domain
```

includes/wikia/VariablesBase.php
themech:
```suggestion
 * @var bool $wgWikiaOrgMigrationScheduled
```

includes/wikia/VariablesBase.php
```diff
@@ -8889,6 +8889,13 @@
...
+ * @see SEO-669
```
themech:
```suggestion
```

includes/wikia/VariablesBase.php
Grunny: :blush:

themech (approved):

PR #16709, opened 2019-02-19
ADEN-8341 Add hivi_leaderboard slot and call top_leaderboard on collapse

fraszczakszymon:


skins/oasis/modules/templates/Ad_Top.php
```diff
@@ -3,10 +3,16 @@
 	<div class="WikiaTopAdsInner">
 
 		<?= $app->renderView('Ad', 'Index', [
```
fraszczakszymon, line 5: We don't need to inject hivi_leaderboard for AE2

includes/wikia/DefaultSettings.php
```diff
@@ -979,6 +979,12 @@
  */
 $wgAdDriverAdEngine3EnabledOnOasisSearchPages = false;
 
+/**
+ * @name $wgAdDriverAdEngine3EnabledOnOasisArticlePages
+ * Enables AdEngine3 extension on article pages
+ */
+$wgAdDriverAdEngine3EnabledOnOasisArticlePages = true;
```
fraszczakszymon, line 986: Don't merge me

PR #16712, opened 2019-02-19
SER | Fix image review on DEV

slayful:
In some cases images on dev (possibly on prod as well) do not return a proper revision in this code:
```
$revisionId = $db->selectField( [ 'revision' ], 'rev_id', [
			'rev_page' => $title->getArticleID(),
			'rev_timestamp' => $timestamp,
		] );
```

This PR adds a fallback to article ID for revision.


mszabo-wikia (approved):

slayful:
The root cause of the issues is most likely due to  timestamp being different if there’s enough latency - and in dev there is probably enough latency.

Closing this PR since this is a workaround for a workaround. @ludwikkazmierczak will create a ticket to address this properly.

PR #16713, opened 2019-02-19
ADEN-6934 Set test src based on configuration

nandy-andy:
https://wikia-inc.atlassian.net/browse/ADEN-6934

Connected with: https://github.com/Wikia/config/pull/2529


fraszczakszymon:
What about AE3?

fraszczakszymon:
There is one usage of legacy ad unit in app:
https://github.com/Wikia/app/blob/dev/extensions/wikia/AdEngine/src/context.js#L2
Doubled src will break it. Can we clean up legacy ad unit?

nandy-andy:
Good catch, thank you! 🙇 I'll try to clean it up

fraszczakszymon (approved):

extensions/wikia/AdEngine3/AdEngine3.setup.php
```diff
@@ -22,3 +21,16 @@
 // Hooks
 $wgHooks['OasisSkinAssetGroupsBlocking'][] = 'AdEngine3::onOasisSkinAssetGroupsBlocking';
 $wgHooks['WikiaSkinTopScripts'][] = 'AdEngine3::onWikiaSkinTopScripts';
+
+// Mirrors logic
+if ( !empty( $_SERVER['HTTP_FASTLY_FF'] ) && !empty($_SERVER[ 'X-Staging' ]) ) {
```
fraszczakszymon: What do you think about moving it to separate file `extensions/wikia/AdEngine3/AdHostMirrors.setup.php` and include new file in `includes/wikia/Extensions.php`. Then we don't need to duplicate this code in AdEngine2/3 extensions

nandy-andy: Perfection! ;-)

fraszczakszymon (approved):

PR #16723, opened 2019-02-21
CORE-94 | Community Central's "Blog:Recent posts" has stopped listing posts since May 2018

macbre:
https://wikia-inc.atlassian.net/browse/CORE-94

Due to the way we were fetching list of blog posts IDs from a specific
category, we were unable to get all of them. This has became visible on the
biggest communities lately.

Instead of using `GROUP_CONCAT` to fetch comma separates list of IDs
(that limits returned value to 64000 characters), simply fetch all matching
blog posts IDs. This can be quite big (over 20k entries) on busy communities.
However, the list is cached in the parser cache.

@Wikia/core-team

### DIFF
--- a/extensions/wikia/Blogs/BlogTemplate.php
+++ b/extensions/wikia/Blogs/BlogTemplate.php
@@ -301,12 +301,12 @@ class BlogTemplateClass {
 	 * @param Parser $parser
 	 * @return bool
 	 */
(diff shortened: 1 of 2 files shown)


ghost (approved):
Mi się podoba, a kotu nie przeszkadza :+1:

PR #16741, opened 2019-02-28
SER-3055 | Publish events for images in MediaWiki

slayful:
The goal is to upload all MW images to Google Cloud Storage.
https://wikia-inc.atlassian.net/browse/SER-3035


maintenance/wikia/migrateImagesToGcs.php
```diff
@@ -0,0 +1,75 @@
...
+			return $this->repo->newFile( $row->page_title );
+		} else {
+			return $this->repo->newFile( $row->page_title, $row->rev_timestamp );
+		}
+	}
+
+	private function publishImageUploadRequest( LocalFile $file, $pageId, $revisionId, $uploaderId ) {
+		global $wgGoogleCloudUploaderPublisher;
+
+		$sha1 = $this->repo->getFullSha1( $file->getPath() );
```
mszabo-wikia: I'm guessing `image.img_sha1` is not suitable here, right?

slayful: Yeah, img_sha1 is a weird hash of a sha1 and we've decided to store full sha1 which we can always hash.

mszabo-wikia: kk

mkalinowski-wikia (approved):

PR #16754, opened 2019-03-01
 ADEN-8464 Configure VMG bidder

fraszczakszymon:


extensions/wikia/AdEngine3/src/ad-context.js
```diff
@@ -360,6 +360,16 @@ export default {
 					}
 				}
 			},
+			vmg: {
+				enabled: false,
+				slots: {
+					TOP_LEADERBOARD: {
+						sizes: [
+							[3, 3]
```
jacekbj, line 368: Just asking,  it's `2x2` on mobile and `3x3` on desktop?

jacekbj (approved):

extensions/wikia/AdEngine3/src/ad-context.js
fraszczakszymon: Yes, like UAP sizes :+1:

PR #16759, opened 2019-03-01
ADEN-8383 Adding FMR slot rotator

dmnsobczak:


extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
```diff
@@ -0,0 +1,257 @@
...
+const refreshInfo = {
+	adVisible: false,
+	heightScrolledThreshold: 10,
+	lastRefreshTime: new Date(),
+	refreshAdPosition: 0,
+	refreshDelay: 10000,
+	startPosition: 0
+};
+
+let btRec = false;
```
jacekbj: We need to talk about hiding module state like it's been done here vs extracting class instance which keeps state itself. The current approach makes writing unit tests virtually impossible.

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
```diff
@@ -0,0 +1,257 @@
...
+let recirculationElement = null;
+
+/***********************
+ * BT rec related logic
+ ***********************/
+
+/**
+ * Duplicates ad slot and injects BT rec placement
+ *
+ * @param {function} onSuccess
```
jacekbj: Please mark parameter as optional `@param {function} [onSuccess]`

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
```diff
@@ -0,0 +1,257 @@
...
+ * Removes BT rec node from DOM
+ *
+ * @returns {void}
+ */
+function removeRecNode() {
+	const recNode = document.querySelector(recSelector) || currentRecNode;
+
+	if (recNode) {
+		recNode.style.display = 'none';
+		recNode.remove();
```
jacekbj: Please set recNode to `false` - it is still present after removal from DOM.

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
```diff
@@ -0,0 +1,257 @@
...
+function enoughTimeSinceLastRefresh() {
+	return ((new Date()) - refreshInfo.lastRefreshTime) > refreshInfo.refreshDelay;
+}
+
+/**
+ * Checks UAP is loaded to disable FRM logic in this case
+ *
+ * @returns {boolean}
+ */
+function fanTakeoverLoaded() {
```
jacekbj: This does not seem to be necessary, you can use `universalAdPackage.isFanTakeoverLoaded()` directly.

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
```diff
@@ -0,0 +1,257 @@
...
+	return universalAdPackage.isFanTakeoverLoaded();
+}
+
+/**
+ * Checks and returns page navigation bar height
+ *
+ * @returns {int}
```
jacekbj: s/int/number

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
```diff
@@ -0,0 +1,257 @@
...
+ */
+function getNavbarHeight() {
```
jacekbj: Please refactor this function to a separate file and replace it's use in other places.

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
```diff
@@ -0,0 +1,257 @@
...
+ *
+ * @param {string} slotName
+ * @param {string} slotStatus
+ * @returns {void}
+ */
+function slotStatusChanged(slotName = fmrPrefix, slotStatus = 'success') {
+	if (!fanTakeoverLoaded() && slotName.substring(0, 16) === fmrPrefix) {
+		if (!btRec) {
+			currentAdSlot = slotService.get(currentSlotName);
+			currentSlotName = fmrPrefix + (currentAdSlot.getConfigProperty('repeat.index') + 1);
```
jacekbj: Keeping slot state in config is a bad code smell for me. Unluckily we cannot remove it, at least yet.

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
```diff
@@ -0,0 +1,257 @@
...
+ * @param {function} onSuccess
+ * @returns {void}
+ */
+function applyRec(onSuccess) {
+	if (!btRec) {
+		return;
+	}
+
+	if (recSelector === null) {
+		recSelector = 'div[id*="' + btLoader.getPlacementId(`${fmrPrefix}1`) + '"]';
```
jacekbj: `1` can be replaces with config property `index` or `currentSlotName` (if it gets a default value of 1).

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
```diff
@@ -0,0 +1,257 @@
...
+	}
+
+	currentRecNode = btLoader.duplicateSlot(`${fmrPrefix}1`);
```
jacekbj: ditto

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
```diff
@@ -0,0 +1,257 @@
...
+/**
+ * Checks if slot can be refreshed or limit is reached
+ *
+ * @returns {boolean}
+ */
+function refreshLimitAvailable() {
+	const limitAvailable = btRec || (currentAdSlot && (currentAdSlot.getConfigProperty('repeat.index') < 20));
+
+	if (!limitAvailable) {
+		scrollListener.removeCallback(rotatorListener);
```
jacekbj: Please move this side effect to `rotateSlots`.

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
```diff
@@ -0,0 +1,257 @@
...
+}
+
+/**
+ * Shows or hides recirculation element
+ *
+ * @param {boolean} visible
+ * @returns {void}
+ */
+function swapRecirculation(visible) {
+	recirculationElement.style.display = visible ? 'block' : 'none';
```
jacekbj: `recirculationElement` may be `null`?

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
```diff
@@ -0,0 +1,257 @@
...
+	refreshInfo.adVisible = !visible;
```
jacekbj: `refreshInfo.adVisible = visible`?

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
```diff
@@ -0,0 +1,257 @@
...
+}
+
+/**
+ * Checks if user is in the right position to rotate FMR
+ *
+ * @returns {boolean}
+ */
+function userInRightPosition() {
+	return utils.isInViewport(recirculationElement) || (btRec && currentRecNode && utils.isInViewport(currentRecNode))
```
jacekbj: Please split this into parts :D

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
```diff
@@ -0,0 +1,257 @@
...
+function userReachedStartPosition() {
+	return refreshInfo.startPosition <= window.scrollY;
+}
+
+/**
+ * Checks if user scrolled some distance before slot reload
+ *
+ * @returns {boolean}
+ */
+function userScrolledEnoughDistance() {
```
jacekbj: I would change the name from `userScrolledEnoughDistance` to `userScrolledEnoughDistanceToReload`

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
```diff
@@ -0,0 +1,257 @@
...
+	return Math.abs(window.scrollY - refreshInfo.refreshAdPosition) > refreshInfo.heightScrolledThreshold;
+}
+
+/**********************************
+ * Watchers and exported functions
```
jacekbj: Only one of the functions below is exported

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
```diff
@@ -0,0 +1,257 @@
...
+}
+
+/**
+ * Callback executed on ad slot or BT rec placement load
+ *
+ * @param {string} slotName
+ * @param {string} slotStatus
+ * @returns {void}
+ */
+function slotStatusChanged(slotName = fmrPrefix, slotStatus = 'success') {
```
jacekbj: Please use `AdSlot.STATUS_SUCCESS` instead of string literal "success".

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
```diff
@@ -0,0 +1,257 @@
...
+	if (!fanTakeoverLoaded() && slotName.substring(0, 16) === fmrPrefix) {
+		if (!btRec) {
+			currentAdSlot = slotService.get(currentSlotName);
+			currentSlotName = fmrPrefix + (currentAdSlot.getConfigProperty('repeat.index') + 1);
+		}
+
+		updateAdRefreshInformation();
+
+		if (slotStatus === 'success') {
```
jacekbj: ditto

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
```diff
@@ -0,0 +1,257 @@
...
+ * Callback executed on ad slot or BT rec placement load
+ *
+ * @param {string} slotName
+ * @param {string} slotStatus
+ * @returns {void}
+ */
+function slotStatusChanged(slotName = fmrPrefix, slotStatus = 'success') {
+	if (!fanTakeoverLoaded() && slotName.substring(0, 16) === fmrPrefix) {
+		if (!btRec) {
+			currentAdSlot = slotService.get(currentSlotName);
```
jacekbj: `currentAdSlot` actually has a different name then `currentSlotName`. I think `currentSlotName` should be `nextSlotName`.

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
```diff
@@ -0,0 +1,257 @@
...
+	return navbar ? navbar.offsetHeight : 0;
+}
+
+/**
+ * Checks if slot can be refreshed or limit is reached
+ *
+ * @returns {boolean}
+ */
+function refreshLimitAvailable() {
+	const limitAvailable = btRec || (currentAdSlot && (currentAdSlot.getConfigProperty('repeat.index') < 20));
```
jacekbj: Why do we skip checking the `currentAdSlot.getConfigProperty('repeat.index') < 20` condition if `btRec` is truthy?

jacekbj (commented):
I have a bunch of remarks but they mainly related to code style and edge cases. Logic-wise it looks very complete 👍

extensions/wikia/AdEngine3/README.md
```diff
@@ -13,15 +13,9 @@ All `*.js` files are build into `dist/` using webpack command.
 - missing DW params
   - scroll_y
   - bidder_won, bidder_won_price
-- floating medrec refresh
-  - with BT integration
 - inline video ads
```
fraszczakszymon: not needed anymore

extensions/wikia/AdEngine3/README.md
```diff
@@ -13,15 +13,9 @@ All `*.js` files are build into `dist/` using webpack command.
 - missing DW params
```
fraszczakszymon: already done

extensions/wikia/AdEngine3/README.md
```diff
@@ -13,15 +13,9 @@ All `*.js` files are build into `dist/` using webpack command.
...
   - scroll_y
   - bidder_won, bidder_won_price
-- floating medrec refresh
-  - with BT integration
 - inline video ads
-- nielsen
 - steam browser
 - templates:
```
fraszczakszymon: already done

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
fraszczakszymon:
```suggestion
		const currentSlotName = `${fmrPrefix}1`;
		recSelector = `div[id*="${btLoader.getPlacementId(currentSlotName)}"]`;
```

fraszczakszymon: It seems we forget to configure eslint in adengine3 extension 🙈

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
```diff
@@ -0,0 +1,257 @@
...
+
+	return navbar ? navbar.offsetHeight : 0;
+}
+
+/**
+ * Checks if slot can be refreshed or limit is reached
+ *
+ * @returns {boolean}
+ */
+function refreshLimitAvailable() {
```
fraszczakszymon: s/refreshLimitAvailable/isRefreshLimitAvailable

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
```diff
@@ -0,0 +1,257 @@
...
+	recirculationElement.style.display = visible ? 'block' : 'none';
+	refreshInfo.adVisible = !visible;
+}
+
+/**
+ * Checks if user is in the right position to rotate FMR
+ *
+ * @returns {boolean}
+ */
+function userInRightPosition() {
```
fraszczakszymon: s/userInRightPosition/isInViewport ?

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
```diff
@@ -0,0 +1,257 @@
...
+	refreshInfo.lastRefreshTime = new Date();
+	refreshInfo.refreshAdPosition = window.scrollY;
+}
+
+/**
+ * Checks if user reached certain scroll position to enable FMR logic
+ *
+ * @returns {boolean}
+ */
+function userReachedStartPosition() {
```
fraszczakszymon: s/userReachedStartPosition/isStartPositionReached

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
fraszczakszymon: What about `isScrolledEnoughToReload`?

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
```diff
@@ -0,0 +1,257 @@
...
+			} else {
+				slotTweaker.hide(currentAdSlot);
+			}
+
+			updateAdRefreshInformation();
+			swapRecirculation(true);
+		} else {
+			scrollListener.removeCallback(rotatorListener);
+
+			context.push('state.adStack', { id: currentSlotName });
```
fraszczakszymon: Shouldn't SlotRepeater take care of that?

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
dmnsobczak: I agree with you, but I'm not sure if this module is not too complex to struggle with unit tests instead of using integration tests.

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
dmnsobczak: Done!

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
dmnsobczak: I don't understand what's the problem - recNode is just a local variable.

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
dmnsobczak: You right, this function was more complex before and I forgot to refactor it after simplification.

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
dmnsobczak: 👍

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
dmnsobczak: I was also bothering me, I don't know why I didn't refactored this. Done!

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
dmnsobczak: It's not a config I think, every ICB slot has different value here and it's used to build slot name.

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
dmnsobczak: Changed!

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
dmnsobczak: Yep!

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
dmnsobczak: Changes applied.

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
dmnsobczak: It's div is always added to page template, so the answer is No.

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
dmnsobczak: No, it's correct. If recirculation is visible, then ad is not.

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
dmnsobczak: It's perfectly readable, what do you mean? :D Splitted!

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
dmnsobczak: I choosed `isScrolledEnoughToRotate`

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
dmnsobczak: Changed `and` to `or` :grammar:

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
dmnsobczak: Done.

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
dmnsobczak: Done.

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
dmnsobczak: The idea was that `currentAdSlot` is slot that is pushed to ads queue and `currentSlotName` is just a name of the next slot, so yes, you right, rename should add some readability.

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
dmnsobczak: We can't read proper index from context, because AdBlock prevents proper AdEngine run, so instead of keeping the index duplicated in fmr-rotator module I decided to enable infinite rotation for AdBlock users since we don't make DFP requests anyway.

extensions/wikia/AdEngine3/README.md
dmnsobczak: Removed.

extensions/wikia/AdEngine3/README.md
dmnsobczak: Removed.

extensions/wikia/AdEngine3/README.md
dmnsobczak: Removed too

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
dmnsobczak: Changed

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
dmnsobczak: Done.

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
dmnsobczak: Done

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
dmnsobczak: Not really. SlotRepeater is not too clever and works only onScroll and we don't inject copies of ICB below but on the same viewport height. We need to move some fmr-rotate logic into it and I don't wanna do this, so I decided to push slots to ads queue manually.

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
```diff
@@ -0,0 +1,237 @@
...
+	}
+}
+
+/**
+ * Enable floating medrec rotating logic
+ *
+ * @param {string} slotName
+ * @returns {void}
+ */
+export function rotateIncontentBoxad(slotName) {
```
fraszczakszymon: It isn't really template (whole file). Could we place it somewhere else?

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
fraszczakszymon: SlotRepeter repeats the slot. 

We can make it more clever, refactor the code. In my opinion fmr-rotator should only pass the conditions (promises?) when slot is going to be repeated, but repeating logic should be included somewhere in the ad-engine. 

I can see that it's going to end with rewriting whole code here, so let's leave it for now. But I believe it should be moved to ad-engine when there is a second use case.

dmnsobczak: 👍

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
dmnsobczak: Moved.

fraszczakszymon (approved):

extensions/wikia/AdEngine3/src/templates/fmr-rotator.js
jacekbj: What I meant is: will recNode be reattached or a new node will be created? If a new one will be created then please set the value to false to make sure that the rest of the code is aware of it's removal.

PR #16768, opened 2019-03-04
SER-3058 | When enabled use a multi-write backend that only uses CEPH

slayful:
https://wikia-inc.atlassian.net/browse/SER-3058


includes/wikia/DefaultSettings.php
```diff
@@ -1833,6 +1833,12 @@
  */
 $wgUseGoogleCloudStorage = false;
 
+/**
+ * https://wikia-inc.atlassian.net/browse/SER-3033
+ * If enabled, a multi write will be used for storing files.
+ */
+$wgUseGcsMigrationStorage = false;
```
pwojcik86: I would use a bucket prefix here, since it will be easier to control than setting flags on multiple wikis. We can set a prefix of out "global wiki" and retrieve it like this:

    \WikiFactory::getVarValueByName("wgUseGcsMigrationBucketPrefix", static::DEFAULT_WIKI_ID );

includes/wikia/Wikia.php
```diff
@@ -1522,13 +1522,16 @@ public static function invalidateUser( User $user, $disabled = false, $keepEmail
...
-		global $wgFSSwiftContainer, $wgFSSwiftServer, $wgUploadPath, $wgUseGoogleCloudStorage;
+		global $wgFSSwiftContainer, $wgFSSwiftServer, $wgUploadPath, $wgUseGoogleCloudStorage,
+			   $wgUseGcsMigrationStorage;
 
 		$path = trim( parse_url( $wgUploadPath, PHP_URL_PATH ), '/' );
 		$wgFSSwiftContainer = substr( $path, 0, -7 );
 
 		if ( $wgUseGoogleCloudStorage ) {
 			$repo['backend'] = 'gcs-backend';
+		} elseif ( $wgUseGcsMigrationStorage ) {
```
pwojcik86: Here I would compare `$wgFSSwiftContainer` to the configured prefix

pwojcik86 (changes requested):

pwojcik86 (approved):

slayful (commented):
LGTM

PR #16774, opened 2019-03-04
IW-1670 | Add limit param to linkSuggest API

bkoval:
@Wikia/iwing @Grunny 

Hi! This change allows us to pass a limit param to search suggestions API. Limit param is restricted to a number between 1 and 100.

**What it means to our code?**

- There's an additional 'limit' param that can be set to a number between 1 and 100.
- Our memcache might cache more than 1 version of searchSuggestions for a given Phrase (up to a 100) for an hour, depending on which limits are going to be used

**Why do we need it?**
In feeds we will be suggesting tags that can be associated with a post by displaying most popular article names. Once the article is picked as a tag it should be filtered from the subsequent suggestion lists, so we want to achieve this by requesting bigger number of suggestions and filtering out the already used on the front-end side.

We could achieve it by passing a query param with already used article ids and modifying the query on the mediawiki side to exclude these, but it might lead to a potentially bigger number of CDN objects being cached for the same responses + it is much less trivial to implement in mediawiki code.

I am open to suggestions though. :)

### DIFF
--- a/extensions/wikia/LinkSuggest/LinkSuggest.class.php
+++ b/extensions/wikia/LinkSuggest/LinkSuggest.class.php
@@ -219,7 +219,7 @@ class LinkSuggest {
 			$sql2Measurement = T::start([ __FUNCTION__, "sql-2" ]);
 			$res = $db->query($sql, __METHOD__);
 
-			self::formatResults($db, $res, $query, $redirects, $results, $exactMatchRow);
+			self::formatResults($db, $res, $query, $redirects, $results, $exactMatchRow, $limit);
 			$sql2Measurement->stop();
 		}
 
--- a/extensions/wikia/LinkSuggest/LinkSuggest.class.php
+++ b/extensions/wikia/LinkSuggest/LinkSuggest.class.php
@@ -13,7 +13,7 @@ use Wikia\Measurements\Time as T;
  * @author Robert Elwell <robert@wikia-inc.com>
  */
 class LinkSuggest {
-
+	const MAX_LINK_SUGGESTIONS_LIMIT = 100;
 	/**
 	 * Get list of suggested images
 	 *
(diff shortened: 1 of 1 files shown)


extensions/wikia/LinkSuggest/LinkSuggest.class.php
```diff
@@ -364,13 +369,13 @@ static private function replaceResultIfRedirected(&$results, &$redirects) {
...
 	 * @param $exactMatchRow
+	 * @param $limit
 	 *
 	 * @author dymsza
 	 */
 
-	static private function formatResults($db, $res, $query, &$redirects, &$results, &$exactMatchRow) {
-		global $wgLinkSuggestLimit;
-		while(($row = $db->fetchObject($res)) && count($results) < $wgLinkSuggestLimit ) {
+	static private function formatResults($db, $res, $query, &$redirects, &$results, &$exactMatchRow, $limit) {
```
jsutterfield, line 377: Why do we need to add `$limit` here? That should have been handled in the query, right?

bkoval: Hmmm, are you suggesting to remove the count($results) < $limit condition whatsoever?

jsutterfield: Initially, yes, b/c I saw that the function signature changed so I had assumed that `count($results) < $limit ` logic was new, but now I see the old implementation is doing the same thing but was using the global. So disregard my comment :)

Unrelated, looks like we probably need to updated [this call](https://github.com/Wikia/app/pull/16774/files/b184b21848d3402726327580163b8e94b42f52a4#diff-2ac62cf91f10b9c15539d927c47d7ec7R231) with the `$limit` parameter as well.

bkoval: Hmmm, the linked line does not contain any calls that need a limit :)

jsutterfield: Sorry, I meant to link this other `self::formatResults` call on L222... https://github.com/Wikia/app/pull/16774/files/b184b21848d3402726327580163b8e94b42f52a4#diff-2ac62cf91f10b9c15539d927c47d7ec7R222

bkoval: Ah, good catch!

jsutterfield (approved):
LGTM!

Also, great job on the PR description :D

mszabo-wikia (approved):

PR #16779, opened 2019-03-05
PLATFORM-3966 legacy domains

harnash:
Properly set alternative domain for non english language wikis created on fandom.com domain.

### DIFF
--- a/maintenance/wikia/HttpsMigration/fixWikiFactoryDomains.php
+++ b/maintenance/wikia/HttpsMigration/fixWikiFactoryDomains.php
@@ -25,16 +25,24 @@ class FixWikiFactoryDomains extends Maintenance {
 			__METHOD__
 		);
 
+		$totalCount = $res->numRows();
+		$this->output( "Processing wikis, $totalCount in total\n");
+
+		$currentIndex = 0;
 		foreach ( $res as $row ) {
+			++$currentIndex;
 			$cityId = $row->city_id;
+			$originalDomain = $row->city_domain;
+
+			$this->output("$currentIndex/$totalCount: Processing wiki (id: $cityId, domain: $originalDomain): ");
 			if ( !WikiFactory::isPublic( $cityId ) ) {
+				$this->output( "wiki not public - skipping\n");
 				continue;
 			}
-			$originalDomain = $row->city_domain;
 
 			$currentUrl = WikiFactory::cityIDtoUrl( $cityId );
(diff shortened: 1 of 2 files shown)


maintenance/wikia/HttpsMigration/fixWikiFactoryDomains.php
```diff
@@ -0,0 +1,96 @@
...
+				"updating WF domain for wiki: {$this->COLOR_GREEN}$originalDomain{$this->COLOR_NONE} -> " .
+				"{$this->COLOR_CYAN}$finalDomain{$this->COLOR_NONE}: "
+			);
+			if ( empty( $finalDomain) ) {
+				$this->output( "no change computed for wiki - {$this->COLOR_CYAN}skipping{$this->COLOR_NONE}\n" );
+				continue;
+			}
+
+			if ( $saveChanges ) {
+				if ( !WikiFactory::removeDomain( $cityId, $originalDomain, [ 'fixing legacy domain after fandom.com migration' ] ) ) {
```
Grunny: The reason parameter should just be a string and not an array, right?

maintenance/wikia/HttpsMigration/fixWikiFactoryDomains.php
```diff
@@ -0,0 +1,96 @@
...
+					$this->output( "failed to remove original domain: {$this->COLOR_RED}$originalDomain{$this->COLOR_NONE}\n" );
+					continue;
+				}
+				if ( !WikiFactory::addDomain( $cityId, $finalDomain, [ 'fixing legacy domain after fandom.com migration' ] ) ) {
```
Grunny: Same here.

Grunny (changes requested):

maintenance/wikia/HttpsMigration/fixWikiFactoryDomains.php
harnash: Right my bad

harnash: Fix in e5c5312763f

maintenance/wikia/HttpsMigration/fixWikiFactoryDomains.php
harnash: Yup: e5c5312763f

Grunny (approved):

PR #16790, opened 2019-03-07
SER | Don't interrupt on failures

slayful:


maintenance/wikia/migrateImagesToGcsForWikis.php
```diff
@@ -103,7 +109,7 @@ private function runMigrateImagesToGcs( $wikiId, $uploadPath ) {
 			$this->output( $output . "\n\n" );
 		} else {
 			$this->error( "Migration failure for {$wikiId}:{$output}\n" );
-			throw new RuntimeException( "Failure to migrate images for wiki {$wikiId}\n" );
+//			throw new RuntimeException( "Failure to migrate images for wiki {$wikiId}\n" );
```
pwojcik86: I think it's fine to just remove this line.

PR #16796, opened 2019-03-07
PLATFORM-3970

abador:
Fix problem with video placeholder


extensions/wikia/FilePage/FilePageHooks.class.php
```diff
@@ -199,6 +199,10 @@ public static function onUndeleteComplete( Title $title ) {
 	 * @return true -- because it's hook
 	 */
 	public static function onArticleSave( WikiPage $page ) {
+		if(get_class($page) == "WikiFilePage" && $page->getFile()->media_type &&
```
harnash:
```suggestion
		if( get_class( $page ) == "WikiFilePage" && $page->getFile()->media_type &&
```

harnash: Usually when checking what type of page we are dealing with we do something like this:
```php
if( $page->getTitle()->getNamespace() == NS_FILE ) {
    $page->getFile();
}
```

extensions/wikia/FilePage/FilePageHooks.class.php
```diff
@@ -199,6 +199,10 @@ public static function onUndeleteComplete( Title $title ) {
...
+		   $page->getFile()->media_type == "VIDEO" ) {
```
harnash: https://github.com/Wikia/app/blob/dee8bc271b635de7a40ac0eb7cc39af2428ffe16/includes/Defines.php#L109

harnash (changes requested):

extensions/wikia/FilePage/FilePageHooks.class.php
harnash: and
```php
$page->getFile()->getMediaType() ==MEDIATYPE_VIDEO
```

extensions/wikia/FilePage/FilePageHooks.class.php
```diff
@@ -199,6 +199,10 @@ public static function onUndeleteComplete( Title $title ) {
 	 * @return true -- because it's hook
 	 */
 	public static function onArticleSave( WikiPage $page ) {
+		if( $page->getTitle()->getNamespace() === NS_FILE &&
+		   $page->getFile()->getMediaType() === MEDIATYPE_VIDEO ) {
```
harnash:
```suggestion
		    $page->getFile()->getMediaType() === MEDIATYPE_VIDEO ) {
```

harnash (approved):

PR #16799, opened 2019-03-08
Platform 3852 - run wiki in wikia.org domain

abador:
Opening wikia.org wiki


extensions/wikia/CreateNewWiki/tasks/SetupWikiCities.php
```diff
@@ -73,7 +73,7 @@ public function addToCityList() {
 	}
 
 	public function addToCityDomains() {
-		global $wgFandomBaseDomain, $wgWikiaBaseDomain;
+		global $wgFandomBaseDomain, $wgWikiaBaseDomain, $wgWikiaOrgBaseDomain;
```
Grunny: Unused?

extensions/wikia/WikiFactory/WikiFactory.php
```diff
@@ -1408,6 +1409,10 @@ static public function getLocalEnvURL( $url, $forcedEnv = null ) {
 			$server = str_replace( ".{$wgFandomBaseDomain}", '', $server );
 			$fandomDomainUsed = true;
 		}
+		if ( endsWith( $server, ".{$wgWikiaOrgBaseDomain}" ) ) {
+			$server = str_replace( ".{$wgWikiaOrgBaseDomain}", '', $server );
+			$wikiaOrgDomainUsed = true;
```
Grunny: `wikiaOrgDomainUsed` is not being used so could be removed.

includes/wikia/GlobalFunctions.php
```diff
@@ -1609,12 +1609,15 @@ function wfGetEffectiveHostname() {
...
+	global $wgWikiaBaseDomain, $wgWikiaOrgBaseDomain, $wgFandomBaseDomain;
 	if ( strpos( $host, ".{$wgFandomBaseDomain}" ) !== false ) {
 		return $wgFandomBaseDomain;
 	}
+	if ( strpos( $host, ".{$wgWikiaBaseDomain}" ) !== false ) {
+		return $wgWikiaBaseDomain;
+	}
 
-	return $wgWikiaBaseDomain;
+	return $wgWikiaOrgBaseDomain;
```
Grunny: Since most wikis are no on fandom.com, I'd make `wgFandomBaseDomain` the default return now, rather than `wikia.org`.

includes/wikia/DefaultSettings.php
```diff
@@ -1773,7 +1773,7 @@
  * A central regex string for use in domain checking, so we can easily
  * update/add/change domains in the future
  */
-$wgWikiaBaseDomainRegex = '((wikia|fandom)\\.com|(wikia|fandom)-dev\\.(com|us|pl))';
+$wgWikiaBaseDomainRegex = '((wikia|fandom)\\.(com|org)|(wikia|fandom)-dev\\.(com|us|pl))';
```
Grunny: I'm not sure if we actually own `fandom.org`, so, given this can be used for validating inputs, I think it would be safer to separate this (we could have `wikia\\.(com|org)|fandom\\.com` for example).

Grunny (changes requested):

includes/wikia/GlobalFunctions.php
abador: @Grunny it seems that this change breaks one of the tests. The default behavior should be that sitemap links for other domains should be generated with the http address extensions/wikia/RobotsTxt/tests/WikiaRobotsTest.php:testSitemap() . Do we want to change it or should it stay the way it is?

Grunny: Leaving it as `wgWikiaBaseDomain` is probably OK, but just changing the test to `https` may make more sense since long-term we won't have anything left on `wikia.com`, my main concern here was `wgWikiaOrgBaseDomain` being the default (and `wikia.org` will also be HTTPS-only anyway). So, I'll leave it up to you. :)

Grunny (approved):

PR #16808, opened 2019-03-11
Revert "CORE-95 | SMWSQLStore3Readers - force an index on smw_object_ids"

macbre:
Reverts Wikia/app#16805, forcing an index was not the best idea.

## An example

```sql
mysql@geo-db-smw-slave.query.consul[smw+genealogy]>explain SELECT /* SMWSQLStore3Readers::getPropertySubjects 5.255.250.112 - 5fc13369-7d42-497a-b920-9719a8628d0a */  smw_title, smw_namespace, smw_iw, smw_sortkey, smw_subobject  FROM `smw_object_ids` FORCE KEY(smw_sortkey) INNER JOIN `smw_di_blob` AS t1 ON t1.s_id=smw_id  WHERE t1.p_id='437' AND smw_iw!=':smw' AND smw_iw!=':smw-delete'  GROUP BY smw_sortkey, smw_id ORDER BY smw_sortkey LIMIT 25  ;
+----+-------------+----------------+------------+------+--------------------+------+---------+-------+----------+----------+----------------------------------------------------+
| id | select_type | table          | partitions | type | possible_keys      | key  | key_len | ref   | rows     | filtered | Extra                                              |
+----+-------------+----------------+------------+------+--------------------+------+---------+-------+----------+----------+----------------------------------------------------+
|  1 | SIMPLE      | t1             | NULL       | ref  | s_id,p_id,s_id_2   | p_id | 4       | const |        2 |   100.00 | Using temporary; Using filesort                    |
|  1 | SIMPLE      | smw_object_ids | NULL       | ALL  | smw_id,smw_sortkey | NULL | NULL    | NULL  | 34551857 |     0.00 | Using where; Using join buffer (Block Nested Loop) |
+----+-------------+----------------+------------+------+--------------------+------+---------+-------+----------+----------+----------------------------------------------------+
2 rows in set, 1 warning (0.00 sec)
```

PR #16788, opened 2019-03-07
CORE-95: SemanticMediaWiki performance fix backport

macbre:
https://wikia-inc.atlassian.net/browse/CORE-95

Cherry-picking from https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/3142 (as found by @mszabo-wikia - kudos!).

The use of `DISTINCT` makes the following query trigger `filesort` and` temporary` and causes serious performance problems on SMW database cluster. According to anemometer this query is the one with the highest median query time.

## Explain for an old query

```sql
SELECT /* SMWSQLStore3Readers::getPropertySubjects 66.249.64.63 - 1ab29fdf-071d-4bd6-9fe4-a4215604a1a8 */  DISTINCT smw_title, smw_namespace, smw_iw, smw_sortkey, smw_subobject  FROM `smw_object_ids` INNER JOIN `smw_di_blob` AS t1 ON t1.s_id=smw_id  WHERE t1.p_id='63' AND smw_iw!=':smw' AND smw_iw!=':smw-delete'  ORDER BY smw_sortkey LIMIT 13320,50

-- EXPLAIN
+----+-------------+----------------+------------+------+-----------------------+------+---------+---------------------------------------------+-------+----------+----------------------------------------------+
| id | select_type | table          | partitions | type | possible_keys         | key  | key_len | ref                                         | rows  | filtered | Extra                                        |
+----+-------------+----------------+------------+------+-----------------------+------+---------+---------------------------------------------+-------+----------+----------------------------------------------+
| 1  | SIMPLE      | smw_object_ids |            | ALL  | PRIMARY,smw_id,smw_iw |      |         |                                             | 75042 | 50.94    | Using where; Using temporary; Using filesort |
+----+-------------+----------------+------------+------+-----------------------+------+---------+---------------------------------------------+-------+----------+----------------------------------------------+
| 1  | SIMPLE      | t1             |            | ref  | s_id,p_id,s_id_2      | s_id | 8       | smw+devroniplag.smw_object_ids.smw_id,const | 1     | 100.00   | Using index; Distinct                        |
+----+-------------+----------------+------------+------+-----------------------+------+---------+---------------------------------------------+-------+----------+----------------------------------------------+
```

### Explain for a new, optimised query

```sql
[...]

PR #16810, opened 2019-03-11
Revert "CORE-95 | SMWSQLStore3Readers - force an index on smw_object_ids [prod]"

macbre:
Reverts Wikia/app#16807,  backporting #16808


mwjames:
@macbre [0] contains some notes about index hinting for this particular select query.

PS: The best option was to use `GROUP BY smw_sort, smw_id` (or in your case `smw_sortkey`) to avoid any reference to `DISTINCT`, `ORDER BY` and `LIMIT` as this will most certainly cause a filesort. This particular query should always (aka implicitly) sort `ASC` which on MySQL is done so by relying on a compound index when using `GROUP BY`.

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/SQLStore/EntityStore/PropertySubjectsLookup.php#L279-L309

macbre:
@mwjames , and that's what we did in #16788 :)

Thanks for the link on index hinting using the number of rows with a given `p_id` value. I'll give a try on our most time consuming queries.

PR #16823, opened 2019-03-12
Purge only fastly cache, PLATFORM-3973

abador:


extensions/wikia/FilePage/FilePageHelper.class.php
```diff
@@ -21,13 +21,16 @@ public static function stripCategoriesFromDescription( $content ) {
 	 *
 	 * @return string $url - url to redirect to
 	 */
-	public static function getFilePageRedirect( Title $title ) {
+	public static function getFilePageRedirect( Title $title , bool $onlyDB=false) {
```
themech: can you please add parameters description for the method doc? especially for `onlyDB`

themech: Is this method supposed to be used externally? I'm wondering about the public access modifier

extensions/wikia/FilePage/FilePageHelper.class.php
```diff
@@ -43,22 +46,22 @@ public static function getFilePageRedirect( Title $title ) {
 		}
 
 		$urlMem = $wgMemc->get( $redirKey );
```
themech: This call is not needed when `onlyDB` is set to true

extensions/wikia/FilePage/FilePageHelper.class.php
```diff
@@ -67,10 +70,66 @@ public static function getFilePageRedirect( Title $title ) {
 				'file' => $title->getText()
 			] );
 		}
-		$wgMemc->set( $redirKey, $url );
+		if( !$onlyDB ) {
```
themech: I'm a little confused here. Even if you don't want to used a cached value, why do we decide not to store it in memcache in case it was not there?

extensions/wikia/FilePage/FilePageHelper.class.php
```diff
@@ -67,10 +70,66 @@ public static function getFilePageRedirect( Title $title ) {
...
+			$wgMemc->set( $redirKey, $url );
+		}
 		return $url;
 	}
 
+	/**
+	 * Returns list of surrogate keys for purging.
+	 *
+	 * @requestParam Title
```
themech: it should be `@param`, `@ requestParam ` is used when a method fetches param values from the current request

extensions/wikia/FilePage/FilePageHelper.class.php
```diff
@@ -67,10 +70,66 @@ public static function getFilePageRedirect( Title $title ) {
...
+	 *
+	 * @return array $keys - surrogate keys to purge
+	 */
+	public static function getSurrogateKeys( Title $title ) {
+		$keys = [];
+		$keys[] = self::getRedirSurrogateKey( $title );
+		if ( $title->inNamespace( NS_FILE ) ) {
+			$url = self::getFilePageRedirect( $title );
+			if( isset($url) ){
```
themech:
```suggestion
			if( isset( $url ) ){
```

extensions/wikia/FilePage/FilePageHelper.class.php
```diff
@@ -67,10 +70,66 @@ public static function getFilePageRedirect( Title $title ) {
...
+				$pageTitle = self::rawURLToTitle( $url );
+				if( $pageTitle ){
```
themech:
```suggestion
				if( $pageTitle ) {
```

extensions/wikia/FilePage/FilePageHelper.class.php
```diff
@@ -67,10 +70,66 @@ public static function getFilePageRedirect( Title $title ) {
...
+				}
+			}
+		}
+		return $keys;
+	}
+
+	/**
+	 * Returns surrogate keys for purging
+	 *
+	 * @requestParam Title
```
themech: should be `@param`

extensions/wikia/FilePage/FilePageHelper.class.php
```diff
@@ -67,10 +70,66 @@ public static function getFilePageRedirect( Title $title ) {
...
+	 *
+	 * @return string $key - surrogate key to purge
+	 */
+	public static function getRedirSurrogateKey( Title $title ) {
```
themech: `protected`?

extensions/wikia/FilePage/FilePageHelper.class.php
```diff
@@ -67,10 +70,66 @@ public static function getFilePageRedirect( Title $title ) {
...
+		return 'redirect-' . $title->getPrefixedText();
+	}
+
+	/**
+	 * Returns Title object based on araw URL
+	 *
+	 * @requestParam string
```
themech: `@param`

extensions/wikia/FilePage/FilePageHelper.class.php
```diff
@@ -67,10 +70,66 @@ public static function getFilePageRedirect( Title $title ) {
...
+	 *
+	 * @return Title $title or null
+	 */
+	private static function rawURLToTitle( string $url ) {
```
themech: Oh dear. `https://muppet.fandom.com/wiki/Cerrado/Abierto` will not work correctly.

extensions/wikia/FilePage/FilePageHelper.class.php
```diff
@@ -67,10 +70,66 @@ public static function getFilePageRedirect( Title $title ) {
...
+
+	/**
+	 * Returns surrogate keys for purging
+	 *
+	 * @requestParam Title
+	 *
+	 * @return string $key - surrogate key to purge
+	 */
+	public static function getRedirSurrogateKey( Title $title ) {
+		return 'redirect-' . $title->getPrefixedText();
```
themech: This key does not contain wiki ID and probably it should. Try using the `Wikia::surrogateKey` method

extensions/wikia/FilePage/FilePageHelper.class.php
```diff
@@ -67,10 +70,66 @@ public static function getFilePageRedirect( Title $title ) {
...
+	 *
+	 * @return array $keys - surrogate keys to purge
+	 */
+	public static function getSurrogateKeys( Title $title ) {
+		$keys = [];
+		$keys[] = self::getRedirSurrogateKey( $title );
+		if ( $title->inNamespace( NS_FILE ) ) {
+			$url = self::getFilePageRedirect( $title );
+			if( isset($url) ){
+				$pageTitle = self::rawURLToTitle( $url );
```
themech: as this converts any url to a local title object, I'd like to make sure - the url is always pointing to the current wiki, right?

extensions/wikia/FilePage/FilePageHelper.class.php
```diff
@@ -67,10 +70,66 @@ public static function getFilePageRedirect( Title $title ) {
...
+				if( $pageTitle ){
+					$keys = array_merge( $keys, self::getSurrogateKeys( $pageTitle ) );
+				}
+			}
+			$url2 = self::getFilePageRedirect( $title, true );
```
themech: `url2` is not the best name, maybe something like `noncacheUrl`?

extensions/wikia/FilePage/FilePageHelper.class.php
```diff
@@ -67,10 +70,66 @@ public static function getFilePageRedirect( Title $title ) {
...
+			if( isset($url2) && $url != $url2 ){
```
themech:
```suggestion
			if ( isset( $url2 ) && $url != $url2 ){
```

extensions/wikia/FilePage/FilePageHooks.class.php
```diff
@@ -181,10 +181,14 @@ static public function onGlobalUsageLinksUpdateComplete( &$images ) {
 	 * @param User $user -- current user
 	 * @param string $reason -- undeleting reason
 	 *
-	 * @return true -- because it's hook
+	 * @return true -- because it's a hook
 	 */
 	public static function onUndeleteComplete( Title $title ) {
-		self::purgeTitle( $title );
+		global $wgRedirectFilePagesForAnons;
```
themech: you seem to be checking this in all hook listeners. I'm wondering - if we want to limit this to selected wikis, can we include `extensions/wikia/FilePage/FilePage.setup.php` only if `$wgRedirectFilePagesForAnons` is true?

extensions/wikia/FilePage/FilePageHooks.class.php
```diff
@@ -181,10 +181,14 @@ static public function onGlobalUsageLinksUpdateComplete( &$images ) {
...
+		if( !$wgRedirectFilePagesForAnons ){
+			return true;
+		}
+		self::clearLinkedFilesCache( $title , true);
```
themech: another space missing. can you please reformat the PR?

extensions/wikia/FilePage/FilePageHooks.class.php
```diff
@@ -260,62 +257,32 @@ public static function onGetFileLinks( $id, &$links ) {
...
+				$title = Title::makeTitleSafe( NS_FILE, $row->il_to );
+				self::purgeMemcache( $title );
+			}
 		}
-		foreach ( $results as $row ) {
-			$title = Title::makeTitleSafe( NS_FILE, $row->il_to );
-			self::purgeRedir( $title );
+		if( !$memcacheOnly ){
+			foreach( $keys  as $key) {
+				Wikia::purgeSurrogateKey( $key );
```
themech, line 303: Can you please log the keys we're purging to kibana so we get a more info about the scope of purges when testing/releasing.

extensions/wikia/FilePage/WikiaFilePage.php
```diff
@@ -153,6 +153,8 @@ protected function newPage( Title $title ) {
 	 */
 	public function view() {
 		global $wgRedirectFilePagesForAnons;
+		Wikia::setSurrogateKeysHeaders( FilePageHelper::getSurrogateKeys( $this->getTitle()	) ,
```
themech: if we decide to include file pages redirects only on selected wikis, this should be replaces with a hook call.

extensions/wikia/MercuryApi/MercuryApiController.class.php
```diff
@@ -264,8 +264,10 @@ public function getPage() {
 
 		try {
 			$title = $this->getTitleFromRequest();
-			Wikia::setSurrogateKeysHeaders( ( new LightboxHelper )->getShareSurrogateKey( $title ),
+			Wikia::setSurrogateKeysHeaders( FilePageHelper::getSurrogateKeys( $title
+			) ,
```
themech: formatting is off

themech (changes requested):

extensions/wikia/FilePage/FilePageHelper.class.php
themech: Just a thought after analyzing the flow. maybe it would be easier to just store page title prefixed text in the memcache instead of the full url? This way we would avoid doing this magic here.

extensions/wikia/FilePage/FilePageHelper.class.php
abador: We don't want to purge memcache prematurely. It was a source of problems with fastly/memcache and even if someone uses this method elsewhere it shouldn't break the purging logic.

extensions/wikia/FilePage/FilePageHelper.class.php
abador: Yep it's used in 3 places that's why it's a helper

extensions/wikia/FilePage/FilePageHooks.class.php
abador: I tried it but unfortunately at the point of including setup only defaults are loaded and wgRedirectFilePagesForAnons is false :/

extensions/wikia/FilePage/FilePageHelper.class.php
```diff
@@ -17,58 +17,133 @@ public static function stripCategoriesFromDescription( $content ) {
...
+			return "";
+		}
+		$redirTitle = Title::newFromText( $prefixedText );
+		$url = $redirTitle->getFullURL();
+		if( Title::newMainPage()->getPrefixedText() == $prefixedText ){
+			$url = wfAppendQuery( $url, [
+				'file' => $title->getText(),
+			] );
+		}
+		\Wikia\Logger\WikiaLogger::instance()->info( __FUNCTION__, [
```
harnash: How often this will be called? Is there a risk of flooding the logs with this?

harnash (commented):
One comment. Rest looks good

harnash (approved):

PR #16841, opened 2019-03-14
Don't hide fandom.com wikis from global sitemap

Grunny:


Grunny:
Reviewed in https://github.com/Wikia/app/pull/16840

PR #16847, opened 2019-03-19
IW-1659 | [feeds] allow for using title instead of article id for getting article details

rybmat:
@Wikia/iwing

### DIFF
--- a/extensions/wikia/FeedsAndPosts/FeedsAndPostsController.class.php
+++ b/extensions/wikia/FeedsAndPosts/FeedsAndPostsController.class.php
@@ -51,7 +51,7 @@ class FeedsAndPostsController extends WikiaApiController {
 			$title = Title::newFromText( $articleTitle );
 		}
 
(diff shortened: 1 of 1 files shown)


extensions/wikia/FeedsAndPosts/FeedsAndPostsController.class.php
```diff
@@ -38,10 +38,20 @@ public function getWikiVariables() {
 	}
 
 	public function getArticleData() {
-		$articleId = intval( $this->getRequiredParam( 'id' ) );
+		$articleId = intval( $this->getVal( 'id', null ) );
```
mszabo-wikia: If `id`  won't end up being used I think we should remove it—afaik this endpoint will only be used by the tags page.

rybmat: yeah, it makes sense, thanks!

mszabo-wikia (approved):

PR #16849, opened 2019-03-20
ADEN-8501 | Add likho key-val to ae3 targeting

bckatarzyna:


extensions/wikia/AdEngine3/src/targeting.js
```diff
@@ -165,6 +165,21 @@ function getZone(adsContext) {
 	};
 }
 
+function getLikhoParams() {
+  const likhoStorage = JSON.parse(localStorage.getItem('likho')) || [];
+  const params = [];
+
+  likhoStorage.forEach((item, index, object) => {
```
Bielik20: `object` -> `array`

It may be confusing to call it object.

extensions/wikia/AdEngine3/src/targeting.js
```diff
@@ -165,6 +165,21 @@ function getZone(adsContext) {
 	};
 }
 
+function getLikhoParams() {
+  const likhoStorage = JSON.parse(localStorage.getItem('likho')) || [];
```
Bielik20: 💎 for default value `[]`

extensions/wikia/AdEngine3/src/targeting.js
```diff
@@ -165,6 +165,21 @@ function getZone(adsContext) {
...
+  const params = [];
+
+  likhoStorage.forEach((item, index, object) => {
+    if (item.expirationTime > Date.now()) {
+      params.push(item.likhoType);
+    } else {
+      object.splice(index, 1);
```
Bielik20: I am not sure about that. It modifies array on which it performs actions. I am not saying no explicitly but I am not sure what will happen .

Bielik20 (approved):

dmnsobczak (approved):

PR #16854, opened 2019-03-21
ADEN-8511 Change Quantcast label payload

dmnsobczak:


extensions/wikia/AnalyticsEngine/AnalyticsProviderQuantServe.php
```diff
@@ -1,75 +1,84 @@
...
+		$keyValues = F::app()->wg->DartCustomKeyValues;
+		$quantcastLabels = array();
+		$labels = array(
+			'gnre' => 'Genre',
+			'media' => 'Media',
+			'theme' => 'Theme',
+			'tv' => 'TV'
+		);
+
+		if (gettype($keyValues) === 'string') {
```
fraszczakszymon: What about `is_string($keyValues)`?

extensions/wikia/AnalyticsEngine/AnalyticsProviderQuantServe.php
```diff
@@ -1,75 +1,84 @@
...
+			foreach (explode(';', $keyValues) as $keyValue) {
+				$keyValue = explode('=', $keyValue);
```
fraszczakszymon:
```suggestion
				list($key, $value) = explode('=', $keyValue);
```
Then `$key`/`$value` will be `string|null`

fraszczakszymon (approved):

extensions/wikia/AnalyticsEngine/AnalyticsProviderQuantServe.php
dmnsobczak: Done.

extensions/wikia/AnalyticsEngine/AnalyticsProviderQuantServe.php
dmnsobczak: Oh yeah, I forgot about that! Changed!

PR #16877, opened 2019-03-28
Vary on X-Mw-Wiki-Id header for LinkSuggest

jsutterfield:
The discussion service queries article titles using the LinkSuggest MW API. It makes all of those requests to community.fandom.com and differentiates the community to query using the `X-Mw-Wiki-Id` header. The problem with that approach is the LinkSuggest API doesn't include `X-Mw-Wiki-Id` in its Vary headers, so we were getting cached responses from different communities.

The fix is to update the response so that Vary header includes `X-Mw-Wiki-Id `.

### DIFF
--- a/extensions/wikia/LinkSuggest/LinkSuggest.php
+++ b/extensions/wikia/LinkSuggest/LinkSuggest.php
@@ -66,8 +66,9 @@ function getLinkSuggest() {
 
 	$out = LinkSuggest::getLinkSuggest($wgRequest);
 
-	$ar = new AjaxResponse($out);
+	$ar = new AjaxResponse( $out );
 	$ar->setCacheDuration(60 * 60);
+	$ar->setVary( implode( ",",[ "Accept-Encoding", WebRequest::MW_WIKI_ID_HEADER ] ) );
 
 	if ($wgRequest->getText('format') == 'json') {
 		$ar->setContentType('application/json; charset=utf-8');
--- a/extensions/wikia/LinkSuggest/LinkSuggest.php
+++ b/extensions/wikia/LinkSuggest/LinkSuggest.php
@@ -68,7 +68,7 @@ function getLinkSuggest() {
 
 	$ar = new AjaxResponse( $out );
 	$ar->setCacheDuration(60 * 60);
-	$ar->setVary( implode( ",",[ "Accept-Encoding", WebRequest::MW_WIKI_ID_HEADER ] ) );
+	$ar->setVary( implode( ",", [ "Accept-Encoding", WebRequest::MW_WIKI_ID_HEADER ] ) );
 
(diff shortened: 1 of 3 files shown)


extensions/wikia/LinkSuggest/LinkSuggest.php
```diff
@@ -66,8 +66,9 @@ function getLinkSuggest() {
 
 	$out = LinkSuggest::getLinkSuggest($wgRequest);
 
-	$ar = new AjaxResponse($out);
+	$ar = new AjaxResponse( $out );
 	$ar->setCacheDuration(60 * 60);
+	$ar->setVary( implode( ",",[ "Accept-Encoding", WebRequest::MW_WIKI_ID_HEADER ] ) );
```
jsutterfield: We're already Varying on `X-Mw-Wiki-Id` [here](https://github.com/Wikia/app/pull/16877/files#diff-5409d922c826c27900d04133bd41e52bR92), but AjaxResponse's use their own Vary logic so we need to add it here as well.

mszabo-wikia (approved):

extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
```diff
@@ -89,7 +89,7 @@ public function  __construct( array $server, array $environment, array $wikiFact
 			$this->mWikiIdForced = true;
 
 			// differ CDN caching on X-Mw-Wiki-Id request header value
-			RequestContext::getMain()->getOutput()->addVaryHeader( 'X-Mw-Wiki-Id' );
+			RequestContext::getMain()->getOutput()->addVaryHeader( WebRequest::MW_WIKI_ID_HEADER );
```
rybmat, line 92: Now I started to wander why we don't have issues with other MW api endpoints we call from pandora services with this header...

rybmat (approved):

extensions/wikia/WikiFactory/Loader/WikiFactoryLoader.php
jsutterfield: So this issues only arises for requests which use `action=ajax`, and it looks like our LinkSuggest request is the only one to use it so far...
```bash
james-s:pandora (master) jamessutterfield$ pwd
/Users/jamessutterfield/wikia/pandora
james-s:pandora (master) jamessutterfield$ grep -r 'action=ajax' *
Binary file service/discussion/build/classes/java/main/com/wikia/discussionservice/gateway/mediawiki/MediaWikiApi.class matches
service/discussion/src/main/java/com/wikia/discussionservice/gateway/mediawiki/MediaWikiApi.java:  @GET("/index.php?action=ajax&rs=getLinkSuggest&format=json&nsfilter=0&limit=20")
```

PR #16901, opened 2019-04-02
[feeds] filter out null titles

rybmat:
@Wikia/iwing

### DIFF
--- a/extensions/wikia/FeedsAndPosts/RecentChanges.class.php
+++ b/extensions/wikia/FeedsAndPosts/RecentChanges.class.php
@@ -78,6 +78,10 @@ class RecentChanges {
 				return $article['pageid'];
 			}, $resultArticles ), DB_SLAVE );
 
+		$resultTitles = array_filter($resultTitles, function ( $title ) {
+			return $title != null;
+		});
+
 		return array_map( function ( $article ) use ( $resultTitles ) {
 			$title = $resultTitles[$article['pageid']];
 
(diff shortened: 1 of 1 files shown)


jsutterfield (approved):

extensions/wikia/FeedsAndPosts/RecentChanges.class.php
```diff
@@ -78,6 +78,10 @@ private function filterOutSmallChangesAndCleanUp( $articles ) {
 				return $article['pageid'];
 			}, $resultArticles ), DB_SLAVE );
 
+		$resultTitles = array_filter($resultTitles, function ( $title ) {
+			return $title != null;
```
mszabo-wikia: What about doing something like
```
$pageIds = [];
$result = [];

for ( $resultArticles as $article ) {
    $pageIds[] = $article['pageid'];
}

$titles = Title::newFromIds( $pageIds );

foreach ( $titles as $title ) {
    $result[] = [
         'title' => $title->getText(),
	'url' => $title->getLocalURL(),
    ];
}
```
`Title::newFromIds`  will never return null titles.

mszabo-wikia (changes requested):

extensions/wikia/FeedsAndPosts/RecentChanges.class.php
latata: @mszabo-wikia Will `$titles` be sorted in right order then?

mszabo-wikia: True, we're operating on sorted data. In that case we could just convert the `array_map` call to a `foreach` loop & do an `isset` check there - then we won't iterate twice.

mszabo-wikia (approved):

PR #16918, opened 2019-04-04
(PLATFORM-4008) Update links to explore pages

Grunny:
Update links to explore pages to fandom.com versions.

Also remove short URL setting on non-English home pages that are now on fandom.com.

/cc @Wikia/core-platform-team


themech (approved):

includes/wikia/DefaultSettings.php
```diff
@@ -1817,13 +1817,6 @@
  */
 $wgShortArticlePathWikis = [
```
themech: can we get rid of this variable and in `WikiFactory::cityUrlToArticlePath` simply check for the `CORPORATE_WIKI_ID`?

themech (approved):

PR #16919, opened 2019-04-04
ADEN-8547 Remove TOP_RIGHT_BOXAD pos.

jacekbj:


extensions/wikia/AdEngine3/src/slots.js
```diff
@@ -116,7 +116,7 @@ export default {
 				defaultSizes: [[300, 250], [300, 600], [300, 1050]],
 				targeting: {
 					loc: 'top',
-					pos: ['TOP_BOXAD', 'TOP_RIGHT_BOXAD'],
+					pos: ['top_boxad'],
```
fraszczakszymon: In AE3 `pos` is not needed if it's one-element array and it matches slot id

fraszczakszymon (approved):

PR #16926, opened 2019-04-05
SER-3129 Add HandleDeletedImages maintenance script

mkalinowski-wikia:
The goal is to remove files that might have been recovered (after deletion) in the process of migrating files to GCS.

https://wikia-inc.atlassian.net/browse/SER-3129

Consumer implemented in https://github.com/Wikia/pandora/pull/6258
The consumer will backup the images and remove them.


pwojcik86 (changes requested):
Please add a description to this PR. Even though this code may be removed, it's good to have some context here.

pwojcik86 (approved):

slayful (approved):

PR #16934, opened 2019-04-09
IW-1726 | add support for DE recommendation service

xkxd:
## Links
* http://wikia-inc.atlassian.net/browse/IW-1726

## Reviewers
@Wikia/iwing


extensions/wikia/Recirculation/js/helpers/recommendedContent.js
```diff
@@ -0,0 +1,97 @@
...
+            url: w.wgServicesExternalDomain + 'recommendations/recommendations',
+            data: {
+                wikiId: w.wgCityId,
+                articleId: w.wgArticleId,
+                beacon: w.beacon_id,
+            }
+        }).done(function (result) {
+            deferred.resolve(mapExperimentalDataResponse(result));
+        }).fail(function (err) {
+            log('Failed to fetch experimental recommended data, using getPopularPages as backup' + err, log.levels.error);
```
rybmat, line 74: does this log method send data to event-logger? I think there was requirement about it

extensions/wikia/Recirculation/js/recirculation.js
```diff
@@ -45,7 +42,9 @@ require([
 			data: {
 				limit: numberOfFandomPostFooterSlots,
 			}
-		}).then(mapAjaxCall);
+		}).then(function (data) {
```
latata, line 45: I think `then(...)` is not needed.

extensions/wikia/Recirculation/js/recirculation.js
```diff
@@ -22,6 +23,7 @@ require([
              mixedFooter,
              discussions,
              sponsoredContentHelper,
+			 recommendedContent,
```
latata: is here something wrong with indentation?

extensions/wikia/Recirculation/js/helpers/recommendedContent.js
```diff
@@ -0,0 +1,97 @@
...
+    var numberOfArticleFooterSlots = $('#mixed-content-footer').data('number-of-wiki-articles');
+    var REQUEST_ID;
+
+    function getFilteredItems(response) {
+        var blacklistedItems = blacklist.get();
+
+        return response.article_recommendation
+            .concat(response.wiki_recommendation)
+            .filter(function (el) {
+                return blacklistedItems.indexOf(el.item_id) === -1;
```
latata, line 21: is `item_id` unique for article_recommendation & wiki_recommendation? I mean is there a possibility that one wiki_recommendation item has the same id as article_recommendation item?

extensions/wikia/Recirculation/js/helpers/recommendedContent.js
```diff
@@ -0,0 +1,97 @@
...
+                    title: el.article_title || el.wiki_title,
+                };
+            });
+    }
+
+    function getExperimentalRecommendedData() {
+        var deferred = $.Deferred();
+        var itemId = w.wgCityId + '_' + w.wgArticleId;
+
+        blacklist.update(itemId);
```
latata: Not sure if we should care about such case but what if the user visits the article but won't scroll down to see Mixed Content Footer and this function won't be called. Should we add such article to the blacklist? :)

extensions/wikia/Recirculation/js/views/mixedFooter.js
```diff
@@ -99,17 +100,28 @@ define('ext.wikia.recirculation.views.mixedFooter', [
...
 				labels = $this.data('tracking').split(','),
-				href = $this.attr('href');
+				href = $this.attr('href'),
+				additionalParams;
+
+			if (shouldSendAdditionalData && labels.indexOf('wiki-article') !== -1) {
+				additionalParams = {
+					recommendation_request_id: recommendedContent.getRequestId(),
+					item_id: $this.attr('data-item-id'),
+					item_type: 'wiki_article'
```
latata, line 117: will it be always the wiki article?

extensions/wikia/Recirculation/js/helpers/recommendedContent.js
xkxd: item_id is in fact `wikiId_articleId`

extensions/wikia/Recirculation/js/recirculation.js
xkxd: if not filtered `when` receives `[data, status, jqXHR]` as an argument

extensions/wikia/Recirculation/js/helpers/recommendedContent.js
xkxd: based on the code `prepareFooter` is called every time, we just fetch additional data (from discussions) when scrolling to the bottom.

extensions/wikia/Recirculation/js/views/mixedFooter.js
xkxd: > item_type (possible values are: wiki_article, wiki, ns_article etc. for now only wiki_article is used).

I decided to hardcode it for now, since the approach may change later

extensions/wikia/Recirculation/js/helpers/recommendedContent.js
xkxd: I've added support for event logger here https://github.com/Wikia/app/pull/16934/files#diff-3d840aaade76d8d66ebe3fae6b6d0f66

and here https://github.com/Wikia/app/pull/16934/files#diff-3bdaa9c4c0484bb666c559d8cd1212e7R33
as they want to get that event in case of not enough items

latata (approved):

extensions/wikia/Recirculation/js/helpers/blacklist.js
```diff
@@ -0,0 +1,59 @@
+define('ext.wikia.recirculation.helpers.blacklist', [
+    'wikia.cache',
+    'wikia.window',
+    'wikia.cookies'
+], function (cache, window, cookies) {
+    'use strict';
+
+    var CACHE_KEY = 'recommendations_blacklist';
```
bkoval: Is there no way to share logic for this between mobile-wiki and app?

extensions/wikia/Recirculation/js/helpers/recommendedContent.js
```diff
@@ -0,0 +1,107 @@
...
+    'wikia.window',
+    'wikia.nirvana',
+    'ext.wikia.recirculation.helpers.blacklist',
+    'wikia.log',
+    'wikia.eventLogger'
+], function ($, w, nirvana, blacklist, log, eventLogger) {
+    'use strict';
+
+    var numberOfArticleFooterSlots = $('#mixed-content-footer').data('number-of-wiki-articles');
+    var REQUEST_ID;
```
bkoval: I thought capses are reserved for consts. How about using a regular camelcase for a variable?

extensions/wikia/Recirculation/js/helpers/recommendedContent.js
```diff
@@ -0,0 +1,107 @@
...
+
+    function getFilteredItems(response) {
```
bkoval: getItemsFilteredByBlacklist?

extensions/wikia/Recirculation/js/helpers/recommendedContent.js
```diff
@@ -0,0 +1,107 @@
...
+            .concat(response.wiki_recommendation);
+    }
+
+    function mapExperimentalDataResponse(response) {
+        var filteredItems = getFilteredItems(response);
+
+        REQUEST_ID = response.recommendation_request_id;
+
+        if (filteredItems < numberOfArticleFooterSlots) {
+            blacklist.remove(5);
```
bkoval: Why 5?

extensions/wikia/Recirculation/js/helpers/recommendedContent.js
```diff
@@ -0,0 +1,107 @@
...
+
+            eventLogger.logError(
```
bkoval, line 34: Is this really an error? Or just a debug information? Maybe we should use another resource for that?

bkoval (changes requested):

extensions/wikia/Recirculation/js/helpers/recommendedContent.js
xkxd: based on the conversation with DE this _should_ not happen (as we'll use "wiki_recommendation" to fill the remaining slots") so when this will actually happen we should treat it as an error

extensions/wikia/Recirculation/js/helpers/blacklist.js
xkxd: unfortunately not.

- different approaches to caching (localStorageConnector / wikia.cache)
- require/ES modules

and extracting such small module to separate repo did not make sense for me

extensions/wikia/Recirculation/js/helpers/recommendedContent.js
xkxd: I've assigned it to a variable, 5 seemed reasonable for me as we'd have slightly more recommendations to choose from then

and DE approved such value.

bkoval (approved):

PR #16936, opened 2019-04-09
PLATFORM-4013| Send MW logs on dev-box to ES

harnash:
On new dev-boxes we need to send logs from docker instance
to ES. Easiest and most reliable solution was to use Fluentd
as we do for our k8s cluster.

### DIFF
--- a/docker/devbox/Dockerfile-fluentd
+++ b/docker/devbox/Dockerfile-fluentd
@@ -8,7 +8,7 @@ USER root
 RUN apk add --no-cache --update --virtual .build-deps \
         build-base ruby-dev \
  && gem install \
-        fluent-plugin-elasticsearch \
+        fluent-plugin-elasticsearch fluent-plugin-throttle \
  && gem sources --clear-all \
  && apk del .build-deps \
  && rm -rf /home/fluent/.gem/ruby/2.5.0/cache/*.gem
--- a/docker/devbox/Dockerfile-fluentd
+++ b/docker/devbox/Dockerfile-fluentd
@@ -6,10 +6,10 @@ USER root
 # below RUN includes plugin as examples elasticsearch is not required
 # you may customize including plugins as you wish
 RUN apk add --no-cache --update --virtual .build-deps \
-        sudo build-base ruby-dev \
- && sudo gem install \
+        build-base ruby-dev \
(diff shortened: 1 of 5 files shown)


docker/devbox/Dockerfile-fluentd
```diff
@@ -0,0 +1,16 @@
+FROM fluent/fluentd:v1.4-onbuild
+
+# Use root account to use apk
+USER root
+
+# below RUN includes plugin as examples elasticsearch is not required
+# you may customize including plugins as you wish
+RUN apk add --no-cache --update --virtual .build-deps \
+        sudo build-base ruby-dev \
```
themech: do we need to use `sudo` if we're running as root?

docker/devbox/fluent.conf
```diff
@@ -0,0 +1,84 @@
...
+  	@type tcp
+  	tag mediawiki
+  	port 9999
+  	<parse>
+		@type json
+		time_format %Y-%m-%dT%H:%M:%S.%N%z
+		time_key @timestamp
+	</parse>
+</source>
+
```
themech, line 11: with syslog-ng, we are throttling the messages: https://github.com/Wikia/chef-repo/blob/37e07590b35b632eb53005ffbb9bc8bb2bb5fb70/cookbooks/syslog-ng-18/attributes/default.rb#L12. What do you think about using something like https://github.com/rubrikinc/fluent-plugin-throttle to limit the messages on dev (in case of some infinite loops or similar)?

docker/devbox/Dockerfile-fluentd
harnash: We are installing `sudo` tool not running as one :-)
`/bin/sh: sudo: not found`

harnash: And this is enforced by the base image. I did not diff too deep into the base image recipe

harnash: PoC in: fa3e51fb62c

docker/devbox/fluent.conf
harnash: Done @themech

themech (approved):

themech (approved):

PR #16946, opened 2019-04-11
ADEN-8651 Mark slot which should be always bidded

dmnsobczak:


jacekbj (approved):

extensions/wikia/AdEngine3/src/slots.js
```diff
@@ -72,6 +72,7 @@ export default {
 			},
 			top_leaderboard: {
 				aboveTheFold: true,
+				alwaysBid: true,
```
fraszczakszymon: I have mixed feelings about this. It creates another flag/"if" path. What about enabling tlb at the beginning:
https://github.com/Wikia/app/blob/dev/extensions/wikia/AdEngine3/src/slots.js#L275
and then disable it on hivi_leaderboard `AdSlot.STATUS_SUCCESS`:
https://github.com/Wikia/app/blob/dev/extensions/wikia/AdEngine3/src/slots.js#L332
```
			slotService.on('hivi_leaderboard', AdSlot.STATUS_SUCCESS, () => {
				slotService.setState('top_leaderboard', false);
				context.set('slots.top_leaderboard.firstCall', false);
			});
```

jacekbj: That's a very good idea. We just need to make sure that second call isn't started before this piece is executed.

dmnsobczak: That's what I'm gonna do, thank you guys!

fraszczakszymon (approved):

fraszczakszymon (approved):

PR #16956, opened 2019-04-16
SER-3089 Change RTBF subroutines order

mkalinowski-wikia:
We need to postpone removing global data until all wiki data is removed.

### DIFF
--- a/extensions/wikia/Privacy/RemovalAuditLog.php
+++ b/extensions/wikia/Privacy/RemovalAuditLog.php
@@ -10,7 +10,8 @@ class RemovalAuditLog {
 
 	public static function createLog( $userId ) {
 		$db = self::getDb( DB_MASTER );
-		$db->insert( self::LOG_TABLE, ['user_id' => $userId], __METHOD__ );
+		$db->insert( self::LOG_TABLE, [ 'user_id' => $userId ], __METHOD__ );
+
 		return $db->insertId();
 	}
 
@@ -46,6 +47,7 @@ class RemovalAuditLog {
 
 	private static function getDb( $type ) {
 		global $wgSpecialsDB;
+
 		return wfGetDB( $type, [], $wgSpecialsDB );
 	}
 
--- a/extensions/wikia/Privacy/RemovalAuditLog.php
+++ b/extensions/wikia/Privacy/RemovalAuditLog.php
@@ -14,27 +14,34 @@ class RemovalAuditLog {
 		return $db->insertId();
 	}
 
+	public static function getNumberOfWikis( $logId ) {
+		return (int)self::getDb( DB_MASTER )
+			->selectField( self::LOG_TABLE, 'number_of_wikis', [ 'id' => $logId ], __METHOD__ );
+	}
+
 	public static function setNumberOfWikis( $logId, $numberOfWikis ) {
-		self::getDb( DB_MASTER )->update(
-			self::LOG_TABLE,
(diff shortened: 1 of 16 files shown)


slayful (approved):
More or less LGTM, but perhaps @pwojcik86 could take another look?

extensions/wikia/Privacy/RemovalAuditLog.php
```diff
@@ -10,36 +10,74 @@ class RemovalAuditLog {
...
+	 * Mark that global user data was removed for the given log id.
+	 *
+	 * @param $logId
+	 */
+	public static function markGlobalDataRemoved( $logId ) {
+		self::getDb( DB_MASTER )
+			->update( self::LOG_TABLE, [ 'global_data_removed' => true ], [ 'log_id' => $logId ], __METHOD__ );
+	}
+
+	public static function allWikiDataWasRemoved( $logId, $detailsDbType = DB_MASTER ) {
```
slayful: Do we need to have a default here? Can't we always explicitly use one? Or is this some kind of a style in MW?

maintenance/wikia/sql/update-schemas.sh
```diff
@@ -6,5 +6,5 @@ do
 	echo "Updating schema of tables in $DATABASE database ..."
 
 	DB_PARAMS=`dbparams.pl --type slave --name $DATABASE`
-	mysqldump --set-gtid-purged=OFF --no-data $DB_PARAMS | sed 's/\/\*\!50100/\/*/g' | grep --invert-match --regexp='^\/\*\!' | sed 's/AUTO_INCREMENT=[0-9]* //g' > ${DATABASE}-schema.sql
```
slayful: What's happening here?

extensions/wikia/Privacy/RemovalAuditLog.php
```diff
@@ -10,36 +10,74 @@ class RemovalAuditLog {
...
+		$expectedWikis = (int)self::getDb( DB_SLAVE )
+			->selectField( self::LOG_TABLE, 'number_of_wikis', [ 'id' => $logId ], __METHOD__ );
+
+		// since we're using this right after updating a row
```
slayful: Then what? Then we're using primary DB? Not always, since it can be overridden in params, right?

extensions/wikia/Privacy/RemoveUserDataOnWikiTask.php
```diff
@@ -220,28 +251,52 @@ public function removeUserDataOnCurrentWiki( $auditLogId, $userId, $username, $o
...
-		RemovalAuditLog::markTaskAsFinished(
-			$auditLogId,
-			$wgCityId,
-			array_reduce( $results, function( $acc, $res ) { return $acc && $res; }, true ) );
+		RemovalAuditLog::markTaskAsFinished( $auditLogId, $wgCityId,
+			array_reduce( $results, function ( $acc, $res ) { return $acc && $res; }, true ) );
 
 		$this->info( "Removed user data from wiki" );
+
+		if ( RemovalAuditLog::allWikiDataWasRemoved( $auditLogId ) ) {
```
slayful: Would you mind adding context here, as to why are we waiting for all data to be removed?

slayful: Comment would be just fine.

slayful (approved):

PR #16958, opened 2019-04-16
SER-3089 Tests fix

mkalinowski-wikia:


pwojcik86:
Tests will be fixed in #16956

PR #16964, opened 2019-04-18
ADEN-8539 Update ad-engine (48c9c9dc24f79450bca21b9bf834369b3e091664)

dmnsobczak:


extensions/wikia/AdEngine3/src/slots.js
```diff
@@ -274,7 +274,7 @@ export default {
 		slotService.setState('hivi_leaderboard', false);
 		slotService.setState('top_leaderboard', true);
 		slotService.setState('top_boxad', isTopBoxadApplicable());
-		slotService.setState('incontent_boxad_1', true);
+		slotService.setState('incontent_boxad_1', isTopBoxadApplicable());
```
bckatarzyna: As now the function refers to two slots maybe we should call it `isRightRailApplicable` or sth like this

bckatarzyna (changes requested):

extensions/wikia/AdEngine3/src/slots.js
dmnsobczak: Done.

bckatarzyna (approved):

PR #16982, opened 2019-04-25
IW-1959 | enable feeds on non-en communities

xkxd:
@Wikia/iwing


rybmat (approved):

rybmat (commented):
can we also switch all links from /d to /f in this PR? like one in DS api

bkoval:
@rybmat This sounds like a tech debt adjustment we can do later (just to be safe from producing bugs),

rybmat (approved):

PR #16992, opened 2019-04-29
SER-3189 don't use FSFile in GcsBackend

pwojcik86:
`GcsFileBackend` has its own `extensionFromPath`. We don't have to load `FSFile`.


slayful (approved):

bkoval:
🙇

PR #17007, opened 2019-05-06
Configurable consul suffix with datacenter and stage validation, PLAT…

abador:
Based on WIKIA_OVERWRITE_CONSUL_SUFFIX env it checks what stage and dc is used and validates it. After that It overwrites all service.consul sufficex and stage prefixes to the set values.
You can check by setting WIKIA_OVERWRITE_CONSUL_SUFFIX=true SERVER_DBNAME=abador php $PHP_PARAMS /usr/wikia/source/app/maintenance/wikia/HttpsMigration/migrateCheckUser.php

### DIFF
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -217,6 +217,12 @@ $wgMemCachedDebug = ! $wgCommandLineMode; // true unless in command line mode
  * Elementary variables.
  */
 require_once "$IP/includes/wikia/VariablesBase.php";
+/**
+ * Override some of the consul url's.
+ */
+if( $wgForceConsulDatacenter ){
+	require_once "$IP/includes/wikia/VariablesDatacenterOverrides.php";
+}
 
 /**
  * Access credentials from private repository.
--- a/LocalSettings.php
+++ b/LocalSettings.php
@@ -42,6 +42,30 @@ if ( empty( $wgWikiaDatacenter ) ) {
  */
 $wgWikiaEnvironment = getenv( 'WIKIA_ENVIRONMENT' );
 
+/**
+ * Force override consul suffix
+ * @var string $wgOverwriteConsulSuffix
+ */
+$wgOverwriteConsulSuffix = getenv( 'WIKIA_OVERWRITE_CONSUL_SUFFIX' );
+if( $wgOverwriteConsulSuffix ){
+	switch($wgWikiaEnvironment){
+		case 'dev':
+			if( $wgWikiaDatacenter != 'sjc' && $wgWikiaDatacenter != 'poz' ){
+				throw new RuntimeException( 'Combination of $wgWikiaEnvironment & $wgWikiaDatacenter not permitted: ' .
+											$wgWikiaEnvironment . '-' . $wgWikiaDatacenter );
+			}
+			break;
+		case 'prod':
+			if( $wgWikiaDatacenter != 'sjc' && $wgWikiaDatacenter != 'res' ){
+				throw new RuntimeException( 'Combination of $wgWikiaEnvironment & $wgWikiaDatacenter not permitted: ' .
+											$wgWikiaEnvironment . '-' . $wgWikiaDatacenter );
+			}
+			break;
+		default:
+			throw new RuntimeException( 'Unknown $wgWikiaEnvironment: ' . $wgWikiaEnvironment );
+	}
+}
+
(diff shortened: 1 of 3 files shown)


LocalSettings.php
```diff
@@ -42,6 +42,30 @@
  */
 $wgWikiaEnvironment = getenv( 'WIKIA_ENVIRONMENT' );
 
+/**
+ * Force override consul suffix
+ * @var string $wgOverwriteConsulSuffix
+ */
+$wgOverwriteConsulSuffix = getenv( 'WIKIA_OVERWRITE_CONSUL_SUFFIX' );
```
harnash: This name is a bit misleading since it does not override only suffixes but also forces app to use consul tags as well. I would opt for something in tune with `wgForceConsulDatacenter` or similar

LocalSettings.php
```diff
@@ -42,6 +42,30 @@
...
+if( $wgOverwriteConsulSuffix ){
+	switch($wgWikiaEnvironment){
+		case 'dev':
+			if( $wgWikiaDatacenter != 'sjc' && $wgWikiaDatacenter != 'poz' ){
```
harnash: Although I appreciate the effort I do not think that this check is needed. If set improperly the app will fail anyway and it will not safeguard us from all possible/incorrect variations. Moreover when we decide to add new DC (i.e. we might have some apps running in FRA or in the cloud) it will add to the things we need to update in order to deploy the application there. I would keep the config as opaque (from the running ENV) as possible and not double check those values (they should be provided and be correct).

includes/wikia/VariablesBase.php
```diff
@@ -7518,20 +7542,28 @@
  */
 $wgSolrHost = 'prod.search-fulltext.service.consul';
 
+if($wgOverwriteConsulSuffix) {
+	$wgSolrHost = $wgWikiaEnvironment . '.search-fulltext.' . $wgWikiaDatacenter . '.consul';
```
harnash: I'm not sure we do have those SOLR instances on DEV

includes/wikia/VariablesBase.php
```diff
@@ -7518,20 +7542,28 @@
...
+
 /**
  * Solr host for key-value storage for ArticleService.
  * @see includes/wikia/services/ArticleService.class.php
  * @var string $wgSolrKvHost
  */
 $wgSolrKvHost = 'prod.search-kv.service.consul';
 
+if($wgOverwriteConsulSuffix) {
+	$wgSolrKvHost = $wgWikiaEnvironment . '.search-kv.' . $wgWikiaDatacenter . '.consul';
```
harnash: same as above

harnash (changes requested):

includes/wikia/VariablesBase.php
```diff
@@ -6393,6 +6400,10 @@
  */
 $wgPoolCounterServers = [ 'prod.kubernetes-lb-l4.service.consul' ];
 
+if($wgOverwriteConsulSuffix) {
+	$wgPoolCounterServers = [ $wgWikiaEnvironment . '.kubernetes-lb-l4.' . $wgWikiaDatacenter . '.consul' ];
```
harnash:
```suggestion
	$wgPoolCounterServers = [ $wgWikiaEnvironment . '.kubernetes-lb-l4.service.' . $wgWikiaDatacenter . '.consul' ];
```

includes/wikia/VariablesBase.php
```diff
@@ -7510,6 +7530,10 @@
 	'IDHost' => ''
 ];
 
+if($wgOverwriteConsulSuffix) {
+	$wgSMTP['host'] = $wgWikiaEnvironment . '.smtp.' . $wgWikiaDatacenter . '.consul';
```
harnash:
```suggestion
	$wgSMTP['host'] = $wgWikiaEnvironment . '.smtp.service.' . $wgWikiaDatacenter . '.consul';
```

includes/wikia/VariablesBase.php
harnash: Remember about `.service.` part

harnash (approved):

mszabo-wikia (changes requested):
In the config refactor it was laid out that `VariablesBase.php` should only contain simple declarations. Could we perhaps move these overrides into a conditionally included override file?

Grunny (approved):

mszabo-wikia (approved):

PR #17016, opened 2019-05-06
LORE-495 - Send article edits

jamieshepherd:
# Description
For our ML efforts, we want a constant stream of edits of article content sent to a RabbitMQ queue. At the moment we are sending the entire plaintext content _in the message_, which is preferred. We will be monitoring the queue size as this builds to see if it's sustainable, as the ML classifier service will be consuming them as they come in (at a rate of around 3/s on average). I'm keeping the `Don't merge` label until we decide if we want to get this service up ahead of the classifier being completed.

I tried a few approaches to this but ended up going with the simplest implementation, running the preview article content through `strip_tags`, which seems to get us 99% of the way there. I think this will suffice until we need a more nuanced solution.

@Wikia/platform-team 
@Wikia/lore


extensions/wikia/TaxonomyArticleHooks/TaxonomyArticleHooks.class.php
```diff
@@ -0,0 +1,43 @@
...
+			// 'articleWikitext' => $page->getText(),
+			'articlePlaintext' => html_entity_decode( strip_tags( $editInfo->output->getText() ), ENT_COMPAT, 'UTF-8' )
+		];
+
+		self::getRabbitConnection()->publish( self::ROUTING_KEY, $message );
+
+		return true;
+	}
+
+	protected static function getRabbitConnection() {
```
nmonterroso, line 34: minor: the keywords here are flipped compared to the function above (`protected static` vs `static public`).

extensions/wikia/TaxonomyArticleHooks/TaxonomyArticleHooks.php
```diff
@@ -0,0 +1,11 @@
+<?php
+/**
+ * @author Lore team
+ *
+ * Something
```
nmonterroso: :jamie:

includes/wikia/Extensions.php
```diff
@@ -1847,3 +1847,5 @@
     include "$IP/extensions/wikia/TriviaQuizzes/TriviaQuizzes.setup.php";
 }
 
+include_once( "$IP/extensions/wikia/TaxonomyArticleHooks/TaxonomyArticleHooks.php" );
```
nmonterroso: I think we can hide this behind a configuration variable. That way we can merge this before we have the other pieces in, and we can enable locally for testing on `dev`.

includes/wikia/VariablesBase.php
```diff
@@ -8978,3 +8978,13 @@
  * @var string[] $wgTriviaQuizzesEnabledPages
  */
 $wgTriviaQuizzesEnabledPages = [];
+
+/**
+ * ArticleSaveHooks RabbitMQ configuration.
+ * @see extensions/wikia/TaxonomyArticleHooks
+ * @var array $wgTaxonomyArticleExchange
+ */
+$wgTaxonomyArticleExchange = [
```
nmonterroso, line 8994: I wonder if it's better to keep this `TaxonomyArticleHooks.php`. That way we can keep all the extensions configuration in a single place. What do you think?

extensions/wikia/TaxonomyArticleHooks/TaxonomyArticleHooks.class.php
```diff
@@ -0,0 +1,43 @@
...
+		global $wgCityId;
+
+		// We currently do not need wikitext for the classifier, but it may be useful
+		//  to have in the future. Uncomment this line to include wikitext in the message
+		$message = [
+			'wikiId' => $wgCityId,
+			'articleId' => $page->getId(),
+			'articleTitle' => $page->getTitle()->getText(),
+			// 'articleWikitext' => $page->getText(),
+			'articlePlaintext' => html_entity_decode( strip_tags( $editInfo->output->getText() ), ENT_COMPAT, 'UTF-8' )
```
harnash, line 26: I wonder if that is not an overkill for RabbitMQ. Some pages tend to be quite complex and putting the whole content into AMPQ message can cause problems (memory/cpu usage spikes). I would prefer to fetch content asynchronously on the consumer side and not pump all the data through the RabbitMQ

harnash (changes requested):

extensions/wikia/TaxonomyArticleHooks/TaxonomyArticleHooks.class.php
jamieshepherd: It's something I've addressed in the PR. Rabbit can handle messages up to around half a gig, and even complex articles should only be around a few kb. We'll be monitoring the queue and have discussed this implementation with Frank ahead of time. Since messages that go in should be immediately consumed I can't see any immediate problems here.

includes/wikia/Extensions.php
jamieshepherd: That's a good idea.

includes/wikia/VariablesBase.php
jamieshepherd: I think we should stick with convention here, as it looks like most extensions that communicate with Rabbit also store their configuration in the global VariablesBase. As we're now also going to have `wgEnableTaxonomyArticles` we'll have _some_ external config in there anyway.

jamieshepherd:
Edit: We're rethinking how we want to approach the end storage part of this, and will likely be moving the article content out of Rabbit messages now. I'm going to close this PR and open another with a much simpler version (article content omitted).

extensions/wikia/TaxonomyArticleHooks/TaxonomyArticleHooks.class.php
harnash: I'm mildly optimistic about that ;-) It's the volume of those messages that worries me. 

Either way remember to put alerts in place like this one: https://github.com/Wikia/chef-repo/blob/master/cookbooks/kapacitor-18/attributes/platform-core.rb#L33-L34 before you go live on prod.

PR #17025, opened 2019-05-08
IW-1970 | Fix redirect resolution in LinkSuggest

mszabo-wikia:
Fix resolution of redirect target article IDs in LinkSuggest that was preventing users from adding redirected pages as article tags in Feeds.

There were 3 issues at play:
* If there was an exact match for the search query and it was a redirect, it would be returned with the title of the redirect target, but the article ID of the redirect page.
* If there were non-exact matches for the search query that were redirects, they would be appended to the results array instead of being indexed by redirect target ID. This caused PHP to assign an array index to them, essentially resulting in a bogus article ID that could point to a different page or no page at all.
* Cross-namespace redirect target exact matches were not formatted properly.

This PR fixes all those and adds an integration test for these cases. As an added bonus, by fetching the redirect target article ID from the `page` table, we can get rid of the highly inefficient `JOIN` on the `querycache` table that was used to filter out broken redirects, since we can now test for the absence of the redirect target article ID as an indicator of whether the redirect is broken or not. Here are SQL explains for comparison:

**Before**
image

**After**
```
+----+-------------+-----------------+------------+--------+-----------------------------------------------------------+--------------------+---------+---------------------------------------------------------+-------+----------+--------------------------+
| id | select_type | table           | partitions | type   | possible_keys                                             | key                | key_len | ref                                                     | rows  | filtered | Extra                    |
+----+-------------+-----------------+------------+--------+-----------------------------------------------------------+--------------------+---------+---------------------------------------------------------+-------+----------+--------------------------+
|  1 | SIMPLE      | page            | NULL       | ref    | name_title,page_redirect_namespace_len,page_ns_latest_idx | page_ns_latest_idx | 4       | const                                                   | 40964 |   100.00 | Using where              |
|  1 | SIMPLE      | redirect        | NULL       | eq_ref | PRIMARY                                                   | PRIMARY            | 4       | pokemon.page.page_id                                    |     1 |   100.00 | Using where              |
|  1 | SIMPLE      | redirect_target | NULL       | eq_ref | name_title,page_ns_latest_idx                             | name_title         | 261     | pokemon.redirect.rd_namespace,pokemon.redirect.rd_title |     1 |   100.00 | Using where; Using index |
+----+-------------+-----------------+------------+--------+-----------------------------------------------------------+--------------------+---------+---------------------------------------------------------+-------+----------+--------------------------+
```

https://wikia-inc.atlassian.net/browse/IW-1970

### DIFF
--- a/extensions/wikia/LinkSuggest/LinkSuggest.class.php
+++ b/extensions/wikia/LinkSuggest/LinkSuggest.class.php
@@ -256,7 +256,7 @@ class LinkSuggest {
 
 			} else {
 
-				$redirTitleFormatted = self::formatTitle($row->page_namespace, $row->rd_title);
+				$redirTitleFormatted = self::formatTitle( $row->rd_namespace, $row->rd_title );
 
 				// remove any instances of original array's value
 				unset( $results[$row->page_id] );
--- a/extensions/wikia/LinkSuggest/LinkSuggest.class.php
+++ b/extensions/wikia/LinkSuggest/LinkSuggest.class.php
@@ -167,10 +167,10 @@ class LinkSuggest {
 		self::formatResults($db, $res, $query, $redirects, $results, $exactMatchRow, $limit);
 		$sql1Measurement->stop();
 		if (count($namespaces) > 0) {
-			$commaJoinedNamespaces = count($namespaces) > 1 ?  array_shift($namespaces) . ', ' . implode(', ', $namespaces) : $namespaces[0];
+			$commaJoinedNamespaces = count($namespaces) > 1 ?  $db->makeList( $namespaces ) : $db->addQuotes( $namespaces[0] );
 		}
 
-		$pageNamespaceClause = isset($commaJoinedNamespaces) ?  'page_namespace IN (' . $commaJoinedNamespaces . ') AND ' : '';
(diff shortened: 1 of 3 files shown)


extensions/wikia/LinkSuggest/LinkSuggest.class.php
```diff
@@ -341,27 +347,6 @@ static function getEmptyResponse($format) {
...
-	 *
-	 */
-
-	static private function replaceResultIfRedirected(&$results, &$redirects) {
-		foreach($results as &$val){
-			if (isset($redirects[$val])) {
-				$val = $redirects[$val];
-			}
-		}
-	}
```
mszabo-wikia, line 363: This helper did nothing. We already resolve redirects in `formatResults`.

extensions/wikia/LinkSuggest/LinkSuggest.class.php
```diff
@@ -207,20 +203,33 @@ static function getLinkSuggest(WebRequest $request) {
 				$pageTitleLikeClause = "'{$queryLower}%'";
 			}
 
-			// TODO: use $db->select helper method
```
jsutterfield, line 210: This TODO has been here for 7 years...It probably thought it was long forgotten. Well done addressing it :)

jsutterfield (approved):
LGTM. Great PR description and kudos for adding the test!

rybmat (approved):

PR #17033, opened 2019-05-08
LORE-519 - Implement ArticleExporter API

jamieshepherd:
# Description
For our ML classification we want to be able to get plaintext version of article content, along with some associated metadata. Similar functionality is available in `WikiaSearchIndexer`, but for large pages it's especially slow. This is a fast, not-so-thorough API that exports only the fields we need for our ML classifier.

# Example
https://gameofthrones.jshepherd.fandom-dev.us/wikia.php?controller=ArticleExporterApi&method=getArticles&ids[]=201

@Wikia/lore

PR #17034, opened 2019-05-08
LORE-495 - Send article updates to RabbitMQ

jamieshepherd:
# Description
- Using a similar structure and under the same extension as https://github.com/Wikia/app/pull/17033
- This PR takes article saves and pushes them to a queue for our ML classifier to consume

@Wikia/lore


extensions/wikia/ArticleExporter/ArticleExporter.setup.php
```diff
@@ -0,0 +1,14 @@
...
+ * @author Lore team
+ */
+
+$dir = __DIR__ . '/';
+
+// Autoload
+$wgAutoloadClasses['ArticleExporterHooks'] = dirname( __FILE__ ) . "/ArticleExporterHooks.class.php";
+
+// Hooks
+$wgHooks['ArticleSaveComplete'][] = 'ArticleExporterHooks::onArticleSaveComplete';
```
nmonterroso: I wonder if we should gate this hook behind a global so that we can turn it off without affecting the API. What do you think?

extensions/wikia/ArticleExporter/ArticleExporterHooks.class.php
```diff
@@ -0,0 +1,38 @@
...
+use Wikia\Rabbit\ConnectionBase;
+
+/**
+ * @author Lore team
+ *
+ * Send plaintext versions of articles to RabbitMQ as they are edited on Fandom
+ */
+
+class ArticleExporterHooks {
+    const ROUTING_KEY = "taxonomy.article-edits";
```
nmonterroso: ~~The rest of the queues we've been using have been singular. Can we change this?~~

minor: it'd be nice if this was configureable so that like the rest of our flows we can choose to have a personalized queue here. What do you think?

drsnyder:
Can we add some tests?

jamieshepherd:
I've added tests to the API PR, but I think they're a little redundant here since the code is really just listening for a hook and calling our rabbit lib - which I _assume_ are already battle tested.

extensions/wikia/ArticleExporter/ArticleExporterHooks.class.php
```diff
@@ -0,0 +1,37 @@
...
+
+        return true;
+    }
+
+    protected static function getRabbitConnection() {
+        global $wgArticleExporterExchange;
+
+        if ( !isset( self::$rabbitConnection ) ) {
+            self::$rabbitConnection = new ConnectionBase( $wgArticleExporterExchange );
+        }
```
drsnyder, line 42: Can this be injected? I thought we had some form of dependency injection in MW...

extensions/wikia/ArticleExporter/ArticleExporterHooks.class.php
```diff
@@ -0,0 +1,37 @@
...
+    public static function onArticleSaveComplete( WikiPage $page ) {
+        global $wgCityId;
+        global $wgArticleExporterExchange;
+
+        $message = [
+            'wikiId' => $wgCityId,
+            'pageIds' => [ strval($page->getId()) ],
+        ];
+
+        self::getRabbitConnection()->publish( $wgArticleExporterExchange['routing'], $message );
```
drsnyder: Are there any namespaces we should ignore?

extensions/wikia/ArticleExporter/ArticleExporterHooks.class.php
nmonterroso: We do, but I'm not sure anyone uses it :(

With the unified platform I could see that being one bit that goes away, maybe.

jamieshepherd: Yeah this is generally the way we've been doing things (at least from all of the extensions I've seen).

extensions/wikia/ArticleExporter/ArticleExporterHooks.class.php
jamieshepherd: Could you clarify this comment? Thanks

drsnyder: I _think_ we probably only care about namespace 0 but I could be wrong. This [page](https://www.mediawiki.org/wiki/Manual:Namespace) has the list of all of the namespaces.

For now, we can probably safely ignore the category, talk, file, and user pages?

drsnyder: It might also be ok to ignore them downstream but if we're concerned about message volume we may want to consider filtering here. I don't have any insight into what kind of volume those pages generate. Generally, I would guess it's pretty low except for maybe the talk pages on some communities?

nmonterroso: I agree. I think we should do both -- ignore non `0` namespaces and include the namespace for each page. That might cause us to reconsider the payload to something like:
```json
{
  "wikiId": "...",
  "pages": [{"pageId": "...", "namespace": 0}]
}
```

If we choose to do that, I can make the corresponding change in Wikia/pandora#6427 to take in the new structure and ignore non-`0` namespaces. What do you think @drsnyder @jamieshepherd ?

jamieshepherd: That's a good callout @drsnyder - updated in ef76cd7 - also updated the payload with your suggestion @nmonterroso

drsnyder: Sounds good.

nmonterroso (approved):

PR #17038, opened 2019-05-09
IW-2000 | Update Oasis recommendations to handle community-specific recommendations

mszabo-wikia:
Update Oasis recommendations to handle community-specific recommendations from the `wiki-recommendations` service. If we have community-specific recommendations available, we will only choose from them based on provided weights, and discard other recommendations that are not specifically targeted to the current community.

https://wikia-inc.atlassian.net/browse/IW-2000


extensions/wikia/Recirculation/js/helpers/sponsoredContent.js
```diff
@@ -38,9 +38,19 @@ define('ext.wikia.recirculation.helpers.sponsoredContent', [
 	}
 
 	function getApplicableContent(sponsoredContent) {
-		return sponsoredContent.filter(function (el) {
-			return !el.geos.length || el.geos.indexOf(userGeo) !== -1;
+		var applicableContent = sponsoredContent.filter(function (el) {
+			return (!el.geos.length || el.geos.indexOf(userGeo) !== -1) && (!el.wikiIds.length || el.wikiIds.indexOf(w.wgCityId) !== -1);
```
xkxd: considering the repetition, I'd write it like
```
var filterBy = function(arr, searchedItem) {
    return arr.filter(function(el) {
        return (!arr.length || arr.indexOf(searchedItem) !== -1)
    })
};

var contentByGeo = filterBy(sponsoredContent, userGeo);
var contentByWikiId = filterBy(sponsoredContent, w.wgCityId);
var contentByWikiIdAndGeo = filterBy(contentByGeo, w.wgCityId);

if (contentByWikiIdAndGeo.length) {
    return contentByWikiIdAndGeo;
}

if (contentByWikiId.length) {
    return contentByWikiId;
}

return contentByGeo;
```
bc currently 
- `applicableContent` is imo always targeted for specific wiki, so `siteSpecificContent` == `applicableContent` 
- `if (siteSpecificContent) {` will be `true` even for `[]`
- we don't support "wikiId without any geo" scenario

xkxd (changes requested):

extensions/wikia/Recirculation/js/helpers/sponsoredContent.js
rybmat: what about the case when there is no content with given geo nor with given wiki ID and we should fallback to items that do not have targeting specified at all?

mszabo-wikia: it seems that the existing logic won't return a sponsored item if none match the current geo.

rybmat: i think we tested that and it worked, but anyways, we should always fall back to item that is not targeted at all if none better match is found

xkxd: @rybmat 
true, we'll need to change/add
```
if (contentByGeo.length) {
    return contentByGeo;
}

return sponsoredContent;
```
to the code I've added above

extensions/wikia/Recirculation/js/helpers/sponsoredContent.js
```diff
@@ -38,9 +39,28 @@ define('ext.wikia.recirculation.helpers.sponsoredContent', [
 	}
 
 	function getApplicableContent(sponsoredContent) {
-		return sponsoredContent.filter(function (el) {
-			return !el.geos.length || el.geos.indexOf(userGeo) !== -1;
+		var geoSpecificContent = sponsoredContent.filter(function (el) {
```
xkxd: I'm still confused with this logic. We have "geoSpecific" items here which are filtered by both geo and wikiId.

extensions/wikia/Recirculation/js/helpers/sponsoredContent.js
```diff
@@ -38,9 +39,28 @@ define('ext.wikia.recirculation.helpers.sponsoredContent', [
...
+			return (!el.geos.length || el.geos.indexOf(userGeo) !== -1) && (!el.wikiIds.length || el.wikiIds.indexOf(w.wgCityId) !== -1);
 		});
+		var siteSpecificContent = sponsoredContent.filter(function (item) {
```
xkxd: here we have `item` but in first func `el` is defined. this can be unified.

~also - shouldn't we have `item.wikiIds.length` check here as well?~ nah, we shouldn't

extensions/wikia/Recirculation/js/helpers/sponsoredContent.js
```diff
@@ -38,9 +39,28 @@ define('ext.wikia.recirculation.helpers.sponsoredContent', [
...
+			return item.wikiIds.indexOf(w.wgCityId) !== -1;
+		});
+
+		var geoAndSiteSpecificContent = siteSpecificContent.filter(function (item) {
```
xkxd: I think this could be done without `some`, by filtering those items against `userGeo`

xkxd (changes requested):

rybmat (approved):

xkxd (approved):

xkxd (approved):

PR #17045, opened 2019-05-13
Use HOSTNAME_OVERRIDE when loading config

themech:
We're using `HOSTNAME_OVERRIDE` to override the host name for example when running in docker. Take this into account it when loading devbox-specific config.


Grunny (approved):

nmonterroso (approved):
Awesome, thanks!

PR #17056, opened 2019-05-16
SER-3198 Use unified-search in SearchApiController

mkalinowski-wikia:


extensions/wikia/Search/SearchApiController.class.php
```diff
@@ -128,21 +132,55 @@ protected function setResponseFromConfig( Wikia\Search\Config $searchConfig ) {
...
+			'batches' => $result->getNumPages(),
+			'currentBatch' => $result->getPage(),
+			'next' => $result->getPage() * $config->getLimit() + 1,
+			'total' => $result->getResultsFound(),
+			'items' => $result->toArray( [
+				'pageid' => 'id',
+				'title',
+				'url',
+				'ns',
+				'article_quality_i' => 'quality',
```
slayful: Do we want to map article_quality if we don't support it?

slayful (changes requested):
Tests aren't passing and article quality is not supported (and likely never used).

slayful (approved):

PR #17058, opened 2019-05-16
PLATFORM-4090| temporarily increase cpulimit for scribunto on runescape wiki

harnash:
Increasing the cpuLimit from `7` to `15` (temporarily)


includes/wikia/Extensions.php
```diff
@@ -428,6 +428,10 @@
 	$wgScribuntoUseGeSHi = $wgEnableSyntaxHighlightGeSHiExt;
 	$wgWysiwygDisabledNamespaces[] = NS_MODULE;
 
+	if ( $wgServer === 'runescape' ) {
```
Grunny: `wgServer` is the domain, so for Runescape it would be `http://runescape.fandom.com` and due to language paths, `wgServer` is the same for all language versions. So I suggest changing this to:
```suggestion
	// Change limit on Runescape temporarily (PLATFORM-4090)
	if ( $wgCityId == 304 ) {
```

Grunny (changes requested):

includes/wikia/Extensions.php
harnash: Yeah. Obviously. I must have been drunk when I did this change :(

Grunny (approved):

PR #17061, opened 2019-05-16
ADEN-8794 Fix resource loaders

fraszczakszymon:


fraszczakszymon:
"Better ask for forgiveness than permission" 🙈
Merging because of losing $$$

PR #17064, opened 2019-05-17
ADEN-8792 Clean AnalyticsEngine and leftovers of AE2

dmnsobczak:


extensions/wikia/ArticleVideo/scripts/featured-video.jwplayer.autoplay.js
```diff
@@ -1,18 +1,13 @@
 define('wikia.articleVideo.featuredVideo.autoplay', [
 	'wikia.abTest',
 	'wikia.articleVideo.featuredVideo.cookies',
-	require.optional('ext.wikia.adEngine.ml.billTheLizardExecutor'),
 	require.optional('ext.wikia.adEngine3.api')
-], function (abTest, featuredVideoCookieService, billTheLizardExecutor, adsApi) {
+], function (abTest, featuredVideoCookieService, adsApi) {
```
nandy-andy, line 5: Why don't we need Bill here anymore? We don't have any active models for autoplay in the service?

nandy-andy (approved):

extensions/wikia/ArticleVideo/scripts/featured-video.jwplayer.autoplay.js
dmnsobczak: Autoplay models are not available for AE3 on Oasis. That's the missing part we need to reimplement:
https://github.com/Wikia/app/blob/b85b96899ec1d0d597e3ecd4971b8430319ea60f/extensions/wikia/AdEngine3/src/ml/executor.js#L4

PR #17067, opened 2019-05-19
Update QuickToolsController.class.php

LunarTwilight:
This change makes it so that when you use QuickTools to block a user, talk page access is removed.


LunarTwilight:
(This change makes it so that when you use QuickTools to block a user, talk page access is removed, and you watch their contribs.)

LunarTwilight:
Per S:C, this would be better done as an option.

PR #17074, opened 2019-05-21
Platform 4088

abador:
When staff changes usernames ignore phalanx and spoof checks. Show warning before confirm

### DIFF
--- a/extensions/wikia/UserRenameTool/RenameUserProcess.class.php
+++ b/extensions/wikia/UserRenameTool/RenameUserProcess.class.php
@@ -200,7 +200,7 @@ class RenameUserProcess {
 		if( count( $errors ) > 0 && $stringErrors ){
 			$stringErr = [];
 			foreach( $errors as $err ){
-				$stringErr[] = $err->toString();
+				$stringErr[] = (string)$err;
 			}
 			return $stringErr;
 		}
--- a/extensions/wikia/UserRenameTool/RenameUserProcess.class.php
+++ b/extensions/wikia/UserRenameTool/RenameUserProcess.class.php
@@ -153,21 +153,14 @@ class RenameUserProcess {
 	 *
 	 * @return mixed false if all prerequisites are met else list of errors
 	 */
-	public function testSpoofAndPhalanx( $stringErrors = false ) {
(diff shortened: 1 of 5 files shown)


Grunny (commented):
I believe this completely skips the checks for staff users, right?

As I understand it, the goal was to still surface these errors to staff users but allow them to override it. So, I would say how this would be have is converting these to `addWarning` for staff users, but leave them as errors for self-renames, so they still see the warning and then can make a decision to confirm before proceeding (check `mActionConfirmed`) so they at least know there's a potential issue.

abador:
> I believe this completely skips the checks for staff users, right?
> 
> As I understand it, the goal was to still surface these errors to staff users but allow them to override it. So, I would say how this would be have is converting these to `addWarning` for staff users, but leave them as errors for self-renames, so they still see the warning and then can make a decision to confirm before proceeding (check `mActionConfirmed`) so they at least know there's a potential issue.

It disables the check for staff, but still it returns the errors and shows them in the confirm modal. When the confirm modal is shown it overlays the page, so I've added those errors in userrenametool-confirm-intro-staff

extensions/wikia/UserRenameTool/RenameUserProcess.class.php
```diff
@@ -174,32 +175,67 @@ public function setup() {
...
-			$this->addError( $error );
-			return false;
+		$err = self::testBlock( $nun );
+		if ( !empty( $err ) ) {
+			$errors[] = $err;
+		}
+		if( count( $errors ) > 0 && $stringErrors ){
+			$stringErr = [];
+			foreach( $errors as $err ){
+				$stringErr[] = (string)$err;
```
Grunny: Casting to string here shouldn't be needed as long as the `wfMessage` calls explicitly state their output format (otherwise `parse` is assumed). And as a best practice, we should always specify an explicit output format with `wfMessage`.

Grunny (changes requested):
Ah, I misread that part. :)

Minor comment, and needs some coding conventions cleanup throughout, otherwise looks good.

Grunny (approved):

Grunny (approved):

PR #17080, opened 2019-05-22
CAKE-4855 | bump content-types library and add theming

vforge:
@Wikia/cake


extensions/wikia/TriviaQuizzes/TriviaQuizzes.setup.php
```diff
@@ -1,24 +1,25 @@
 <?php
-$wgExtensionCredits['other'][] = [
+$wgExtensionCredits[ 'other' ][] = [
```
mklucsarits: We don't use spaces inside arrays in PHP, only inside parentheses.

mklucsarits: I should clarify, when assigning to an existing array, the spaces aren't used. I've looked in other Wikia extensions and it's the same.

vforge: I've seen both in different files (especially `.setup.php` are very inconsistent). I went with your suggestion.

extensions/wikia/TriviaQuizzes/TriviaQuizzesHooks.class.php
```diff
@@ -1,13 +1,24 @@
...
 		}
 
+		wfProfileOut( __METHOD__ );
+		return true;
+	}
+
+	public static function onWikiaSkinTopScripts( Array &$vars, &$scripts ) {
+		wfProfileIn( __METHOD__ );
+		$td = (new ThemeSettings())->getSettings();
+		$vars['wgTriviaQuizzesPrimaryColor'] = $td['color-community-header'];
```
mklucsarits, line 19: This is probably fine, but I'm wondering if there's another naming syntax for these kinds of variables that aren't really wikia global (`$wgVariable`) types but are sort of faked for the front-end. It probably doesn't matter though.

mklucsarits (approved):

PR #17098, opened 2019-05-30
Platform 4100

abador:


docker/devbox/nginx_wrapper.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+mkfifo /var/log/nginx/std.log
+nginx -g "daemon off;" > /var/log/nginx/std.log
```
harnash, line 3: Can't it be achievable through: https://nginx.org/en/docs/http/ngx_http_log_module.html#access_log? This way we could get rid of this wrapper (`mkfifo` could be on the nginx dockerfile level?)

docker/devbox/prometheus-nginxlog-exporter.hcl
```diff
@@ -0,0 +1,18 @@
...
+}
+
+
+namespace "mediawiki_nginx_devbox" {
+  format = "{ \"appname\": \"mediawiki-nginx-access-logs\", \"time_local\": \"$time_local\", \"remote_addr\": \"$http_fastly_client_ip\", \"remote_user\": \"$remote_user\", \"method\": \"$request_method\", \"request\": \"$request_uri\", \"status\": $status, \"body_bytes_sent\": $body_bytes_sent, \"request_time\": $request_time, \"http_host\": \"$http_x_original_host\", \"http_referrer\": \"$http_referer\", \"http_user_agent\": \"$http_user_agent\" }"
+  source_files = [
+    "/var/log/nginx/std.log"
+  ]
+  labels {
+    app = "mediawiki_nginx_devbox"
```
harnash:
```suggestion
    app = "nginx_"
```
should be enough - more detailed description should be in the labels. Imagine that you want to pull all the metrics from the NGINX instances and having them the same metric makes more sense, you can filter them later using tags/labels.

docker/prod/Dockerfile-nginxexporter
```diff
@@ -0,0 +1,6 @@
+# This is a base nginx image with Wikia's directory structure
+FROM quay.io/martinhelmich/prometheus-nginxlog-exporter
+
+COPY ./prometheus-nginxlog-exporter.hcl /etc/prometheus-nginxlog-exporter.hcl
```
harnash: We should be able to mount the configuration file directly using method like this:
https://github.com/martin-helmich/prometheus-nginxlog-exporter#docker
This will allow us to use the original image and limit the work needed to get this running.

```bash
$ docker run --name nginx-exporter -p 4040:4040 -v logs:/mnt/nginxlogs -v /path/to/config.hcl:/etc/prometheus-nginxlog-exporter.hcl quay.io/martinhelmich/prometheus-nginxlog-exporter -config-file /etc/prometheus-nginxlog-exporter.hcl
```

docker/prod/prod.template.yaml
```diff
@@ -121,6 +124,29 @@ spec:
             requests:
               cpu: 50m
               memory: 50Mi
+        - name: exporter
+          image: artifactory.wikia-inc.com/platform/nginx-exporter-prod:${IMAGE_TAG}
+          livenessProbe:
+            httpGet:
+              path: /metrics
```
harnash, line 152: that's a hack ;-)

docker/prod/prod.template.yaml
```diff
@@ -370,8 +400,29 @@ spec:
       type: mediawiki
   namespaceSelector:
     matchNames:
-    - prod
+      - prod
```
harnash: indentation is messed up here

docker/prod/prod.template.yaml
```diff
@@ -370,8 +400,29 @@ spec:
...
   endpoints:
-  - port: main
-    path: /metrics.php
-    interval: 10s
+    - port: main
+      path: /metrics.php
+      interval: 10s
+---
+apiVersion: monitoring.coreos.com/v1
```
harnash: There is already one `ServiceMonitor` defined above - you should probably add a new entry there instead of creating a new one

harnash (changes requested):

harnash:
I was also thinking if we could use https://github.com/fluent/fluent-plugin-prometheus so we do not introduce one more tool to the stack but use something more flexible? Would it meet our needs in terms of nginx logs->prom metrics? (see: https://devhub.io/repos/kazegusuri-fluent-plugin-prometheus)

Also I believe OPS are already using some tool to scrape metrics: https://github.com/discordianfish/nginx_exporter and there is https://github.com/vozlt/nginx-module-vts

docker/devbox/nginx_wrapper.sh
abador: So this one of the things I've been struggling with. It seems that nginx docker is set up to write to /dev/stdout, so on nginx you have a symlink /var/log/nginx/access.log => /dev/stdout (error log is similar). The problem with that setup caused the exporter to throw a panic(when log writing stops for some reason the app tries to seek in that file, but you can't seek in pipe). This solution blocks the pipe, so the app waits for the pipe instead of seeking the new line after write is finished. I've added this mkfifo here, because it all works on mounted volumes. The lib used to doing this is quite stable, and I couldn't get it to work the other way:/

docker/prod/prod.template.yaml
abador: Yes but It beats adding code to the app ;)

harnash (approved):

PR #17099, opened 2019-05-30
SER-3241 | Remove HD option from searching

slayful:


slayful:
The HD Only option is buggy (checkbox not always stays selected) and may not work. The removal should be done due to same issues as  - option is not much used.

mkalinowski-wikia (approved):

PR #17110, opened 2019-05-31
IW | Allow to use Host header in internal requests

mszabo-wikia:
This aims to make the life of clients trying to call MW from inside the k8s cluster easier. It makes it possible to omit `X-Original-Host` header when calling MW, so redirect handling becomes automatic.

Changes made:
* Introduce and use `wgEnvironmentDomainMappings` that defines a map of env-specific domains to env-agnostic equivalents.
* Allow MW to handle sandbox etc. URLs properly without relying on VCL's normalization.
* Remove support to get URLs for non-current environment from `WikiFactory::getLocalEnvUrl` —this functionality was only used in 1 place, where it can be replaced by `wfNormalizeHost`
* Make NGINX use host from URL if `X-Original-Host` is absent.
* Remove unused `wgServicesExternalAlternativeDomain` setting.


docker/base/base.inc
```diff
@@ -1,6 +1,12 @@
 listen 8080;
 root /usr/wikia/slot1/current/src;
 
+set $fastcgi_server_name $host;
+
+if ($http_x_original_host) {
+	set $fastcgi_server_name $http_x_original_host;
```
themech, line 7: does it mean we should to Vary the response on X-Original-Host?

mszabo-wikia: I don't think we need to make such a change, as the part inside the if block is the current behavior. All MW requests coming into the cluster have `X-Original-Host` header set already.

themech: 👍

docker/prod/site.conf.template
```diff
@@ -17,10 +17,11 @@ server {
         # make MediaWiki think we're running on a standard port
         # otherwise we end up with redirects to port 8080
         fastcgi_param SERVER_PORT 80;
-        # use the original request host so MW can identify the specific wiki
-        # see /cookbooks/varnish4/templates/default/control_stage_deliver.vcl.erb in chef-repo
-        fastcgi_param HTTP_HOST $http_x_original_host;
-        fastcgi_param SERVER_NAME $http_x_original_host;
+
+		# use the original request host so MW can identify the specific wiki
```
Grunny, line 21: Minor: we should keep whitespace consistent in these nginx configs.

Grunny (approved):

PR #17116, opened 2019-06-03
Increase NGINX FastCGI buffers size

mszabo-wikia:
Increase NGINX FastCGI buffers size from defaults—based on sandbox testing, these values provide a more consistent and better performance.

### DIFF
--- a/docker/base/nginx.conf
+++ b/docker/base/nginx.conf
@@ -22,6 +22,9 @@ http {
     fastcgi_connect_timeout 3s;
     fastcgi_send_timeout 10s;
     fastcgi_next_upstream off; # do not retry failed requests, CDN handles that for us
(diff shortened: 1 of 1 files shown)


docker/base/nginx.conf
```diff
@@ -22,6 +22,9 @@ http {
     fastcgi_connect_timeout 3s;
     fastcgi_send_timeout 10s;
     fastcgi_next_upstream off; # do not retry failed requests, CDN handles that for us
+	fastcgi_buffer_size 16k;
+	fastcgi_buffers 32 16k;
+	fastcgi_busy_buffers_size 32k;
```
Grunny: Minor: Whitespace should be consistent.

Grunny (approved):

PR #17130, opened 2019-06-05
ADEN-8570 Add Confiant integration

dmnsobczak:


extensions/wikia/AdEngine3/src/ad-context.js
```diff
@@ -464,6 +464,12 @@ export default {
 			parameters: {},
 			timeout: 2000,
 		},
+		confiant: {
```
nandy-andy, line 467: Should it be exactly the same as on mobile?

dmnsobczak: I asked Nef in ticket about this and we're planning to move QA of this task to Confiant, so they will verify that and we will change those values if required :)

extensions/wikia/AdEngine3/src/setup.js
```diff
@@ -108,6 +108,7 @@ function setupAdContext(wikiContext, isOptedIn = false, geoRequiresConsent = tru
 		isGeoEnabled('wgAdDriverUnstickHiViLeaderboardAfterTimeoutCountries')
 	);
 
+	context.set('services.confiant.enabled', isGeoEnabled('wgAdDriverConfiantDesktopCountries'));
```
jacekbj, line 111: `isGeoEnabled('wgAdDriverConfiantCountries') && isGeoEnabled('wgAdDriverConfiantDesktopCountries')` to be consistent with mobile?

dmnsobczak: No, because `wgAdDriverConfiantCountries` is a copy of mobile wgVar. By doing this we can accidentally enable it on desktop.

jacekbj: Talked over offline. `wgAdDriverConfiantCountries` will be removed after rollout.

jacekbj (approved):

PR #17135, opened 2019-06-06
PLATFORM-4083 Resolve custom JS import methods to fandom domain

Michal-Chatlas1221:


skins/common/wikibits.js
```diff
@@ -728,7 +728,7 @@ if (skin != 'monaco' && skin != 'oasis') {
 window.importScriptPage = function( page, server ) {
 	var url = '/index.php?title=' + encodeURIComponent(page.replace(/ /g,'_')).replace('%2F','/').replace('%3A',':') + '&action=raw&ctype=text/javascript';
 	if( typeof server == "string" ) {
-		if( server.indexOf( '://' ) == -1  && server.substring( 0, 2 ) !== '//' ) url = 'http://' + server + '.' + mw.config.get('wgWikiaBaseDomain') + url;
+		if( server.indexOf( '://' ) == -1  && server.substring( 0, 2 ) !== '//' ) url = 'http://' + server + '.' + mw.config.get('wgFandomBaseDomain') + url;
```
Grunny: It won't change the functionality much since `importScriptURI` would have made it protocol-relative if it is plain HTTP, but we can change these to use `https` now while changing them to us `fandom.com`.

Grunny (changes requested):
Minor suggestion, otherwise looks good. :)

skins/common/wikibits.js
Michal-Chatlas1221: updated

Grunny (approved):

PR #17155, opened 2019-06-10
SER-3242 Regex fix for Phalanx

mkalinowski-wikia:


slayful (approved):
Tests should be fixed.
This is good enough to resolve the issue, we need to fix this properly in the future though.

